PR #38, opened 2022-11-26
refactoring frontend (add eslint, absolute import)

gangjun06:
## Absolute Import
Add jsconfig.json file and setup paths for absolute import.
Ex) `../../../components` -> `~/components`
## eslint
Add eslint script and eslint plugins to package.json and extends `next/core-web-vitals` in .eslintrc.json.
Using eslint-simple-import-sort to sort imports.
## Etc
I'm sorry for suddenly creating PR without any issues.
+ Also, I'm wondering that uses tab instead of two space. I think two space looks nice then tab in JSX

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,3 +1,6 @@
 {
-  "extends": "next/core-web-vitals"
+  "extends": "next/core-web-vitals",
+  "rules": {
+    "react-hooks/exhaustive-deps": "off"
+  }
 }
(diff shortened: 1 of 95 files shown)


mv-turtle (approved):
This is really great! Thank you so much for your help! Really appreciate it!

mv-turtle:
Thank you again! Just add an issue next time, if possible. 

Also, I think 2 spaces instead of tabs could be a good idea. Feel free to send a PR if you want to.

PR #45, opened 2022-11-27
Edit frontend .prettierrc config

gangjun06:
## Edit .prettierrc
tab -> two space for legibility. https://github.com/Infisical/infisical/issues/38#issuecomment-1328066415

## Update absolute import
Add `~/const` as import path. `../../../const` -> `~/const`


mv-turtle (approved):
Looks good! Thank you so much @gangjun06!

Merged!

Issue #39, opened 2022-11-26
Localize Web UI


gangjun06:
### Feature description
Add localizing package to frontend. Such as [next-translate](https://github.com/aralroca/next-translate) [next-i18next](https://github.com/i18next/next-i18next).

### Why would it be useful?
People in many countries will be able to use the web UI more convenient.

### Additional context

mv-turtle:
This is an awesome idea! Would you be able to help with Korean?

gangjun06:
Of course! I can translate english into korean. And if you want, I will create PR for apply i18 to frontend project.

mv-turtle:
Yes, this is great! Thank you so much for your help!

mv-turtle:
Also, if you want, we have a community Slack channel for our contributors. We would love to have you join!

Here is the link: https://join.slack.com/t/infisical-users/shared_invite/zt-1kdbk07ro-RtoyEt_9E~fyzGo_xQYP6g

PR #46, opened 2022-11-27
Localize Web UI #39 

gangjun06:
## Check Lists
- [x] Setup localization module
- [x] Move texts to json file
- [x] Translate english into korean
- [x] Resolve Conflict
- [x] Language select in Login page and setting page

### package.json
- `next-translate` for support multi languages.
### Locale files in `frontend/locales`
All language files are located under `frontend/locales/{language}.json`.
Language file is like this;
```json
{
  "namespace": {
     "key": "value",
     "nested-key": {
       "key": "nested-value"
    }
  }
}
```
### Texts in `frontend/pages/` and `frontend/components`
```js
import useTranslation from "next-translate/useTranslation";
......
 const { t } = useTranslation("");
 ......
 <div>{t("common:hello")}</div>
 ......
 ```
 ### Define namespaces and locales in `frontend/i18n.js`
 Defines the namespaces to use per page
### docker-compose.dev.yml
https://github.com/gangjun06/infisical/blob/4c94ddd1b2635143ef93c941962c2bd29dfdd828/docker-compose.dev.yml#L52-53
To change language files without recreate full container with just restart, Added this lines.
## Translate key convention?
I didn't make it clearly, but i followed some rules.
- use `-` for space in key.
- seperate namespace by page and section.
- make keys simple as possible
## ETC
For vscode, i18n-ally is very useful plugin for translated file.
It can be able to preview value of the translate key.

### CODE REVIEW

**Comment by gangjun06** [comment]
I'm sorry that it is taking a long time.
I'll make it mergeable without conflict in a few days.

**Comment by mv-turtle** [comment]
Hi @gangjun06! Don't worry about it! Just take as much time as you need! We really appreciate your contributions. Thank you.

**Comment by mv-turtle** [comment]
HI @gangjun06! This is really awesome! I'm still looking into this, it's a lot of code 😃 

Do you think you could fix the merge conflicts?

**Comment by gangjun06** [comment]
I fixed conflicts 5910bfbb4d5aaa3e2bb219b6674f07559dc2f80b

File: frontend/next.config.js [hunk 1]
```diff
@@ -49,7 +50,7 @@ const securityHeaders = [
   },
 ];
 
-module.exports = {
+module.exports = nextTranslate({
```
mv-turtle, line 53: Thank you so much! I think it's really great! 

I'm a little bit bit concerned with this package. It doesn't have that many downloads. Do you think it is a potential security vulnerability that you wrap this whole thing in nextTranslate? 

I'm looking at some other projects. It seems like they are using next-i18next. For example, cal.com (https://github.com/calcom/cal.com/blob/main/apps/web/next.config.js). It seems like it has 3 times more downloads. 

Are there any reasons why you chose next-translate over the other one?

mv-turtle: Other than that it looks awesome! 🔥🚀

gangjun06: Sorry, actually I haven't thought about security problem 🥲.
I used next-translate because  I used this in other project and next-translate was more convenient to use and.
For example, we don't need to write getStaticSideProps or getServerSideProps in every pages.
Also, I think nextTranslate source code, https://github.com/aralroca/next-translate/blob/master/src/plugin/index.ts looks just normal next.js plugin.
But I don't have any security background and If you worry, I can change next-translate into next-i18next.

mv-turtle: Yeah, I'm really sorry for asking you to change it... 😭 If you think you could switch it to next-i18next, this would be awesome! Do you think it would take you a lot of time? 

I think we just need to be very cautious because of how sensitive the data Infisical stores is. If there is a security vulnerability, this could lead to many people losing their secrets and data. So i'm a bit worried that no big projects are using next-translate, right? With next-i18next, I found that Cal.com is using them which is why it seems more credible. Super sorry again 😭

mv-turtle: Hi @gangjun06! Just wanted to check in. Let me know if you need any help at all! :)

gangjun06: @mv-turtle  Oops! Sorry for big late.
I didn't push after change library and I didn't care about it for a week.
Also there're some error apply library.
I'll fix them and make mergeable ASAP!

mv-turtle: Thank you so much @gangjun06! ❤️

gangjun06:
Updated module to next-i18next and resolved all merge conflicts.

# Some Notes
## Apply translation to page
Next-i18next needs getServersideProps for all pages.
So more convenient, I created some utility on `frontend/components/utilities/withTranslateProps.ts`
You can use it like this:
```tsx
import { getTranslatedServerSideProps } from "~/utilities/withTranslateProps";
export const getServerSideProps = getTranslatedServerSideProps(["dashboard"]);
---
import { getTranslatedStaticProps } from "~/components/utilities/withTranslateProps";
export const getStaticProps = getTranslatedStaticProps(["auth", "login"]);
```

## Too many locale files
I couldn't find good way for manage namespaces in one file. So, I just created lots of file in locales/{language}.
Maybe make all locale texts to same namespace will be better way. (But I created multiple namespaces because next-i18next doesn't cache locale text, just load all texts in every time page load.)

## How language change works?
I believe next-i18next doesn't have language change function. It's just working with dynamic locale path. (/en/... , /ko/.......)
I can't sure it's good way, Because I didn't looked up much about this library. I saved locale on the localStorage, and if the URL doesn't match with current locale, It will be redirect to correct url by useEffect in _app. 

Currently, this PR doesn't cover all texts. If no one contributes to this, I'll create new PR later when I am free(Maybe Jan 3rd week~).

mv-turtle:
Sounds good! Thank you so much, I will have a look soon! 

Quick question. If a translation is not available, does the text default to English?

gangjun06:
Yes, default locale is English.
/login -> English
/en/login -> English

mv-turtle:
Do you maybe know why this error occurs on all pages? Trying to look into this right now

Screenshot 2022-12-26 at 16 15 18

gangjun06:
> Do you maybe know why this error occurs on all pages? Trying to look into this right now
> 
> Screenshot 2022-12-26 at 16 15 18
Oooops It was my mistake... I fixed it by using router.asPath instead of router.pathname in _app.js.

mv-turtle:
I think all of this makes sense! Thank you so much for the contribution ❤️
Would you be able to help fix the merge conflicts? there seem to be quite a lot :)

mv-turtle:
I solved all the merge conflicts! Merging now :)
Thank you so much for the contribution!!! 🚀❤️

@gangjun06 I will work on adding translation to some part of the app where it's missing right now. Would you be able to help translating them to Korean when you have time? Also, do you think you could help translate our Readme? 😃 I think we should try to do something like Supabase does (https://github.com/supabase/supabase/blob/master/i18n/languages.md)

mv-turtle (approved):

Issue #36, opened 2022-11-25
Build frontend within image build process and inject envars into containers at runtime


dangtony98:
### Feature description

Instead of building the frontend separately from the image, Infisical should pre-build the frontend within the image build process and make it so that envars can be injected into containers at runtime.

### Why would it be useful?

This would expedite the frontend deployment process and reduce downtime between deployments by 1-2 mins.

### Additional context

The difficulty stems from Next.js baking environment variables in at build time which wouldn't work for self-hosting since everyone uses their own environment variables; passing in environment variables into Next.js at runtime results in undefined environment variables.

reginaldbondoc:
Hi @dangtony98,

I am taking a jab at this and I would like to take a different approach.

Instead of going for runtime access (which will prevent optimization of built frontend package), let's remove the dependency. From the current frontend codebase, I see these:
1. Only the environment variable `NEXT_PUBLIC_WEBSITE_URL` is actually being used
2. One is in `next.config.js` for CSP and we can use `self` value for that - you already merged the change re: this
3. Another is usage as `PATH` const

I prepared a pull request to fulfill item 3 above. If that is fine, then the only remaining thing is to actually use the frontend image, no need for environment variables at all.

Unless you see the need for environment variable at runtime? In that case, we'll be forced to use a different approach, but that will prevent frontend build optimization. Until then, maybe my approach here will suffice?

dangtony98:
Hey egi!

I checked the PR and it looked good to me! — Good optimizations to frontend by removing `PATH` and `NEXT_PUBLIC_WEBSITE_URL`.

That said, I think there's still one variable that's `NEXT_PUBLIC_TELEMETRY_ENABLED` that's being used on the frontend for people to opt-out of telemetry at run-time; the remaining variables can be baked into the image build for sure.

reginaldbondoc:
Oh, you're right!

In that case, I will continue on finding a solution to this. 🙇

reginaldbondoc:
[Solution](https://github.com/Infisical/infisical/pull/80) has been accepted and merged.

PR #59, opened 2022-11-30
I-36 Remove dependency to PATH const

reginaldbondoc:
Prerequisites to fulfill Issue https://github.com/Infisical/infisical/issues/36 :

1. Remove usage of `NEXT_PUBLIC_WEBSITE_URL` in `next.config.js` and hard-code `self` value instead - this was already implemented: https://github.com/Infisical/infisical/commit/c3427d110a8fcb11a80cdc014a7ef36572c3c6ac
2. Remove usage of `NEXT_PUBLIC_WEBSITE_URL` as `PATH` constant

This PR deals with item 2.

`SecurityClient` depends on `PATH` constant to make sure to add JWT as `Authorization` header only on trusted domains. This `PATH` constant requires building of frontend image to get the client domain `NEXT_PUBLIC_WEBSITE_URL` from environment variable.

Here, trusted domain is the origin of the script. I understand that we don't want to send JWT to external APIs. I assume that this is the reason we have this check.

But..
1. We don't have to prepend `PATH` to every call to the origin, as the origin is automatically prepended when URL consists only of path (e.g. `/api/v1/..` will automatically be `https://domain.com/api/v1/..`)
2. We can inject the JWT for every request that courses through the `SecurityClient` and remove the check whether the caller is trying to reach origin domain. What then if we need to call external API? Just don't use `SecurityClient` 😉 

If I miss the rationale of this utility `SecurityClient`, please let me know. Maybe we can find a compromise.

### DIFF
--- a/frontend/components/utilities/SecurityClient.js
+++ b/frontend/components/utilities/SecurityClient.js
@@ -1,8 +1,6 @@
-import { PATH } from "~/const";
 import token from "~/pages/api/auth/Token";
 
 export default class SecurityClient {
-  static authOrigins = [PATH];
   static #token = "";
 
   contructor() {}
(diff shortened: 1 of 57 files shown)


dangtony98 (approved):
Looks good to me!

Thanks so much for the contribution! Good stuff 😄

PR #63, opened 2022-12-03
Fix babel error

SH5H:
Every files in the frontend folder has an error like this `Parsing error : Cannot find module 'next/babel'`
So I fix it and fix the typo in readme


mv-turtle (approved):
This makes sense! Thank you so much! Merging it

Issue #47, opened 2022-11-27
Convert Frontend to TypeScript


mv-turtle:
### Feature description
Convert all the files in frontend to TypeScript

### Why would it be useful?
Mainly because of the improved developer experience

### Additional context
There are quite a few JS files, so it might take a while to convert all of them

rabbanirizwan:
@mv-turtle Please let me know about some components that you haven't yet converted so that I can assist you.

mv-turtle:
@rabbanirizwan thank you so much for your help! Could you please help with any file in the `frontend > components > utilities` folder? For example, `file.js`, `randomId.js`, `generateBackupPDF.js`, `cryptography > aes-256-gcm.js`, etc. 

If you want, you can also look into `frontend > components > dashboard > DasboardInputField.js` or `frontend > components > basic > popups > BottomRightPopup.js`

This would be super useful! Also, if you want, you should join our community Slack: https://join.slack.com/t/infisical-users/shared_invite/zt-1kdbk07ro-RtoyEt_9E~fyzGo_xQYP6g. You can ask any questions there, and someone will reply almost immediately :)

mv-turtle:
Hi @rabbanirizwan! Would you still be interested in helping out here? Please let me know :)

SH5H:
I'm gonna handle this issue.

mv-turtle:
Thank you so much @SH5H!

mv-turtle:
This has now been completed! Thank you everyone for your help :)

PR #66, opened 2022-12-04
Update dashboard component

SH5H:
issue: #47 

I convert components in the dashboard to typescript.

and I check the drag & drop feature 

https://user-images.githubusercontent.com/25437192/205468840-c0590f91-0a4e-4906-8ac1-433ffbaa59c8.mov

### DIFF
--- a/frontend/components/basic/buttons/Button.tsx
+++ b/frontend/components/basic/buttons/Button.tsx
@@ -1,25 +1,28 @@
 import React from "react";
 import Image from "next/image";
 import { IconProp } from "@fortawesome/fontawesome-svg-core";
-import { FontAwesomeIcon, FontAwesomeIconProps } from "@fortawesome/react-fontawesome";
+import {
+  FontAwesomeIcon,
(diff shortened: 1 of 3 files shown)


mv-turtle:
@SH5H Quick question. Do UIEvent and WheelEvent need to be imported? It seems like they are not used anywhere if I understand correctly. Otherwise looks very good! Thank you so much :)

Screenshot 2022-12-03 at 23 29 01

SH5H:
Oh, that's my mistake.
Maybe we have to add a lint rule to avoid this happening.
May I add some lint rules??

SH5H:
I remove useless types.
And at last time, I forget to submit the Dropzone component So I add it this time

mv-turtle (approved):
This makes sense! Thank you so much for contributing :)

mv-turtle:
@SH5H Definitely! Feel free to add some lint rules, I think we can make our lint a bit stricter. Thank you!

PR #70, opened 2022-12-04
Add Lint rules

SH5H:
Add some lint rules.

I only add eslint:recommend plugin for typescript

you can check whole rule list in https://eslint.org/docs/latest/rules/


mv-turtle:
Looks good to me! Thank you!

mv-turtle (approved):

PR #77, opened 2022-12-05
Convert JS to TS 

SH5H:
Convert JS to TS several components.

### DIFF
--- a/frontend/components/utilities/attemptLogin.js
+++ b/frontend/components/utilities/attemptLogin.js
@@ -32,7 +32,6 @@ const attemptLogin = async (
   isLogin
 ) => {
   try {
-    let userWorkspace, userOrg;
     client.init(
       {
         username: email,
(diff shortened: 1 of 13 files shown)


frontend/components/utilities/attemptLogin.js
```diff
@@ -140,14 +139,8 @@ const attemptLogin = async (
...
-                DB_USERNAME: [
-                  "user1234",
-                  "personal",
-                ],
-                DB_PASSWORD: [
-                  "ah8jak3hk8dhiu4dw7whxwe1l",
-                  "personal",
-                ],
+                DB_USERNAME: ["user1234", "personal"],
+                DB_PASSWORD: ["ah8jak3hk8dhiu4dw7whxwe1l", "personal"],
```
SH5H, line 143: I think this kind of code has to be saved in .env file

mv-turtle: Oh, this is just dummy data :) 
We populate this for every new users so that they have some secrets to play around with in their dashboard. It's not real secrets or environment variables.

mv-turtle (approved):
Thank you so much for this contribution! I think it makes sense! :)

frontend/components/utilities/attemptLogin.js
mv-turtle: GitGuardian is blocking it even though it's not a real secret. Do you mind changing `ah8jak3hk8dhiu4dw7whxwe1l` to `example_password` in your branch?

This is just an example secret

mv-turtle:
Actually, I'm just gonna merge, because this is a mistake on the GitGuardian side

PR #80, opened 2022-12-05
I-36 Use pre-built frontend image instead of building Next.js app on boot

reginaldbondoc:
### Frontend changes
- New `Dockerfile` for building images
  - Now multi-stage
  - `NODE_ENV` value `production` pre-baked
  - `POSTHOG_HOST` pre-baked with value `https://app.posthog.com` (can be overwritten during build via `build-arg`)
  - `POSTHOG_API_KEY` will be pre-baked with "public" PostHog API key during build via `build-arg`. Now, self-hosted clients will already have this set in the image they pull.
  - Final production image has the optimized Next.js application, i.e. using the `standalone` files
  - On boot, some environment variables may override during runtime
    - `NEXT_PUBLIC_POSTHOG_API_KEY` - when set, will replace the pre-baked one. One application of this is for the cloud-hosted clients, where we want to use the "private" PostHot API key
    - `INFISICAL_TELEMETRY_ENABLED` (`TELEMETRY_ENABLED` in `.env` file setting the value of `INFISICAL_TELEMETRY_ENABLED` in `docker-compose`) - when set to `false`, will disable Infisical telemetry. Enabled by default.
    - Note on value of `INFISICAL_TELEMETRY_ENABLED` or `TELEMETRY_ENABLED`: Only the string `false` will disable telemetry. Any other value, including empty string, will enable telemetry.
  - Starts the application via `node` instead of `next` (the latter has issue with `SIGTERM`)
- Created separate singleton utility class `Telemetry`
  - Initialization of PostHog API (only when telemetry is enabled) will be done only once in this class
- Modified usages of posthog API to use the singleton `Telemetry` class

### Compose files changes
- `docker-compose.yml`
  - Removed the `platform` qualifier as this PR includes support for building multi-platform images (see changes to GitHub Actions workflow below)
  - Removed `build` definition for both backend and frontend services - we should be using the images instead of building (please correct me here if I'm wrong)
  - Removed unused environment variables
- `docker-compose.dev.yml`
  - Removed `image` property as we're building on boot anyway
  - Removed unused environment variables
  - Added dependency to `mongo` of `mongo-express`

### GitHub Actions workflow changes
- Improved `docker-image` workflow
 - Separated into backend and frontend jobs. Now, frontend can build in parallel even when some backend build steps fail.
 - Support for multi-platform (`amd64` & `arm64`)
 - Added test step to run the built image before actually pushing to Docker Hub (Note: the second build step does not actually build from scratch, as the layers for `amd64` are cached. Only the `arm64` will be built anew)


maidul98:
Hey Egi, i am still reviewing the changes! Really good stuff

frontend/scripts/replace-variable.sh
```diff
@@ -0,0 +1,16 @@
...
+REPLACEMENT=$2
+
+if [ "${ORIGINAL}" = "${REPLACEMENT}" ]; then
+    echo "Replacement is the same, skipping.."
+    exit 0
+fi
+
+echo "Replacing pre-baked value.."
+
+find /app/public /app/.next -type f ! -name "*.png" ! -name "*.svg" ! -name "*.gif" ! -name "*.jpg" ! -name "*.jpeg" ! -name "*.ico" |
```
maidul98, line 13: Instead of doing if file name not a certain extension, do you think it would be better to just scope to only. tsx and .js files? I don't think we will have any other file types that will need to be executed since this is a JS project. What do you think?

maidul98: Sounds good, let me know your finding. Also what performance did you notice when doing the text replace when a override is provided? 

Sidenote, i have also added the `PUBLIC_POSTHOG_API_KEY` to our secrets so that is good to go once we push this one out.

reginaldbondoc: > let me know your finding

I found out that just `.js` files 😄. So I updated this PR to limit replacement to only these.

> what performance did you notice when doing the text replace

With the last replace script, it took ~2s for the frontend to boot up. I don't have the timings of the actual replace though 😅. My latest change is limited to only `.js` files so definitely a lot faster now I hope ✌️

> i have also added the PUBLIC_POSTHOG_API_KEY to our secrets

Cool! Thanks!

maidul98 (dismissed):
Looks go to me! Ping me if the git workflow is not working as expected so we can take a look together

maidul98 (approved):
Looks good!

PR #84, opened 2022-12-06
Convert JS to TS

SH5H:
conversion JS to TS
- login1.js
- login2.js

refactoring
- attemptLogin


mv-turtle (approved):
Looks good to me! Thank you. Merging!

Issue #62, opened 2022-12-02
Add an `export` command with a few format options


edgarrmondragon:
### Feature description

A new `export` command that exports secrets in one of a few available formats. Some good initial formats may be:

- `dotenv`
- `json`
- `csv`

### Why would it be useful?

This would allow users to process secrets into a structure compatible with other tools.

### Additional context

This is similar to and inspired by [chamber](https://github.com/segmentio/chamber#exporting)'s export command.

mv-turtle:
Hi @edgarrmondragon! This is a great idea! We are currently working on setting up a series integrations for Infisical with other staging/product 3rd party services, but we will try to work on this issue right after that. 

If you have time and interest, feel free to send a PR for this. Any help is appreciated :)

PR #90, opened 2022-12-07
Add an `export` command

edgarrmondragon:
- I refactored some of the code in `cli/packages/cmd/run.go` so it could be
  reused for the new command.
- I added a docs page for the new command.

Closes https://github.com/Infisical/infisical/issues/62

### DIFF
--- a/cli/packages/cmd/export.go
+++ b/cli/packages/cmd/export.go
@@ -25,7 +25,7 @@ var exportCmd = &cobra.Command{
 	Use:                   "export",
 	Short:                 "Used to export environment variables to a file",
 	DisableFlagsInUseLine: true,
-	Example:               "infisical export --env=prod --format=json",
+	Example:               "infisical export --env=prod --format=json > secrets.json",
 	Args:                  cobra.NoArgs,
 	PreRun:                toggleDebug,
 	Run: func(cmd *cobra.Command, args []string) {
@@ -67,13 +67,13 @@ var exportCmd = &cobra.Command{
 		var output string
 		if shouldExpandSecrets {
 			substitutions := util.SubstituteSecrets(envsFromApi)
-			output, err = formatEnvs(envName, substitutions, format)
+			output, err = formatEnvs(substitutions, format)
 			if err != nil {
 				log.Errorln(err)
 				return
 			}
 		} else {
-			output, err = formatEnvs(envName, envsFromApi, format)
+			output, err = formatEnvs(envsFromApi, format)
 			if err != nil {
 				log.Errorln(err)
 				return
@@ -92,7 +92,7 @@ func init() {
 }
 
 // Format according to the format flag
-func formatEnvs(_ string, envs []models.SingleEnvironmentVariable, format string) (string, error) {
+func formatEnvs(envs []models.SingleEnvironmentVariable, format string) (string, error) {
 	switch strings.ToLower(format) {
 	case FormatDotenv:
 		return formatAsDotEnv(envs), nil
--- a/cli/packages/cmd/export.go
+++ b/cli/packages/cmd/export.go
@@ -1,0 +1,137 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+*/
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/Infisical/infisical-merge/packages/models"
+	"github.com/Infisical/infisical-merge/packages/util"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const (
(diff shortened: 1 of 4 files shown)


cli/packages/cmd/export.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// Format environment variables as a CSV file
+func formatAsCSV(envs []models.SingleEnvironmentVariable) string {
+	var csv string
+	csv += "Key,Value\n"
+	for _, env := range envs {
+		csv += fmt.Sprintf("%s,\"%s\"\n", env.Key, env.Value)
+	}
+	return csv
```
edgarrmondragon, line 115: This a very naive way of producing a CSV string, but it's probably good enough for a first iteration.

cli/packages/cmd/export.go
```diff
@@ -0,0 +1,137 @@
...
+			substitutions := util.SubstituteSecrets(envsFromApi)
+			output, err = formatEnvs(envName, substitutions, format)
+			if err != nil {
+				log.Errorln(err)
+				return
+			}
+		} else {
+			output, err = formatEnvs(envName, envsFromApi, format)
+			if err != nil {
+				log.Errorln(err)
```
maidul98, line 78: We should change this to a user friendly message and add the full message to debug log

cli/packages/cmd/export.go
```diff
@@ -0,0 +1,137 @@
...
+func init() {
+	rootCmd.AddCommand(exportCmd)
+	exportCmd.Flags().StringP("env", "e", "dev", "Set the environment (dev, prod, etc.) from which your secrets should be pulled from")
+	exportCmd.Flags().String("projectId", "", "The project ID from which your secrets should be pulled from")
+	exportCmd.Flags().Bool("expand", true, "Parse shell parameter expansions in your secrets")
+	exportCmd.Flags().StringP("format", "f", "dotenv", "Set the format of the output file (dotenv, json, csv)")
+}
+
+// Format according to the format flag
+func formatEnvs(_ string, envs []models.SingleEnvironmentVariable, format string) (string, error) {
```
maidul98, line 95: is there a reason why the parameter `_ string` is here but not used?

cli/packages/cmd/export.go
maidul98: Never mind, i see that you set the custom error message in the method

cli/packages/cmd/export.go
maidul98: Yeah, i think this is fine for first iteration. If you would like to use the standard csv package from go lang check this out https://pkg.go.dev/encoding/csv#example-Writer

docs/cli/export.mdx
```diff
@@ -0,0 +1,33 @@
...
+## Description
+
+Export environment variables from the platform into a file format.
+
+## Options
+
+| Option        | Description                                                                                                 | Default value |
+| ------------- | ----------------------------------------------------------------------------------------------------------- | ------------- |
+| `--env`       | Used to set the environment that secrets are pulled from. Accepted values: `dev`, `staging`, `test`, `prod` | `dev`         |
+| `--projectId` | Used to link a local project to the platform (required only if injecting via the service token method)      | `None`        |
```
maidul98, line 18: I think we should change this to `Used to determine from which infisical project your secrets will be exported from (only required if injecting via the service token method).`. What do you think?

cli/packages/cmd/export.go
```diff
@@ -0,0 +1,137 @@
...
+	FormatJson   string = "json"
+	FormatCSV    string = "csv"
+)
+
+// exportCmd represents the export command
+var exportCmd = &cobra.Command{
+	Use:                   "export",
+	Short:                 "Used to export environment variables to a file",
+	DisableFlagsInUseLine: true,
+	Example:               "infisical export --env=prod --format=json",
```
maidul98, line 28: I think also adding the example to export to a file will be helpful. Thoughts?

maidul98:
By the way, you should join our slack to communicate easier! https://join.slack.com/t/infisical-users/shared_invite/zt-1kdbk07ro-RtoyEt_9E~fyzGo_xQYP6g

cli/packages/cmd/export.go
edgarrmondragon: Yeah, leftover from a refactor when I was planning to use the environment name. I'll remove it

docs/cli/export.mdx
edgarrmondragon: Yeah, makes sense. Bad copy & paste 😅

cli/packages/cmd/export.go
edgarrmondragon: Makes sense. I'll add that.

PR #92, opened 2022-12-07
Add Notification Timeout

### LINKED ISSUE #89: Improve (Error) Notifications
### Feature description
It would be really great if notification auto-disappeared (e.g., a user had a notification saying that they have some duplicate keys and then it would disappear like int he attached example) so that they don't need to remove those themselves 

The following is an example of how PostHog does. Ours doesn't need to be so well-animated, but I think this would an end goal :) 
https://user-images.githubusercontent.com/78047717/206074042-f1b9c508-0993-4cc0-a33a-d7f0e3b3a1a2.mov

I also think that we generally have to add more notification (when secrets are saved, we should tell people something like "Successfully saved the secrets"). The auto-disappearence would be amazing in this case. 

### Why would it be useful?
It would a huge improvement for UX. I think many people would benefit from it.

asharonbaltazar:
closes #89 

## List of Changes
- Add `info` type for notifications. It's blue-colored.
- Require `notif.text` for `clearNotification`. It didn't make sense to have `clearNotification` to clear _all_ notifications. 
- `Notification` type within state is now `Required<>`. We need to ensure notifications within the `NotificationProvider` context have a `type` and `timeoutMs`
- Add optional `timeoutMs` to create a new `Notification`
- Add timeout capabilities for notifications to auto-hide after a specified amount of time (using the passed `timeoutMs`)
- Move notifications to bottom right of screen and hide notification wrapper if `!notifications.length`

### DIFF
--- a/frontend/components/context/Notifications/Notification.tsx
+++ b/frontend/components/context/Notifications/Notification.tsx
@@ -6,7 +6,7 @@ import { Notification as NotificationType } from "./NotificationProvider";
 
 interface NotificationProps {
   notification: NotificationType;
-  clearNotification: (text?: string) => void;
+  clearNotification: (text: string) => void;
 }
 
 const Notification = ({
--- a/frontend/components/context/Notifications/Notification.tsx
+++ b/frontend/components/context/Notifications/Notification.tsx
@@ -6,7 +6,7 @@ import classnames from "classnames";
 import { Notification as NotificationType } from "./NotificationProvider";
 
 interface NotificationProps {
-  notification: NotificationType;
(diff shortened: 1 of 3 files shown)


asharonbaltazar:
https://user-images.githubusercontent.com/58940073/206247938-7a22f43d-574f-4951-a82f-f6112a8a1bc4.mov

asharonbaltazar:
I still have to add `createNotification` after specified actions, per #89.

mv-turtle (approved):
This is really cool! I love it!

mv-turtle:
Thank you so much :)

PR #98, opened 2022-12-07
Fix workspace type definition

reginaldbondoc:


mv-turtle (approved):
Looks great! Thank you :)

PR #99, opened 2022-12-07
Add PR gate workflows

reginaldbondoc:
Two separate workflows that run on creation of pull requests:
* Frontend workflow watches for changes under `frontend` directory
* Backend workflow watches for changes under `backend` directory

For now, only executes `npm run build`. But the step for running tests is already included but commented out.

Also squeezed in a fix on build-arg for frontend image build.


maidul98 (approved):
approved

Issue #75, opened 2022-12-05
Better Contributing Getting Started Guide


dangtony98:
There's currently a lot of friction involved in the process to start contributing to Infisical. This thread exists to compile issues that contributors have been having whilst setting up Infisical — We hope to eliminate the friction entirely and make a better step-by-step contributing guide to help folks ease into development this week.

Notable issues:

- Environment variable setup: Difficulty configuring environment variables (which are optional/not-optional, how to generate keys, etc.)
- SMTP email setup: Difficulty configuring SMTP environment variables and "less secure apps" issue for gmail.

PR #108, opened 2022-12-09
Updated the start developing guide as suggested in issue #75

adrianmarinwork:
I've updated the content of the developing.mdx file that explains how to start and run the localhost of Infiscal.
The changes are mainly done in the **Set up environment variables** as it's the one that gave me more problems and because the rest it's clear enough already.
I also recommend referring to this file somewhere in the main ReadMe of this repository or as you already have linked, in the [get started contribution guide](https://infisical.com/docs/contributing/overview), because when I started trying to run the localhost I didn't know that this file existed.

If you have any suggestions or changes that you want to make just tell me.

### DIFF
--- a/docs/contributing/developing.mdx
+++ b/docs/contributing/developing.mdx
@@ -16,17 +16,48 @@ cd infisical
 
 ## Set up environment variables
 
-Tweak the `.env` according to your preferences. Refer to the available [environment variables](/self-hosting/configuration/envars).
+Before running the docker-compose we have to generate the .env file with the environment variables, you can create your own file or start with the
+`.env.example` as an example guide.
 
-```bash
-cp .env.example .env
+Mandatory variables in the `.env` file:
+
+1. Keys and JWT variables
+
+![image](https://user-images.githubusercontent.com/118568289/206791534-9c9d1431-e83d-49c0-8a54-b373ed0df820.png)
+
+The `.env.example` has these variables empty, you can self generate the `JWT and ENCRYPTION_KEY` with this [32-byte random hex strings generator](https://www.browserling.com/tools/random-hex).
+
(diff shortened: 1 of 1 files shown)


docs/contributing/developing.mdx
```diff
@@ -16,17 +16,48 @@ cd infisical
...
 ```
+PRIVATE_KEY='oGVv5rThrpZ7WLgQW27chY1cXngr4wLQIZnGfSKgHPk='
+PUBLIC_KEY='ldr6JaC7AY+tun3omGLdE4SWpkJbtVBOI54KfUP53Xc='
+```
+
+2. Mongo variables and site URL
+
+![image](https://user-images.githubusercontent.com/118568289/206792171-3376e3c6-c3ac-4d5d-8776-d78ee089b520.png)
+
+These variables are used to connect the MongoDB and set the URL for the localhost.
```
mv-turtle, line 41: Something to also add here is that `MONGO_USERNAME` and `MONGO_PASSWORD` should appear in `MONGO_URL` after `//` in the form of `MONGO_USERNAME:MONGO_PASSWORD`. The recommendation is to just use `root:example` for development purposes though.

@adrianmarinwork do you think this makes sense to be added here?

mv-turtle:
I think the rest looks great! Let me know what you think about my comment, and we can then merge :)

docs/contributing/developing.mdx
adrianmarinwork: Yeah, it could be a good thing to pinpoint in the case that the user uses his own username and password.

adrianmarinwork: Done!

mv-turtle: This is perfect!

mv-turtle (approved):
Thank you so much @adrianmarinwork! This is great!

PR #112, opened 2022-12-10
Add local SMTP server for development

reginaldbondoc:
I've noticed that new contributors are having some difficulty in configuring SMTP.

This PR adds a local SMTP server for testing purposes. It will be spawned as part of `docker-compose.dev.yml`

Developers who want to use this will just have to set the following in their `.env` file:
```
SMTP_HOST=smtp-server
SMTP_PORT=1025
SMTP_NAME=<whatever you like>
SMTP_USERNAME=team@infisical.com
SMTP_PASSWORD=
```

Sample:
Screenshot 2022-12-10 at 20 39 06


dangtony98 (approved):
Good stuff! — This really helps simplify emails on local development (not sure why we didn't use something like this before tbh).

Tested it out and works nicely with mails viewable on `http://localhost:8025` 🚀

PR #116, opened 2022-12-11
Add healthchecks and test image before push

reginaldbondoc:


backend/healthcheck.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable no-console */
+/* eslint-disable no-undef */
```
reginaldbondoc, line 3: Is there a way to exclude some files in linting process?

backend/src/index.ts
```diff
@@ -86,6 +92,31 @@ app.use('/api/v1/stripe', stripeRouter);
...
 app.use('/api/v1/integration-auth', integrationAuthRouter);
 
-app.listen(PORT, () => {
-	console.log('Listening on PORT ' + PORT);
+const server = http.createServer(app);
+
+const onSignal = () => {
+  console.log('Server is starting clean-up');
+  return Promise.all([
+    // your clean logic, like closing database connections
```
reginaldbondoc, line 100: @dangtony98 You  might want to add clean-up logic here. Maybe shut down database pool, if any?

backend/healthcheck.js
maidul98: I think the way you have it should disable those roles from the file

.github/workflows/docker-image.yml
```diff
@@ -3,83 +3,79 @@ name: Push to Docker Hub
...
-      #   run: |
-      #     docker run --rm infisical/backend:test
-      -
-        name: 🏗️ Build backend and push
+      - name: 📦 Build backend and export to Docker
+        uses: docker/build-push-action@v3
+        with:
+          load: true
+          context: backend
+          tags: infisical/backend:test
```
maidul98, line 27: Double checking, but this creates the image with tag `test` locally right?

maidul98: actaully, i see you have `load: true` so it only exports it locally

.github/resources/healthcheck.sh
```diff
@@ -0,0 +1,26 @@
+# Name of the target container to check
+container_name="$1"
```
maidul98, line 2: What do you think about making timeout as a second argument? That would allow us to have different timeout for different containers. Let me know if you think it will be useful at this point or not

backend/healthcheck.js
```diff
@@ -0,0 +1,27 @@
...
+  console.log(`HEALTHCHECK STATUS: ${res.statusCode}`);
+  if (res.statusCode == 200) {
+    process.exit(0);
+  } else {
+    process.exit(1);
+  }
+});
+
+healthCheck.on('error', function (err) {
+  console.error(err);
```
maidul98, line 23: Should we prepend this log with "HEALTH CHECK ERROR" because printing the log?

backend/src/index.ts
```diff
@@ -7,63 +9,67 @@ import dotenv from 'dotenv';
...
 const app = express();
 
 Sentry.init({
-	dsn: SENTRY_DSN,
-	tracesSampleRate: 1.0,
-	debug: NODE_ENV === 'production' ? false : true,
-	environment: NODE_ENV
+  dsn: SENTRY_DSN,
+  tracesSampleRate: 1.0,
+  debug: NODE_ENV === 'production' ? false : true,
```
maidul98, line 21: Why use `===` here? I think the three equlas check if the operands are equal not the string themselves

egibondoc-s2: Oh, I did not change this line.

I think `===` checks for value and type equality.

backend/healthcheck.js
egibondoc-s2: Good idea!

.github/workflows/docker-image.yml
egibondoc-s2: Yup, build and tag `test`. But load only so we can use it. On succeeding build (with push), image will be reused but tagged as `latest`.

backend/Dockerfile
```diff
@@ -2,11 +2,14 @@ FROM node:16-bullseye-slim
...
-COPY package*.json .
+COPY package.json package-lock.json ./
 
-RUN npm install
+RUN npm ci --only-production
 
 COPY . .
 
-CMD ["npm", "run", "start"]
+HEALTHCHECK --interval=10s --timeout=3s --start-period=10s \
```
maidul98, line 11: I wonder if this will get rate limited because of this https://github.com/Infisical/infisical/blob/4d0969fbc32a759e469745e621652cd85f0eda1d/backend/src/index.ts#L65 🧐

backend/healthcheck.js
egibondoc-s2: Yup. I just think that it will be cleaner if we have some whitelist of files not to be processed by linter. But that's just me. If we're okay with the first lines of file disabling lints, then all good. :bow:

maidul98: Feel free to  create a `.eslintignore` file to the select files completely from the lint

.github/workflows/docker-image.yml
maidul98: Awesome

backend/src/index.ts
dangtony98: @reginaldbondoc is right here. We should be using `===` in most if not all cases.

maidul98: ack

backend/Dockerfile
maidul98: Can we remove rate limit for a given endpoint @dangtony98

egibondoc-s2: Oh, good catch! I'll look into this, if this path is also rate limited.

backend/healthcheck.js
reginaldbondoc: Created base ignore file

backend/src/index.ts
reginaldbondoc: Added DB connection clean-up

backend/healthcheck.js
reginaldbondoc: Added

backend/Dockerfile
reginaldbondoc: Added skip in rate-limiter

backend/src/helpers/rateLimiter.ts
```diff
@@ -2,34 +2,35 @@ import rateLimit from 'express-rate-limit';
...
 const apiLimiter = rateLimit({
-	windowMs: 15 * 60 * 1000,
-	max: 400,
-	standardHeaders: true,
-	legacyHeaders: false
+  windowMs: 15 * 60 * 1000,
+  max: 400,
+  standardHeaders: true,
+  legacyHeaders: false,
+  skip: (request) => request.path === '/healthcheck'
```
maidul98, line 9: nicee

maidul98 (approved):
Looks good! 💥

maidul98:
I'll ping you on slack and we can test together

PR #117, opened 2022-12-11
:bug: Dashboard bugfix - Jumping Secrets #51

### LINKED ISSUE #51: Secrets in the dashboard are constantly jumping while being edited
### Describe the bug
When you are changing the names of the secrets, they are automatically jumping to maintain the alphabetical order. 

### Expected behavior
Ideally the order is only specified on the first load. All the further secrets that are added should, for example, just appear at the top. 

### Additional context
This might require in general how we approach the loading of secrets to the dashboard (the data structure), and/or how we track the order of secrets.

LemmyMwaura:
# Description
Fixes the jumping secrets on the dashboard while sorting. Works as expected.

Fixes #51.

Changes & Features:
 - [x] Secrets are only sorted when the sort button is clicked.
 - [x] Replaced the list containing keypair values with a map. It's now more intuitive and easier to work with.

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)


LemmyMwaura:
Awesome. Glad to help.

PR #127, opened 2022-12-17
Adding automatic linter for checking commit

SH5H:
delete the prettier option in the automatic linter
and add `--ignore-scripts` for every `npm install` command

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,7 +1,0 @@
-{
-  "semi": true,
-  "trailingComma": "none",
-  "singleQuote": true,
-  "printWidth": 80,
-  "useTabs": false
-}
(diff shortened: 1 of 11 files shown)


mv-turtle:
Hi @SH5H! Why do you want to delete prettier? :)

maidul98:
> Why do you want to delete prettier? :)

Because its style is too restrictive and causes unnecessary changes during commit. I will review this @SH5H. I'll update you once I finish!

maidul98:
Small nit, can you pease delete any pretier related files such as `.prettierrc` and resubmit PR? Otherwise, looks good!

SH5H:
I can't update the frontend package file.
Even I update the lock file, the CI Process is always failed.
Can you delete the `prettier` in package dev dependency and update the frontend lock file instead of me?

maidul98 (approved):
You have to run `npm install` so that the lock file auto updates but sounds good, i'll do it! Thank you so much

SH5H:
> You have to run `npm install` so that the lock file auto updates but sounds good, i'll do it! Thank you so much

I did, but it didn't work

PR #140, opened 2022-12-20
Backend test setup

reginaldbondoc:
* Initial test configuration
* Sample test (just the healthcheck endpoint for now)
* Moved things from `index.ts` to separate parts for easier test setup
* Initial coverage configuration
* GH Action workflow
  * Modified Backend PR check to run test
  * New workflow to post comment/report of test result and coverage to PR (Note: this new workflow will not successfully run until merged into target main branch)

Sample:

Test result:
image

Coverage annotations:
image

  
Note: some files were automatically modified by prettier on commit (husky?).


maidul98:
For code coverage, will the missed branch or line be shown to the user? I'm curious how they can know what has caused the coverge to fail

.github/workflows/check-be-pull-request.yml
```diff
@@ -1,41 +1,43 @@
...
-      #   name: 🧪 Run tests
-      #   run: npm run test:ci
-      #   working-directory: backend
-      -
-        name: 🏗️ Run build
+      - name: 🧪 Run tests
+        run: npm run test:ci
+        working-directory: backend
+      - name: 📁 Upload test results
+        uses: actions/upload-artifact@v3
```
maidul98, line 34: Does this test result get deleted at some point?

.github/workflows/be-test-report.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: 📁 Download test results
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          name: be-test-results
+          path: backend
+          workflow: check-be-pull-request.yml
```
maidul98, line 22: Making sure I understand, the `check-be-pull-request`  will be ran and its result on `success` will then be passed on to the next step?

reginaldbondoc:
> For code coverage, will the missed branch or line be shown to the user? I'm curious how they can know what has caused the coverge to fail

Yes:
image

(But the link there just points to commit, not actual file/line in GitHub). But this shall provide proper feedback on which line/branch is not covered.

.github/workflows/check-be-pull-request.yml
reginaldbondoc: Ah, because GitHub cannot access files generated by other jobs/workflows. And the only way to "share" files is to upload them from one job, and download to another job/workflow.

.github/workflows/be-test-report.yml
reginaldbondoc: Close 😅 

To be accurate: `be-test-report` listens for successful completion of workflow `check-be-pull-request`. (The former does not trigger the latter)

But I guess that's what you're already thinking, just wanted to be accurate 🙇

Zamion101:
Will be the test only available to run in GH CI/CD or is it possible to run test locally? Example `npm run test:backend`

reginaldbondoc:
> Will be the tests only available to run in GH CI/CD or is it possible to run test locally? Example `npm run test:backend`

Possible to run test locally. Dev can do: `npm test`

.github/workflows/check-be-pull-request.yml
maidul98: Will the upload override the release artifacts here? https://github.com/Infisical/infisical/releases/tag/v0.1.11 
Because I noticed that when i tried to upload artifacts, it would override

maidul98: So for each PR that is created, the artifacts of the test results will be uploaded in order to share the results with the user ?

maidul98:
Lastly, i didn't see a coverge threshold defined unless there is a default? https://jestjs.io/docs/configuration#coveragethreshold-object

@dangtony98 can you also review the index.ts file to make sure you are up to date with the new restructure?

backend/src/services/database.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const initDatabase = (MONGO_URL: string) => {
+  mongoose
+    .connect(MONGO_URL)
+    .then(() => console.log('Successfully connected to DB'))
+    .catch((e) => {
+      console.log('Failed to connect to DB ', e);
+      setTimeout(() => {
+        console.log(e);
+      }, 5000);
```
Zamion101, line 12: Is there specific reason of waiting 5 seconds before writing error to console?

backend/src/services/health.ts
```diff
@@ -0,0 +1,32 @@
...
+export const setUpHealthEndpoint = <T>(server: T) => {
+  const onSignal = () => {
+    console.log('Server is starting clean-up');
+    return Promise.all([
+      () => {
+        if (mongoose.connection && mongoose.connection.readyState == 1) {
+          mongoose.connection.close();
+          () => {
+            console.info('Database connection closed');
+          };
```
Zamion101, line 14: First Promise.all has only 1 function and it's not `async () => {}`. It's not a problem but it will be better to use `new Promise((resolve,reject)=>{...})` if you need this function to be async.
Second thing is why not use `console.info(...)` directly rather than making it function, also is it even working? I'm doubting it will work it's just a function there is no call present in this function.

Here is what I mean:
image

Zamion101 (commented):
There are some questions  I'm curious about. Overall not seeing any major security or design flaw. There needs to be more work on some parts towards consistent type definition and error handling.

reginaldbondoc:
> Lastly, i didn't see a coverge threshold defined unless there is a default? https://jestjs.io/docs/configuration#coveragethreshold-object

That's correct. Because there's no test yet, aside from the initial one that I added. If we define the threshold, the run will fail. 😉 

For now, I plan to just start with this, and add more tests as we move forward. Once we reach 50%, I would suggest we set the threshold and then aim higher. 👍

reginaldbondoc:
> I will be working on central error handling

❤️ 

> first I need this PR to be merged into main for me to start working on some changes

That's consistent with my rationale of opening this PR as early as now 👍

.github/workflows/check-be-pull-request.yml
reginaldbondoc: > Will the upload override the release artifacts here

The upload shouldn't override. They're artifacts of PR run.

> test results will be uploaded in order to share the results with the user

Yes

backend/src/services/database.ts
reginaldbondoc: Good catch! I actually just moved the entire function from `index.ts`.

I will remove this additional logging statement. Thanks for spotting!

backend/src/services/health.ts
reginaldbondoc: Indeed, good catch as well! I will align. 🙇

backend/src/services/database.ts
Zamion101: Don't work too much on Logging. I will be adding Central Logging as well using `winston` and `Loki` by Grafana.

Zamion101 (approved):
I'm okay with current state of this code. Waiting for your review @dangtony98

maidul98 (approved):

Issue #142, opened 2022-12-21
JWT_SERVICE_SECRET falls under `optional` in .env.example


Zamion101:
### Describe the bug
Environment variable `JWT_SERVICE_SECRET` needs to present in order to create Service Tokens, but it in `.env.example` it is under the `# JWT lifetime` and has `Optional` keyword in description. One can assume that you don't need to enter anything here but thats not the case. It needs to be under `# JWT` section in order to work without problem.

### To Reproduce
1. Go to `.env`
2. Delete the content from following variable 'JWT_SERVICE_SECRET'
3. Run `docker-compose`
4. Navigate to 

### Expected behavior
Variable should be under `# JWT` and must be indicated as `Required`.

### Screenshots
![image](https://user-images.githubusercontent.com/8071263/208794591-536e267c-ad44-4520-885c-b5ad45e5120b.png)
![image](https://user-images.githubusercontent.com/8071263/208794638-2775726c-b816-4e1a-9d39-148d6cae6f38.png)
![image](https://user-images.githubusercontent.com/8071263/208795633-4f054e5a-31ef-496f-ba87-9cbaf9245801.png)

### Additional context
No additional context available.

dangtony98:
Hey @Zamion101 thanks for pointing this out!

Will fix this and update in docs soon 😁

PR #144, opened 2022-12-21
Fix `JWT_SERVICE_SECRET` not present in docs

Zamion101:
According to issue #142, relocated `JWT_SERVICE_SECRET` and updated example image in `/docs/contributing/developing.mdx` to contain `JWT_SERVICE_SECRET` as well.


mv-turtle (approved):
Looks great! Thank you @Zamion101 :)

PR #152, opened 2022-12-22
Refactoring backend codebase to use central error handling and logging

Zamion101:
#### [21.12.2022]
- Added Midleware to capture and handle all errors
    - will catch error from `throw ...` and `next(...)`
- Added RequestError base error class to build consistent error objects
- Added `VERBOSE_ERROR_OUTPUT` environment variable to control `stacktrace` of error mesages, with this variable it's easier to locate error in developmen environment.
- Added common errors like `UnauthorizedRequestError`, `BadRequestError`
- Added consistent Logging solution using `winston`
    - Supports Loki Transporter using `winston-loki`
- Outputing Legal disclaimer to console when `TELEMETRY_ENABLED=true`
- Added Sentry.captureException to Error Handler
#### [22.12.2022]
- Added middleware to capture unrouted requests for responding with proper error messages and status code
#### [24.12.2022]
- New error types such as `IntegrationNotFoundError`, `WorkspaceNotFoundError`, `AccountNotFoundError' and more.

#### [21.12.2022]
- Changed console.log to getLogger() favor of using consistent logging
#### [24.12.2022]
- Some of the error types in middlewares changed to more related error types.
- Environment variable of 'VERBOSE_ERROR_OUTPUT' changed to more reliable validation method in `config/index.ts` as per @dangtony98 requested.

#### New error messages for unrouted requests (stacktrace can be omitted using `VERBOSE_ERROR_OUTPUT=false`)
image


backend/environment.d.ts
```diff
@@ -14,6 +14,8 @@ declare global {
 			JWT_SIGNUP_SECRET: string;
       MONGO_URL: string;
       NODE_ENV: 'development' | 'staging' | 'testing' | 'production';
+      VERBOSE_ERROR_OUTPUT: boolean;
```
dangtony98, line 17: VSCode is showing me an error on this one; I think it has to be a `string` type instead because everything passing through as envars enters in `string` format.

Screen Shot 2022-12-23 at 12 20 37 PM

For this reason we have this line in the backend config file for `TELEMETRY_ENABLED`:

`const TELEMETRY_ENABLED = process.env.TELEMETRY_ENABLED! !== 'false' && true;`

Let me know what you think.

Zamion101: Interesting for me VSCode showing no error and as far as I now Dotenv will parse it to boolean. But will change for to be sure. 
image

backend/src/middleware/requireBotAuth.ts
```diff
@@ -15,30 +16,22 @@ const requireBotAuth = ({
...
-        } catch (err) {
-            Sentry.setUser(null);
-			Sentry.captureException(err);
-			return res.status(401).send({
-				error: 'Failed bot authorization'
-			});
+        const bot = await Bot.findOne({ _id: req[location].botId });
+        
+        if (!bot) {
+            return next(UnauthorizedRequestError({message: 'Failed to locate Bot account'}))
```
dangtony98, line 22: To provide more context on this `Bot` stuff (it's a new feature/mechanism that we've been working on this past 2 weeks) — Every project/workspace has a `Bot` that has its own public-private key pair. The purpose of the bot is for people to be able to share their project/workspace key with it so it can perform various automations including syncing secrets to cloud integrations (can elaborate more on this when we call soon).

This middleware route is basically to check that for whatever `botId` being passed through likely as route params, that it matches to a `Bot` in the database and that `Bot` is a bot of a project that the user on the request is a member of.

Wondering if this would fall under the `UnauthorizedRequestError`? It could be that no `Bot` exists for that `botId` being passed in? I'm cool with leaving it as is but just wanted to provide more context anyways!

backend/src/middleware/requireIntegrationAuth.ts
```diff
@@ -21,48 +22,40 @@ const requireIntegrationAuth = ({
...
-				_id: integration.integrationAuth
-			}).select(
-				'+refreshCiphertext +refreshIV +refreshTag +accessCiphertext +accessIV +accessTag +accessExpiresAt'
-			);
-
-			if (!integrationAuth) {
-				throw new Error('Failed to find integration authorization');
-			}
+		if (!integration) {
+			return next(UnauthorizedRequestError({message: 'Failed to locate Integration'}))
```
dangtony98, line 33: Same inquiry as previous `Bot` inquiry regarding error type. Should we be using `UnauthorizedRequestError` here or should there be a more specific Error type that implies that resource could not be found?

dangtony98:
Just left a few minor comments above to sort out. Otherwise, great work with improving the logging for Infisical!

Looking forward to merging this 😁

backend/src/middleware/requireBotAuth.ts
Zamion101: I wasn't sure to use which error message or should I create a new types of error. So just used `UnauthorizedRequestError` while it was inside of Authentication logic. It's my mistake not to write TODO/FIXME or comment about it. Thanks for giving more context on the topic. Will change to correct erros with next commit. Any suggestions for types, names, or context for this type of error would be appreciated

backend/src/middleware/requireIntegrationAuth.ts
Zamion101: Same thing here as well. I answered in previous one.

Zamion101:
At this point, I think we can start merging these changes to take small steps forward. We still have a lot of work to do to completely refactor all legacy errors to the new RequestError. If I were to do it by myself, it would take much longer than necessary. That's why I think we can merge this pull request and encourage existing and new contributors to use getLogger() for logging, `next(RequestError())` in routes and `throw RequestError()` in any other place.

Even though I tested in Web GUI and HTTP Request, please give it a try and give me some feedback. I will make corrections as soon as possible of any given. @dangtony98 @maidul98

backend/src/app.ts
```diff
@@ -69,6 +73,37 @@ app.use('/api/v1/stripe', stripeRouter);
 app.use('/api/v1/integration', integrationRouter);
 app.use('/api/v1/integration-auth', integrationAuthRouter);
 
+
+//* Handle unrouted requests and respond with proper error message as well as status code
+app.use((req, res, next)=>{
```
dangtony98, line 78: Would be great if you could modularize this last middleware into a separate file as it feels a little chunky here.

Zamion101: No problem.

Zamion101: Split error handler middleware to a seperate file as your request.

Zamion101: Oh no I messed up  the naming :|

backend/src/app.ts
```diff
@@ -29,8 +29,9 @@ import {
   integrationAuth as integrationAuthRouter
 } from './routes';
 import { getLogger } from './utils/logger';
-import { catchUnroutedRequests } from './middleware/catchUnroutedRequests';
 import { RouteNotFoundError } from './utils/errors';
+import { errorHandler } from '@sentry/node/types/handlers';
```
Zamion101, line 33: Ooops.

dangtony98 (dismissed):
Looks good to me!

There're definitely a lot of parts that need to be re-classified/renamed (e.g. there may need to be a BotNotFound Error) and ironed out but I think this is a great start for us to start re-working the backend and getting more contributors involved in that process :)

Great stuff.

PR #155, opened 2022-12-22
update env vars for smtp

### LINKED ISSUE #153: Certain SMTP servers do not work
### Describe the bug
Currently, if the SMTP authentication username is not the SMTP sender address, emails will not send. 

### To Reproduce
Use an SMTP server that has a different authentication username and sender address.

### Expected behavior
Emails send successfully.

### Additional context
The SMTP authentication username and sender address should be separate in the environmental variables. I will create a PR to fix this issue.

jonerrr:
Closes #153

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,12 +38,15 @@ SITE_URL=http://localhost:8080
 
 # Mail/SMTP
 # Required to send emails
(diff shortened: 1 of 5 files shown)


dangtony98 (approved):
Looks good to me!

Glad that this SMTP configuration will be more versatile to folks 😁

PR #166, opened 2022-12-24
feat(frontend/signup): add resend code functionality

naorpeled:
image
image


mv-turtle (approved):
This is really useful! I completely forgot about resending these emails :)

Thank you for helping! Merging...

PR #168, opened 2022-12-25
Issue 159 - Adding docker image versioning

cerrussell:
Modified docker workflow to push git version tags with image. If this part works, I will then probably need to create and add an additional version specifier onto this metadata since multiple images are being pushed for the same version.


maidul98:
Thnaks for the PR, i'll review this tomorrow

.github/workflows/docker-image.yml
```diff
@@ -10,6 +10,17 @@ jobs:
...
+      - name: Docker metadata action
+        id: meta_backend
+        uses: docker/metadata-action@v4.1.1
+        with:
+          images: |
+            infisical/backend
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
```
maidul98, line 22: I understand `type=ref,event=branchtype=ref,event=pr` but what will the outputs of 

`semver,pattern={{major}}.{{minor}}` and
`type=semver,pattern={{version}}` be?

Where is it getting major and minor from?

cerrussell: It is pulling the git tags that already exist for releases. However, I now realize this won't really work because it is a manually run workflow and will probably just say the version is master or main.

I have another couple ideas that don't rely on the git info. I'll try them out on my own docker hub and see which is the most simple and sustainable. Using the git info just would have been convenient for creating a mechanism to manage the latest tag (I'm specifically thinking of when the most recently pushed image is not the one you want Docker offering as the default).

maidul98: I wonder if we can just use current time somehow for versioning. @reginaldbondoc do you have any thoughts on this?

cerrussell: That seems reasonable if semantic versioning isn't necessary and you will never be pushing images of previous builds. If you wanted to have an image of a previous release, using current time will make it difficult to figure out which image is actually the most recent version.

.github/workflows/docker-image.yml
```diff
@@ -10,6 +10,17 @@ jobs:
     steps:
       - name: ☁️ Checkout source
         uses: actions/checkout@v3
+      - name: Docker metadata action
```
reginaldbondoc, line 13: Minor and just nitpicking:

Any reason for having this as 2nd step and not after all the setup (QEMU, BuildX, Login), right before the actual docker build step?

.github/workflows/docker-image.yml
```diff
@@ -10,6 +10,17 @@ jobs:
...
+        id: meta_backend
+        uses: docker/metadata-action@v4.1.1
+        with:
+          images: |
+            infisical/backend
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
```
reginaldbondoc, line 21: I just read the action documentation, and IMHO, labeling the images with `PR-n` is not something I would advise.

reginaldbondoc (commented):
I think @cerrussell has some good idea here. It's just that he's limited with the current trigger: `workflow_dispatch`.

I did not dare change the trigger before as I don't have a grasp of the release process of Infisical just yet. But given that we now want to version our images, I think it's high time to rethink when/how we want to create/tag images.

What do you think about:
* Creating image versions out of Git tags. This will simplify the usage of action into just `type=semver,pattern={{version}}`
* Changing the trigger of this workflow from `workflow_dispatch` into
```
on:
  push:
    tags:
      - '[0-9]+.[0-9]+.[0-9]+*'
```
?

The "release manager/developer" will then need to create a tag, preferably following SemVer, and that should deal with versioning the code and the image that contains that code. It will then be easy to track which image has the commit.

.github/workflows/docker-image.yml
cerrussell: @reginaldbondoc Honestly just because that's where they put it in the example in the [documentation](https://github.com/marketplace/actions/docker-metadata-action) for the action. No issues changing the order though. And regarding the pr-n label, that was again just one of the tags they included in the default semver configuration. Easily modified.

maidul98: @cerrussell If you have slack please join so that we can all discuss more easily. Here is our slack https://join.slack.com/t/infisical-users/shared_invite/zt-1kdbk07ro-RtoyEt_9E~fyzGo_xQYP6g

cerrussell: @maidul98 Joined.

PR #169, opened 2022-12-25
Dashboard sidebar

mv-turtle:
Still need to fix some bugs though this should be ready to merge after the CLI is updated to prioritize overridden secrets

What this adds? 
- A sidebar instead of a popup for each secret. 
Screenshot 2022-12-25 at 10 18 22
- Conversion of some files to TS
- Performance optimizations for the dashboard


Zamion101:
Do you think it will be a good idea to store history of secrets? Would that not be a security risk or violates the security practices?

PR #172, opened 2022-12-26
chore(frontend): removed unneccessary react import in project

akhilmhdh:
### What does it do?

1. Removed unncessary react import made in project

### Why is it needed?

Slight reduction in bundle size 

Ref: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html


akhilmhdh:
I can also attach all the eslint error fixes with this, if no one is working on that

mv-turtle (approved):
Oooo, sounds good! This makes a lot of sense. Thank you so much :)

We do need to run eslint at some point, but I'm afraid this will lead to a lot of merge conflicts 🥲

PR #177, opened 2022-12-26
ci: add autocompletion and manpage

jon4hz:
This adds support for the automatic generation of shell autocompletion files and linux manpages.

### DIFF
--- a/cli/scripts/manpages.sh
+++ b/cli/scripts/manpages.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+set -e
+rm -rf manpages
+mkdir manpages
+cd cli
+go run . man | gzip -c > "../manpages/infisical.1.gz"
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ node_modules
 .DS_Store
 
 /dist
+/completions/
+/manpages/
 
 # frontend
 
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -6,6 +6,11 @@
 #     - cd cli && go mod tidy
 #     # you may remove this if you don't need go generate
 #     - cd cli && go generate ./...
+before:
+  hooks:
+    - ./cli/scripts/completions.sh
+    - ./cli/scripts/manpages.sh
+
 builds:
   - id: darwin-build
     binary: infisical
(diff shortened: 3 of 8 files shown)


jon4hz:
1080f2581c72b207dcb40355c9ebfc8ed7c7d2e7 also changes the archive format to `zip` on windows. 
This is more convenient, since windows doesn't support `tar.gz` out of the box.

maidul98:
I'm wondering if a man page is necessary? I feel like the current overview of commands and flag is sufficient. Is there something that man will be more helpful for than the current overview when you only type in `infisical` in the terminal?

Screen Shot 2022-12-26 at 8 44 35 PM

jon4hz:
I always like manpages because it gives you an overview of all available commands including all flags.
If you need to search for something specific, you can simply `/search` for it. 

image


`infisical help` just shows you the available subcommand and current command flags but not all flags from subcommand, too.

maidul98:
Is there a reason why I was not able to load the man page like you show in the screenshot when i did `infisical man`?

jon4hz:
`infisical man` will generate the manpage in it's raw roff format (which the `man` command will display in a readable way).
If you want it in a readable format ad hoc, you can use `infisical man | roff2text | less`

maidul98:
If we can make the man page self sufficient meaning, it requires no other dependency such as `roff2text` or other tools to conevert it into a readable format, let's add the man page. I think having it more like `man ls` where you just type in the command and it shows you the full text without having to pipe to a converter is more convenient. 

We can also leave the man page for the future if you would like and only add your other changes in this PR. Let me know what you think

jon4hz:
That's exactly how it will work :)

[This](https://github.com/Infisical/infisical/pull/177/files#diff-7326b55c062b0f46fe9e39aace0a25f4515cf206040fb91a6fd2cae839f5e826R110) deploys he manpage accordingly when the user installs infisical through an nfpm build. Same for an AUR installation [here](https://github.com/Infisical/infisical/pull/177/files#diff-7326b55c062b0f46fe9e39aace0a25f4515cf206040fb91a6fd2cae839f5e826R152). The manpage will be directly accessible using `man infisical` using these installation methods. 

The `infisical man` command is more like a build hook to generate the page and shouldn't be used by the users in the first place. The command is hidden and wont appear when you type `infisical help`

maidul98:
Ah i see, i misunderstood! Okay, sounds good. I'll review full PR in a bit

cli/scripts/manpages.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+set -e
+rm -rf manpages
+mkdir manpages
+cd cli
+go run . man | gzip -c > "../manpages/infisical.1.gz"
```
maidul98, line 6: oh i see, the hidden man command is used here. Nice

maidul98 (approved):
Looks good to me! 🚀  Will be releasing new CLI update soon

PR #180, opened 2022-12-27
fix(backend): resolved workspace membership invite failure

akhilmhdh:
### What does it do?

1. The membership invite to workspace was failing because

https://github.com/Infisical/infisical/blob/main/backend/src/controllers/v1/membershipController.ts#L187

The above line expects the middleware `requireWorkspaceAuth` to have workspace details. But due to not having populate query in it. Workspace id is only available causing the access to be undefined. 

### Why is it needed?

1. This will fix workspace membership invite


mv-turtle (approved):
This makes a lot of sense! Thank you for fixing this :)

PR #183, opened 2022-12-28
Helm updates

jon4hz:
Hey, 
I was missing some configuration options while deploying infisical with helm. 

The following features were added:
- support for custom annotations
- support for custom service type for the frontend
- allow image image overrides

I also added some default helper templates and refactored the naming, labeling and selectors in a more "helmish" way. 
Your service currently uses selectors like `app=backend` which isn't very unique and could lead to troubles, e.g. if more than one infisical installation is running in the same namespace.


maidul98:
Awsome, i'll take a look at this soon

maidul98:
I was unable to run this on my minikube cluster. Were you able to test it out? I also noticed that the backend is unable to connect to the mongo db instance in the deployment anymore

mv-turtle:
Hey @jon4hz! Have you maybe had time to look at Maidul's comment above ^ :)

maidul98:
@mv-turtle  Gave me an update via Slack. An update from my side @jon4hz , I did not have time to fix the issue with deployment yet. If you are free, feel free to dig back into it or let me know and I can also do it

jon4hz:
Sorry for the late response. 
I'm not sure how to handle this correctly. 
The mongodb service is now named according to `{{ include "infisical.mongodb.fullname" . }}` and we can't access named templates from the value file. However I wouldn't rename the service to anything static. Otherwise you can't `helm install` this chart multiple time in the same namespace. 

What do you think about adding another block in the values like `mongodbConnection`?
```yaml
mongodbConnection:
  host: xxx # defaults to the internal mongo service
  port: 27017
  username: xxx
  password: xxx
```

maidul98:
@jon4hz i will get back to this tomorrow

maidul98:
I have looks around quite a bit to see if we can access the named templated in the values file but it looks like this is not possible as you mentioned. 

With your suggestion, it looks like the user would be responsible for making sure if there are two instances of Infisical installed that they do not have colliding host names for the mongo db service. Are there any other down sides? 

```
mongodbConnection:
  host: xxx # defaults to the internal mongo service
  port: 27017
  username: xxx
  password: xxx
```

jon4hz:
Actually the user doesn't have to care about any colliding name because all services will include the helm release name which is unique by default. 

[c23b291](https://github.com/Infisical/infisical/pull/183/commits/c23b291f250a3f0d9671872c5ae826e42e4848ac) changes the way how the `MONGO_URL` variable is set. The `MONGO_URL` will use the internal mongodb by default with options to override any default values. 

But as I already mentioned, this change is breaking but without it, the mongodb connection will always be very static.

jon4hz:
Btw, what do you think about changing the backend, so it accepts env vars like `MONGO_USER` and `MONGO_PASSWORD`? 
This would allow passing only the password as kubernetes secret. 
With the current options, you'd have to pass the entire `MONGO_URL` as secret which isn't very practical imo.

maidul98:
I think the reason for putting the entire `MONGO_URL ` was because mongo db cloud has a very different format and putting just the username and password would not cover that use case. For the new block you wanted to add to the values file, how would it look with a mongo cloud url?

maidul98:
Let me know if you have any concerns for https://github.com/Infisical/infisical/pull/183/commits/375412b45d0b66cf82f6907dd26e76cd46758b8c other wise i'll get it merged in!

maidul98 (approved):

Issue #131, opened 2022-12-19
.NET integration


gbro3n:
### Feature description
An integration with .NET's configuration tooling would be great to see.

### Why would it be useful?
There's a lot of .NET projects out there that could benefit from this.

mv-turtle:
Infisical mostly likely supports .NET in its current form. We just don't have the docs for it yet. We need to test it out, and I will get back to you.

jon4hz:
That's definitely possible and .NETs [ConfigurationBuilder](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationbuilder?view=dotnet-plat-ext-7.0) makes it fairly easy.

```c#
public IConfigurationRoot ConfigureConfiguration()
{
    var builder = new ConfigurationBuilder()
        .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
        .AddEnvironmentVariables();
    return builder.Build();
}
```

mv-turtle:
Oh interesting! This makes sense - I haven't found the time to test it out and write the docs yet (not very familiar with .NET) :(

@jon4hz If you're interested in helping with this, that would be so amazing 🙏 No pressure though!

jon4hz:
Well I'm no expert on .NET either, I just recently started learning c#. :) 
But yeah sure, what would you need for the docs? The snipped I posted should be more or less everything that's needed to parse the environment variables.

mv-turtle:
Kinda like [this](https://infisical.com/docs/integrations/frameworks/nextjs). First step is going to be the same, and then you just need to run the start command and see if environment variables are inserted :)

The frameworks are [here](https://github.com/Infisical/infisical/tree/main/docs/integrations/frameworks).

mv-turtle:
@gbro3n @gerwim here are the instructions for how to use Infisical with .NET: https://infisical.com/docs/integrations/frameworks/dotnet

Let us know if you have any questions

PR #184, opened 2022-12-28
docs: add dotnet example

jon4hz:
ref #131


mv-turtle:
Awesome that it works! Thank you so much :)

mv-turtle (approved):

PR #185, opened 2022-12-28
ci: auto build and tag images

jon4hz:
Currently all docker images are tagged with `latest` and overwritten on new builds. 
This is a bit problematic in production environments since it prevents version control. 

I'd like to propose this build pipeline for the docker images. 

The pipeline will be triggered automatically on a new tag and create images with the according tags. 
Pull requests to `main` will trigger the pipeline but only to test if the image still works. The image wont be pushed to dockerhub. 

The docker image with tag `latest` will always be the latest tag on github. Builds triggered by `workflow_dispatch` will have the tag `main`.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -29,7 +29,7 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: 🏷️ Meta
         id: meta
-        uses: docker/metadata-action@v3
+        uses: docker/metadata-action@v4
         with:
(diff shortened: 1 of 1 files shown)


maidul98:
Thank you! This is actually being discussed in https://github.com/Infisical/infisical/pull/168. I believe she started working on it first so let's give her the opportunity to finish that PR

jon4hz:
> Thank you! This is actually being discussed in #168. I believe she started working on it first so let's give her the opportunity to finish that PR

Oh whoops, didn't see that :sweat_smile:

Issue #31, opened 2022-11-25
Feature: Have the ability to add/delete/rename environments


mv-turtle:
### Describe the feature
I would be awesome to have a project setting in which you would be able to add/remove/edit available environments. 

### Why would it be useful?
Infisical currently support supports 4 default environments (Development, Staging, Production, and Testing). Many users have requested to have their own custom environments.

mv-turtle:
FYI @icaruk; I created an issue

gmgale:
I would be happy to take a look as a first time issue for me. Let mew know what you think. Thanks!

mv-turtle:
I created a front-end for this in the above commit. Just the backend left

akhilmhdh:
Hey there.

I would love to work on this issue.

maidul98:
This has been completed by @akhilmhdh

PR #186, opened 2022-12-28
feat(#31): implemented api for environment crud operations

akhilmhdh:
## What does it do

This will enable workspace environment crud operations

## Screenshots

<details>

<summary> Expand</summary>

## List Env
Screenshot 2023-01-11 at 11 06 19 PM

## Create Env
Screenshot 2023-01-11 at 11 06 24 PM

### Update Env
Screenshot 2023-01-11 at 11 06 47 PM

### Delete Env
Screenshot 2023-01-11 at 11 06 59 PM

### Dashboard
Screenshot 2023-01-11 at 11 07 16 PM

## Service Token
Screenshot 2023-01-11 at 11 07 40 PM

## Integration Page

Screenshot 2023-01-13 at 9 06 59 PM

Screenshot 2023-01-13 at 9 07 02 PM

</details>


mv-turtle:
This is awesome! Makes total sense to me! 

I feel like @dangtony98 might say that the routes should be like 'workspace/:workspaceId/environment'. I know that he is doing some v2 revamp which should create a common way of naming routes. @dangtony98 do you have any opinions here?

backend/src/app.ts
```diff
@@ -92,6 +93,7 @@ app.use('/api/v1/integration-auth', v1IntegrationAuthRouter);
 // v2 routes
 app.use('/api/v2/workspace', v2WorkspaceRouter);
 app.use('/api/v2/secret', v2SecretRouter);
+app.use('/api/v2/environment', v2EnvironmentRouter);
```
Zamion101, line 96: should be `/api/v2/environments`

backend/src/controllers/v2/environmentController.ts
```diff
@@ -0,0 +1,167 @@
...
+    );
+    if (!workspace) {
+      throw new Error('Failed to update workspace');
+    }
+
+    await Secret.updateMany(
+      { workspace: workspaceId, environment: oldEnvironmentSlug },
+      { environment: environmentSlug }
+    );
+    await ServiceToken.updateMany(
```
Zamion101, line 87: Above and following database operations should follow `Atomicity` from ACID. This applies for similar operations further in the PR.

backend/src/routes/v2/environment.ts
```diff
@@ -0,0 +1,54 @@
...
+    acceptedStatuses: [COMPLETED, GRANTED],
+  }),
+  param('workspaceId').exists().trim(),
+  body('environmentSlug').exists().trim(),
+  body('environmentName').exists().trim(),
+  validateRequest,
+  environmentController.createWorkspaceEnvironment
+);
+
+router.put(
```
Zamion101, line 26: Correct HTTP Method should be `PATCH`.

"When a client needs to replace an existing Resource entirely, they can use PUT. When they're doing a partial update, they can use HTTP PATCH."

We're renaming just **one** environment variable and not entire list of environments.

Zamion101 (changes requested):
There needs some changes and further discussion on how to handle envrionments in upcoming V2 API by @dangtony98

backend/src/controllers/v2/environmentController.ts
akhilmhdh: Can you be more elaborate your point. All the above operations are atomic by nature from Mongodb 4.0 onwards.

backend/src/routes/v2/environment.ts
akhilmhdh: Not really. GIven the route scope is environments and not list of environments. 

An environment here contains slug and name. Patch would have asked user to update either slug or name, but that's not intentded here.  We are asking user to provide both the parameters together that is all entity. That is PUT not patch.

akhilmhdh:
> There needs some changes and further discussion on how to handle envrionments in upcoming V2 API by @dangtony98

@Zamion101 And yes. I have talked with tony and now this PR is in a waiting for v2 routes to come inti picture and launched

backend/src/controllers/v2/environmentController.ts
akhilmhdh: @Zamion101 Ohhh sure that's valid issue. But the catch is for mongodb to support transactions, its not simple as sql db. U need to enable replication to support transactions. For that we need to change the image's doc and also prod i guess.

I think then the scope gets bigger for this PR. Would love to get input from @dangtony98 @maidul98 on this.

Zamion101: I don't think we will have any problem as it is right now. I checked the dependencies and package versions and we're good to use following method from Mongoose documents;

https://mongoosejs.com/docs/transactions.html

akhilmhdh: @Zamion101 Yah library wise is ok all good.

But to support transactions in mongodb there is a architecture change in db required.

Check this: https://www.mongodb.com/community/forums/t/why-replica-set-is-mandatory-for-transactions-in-mongodb/9533

mv-turtle:
Looking into this right now! Thank you for such detailed description!

frontend/pages/api/workspace/getAWorkspace.ts
```diff
@@ -0,0 +1,31 @@
...
+  name: string;
+  organization: string;
+  environments: Array<{ name: string; slug: string }>;
+}
+
+/**
+ * This route lets us get the workspaces of a certain user
+ * @returns
+ */
+const getAWorkspace = (workspaceID:string) => {
```
mv-turtle, line 15: I think this is a duplicate of GetProjectInfo.js
We kinda use workspaces and projects interchangeably which id pretty confusing

mv-turtle (approved):

PR #188, opened 2022-12-30
add ability to import/export secrets with comments

### LINKED ISSUE #74: Add import/export of .yml files
### Feature description
1. A user should be able to drag and drop a .yml file into the dashboard and then it is parsed into secrets immediately (similar how .env is right now)
2.  A use should be able to download secrets as a .yml file (currently only .env).

### Why would it be useful?
Currently only .env files are supported. Frameworks like Rails use .yml.

### Additional context
This is more of a frontend issue but a similar thing could be done in the CLI

mocherfaoui:
closes #74 

few things to note:
- it would be better if some functions are moved to separate files like `getSecrets` and `checkOverrides`
- I used `Menu` as a dropdown menu from headless/ui, I think it would be better if it's extracted to a reusable component for future use
	- there is an issue with the dropdown menu, the menu doesn't close after a button is clicked. i'm investigating this issue at the moment
- it would be better the `onButtonPressed` prop from `Button` component is optional so it can be used as a trigger like the case above


mv-turtle:
Hi @mocherfaoui! Sorry for taking so long to review. This looks awesome! 

Merging this right now. Meanwhile, a few things to mention: 
- as you mentioned, it would be nice to put those secret function into corresponding files. Feel free to do this, or I will try to do it in one of the future commits. 
- What do you think we should do about nested .yml files? Cyrrently it just converts a nested value to: 
Screenshot 2023-01-06 at 15 59 20

mv-turtle:
@mocherfaoui I can't pass this Frontend Check... :(((

Would you be able to have a look? Then it should be ready to merge

mv-turtle (approved):
I fixed it! Merging right now 🚀
Though please let me know what you think of the questions above :)

Issue #175, opened 2022-12-26
Add popup before secret delete 


maidul98:
### Feature description
Users should be warned before deleting any secrets. Currently, it is too easy to delete secrets and users are not informed of impact that deleting secrets can have. Users should be informed that any downstream integration which relies on the secret will be impacted. We can achieve this with a pop up of some kind

### Why would it be useful?
This would make users think twice before deleting a secret.

JoaoVictor6:
I want to resolve this issue, I would create a component named `DeleteKeyDialog` or similar. Does make sense this approach?

mv-turtle:
Hi @JoaoVictor6! Yes, I think that makes sense. We just need to ask them something like "Are you sure you want to delete this secret? This cannot be undone." And the buttons would be 'Cancel' and 'Delete'. 

I'm currently making some changes to the deletions. Would you be fine with waiting ~1 day before starting? Thank you so much for your help :) 

Also, please, join our community Slack for our contributors and users. You can ask any questions you might have there. Here is the link: https://join.slack.com/t/infisical-users/shared_invite/zt-1kdbk07ro-RtoyEt_9E~fyzGo_xQYP6g

JoaoVictor6:
Okay! :)

mv-turtle:
But I think we should still add a modal right? For example, Slack also has a sidebar, but they ask you any way if you want to delete a message

JoaoVictor6:
Certainly. The confirm delete action is very simple, in this case make more sense create a dialog

PR #197, opened 2023-01-06
Add popup before secret delete

JoaoVictor6:
## Add popup before secret delete
This is my solution for issue #175 :)

### Goals
- [ ] Create a dialog
- [ ] Isolate logic for delete action in a component
- [ ] Improve types on delete scope

### Demo
https://user-images.githubusercontent.com/68869379/210918265-cfc7078f-2013-4b94-b131-d78ec7231f19.mp4
> The bug on the version history box on sidebar has because when recording the server wasn't running :p

### DIFF
--- a/frontend/components/dashboard/SideBar.tsx
+++ b/frontend/components/dashboard/SideBar.tsx
@@ -9,6 +9,7 @@ import Button from '../basic/buttons/Button';
 import Toggle from '../basic/Toggle';
 import CommentField from './CommentField';
 import DashboardInputField from './DashboardInputField';
+import { DeleteActionButton } from './DeleteActionButton';
 import GenerateSecretMenu from './GenerateSecretMenu';
 
 
@@ -28,6 +29,10 @@ interface OverrideProps {
   pos: number;
   comment: string;
 }
+export interface DeleteRowFunctionProps { 
+  ids: string[]; 
+  secretName: string; 
+}
 
 interface SideBarProps {
   toggleSidebar: (value: string) => void; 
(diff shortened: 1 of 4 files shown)


frontend/components/dashboard/SideBar.tsx
```diff
@@ -28,6 +29,10 @@ interface OverrideProps {
   pos: number;
   comment: string;
 }
+export interface DeleteRowFunctionProps { 
+  ids: string[];
```
JoaoVictor6, line 33: I centralize this type. I noticed an inconsistency in the data property. In all cases, this is one item, why is it an array and not an object?

frontend/public/locales/en/dashboard.json
```diff
@@ -25,6 +25,11 @@
     "comments": "Comments & Notes",
     "personal-explanation": "This secret is personal. It is not shared with any of your teammates.",
     "generate-random-hex": "Generate Random Hex",
-    "digits": "digits"
+    "digits": "digits",
+    "delete-key-dialog": {
```
JoaoVictor6, line 29: this structure is valid?

mv-turtle: yeah, I think that's good. We will need to modify this to make it more consistent

frontend/components/dashboard/SideBar.tsx
mv-turtle: This is very strange. Logically, it should be an ObjectId, I don't know why it is an array

mv-turtle (approved):
Looks really great! Thank you so much for the contribution :)

frontend/components/dashboard/SideBar.tsx
JoaoVictor6: I made some refactoring in the sidebar component. maybe I can create a pr with this change for your review?

mv-turtle: @JoaoVictor6 Sounds good! Looking forward

PR #202, opened 2023-01-07
Draft: Telemetry for v2 routes

mv-turtle:


backend/src/controllers/v2/secretController.ts
```diff
@@ -61,15 +86,15 @@ export const batchCreateSecrets = async (req: Request, res: Response) => {
...
+      environment,
       type: rawSecret.type,
       user: new Types.ObjectId(req.user._id)
     }
 
     sanitizedSecretesToCreate.push(safeUpdateFields)
   })
 
-  const [bulkCreateError, newlyCreatedSecrets] = await to(Secret.insertMany(sanitizedSecretesToCreate).then())
+  const [bulkCreateError, secrets] = await to(Secret.insertMany(sanitizedSecretesToCreate).then())
```
Zamion101, line 97: Is there a Typo? Isn't the correct form is "sanitizedSecretsToCreate"

backend/src/controllers/v2/secretController.ts
```diff
@@ -27,23 +31,44 @@ export const createSingleSecret = async (req: Request, res: Response) => {
...
+      distinctId: req.user.email,
+      properties: {
+        numberOfSecrets: 1,
+        workspaceId,
+        environment,
+        channel: req.headers?.['user-agent']?.toLowerCase().includes('mozilla') ? 'web' : 'cli',
+        userAgent: req.headers?.['user-agent']
+      }
+    });
+  }
```
Zamion101, line 57: This PostHog boilerplate takes a lot of space and not easy to change in bulk in the future. Is it possible to seperate to function and use that function?

backend/src/controllers/v2/secretsController.ts
```diff
@@ -0,0 +1,472 @@
...
+    const channel = req.headers?.['user-agent']?.toLowerCase().includes('mozilla') ? 'web' : 'cli';
+    
+    const readAction = await EELogService.createActionSecret({
+        name: ACTION_READ_SECRETS,
+        userId: req.user._id.toString(),
+        workspaceId: workspaceId as string,
+        secretIds: secrets.map((n: any) => n._id)
+    });
+
+    readAction && await EELogService.createLog({
```
Zamion101, line 186: Can this code be simplified and take advantage of Promises?
```ts
await EELogService.createActionSecret({
        name: ACTION_READ_SECRETS,
        userId: req.user._id.toString(),
        workspaceId: workspaceId as string,
        secretIds: secrets.map((n: any) => n._id)
    })
    .then((readAction)=>{
        EELogService.createLog({
             userId: req.user._id.toString(),
             workspaceId: workspaceId as string,
             actions: [readAction],
             channel,
             ipAddress: req.ip
        });
    })
    .then(()=>{return res.status(200).json({secrets})})
```

or just be entirely async without await, it will speed up response time because response is not dependent to Audit Logs success status.

backend/src/controllers/v2/secretsController.ts
```diff
@@ -0,0 +1,472 @@
...
+            secretKeyCiphertext,
+            secretKeyIV,
+            secretKeyTag,
+            secretValueCiphertext,
+            secretValueIV,
+            secretValueTag,
+            secretCommentCiphertext,
+            secretCommentIV,
+            secretCommentTag
+        } = secret;
```
Zamion101, line 233: I prefer to just use `secret.xyz` in required fields rather then deconstructing to seperate variables like here. It's just makes code hard to read because of the variables that used once and never again below code.

backend/src/routes/v2/secrets.ts
```diff
@@ -0,0 +1,177 @@
...
+    ADMIN, 
+    MEMBER,
+    SECRET_PERSONAL,
+    SECRET_SHARED
+} from '../../variables';
+
+router.post(
+    '/',
+    body('workspaceId').exists().isString().trim(),
+    body('environment').exists().isString().trim().isIn(['dev', 'staging', 'prod', 'test']),
```
Zamion101, line 21: So what about custom envrionment names?

backend/src/routes/v2/secrets.ts
```diff
@@ -0,0 +1,177 @@
...
+                        !secret.secretKeyIV ||
+                        !secret.secretKeyTag ||
+                        !secret.secretValueCiphertext ||
+                        !secret.secretValueIV ||
+                        !secret.secretValueTag
+                    ) {
+                        throw new Error('secrets array must contain objects that conform to the Secret interface');
+                    }
+                }
+            } else if (typeof value === 'object') {
```
Zamion101, line 111: I have seen this check **a lot** in this PR. If we're gonna use this check over and over again we need to make it seperate function and use that function here. It's really not easy to manage in this scale.

backend/src/routes/v2/secrets.ts
```diff
@@ -0,0 +1,177 @@
...
+    }),
+    requireSecretsAuth({
+        acceptedRoles: [ADMIN, MEMBER]
+    }),
+    secretsController.updateSecrets
+);
+
+router.delete(
+    '/',
+    [
```
Zamion101, line 146: Is this correct? The check is inside of Array but the Array is not deconstructed. Should the correct form be `...[check('secretIds')]`

Zamion101 (changes requested):
There needs some changes to this PR and associated commits. There are some parts that we can optimize and some parts that we can seperate logics to their own functions for the purpose of easier to manage in the future. I would like to get your opininons on my reviews.

backend/src/routes/v2/secrets.ts
mv-turtle: I think we don't have customer environment names yet. But agree, when we implement that we should change that

backend/src/controllers/v2/secretController.ts
mv-turtle: I think it might be, but this controller is not user anymore. We will deprecate it soon, but for now the correct version is in `secretsContoller.ts`

backend/src/controllers/v2/secretController.ts
mv-turtle: Hmmm interesting! How would you do it? Personally I'd say it's already pretty concise, though it does take a lot of space

backend/src/controllers/v2/secretController.ts
```diff
@@ -93,10 +139,12 @@ export const batchDeleteSecrets = async (req: Request, res: Response) => {
...
   const deleteOperationsToPerform: AnyBulkWriteOperation<ISecret>[] = []
 
+  let numSecretsDeleted = 0;
   secretIdsToDelete.forEach(secretIdToDelete => {
     if (secretsUserCanDeleteSet.has(secretIdToDelete)) {
       const deleteOperation = { deleteOne: { filter: { _id: new Types.ObjectId(secretIdToDelete) } } }
       deleteOperationsToPerform.push(deleteOperation)
+      numSecretsDeleted++;
     } else {
       throw RouteValidationError({ message: "You cannot delete secrets that you do not have access to" })
```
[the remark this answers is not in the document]
mv-turtle: Yeah, as mentioned above, I think this controller is about to be deprecated

backend/src/controllers/v2/secretController.ts
Zamion101: What about we create a function to get `distinctId`, `channel`, `userAgent` from `req` and `context` variable which extends the default. Something like this;
```ts
type PostHogEventData = {
    event: string
    distinctId: string
    properties: Record<any,any> & {
        channel?: "web"|"cli"
        userAgent?: string
    }
}

function capturePostHogEvent(eventName: string, req: Request, context: Partial<PostHogEventData>){
    if(!postHogClient) return;

    const postHogData = Object.assign({
        event: eventName,
        distinctId: req.user.email,
        properties: {
            channel: req.headers?.['user-agent']?.toLowerCase().includes('mozilla') ? 'web' : 'cli',
            userAgent: req.headers?.['user-agent']
        }
    }, context)

    postHogClient.capture(postHogData);
}

...

capturePostHogEvent('secrets_added', req, {
    properties: {
        numberOfSecrets: 1,
        workspaceId,
        environment
    }
})
```

dangtony98: Agree with this!

I think we can add this as an extension to `/services/PostHogClient` in the backend. We can convert it to a class and have some useful methods on there like this one to capture events.

In the interest of time and un-blocking some people, we're going to merge this PR and add in this change in a separate PR. Feel free to submit a new PR with this modularization after the merge!

backend/src/controllers/v2/secretController.ts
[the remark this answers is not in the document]
dangtony98: Yup this controller will be deprecated soon pretty much after the merge as we transition to the new endpoints in `v2/secrets`.

backend/src/controllers/v2/secretsController.ts
dangtony98: Ah yea we can do without await then because you're right response is not dependent on the audit log success status — Let's submit that in the next PR

backend/src/routes/v2/secrets.ts
dangtony98: Yeah Akhil will add these right after.

backend/src/controllers/v2/secretsController.ts
dangtony98: Sounds good I'll make this adjustment after!

backend/src/routes/v2/secrets.ts
dangtony98: Sounds good. I'll throw this check into a helper function 👌

backend/src/routes/v2/secrets.ts
dangtony98: Yeah I believe it's correct. I've tested it with and without correct `secretIds`

dangtony98: Oh actually hmm maybe it would also work without the array..

dangtony98: Alrighty fixed.

dangtony98 (approved):

PR #204, opened 2023-01-08
Issue 159 - Docker image tags

cerrussell:
#168 

Added the short commit hash as a Docker tag. Kept the latest tag as well so that old images don't continue to be offered as latest on Docker Hub (there is nothing automatically updating this and Docker just assumes latest is truly latest).


Zamion101 (approved):
This changes looks good to me.

maidul98 (approved):
Awsome! Mergeing

PR #219, opened 2023-01-13
feat(CLI): adding new export format (dotenv-export)

imakecodes:
# Why
I want to have a way to export the variables with the word `export ` at the beggning of each variable.

This is helpful to permit use the variables in bash subprocesses

The diference between use or not the `export` is described on [this post](https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export)

# How to use
```
infisical export --format=dotenv-export

export INFISICAL_DEV_INT=1
export INFISICAL_DEV_STRING=some-string
```


maidul98 (approved):
awesome, looks good!

PR #222, opened 2023-01-14
migration(frontend): migrated frontend files to ts execpt dialog component

akhilmhdh:
1. Completed migration to ts except dialog components  and removed ts errors of all files till now.
2. Some i have put any type which are basically API calls over state which will soon be replaced with react-query.

Screenshot 2023-01-15 at 12 12 53 AM


akhilmhdh:
SIdenote: Don't mind the non uniform linting codebase having. Let's do that after ts migration complete. Ill put with prettier

mv-turtle (approved):
Just checked everything. Looks great!

PR #226, opened 2023-01-15
chore(frontend): changed source code to src folder from root

akhilmhdh:
Why this change.

1. Moving the source code to src folder and keeping all frontend configs will immensely help to have a clear separation of concern. 
2. Like docker mounting, tailwind config all those will be easier to configure
3. Patched lockfile from swc issue
4. Added a new ts path called `@app` -> src. Thus call imports can regarding file be made like from now on: 

```ts
import {} from "@app/folder1/folder2/"
```


akhilmhdh:
@mv-turtle If this was decided to keep due to pages folder file based routing.

No worries cause nextjs will identify pages inside src

Ref: https://nextjs.org/docs/advanced-features/src-directory

mv-turtle (approved):

PR #238, opened 2023-01-18
feat(frontend): added storybook with tailwind integration

akhilmhdh:
1. Added storybook integration
2. Installed some radix component needed for new component library

Screenshot 2023-01-18 at 11 34 17 PM


frontend/src/components/v2/Button/Button.tsx
```diff
@@ -0,0 +1,35 @@
...
+export type ButtonProps = ButtonHTMLAttributes<HTMLButtonElement> & Props;
+
+export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
+  ({ children, isDisabled = false, className, ...props }, ref): JSX.Element => {
+    return (
+      <button
+        ref={ref}
+        aria-disabled={isDisabled}
+        type="button"
+        className={twMerge(
```
mv-turtle, line 22: Why don't we use classnames (we already use it for some other components) instead of tw-merge? It is much more popular package and I'd be more confident in adding it. What do you think?

akhilmhdh: I think the answer to that would be this. Try to override and the classname priority would hit u @mv-turtle


https://github.com/dcastil/tailwind-merge/blob/v1.8.1/docs/what-is-it-for.md

mv-turtle (approved):

PR #242, opened 2023-01-21
Allow INFISICAL_URL to use domain for self-hosted

asheliahut:
## Description

Updates the cli to allow for an environment variable named `INFISICAL_URL` this way when a self hosted user is using the CLI they don't need to specify the domain over and over.

The order of respect for domains is as follows
Infisical Cloud Domain -> Environment Variable -> CLI Argument

This way the CLI argument can always override even the environment variable.

### DIFF
--- a/cli/packages/cmd/root.go
+++ b/cli/packages/cmd/root.go
@@ -6,8 +6,10 @@ package cmd
 import (
 	"os"
 
-	"github.com/Infisical/infisical-merge/packages/config"
 	"github.com/spf13/cobra"
+
+	"github.com/Infisical/infisical-merge/packages/config"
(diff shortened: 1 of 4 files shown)


asheliahut:
@Grraahaam I like your idea, but sadly what you are suggesting wouldn't help in the case of global usage which is what I was looking for here. What if you don't have a project setup yet at all? I think having both options would be neat in the long run.

maidul98 (approved):
Awsome PR! A lot of folks have asked for this. Let me know if you have ideas for further improving the CLI

asheliahut:
@Grraahaam that still isn't flexible enough for my use case. I need to be able to change it quickly between runs and I'd prefer not having to use a tool like jq or something to edit a config file constantly.

PR #249, opened 2023-01-23
Allow changing backend domain

### LINKED ISSUE #248: Allow configuring `API_URL` for frontend
Thanks for the wonderful project!

### Feature description

Right now the frontend makes always makes backend requests to the same domain. This should be made configurable via an `API_URL` env var.

### Why would it be useful?

I have the backend hosted on a different domain than the one frontend is in. Thus, when the frontend makes requests to `https://infisical.<domain>/api/v1/auth/login1`, the request fails. I have the backend live on `https://infisical-backend.<domain>`.

Perhaps it can be configured on the frontend image via an `API_URL` env var which defaults to the current domain to preserve backwards compatibility.
### Additional context

I am using dokku to self host infisical.

IgnisDa:
Fixes #248.


IgnisDa:
cc @mv-turtle

IgnisDa:
@maidul98 Could you please approve the workflow runs so that I can test if the changes are correct?

PR #256, opened 2023-01-25
feat(ui): added new auth guard with react-query and axios

akhilmhdh:
1. Added new auth guard
2. Added react-query with axios and interceptor to auto refresh
3. Setup for react-query folder structure


mv-turtle:
@dangtony98 would you be able to have a look please?

dangtony98 (approved):

PR #257, opened 2023-01-25
new secrets are added to the top

### LINKED ISSUE #237: Move the view to the secret that was just inserted
### Feature description
When a user adds a new key, it is currently added to the bottom of the dashboard which results in users not realizing that a new secret has been added successfully. We want to scroll the dashboard to the bottom of the list when this happens. Ideally, we might want to add new secrets at the top, and then we would need to scroll the dashboard to the top.

### Why would it be useful?
This feature would improve the suer experience and remove confusion for users.

mocherfaoui:
closes #237


mv-turtle (approved):
Just tested it out and it works like magic! :)
Thank you so much for helping out with this @mocherfaoui!!

Issue #182, opened 2022-12-28
Helm: use bitnamis mongodb chart


jon4hz:
### Feature description
What do you think about adding [bitnamis mongodb chart](https://github.com/bitnami/charts/tree/main/bitnami/mongodb) as helm dependency instead of managing the mongodb chart yourself? 

### Why would it be useful?
Currently it isn't possible to setup a productive infisical instance using helm unless you outsource the mongodb. Bitnamis Chart is extremely good and brings all the customization one would need.

maidul98:
This is great suggestion and will allow people to have mongo db StatefulSet. Adding this should also this issue https://github.com/Infisical/infisical/issues/173

If you are familiar with that mongo chart, please feel free to post a PR

jon4hz:
To be honest, I never used that chart myself but I can take a look at it. Will most likely be next year when I get to that though :)

maidul98:
Sounds good

maidul98:
Yes!

PR #273, opened 2023-01-29
fix(charts): mongodb persistence + ingressClassName support and configuration improvements suggestion 

Grraahaam:
# Description 📣

<kbd>fix</kbd>
- Replaced `metadata.annotations.kubernetes.io/ingress.class: "nginx"` with the recommended `spec.ingressClassName` in the `ingress` resource (details [here](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation) and [there](https://stackoverflow.com/a/73917249))

- Simplified value paths in templates (e.g. `.Values.backend` to `$backend`) reducing the paths length

<kbd>feat</kbd>
- MongoDB storage persistence by using PVC and linking the database storage to it

<kbd>refacto</kbd>
- Refactored the `values.yaml` and the way we pass configuration variables to our pods by preferring native k8s secrets, and using `envFrom`

<kbd>docs</kbd>
- Added more detailed installation instructions

*I admit that this PR might contain a lot of different changes, but all related to our helm-charts. Ping me if you rather want me to split those changes into dedicated PRs*

## Type ✨

|  |  |
| --------------: | :-: |
|         Bug fix | ❌  |
|     New feature | ✅  |
| Breaking change | ✅  |
|   Documentation | ✅  |

# Tests 🛠️

### Helm

```sh
# Infisical
cd ./helm-charts/infisical
helm lint
helm template infisical .

# Infisical Secrets Operator
cd ./helm-charts/secrets-operator
helm lint
helm template secrets-operator .
```

---

### DIFF
--- a/helm-charts/infisical/templates/backend-deployment.yaml
+++ b/helm-charts/infisical/templates/backend-deployment.yaml
@@ -37,7 +37,7 @@ spec:
         - containerPort: 4000
         envFrom:
         - secretRef:
-            name: {{ $backend.kubeSecretRef | default "infisical-backend" }}
+            name: {{ $backend.kubeSecretRef | default (include "infisical.backend.fullname" .) }}
         env:
         - name: MONGO_URL
           value: {{ include "infisical.mongodb.connectionString" . | quote }}
@@ -68,7 +68,7 @@ spec:
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ $backend.kubeSecretRef | default "infisical-backend" }}
+  name: {{ $backend.kubeSecretRef | default (include "infisical.backend.fullname" .) }}
 type: Opaque
 data:
   {{- range $key, $value := $backend.secrets }}
--- a/helm-charts/infisical/templates/frontend-deployment.yaml
+++ b/helm-charts/infisical/templates/frontend-deployment.yaml
@@ -35,7 +35,7 @@ spec:
           periodSeconds: 10
         envFrom:
         - secretRef:
-            name: {{ $frontend.kubeSecretRef | default "infisical-frontend"  }}
+            name: {{ $frontend.kubeSecretRef | default (include "infisical.frontend.fullname" .)  }}
         ports: 
         - containerPort: 3000
 
@@ -69,7 +69,7 @@ spec:
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ $frontend.kubeSecretRef | default "infisical-frontend" }}
+  name: {{ $frontend.kubeSecretRef | default (include "infisical.frontend.fullname" .) }}
 type: Opaque
 data:
   {{- range $key, $value := $frontend.secrets }}


maidul98:
Hey @Grraahaam I'll review this soon!

helm-charts/infisical/templates/backend-deployment.yaml
```diff
@@ -9,23 +10,23 @@ metadata:
...
       annotations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
       containers:
-      - name: {{ template "infisical.name" . }}-{{ .Values.backend.name }}
-        image: "{{ .Values.backend.image.repository }}:{{ .Values.backend.image.tag | default .Chart.AppVersion }}"
-        imagePullPolicy: {{ .Values.backend.image.pullPolicy }}
+      - name: {{ template "infisical.name" . }}-{{ $backend.name }}
+        image: "{{ $backend.image.repository }}:{{ $backend.image.tag | default .Chart.AppVersion }}"
```
maidul98, line 28: I'm wondering if we should get rid of the `default .Chart.AppVersion`.Because we would never have a image tag that is the `.Chart.AppVersion`. Those aren't related. What do you think

helm-charts/infisical/templates/backend-deployment.yaml
```diff
@@ -34,32 +35,24 @@ spec:
...
         envFrom:
         - secretRef:
-            name: {{ .Values.backend.kubeSecretRef }}
+            name: {{ $backend.kubeSecretRef }}
         {{- end }}
         env:
         - name: MONGO_URL
           value: {{ include "infisical.mongodb.connectionString" . | quote }}
-        {{- if .Values.backendEnvironmentVariables }}
-        {{- range $key, $value := .Values.backendEnvironmentVariables }}
```
maidul98, line 46: is there a reason why you deleted the abality to write envs right into the deployment file? I'm gussing because it is more secure and can be stored in `kubeSecretRef`? Or were you thinking or something else

maidul98: Oh i see, you put them in a secret below

helm-charts/infisical/templates/backend-deployment.yaml
```diff
@@ -70,3 +63,15 @@ spec:
...
+
+---
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $backend.kubeSecretRef | default "infisical-backend" }}
+type: Opaque
+data:
+  {{- range $key, $value := $backend.secrets }}
```
maidul98, line 75: I'm curious, let's say I removed some secrets from the values file for backend. Would upgrading the helm chart also refect a one to one to this secret? Or do old values that no longer exist in the values file continue to exist but only only overwritten

helm-charts/infisical/templates/frontend-deployment.yaml
```diff
@@ -9,65 +10,69 @@ metadata:
...
+        image: "{{ $frontend.image.repository }}:{{ $frontend.image.tag | default .Chart.AppVersion }}"
+        imagePullPolicy: {{ $frontend.image.pullPolicy }}
         readinessProbe:
           httpGet:
             path: /
             port: 3000
           initialDelaySeconds: 10
           periodSeconds: 10
-        {{- if .Values.frontend.kubeSecretRef }}
+        {{- if $frontend.kubeSecretRef }}
```
maidul98, line 36: It looks like the user has to define a kubeSecretRef + define the environment variables in values.frontend.secrets ? I feel like if they are defining `values.frontend.secrets` then we should just put those in a secret and auto reference them. Thoughts?

helm-charts/secrets-operator/templates/deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- $ctrlManager := .Values.controllerManager }}
```
maidul98, line 1: All changes for the `secrets-operator` helm chart gets auto generated by Kubebuilder. This means that these changes will be overwritten on next release. So feel free to remove or leave it as is

helm-charts/infisical/templates/backend-deployment.yaml
maidul98: Yes, perfect. Let's do the same for frontend as well

helm-charts/infisical/templates/backend-deployment.yaml
maidul98: I agree, this is better approach

helm-charts/infisical/templates/backend-deployment.yaml
maidul98: Awesome

helm-charts/infisical/templates/frontend-deployment.yaml
```diff
@@ -9,65 +10,69 @@ metadata:
...
+
+{{ if not $frontend.kubeSecretRef }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $frontend.kubeSecretRef | default "infisical-frontend" }}
+type: Opaque
+data:
+  {{- range $key, $value := $frontend.secrets }}
+  {{ $key }}: {{ $value | b64enc }}
```
maidul98, line 76: I think we should wrap the value in quotes before `b64enc` otherwise we get 

```
rror: UPGRADE FAILED: template: infisical/templates/frontend-deployment.yaml:76:26: executing "infisical/templates/frontend-deployment.yaml" at <b64enc>: wrong type for value; expected string; got float64
```

maidul98:
Everything else looks great! I was able to run it on digital ocean. However, when i did 

```
helm upgrade infisical .  --values /Users/maidulislam/Desktop/k8-values/dev/values.yaml --recreate-pods
```

I expected the data in the volume to presist, but reloading my page seems like the data was lost. Fyi this is on Digital ocean (I was able to see the PVC there so it definitely got created)

helm-charts/infisical/templates/backend-deployment.yaml
```diff
@@ -70,3 +61,17 @@ spec:
...
       port: 4000
       targetPort: 4000 # container port 
+
+---
+
+{{ if not $backend.kubeSecretRef }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $backend.kubeSecretRef | default "infisical-backend" }}
```
jon4hz, line 71: Shouldn't the name default to `include "infisical.backend.fullname" .`? Otherwise multiple helm install without customization will have colliding names.

helm-charts/infisical/templates/frontend-deployment.yaml
```diff
@@ -9,65 +10,69 @@ metadata:
...
+      nodePort: {{ $frontend.service.nodePort }}
+      {{- end }}
+
+---
+
+{{ if not $frontend.kubeSecretRef }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $frontend.kubeSecretRef | default "infisical-frontend" }}
```
jon4hz, line 72: Same as for the backend secret. The default should probably be `include "infisical.frontend.fullname" .`

helm-charts/infisical/templates/backend-deployment.yaml
jon4hz: I think the variable `backendEnvironmentVariables` was completely removed. This will break existing installations.

jon4hz:
While I appreciate the effort regarding the mongodb persistent, I suggest to refactor that part completely and implement #182 instead. Bitnamis mongodb chart is extremely good and offers much more flexibility. 

If the bitnami mongodb chart is is on infisicals roadmap anyway, it would be easier to implement that now. Currently people probably don't use the mongo part of this chart due to the missing persistence. But if that changes, it will be much harder to switch later.

helm-charts/infisical/templates/mongodb-deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- $mongodb := .Values.mongodb }}
 apiVersion: apps/v1
```
zifeo, line 2: Can we make this deployment optional (in case we want to use an external database)?

maidul98: I think what we can do is if `externalMongoDBConnectionString` is defined then no need to spin this deployment up. What do you think @Grraahaam

helm-charts/infisical/templates/backend-deployment.yaml
jon4hz: well, it's up to @maidul98 if you want to keep this non breaking or not. I just wanted to point it out 😅

PR #279, opened 2023-02-01
CI Improvements 

jon4hz:


.goreleaser.yaml
```diff
@@ -14,7 +14,9 @@ before:
 builds:
   - id: darwin-build
     binary: infisical
-    ldflags: -X github.com/Infisical/infisical-merge/packages/util.CLI_VERSION={{ .Env.CLI_VERSION }}
+    ldflags: -X github.com/Infisical/infisical-merge/packages/util.CLI_VERSION={{ .Version }}
```
maidul98, line 17: Does .Version hold the tag by default? Any documentation on this

.goreleaser.yaml
```diff
@@ -25,10 +27,14 @@ builds:
       - goos: darwin
         goarch: "386"
     dir: ./cli
+
   - id: all-other-builds
     env:
       - CGO_ENABLED=0
     binary: infisical
+    ldflags: -X github.com/Infisical/infisical-merge/packages/util.CLI_VERSION={{ .Version }}
```
maidul98, line 35: good catch, i forgot to add it here

.goreleaser.yaml
jon4hz: btw that also allows you to test the goreleaser build locally, because you don't have to depend on any variable injection from github action :)

maidul98: nice

jon4hz:
ffda30b might be a bit opinionated but adding a `-devel` prefix allows you to distinguish actual releases from snapshots built with `goreleaser build --clean --snapshot`.

.goreleaser.yaml
maidul98: I'm wondering how tag github tag maps to .Version. Where is the defined os is this parsed automatically and set to .Version

jon4hz: It's part of goreleasers template magic :)
Actually I think it just reads the last git tag and removes the "v" prefix.

maidul98 (approved):
good stuff! 🚀

maidul98:
It looks like when i install via brew with your change, we get
```
Error: An exception occurred within a child process:
  NameError: undefined local variable or method `infisical' for #<Formulary::FormulaNamespaceba3a130640a0e57a00d3373d0bbc296b::Infisical:0x00007fca3cc55b08>
```
I will roll back so our users do not downloaded this version

jon4hz:
yikes, sorry for the trouble. I'll take a look this evening an make a PR

PR #281, opened 2023-02-01
fix homebrew

jon4hz:
Hi, I'm not 100% sure but this should fix the homebrew error.


jon4hz:
I think missing quotes in the `install` section.

maidul98 (approved):
Okay, giving it a go

PR #295, opened 2023-02-03
add new modal to compare secrets across environments

mocherfaoui:
demo: https://www.loom.com/share/21139494f7e2435c9d2cd29296af4e75


frontend/src/components/dashboard/SideBar.tsx
```diff
@@ -7,9 +7,12 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import SecretVersionList from '@app/ee/components/SecretVersionList';
 
+import { WorkspaceEnv } from '~/pages/dashboard/[id]';
+
```
mocherfaoui, line 11: @mv-turtle 
wouldn't be better to move some types to `src/utils/types.ts` for example? the line above will throw an eslint error

frontend/src/components/utilities/secrets/getSecretsForProject.ts
```diff
@@ -29,8 +29,8 @@ interface SecretProps {
 
 interface FunctionProps {
   env: string;
-  setIsKeyAvailable: any;
-  setData: any;
+  setIsKeyAvailable?: any;
+  setData?: any;
```
mocherfaoui, line 33: imo, `setIsKeyAvailable` and `setData` are unnecessary here. we can return `{ secrets, isKeyAvailable }` and set state after getting the return value instead of passing these functions.
what do you think @mv-turtle ?

frontend/src/components/dashboard/SideBar.tsx
mv-turtle: Yes, I just fixed that. I think that was a bug, it shouldn't have been this way I think

frontend/src/components/utilities/secrets/getSecretsForProject.ts
mv-turtle: I agree! Just need to make sure we check that it works in all the cases. We should definitely try to optimize this!

mv-turtle (approved):
This is really awesome! Really love this! Thank you :)

PR #315, opened 2023-02-08
Feat new layout synced with api changes

akhilmhdh:


mv-turtle (approved):
There were a couple bugs related to organizations and projects, but I have fixed them now. I have also changed the style quite a bit - let me please know if you think it's good or want to revert something :)

PR #328, opened 2023-02-12
feat(ui): add new button style, improved select ui and linted app layout

akhilmhdh:
@mv-turtle As discussed with you


mv-turtle (approved):
Looks great to me!!

Issue #325, opened 2023-02-10
Append environment name to the dashboard page url


maidul98:
Currently, the url for projects page looks like this regradress of what environment you are in 

```
/dashboard/63e5ea1e7a65573dc1e20ee9
```

This is fine, but when you refresh the page you lose the environment you were in and then have to manually navigate to it. It would be better if it was more like this

```
/dashboard/63e5ea1e7a65573dc1e20ee9/{env name}
```

Example 

```
/dashboard/63e5ea1e7a65573dc1e20ee9/dev
```

RashidUjang:
Just throwing this out there, would it be sufficient if we were to use some sort of browser storage (IndexedDB or Local Storage) to store the environment of the user for this instead of changing the URL?

dangtony98:
@RashidUjang Why not include it in the URL?

RashidUjang:
@dangtony98 I'm just exploring other ideas as I think it'd look neater not having in the URL as its sort of a temporary state that's useful only in the context of the user's current session (compared it with the Postman web client).

Just as I'm typing this though, having it in the URL does have an added benefit of making it easy to save & share a URL to go directly to an environment of a particular ID. 

Feel free to dictate the direction, just adding some discourse!

dangtony98:
Ah ya that makes sense.

The main benefit I see from having the `env` be in the URL is having the shareability component as you mentioned for sure!

And yea down for discourse as well; I think @maidul98 can chime in here.

esau-morais:
May I work on this issue?

maidul98:
@RashidUjang Yes, I think having the env in the URL allows for better readability and keeps place on the dashboard if the user refreshes.

@esau-morais please, that would be great!

RashidUjang:
@maidul98 Sounds good and yes that definitely would ease sharing!

mv-turtle:
This is implemented now thanks to @esau-morais :)

Issue #181, opened 2022-12-28
🌎 Help us translate Infisical 


vmatsiiako:
### 🌎 We want to let as many people use Infisical as possible! Help us translate it! 🚀

There are two main components that need to be translated for Infisical: README.md and Web UI. README.md is probably higher priority as it allows more people to initially understand what Infisical is. 

Here are some of the languages we are looking for: (👀 - looking for; ✅ - ready; ⚠️ - needs minor revision)

(if your language is not in this list, we would still love to have it! Please, let us know :))

|  Language | README.md status | Web UI status |
| ------------- | ------------- | ------------- |
| Arabic | 👀 | 👀 |
| English | ✅  | ✅ |
| French | 👀 | ✅ |
| German | ✅ | 👀 |
| Hindi | 👀 | 👀 |
| Indonesian | ✅ | 👀 |
| Italian | 👀 | 👀 |
| Japanese | 👀 | 👀 |
| Korean | ✅ | ✅ |
| Malay | 👀 | 👀 |
| Portuguese | 👀 | 👀 |
| Portuguese - Brazil | ✅ | ✅ |
| Spanish | ✅ | 👀 |
| Simplified Chinese | 👀 | 👀 |
| Traditional Chinese | 👀 | 👀 |
| Turkish | ✅ | ✅ |
| Ukrainian | 👀 | 👀 |
| Vietnamese | 👀 | 👀 |

For Web UI, you would need to create a new directory [here](https://github.com/Infisical/infisical/tree/main/frontend/public/locales).
For README.md, you can add them to [this folder](https://github.com/Infisical/infisical/tree/main/i18n). 

If you have any questions, just shoot them in the comments to this issue or [join our Slack](https://join.slack.com/t/infisical-users/shared_invite/zt-1kdbk07ro-RtoyEt_9E~fyzGo_xQYP6g).

P.S. In future, we will also have docs in different languages, but that seems like too much work right now :)

JoaoVictor6:
I can translate the UI into Portuguese!

mv-turtle:
That would be really amazing @JoaoVictor6! Appreciate your help :) Let me know if you need anything

Gabriellopes232:
Hello, I'm Brazilian too! I would love to be able to help @JoaoVictor6 in this issue to translate the UI to Portuguese!

mv-turtle:
[...]

PR #333, opened 2023-02-15
Add missing `forgot-password` in pt-BR and save selected env on URL

esau-morais:
### Description

1. Add missing `forgot-password` on pt-BR #181
2. Save selected workspace environment on URL using `?env=` query param, so it can be sharable and prevent from resetting on refresh -> #325 

### Screenshots

Before:
before

After:
image
---
2

Before
image

After
image


mv-turtle:
This is really awesome! I found one minor bug: if you have a query param in your URL, layout will not should what tab is currently active. I just fixed it though :)

Screenshot 2023-02-20 at 20 26 39

mv-turtle (commented):
Thank you so much for helping with this! This is really amazing! :)

mv-turtle (approved):

PR #337, opened 2023-02-17
Revamped the org settings page

akhilmhdh:


mv-turtle:
Hey @akhilmhdh! It looks really good. The only problem I found is that deleting a user doesn't work as expected - it starts the following infinite spinner. If I wait long enough, a user is deleted but the spinner still keeps running. Have you experienced it? 

Screenshot 2023-02-18 at 13 00 24

akhilmhdh:
@mv-turtle bingo. That's what i was stating. That spinner is staying like that because network call is not yet finished. 

Check ur network tab when ur making it. It stays until timeout

mv-turtle:
Hmmmm, this is very strange. @dangtony98 do you maybe know what could be going wrong?

mv-turtle (approved):
Merging! Works really great now! There is one minor issue because there is no loading for when you're adding a user, but I will try to fix this now after merging.

PR #341, opened 2023-02-20
Refactored and improved PDF backup generation

umrak11:
Added parameters to jsPDF so the generation of PDF is much smaller than before - around (40kb vs 4.8 mb before).
Refactored to more functions, so you know better where header is generated and other content.


mv-turtle:
Just went over everything - works perfectly! Thank you so much for your contribution @umrak11  :)

mv-turtle (approved):

PR #354, opened 2023-02-22
Table loading state and empty states

akhilmhdh:
Fix: missing loading state on member addition in button
fix: whitespace issue in project settings page as discussed with @mv-turtle


mv-turtle (approved):
Looks great!

PR #371, opened 2023-02-24
Improve env filter

jon4hz:
# Description 📣

This PR adds some more reserved env vars as well as support for reserved prefixes like `XDG_` which shouldn't be overwritten by infisical. 
There are probably more vars that should be reserved but I think the current ones should cover most cases. 

The PR also micro-improves the generation of the secrets map by preallocating its size.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

https://github.com/jon4hz/infisical/blob/4a72d725b1038ea86d1cc9888bf78216658682bc/cli/packages/cmd/cmd_test.go

---

### DIFF
--- a/cli/packages/cmd/cmd_test.go
+++ b/cli/packages/cmd/cmd_test.go
@@ -1,0 +1,49 @@
+package cmd
+
+import (
+	"testing"
+
+	"github.com/Infisical/infisical-merge/packages/models"
+)
+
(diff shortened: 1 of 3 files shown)


maidul98 (approved):
Excellent! This was much needed 💥

PR #373, opened 2023-02-24
fix: properly support default environment

jon4hz:
# Description 📣

I'm not sure if this is intended but at the moment `infisical` completely ignores the `defaultEnvironment` from the workspace file. This PR adds proper support for this config option. At first, it checks the env-branch mapping and if it can't find anything there, it returns the `defaultEnvironment` instead of an empty string. 

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Added some test cases to cover the changes. 

---


maidul98 (approved):
Looks good! I added the docs to complete it

PR #374, opened 2023-02-24
fix: always execute cmd in subshell

jon4hz:
# Description 📣

Currently `infisical` doesn't support executing aliases as single command. 
The following script fails but should not imo:

```bash
alias ll="ls -l"
infisical run -- ll # fails
infisical run --command ll # works
```

The script fails because infisical tries to lookup `ll` in the `$PATH` which is impossible, because it's only an alias. If I executed it using `infisical run --command ll` it would have worked, because `infisical` seems to spawn a subshell when the `--command` arg is used. I think this is a bit intuitive and confusing.

This can be fixed if we always start a subshell to execute the command. Is there any reason that speaks against doing that? 

I also fixed a bug on windows, if you use something like "Git Bash", which sets the `$SHELL`  variable. At the moment `infisical` ignores that, even though it would be perfectly valid. 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Can be tested with the script in the description. 

---

### DIFF
--- a/cli/packages/cmd/run.go
+++ b/cli/packages/cmd/run.go
@@ -161,11 +161,14 @@ func init() {
 
 // Will execute a single command and pass in the given secrets into the process
 func executeSingleCommandWithEnvs(args []string, secretsCount int, env []string) error {
-	command := args[0]
-	argsForCommand := args[1:]
+	shell := subShellCmd()
 	color.Green("Injecting %v Infisical secrets into your application process", secretsCount)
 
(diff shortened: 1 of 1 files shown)


maidul98 (approved):
This is a great improvemen as it allows us to run also run chained commands with just `run --"chain command 1 && chained command 2"`! I agree the --command was not intuitive. Great work!

PR #389, opened 2023-02-27
Revert "fix: always execute cmd in subshell"

maidul98:
Reverts Infisical/infisical#374


maidul98:
Users having issues with injecting secrets

PR #394, opened 2023-02-28
Update create-secrets.mdx

ha-sante:
# Description 📣
I am making this change to draw your attention to this as it seems that some variables are used wrong.

- Specifically this section:
`util.decodeBase64(encryptedProjectKey), 
util.decodeBase64(encryptedProjectKey.nonce), 
util.decodeBase64(encryptedProjectKey.sender.publicKey), 
util.decodeBase64(PSWD)`

- Imported tweetnacl as well so it's easier to understand and for the code to make sense from the get go.

## Type ✨

- [X] Bug fix
- [ ] New feature
- [X] Breaking change
- [X] Documentation

# Tests 🛠️

- I imported the code, executed it and fixed these errors and blind sides before I could use it.

# Alert ⚠️
- I have a bug somewhere in the decrypt function so test out these changes with your context and solve for  if the changes I made doesn't actually work in practice and that we need some new variables or properties etc.

### DIFF
--- a/docs/api-reference/overview/examples/create-secrets.mdx
+++ b/docs/api-reference/overview/examples/create-secrets.mdx
@@ -23,6 +23,7 @@ Prerequisites:
 ```js
 const crypto = require('crypto');
 const axios = require('axios');
+const nacl = require('tweetnacl');
 
 const ALGORITHM = 'aes-256-gcm';
 const BLOCK_SIZE_BYTES = 16;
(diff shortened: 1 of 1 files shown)


ha-sante:
More to this, that I happened to have fixed earlier.

- When calling the decrypt function, the params need to be set right.
- What I met was, the calling of this function passed an object, whiles the params of the function itself was set as normal, variable after variable.


`
const decrypt = (ciphertext, iv, tag, secret) => {
	const decipher = crypto.createDecipheriv(
		ALGORITHM,
		secret,
		Buffer.from(iv, 'base64')
	);
	decipher.setAuthTag(Buffer.from(tag, 'base64'));

	let cleartext = decipher.update(ciphertext, 'base64', 'utf8');
	cleartext += decipher.final('utf8');

	return cleartext;
}

const privateKey = decrypt({
		ciphertext: user.encryptedPrivateKey,
		iv: user.iv,
		tag: user.tag,
		secret: PSWD.slice(0, 32).padStart(32, '0');
});
`

dangtony98:
Merging this but the docs for working with secret CRUD operations will be revised soon entirely to be service-token based instead of via API Key.

There will be a full documentation change here in the next few days!

PR #401, opened 2023-03-03
Update check

jon4hz:
# Description 📣

Hi, I changed a few things regarding the update check infisical does on each run. 
- Use `io` in favor of `ioutil` because ioutil is deprecated
- Added missing error check and log possible errors to the debug log
- Added an env variable to disable the update check

I think it should be possible to disable the update check completely. If you use `infisical` in an air-gapped environment, that check will never succeed and it slows down the execution, because the cli will wait until it hits a timeout while querying github. 

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Run `infisical` in an air-gapped environment with a self hosted instance. 

```sh
infisical run -- echo # will be extremely slow

export INFISICAL_DISABLE_UPDATE_CHECK=1
infisical run -- echo # will be much faster
```

---

### DIFF
--- a/cli/packages/util/check-for-update.go
+++ b/cli/packages/util/check-for-update.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"os"
 	"os/exec"
@@ -55,7 +55,7 @@ func getLatestTag(repoOwner string, repoName string) (string, error) {
 
 	defer resp.Body.Close()
 
-	body, err := ioutil.ReadAll(resp.Body)
+	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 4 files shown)


maidul98:
I think this is a good place to improve the command execution time but it does add a bit of manual work for the user. I think a slightly slower execution for local development is okay. For production runs, i think this is not acceptable. I'm wondering if there is a way to auto detect if its running in production such as in a docker container and then disable the check. Do you have any thoughts for auto detecting production run so that by default we have the check disabled?

jon4hz:
To be honest, I have no idea how we could detect an execution environment reliably or if that's even possible.
Disabling that check by default doesn't make much sense imo, because I don't think much users would use a flag to explicitly enable it. 

Alternatively you could execute that check only when the user runs `infisical --version`. That's what `terraform` does for example.

maidul98:
Right. Let's keep this for now and if we come across anything better along the way, we'll update it. Good stuff! 💥

maidul98 (approved):

PR #424, opened 2023-03-10
docs: python docs for example CRUD

Aashish-Upadhyay-101:
python docs CRUD example.


dangtony98:
Nice job!

I just tested all these CRUD operations and they worked well! 

I hope you learned a lot on how some of the cryptography works with secrets in Infisical 😄

Issue #406, opened 2023-03-06
[CLI] Long secret names/values mess up the output


jon4hz:
### Describe the bug
Long secret names or values mess up the output from the cli when using `infisical secrets` and `infisical secrets get xyz` 

### To Reproduce
1. Create a very long secret
2. `infisical secrets` 

### Screenshots
![image](https://user-images.githubusercontent.com/26183582/222995311-f151ec2f-2266-48fa-a454-d2a56b20d164.png)

### Additional context
Proposal:

What do you think about the following changes:
- `infisical secrets` truncates the value of the secret in the table, to keep a readable table
- `infisical secrets get xyz` doesn't render the table at all and only prints the secret if a single secret is requests.  If multiple secrets are requested, the table also truncates secrets which are too long. 
- Alternatively, the table could support multiline rows but I'm not sure if that is very helpful. For example, if you store something like a certificate as secret and you want to copy it, it would much easier to use `infisical secrets get myCertificate` instead of copy pasting line after line.

PR #426, opened 2023-03-10
Improve infisical secrets command

jon4hz:
# Description 📣

As described in #406, large secrets completely mess up the table rendering. This PR implements parts of my proposed solution by truncating secrets which are to long to be displayed in a single line. The table is only truncated if stdout is a terminal. If you redirect the command to a file, it wont truncate.
By using `github.com/muesli/ansi` to determine the width of a string, we can easily handle all kind of cases like having chars with double width.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/cli/packages/visualize/visualize.go
+++ b/cli/packages/visualize/visualize.go
@@ -31,9 +31,11 @@ const (
 
 // Given headers and rows, this function will print out a table
 func Table(headers [3]string, rows [][3]string) {
-	// if we're not in a terminal, don't truncate the secret value
-	shouldTruncate := isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd())
+	// if we're not in a terminal or cygwin terminal, don't truncate the secret value
+	shouldTruncate := isatty.IsTerminal(os.Stdout.Fd())
 
+	// This will return an error if we're not in a terminal or
+	// if the terminal is a cygwin terminal like Git Bash.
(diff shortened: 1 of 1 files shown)


cli/packages/visualize/visualize.go
```diff
@@ -16,8 +22,35 @@ type TableOptions struct {
...
+const (
+	// combined width of the table borders and padding
+	borderWidths = 10
+	// char to indicate that a string has been truncated
+	ellipsis = "…"
+)
+
 // Given headers and rows, this function will print out a table
-func Table(headers []string, rows [][]string) {
+func Table(headers [3]string, rows [][3]string) {
```
maidul98, line 33: This method was intended to be used for any number of rows and colums. Is there a reason why you are restricting it to 3? I understand that 3 is for the case of displaying secrets, but i feel like we lose the ability to use it for any number of rows/cols

jon4hz: I made the 3 rows explicit because adding new rows will break the rendering. 
Since the terminal has only limited width and there's no horizontal scrolling, you have to calculate available space yourself and explicitly define where to truncate values. 
A generic table will almost never work unless you'd specify rules like "always truncate the first column" for example. But that doesn't make much sense imo.

maidul98 (approved):
Looks good! I think it'd be nice if we had a flag to disable it so that when we get specific secrets, the full secret shows

PR #440, opened 2023-03-16
test setup

Aashish-Upadhyay-101:
# Tests 🛠️


backend/setupTests.ts
dangtony98, line 1: Is there a reason why we need this file to be in the root of the `backend` folder? — Can we not fit this into `backend/tests` folder?

Aashish-Upadhyay-101: so basically, the tests folder is for the actual unit tests, the ```setupTest.ts``` file is responsible for setting up the ***test*** database before the actual tests and make sure to close that database after afterall tests.
so the main reason is to put in in the root of backend is not to mix that with the actual tests, coz its just a setup before the test.

dangtony98: I'm wondering though: Would it make sense to put setupTest.ts in something like `tests/index.ts` and then inside `tests` we can have subfolders for different controllers/services etc.?

```
- tests/
   - index.ts (setup before and after)
   - services/
      - exampleService.ts
```

backend/jest.config.ts
```diff
@@ -0,0 +1,9 @@
+export default {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  collectCoverageFrom: ['src/*.{js,ts}', '!**/node_modules/**'],
+  modulePaths: ['<rootDir>/src'],
+  testMatch: ['<rootDir>/tests/**/*.test.ts'],
+  setupFiles: ['<rootDir>/test-resources/env-vars.js'],
```
dangtony98, line 7: TODO: Add remaining required environment variables for testing (e.g. `ENCRYPTION_KEY`).

backend/setupTests.ts
```diff
@@ -0,0 +1,11 @@
+import { beforeAll, afterAll } from '@jest/globals';
+import { DatabaseService } from './src/services';
+
+beforeAll(async () => {
```
dangtony98, line 4: TODO: Remove because `DatabaseService.initDatabase()` and `DatabaseService.closeDatabase()` are already done inside the `main()` server code.

backend/jest.config.ts
dangtony98, line 7: TODO: update path of `setupSetup.test` to be inside `tests`

Aashish-Upadhyay-101: hey or maybe we add those variables which the test actually needs, I mean define the variables when the corresponding tests need that.

dangtony98: @Aashish-Upadhyay-101 Sounds good!

PR #464, opened 2023-03-26
unit tests for utils/posthog and utils/crypto

Aashish-Upadhyay-101:
unit tests for `utils/posthog.ts` and `utils/crypto.ts`

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -57,7 +57,7 @@
     "lint-and-fix": "eslint . --ext .ts --fix",
     "lint-staged": "lint-staged",
     "pretest": "docker compose -f test-resources/docker-compose.test.yml up -d",
-    "test": "cross-env NODE_ENV=test jest --testTimeout=10000 --detectOpenHandles",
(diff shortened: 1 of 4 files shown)


dangtony98:
Hey @Aashish-Upadhyay-101!

The tests provided seem to cover basic scenarios for the encryption and decryption functions. There's still some room for improvement to make the tests more robust and comprehensive:

1. Test edge cases and input validation:

- Test with empty strings or null inputs for the plaintext, key, publicKey, and privateKey.
- Test with input strings containing special characters, Unicode characters, or very long strings.

2. Test the encryption and decryption functions together:

- Encrypt and decrypt using the same key and verify that the decrypted plaintext is the same as the original plaintext.
- Encrypt and decrypt using different keys and expect the decryption to fail.

3. Test error messages:

- When testing for errors, check if the error message is as expected. This helps ensure that the correct error is being thrown.

4. Test for tampering:

- Modify the ciphertext, nonce, or tag before attempting to decrypt and check if decryption fails or if an error is thrown.

Aashish-Upadhyay-101:
@dangtony98 added  improvement tests ! 🚀

PR #466, opened 2023-03-29
Typo: User remove from Organization notification typo

Aashish-Upadhyay-101:
There is a typo, when a user is deleted from an organization, a notification appears on bottom right corner of the screen which says "***Successfully removed used from org***". 

This PR fixes this typo. "***used -> user***"
🚀 🚀


maidul98 (approved):
Good find!

Issue #340, opened 2023-02-20
Multiple CLI Users


ngregrichardson:
### Feature description
Allow the CLI to support multiple user accounts logged in simultaneously.

### Why would it be useful?
This would allow users to easily switch between scopes (work versus personal projects) instead of re-authenticating anytime they want to run a project in a different scope.

### Additional context
The two ways I imagine it could work are:
- Automatically switching between profiles based on which would work with the current workspace. So if you run `infisical run` in a directory with a workspace id that only has access from some accounts, it would use that account. This would be ideal, but I can imagine problems with multiple accounts with different permissions on different workspaces.
- Adding a new command (maybe `infisical switch` or something similar) that would let you switch between logged-in accounts. Less ideal, but still makes the workflow of re-authenticating with your password easier.

quinton11:
Has this been taken up by anyone yet?

maidul98:
Hey @quinton11 no one has picked this up yet

quinton11:
Sure, will work on this

maidul98:
Awesome! Please join our slack if you'd like to run your ideas by me or have any questions https://join.slack.com/t/infisical-users/shared_invite/zt-1kdbk07ro-RtoyEt_9E~fyzGo_xQYP6g

PR #481, opened 2023-04-04
feat: CLI support for switching between multiple logged in user accounts

quinton11:
See #340. 

This feature allows users  login with multiple accounts and switch between them using [infisical switch] without having to re-authenticate anytime they intend to switch scopes.

### DIFF
--- a/cli/packages/cmd/login.go
+++ b/cli/packages/cmd/login.go
@@ -48,14 +48,24 @@ var loginCmd = &cobra.Command{
 			util.HandleError(err)
 		}
 
-		if currentLoggedInUserDetails.IsUserLoggedIn && !currentLoggedInUserDetails.LoginExpired && len(currentLoggedInUserDetails.UserCredentials.PrivateKey) != 0 {
-			shouldOverride, err := shouldOverrideLoginPrompt(currentLoggedInUserDetails.UserCredentials.Email)
+		addUser := false
+		if currentLoggedInUserDetails.UserCredentials.Email != "" {
+			addUser, err = addNewUserPrompt()
 			if err != nil {
 				util.HandleError(err)
 			}
+		}
 
-			if !shouldOverride {
-				return
+		if !addUser {
+			if currentLoggedInUserDetails.IsUserLoggedIn && !currentLoggedInUserDetails.LoginExpired && len(currentLoggedInUserDetails.UserCredentials.PrivateKey) != 0 {
+				shouldOverride, err := shouldOverrideLoginPrompt(currentLoggedInUserDetails.UserCredentials.Email)
+				if err != nil {
+					util.HandleError(err)
+				}
+
+				if !shouldOverride {
+					return
+				}
 			}
 		}
 
@@ -340,6 +350,19 @@ func getFreshUserCredentials(email string, password string) (*api.GetLoginOneV2R
 	}
 
 	return &loginOneResponseResult, &loginTwoResponseResult, nil
+}
+
+func addNewUserPrompt() (bool, error) {
+	prompt := promptui.Select{
+		Label: "Infisical detects previous logged in users. Would you like to add a new user? Select[Yes/No]",
+		Items: []string{"No", "Yes"},
+	}
+
+	_, result, err := prompt.Run()
+	if err != nil {
+		return false, err
+	}
+	return result == "Yes", err
 }
 
 func shouldOverrideLoginPrompt(currentLoggedInUserEmail string) (bool, error) {
--- a/cli/packages/util/config.go
+++ b/cli/packages/util/config.go
@@ -31,9 +31,25 @@ func WriteInitalConfig(userCredentials *models.UserCredentials) error {
 		return fmt.Errorf("writeInitalConfig: unable to write config file because [err=%s]", err)
 	}
 
+	//if empty
+	if existingConfigFile.LoggedInUsersEmail == nil {
+		existingConfigFile.LoggedInUsersEmail = []string{}
+	}
+
+	//if profile exists
+	if len(existingConfigFile.LoggedInUsersEmail) > 0 {
+		ok := Contains(existingConfigFile.LoggedInUsersEmail, userCredentials.Email)
+		if !ok {
+			existingConfigFile.LoggedInUsersEmail = append(existingConfigFile.LoggedInUsersEmail, userCredentials.Email)
+		}
+	} else {
+		existingConfigFile.LoggedInUsersEmail = append(existingConfigFile.LoggedInUsersEmail, userCredentials.Email)
+	}
+
 	configFile := models.ConfigFile{
-		LoggedInUserEmail: userCredentials.Email,
-		VaultBackendType:  existingConfigFile.VaultBackendType,
+		LoggedInUserEmail:  userCredentials.Email,
+		VaultBackendType:   existingConfigFile.VaultBackendType,
+		LoggedInUsersEmail: existingConfigFile.LoggedInUsersEmail,
 	}
 
 	configFileMarshalled, err := json.Marshal(configFile)
@@ -176,7 +192,7 @@ func GetConfigFile() (models.ConfigFile, error) {
 	return configFile, nil
 }
 
-// Write a ConfigFile to disk. Raise error if unable to save the model to ask
+// Write a ConfigFile to disk. Raise error if unable to save the model to disk
 func WriteConfigFile(configFile *models.ConfigFile) error {
 	fullConfigFilePath, fullConfigFileDirPath, err := GetFullConfigFilePath()
 	if err != nil {
--- a/cli/packages/cmd/switch.go
+++ b/cli/packages/cmd/switch.go
@@ -1,0 +1,93 @@
+package cmd
+
+import (
+	"errors"
+
+	"github.com/Infisical/infisical-merge/packages/util"
+	"github.com/manifoldco/promptui"
+	"github.com/spf13/cobra"
+)
+
+var switchCmd = &cobra.Command{
+	Use:                   "switch",
+	Short:                 "Used to switch between Infisical profiles",
+	DisableFlagsInUseLine: true,
+	Example:               "infisical switch",
+	Args:                  cobra.ExactArgs(0),
+	PreRun: func(cmd *cobra.Command, args []string) {
+		util.RequireLogin()
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		//get previous logged in profiles
+		loggedInProfiles, err := getLoggedInUsers()
+		if err != nil {
+			util.HandleError(err, "[infisical switch]: Unable to get logged Profiles")
+		}
+
+		//prompt user
+		profile, err := LoggedInUsersPrompt(loggedInProfiles)
+		if err != nil {
+			util.HandleError(err, "[infisical switch]: Prompt error")
+		}
+
+		//write to config file
+		configFile, err := util.GetConfigFile()
+		if err != nil {
+			util.HandleError(err, "[infisical switch]: Unable to get config file")
+		}
+
+		configFile.LoggedInUserEmail = profile
+		ok := util.Contains(configFile.LoggedInUsersEmail, profile)
+		if !ok {
+			configFile.LoggedInUsersEmail = append(configFile.LoggedInUsersEmail, profile)
+		}
+
+		err = util.WriteConfigFile(&configFile)
+		if err != nil {
+			util.HandleError(err, "")
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(switchCmd)
+}
+
+func getLoggedInUsers() ([]string, error) {
+	loggedInProfiles := []string{}
+
+	if util.ConfigFileExists() {
+		configFile, err := util.GetConfigFile()
+		if err != nil {
+			return loggedInProfiles, err
+		}
+
+		//get logged in profiles
+		//
+		if configFile.LoggedInUsersEmail == nil {
+			loggedInProfiles = append(loggedInProfiles, configFile.LoggedInUserEmail)
+		} else {
+			if len(configFile.LoggedInUsersEmail) > 0 {
+				loggedInProfiles = append(loggedInProfiles, configFile.LoggedInUsersEmail...)
+			}
+		}
+		return loggedInProfiles, nil
+	} else {
+		//empty
+		return loggedInProfiles, errors.New("couldn't retrieve config file")
+	}
+}
+
+func LoggedInUsersPrompt(profiles []string) (string, error) {
+	prompt := promptui.Select{Label: "Which of your Infisical profiles would you like to use",
+		Items: profiles,
+		Size:  7,
+	}
+
+	idx, _, err := prompt.Run()
+	if err != nil {
+		return "", err
+	}
+
+	return profiles[idx], nil
+}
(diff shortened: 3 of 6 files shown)


maidul98:
Thanks for the PR. I will review it this week!

maidul98:
Just reviewd the PR, works great! I'm going to make some minor changes and then i'll merge it

maidul98:
Just wanted to also mention that the community has been asking for the host API to be persisted, and I think it's a good idea. By attaching the domain to use for APIs when we add a new user, we can fulfill this request. It would be great if we could do this sooner rather than later because we'll need to make sure everything is compatible with the Config data structure we save. Would you like to give it a shot? If you're busy, no worries, just let me know, and I'm happy chime in on it.

I think what should happen is we should allow the user to choose a host, the default being infisical cloud before login and then continue the login steps. Once done, save the host along with the user email and then we can change this to point to the config https://github.com/Infisical/infisical/blob/main/cli/packages/cmd/root.go#L35

quinton11:
Great, it wouldn't be a problem.. I'll look into it. Thanks!

> Just wanted to also mention that the community has been asking for the host API to be persisted, and I think it's a good idea. By attaching the domain to use for APIs when we add a new user, we can fulfill this request. It would be great if we could do this sooner rather than later because we'll need to make sure everything is compatible with the Config data structure we save. Would you like to give it a shot? If you're busy, no worries, just let me know, and I'm happy chime in on it.
> 
> I think what should happen is we should allow the user to choose a host, the default being infisical cloud before login and then continue the login steps. Once done, save the host along with the user email and then we can change this to point to the config https://github.com/Infisical/infisical/blob/main/cli/packages/cmd/root.go#L35

quinton11:
Changes:
On user logins, user's are prompted for the preffered hosting domain associated with the account (default being infisical cloud). Config data structure supports multiple users along with their domains. 
An added user sub command allowing users change the domain for a logged in account (in case of a domain url change)

cli/packages/cmd/login.go
```diff
@@ -48,8 +54,17 @@ var loginCmd = &cobra.Command{
 			util.HandleError(err)
 		}
 
+		// addUser := false
```
maidul98, line 57: I think you forgot to delete these?

cli/packages/cmd/login.go
```diff
@@ -252,6 +274,51 @@ func init() {
...
+func askForDomain() error {
+	//query user to choose between Infisical cloud or self hosting
+	options := []string{"Infisical Cloud", "Self Hosting"}
+	optionsPrompt := promptui.Select{
+		Label: "Select your hosting option",
+		Items: options,
+		Size:  2,
+	}
+
+	idx, _, err := optionsPrompt.Run()
```
maidul98, line 286: nit: is there a reason why you have choosen `idx`? I would think it's better to call it something like `selectedHostingOption`

cli/packages/cmd/login.go
```diff
@@ -252,6 +274,51 @@ func init() {
...
+		Validate: urlValidation,
+	}
+
+	domain, err := domainPrompt.Run()
+	if err != nil {
+		return err
+	}
+
+	//set api url
+	config.INFISICAL_URL = domain
```
maidul98, line 317: we should add a `/api` at the end because all of our default nginx configs route traffic from `/api` to the backend services

cli/packages/cmd/login.go
```diff
@@ -252,6 +274,51 @@ func init() {
...
+		_, err := url.ParseRequestURI(input)
+		if err != nil {
+			return errors.New("this is an invalid url")
+		}
+		return nil
+	}
+
+	//else run prompt to enter domain
+	domainPrompt := promptui.Prompt{
+		Label:    "Domain",
```
maidul98, line 307: I wonder if domain is clear, when i think of domain i think of google.com not the https or http before it. Do you know if there is a way to provide a placeholder text with an example so users know?

cli/packages/cmd/login.go
```diff
@@ -342,16 +409,16 @@ func getFreshUserCredentials(email string, password string) (*api.GetLoginOneV2R
 	return &loginOneResponseResult, &loginTwoResponseResult, nil
 }
 
-func shouldOverrideLoginPrompt(currentLoggedInUserEmail string) (bool, error) {
+func userLoginMenu(currentLoggedInUserEmail string) (bool, error) {
 	prompt := promptui.Select{
-		Label: fmt.Sprintf("There seems to be a user already logged in with the email: %s. Would you like to override that login? Select[Yes/No]", currentLoggedInUserEmail),
-		Items: []string{"No", "Yes"},
+		Label: fmt.Sprintf("Current logged in user email: %s", currentLoggedInUserEmail),
```
maidul98, line 414: what do you think about also adding which host they are on such as `current logged in user email: maidul@ss.com on domain: app.infisical.com`?

cli/packages/cmd/user.go
```diff
@@ -0,0 +1,212 @@
...
+	},
+}
+
+func init() {
+	userCmd.AddCommand(domainCmd)
+	userCmd.AddCommand(switchCmd)
+	rootCmd.AddCommand(userCmd)
+}
+
+func getLoggedInUsers() ([]string, error) {
```
maidul98, line 151: i think it'd be great if you can write quick function descriptions in 1-2 sentences for all functions you create. It would help future contributors to quickly understand

cli/packages/cmd/user.go
```diff
@@ -0,0 +1,212 @@
...
+
+		err = util.WriteConfigFile(&configFile)
+		if err != nil {
+			util.HandleError(err, "")
+		}
+	},
+}
+
+var domainCmd = &cobra.Command{
+	Use:                   "domain",
```
maidul98, line 83: what do you think about putting command under `update` subcommand like this `infisical user update domain`?

cli/packages/util/config.go
```diff
@@ -31,9 +32,25 @@ func WriteInitalConfig(userCredentials *models.UserCredentials) error {
 		return fmt.Errorf("writeInitalConfig: unable to write config file because [err=%s]", err)
 	}
 
+	//if profiles exists
+	loggedInUser := models.LoggedInUser{
+		Email:  userCredentials.Email,
+		Domain: config.INFISICAL_URL,
+	}
+	if len(existingConfigFile.LoggedInUsers) > 0 {
```
maidul98, line 40: i think this can be simplified to:

```
if len(existingConfigFile.LoggedInUsers) == 0 || !ConfigContainsEmail(existingConfigFile.LoggedInUsers, userCredentials.Email) {
    existingConfigFile.LoggedInUsers = append(existingConfigFile.LoggedInUsers, loggedInUser)
}
```

cli/packages/util/config.go
```diff
@@ -31,9 +32,25 @@ func WriteInitalConfig(userCredentials *models.UserCredentials) error {
...
+		}
+	} else {
+		existingConfigFile.LoggedInUsers = append(existingConfigFile.LoggedInUsers, loggedInUser)
+	}
+
 	configFile := models.ConfigFile{
-		LoggedInUserEmail: userCredentials.Email,
-		VaultBackendType:  existingConfigFile.VaultBackendType,
+		LoggedInUserEmail:  userCredentials.Email,
+		LoggedInUserDomain: config.INFISICAL_URL,
```
maidul98, line 51: we set the current logged in user domain here but i don't think we update it in `LoggedInUsers` domain property  if the user email already exists. Let me know if i misunderstood

maidul98:
Overall, excellent work @quinton11! I think the Infisical community will really love this. I left some inital feedback, tomorrow i'll do more thorough testing to make sure all all edge cases are covered and existing users won't have any issues. 

Also I think both of these methods of setting the domain should override the user set domain, what do you think?
Screenshot 2023-04-18 at 9 14 11 PM

cli/packages/cmd/login.go
quinton11: Yep, my bad.

cli/packages/cmd/login.go
quinton11: `idx` referring to the selected index of the options. However yes `selectedHostingOption` would be more readable. Instead of dealing with the index, comparing the selected value directly should be better. Thanks

cli/packages/cmd/login.go
quinton11: Yeah, my initial thought was that will be added directly by the users. However adding a placeholder text example you suggested, as to the format they're to input would resolve the ambiguity.

cli/packages/cmd/login.go
quinton11: Yeah we could provide the `default` value in the prompt object. Example: `Default: 'Example: https://my-domain-example.com`

cli/packages/cmd/login.go
quinton11: Yeah this shouldn't be an issue, there is an added check for the current logged in user domain in https://github.com/Infisical/infisical/pull/481/files/be2cf54d6e138b8d82a5310be941f43310b9823f#diff-eb3a7252b09a5793a44fec4d5e033dda44f2fc86484c3d1a728bb57208058fda.

cli/packages/cmd/user.go
quinton11: Yeah, thanks

cli/packages/cmd/user.go
quinton11: Right, wouldn't be an issue, its more user readable

cli/packages/util/config.go
quinton11: Right, good eye! Will amend

quinton11:
> Overall, excellent work @quinton11! I think the Infisical community will really love this. I left some inital feedback, tomorrow i'll do more thorough testing to make sure all all edge cases are covered and existing users won't have any issues.
> 
> Also I think both of these methods of setting the domain should override the user set domain, what do you think? Screenshot 2023-04-18 at 9 14 11 PM

Thanks!
Yes I agree, however dealing with multiple profiles, the sub command was to allow users change the domain of a particular profile. What could happen is we could still use the command but check for the INFISICAL_API_URL exported variable, if there is a value, ask user to either use or override

maidul98:
> > Overall, excellent work @quinton11! I think the Infisical community will really love this. I left some inital feedback, tomorrow i'll do more thorough testing to make sure all all edge cases are covered and existing users won't have any issues.
> > Also I think both of these methods of setting the domain should override the user set domain, what do you think? Screenshot 2023-04-18 at 9 14 11 PM
> 
> Thanks! Yes I agree, however dealing with multiple profiles, the sub command was to allow users change the domain of a particular profile. What could happen is we could still use the command but check for the INFISICAL_API_URL exported variable, if there is a value, ask user to either use or override

I'm thinking we override by default for both `INFISICAL_API_URL` exported variable and and `--domain` flag. If either of those are present, we ignore the domain for the user. This is because existing users already expect these methods of setting the domain

cli/packages/cmd/login.go
maidul98: yes, good thought. Let's add the /api in the place holder text so they know

cli/packages/cmd/login.go
maidul98: or maybe `https://my-self-hosted-instance.com`? Feel free to go with what you think is best for developers to understand

cli/packages/cmd/login.go
maidul98: yep, no worries

maidul98:
Shoot me a ping when you have made new changes for the comments we disccused ^ and i'll review again! Thanks again

cli/packages/util/config.go
quinton11: Yep, addressed this

cli/packages/cmd/user.go
```diff
@@ -0,0 +1,248 @@
...
+
+		//prompt user
+		profile, err := LoggedInUsersPrompt(loggedInProfiles)
+		if err != nil {
+			util.HandleError(err, "[infisical user update domain]: Prompt error")
+		}
+
+		domain := ""
+		domainQuery := true
+		if config.INFISICAL_URL_MANUAL_OVERRIDE != util.INFISICAL_DEFAULT_API_URL {
```
maidul98, line 115: Can you explain why you choose to use `INFISICAL_URL_MANUAL_OVERRIDE` variable instead of   `config.INFISICAL_URL`? I see in `GetCurrentLoggedInUserDetails`, you set `INFISICAL_URL_MANUAL_OVERRIDE=config.INFISICAL_URL` so i'm curious if we need this or if there is a use case I didn't catch

cli/packages/util/credentials.go
```diff
@@ -87,6 +88,13 @@ func GetCurrentLoggedInUserDetails() (LoggedInUserDetails, error) {
 			SetAuthToken(userCreds.JTWToken).
 			SetHeader("Accept", "application/json")
 
+		config.INFISICAL_URL_MANUAL_OVERRIDE = config.INFISICAL_URL
```
maidul98, line 91: Can you explain why you choose to use `INFISICAL_URL_MANUAL_OVERRIDE` variable instead of   `config.INFISICAL_URL`? I see in `GetCurrentLoggedInUserDetails`, you set `INFISICAL_URL_MANUAL_OVERRIDE=config.INFISICAL_URL` so i'm curious if we need this or if there is a use case I didn't catch

cli/packages/cmd/user.go
maidul98, line 115: Can you explain why you choose to use `INFISICAL_URL_MANUAL_OVERRIDE` variable instead of   `config.INFISICAL_URL`? I see in `GetCurrentLoggedInUserDetails`, you set `INFISICAL_URL_MANUAL_OVERRIDE=config.INFISICAL_URL` so i'm curious if we need this or if there is a use case I didn't catch

quinton11: Right, so with the two manual override methods, `config.INFISICAL_URL` gets updated with the user self-hosted-instance url from either `--domain` or `INFISCAL_API_URL` env var  before it hits the run function of any infisical command, now if you notice in the `PreRun` or `Run` of most commands there is a check for a user's logged in status in `GetCurrentLoggedInUserDetails`. For multi-profiles the config file holds the current logged in user as well as the user's domain. Now in here: https://github.com/Infisical/infisical/pull/481/commits/d4206cdbd82707d5ed3e655bf5e1012c1241d619#diff-eb3a7252b09a5793a44fec4d5e033dda44f2fc86484c3d1a728bb57208058fdaR93 , there is a check for if the config data structure holds a `LoggedInUserDomain` value. If there is it overrides the `config.INFISICAL_URL`. To still support the two manual override methods we don't want that. So the `config.INFISICAL_URL_MANUAL_OVERRIDE` gets assigned the `config.INFISICAL_URL` value before its updated. This is necessary because we can use that to check if there was a domain override from those two methods in the `Run` functions of the relevant commands, in this case `login` and `user update domain`.

quinton11: Then we check if there was an override by comparing `config.INFISICAL_URL_MANUAL_OVERRIDE` with the default api

maidul98: Oh i see, makes sense. So INFISICAL_URL_MANUAL_OVERRIDE holds the value defined by the two methods and is used to detect if those two methods are used. Thanks for explaining

maidul98 (approved):

Issue #198, opened 2023-01-06
Add slack integration into CLI release workflow so users can subscribe to be notified of updates


maidul98:
Add slack integration into CLI release workflow so users can subscribe to be notified of updates. We currently use go releaser to build our CLI artifacts and use github action to automate the whole process. By having slack integration, users can stay update date with new features and bug fixes

quinton11:
Is there anyone working on this?

maidul98:
Not yet, would you like to take a stab?

quinton11:
Yeah I would, thanks

akhilmhdh:
Closing as not required! We have a release page for CLI changes

PR #498, opened 2023-04-13
Feat/slack integration

quinton11:
# Description 📣

Slack integration into CLI release workflow so users can subscribe to be notified of updates

Issue: See #198 

Motivation: This is to allow slack users subscribe to be alerted on latest releases relating to the infisical cli.

Dependencies:
- Requires a SLACK_WEBHOOK and SLACK__ICON github secret.


maidul98:
With this approach, users will only know of new CLI updates if they are in our slack. I think it'd be more useful if this was a slack bot that people can install in their workspaces. Let me know what you think

quinton11:
Updated PR. To use you would have to provide a SLACK_BOT_TOKEN env variable after creating your slack app. As well as the message you intend to broadcast. Currently its just a basic text.

How it works?
It uses a custom github action (see https://github.com/Infisical/infisical/pull/498/files#diff-33ba490cb8145b58db19f4dd0f72ccfddae4b86d5fde272a1d3c6c9ce964514a) to get all channels, both private and public associated with your bot token (When users install your bot and invite them into channels) and broadcasts the message you provide to each channel.

maidul98:
Closing for now as it is not a top priority

Salmaan-M:
Can we open this issue again ?, i would love to work on it.

Issue #214, opened 2023-01-13
Add IP-whitelisting to projects


mv-turtle:
### Feature description
Project members should only be allowed to perform any operations (especially the ones related to secrets) if they have an allowed IP-address. 

### Why would it be useful?
For security purposes - would be very useful for many users.

imakecodes:
I'm using behind Nginx and I'm using it to secure my instance but it not easy to handle sometimes

mv-turtle:
Hey @uncle-tee! Thank you so much for your help! That would be amazing! Have you already joined our Slack? Here is the link: https://infisical.com/slack. Let us know if me have any questions regarding the codebase or while setting up

You should also talk to @dangtony98. because I know that he was working on this feature at some point

akhilmhdh:
Closing as there is already one in discussion thread for this.

https://github.com/Infisical/infisical/discussions/5712

PR #509, opened 2023-04-18
Feat(Backend): Add IP-whitelisting to projects

uncle-tee:
# Description 📣
This PR allows workspace members to add IP addresses, it also creates a middleware to validate if the IP has been whitelisted. Based on this issue [here](https://github.com/Infisical/infisical/issues/214).

- [ ✅] Endpoint to create an IP for a workspace
- [ ✅] Endpoint to get all IPs in a workspace
- [✅ ] Endpoints to point to delete IPs  from a workspace
- [ ✅] Middleware to validate if the incoming request has a valid IP in its workspace.

## Type ✨

- [ ] Bug fix
- [ 👍🏿] New feature
- [ ] Breaking change
- [ ] Documentation


backend/src/routes/v2/ipAddress.ts
dangtony98, line 1: Can you please move these endpoints into `routes/v2/workspaces`?

We'll work on better folder modularization in the future but this would be inline with existing endpoints.

backend/src/models/ipAddress.ts
```diff
@@ -0,0 +1,36 @@
+import { model, Schema, Types } from "mongoose";
+
+export interface IIpAddress {
```
dangtony98, line 3: Nit-picky detail but let's do `IIPAddress` here.

docker-compose.dev.yml
dangtony98, line 1: Is there a reason for exposing this port?

backend/src/controllers/v2/ipAdressController.ts
dangtony98, line 1: Let's put this int `controllers/v2/workspacesController.ts` to be consistent with how we moved the workspace IP routing to `routes/v2/workspace`.

dangtony98:
This has been completed in #785 since this PR took too long 😞

Issue #594, opened 2023-05-22
GUI bug and no scrollbar in project tab.


Shabbirali-ali:
![image](https://github.com/Infisical/infisical/assets/73350953/b6ee8e73-354d-45bd-a0c4-951c7d0a78a1)
Above image shows GUI bug when delete tag.
![image](https://github.com/Infisical/infisical/assets/73350953/caffbce9-bbf0-490b-8568-f7208c6d3a35)
no Scrollbar in project list.

vmatsiiako:
This should be fixed now thanks to @piyushchhabra!

PR #602, opened 2023-05-25
fix(ui): fixed scroll on project list selection

piyushchhabra:
# Description 📣

**Fixes**: https://github.com/Infisical/infisical/issues/594

The project list dropdown in the main dashboard is not having any scroll as of now. This makes the UI a bit bloated when a user adds more than a few projects. I am now adding a scroll bar to the list to improve the user experience.

Along with this, I have updated the up and down arrow icons in the `Select` dropdown component. It was using a different icon for the up arrow and a different icon for the down arrow. I have made it consistent now
```diff
- faChevronUp
+ faCaretUp
```

> I have also moved the `Add Project` button to the top of the list instead of the bottom otherwise It will not be visible when there are multiple projects. Ideally, this item should be fixed during the dropdown scroll but that change is gonna take me some time to figure out

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Demo 🛠️

**Before Fix:** 
withoutFix

**After Fix:**
fixedScroll

```sh
```

---

### DIFF
--- a/frontend/src/components/v2/Select/Select.tsx
+++ b/frontend/src/components/v2/Select/Select.tsx
@@ -1,6 +1,6 @@
 import { forwardRef, ReactNode } from 'react';
 import { IconProp } from '@fortawesome/fontawesome-svg-core';
-import { faCaretDown, faCheck, faChevronUp } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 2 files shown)


vmatsiiako (approved):
Thank you so much @piyushchhabra! Merging this right now

PR #630, opened 2023-06-08
fix domain selection login for cli-login

adnull:
# Intuitive logic in selecting proper domain URL 📣
there is no way to set api domain by the --domain command argument or the INFISICAL_API_URL environment variable in the login command for now, since there is wrong condition in the login code. 

So, i propose this logic: 
if config.INFISICAL_URL != util.INFISICAL_DEFAULT_API_URL it means that it was changed by command argument or env variable.

If there is any saved config and we've got domain from it, set config.INFISICAL_URL_MANUAL_OVERRIDE to it, but keep config.INFISICAL_URL being set by command argument or env var.

Then, if we have URL loaded from the config and the real URL wasn't set by command argument - ask for domain override, otherwise ask for new domain only if URL wasn't set by command argument

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Tested in various conditions  - with URL set by argument or by env var, with existing config and without it, looks like this logic is intuitive 

```sh
```

---


maidul98:
Hey @adnull i'll review this soon!

maidul98:
@adnull I have checked and `--domain` seems to work as expected. Closing for now.

PR #631, opened 2023-06-08
Fix domain prompt in cli

adnull:
# Remove "Example - " from  domain input 📣

If user selects a self-hosted instance, cli asks for a domain url, which contains "Example - ". So the user just starts changing example url instead of typing it from scratch. And it leads to "wrong url" message, which is non-obvoius to understand why does it happen
изображение

So it is easy to fix by just removing the "Example - " substring from the input. 

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/cli/packages/cmd/login.go
+++ b/cli/packages/cmd/login.go
@@ -331,6 +331,7 @@ func askForDomain() error {
 	}
 
 	urlValidation := func(input string) error {
+		input = strings.Replace(input, "Example - ", "", 1)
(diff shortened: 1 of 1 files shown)


vmatsiiako:
Hi @adnull! Thank you so much for contributing! @maidul98 will take a look at this soon

maidul98:
No reports of this confusion has been noted. Closing for now

PR #633, opened 2023-06-09
Folder scoped service token

akhilmhdh:
# Description 📣

1. Folder scoped service token support in API
2. Service token ui added secretpath option

## Type ✨

- [x] New feature

---

### DIFF
--- a/backend/src/controllers/v2/secretsController.ts
+++ b/backend/src/controllers/v2/secretsController.ts
@@ -1,6 +1,6 @@
 import { Types } from "mongoose";
 import { Request, Response } from "express";
-import { ISecret, Secret } from "../../models";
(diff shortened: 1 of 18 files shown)


dangtony98:
Looks good to me!

maidul98 (approved):
It looks like we haven't yet added the restrictions to v2 secrets. Let's also add it to that next. Otherwise, the PR looks good. 

I added some changes like verifying the path before service token creation and updated error types. Maybe we should add frontend notification for that as well next, but not urgent issue. 

We should also display folder names as is because right now they are shown as capital but under the hood they are lower case so it becomes confusing when trying to form the path. Google drive also does this.

We can  merge this one and open a new PR for the additional changes. Up to you.

Issue #336, opened 2023-02-17
Login redirect from CLI to browser


vmatsiiako:
### Feature description
When I need to log in in the CLI, it would be really awesome to be redirected to the browser. Then I can really easily login with 1password Chrome extension or the Apple Keychain, and I don't need to paste my password to the CLI. 

### Why would it be useful?
This would save a lot of time, reduce friction, and improve general experience of using Infisical

animeshdas2000:
Hey @mv-turtle, this seems like a good issue to work on. Although I haven't worked on golang extensively as I could gather from the codebase, I believe we need to write a condition where we give the user to log in via the browser and then it redirects to the browser. And we do this in packages/cmd/login.go?

If this sounds good, I would like to work on it, can you assign it to me?

maidul98:
@animeshdas2000 yes, this should be done in `packages/cmd/login.go`. What you want to do is pop open the browser from the CLI, and have the CLI listen to a call back (usually on a port). When the user completes a login in the browser, send over the private key, public key, JWT token, and any other info you see needed in login cmd. The rest of the workflow should be to save the details to keyring like we already do. 

By the way, I am making some changes in `packages/cmd/login.go` so let me push out these changes before you have a large merge conflict. Feel free to explore though and let me know what your approach is.

maidul98:
Hey @animeshdas2000 just checking in. Let us know if you need a bit of help!

animeshdas2000:
Hey @maidul98, I am on it. Was stuck due to some personal commitment but I'm back and I should be able to raise a PR soon. I'll reach out in case I'm stuck. Thanks for checking on this

akhilmhdh:
This has been released in our recent release. Thus closing it for now, feel free to create a new issue if its not working.

PR #648, opened 2023-06-14
feat: cli login via browser

quinton11:
# Description 📣

This feature allows cli users login into their profiles without inputting their credentials into the terminal.
See #336


quinton11:
@vmatsiiako @maidul98  After user logs in, the browser is to route to a `cli-redirect` page. Currently the page is just the 404 page with a different text indicating a successful login.

frontend/src/components/login/InitialLoginStep.tsx
```diff
@@ -31,33 +33,60 @@ export default function InitialLoginStep({
...
-    
-            // case: login does not require MFA step
-            router.push(`/dashboard/${localStorage.getItem('projectData.id')}`);
-          }
-    
+
+            setIsLoading(true);
+            const queryParams = new URLSearchParams(location.search)
+            if (queryParams) {
+                const callbackPort = queryParams.get("callback_port")
```
maidul98, line 43: I think this line should go above the if and we should only go into it if 

```
const callbackPort = queryParams.get("callback_port")
```
has a value

frontend/src/components/login/InitialLoginStep.tsx
```diff
@@ -31,33 +33,60 @@ export default function InitialLoginStep({
...
+                    password,
+                })
+
+                if (isCliLoginSuccessful && isCliLoginSuccessful.success) {
+                    // case: login was successful
+                    const cliUrl = `http://localhost:${callbackPort}`
+
+                    //send request to server endpoint
+                    const instance = axios.create()
+                    const cliResp = await instance.post(cliUrl,{...isCliLoginSuccessful.loginResponse,email,password})
```
maidul98, line 57: hmm, i wonder if this is safe to do to. Let me get back to this

maidul98: also since we are sending back login  2 response, that means that users will still have to retrun to CLI to enter their MFA code. I expected even MFA workflow to happen on the browser

maidul98: I wonder if we can somehow add a secure transport layer so that the data isn't just sent back over http to the local server

frontend/src/components/login/InitialLoginStep.tsx
quinton11: Wouldn't `queryParams` if undefined throw an error if we call `.get()` on it?

frontend/src/components/login/InitialLoginStep.tsx
quinton11: Hm, I didn't consider the MFA workflow.
I don't believe that would be an issue(browser handling it) however what would be returned to the CLI?

frontend/src/components/login/InitialLoginStep.tsx
quinton11: Maybe an extra check after the line for the callback_port value instead?

frontend/src/components/login/InitialLoginStep.tsx
maidul98: We return the same details back to the CLI  that would be retruned by the login2 call. You can see this being done here: https://github.com/Infisical/infisical/blob/main/cli/packages/cmd/login.go#L136. So what you can do is when it is MFA, have the user complete the MFA in the browser and once done, you can send back the details from the MFA verify call to the CLI

frontend/src/components/login/InitialLoginStep.tsx
maidul98: I think this should actually be fine

maidul98:
- I know the current time out is set to 1 minute, but i think it should be running for more like at least 10 minutes and give the option for user to quit it. Here is how heroku does it. 
Screenshot 2023-06-15 at 6 07 19 PM


- I also noticed that there is always this strange pop up and I wonder if there is any way to get rid of it because Heroku and other CLI based login don't seem to have it 
Screenshot 2023-06-15 at 6 03 28 PM

-  I think having the login via browser as a flag makes this feature really hidden. I think we should default to browser login and if it fails for any reason, then fall back to CLI based login. What do you think?
- Also just remember, all the CLI needs is 
```
Email:      email,
PrivateKey: string(decryptedPrivateKey),
JTWToken:   loginTwoResponse.Token,
```

so I think we should just pass these back to the CLI and skip the other CLI login logic if we get these details back successfully from the call back. This means that if the user is already logged in the browser, then these details would be sent back to the CLI without requiring any logins again.

Let me know you have any questions

frontend/src/components/login/InitialLoginStep.tsx
maidul98: Added details about what should be sent here https://github.com/Infisical/infisical/pull/648#issuecomment-1593815018

maidul98: In terms of security, we should be okay since we are communicating with the client itself

maidul98:
Great work! I'll take a look at the issue with the welcome text being pushed to the right. One case we are missing is when the user is already signed into Infisical web app. If the user is signed in, then we should fetch the `model.UserCredentials` details from local storage and send it back

Screenshot 2023-06-16 at 1 39 27 PM

quinton11:
Regarding the case where the user is already signed into the infisical web app, looks like the email isn't stored in localStorage. With that what we can do without tampering with the localStorage is send the `privateKey` with the `token` back to the cli and update the `userCredentials.Email`  with the `config.LoggedInUserEmail` value  on the CLI side. However there are issues with that regarding handling multiple profiles.

In the case where a user logs in a different profile on the browser and doesn't logout and attempts to login via the cli which has a different current logged in profile, the wrong privateKey and token are sent back.

Another way is to make a request to the server for user details before responding to the CLI.
Or just add the email to the localStorage in the login process. What do you think?

maidul98:
The user details including the email can be grabbed by using this context https://github.com/Infisical/infisical/blob/main/frontend/src/context/UserContext/UserContext.tsx#L31

quinton11:
Yeah so the CLI pops the browser open with the `hostUrl/login` endpoint and the check for previous logins happens here https://github.com/Infisical/infisical/blob/main/frontend/src/pages/login.tsx#L42 right? The login page doesn't fall under the UserContext.
Screen Shot 2023-06-17 at 16 30 32

maidul98:
I have asked @akhilmhdh to help you with this since he has more context with frontend than I do

akhilmhdh:
@quinton11 can you wrap [this part](https://github.com/Infisical/infisical/blob/main/frontend/src/pages/_app.tsx#L68) with user context provider and check it out. 

I don't think it will break anymore. Kindly try both logged out user logging in too after this change

quinton11:
@maidul98 Thanks! Btw I believe I've fixed the unaligned text issue, I'll test it some more.
@akhilmhdh It doesn't necessarily break now the user is however undefined even after logging in. That is from `useUser` after wrapping that line with the `UserProvider`. Or do I need to pull recent changes?

akhilmhdh:
Ah yah. I think its because the user context is not re validated when the user is logged in, its on previous state, 

In other components it works without re validating as on login a remount happens.

You can do two options.
1. Put a dependency between auth provider and user provider so that when auth change user also changes. May be u can join these two providers also together.

2. Just do the server call of the user provider that is the getUser function inside here and finish the process.

Step 2 would be the easiest as the other one may increase the scope of this PR a bit big. Ill just later do the step 1.

quinton11:
Yeah with step 2 do you mean the `useGetUser` hook? it makes a useQuery call for the user details right? I tried that before but it appears the call doesn't complete in time..i.e at the check for user login the data is undefined

akhilmhdh:
yah u can import that axios function here and call it here. If that is also undefined. Then i dont think your logged in.

akhilmhdh:
yah u can import that axios function here and call it here. If that is also undefined. Then i dont think your logged in.

quinton11:
Right, so if user is still logged in the details are fetched and sent back to the CLI. Also includes a fix for the terminal text alignment bug. Thanks!

maidul98:
Login from CLI works in both logged in / unlogged in state, nice! I noticed that when you quit, and choose self hosted, the backend is not pointing to the self hosted URL anymore. I expected the login to point to the self hosted URL that i choose after pressing quit. 

Also, let's add in a flag `-i, --interactive` that will trigger username/password login

cli/packages/cmd/login.go
```diff
@@ -326,7 +347,8 @@ func askForDomain() error {
 
 	if selectedHostingOption == INFISICAL_CLOUD {
 		//cloud option
-		config.INFISICAL_URL = util.INFISICAL_DEFAULT_API_URL
+		config.INFISICAL_URL = fmt.Sprintf("%s/api", util.INFISICAL_DEFAULT_URL)
```
maidul98, line 350: Is there a reason we need to define a new `INFISICAL_DEFAULT_URL` varable? I think we can just re use `INFISICAL_DEFAULT_API_URL`?

maidul98: Oh i see, you need it for the login url

quinton11: Yep

quinton11:
> Login from CLI works in both logged in / unlogged in state, nice! I noticed that when you quit, and choose self hosted, the backend is not pointing to the self hosted URL anymore. I expected the login to point to the self hosted URL that i choose after pressing quit.
> 
> Also, let's add in a flag `-i, --interactive` that will trigger username/password login

You mean when you quit the browser login?

quinton11:
> Login from CLI works in both logged in / unlogged in state, nice! I noticed that when you quit, and choose self hosted, the backend is not pointing to the self hosted URL anymore. I expected the login to point to the self hosted URL that i choose after pressing quit.
> 
> Also, let's add in a flag `-i, --interactive` that will trigger username/password login

Hm, I tested it again but I don't see this behaviour..what url does it point to in your set up? That is after hitting 'q'

maidul98:
Tested and all looks good to me from the CLI side. Please address the merge conflicts and then @vmatsiiako will review the frontend side of things. Once those are done, i'm happy to merge it in. Good work @quinton11 !

maidul98 (approved):

Issue #58, opened 2022-11-30
Change 6-digit code to an email link during signup


mv-turtle:
### Feature description
During signup, users get an email with a 6-digit code that they later need to copy and paste into the next signup step. Ideally, we would not to substitute it to an email link that people click and it automatically opens up the next signup step (after verifying that the link is correct).

### Why would it be useful?
Links are both more secure and effortless than 6-digit codes. This would be very good to have!

khoa165:
@vmatsiiako hi Vlad, I am new to OSS contributing, hoping to pick up my first issue. Do you know if this has been done yet?

vmatsiiako:
hey @khoa! Excited for you to join our community and hopefully contribute to Infisical. You should also consider joining our Slack for contributors. Here is the link: https://infisical.com/slack

I think this would be a great issue. I'd love to hear what @dangtony98 thinks about it too.

khoa165:
@vmatsiiako do you know how I can setup the email/SMTP? I used to just use my own personal gmail/password and put it in .env file. However, Google recently disabled/deprecated the ability to allow for less secure app access, so I am not sure if there is alternative or if there is shared email account from Infisical for this purpose.

<img width="1556" alt="Screenshot 2023-06-19 at 2 16 57 PM" src="https://github.com/Infisical/infisical/assets/46258781/6140b9c7-6619-48c4-b88d-4db1ff35b136">

khoa165:
@vmatsiiako from what I just dig around, sendgrid seems to offer SMTP server instead of giving you a free email.
Specifically, I think I am looking for maybe alternative of what I should populate these fields in .env with:
```
SMTP_HOST=Gmail
SMTP_PORT=
SMTP_NAME=
SMTP_USERNAME=dev.khoa165@gmail.com
SMTP_PASSWORD=[my email password]
```
I made sure to put the correct email/password yet it did not send me the email for 6-digit code (step 2 of signup)

maidul98:
If you want to configure SMTP for local development, you can simply use the mailhog details below 

```
SMTP_HOST=smtp-server
SMTP_PORT=1025
SMTP_NAME=<whatever you like>
SMTP_USERNAME=team@infisical.com
SMTP_PASSWORD=
```

All mail will then be sent here http://localhost:8025/

dangtony98:
@khoa165 You should the documentation we have for configuring Infisical with various email providers here: https://infisical.com/docs/self-hosting/configuration/email

[...]

PR #683, opened 2023-06-24
FEAT: replace digit code with email signup link

khoa165:
# Description 📣
FEAT: replace digit code with email signup link
[Issue 58](https://github.com/Infisical/infisical/issues/58)

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
https://www.loom.com/share/3ed03c737c5f4757bd29565867a54670?sid=0e46e618-55b3-4519-9bd0-234a5060a4aa
---


dangtony98:
Looks good to me from the backend; the functionality worked well!

I glossed over the frontend but @vmatsiiako can you give it a look? — Prolly more familiar with the signup sequence.

khoa165:
Hope the Loom video is useful to peek at frontend. Lemme know if you want me to make any changes @vmatsiiako

maidul98:
Can you please verify that when SMTP related environments are not added to the backend, the step to confirm the users email is skipped?

vmatsiiako:
This looks great to me! The only small thing is: in the email, should we give people both the button and the link? This way people can copy it easier

dangtony98:
I think the button should be good (people can right-click it to get the link); I don't think having the link explicitly there is common-practice.

I'll do one last run through it and check for the edge-case @maidul98 highlighted today, then merge :)

khoa165:
@dangtony98 Hi Tony, indeed I didn't skip step 2 when smtp is not configured. My latest commit should have that fix. Before, step 2 always made the request to backend regardless if there is smtp or not, and even without the code, the server still returns 200 (which does not make sense) hence step 2 is skipped to step 3 regardless. Anyway, I fixed that, hope all should be good to merge. My latest commit changed like 1 or 2 lines so should be easy to review

dangtony98:
@maidul98 Can you take a look at this?

I've tried the flow without SMTP configured and it didn't work; it skipped the user email confirmation step but when completing the user's account, it errored out on this in `v3/signupController.ts`:

```js
if (!user || (user && user?.publicKey)) {
	return res.status(403).send({
		error: "Failed to complete account for complete user",
	});
}
```

Is the flow typically supposed to work when SMTP is not configured?

maidul98:
The issue is you never called `checkEmailVerificationCode`.  `checkEmailVerificationCode` needs to be called with a code even when smtp is disabled. This is because when this api call is made, it will create the user with a email which is needed for next step. 

Because SMTP is disabled, the server will not do a code verification match as shown here https://github.com/Infisical/infisical/blob/6d807c0c742e724362fd06e49cee502b683045cd/backend/src/controllers/v1/signupController.ts#L75

maidul98:
Closing due to inactivity

vikramnagwal:
@maidul98 , is this issue still open? I think adding a verify account button will resolve this issue, as well as #1582.

PR #684, opened 2023-06-24
integrations page revamp

akhilmhdh:
# Description 📣

1. New integration page
2. Updated storybook to latest version

## Type ✨

- [x] Bug fix
- [x] Optimization

---

- [ x I have read the [contributing guide](https://infisical.com/docs/contributing/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/code-of-conduct). 📝


akhilmhdh:
@dangtony98 kindly test some more integrations from your setup before merging.

maidul98:
Seems much faster! 

For the cloud integration section, i think we should put the list of integrations in the same container so that  it doesn't stick out far right. Or we can make the list of active integrations longer? 

Screenshot 2023-06-25 at 12 20 57 PM

maidul98:
I have tested with Netlify, Github, Gitlab, and Vercel and eveything seems to work nice. Let's wait for @vmatsiiako to see if he has any suggestions for the UI

vmatsiiako:
I was just going through this PR and there is another things that i thought of – there is not indication that a particular integration is `working`. We previoulsy had the 'In Sync' label. Maybe we should have some kind of indication like that. Ideally, we'd also have a manual sync button

vmatsiiako:
Another thought is that it could be a good idea to split infra tools (Docker, Kubernetes, etc) apart from the frameworks (React, Vue, etc) 

Screenshot 2023-06-25 at 21 41 22

vmatsiiako:
I pushed a super minor update, but otherwise I like everything!

dangtony98:
I agree regarding @maidul98 comment on keeping things flushed left; ideally, we pick either we go with center or left but having it be sporadic across different pages makes for confusing styling consistency.

I prefer flushing left because it seems more consistent for the breadcrumbs.

akhilmhdh:
@dangtony98 its kinda of bad ux to keep it to left. 

Because this issue is when you have big screen and if its on left kinda stressful to the users to look towards the left all time. 

Also symmetry is another.

I didn't extend it because i thought @dangtony98 was reducing the of containers.

dangtony98:
I think we should replace the id of the integration here with the name of the integration or something like that @akhilmhdh 

Screen Shot 2023-06-26 at 12 26 37 PM

akhilmhdh:
Huhh. My bad. I thought i put name here. Yah it should be name. My bad. Good catch @dangtony98

dangtony98:
> I was just going through this PR and there is another things that i thought of – there is not indication that a particular integration is `working`. We previoulsy had the 'In Sync' label. Maybe we should have some kind of indication like that. Ideally, we'd also have a manual sync button

I feel like if the integration is set up under "current integrations" then it can be assumed that it's working.

dangtony98:
> Seems much faster!
> 
> For the cloud integration section, i think we should put the list of integrations in the same container so that it doesn't stick out far right. Or we can make the list of active integrations longer?
> 
> Screenshot 2023-06-25 at 12 20 57 PM
> Also i think we should shift thigns to the left. Putting it on the center looks odd with the breadcrub

Agree. I think we can keep integrations to 4 cols and have it be the same width as the current integrations up-top; it would feel more consistent.

@akhilmhdh can you make this change?

akhilmhdh:
Ofcourse.

akhilmhdh:
Resolved. It should be now uniform width. @vmatsiiako did you made it back to leftie 😅. Check it out now. Its uniform size and centered well.

PR #713, opened 2023-07-05
secret reference

akhilmhdh:
# Description 📣

1. Implemented secret reference with multiple environment and multiple folder reference
2. Updated service token to support multi env and glob matching in folders
3. Updated documentation
4. Added new backfill for service token changes

## Type ✨

- [x] New feature

# Tests 🛠️

```sh
```

---


backend/src/helpers/secrets.ts
```diff
@@ -323,38 +297,39 @@ export const createSecretHelper = async ({
...
   });
 
   // if using service token filter towards the folderId by secretpath
   if (authData.authPayload instanceof ServiceTokenData) {
-    const { secretPath: serviceTkScopedSecretPath } = authData.authPayload;
-    if (secretPath !== serviceTkScopedSecretPath) {
+    const { scopes: tkScopes } = authData.authPayload;
+    const validScope = tkScopes.find(
+      (scope) =>
+        picomatch.isMatch(secretPath, scope.secretPath, { strictSlashes: false }) &&
```
maidul98, line 312: i think this should `secretPath || "/"` ? Because for the other ones i see you have made it so that if no path is defined then the secret is created on root

maidul98: Oh i see, secretPath is already defaut to '/'

backend/src/helpers/secrets.ts
```diff
@@ -833,26 +817,27 @@ export const deleteSecretHelper = async ({
...
 
   // if using service token filter towards the folderId by secretpath
   if (authData.authPayload instanceof ServiceTokenData) {
-    const { secretPath: serviceTkScopedSecretPath } = authData.authPayload;
+    const { scopes: tkScopes } = authData.authPayload;
+    const validScope = tkScopes.find(
+      (scope) =>
+        picomatch.isMatch(secretPath, scope.secretPath, { strictSlashes: false }) &&
+        scope.environment === environment
+    );
```
maidul98, line 834: I see that we do a lot of repeating of the same few lines. Is it possible to make it into a function or is there a reason you didn't put it into a function?

```
const findValidScope = (authPayload, secretPath, environment) => {
  const { scopes: tkScopes } = authPayload;
  const validScope = tkScopes.find(
    (scope) =>
      picomatch.isMatch(secretPath, scope.secretPath, { strictSlashes: false }) &&
      scope.environment === environment
  );
  return validScope;
};
```

maidul98:
- Left comments on slack

backend/src/helpers/secrets.ts
akhilmhdh: Done

maidul98 (approved):

Issue #679, opened 2023-06-22
Windmill Integration


dangtony98:
### Feature description
It would be great to have an integration with [Windmill](https://github.com/windmill-labs/windmill).

### Why would it be useful?
This would enable users of Windmill to store and sync their environment variables from Infisical to Windmill.

### Additional context
This should be fairly simple with a ton of examples to go off of 😄

nevvada:
hey @dangtony98, I'd be interested in working on this!

PR #765, opened 2023-07-20
Feature: Windmill.dev cloud Integeration

sunilk4u:
# Description 📣

This PR (feature #679) adds Windmill.dev Cloud integration to infisical. Sync secrets from infisical to windmill.dev cloud using windmill's api key token. Authorize using windmill api key in infisical, create integration according to environment. Add secrets to indfisical with comments (which will reflect in windmill's variable description). 

This feature supports two paths of secrets for windmill (user / folder). Just use the proper format of windmill variables (additional checks are put in backend before making request).

```text
User path secrets format: u/user_name/secret_name
Folder path secrets format: f/folder_name/secret_name
```

This feature will not sync secrets with invalid windmill variable format. Workspaces/apps of windmill where user(api key) is admin are allowed to integrate and sync.

ezgif com-video-to-gif

## Important points to note
- Secrets with correct windmill variable format will sync. Rest will be ignored.
- Description for windmill variables can be provided in infisical secrets comments. It is not mandatory.
- During integration only those workspaces/apps will show where user (api token) has write access. Backend will create a simple secret in windmill and check write access. It will delete the secret after write access is verified.

## Type ✨

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

- Scenario 1: Create secret in infisical and with valid and invalid format to see if correct secrets are syncing.
- Scenario 2: Create and sync scerets with and without comments.
- Scenario 3: Update secrets key and comments to check if there is any residue in windmill variable dashboard.
- Scenario 4: Check for workspaces/apps not showing for user without write access during integration process.

Note: At the time of pull request, code has been merged with latest infisical/main.
---


rubenfiszel:
Hello, founder of windmill here.

> Only admin privilege user can create variables in windmill.

That's not true, windmill has complex permissions and users have different read and write permissions to different folders based on the permissions set in windmill. It is however true that only admins will never have permissioning issue.

backend/src/integrations/sync.ts
```diff
@@ -2007,4 +2020,123 @@ const syncSecretsCodefresh = async ({
...
+  // convert secret results to [key] format
+  const secretsResList = getSecretsRes.map((secretObj: any) => (secretObj.path));
+    
+  // convert the secrets to [{}] format
+  const modifiedFormatForCreateSecretInjection: WindmillSecret[] = [];
+  const modifiedFormatForUpdateSecretInjection: WindmillSecret[] = [];
+  
+  Object.keys(secrets).forEach(
+    (key) => {
+        const pattern = new RegExp('^([a-zA-Z0-9])\/([a-zA-Z-0-9])+\/([a-zA-Z0-9])+')
```
rubenfiszel, line 2070: underscores and dash are also allowed

sunilk4u:
> Hello, founder of windmill here.
> 
> > Only admin privilege user can create variables in windmill.
> 
> That's not true, windmill has complex permissions and users have different read and write permissions to different folders based on the permissions set in windmill. It is however true that only admins will never have permissioning issue.

Thanks for letting me know, I will look into this. Is there a way through API I can check if user(api key) has priviledges to create variables? I couldn't find a proper way for that.

backend/src/integrations/sync.ts
sunilk4u: will incorporate this as well.

rubenfiszel:
>      Thanks for letting me know, I will look into this. Is there a way through API I can check if user(api key) has priviledges to create variables? I couldn't find a proper way for that.

Anyone can create variables, but not everywhere. A user always has write-access on his user-space. For folders, it's a bit more complex and require checking the permissions on the folder which you can retrieve from the PAI.

I would suggest simply handling errors if user has no write access to a variable (which will return a 401 Unauthorized)  by attempting to write/update rather than checking in advance.

backend/src/integrations/apps.ts
```diff
@@ -760,4 +765,107 @@ const getAppsCodefresh = async ({
...
+
+
+/**
+ * Return list of projects for Windmill integration
+ * @param {Object} obj
+ * @param {String} obj.accessToken - access token for Windmill API
+ * @returns {Object[]} apps - names of Windmill workspaces
+ * @returns {String} apps.name - name of Windmill workspace
+ */
+const getAppsWindmill = async ({ accessToken }: { accessToken: string }) => {
```
dangtony98, line 777: Hi @rubenfiszel!

This function is meant to fetch back a list of apps in Windmill that the user can select from to sync their secrets to.

Wondering is this the best way to check that the user under the `accessToken` has write access to user/folder-path based variables? The current implementation attempts to write a sample variable for both cases and keeps apps where the user has sufficient permissions but this seems rather lengthy.

Perhaps there is a better way to check for sufficient permissions rather than having to make a sample secret for each app?

rubenfiszel: The permissions are too dynamic in windmill for this to work well. I would just list the workspaces like you do and assume one has write access on all the secrets he can see.

Then if the write-access is not working, then there should be a clear error during sync.

In most cases, the sync will be done by admins so this won't be a common case.

rubenfiszel: I think it would even be fair to require that the token is one of an admin and thus mention that behavior is somewhat undefined if that's not the case. It's not as precise as it could be but there won't be a lot of cases where a non-admin would sync workspace-level secrets.

PR #778, opened 2023-07-23
docs: add running infisical docs locally guide

afrieirham:
# Description 📣

image

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---


maidul98:
Great idea but @afrieirham I think it might be better to add these docs here instead: https://infisical.com/docs/contributing/developing

dangtony98:
Agree. I think we should put it into the contributing guide/docs 😄

afrieirham:
@maidul98 cool, will do that instead. 🙌

PR #836, opened 2023-08-09
feat: Add Parse copy-pasted secrets in the dashboard

### LINKED ISSUE #231: Parse copy-pasted secrets into the dashboard
### Feature description
If a user pastes `SECRET_X=VALUE_Y` into the dashabord, we should immediately split the key and the values into two input fields. 

### Why would it be useful?
This a QoL feature and it would be dramatically improve the user experience for many users who paste the environment variables / secrets into the dashaboard.

alceil:
# Description 📣

Feature to Parse copy-pasted secrets in the dashboard
Resolves #231 

[screen-capture.webm](https://github.com/Infisical/infisical/assets/47685349/df257771-dbc6-45d8-af62-cceb47b2aa66)

## Type ✨

- [x] New feature

# Tests 🛠️

```sh
```

### DIFF
--- a/frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
+++ b/frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
@@ -199,6 +199,15 @@ export const SecretInputRow = memo(
 
     const isCreatedSecret = !secId;
     const shouldBeBlockedInAddOnly = !isCreatedSecret && isAddOnly;
+    
+    const hasEqualSign = (str:string) => str.includes("=");
+  
+    const splitKeyValue = (str:string) => {
+      const [key, ...rest] = str.split("=");
+      const value = rest.join("=");
+      setEditorRef(value);
+      return key;
+    };
 
     // Why this instead of filter in parent
     // Because rhf field.map has default values so basically
@@ -227,7 +236,7 @@ export const SecretInputRow = memo(
           control={control}
           defaultValue=""
           name={`secrets.${index}.key`}
-          render={({ field }) => (
+          render={({ field:{onChange,...field}}) => (
             <HoverCard openDelay={0} open={isKeyError ? undefined : false}>
               <HoverCardTrigger asChild>
                 <td className={cx(isKeyError ? "rounded ring ring-red/50" : null)}>
@@ -244,6 +253,9 @@ export const SecretInputRow = memo(
                       onBlur={() => {
                         isKeySubDisabled.current = false;
                         field.onBlur();
+                      }}
+                      onChange={(val)=>{
+                        onChange(hasEqualSign(val.target.value)?splitKeyValue(val.target.value):val);
                       }}
                       autoCapitalization={autoCapitalization}
                     />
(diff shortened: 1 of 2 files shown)


frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
```diff
@@ -199,6 +199,15 @@ export const SecretInputRow = memo(
 
     const isCreatedSecret = !secId;
     const shouldBeBlockedInAddOnly = !isCreatedSecret && isAddOnly;
+    
+    const hasEqualSign = (str:string) => str.includes("=");
```
akhilmhdh, line 203: can you keep this function outside the component if its just a utility function

frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
```diff
@@ -245,6 +254,9 @@ export const SecretInputRow = memo(
                         isKeySubDisabled.current = false;
                         field.onBlur();
                       }}
+                      onChange={(val)=>{
+                        onChange(hasEqualSign(val.target.value)?splitKeyValue(val.target.value):val);
```
akhilmhdh, line 258: The issue with this logic is whenever secret value has = it gets split into two parts which can confuse user. 

This should happen only when user is pasting into the input field.

akhilmhdh: Also are you sure value is getting saved, just asking to be sure. Coz the editorRef change may be affected only in ui not in state

frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
alceil: sure @akhilmhdh

alceil:
I have added some relevant changes can you please have a look👀 @akhilmhdh

frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
```diff
@@ -218,6 +225,15 @@ export const SecretInputRow = memo(
       return <></>;
     }
 
+    let isPasted = false; 
+    
+    const handlePaste = () => {
```
akhilmhdh, line 230: Using `onPaste` event for this is indeed a good move. 

But using this way makes it so complicated. I would request you to revert back the onchange things and instead here you will get the event and then using that event u will get clipboard content. Using `setValue` function from react hook form and setEditor ref set the things accordingly

frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
```diff
@@ -218,6 +225,15 @@ export const SecretInputRow = memo(
       return <></>;
     }
 
+    let isPasted = false;
```
akhilmhdh, line 228: Never put state change like this its a bad practice

akhilmhdh (changes requested):

maidul98:
Changes have not been be addressed. Closing for now

PR #846, opened 2023-08-11
fix: backend image failed to start due to npm cache permission

narindraditantyo:
# Description 📣

Backend image is always failed to start while running on non-root environment (e.g. OpenShift). Previous attempt to add recursive chown to `/app` folder causing way slower build times and it turns out I didn't have to change the ownership and just changed the environment variable for the npm cache folder to the 'node' user home folder.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -14,6 +14,8 @@ FROM node:16-alpine
 
 WORKDIR /app
 
+ENV npm_config_cache /home/node/.npm
+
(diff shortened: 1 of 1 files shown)


maidul98 (approved):
Look good to me, we'll give this a try and let you know if build times stay stable

PR #867, opened 2023-08-15
Improve tag colors and sorting on dashboard page

daninge98:
# Description 📣
Makes 2 changes to the way tags are displayed on the dashboard page.

1. **Displays tags in alphabetical order** (currently it looks like tags are displayed in the order in which they were added, which doesn't seem like a useful ordering)

3. **Makes sure that each tag has a consistent color across all rows** (currently the first tag on each row has one color, the second another, the third another etc., but this is not a useful coloring)

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
Ran locally and compared the main branch to this PRs branch

Before this PR:
Screenshot 2023-08-15 at 6 55 01 PM

After this PR:
Screenshot 2023-08-15 at 6 55 24 PM

I also checked that the colors and the sorting stayed correct when (a) adding and removing existing tags from secrets and (b) creating new tags and adding and removing the new tags

---

### DIFF
--- a/frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
+++ b/frontend/src/views/DashboardPage/components/SecretInputRow/SecretInputRow.tsx
@@ -110,6 +110,11 @@ export const SecretInputRow = memo(
       append
     } = useFieldArray({ control, name: `secrets.${index}.tags` });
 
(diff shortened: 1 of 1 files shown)


vmatsiiako:
@daninge98 this is awesome! Thank you so much for contributing! :)

vmatsiiako (approved):

PR #868, opened 2023-08-16
Adds user customizable environment ordering

### LINKED ISSUE #841: Ability to order environments
### Feature description
User should be able to determine the order in which environments show up in the dashboard

### Why would it be useful?
When having 10+ environments, it may become hard to navigate among those. Currently they are being sorted based on created timestamps, which very often doesn't make sense. 

### Additional context
Need to make sure that it still works for old environments.

daninge98:
# Description 📣
Fixes #841 

Allows users to re-order their environments on the project settings page. These changes are then reflected in the ordering of columns on the secrets dashboard and in other places throughout Infisical.

Users can specify their desired ordering using the new up and down arrows on the project settings page

Screenshot 2023-08-16 at 5 46 38 PM

The up arrow on the top row, and the down arrow on the bottom row are disabled (and you can't click on them)

This is implemented using a new backend route `PATCH /v2/{workspaceId}/environments` which takes two environments (defined by name and slug) and swaps the order of them in the database.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

https://github.com/Infisical/infisical/assets/16603122/d5966031-b47a-449c-a13d-5767cece2ee9

There is technically a bug with the implementation of this feature where if two users are changing the order of their projects' environments at the same time, then they might get strange behavior (e.g. if there are only two environments, and somebody already swapped them and your webpage is out of date, clicking swap will actually just swap them back to the original position and appear to do nothing). However this seems like not a big deal in reality as the ordering is only for aesthetic purposes, and the "swap" way of implementing it was the simplest

---

### DIFF
--- a/backend/src/controllers/v2/environmentController.ts
+++ b/backend/src/controllers/v2/environmentController.ts
@@ -86,6 +86,43 @@ export const createWorkspaceEnvironment = async (
 };
 
 /**
+ * Swaps the ordering of two environments in the database. This is purely for asthetic purposes.
+ * @param req
+ * @param res
+ * @returns
+ */
+export const reorderWorkspaceEnvironments = async (
+  req: Request,
+  res: Response
+) => {
+  const { workspaceId } = req.params;
+  const { environmentSlug, environmentName, otherEnvironmentSlug, otherEnvironmentName } = req.body;
+
+  // atomic update the env to avoid conflict
+  const workspace = await Workspace.findById(workspaceId).exec();
+  if (!workspace) {
+    throw new Error("Failed to create workspace environment");
+  }
+
+  const environmentIndex = workspace.environments.findIndex((env) => env.name === environmentName && env.slug === environmentSlug)
+  const otherEnvironmentIndex = workspace.environments.findIndex((env) => env.name === otherEnvironmentName && env.slug === otherEnvironmentSlug)
+
+  if (environmentIndex === undefined || otherEnvironmentIndex === undefined) {
+    throw new Error("environment or otherEnvironment couldn't be found")
+  }
+
+  // swap the order of the environments
+  [workspace.environments[environmentIndex], workspace.environments[otherEnvironmentIndex]] = [workspace.environments[otherEnvironmentIndex], workspace.environments[environmentIndex]]
+
+  await workspace.save()
+
+  return res.status(200).send({
+    message: "Successfully reordered environments",
+    workspace: workspaceId,
+  });
+};
+
+/**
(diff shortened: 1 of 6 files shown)


maidul98:
cc: @akhilmhdh  @vmatsiiako

frontend/src/views/Settings/ProjectSettingsPage/components/EnvironmentSection/EnvironmentTable.tsx
```diff
@@ -31,6 +35,43 @@ type Props = {
 
 export const EnvironmentTable = ({ handlePopUpOpen }: Props) => {
   const { currentWorkspace, isLoading } = useWorkspace();
+  const { createNotification } = useNotificationContext();
+  const reorderWsEnvironment = useReorderWsEnvironment();
+
+  const reorderEnvs = async (moveUp: boolean, name: string, slug: string) => {
```
akhilmhdh, line 41: can you please change the name to `handleReorderEnv` and first argument to `shouldMoveUp`

backend/src/controllers/v2/environmentController.ts
```diff
@@ -124,7 +161,7 @@ export const renameWorkspaceEnvironment = async (
   if (envIndex === -1) {
     throw new Error("Invalid environment given");
```
akhilmhdh, line 162: throw new BadRequestError

backend/src/controllers/v2/environmentController.ts
```diff
@@ -85,6 +85,43 @@ export const createWorkspaceEnvironment = async (
...
+  req: Request,
+  res: Response
+) => {
+  const { workspaceId } = req.params;
+  const { environmentSlug, environmentName, otherEnvironmentSlug, otherEnvironmentName } = req.body;
+
+  // atomic update the env to avoid conflict
+  const workspace = await Workspace.findById(workspaceId).exec();
+  if (!workspace) {
+    throw new Error("Failed to create workspace environment");
```
akhilmhdh, line 104: throw new BadRequestError

akhilmhdh (changes requested):
Other than slight naming changes this looks awesome to me. Great job

backend/src/controllers/v2/environmentController.ts
daninge98: This one actually isn't in code that I modified, but I switched all the ones I wrote to use a BadRequestError

vmatsiiako (approved):
This is awesome! Merging!

Issue #863, opened 2023-08-14
Secrets containing angle brackets are escaped from Web UI


harveysanders:
### Describe the bug
Web secrets UI escapes angle brackets included in a secret value.

#### Context:
Hi!
I have a few mailbox strings set in my infisical secrets. These strings sometimes contain angle brackets to include a display name along with the email address. If the angle brackets are include in a secret's value, that portion is not displayed in the Web UI. Thankfully, the actual value is not changed. I assume this is due to HTML escaping, but is there a way we can display the values as originally input?

### To Reproduce
1. Create a secret with a value that contains angle brackets `<` `>`
2. View the secrets in the web UI
3. The portion of the value within the brackets is not visible (presumably escaped for HTML sanitization)

### Expected behavior
The web UI should display the secret value as inputed

### Screenshots
```
$ infisical secrets set EMAIL_SUPPORT="Support <support@example.com>" --env=staging

┌───────────────┬───────────────────────────────┬───────────────────────┐
│ SECRET NAME   │ SECRET VALUE                  │ STATUS                │
├───────────────┼───────────────────────────────┼───────────────────────┤
│ EMAIL_SUPPORT │ Support <support@example.com> │ SECRET VALUE MODIFIED │
└───────────────┴───────────────────────────────┴───────────────────────┘
```

![image](https://github.com/Infisical/infisical/assets/9354822/a5256f37-e30d-4c9e-a7bb-a3550b05bf06)

akhilmhdh:
Duplicate of #820

subhajit20:
Hey @harveysanders  @akhilmhdh  I would like to take this issue.
Would you guys assign it to me?

harveysanders:
@subhajit20 It looks like @shlokmehta21 is taking it via #820. However, they needed some help debugging. Maybe you could give them some pointers on how to debug/log the issue

akhilmhdh:
Hey @harveysanders This has been fixed with our recent release. Let me know if your still facing the same issue. Closing this issue

PR #875, opened 2023-08-20
Convert arrow keys to html entities so it'll be visible in the dashboard

### LINKED ISSUE #820: Symbol `<` causes text to be removed
### Describe the bug

The syntax highlighter has an html stripper, the only catch is the stripper removes even if user has value `<` in it. Need to fix this

### To Reproduce
1. Add a value in dashboard with `< >` both. 
2. It will be hidden

### Expected behavior
1. Should be visible

### Screenshots

image

image

wasimTQ:
# Description 📣

Converted the angle brackets with html entities so that the content editable won't escape the characters. I don't know why we need to sanitize html but it's still there in `onBlur`

Fixes #820 and #863

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/components/v2/SecretInput/SecretInput.tsx
+++ b/frontend/src/components/v2/SecretInput/SecretInput.tsx
@@ -7,7 +7,30 @@ import { useToggle } from "@app/hooks";
 
 const REGEX = /\${([^}]+)}/g;
 const stripSpanTags = (str: string) => str.replace(/<\/?span[^>]*>/g, "");
+
+const entitiesToReplace = {
+  '<': '&lt;',
+  '>': '&gt;',
+}
+
+const escapeEntitiesFromString = (htmlString: string) => {
+  for (const [key, value] of Object.entries(entitiesToReplace)) {
+    htmlString = htmlString.replaceAll(key, value)
+  }
+
+  return htmlString;
+}
+
+const convertEscapedEntitiesToHTML = (value: string) => {
+  for (const [key, val] of Object.entries(entitiesToReplace)) {
+    value = value.replaceAll(val, key)
+  }
+
+  return value;
+}
+
 const replaceContentWithDot = (str: string) => {
+  str = convertEscapedEntitiesToHTML(str)
   let finalStr = "";
   let isHtml = false;
   for (let i = 0; i < str.length; i += 1) {
@@ -36,7 +59,7 @@ const syntaxHighlight = (orgContent?: string | null, isVisible?: boolean) => {
       `<span class="ph-no-capture text-yellow">&#36;&#123;<span class="ph-no-capture text-yello-200/80">${b}</span>&#125;</span>`
   );
 
-  return newContent;
+  return escapeEntitiesFromString(content);
 };
 
 const sanitizeConf = {
@@ -60,6 +83,7 @@ export const SecretInput = ({
   ...props
 }: Props) => {
   const [isSecretFocused, setIsSecretFocused] = useToggle();
+  const html = syntaxHighlight(value, isVisible || isSecretFocused)
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


wasimTQ:
@akhilmhdh I'd like to know what are the other changes that needs to be made here. And also whether my code broke anything else.

frontend/src/components/v2/SecretInput/SecretInput.tsx
```diff
@@ -36,7 +59,7 @@ const syntaxHighlight = (orgContent?: string | null, isVisible?: boolean) => {
       `<span class="ph-no-capture text-yellow">&#36;&#123;<span class="ph-no-capture text-yello-200/80">${b}</span>&#125;</span>`
```
wasimTQ, line 59: @akhilmhdh Also can you explain what this line of code does? I'd like to know because I removed the return type of this function.

wasimTQ: Good to know. I can replace it back with not removing the feature.

wasimTQ: And btw the dots are not calculated from the starting of `<` or `>`.

If the secret is `Trigger.dev <hello@trigger>` it has only 12 dots whereas it should have 27. I know where it's been calculated. It's happening because of the check `!isHtml && char !== "\n"` from `replaceContentWithDot`

wasimTQ: All I need to know is whether the dots should replace all the letters inside and including the angle brackets. Or is it better left like now?

wasimTQ: @akhilmhdh Other than the above comment. Anything else needs to be checked?

akhilmhdh: Let me check this and ill get back to you

akhilmhdh (changes requested):
It seems to completely broken for me. 

Sharing my screenshot:

Screenshot 2023-08-21 at 8 26 04 PM


Screenshot 2023-08-21 at 8 26 08 PM

akhilmhdh:
To reproduce kindly create a multi line secret with `${anything works}` and then save. 

Refresh the page and check it out if its still same with syntax highlighing

wasimTQ:
> To reproduce kindly create a multi line secret with `${anything works}` and then save.
> 
> Refresh the page and check it out if its still same with syntax highlighing

Got it. Will check it now

wasimTQ:
Suddenly I'm getting `502 Bad gateway` when trying to sign up.

Launched it using `docker-compose -f docker-compose.dev.yml up --build --force-recreate`
And used the credentials mentioned in the docs

wasimTQ:
@akhilmhdh I hope it's fixed now. BTW do we need to hide the content inside angle brackets including it? Right now it isn't shown as dots when the input is not on focus

wasimTQ:
Do you want it to be look like this

image

image

akhilmhdh:
Hey @wasimTQ 

It seems to broken again. 
Screenshot 2023-08-22 at 11 16 55 AM

When i type something and press enter this generates like this

wasimTQ:
@akhilmhdh I hope there won't be any issue in this PR. Can you please do one last review and let me know what you've found?

akhilmhdh:
Hey @wasimTQ 

It seems causing very much unexpected behavoir. I have attached video regarding it. I think ill look into this bug and raise the bug as it seems a bit harder than i expected to be for a good first issue and more users are complaining about it. 


https://github.com/Infisical/infisical/assets/31166322/35ebf478-7772-4667-bfc2-7908390c7c21

akhilmhdh:
Closing this PR and this has been resolved in our recent release. 

@wasimTQ Thank you so much for your efforts. It was quite insightful for me to get started on this one yesterday, on what was happening. It help me fix it way faster than i anticipated. It's just than more people were reporting and wanted the fix. Again thank you @wasimTQ for your valuable time on working on this. 

Team Infisical thank you. Let me know if you wanna work on another issue

PR #885, opened 2023-08-22
Password fixes - enforce max length, add checks (pwd breach, PII, low entropy), improved UX, deprecate common-passwords api

ragnarbull:
#  Description 📣

This change enforces a max password length (100 chars) and helps prevent a possible DDOS attack from the user accidentally or intentionally setting a very very long password. Although not severe due to Infisical's architecture (ie. using client-side Argon2id KDF), this change should still be implemented as per CWE-521 Weak Password Requirements. It prevents client-side network delays and follows best practices.

Other changes include deprecation of 1 API route (ie. api/v1/auth/getCommonPasswords) and removal of the associated text file of common passwords. These are no longer relevant and none fall within the strict password criteria (min 14 chars, 1 special char, 1 uppercase, 1 lowercase, 1 number, no more than 2 repeated consecutive chars). 

Instead, a secure client-side call to the external haveIBeenPwnd (HIBP) password API has been setup. The URI has been added to the CSP. This action involves hashing (SHA-1) the password in binary and splitting the first 5 hash chars to compare against the ranged hash tables (00000-FFFFF prefixes) that make up the ~700 mln long HIBP db. Random padding is added to prevent MitM attacks where the response body size could infer which of the 16^5 ranged hash tables is called. The API is fast (hosted by Cloudflare at global edge servers), reliable and free. There is no rate limit (confirmed by Troy Hunt of HIBP). The setup in this PR accounts for the possibility of network/Cloudflare errors (eg. 503) and only triggers the error for isBreachedPassword (which prevents the password being used) when the API returns true (ie. pwnd), not when there is an error. This should be reviewed alongside CWE-521 requirements. The fallback position is to the classic password criteria. 

All relevant error messages have been modified for the impacted client side actions (ie. set/reset/change password) and have been translated to the 6 supported languages. I only speak two of them (EN & FR) so have used Google Translate for the others (might want to check these!). 

I've also expanded the password requirements to be more user-friendly and internationally-friendly. eg. the one letter char enforcement now includes the top 25 most spoken languages (eg. Korean, Ukrainian, Farsi, Vietnamese etc) and emojis. Escape char & sequence  has been added as a restriction. At least one number/symbol/emoji has been added instead of at least 1 number AND at least 1 symbol. This is a better UX and follows sec best practices. The repeated consecutive chars includes any char or whitespace now as well. Checks for inclusion of basic low entropy data (email, url) and PII (SSN, passport, driver's license, phone number, DoB) have been added and restrict the user from including these in their new pwd. No external libraries have been used here.

Despite the size of this PR there are no breaking changes (only a deprecation).

**Summary:**

- Frontend:
     - Added 1 external API to CSP >>> https://api.pwnedpasswords.com/range/
     - Added a new folder >>> frontend/src/components/utilities/checks/password (moved checkPassword.ts & PasswordCheck.ts)
     - Added 2 new files >>> *./checkIsPasswordBreached.ts and *./passwordRegexes.ts
- Backend:
     - Deprecated 1 API >>> api/v1/auth/common-passwords
     - Removed 1 file >>> backend/src/data/common_passwords.txt
     
 **My suggested improvements along the same line **
 
 - url encode safe hash prefix string prior to sending to HIBP API (DONE)
 - Add error handling for escape character in pwd fields (DONE)
 - Review error translations (especially Korean & Turkish) & ensure they are implemented to the flow
 - Add unit testing for different attempted passwords (need to add jest to frontend)
 - Clean up repetition in code (eg. the reset password logic and components are effectively duplicated)
 - Improved UI for changePassword to indicate basic password checks have passed and make them responsive as the user types
 - Add a second row to the change password function to ensure the user changes to the password they want (prevents mistakes and lockout)!
 - Add more validation for low entropy data (eg. IP address, GPS coordinates) or PII (eg. physical address) (but not bloat the regexes...)
 - Add [zxcvbn into the mix.](https://github.com/dropbox/zxcvbn) - this looks up common passwords and words from Western media and can be added into a password strength meter. Again, you could get carried away adding non-Western words (the function needs to be very fast). Here's an example I did with different lightsaber bars as your password complexity levels up: https://github.com/Infisical/infisical/assets/83689282/2c146df3-d29f-4dd6-bc87-3ee885766383
 - Change password toggle password visibility (this is really dependent on Infisical security approach - a dev could leave their screen open with the password visibility set to true or be using a browser extension that can see the plaintext password - NB. GitHub does NOT use toggle password functionality (probably for these reasons)
 - If anything needs to be added to documentation, do that

## Type ✨

- [X] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Please see the attached snips showing the error messages and success. Please check the register and reset functions as well as testing has mostly been done with the changePassword action on the personal settings page. See video in comment below.
(as above, unit tests would be a great addition if other tests are added to the frontend)


ragnarbull:
Video showing flow on changing password from personal settings page
https://github.com/Infisical/infisical/assets/83689282/1f53e491-e632-4218-be58-f94d2607a1f1

ragnarbull:
Original issue fixed (no max password length enforced)
https://github.com/Infisical/infisical/assets/83689282/28049fd7-ac94-4707-8646-3765d5486014

dangtony98 (commented):
Just did a preliminary review of the code (I haven't actually tested the implementation) and have a few comments:

1. The height and width of the signup form appears to be distorted in this PR (I think some styling was updated; I'd check that the form looks right across multipe screen sizes).
2. I think we can remove the `validator` library from the frontend (we want to minimize unnecessary dependencies as much as possible as we evolve Infisical); I'd suggest either finding a way to validate email via regex or better yet use the `yup` library that we use for schema validation across the codebase — There is a method exposed on there that should allow you to validate the email.
3. Regarding the `createHash` method from the `crypto` module: I would suggest finding an equivalent method on `window.subtle.crypto` instead since that would be more reliable in the browser; the `crypto` node module should be used on the `backend` but I've run into many issues on the `frontend` with it lately (wasn't designed for it) — this is a gradual transition I'm trying to roll out across the frontend in general as it pertains to other crypto operations.
4. I think we should avoid non-ASCII characters (e.g. chars from other languages) for passwords. Since passwords in Infisical play an even more critical role in certain crypto operations, I want to be careful with this.

Will discuss this in more depth in the community call later!

ragnarbull:
Thanks @dangtony98 ! Working on these fixes now, I'll just send them through one by one (2/4 done). Will comment back when completed.

ragnarbull:
OK @dangtony98 I've fixed it all up. The forms were warped as some err messages were too long.

Here's a video https://www.veed.io/view/15448a80-6d2c-42df-b5d8-dd5ba82d18bd?panel=share (def speed it up 2x) going through the full flow of setting up an account, adding secrets from .env file, changing the pwd, viewing the secrets, resetting pwd, viewing the secrets. It runs through each user action with a test suite of valid and invalid pwds. 
[pwd-tests.txt](https://github.com/Infisical/infisical/files/12435097/pwd-tests.txt)

Can confirm that emojis and other unicode symbols encrypt & decrypt fine your AES-256-GCM implementation (same with the pwd hashing).

NB. I also changed the pwd criteria to be more user friendly. Rather than enforcing uppercase, lowercase, number & symbol I have made it: at least one letter char (from top 50 languages) + at least one number/symbol/emoji. 
Other criteria include 14-100 chars, no PII/low entropy data (eg. a SSN, an email, a URL etc - this list could be expanded in the future), no escape char/sequence, no breached pwd, no more than 3 repeating consecutive chars. 

I've translated all the errors but I couldn't confirm that the translations actually show in the errors - there seems to be an issue with this for everything in the personal settings page (in French anyway). They **should** all be short enough to not warp the register form again.

Let me know what you think and what needs to be fixed!

PS. lurkers only my email is real (same for .env secrets shown).

ragnarbull:
Woo! Great pick ups. Thanks @dangtony98 ! First PR ever 🎉

Issue #878, opened 2023-08-20
Show overridden personal secrets in overview page


decarbonise:
Currently, when viewing secrets in the overview page, the underlying value of secrets is shown and any overrides by personal values is ignored. We should:

1) Put the overridden secret icon in the grid of ticks and crosses where there is a secret that has been overwritten by a personal value, so that it is possible to easily identify overridden secrets. This is particularly necessary when quickly debugging secrets retrieval where results differ from expectations due to overridden secrets.

2) Show the overridden secret value when viewing the secret, the same as how it appears on the environment page when viewing a secret. This allows us to quickly see and edit personal secret overrides

SMLukwiya:
Working on this!🙂

Devmoni:
I've started implementing this and have covered part of the functionality mentioned.

Attached is a screenshot of the current output based on my understanding of the issue.

<img width="1536" height="482" alt="Image" src="https://github.com/user-attachments/assets/6507d0ef-33a8-4c4e-90dd-6e546a3810b4" />

Would appreciate any feedback , am I heading in the right direction? Let me know if there's anything I might have missed or could improve.

Thanks!

@decarbonise

akhilmhdh:
This has been implemented in our new dashboard revamp. Closing this issue.

PR #915, opened 2023-08-30
feat: show overridden secrets in overview

montocoder:
# Description 📣

This PR enhances the "secrets overview page" to display overridden personal secret values and introduces an "overridden secret" icon in the overview grid.

Related Issue: [#878](https://github.com/Infisical/infisical/issues/878)

Demo: https://share.vidyard.com/watch/BV36qSqheWdwNdAEpaietE?

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

1. Navigate to secret environment page
2. Create a secret and save
3. Mark the secret as "overridden" and change value of the secret to a personal overridden value
4. Navigate back to the "secret overview" page and the value should indicate overridden and show the overridden value.

---

- [x] I have read the [contributing guide](https://infisical.com/docs/contributing/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/code-of-conduct). 📝](https://share.vidyard.com/watch/BV36qSqheWdwNdAEpaietE?)


frontend/src/views/SecretOverviewPage/components/SecretOverviewTableRow/SecretOverviewTableRow.tsx
```diff
@@ -81,6 +86,11 @@ export const SecretOverviewTableRow = ({
                       <FontAwesomeIcon icon={faCircle} />
                     </Tooltip>
                   )}
+                  {isOverridden && (
```
akhilmhdh, line 89: So now how will the user's remove the override here

frontend/src/views/SecretOverviewPage/components/SecretOverviewTableRow/SecretOverviewTableRow.tsx
```diff
@@ -5,14 +5,16 @@ import {
...
+  faXmark,
+  faCodeBranch
 } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { twMerge } from "tailwind-merge";
 
 import { Button, TableContainer, Td, Tooltip, Tr } from "@app/components/v2";
 import { useToggle } from "@app/hooks";
 import { DecryptedSecret } from "@app/hooks/api/secrets/types";
+import { SecretActionType } from "@app/views/DashboardPage/DashboardPage.utils";
```
akhilmhdh, line 17: Kindly define it again here again rather than importing from another page as each page is independent

akhilmhdh (changes requested):

frontend/src/views/SecretOverviewPage/components/SecretOverviewTableRow/SecretOverviewTableRow.tsx
SMLukwiya: Thanks for the feedback.
I think the issue was just about displaying it. It's possible to remove the override in the detailed view.

> Put the overridden secret icon in the grid of ticks and crosses where there is a secret that has been overwritten by a personal value, so that it is possible to easily identify overridden secrets.

frontend/src/views/SecretOverviewPage/components/SecretOverviewTableRow/SecretOverviewTableRow.tsx
SMLukwiya: Implemented

frontend/src/views/SecretOverviewPage/components/SecretOverviewTableRow/SecretOverviewTableRow.tsx
akhilmhdh: Hey @SMLukwiya 

The overview page has a edit functionality where user's can edit and save secret. Now even thought your displaying it, when user's edit it your not handling the case where its editing the general or the override value. 

This will make user's confused. So requesting to implement secret override CRUD like dashboard here as well please. We cannot merge without that as this is more confusing for users

SMLukwiya: got it, looking into it.

SMLukwiya: Hey @akhilmhdh, I noticed the behavior in V3 of the api, where deleting an override from the overview page works but reverts the secret to the original value, and deleting the original secret removes the secret from an environment all together. In the dashboard, deleting an override behaves the same as deleting the original secret. Is this difference the current expected behavior? You can view this attachment of the current approach [short video](https://share.vidyard.com/watch/hScpupPjxatqyqvGoVduZ9?)

lokesh3097:
Hi everyone!

Is there any update regarding this PR or the issue #878? Is somebody still working on the issue?

Thanks!

PR #932, opened 2023-09-01
Secret scanning: scan for leaked secrets in historical commits

### LINKED ISSUE #881: Add historical scan capability to github secret scanning app
### Feature description
Currently, only new commits are scanned for exposed secrets. This leaves secrets exposed in old commits hidden. Right after users add git repositories, we should iniate a scan of all available repositories.

### Why would it be useful?
This will he helpful because users can find out leaked secrets from past commits

### Additional context
The repo scan should be initiated here https://github.com/Infisical/infisical/blob/main/backend/src/ee/services/GithubSecretScanning/GithubSecretScanningService.ts#L21

The secret scanning makes use of our CLI scanning, more details about this can be found here https://infisical.com/docs/cli/scanning-overview

daninge98:
Closes #881 

# Description 📣

Currently when a user connects their Github account to Infisical, we start scanning every future commit to see if there are any leaked secrets as part of that commit, but we don't look for leaked secrets in historical commits (i.e. commits made before the user started using Infisical secret scanning). This PR changes that.

Now when the user connects their repo to Infisical, we will schedule a full historical scan of their repo

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Tested locally using my own Github app.

Confirmed that after the user connects a repo, historically leaked secrets were flagged in the UI when the page was refreshed

---


daninge98:
I still need to test the email gets sent ok before merging, I will take a look at this tomorrow

maidul98:
Were you able to explore if it was possible to dowload the bare git repository without downloading the git binary into the docker image? One reason I wanted to avoid the git binary is because it makes the image a bit larger

daninge98:
> Were you able to explore if it was possible to dowload the bare git repository without downloading the git binary into the docker image? One reason I wanted to avoid the git binary is because it makes the image a bit larger

I had a look around but I couldn't work out a way to do it

maidul98 (approved):
I tested and the scan happens as expected, nice work. It would be nice if there was some visual indication that these scans were happening but this is something we can add later

PR #933, opened 2023-09-02
Add resource specification to frontend and backend deployment containers

MohamadTahir:
# Description 📣

You can specify how much resource your frontend/backend containers should use.

This is to assure
1- The Infisical containers don't consume more resources than expected.
2- The Infisical containers always have the resources that they asked for in the specs.

I've also added a `CHANGELOG.md` file to easily track the changes between each release, please have a look.

The resource limit/requests that are set are based on our monitoring of Infisical in our production environment.  

as for why 
1- I set the CPU request but no Memory request
2- I set a memory limit but no CPU limit

Please read the following articles to fully explain the reason 
- https://home.robusta.dev/blog/kubernetes-memory-limit
- https://web.archive.org/web/20220805232857/https://home.robusta.dev/blog/stop-using-cpu-limits/

This is a practice that we have been implementing in all of our clusters and we are greatly benefiting from it.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
To test these changes all you need to do is install the Infisical using the helm chart and then check if the containers have the `resources` annotation in their YAML manifest. you can do so by

```sh
> cd helm-charts/infisical
> helm install --dry-run --debug my-test-infisical .
```
---


helm-charts/infisical/README.md
```diff
@@ -91,6 +92,7 @@ kubectl get secrets -n <namespace> <secret-name> \
...
 | `backend.image.tag`                                    | Backend image tag                                                                                                                                                                                                                   | `latest`                    |
 | `backend.image.pullPolicy`                             | Backend image pullPolicy                                                                                                                                                                                                            | `IfNotPresent`              |
+| `backend.resources`                                    | Backend container resource spesification, check the offical [kubernetes documentations](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)                                                             | ``                          |
```
maidul98, line 95: How did you generate these docs? Please follow these steps to generate docs 

https://github.com/Infisical/infisical/blob/e47912edd79bd379f622a5b36af16bddc620d784/helm-charts/README.md?plain=1#L33-L39

helm-charts/infisical/values.yaml
```diff
@@ -43,6 +43,13 @@ frontend:
     ## @param frontend.image.pullPolicy Backend image pullPolicy
     ##
     pullPolicy: IfNotPresent
+  ## container resource specifications [check the offical kubernetes documentations](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/)
+  ##
+  resources:
+    limits:
+      memory: 100Mi
```
maidul98, line 50: Can you provide the metrics you used to come to this number?

MohamadTahir: I've set these numbers based on the metrics we have from our infisical deployment on our production cluster. I will send you a screenshot in the DM.

Frontend is typically using 0.5mCPU & 75MB of Memory

The backend is using around 70mCPU & 140MB of Memory.

maidul98 (approved):
Tested and looks good to me, thanks for the pr!

PR #955, opened 2023-09-08
Infisical Radar - ignore false positives

ragnarbull:
# Description 📣

User can flag leaked secrets on the frontend & notifications won't go out. Works for secrets with the same SHA3-512 hash in the repo. Checks .infisicalignore file and flags all fingerprints as false positives and persists them. Works for both the historical scan and new commits and displays the value on the secret scanning table. Added some filters on the frontend, the ability to download the secret scanning log table as a .csv file, and sorting of the date found. NB. removed two values from GitRisk schema (isFalsePositive and isResolved) as it is easier to use enum of GitRisk.RiskStatus.

Improvements needed:

- Fix unnecessary updating of .infisicalignore file on pushes (need to check if the risk status is already flagged as a false positive)
- UI needs work but is the start of a dashboard
- Add improvements to documentation incl. setup for self-hosting
- add instances of // infisical-scan:ignore as false positives 
- SHA3-512 hash is not exposed on the frontend but it could be if desired and included in the .csv download
- Append the hash to the fingerprint for more efficient lookups (ie. avoid need to query db)
- Break out some code for the historical and new push scans into helpers

## Type ✨

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Tested locally. Likely to be conflicts with risk status from // infisical-scan:ignore comment, .infisicalignore file, cloud secret scanning table etc


backend/src/queues/secret-scanning/githubScanFullRepository.ts
```diff
@@ -22,26 +24,77 @@ type TScanPushEventQueueDetails = {
...
-      }).lean()
+
+      // Create a SHA3-512 hash of the secret & immediately clear the plaintext secret from memory
+      // Used in push events to collate all SHA-3 hashes of the secret in the same repo's file (prevents case where the secret moves)
+      // Also useful to append to the fingerprint (eventually) & for user audit secret scanning log table .csv download
+      let secret = finding.Secret;
+      const sha512Hash = createHash("sha3-512");
+      sha512Hash.update(secret);
+      const hashResult = sha512Hash.digest("hex");
+      secret = "";
```
maidul98, line 52: I don't think this is necessary because garbage collection should take care of it

ragnarbull: You're right - only added this to ensure the secret is overridden as soon as it's not needed.

maidul98:
I love the filters, maybe you can look into how audit logs page is styled and copy some styles from there into the scanning page

ragnarbull:
> I love the filters, maybe you can look into how audit logs page is styled and copy some styles from there into the scanning page

Thanks! I've made them a bit better. Tried to use the form control from the audit logs but I couldn't get it the same due to how I'm doing the filtering. Everything else is fixed for now.

maidul98 (changes requested):

backend/src/queues/secret-scanning/githubScanFullRepository.ts
```diff
@@ -22,26 +24,77 @@ type TScanPushEventQueueDetails = {
...
+      const updateOperation = {
+        updateOne: {
+          filter: { fingerprint: finding.Fingerprint },
+          update: {
+            ...convertKeysToLowercase(finding),
+            installationId: installationId,
+            organization: organizationId,
+            repositoryFullName: repository.fullName,
+            repositoryId: repository.id,
+            hashedSecret: hashResult,
```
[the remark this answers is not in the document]
ragnarbull: @dangtony98 do you have some thoughts on this? I've used SHA3-512 here. The idea of putting the secret hash in is to more efficiently look through the repo and eventually search other repos for that secret. GitGuardian for example returns a 64 char base64 string ("secret_hash") when you download their .csv table. It's currently not exposed on frontend but maybe with RBAC an admin could get the table with the hashes?

backend/src/queues/secret-scanning/githubScanPushEvent.ts
```diff
@@ -51,40 +58,118 @@ githubPushEventSecretScan.process(async (job: Job, done: Queue.DoneCallback) =>
...
-
+          const fingerPrintWithCommitId = `${commit.id}:${filepath}:${finding.RuleID}:${finding.StartLine}`;
+          const fingerPrintWithoutCommitId = `${filepath}:${finding.RuleID}:${finding.StartLine}`;
+
+          // Create a SHA3-512 hash of the secret & immediately clear the secret from memory
+          let secret = finding.Secret;
+          const sha512Hash = createHash("sha3-512");
+          sha512Hash.update(secret);
+          const hashResult = sha512Hash.digest("hex");
+          secret = "";
```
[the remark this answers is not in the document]
ragnarbull: Taken out the set to empty str but  left the hashing

akhilmhdh:
@maidul98 @ragnarbull can you change the API calls for secret scanning from the old API client to the new one inside `hooks/api` which  uses react query. 

The one secret scanning uses is depreciated we are planning to remove it

ragnarbull:
> @maidul98 @ragnarbull can you change the API calls for secret scanning from the old API client to the new one inside `hooks/api` which uses react query.
> 
> The one secret scanning uses is depreciated we are planning to remove it

Thanks @akhilmhdh ! I've replaced with the new client and deprecated the old one.

backend/src/queues/secret-scanning/githubScanFullRepository.ts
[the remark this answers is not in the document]
maidul98: I think we can use Blind indexing here so that the secret is encrypted, but we can still query for it. We use blind indexing to encrypt secret names and then query for them later https://github.com/Infisical/infisical/blob/c3696bdbbc06e5bc13c78cc9fe19d321d6092f65/backend/src/services/SecretService.ts#L82

@dangtony98 can you confirm?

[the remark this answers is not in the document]
maidul98: Spoke to Tony, he also agrees that blind indexing should be added for this. You will likely have to create another table in the db to store these blind index details (we have one for secrets). Feel free to ask Tony about the specifics if you need help

ragnarbull (commented):
Added blind indexes for Git secrets instead of storing SHA3-512 hashes

ragnarbull (commented):
All changes should be addressed

backend/src/ee/models/gitRisks.ts
```diff
@@ -26,15 +29,12 @@ export type GitRisks = {
   ruleID: string;
   fingerprint: string;
   fingerPrintWithoutCommitId: string
-
-  isFalsePositive: boolean; // New field for marking risks as false positives
-  isResolved: boolean; // New field for marking risks as resolved
   riskOwner: string | null; // New field for setting a risk owner (nullable string)
```
maidul98, line 32: I see you added a new field but it isn't being used anywhere?

backend/src/helpers/secretScanning.ts
```diff
@@ -0,0 +1,437 @@
...
+ * for the GitHub organization's ID [organizationId].
+ * @param {Object} obj - The input parameters for creating Git secret blind index data.
+ * @param {String} obj.organizationId - GitHub organization's ID for Infisical Radar.
+ * @returns {Promise<string>} - A Promise that resolves to the generated Git secret blind index salt for the organization.
+ */
+export const createGitSecretBlindIndexDataHelper = async ({
+  organizationId
+}: CreateGitSecretBlindIndexDataParams): Promise<string> => { 
+  // initialize random blind index salt for the organization
+  const randomBytesPromisified = promisify(randomBytes);
```
maidul98, line 165: what is promisify and why do we need it? I don't see any instances were we were using it before

backend/src/helpers/secretScanning.ts
```diff
@@ -0,0 +1,437 @@
...
+ */
+export const createGitSecretsHelper = async ({
+  gitSecrets,
+  gitSecretBlindIndexes,
+  organizationId,
+  salt,
+  status
+}: CreateGitSecretsParams): Promise<string[]> => {
+
+  const gitSecretsArray = Array.isArray(gitSecrets) ? gitSecrets : Array.from(gitSecrets); // change from set to array for the push event
```
maidul98, line 265: Will `gitSecrets` be something other than an array?

backend/src/helpers/secretScanning.ts
```diff
@@ -0,0 +1,437 @@
...
+  const encryptedGitSecrets = await encryptGitSecretsHelper({ gitSecrets: gitSecretsArray }); // returns only unique values
+
+  const blindIndexesToInsert = new Set<string>(); // same indexing as the encryptedGitSecrets
+  const allBlindIndexes: string[] = [];
+
+  if (!gitSecretBlindIndexes || (Array.isArray(gitSecretBlindIndexes) && gitSecretBlindIndexes.length === 0)) {
```
maidul98, line 271: same for this, will this be something other than an array at some point?

backend/src/ee/models/gitRisks.ts
ragnarbull: This is old - before I worked on it. I removed it to use the specific risk statuses

backend/src/helpers/secretScanning.ts
ragnarbull: yeah it will either be an array or a set depending on whether it's coming from the full repo scan or the push. The reason is because in the push we need to create the blind index on the fly and so we can just pass the unique Git secrets. the full repo scan needs to still create the blind indexes, only add the unique ones to the GitSecret schema but add all of them against the corresponding GitRisks

maidul98: why don't we convert the set to an array before it gets passed to this function? That way the parameters stay consistent

Issue #938, opened 2023-09-03
Add a `--path` and `--env` ption to `infisical init`


NiklasRosenstein:
### Feature description

It would be nice if a `--path` and `--env` option could be added to the `infisical init` command to store the default path to read the secrets from in `.infisical.json` and not having to pass it to every `run` command.

### Why would it be useful?

Allows you to move more configuration into the `.infisical.json` file and have less to write in the CLI afterwards.

PR #968, opened 2023-09-12
feat: add `path` and `env` option to `init` cmd

quinton11:
## Description 📣

`Path` and `env` options included into init cmd to persist in workspace file. This is to prevent repeatedly passing `path` and `env` on every `run` cmd. [See](https://github.com/Infisical/infisical/issues/938)

### Changes
- Modified `workspaceConfigFile`, now includes the `defaultSecretsPath`


Salman2301:
https://github.com/Infisical/infisical/issues/1075

quinton11:
@maidul98 @akhilmhdh

maidul98:
@rawkode i think this flag might be helpful for you as it allows you to have different .infisical.json files for each of your repos

Issue #937, opened 2023-09-03
`infisical secrets --path` option gives error with `get` subcommand


NiklasRosenstein:
### Describe the bug

`infisical secrets --help` states that there is a `--path` option:

```
$ infisical --version
infisical version 0.12.1

$ infisical secrets --help
Used to create, read update and delete secrets

Usage:
  infisical secrets
  infisical secrets [command]

[...]

Flags:
      --env string          Used to select the environment name on which actions should be taken on (default "dev")
      --expand              Parse shell parameter expansions in your secrets (default true)
  -h, --help                help for secrets
      --include-imports     Imported linked secrets  (default true)
      --path string         get secrets within a folder path (default "/")

[...]
```

However, combining it with the `get` subcommand gives an error that the option does not exist.

```
$ infisical secrets get --path Hetzner/Cloud-Kubernetes --env prod HCLOUD_TOKEN
Error: unknown flag: --path
Usage:
  infisical secrets get [secrets]
[...]

$ infisical secrets --path Hetzner/Cloud-Kubernetes get --env prod HCLOUD_TOKEN
Error: unknown flag: --path
Usage:
  infisical secrets get [secrets]
[...]
```

### To Reproduce

See above.

### Expected behavior

The `--path` argument is accepted by the `infisical secrets` command.

### Platform you are having the issue on:

Linux

### Additional context

Infisical free plan self hosted.

quinton11:
Related https://github.com/Infisical/infisical/issues/910

Salman2301:
Isn't this fix? Close the issue?

maidul98:
Yes, `--path` has been added to `secrets get` command

PR #969, opened 2023-09-12
fix: `path` option for `secret get` sub cmd

quinton11:
# Description 📣

[See](https://github.com/Infisical/infisical/issues/937)


maidul98 (approved):
tested, lgtm

PR #981, opened 2023-09-14
fix: hide current env from dropdown while copying secrets

G3root:
# Description 📣

[Screencast 2023-09-14 16:11:35.webm](https://github.com/Infisical/infisical/assets/84864519/151c54b7-45d0-47f8-9099-d083163731ca)

## Type ✨

- [x] Bug fix


maidul98:
Is there a reason you think copying secrets from the same env should be blocked? I can think of a case where you have folders within the same environment and you want to copy secrets from a path within the same env into your current path. When you remove the current environment, you can't do that anymore

G3root:
This seems to be a quite fitting use case 💯 . I initially believed it to be a bug, and sorry for my wrong assumptions. I will now close this pull request.

Issue #982, opened 2023-09-14
[ENG-93] Alpine Linux CDN is unavailable - no fallback mirror or retry mechanism


ragnarbull:
### Describe the bug
When the Alpine CDN is down/unavailable there is no fallback mirror or retry mechanism. 

This infrequently occurs:
- https://github.com/gliderlabs/docker-alpine/issues/527
- https://stackoverflow.com/questions/53710206/docker-cant-build-because-of-alpine-error
- https://news.ycombinator.com/item?id=30341050

Looks like it has been down in East Coast Australia for >12 hrs (or it's just me).
This would (probably always) impact users in US-sanctioned countries (eg. Iran). not using a VPN.

### To Reproduce
1. Attempt to run Infisically locally using docker-compose
2. Error with building the backend stage

### Expected behavior
Docker should fetch infisical-cli from the official Alpine build hosted on Cloudsmith. 

### Screenshots
![Screenshot 2023-09-14 215918](https://github.com/Infisical/infisical/assets/83689282/cc36e2d1-f29f-4af9-b1d5-79793d1fc888)

### Platform you are having the issue on:
- Windows using bash in VS-code terminal

<sub>[ENG-93](https://linear.app/infisical/issue/ENG-93/alpine-linux-cdn-is-unavailable-no-fallback-mirror-or-retry-mechanism)</sub>

PR #983, opened 2023-09-14
Docs: Update FAQ for Alpine CDN error

ragnarbull:
# Description 📣

Docs to explain local development fix for issue #982 (ie. reconfigure backend Dockerfile with pinned Alpine Linux version,  add retry mechanism and a fallback Alpine Linux mirrors array). Links include archived links in case the global CDN is down.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Documentation

# Tests 🛠️

Checked on localhost docs


ragnarbull:
> This is great! I think it would actually be better under developer setup docs because when the image is built by us for prod, it won't be under the alpine cnd anymore, it will be a new image built on top of alpine hosted on docker hub. Maybe you can add a FAQ page here https://infisical.com/docs/contributing/overview
> 
> let me know if you have any questions

Oh yes that makes more sense. Removed from self-hosted and created new FAQ under contributing.

maidul98 (approved):
nice work!

PR #990, opened 2023-09-15
Update README.md

captain-Akshay:
# Description 📣

i think the place to create an account for a user is in the front-end so the port number should be 8080

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ X] Documentation


dangtony98:
Actually I just realized that this `localhost:80` should be correct cuz we're dealing with `docker-compose.yml` file not the `docker-compose.dev.yml` file here.

PR #991, opened 2023-09-15
feat(rbac): migration script for permission from old permission to new

akhilmhdh:
# Description 📣

1. Migration script from old to new

This will create a new role by inheriting member permission and set permission accordingly

# Tests 🛠️

```sh
```

---


backend/src/controllers/v1/membershipController.ts
```diff
@@ -148,7 +148,12 @@ export const changeMembershipRole = async (req: Request, res: Response) => {
   const membership = await Membership.findByIdAndUpdate(
     membershipId,
     {
-      role
+      $set: {
+        role
+      },
+      $unset: {
```
maidul98, line 154: why was this newly added?

maidul98: Oh i see because the user has set the role back to a default one

backend/src/utils/setup/backfillData.ts
```diff
@@ -679,22 +686,79 @@ export const backfillUserAuthMethods = async () => {
...
+      if (ability === "write") secretAccessRule[environmentSlug].write = false;
+      if (ability === "read") secretAccessRule[environmentSlug].read = false;
+    });
+
+    Object.entries(secretAccessRule).forEach(([envSlug, { read, write }]) => {
+      if (read) {
+        customPermissions.push({
+          subject: ProjectPermissionSub.Secrets,
+          action: ProjectPermissionActions.Read,
+          conditions: { environment: envSlug }
```
maidul98, line 719: we should also set the secrets path to `/**` so users have access to eveything within that environment just like they would right now

backend/src/utils/setup/backfillData.ts
```diff
@@ -679,22 +686,79 @@ export const backfillUserAuthMethods = async () => {
...
+      slug: `custom-role-${crypto.randomBytes(3).toString("hex")}`,
+      permissions: customPermissions
+    });
+    await role.save();
+    await Membership.findByIdAndUpdate(membership._id, {
+      $set: {
+        role: CUSTOM,
+        customRole: role
+      },
+      $unset: {
```
maidul98, line 757: let's leave this incase anything goes wrong, that way we can revert back chages

maidul98:
Added dedup role logic and added Redis for locking on to onc instance to run the backfill

maidul98 (approved):

PR #992, opened 2023-09-15
fix hyphen env variable import bug

serin0837:
# Description 📣

Just realized I can't import the .env file with the env variable named with hyphens and by changing the order of regex to [\w.:-] it fixed it. 
I know the best practice is using an underscore for the env variable name but it's weird that I can add env variable name in infisical dashboard with hyphens but can't import with .env file.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


vmatsiiako (approved):
That's a great catch! Merging!

Issue #926, opened 2023-09-01
Copy secrets from other environment – including folders


vmatsiiako:
### Feature description
Currently you can copy secrets from other environments. 
However, you are only able to copy secrets from a particular path – this action will not copy the secrets within the folders that a part of this particular path. It would be great if there was a feature flag (check box) that would copy all the subfolders from the current folder, too (aka not just the secrets). 

### Why would it be useful?
this would make environment replication so much easier and quicker. 

### Context
Some useful screenshots: 
<img width="1442" alt="Screenshot 2023-08-31 at 23 03 30" src="https://github.com/Infisical/infisical/assets/78047717/68d835be-c0de-4c9a-a669-0970b3373035">
<img width="741" alt="Screenshot 2023-08-31 at 23 03 57" src="https://github.com/Infisical/infisical/assets/78047717/9d32a3a7-f0a4-4991-90f3-6405113fca5f">

G3root:
hi @vmatsiiako i like to give this a try

vmatsiiako:
Hi @G3root! That would be very useful. Please join our Slack community in case you need any help or have any questions: https://infisical.com/slack

vmatsiiako:
@G3root, i believe the PR for this issue was just posted ^ 
Would you maybe want to find another issue?

akhilmhdh:
This is now supported with our new replicate feature. 

![Image](https://github.com/user-attachments/assets/b089d5ec-c61a-4d51-9be5-75afc440922f)

PR #993, opened 2023-09-15
Feat/env folders copy

quinton11:
# Description 📣

- Web UI - allows copying of folders from a  select environment and path - [See](https://github.com/Infisical/infisical/issues/926)

## New
- An endpoint to handle creating batch folders


quinton11:
@maidul98 @akhilmhdh 

Demo: 
https://github.com/Infisical/infisical/assets/70300837/19395034-9504-4215-a1c3-0ec5e3e07524

package.json
```diff
@@ -21,5 +21,8 @@
 	"devDependencies": {
 		"eslint": "^8.29.0",
 		"husky": "^8.0.3"
+	},
+	"dependencies": {
+		"@radix-ui/react-tabs": "^1.0.4"
```
ap0h, line 26: **question:** Shouldn't this be in the `/frontend/package.json`

quinton11: Right, must've run the install pkg cmd in the root directory

maidul98:
This PR is now marked stale. The codebase as changed drastically due to Postgres migration.

PR #999, opened 2023-09-18
fix: cli validating the environment via old method gets removed

akhilmhdh:
# Description 📣

1. Added migration for removal of denied permission field
2. Removed environment check in cli due to change in our new permission system

## Type ✨

- [x] Bug fix


maidul98:
Closing PR as there are two changes in one. I have pushed out these changes separately

PR #1018, opened 2023-09-22
Feat/dashboard v3

akhilmhdh:
# Description 📣

1. Implemented new dashboard v3 version
2. Migrated to v3 secrets api
3. Added new batch functions for v3
4. Updated  folder and import api to only using secret path
5. New ui for rollback screen

## Type ✨

- [x] New feature

# Tests 🛠️

```sh
```

---


maidul98:
I think the styling of the text area to add secrets could be improved. It should be full width to start 
Screenshot 2023-09-23 at 12 38 39 PM

maidul98:
In terms of style, I think tool bar looks a bit unconventional. Usually, when you select multiple items there is a drop down on the right side with actions. Maybe there can be some improvement here? 
Screenshot 2023-09-23 at 12 46 40 PM

For example, this is how github one looks 
Screenshot 2023-09-23 at 12 48 25 PM

maidul98:
I find this save icon unintuitive. Maybe a save button is more clear
Screenshot 2023-09-23 at 12 54 11 PM

akhilmhdh:
> I think the styling of the text area to add secrets could be improved. It should be full width to start. Screenshot 2023-09-23 at 12 38 39 PM
> 
> Also I think having a pop up to add secret creates quite a bit of friction. Is there a reason you added a pop up? i think it is faster to add secrets on our current UI than this one. I don't think we need a pop up, we should just keep the layout as is and add a button on the same row to confirm the change.

@maidul98 Its because we would soon have things like options to select multi environments when creating secrets, provide, create one in a folder etc

backend/src/helpers/secrets.ts
```diff
@@ -1173,8 +1232,557 @@ export const expandSecrets = async (
...
+  const exists = await Secret.exists({
+    workspace: new Types.ObjectId(workspaceId),
+    folder: folderId,
+    environment
+  })
+    .or(
+      secrets.map(({ secretName, type }) => ({
+        secretBlindIndex: secretBlindIndexes[secretName],
+        type: type,
+        ...getAuthDataPayloadUserObj(authData)
```
maidul98, line 1293: what if a different user tried to create the same secret? in that case `user` would be different so exists would be false and the rest of the logic would continue?

backend/src/helpers/secrets.ts
```diff
@@ -1173,8 +1232,557 @@ export const expandSecrets = async (
...
+    )
+  ).then((blindIndexes) =>
+    blindIndexes.reduce<Record<string, string>>((prev, curr, i) => {
+      prev[secrets[i].secretName] = curr;
+      secretBlindIndexToKey[curr] = secrets[i].secretName;
+      return prev;
+    }, {})
+  );
+
+  const secretsToBeUpdated = await Secret.find({
```
maidul98, line 1477: why do we do this check? I'm thinking if the secrets that you want to update do not exist, then no update will be made for that given secret so we don't need this. let me know if i missed something

maidul98: oh i see, it's for createAuditLog

backend/src/helpers/secrets.ts
```diff
@@ -1173,8 +1232,557 @@ export const expandSecrets = async (
...
+
+  const deletedSecrets = await Secret.find({
+    workspace: new Types.ObjectId(workspaceId),
+    folder: folderId,
+    environment
+  })
+    .or(
+      secrets.map(({ secretName, type }) => ({
+        secretBlindIndex: secretBlindIndexes[secretName],
+        type: type === "shared" ? { $in: ["shared", "personal"] } : type,
```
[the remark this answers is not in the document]
maidul98: actually, it makes sense. when the secret you delete is shared, it should delete both personal version of that secret and shared version

maidul98 (approved):
lgtm

PR #1027, opened 2023-09-26
Service Token V3

dangtony98:
# Description 📣

This PR adds the preliminary Service Token V3 (ST V3) feature that is a new authentication method to replace the current Service Token (ST V2) structure over time.

Comparison of ST V3 vs ST V2:
- ST V3 exports a `JSON` file containing 3 components: `privateKey`, `publicKey`, and `serviceToken` where `serviceToken` is a JWT token prefixed (`proj_token`) to access the API and the public-private key pairs are to support cryptographic operations for the client whenever E2EE is needed.
- ST V3 avoids user misuse of ST V2 because it separates encryption/decryption keys from the authentication token into separate components; this is a significant security improvement for E2EE use-cases.
- ST V3 supports IP allowlisting; this means you can now restrict usage of a ST V3 to a specific IP or CIDR range.
- ST V3 supports toggling on/off active states; this removes the need to delete a service token to render it inactive, though this is also supported.
- ST V3 supports granular `read` or `readWrite` permissions down to each path level.
- ST V3 includes `lastUsed` timestamps.
- ST V3 cryptography is designed so that project keys can be rotated without rendering service tokens inactive unless explicitly toggled inactive or removed; it achieves this by ditching one-off symmetric encryption of project key copies and instead providing users the ability to encrypt new project keys under the public key of the service.
- ST V3 makes progress towards standardizing API-wide authentication to JWT format (as opposed to supporting a mix of API keys, JWTs, etc.); in the future, all tokens will be prefixed to indicate what kind of token is being checked.

To note, the feature should be considered in Beta for the time being.

See screenshot below for visual comparison of current Service Token vs the new Service Token V3 (Beta):

Screen Shot 2023-09-30 at 9 55 41 PM

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


frontend/src/views/Settings/ProjectSettingsPage/components/ServiceTokenV3Section/ServiceTokenV3Table.tsx
```diff
@@ -0,0 +1,240 @@
...
+        } catch (err) {
+            console.log(err);
+            createNotification({
+                text: `Failed to ${isActive ? "enable" : "disable"} service token v3`,
+                type: "error"
+            });
+        }
+    }
+    
+    const formatDate = (dateToFormat: string) => {
```
akhilmhdh, line 74: Use date-fn for this

frontend/src/views/Settings/PersonalSettingsPage/APIKeySection/APIKeyTable.tsx
```diff
@@ -49,56 +49,54 @@ export const APIKeyTable = () => {
...
+        <TBody>
+          {isLoading && <TableSkeleton columns={4} innerKey="api-keys" />}
+          {!isLoading &&
+            data &&
+            data.length > 0 &&
+            data.map(({ _id, name, createdAt, expiresAt, lastUsed }) => {
+              return (
+                <Tr className="h-10" key={`api-key-${_id}`}>
+                  <Td>{name}</Td>
+                  <Td>{formatDate(lastUsed)}</Td>
```
akhilmhdh, line 72: date-fns

akhilmhdh (changes requested):

backend/src/config/index.ts
```diff
@@ -26,6 +26,7 @@ export const getJwtSignupLifetime = async () => (await client.getSecret("JWT_SIG
 export const getJwtProviderAuthSecret = async () => (await client.getSecret("JWT_PROVIDER_AUTH_SECRET")).secretValue;
 export const getJwtProviderAuthLifetime = async () => (await client.getSecret("JWT_PROVIDER_AUTH_LIFETIME")).secretValue || "15m";
 export const getJwtSignupSecret = async () => (await client.getSecret("JWT_SIGNUP_SECRET")).secretValue;
+export const getJwtServiceTokenSecret = async () => (await client.getSecret("JWT_SERVICE_TOKEN_SECRET")).secretValue;
```
maidul98, line 29: idea: what if we set all of the jwt secrets to the single encryption key? Feel like we have a lot of JWT secrets. If one JWT gets compromised, surely the others would too?

backend/src/controllers/v3/secretsController.ts
```diff
@@ -59,29 +74,52 @@ export const getSecretsRaw = async (req: Request, res: Response) => {
...
+        requiredPermissions: [Permission.READ]
+      });
+      permissionCheckFn = (env: string, secPath: string) =>
+        isValidScopeV3({
+          authPayload: req.authData.authPayload as IServiceTokenDataV3,
+          environment: env,
+          secretPath: secPath,
+          requiredPermissions: [Permission.READ]
+        });
+      break;
```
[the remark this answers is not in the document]
maidul98: yeah it would be good to reduce some of repetition using a helper function of some sorts

backend/src/ee/controllers/v3/serviceTokenDataController.ts
```diff
@@ -0,0 +1,321 @@
...
+        expiresAt = new Date();
+        expiresAt.setSeconds(expiresAt.getSeconds() + expiresIn);
+    }
+    
+    let user;
+    if (req.authData.actor.type === ActorType.USER) {
+        user = req.authData.authPayload._id;
+    }
+    
+    const isActive = false;
```
maidul98, line 115: i'm thinking this should be true by default because why would you create a inactive service token? Makes more sense to deactivate it later

backend/src/ee/routes/v1/sso.ts
```diff
@@ -30,6 +30,7 @@ router.get(
 router.get("/redirect/github", authLimiter, (req, res, next) => {
   passport.authenticate("github", {
     session: false,
+    scope: [ 'user:email' ],
```
maidul98, line 33: nit: unrelated change right?

backend/src/ee/routes/v3/serviceTokenData.ts
```diff
@@ -0,0 +1,39 @@
+import express from "express";
+const router = express.Router();
+import { requireAuth } from "../../../middleware";
+import { AuthMode } from "../../../variables";
+import { serviceTokenDataController } from "../../controllers/v3";
+
+router.get(
```
maidul98, line 7: i didn't see this getting used anywhere, do we still need it?

backend/src/helpers/auth.ts
```diff
@@ -65,22 +71,29 @@ export const validateAuthMode = ({
...
 
-		switch (tokenValue.split(".", 1)[0]) {
+		const parts = tokenValue.split(".");
+		
+		switch (parts[0]) {
 			case "st":
 				authMode = AuthMode.SERVICE_TOKEN;
+				authTokenValue = tokenValue;
+				break;
+			case "proj_token":
```
maidul98, line 89: nit: maybe we should call it `stv3`

backend/src/helpers/auth.ts
```diff
@@ -382,11 +460,15 @@ export const createToken = ({
...
 }: {
 	payload: any;
-	expiresIn: string | number;
+	expiresIn?: string | number;
 	secret: string;
 }) => {
 	return jwt.sign(payload, secret, {
-		expiresIn,
+		...(
+			(expiresIn !== undefined && expiresIn !== null)
```
maidul98, line 468: nit: you can do short hand `expiresIn != null`

backend/src/models/serviceTokenDataV3.ts
```diff
@@ -0,0 +1,129 @@
...
+    publicKey: string;
+    isActive: boolean;
+    lastUsed?: Date;
+    usageCount: number;
+    expiresAt?: Date;
+    scopes: Array<IServiceTokenV3Scope>;
+    trustedIps: Array<IServiceTokenV3TrustedIp>;
+}
+
+const serviceTokenDataV3Schema = new Schema(
```
maidul98, line 35: I see that we have both a service token data v3 and service token data v3 key. Can you please explain high level how the crypto works for this?

backend/src/validation/serviceTokenDataV3.ts
```diff
@@ -0,0 +1,119 @@
...
+  });
+  
+  if (!serviceTokenData.workspace.equals(workspaceId)) {
+    // case: invalid workspaceId passed
+    throw UnauthorizedRequestError({
+      message: "Failed service token authorization for the given workspace"
+    });
+  }
+  
+  if (environment) {
```
maidul98, line 49: under which case will `environment` be empty? I think isValidScopeV3 should always run unless i'm missing something

backend/src/validation/workspace.ts
```diff
@@ -57,8 +58,11 @@ export const validateClientForWorkspace = async ({
         environment,
         requiredPermissions
       });
-
-      return {};
+      return { membership, workspace};
+    case ActorType.SERVICE_V3:
```
maidul98, line 62: how come we default to UnauthorizedRequestError for this when for v2 service token we try to validate it

backend/src/models/serviceTokenDataV3.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export interface IServiceTokenDataV3 extends Document {
+    _id: Types.ObjectId;
+    name: string;
+    workspace: Types.ObjectId;
+    user: Types.ObjectId;
+    publicKey: string;
+    isActive: boolean;
+    lastUsed?: Date;
```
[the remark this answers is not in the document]
dangtony98: Hmm I’m not sure about this one - I think the Date type here is used pretty commonly throughout the codebase - Will check how this compares to the default time stamp types.

maidul98:
Screenshot 2023-10-04 at 8 59 09 PM

When updating a path, you don't see what the current path is. it just defaults to `/`

backend/src/config/index.ts
dangtony98: I might have a better idea for this part. I need to go over the security implications first - We shouldn’t just set it to the encryption key at all; this would have severe implications on parts like MFA currently that’s contingent on differentiation between the secret for that temporary token and regular JWT token.

An option would be to use the encryption key as a seed to generate 6 other secrets sequentially.

Another option might be to programmatically generate the JWT secrets on the backend randomly and encrypt them under the encryption key.

backend/src/ee/routes/v1/sso.ts
dangtony98: Yeah. It must’ve got caught in the update for this but unrelated.

backend/src/ee/routes/v3/serviceTokenData.ts
dangtony98: I’m not sure I understand which part is not needed - I believe all these new endpoints are needed?

backend/src/helpers/auth.ts
dangtony98: It can still be undefined which is different from null.

backend/src/models/serviceTokenDataV3.ts
dangtony98: Yup different approach than V2.

Summary: ST V3 has a public-private key pair like a user. You share a copy of the project key with ST V3 by encrypting it under the public key of ST V3 - Exact same procedure as what we have for users and bots.

On the client-side, ST V3 decrypts the copy of the project key using its private key.

backend/src/validation/serviceTokenDataV3.ts
dangtony98: In the future case where ST V3 may be provisioned project-level actions that are not bound to a specific environment / path - not applicable to this PR specifically.

backend/src/validation/workspace.ts
dangtony98: Because this function is deprecated.

We are stripping out all middlewares because we moved it to controller-level validation so I figured to just put this in place for now since I will be deleting this function from the codebase entirely.

dangtony98: It’s also not used anywhere anymore and where ST V3 might be relevant.

docs/documentation/platform/token.mdx
```diff
@@ -3,30 +3,97 @@ title: "Service token"
 description: "Infisical service tokens allows you to programmatically interact with Infisical"
 ---
 
-Service tokens play an integral role in allowing programmatic interactions with an Infisical project, functioning as digital token that open access to specific project resources such as secrets.
+Service tokens are authentication credentials that services can use to access designated endpoints in the Infisical API to manage project resources like secrets.
```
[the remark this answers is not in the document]
dangtony98: That works.

docs/internals/service-tokens.mdx
```diff
@@ -5,10 +5,13 @@ description: "Understanding service tokens and their best practices"
 ​
 Many clients use service tokens to authenticate and read/write secrets from/to Infisical; they can be created in your project settings.
 
+On this page, we discuss Service Token V3, the new and improved authentication method.
```
maidul98, line 8: same note for here, let's hide v3 docs for now until the feature is made visible to users (which will be when all the clients are compatible with v3 tokens)

dangtony98:
> I think we should let users copy the service token by defualt and if they want to download it they can. I think downloading adds a bit of friction because now if you want to use the service token for k8s or cli, you have to open it with a text editor then copy it over so you can paste it.
> 
> We should allow them to copy the service token as a single line so it's easy to paste into cli and k8 purpose.
> 
> otherwise, looks great, just a few small comments from my end. After, we can put it behind feature flag until the clients are brought up to speed and merge to main

I was hoping to avoid the copy from clipboard stuff actually; I was going for the model found in (1) GCP where you download the service account JSON and (2) Azure - When setting up SAML SSO, you also download the cert instead of viewing it in browser UI; developers have a way of being able to work with it anyways and shouldn’t find it a problem.

The other consideration I was thinking that’s nice about exporting a file is that it may be fully possible that they want to use that file like in GCP. They may add the file to their codebase even and git-ignore it or something.

I think this one is more of a convenience UX thing and I was looking to ship the feature faster - I’ve been thinking actually to move away from copy to clipboard wherever possible; it’s extra overhead.

dangtony98:
Will make adjustments in the morning; should be ok afterward.

The main thing I wanted to get tested here is that the JWT works and the permissions for it.

backend/src/helpers/auth.ts
dangtony98: Adjusted to `stv3`.

backend/src/controllers/v3/secretsController.ts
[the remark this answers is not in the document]
dangtony98: Refactored in the latest update.

PR #1028, opened 2023-09-27
Feature: MFA methods (authenticator app & recovery codes)

ragnarbull:
# Description 📣

New feature that extends the current 2FA offering (ie. email). New methods are:

1. Authenticator app (either TOTP or the two-factor secret key)
2. MFA recovery codes

The two-factor seed key and the recovery codes are CSPRNG generated base32 strings that are then encrypted with AES-256-GCM, each with their own random salts & IVs. The secret key is reusable, the codes are not and are decremented immediately after use to prevent replays. MFA with email only configured !== recovery codes (ie. if you enable auth app, then disable it, your MFA recovery codes will be removed). The user can set a preference for their option and is redirected to that option after login. If no preference is set, an MFA selection modal will appear. The TOTP codes are seeded by the key and compared to the user's input from their authenticator app - compatible apps are listed in code and a link but will be added to docs (anything using SHA-256 should work). Note that a few emails are sent to the user (eg. enable MFA with auth app/create new MFA recovery codes, MFA recovery code was used, MFA was disabled). 

Libraries added:

1. qrcode & types - generate QR code for auth app on the front end - several dependencies
2. couple of type files were updated

TODO:

1. Ensure CLI login works (especially secret key & recovery codes as may be issues)
3. Fix bug where email MFA token is sent for each MFA login attempt even with non-email option (DONE)
4. Documentation on new methods
5. Ensure existing users with email configured will be able to access the configure MFA modal under personal settings (double check user.mfaMethods change to an enum of MfaMethods)
6. Check links in the 3 emails mentioned above - eg. audit logs. Need to add links for licensed self-hosting
7. Security keys will be added at a later date in a separate PR
8. MFA modal to access authorized resources will be added (as above)
9. Ability for admin to see who has MFA enabled and which methods and track creation/update timestamps (as above)

## Type ✨

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Videos as posted on Slack with improvements. Thoroughly tested flow except through CLI login.

---


ragnarbull:
For anyone wondering, I also added security keys (eg. YubiKey, Windows Hello, Android, Apple FaceID) to this PR but was told to remove them. Would be good if there is an update of when this is getting looked at as you'll probably want me to respond to how this stuff works... @dangtony98

maidul98:
This PR is now stale. The backend structure has changed completely due to Postgres migration. Closing for now

PR #1031, opened 2023-09-27
Secret approval policies feature

akhilmhdh:
# Description 📣

Phase 1 for secret approval with backend api and frontend ui for secret policy 

## Type ✨

- [x] New feature

# Tests 🛠️

```sh
```

---


akhilmhdh:
Two things:

Squash this commit on merge @maidul98 because i don't want that checkpoint commit message, just the message content is correct

Second this is behind a feature flag `NEXT_PUBLIC_SECRET_APPROVAL` set is true then sidebar will have the section

maidul98:
when you enter a number it gives this error 
Screenshot 2023-09-27 at 1 07 28 PM

backend/src/controllers/v1/secretApprovalController.ts
```diff
@@ -0,0 +1,109 @@
...
+  );
+  ForbiddenError.from(permission).throwUnlessCan(
+    ProjectPermissionActions.Edit,
+    ProjectPermissionSub.SecretApproval
+  );
+
+  const updatedDoc = await SecretApproval.findByIdAndUpdate(id, {
+    approvals,
+    approvers,
+    $set: secretPath === "-" ? undefined : { secretPath }
```
maidul98, line 61: would would secret path be `-`?

backend/src/models/secretApproval.ts
```diff
@@ -0,0 +1,44 @@
...
+    workspace: {
+      type: Schema.Types.ObjectId,
+      ref: "Workspace",
+      required: true
+    },
+    approvers: [
+      {
+        // user associated with the personal secret
+        type: Schema.Types.ObjectId,
+        ref: "Membership"
```
maidul98, line 23: curious why you choose `Membership`. I would think approvers are users. I also don't see Membership of secretApprovalSchema being used anywhere yet

backend/src/models/secretApprovalRequest.ts
```diff
@@ -1,81 +1,65 @@
...
-		},
-		requestId: {
-			type: String,
-			required: false,
-		},
-	},
-	{
-		timestamps: true,
-	}
+const secretApprovalSchema = new Schema<ISecretApprovalRequest>(
```
maidul98, line 33: are we using this? i see another model named `SecretApproval` but i see you made changes to this too

maidul98: if we are not using it, please delete it

akhilmhdh: No no we are using it. 

SecretApproval is the configuration is the tule itself. Rrquest contains the changes made wrt to the config. 


So there can be x number of rules. But a change made will be against at a time one rule.

backend/src/models/secretApproval.ts
akhilmhdh: It should be membership because using user id is a idea.

When user leaves org membership  will be also removed. If its user id based even when user leaves they will be able to see it or checks would be complicated.

backend/src/controllers/v1/secretApprovalController.ts
akhilmhdh: This is to remove that particular field.

As its a patch even if undefined it could mean the user just didn't provide  the option. 

Now if we give undefined mongoose omits it too. So i had to set it up like this.

backend/src/models/secretApproval.ts
maidul98: makes sense

backend/src/models/secretApprovalRequest.ts
```diff
@@ -1,81 +1,65 @@
...
+    ]
+  },
+  {
+    timestamps: true
+  }
 );
 
-export const SecretApprovalRequest = model<ISecretApprovalRequest>("SecretApprovalRequest", secretApprovalRequestSchema);
\ No newline at end of file
+export const SecretApproval = model<ISecretApprovalRequest>("SecretApproval", secretApprovalSchema);
```
maidul98, line 65: I was confused because both were called `SecretApproval`. Why don't we call this `SecretApprovalRequest`?

backend/src/models/secretApproval.ts
```diff
@@ -0,0 +1,44 @@
...
+      type: Number,
+      default: 1
+    }
+  },
+  {
+    timestamps: true
+  }
+);
+
+export const SecretApproval = model<ISecretApproval>("SecretApproval", secretApprovalSchema);
```
maidul98, line 44: Let's call this and the file SecretApprovalPolicy

backend/src/models/secretApprovalRequest.ts
akhilmhdh: Done. My bad

backend/src/models/secretApproval.ts
akhilmhdh: Done

maidul98 (approved):
lgtm

PR #1047, opened 2023-10-03
Add version tags to Docker image & display on frontend

ragnarbull:
# Description 📣

Add current Infisical version to the backend & frontend Docker images as tags during the GitHub Actions prod build. Inject that version number into the frontend (and replace it so it's dynamic), then display this on the client at the bottom left of the sidebar.

TODO: update docs, check GHA works to inject the version to Docker image tags

## Type ✨

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Not tested the GitHub Actions workflow as discussed.

---

### DIFF
--- a/.github/workflows/build-docker-image-to-prod.yml
+++ b/.github/workflows/build-docker-image-to-prod.yml
@@ -39,7 +39,10 @@ jobs:
           token: ${{ secrets.DEPOT_PROJECT_TOKEN }}
           load: true
           context: backend
-          tags: infisical/backend:test
+          tags: |
+            infisical/backend:${{ steps.extract_version.outputs.version }}
+          labels: |
+            version=${{ steps.extract_version.outputs.version }}
       - name: ⏻ Spawn backend container and dependencies
         run: |
           docker compose -f .github/resources/docker-compose.be-test.yml up --wait --quiet-pull
@@ -58,9 +61,10 @@ jobs:
           context: backend
           tags: |
             infisical/backend:${{ steps.commit.outputs.short }}
-            infisical/backend:latest
             infisical/backend:${{ steps.extract_version.outputs.version }}
           platforms: linux/amd64,linux/arm64
+          labels: |
+            version=${{ steps.extract_version.outputs.version }}
 
   frontend-image:
     name: Build frontend image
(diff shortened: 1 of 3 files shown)


maidul98:
Just took a look, what you can do instead is:
- set the NEXT_INFISICAL_PLATFORM_VERSION as a build argument in github action. set it equal to the extracted version used for tagging already
- then assign this build argument in the docker file to a Next js environment varable. We do something similar to this [here](https://github.com/Infisical/infisical/blob/main/frontend/Dockerfile#L3) 
- Then in the frontend app, you just reference it as an environment variable proccess.env.NEXT_INFISICAL_PLATFORM_VERSION

let me know if you have questions via slack

ragnarbull:
@maidul98 thanks I've made the changes

frontend/src/layouts/AppLayout/AppLayout.tsx
```diff
@@ -696,6 +697,10 @@ export const AppLayout = ({ children }: LayoutProps) => {
                       </div>
                     </button>
                   )}
+                <div className="mb-2 w-full pl-5 duration-200 hover:text-mineshaft-200">
+                  <FontAwesomeIcon icon={faInfo} className="mr-4 px-[0.1rem]" />
+                    Platform Version: {process.env.NEXT_PUBLIC_INFISICAL_PLATFORM_VERSION || "N/A"}
```
maidul98, line 702: if it's empty we can just avoid showing this whole block

.github/workflows/build-docker-image-to-prod.yml
```diff
@@ -90,9 +94,13 @@ jobs:
           token: ${{ secrets.DEPOT_PROJECT_TOKEN }}
           project: 64mmf0n610
           context: frontend
-          tags: infisical/frontend:test
+          tags: |
+            infisical/frontend:${{ steps.extract_version.outputs.version }}
```
maidul98, line 98: this is for creating a test image before publishing the real thing. you can revert this

.github/workflows/build-docker-image-to-prod.yml
```diff
@@ -58,9 +61,10 @@ jobs:
           context: backend
           tags: |
             infisical/backend:${{ steps.commit.outputs.short }}
-            infisical/backend:latest
             infisical/backend:${{ steps.extract_version.outputs.version }}
           platforms: linux/amd64,linux/arm64
+          labels: |
+            version=${{ steps.extract_version.outputs.version }}
```
maidul98, line 67: I don't think we need this. Any reason you added labels ?

maidul98:
also you can test this change by setting build args in docker compose just like it would be set in the github workflow. check this https://stackoverflow.com/questions/50734271/how-to-define-build-args-in-docker-compose

.github/workflows/build-docker-image-to-prod.yml
ragnarbull: just metadata on the image, agree it's not needed

ragnarbull:
> also you can test this change by setting build args in docker compose just like it would be set in the github workflow. check this https://stackoverflow.com/questions/50734271/how-to-define-build-args-in-docker-compose

tested like this and works.

docker-compose-dev.yml

`    build:
      context: ./frontend
      dockerfile: Dockerfile.dev
      args:
        - NEXT_INFISICAL_PLATFORM_VERSION=v1.2.3`
        
Dockerfile.dev

`# Copy all files 
COPY . .

ARG NEXT_INFISICAL_PLATFORM_VERSION=development
ENV NEXT_PUBLIC_INFISICAL_PLATFORM_VERSION=${NEXT_INFISICAL_PLATFORM_VERSION}
RUN echo "NEXT_PUBLIC_INFISICAL_PLATFORM_VERSION is $NEXT_INFISICAL_PLATFORM_VERSION"

CMD ["npm", "run", "dev"]`

maidul98 (approved):
lgtm

PR #1067, opened 2023-10-09
Delete user, organization, project capabilities feature/update

dangtony98:
# Description 📣

This PR includes new and updated capabilities related to deleting the resources: `user`, `organization`, `workspace` (project). More specifically:

- Delete user account.
- Delete organization.
- Delete workspace/project.

When deleting a user:

- The deletion will fail if the user is part of an organization with >1 where there are no other members with the `admin` role; such case implies that the organization would be left admin-less.
- If the user is part of an organization with 1 member, the organization is deleted.
- The user's data is removed from Infisical (e.g. if they belonged to other workspaces, their `membership` is removed); any audit logged data containing that user's data part of an organization that will not get deleted is kept to preserve the integrity of audit logging — removing this data would imply that a bad actor can perform malicious actions and then remove themselves from Infisical without leaving any traces behind; this can be subject to change.

When deleting an organization:

- If the user is no longer part of any organizations, they are redirected to `/org/none` to create a new organization; this applies upon logging in as well.

Note that:

- Deleting an organization also removes its sub-resources.
- Deleting a workspace/project also removes its sub-resources.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


backend/src/helpers/organization.ts
```diff
@@ -65,6 +104,227 @@ export const createOrganization = async ({
...
+ * @param {Object} obj
+ * @param {Types.ObjectId} obj.organizationId - id of organization to delete
+ * @returns 
+ */
+export const deleteOrganization = async ({
+  organizationId
+}: {
+  organizationId: Types.ObjectId;
+}) => {
+  const session = await mongoose.startSession();
```
akhilmhdh, line 118: To support mongodb transaction. I think the user's need to enable replicaSet in infra level at mongodb. 

Don't you think this would break  self hosted users

backend/src/helpers/user.ts
```diff
@@ -134,3 +156,169 @@ export const checkUserDevice = async ({
...
+
+    if (orgMemberCount > 1 && otherOrgAdminCount === 0) {
+      throw InternalServerError({
+        message: "Failed to delete account because an org would be admin-less"
+      });
+    }
+  }
+
+  // delete workspaces where user is only member
+  for await (const membership of memberships) {
```
akhilmhdh, line 199: Do we really need this check. I mean if org is going to delete the just delete all workspace too.

backend/src/helpers/user.ts
```diff
@@ -134,3 +156,169 @@ export const checkUserDevice = async ({
...
+    await Key.deleteMany({
+      receiver: user._id
+    });
+
+    const membershipOrgs = await MembershipOrg.find({
+      user: userId
+    });
+  
+    // delete organizations where user is only member
+    for await (const membershipOrg of membershipOrgs) {
```
akhilmhdh, line 274: This is a bad practice as this shoots individually a lot of ones. 

Also say i see transactions is started this would lock up the db rows too. causing even delay to rest of the other operations. I would say just do a small aggregation query to fetch all org with just user as member

akhilmhdh (changes requested):

backend/src/helpers/organization.ts
dangtony98: I was not aware actually; will discuss with @maidul98

backend/src/helpers/user.ts
dangtony98: This one actually isn't about the top-level organization being deleted.

It's more-so that the user may be part of projects in another organization (that will not be deleted); we need to check that the project has another `admin` or else it will be admin-less when the user gets deleted.

backend/src/helpers/organization.ts
maidul98: mongodb >= 3.6.0 supports transactions. i also checked document db and it also supports transactions

maidul98: good use of transactions here

backend/src/helpers/user.ts
```diff
@@ -134,3 +156,169 @@ export const checkUserDevice = async ({
...
+  // delete organizations where user is only member
+  for await (const membershipOrg of membershipOrgs) {
+    const orgMemberCount = await MembershipOrg.countDocuments({
+      organization: membershipOrg.organization,
+    });
+    
+    const otherOrgAdminCount = await MembershipOrg.countDocuments({
+      organization: membershipOrg.organization,
+      user: { $ne: userId },
+      role: ADMIN
```
maidul98, line 188: note: what we want to know is if there is at least one user who has permission to change role. This can be `admin` or any custom role. I think for now this check is good enough.

backend/src/ee/models/gitAppInstallationSession.ts
```diff
@@ -29,6 +29,4 @@ const gitAppInstallationSession = new Schema<GitAppInstallationSession>({
 });
 
 
-const GitAppInstallationSession = model<GitAppInstallationSession>("git_app_installation_session", gitAppInstallationSession);
-
-export default GitAppInstallationSession;
+export const GitAppInstallationSession = model<GitAppInstallationSession>("git_app_installation_session", gitAppInstallationSession);
```
maidul98, line 32: nit: unrelated changes. i think it's good to try to keep changes as closely related as possible

backend/src/helpers/user.ts
```diff
@@ -134,3 +156,169 @@ export const checkUserDevice = async ({
...
+
+    await APIKeyData.deleteMany({
+      user: user._id
+    });
+
+    await Action.deleteMany({
+      user: user._id
+    }); 
+    
+    await Log.deleteMany({
```
maidul98, line 257: i don't think we should delete this because the logs can still be useful since it includes user emails in most cases

frontend/src/components/utilities/attemptCliLogin.ts
```diff
@@ -122,17 +120,7 @@ const attemptLogin = async (
                             iv,
                             tag,
                             privateKey
-                          });
-
-                        const userOrgs = await fetchOrganizations();
-                        const orgId = userOrgs[0]._id;
-                        localStorage.setItem("orgData.id", orgId);
```
maidul98, line 129: do we no longer need to set the org to localStorage?

frontend/src/components/utilities/attemptLogin.ts
```diff
@@ -119,20 +117,6 @@ const attemptLogin = async (
               tag,
               privateKey
             });
-            
-            // TODO: in the future - move this logic elsewhere
-            // because this function is about logging the user in
-            // and not initializing the login details
-            const userOrgs = await fetchOrganizations();
```
maidul98, line 126: why did we remove these?

maidul98:
looks good, left some small comments

backend/src/helpers/organization.ts
dangtony98: Sounds like we're good here then.

backend/src/helpers/user.ts
dangtony98: True but according to @akhilmhdh, an organization must have an `admin` I think; perhaps we enforce this rule somewhere at some point.

backend/src/helpers/user.ts
dangtony98: Ah the `Log` collection is actually deprecated. We use `AuditLog` now which I didn't touch.

frontend/src/components/utilities/attemptCliLogin.ts
dangtony98: In the context of the `attemptCliLogin` method specifically, this isn't needed because the user is directed to the "successfully logged in via CLI" page after which they return to their terminal; this was, however, necessary previously for `attemptLogin` method which is similar but for logging in via browser.

Here, we used to fetch user's organizations to navigate to one after logging in. I've replaced all such similar functionality occurrences with one helper function in `Login.utils.tsx`:

```
export const navigateUserToOrg = async (router: NextRouter) => {
    const userOrgs = await fetchOrganizations();

    if (userOrgs.length > 0) {
        // user is part of at least 1 org
        const userOrg = userOrgs[0] && userOrgs[0]._id;
        localStorage.setItem("orgData.id", userOrg);
        router.push(`/org/${userOrg}/overview`);
    } else {
        // user is not part of any org
        router.push("/org/none");
    }
}
```

dangtony98:
Will merge this in a few hours.

PR #1071, opened 2023-10-09
Improve-logging

ragnarbull:
# Description 📣

Add stack trace error logging in the console (with Winston) for prod - this will enable self-hosted users to debug better. Modified that all unexpected errors are sent to Sentry in prod.

## Type ✨

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Tested in a prod env with the change password function by throwing different errors and with different log levels. Checked custom error levels cast to Winston properly. Checked that the correct log level will be sent to Sentry.

---

### DIFF
--- a/backend/src/utils/requestError.ts
+++ b/backend/src/utils/requestError.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-extra-semi */
 import { Request } from "express"
 import { getVerboseErrorOutput } from "../config";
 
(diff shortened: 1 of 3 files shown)


backend/src/utils/requestError.ts
```diff
@@ -12,6 +12,27 @@ export enum LogLevel {
     EMERGENCY = 600,
 }
 
+export const mapToWinstonLogLevel = (customLogLevel: LogLevel): string => {
```
maidul98, line 15: removing as not needed for time being. each error should have it's own log level already

maidul98 (approved):

Issue #1046, opened 2023-10-02
App throws Application error


techemmy:
### Describe the bug
I invited a new user to an organization as an admin. In the same browser in a new tab, I received the invite I sent (I have access to the new user email). In the new user account on the members section, the **Add to projects** button shows beside the new user's email (which I think should show for only admins). 
I clicked the button which threw the Application error prompt. I refreshed the page hoping it was because I was just logged in as an admin before accepting the invite, but the button was still there.
I tried logging in to the Member account on a different browser, but t still shows the same button.

### To Reproduce
Steps to reproduce the behaviour:
1. Log in as an admin and invite a new member
2. Accept the member invite on the same browser
3. Go to the Members tab
4. You should see the **Add to projects** button there even if you refresh.

### Expected behaviour
I don't think the **Add to projects** button should show for just Members.

### Screenshots
<img width="1261" alt="Screenshot 2023-10-02 at 23 20 52" src="https://github.com/Infisical/infisical/assets/43725109/8d0785fa-8c6c-42dd-aff8-3186eb7c8539">
<img width="1268" alt="Screenshot 2023-10-02 at 23 23 25" src="https://github.com/Infisical/infisical/assets/43725109/26f79541-67a9-41ef-aa7e-53ffa5c3a81c">

### Platform you are having the issue on:
On app.infisical.com

Salman2301:
Able to replicate the issue.
The better fix is not letting the user to login. If user is already loggedIn

It happens in the `/sendInvite` page allowing a loggedIn user to register via an invite.

~~2. Accept the member invite on the same browser~~

It's actually a frontend issue.

techemmy:
Welcome @Salman2301 🙌

techemmy:
Well done @Salman2301 🙌

PR #1084, opened 2023-10-12
fix: signup only if the user not logged in #1046

Salman2301:
# Description 📣
As the issue #1046 describe. There was error due to multiple different user logged into the application. To avoid the bug that leads to that error. It's better not to let user to signup via invite. If the user is already logged. Used the same code snippet as signup page to redirect the user.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

1. Start the infisical local server
1. LogIn to the website
1. Goto `/signupinvite`
1. User should be redirect to the org page.

> For not logged in user, They can continue with the signup proccess

---

### DIFF
--- a/frontend/src/pages/signupinvite.tsx
+++ b/frontend/src/pages/signupinvite.tsx
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import crypto from "crypto";
 
-import { useState } from "react";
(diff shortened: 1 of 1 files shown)


Salman2301:
The issue still stays...
Need to hide the `Add to projects` for `Member` roles.
Clicking on the button leads to `/project/undefined/members`

Issue #1058, opened 2023-10-04
Serve next js with express js 


maidul98:
### Feature description
Currently, the standalone Infisical image contains nginx, pm2, frontend and backend running on two different servers. This creates a lot of overhead for a single container and has reduced performance for the end user.

To lower the overhead, we should have the existing backend also serve the next js app

### Additional context
Some possible options
- build/export frontend and serve it with as static content via express js 
- install `next` in express app and serve content directly from `frontend` folder
```
const nextApp = next({ dev: dev, dir: "../frontend"});
  const handle = nextApp.getRequestHandler();
```

maidul98:
cc: @Salman2301

Salman2301:
@maidul98 Will look into it... Just getting familiar with the code base atm. 😅

maidul98:
Sounds good @Salman2301

rtpa25:
Taking a look into it

PR #1089, opened 2023-10-15
feat: serve frontend from backend

G3root:
# Description 📣

fixes: #1058

- [x] New feature
- [x] Breaking change

# Tests 🛠️

```sh
```

---


maidul98:
I see you still have pm2 and nginx in the docker image. I think we should be able to remove nginx because all the traffic should go right to the backend now and after that we'll have one process so we can remove pm2 as well. What do you think?

maidul98:
> I see you still have pm2 and nginx in the docker image. I think we should be able to remove nginx because all the traffic should go right to the backend now and after that we'll have one process so we can remove pm2 as well. What do you think?

I have addressed these. New issue is fixing issue with getting data fetched via `getStaticProps`.

Screenshot 2023-10-16 at 12 50 27 AM

https://github.com/Infisical/infisical/blob/3f0b6dc6c1d9b51fbb4014f059919e3b5e546c51/frontend/src/pages/integrations/%5Bid%5D.tsx#L28

Second issue: something is up with rendering integrations page

G3root:
> > I see you still have pm2 and nginx in the docker image. I think we should be able to remove nginx because all the traffic should go right to the backend now and after that we'll have one process so we can remove pm2 as well. What do you think?
> 
> I have addressed these. New issue is fixing issue with getting data fetched via `getStaticProps`.
> Screenshot 2023-10-16 at 12 50 27 AM
> 
> https://github.com/Infisical/infisical/blob/3f0b6dc6c1d9b51fbb4014f059919e3b5e546c51/frontend/src/pages/integrations/%5Bid%5D.tsx#L28
> 
> Second issue: something is up with rendering integrations page

pushed a fix for that

.dockerignore
```diff
@@ -1,2 +1,11 @@
...
+frontend/node_modules
+backend/frontend-build
+**/node_modules
+**/.next
+.dockerignore
+.git
+README.md
+.dockerignore
+**/Dockerfile
+.next
```
Salman2301, line 11: We can remove this?

Salman2301: Already have `**/.next`

maidul98 (approved):
app

PR #1099, opened 2023-10-19
Update JWT secret scheme, replace many secrets with one secret

dangtony98:
# Description 📣

Previously, Infisical required you to input 5-6 different JWT secrets to get up and running. This PR replaces all 5-6 secrets with one new `AUTH_SECRET`; the primary motivation here is to reduce the usage curve and room for error in self-hosted deployments of Infisical whilst maintaining similar level of security.

The `AUTH_SECRET` is enough to verify that JWT tokens came from Infisical; differentiation between secret types is done via the `authTokenType` claim on the JWT payload which would not be alterable/forgeable without knowledge of the `AUTH_SECRET`.

We recommend setting `AUTH_SECRET` to a 256-bit randomly-generated value.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


backend/src/config/index.ts
```diff
@@ -17,17 +17,13 @@ export const getRootEncryptionKey = async () => {
 }
 export const getInviteOnlySignup = async () => (await client.getSecret("INVITE_ONLY_SIGNUP")).secretValue === "true"
 export const getSaltRounds = async () => parseInt((await client.getSecret("SALT_ROUNDS")).secretValue) || 10;
+export const getAuthSecret = async () => (await client.getSecret("AUTH_SECRET")).secretValue; // new
```
maidul98, line 20: hmm, this will mean that any existing deployments without this token will fail on next deploy. 

i think what we should do is if `JWT_AUTH_SECRET` is defined then use that as `AUTH_SECRET`. 

Option 2 is we don't have to rename `JWT_AUTH_SECRET` to `AUTH_SECRET`, this might be simpler for now and won't break existing deployments

maidul98:
otherwise eveything else looks good

backend/src/config/index.ts
dangtony98: I'll make `AUTH_SECRET` take from `JWT_AUTH_SECRET` if it is defined; I wanna remove `JWT` from the name of the variable.

dangtony98: Okay check again @maidul98

maidul98 (approved):
app

PR #1106, opened 2023-10-21
feat: respect domain override in login

### LINKED ISSUE #939: `infisical init` should respect `--domain` option and `INFISICAL_APP_URL`
### Feature description

When running `infisical init`, it prompts me for whether SaaS or Self-Hosted and then for the instance URL. It would be nice if it would respect the available options to set the infisical URL and not ask these questions.

### Why would it be useful?

Better UX.

### Additional context

Infisical cli v0.12.1

quinton11:
# Description 📣
Closes #939 
After the current logged in user is checked in the login Cmd,
- `config.INFISICAL_URL` is the current logged in user's domain url
- `config.INFISICAL_URL_MANUAL_OVERRIDE` is the value of the manual override set from `--domain` or the `INFISICAL_API_URL` env variable
- We check if the domain flag was set or the env variable exists, if yes, we respect the manual override
- If not we assume `config.INFISICAL_URL` is the default api url then prompt for a domain override ( necessary? )

### DIFF
--- a/cli/packages/cmd/login.go
+++ b/cli/packages/cmd/login.go
@@ -73,16 +73,23 @@ var loginCmd = &cobra.Command{
 				return
 			}
 		}
-		//override domain
+
+		promptOverrideDomain := true
 		domainQuery := true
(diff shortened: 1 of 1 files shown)


quinton11:
@maidul98

PR #1117, opened 2023-10-25
Secret rotation

akhilmhdh:
# Description 📣

1. Implemented basic secret rotation with providers sendgrid, postgres/cockroach db and mysql/mariadb

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/controllers/v3/secretsController.ts
```diff
@@ -467,20 +459,13 @@ export const deleteSecretByNameRaw = async (req: Request, res: Response) => {
 export const getSecrets = async (req: Request, res: Response) => {
   const validatedData = await validateRequest(reqValidator.GetSecretsV3, req);
   const {
-    query: { environment, workspaceId, include_imports: includeImports, folderId }
```
maidul98, line 470: I see that we are removing `folderId` in some of these routes, why is that?

backend/src/ee/secretRotation/service.ts
```diff
@@ -0,0 +1,130 @@
...
+  inputs,
+  outputs
+}: TCreateSecretRotation) => {
+  const rotationTemplate = rotationTemplates.find(({ name }) => name === provider);
+  if (!rotationTemplate) throw BadRequestError({ message: "Provider not found" });
+
+  const formattedInputs: Record<string, unknown> = {};
+  Object.entries(inputs).forEach(([key, value]) => {
+    const type = rotationTemplate.template.inputs.properties[key].type;
+    if (type === "string") {
```
maidul98, line 39: nit: i think a switch statement would be better here 

```
switch (type) {
    case "string":
      formattedInputs[key] = value;
      break;
    case "integer":
      formattedInputs[key] = parseInt(value, 10);
      break;
    default:
      formattedInputs[key] = JSON.parse(value);
  }
```

backend/src/ee/secretRotation/service.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+    formattedInputs[key] = JSON.parse(value as string);
+  });
+  // ensure input one follows the correct schema
+  const valid = ajv.validate(rotationTemplate.template.inputs, formattedInputs);
+  if (!valid) {
+    throw BadRequestError({ message: ajv.errors?.[0].message });
+  }
+
+  const encData: Partial<ISecretRotationEncData> = {
```
maidul98, line 55: nit: i think the more descriptive the better

backend/src/ee/secretRotation/service.ts
```diff
@@ -0,0 +1,130 @@
...
+    creds: []
+  };
+
+  const secretRotation = new SecretRotation({
+    workspace: workspaceId,
+    provider,
+    environment,
+    secretPath,
+    interval,
+    outputs: Object.entries(outputs).map(([key, secret]) => ({ key, secret }))
```
maidul98, line 66: is there a reason why the outputs are not encrypted at rest?

maidul98: Oh i see, the outputs are the mapping to the key name and secret id

backend/src/ee/secretRotation/templates/mysql.ts
```diff
@@ -0,0 +1,84 @@
...
+      "username2",
+      "port"
+    ],
+    additionalProperties: false
+  },
+  outputs: {
+    db_username: { type: "string" },
+    db_password: { type: "string" }
+  },
+  internal: {
```
maidul98, line 39: what are `internal` for and what does it keep track of?

backend/src/ee/secretRotation/queue.ts
```diff
@@ -0,0 +1,444 @@
...
+      default: {
+        throw Error(`Interpolation key not found - ${key}`);
+      }
+    }
+  }
+  const [type, keyName] = key.split(".").map((el) => el.trim());
+  return variables[type as keyof ISecretRotationData][keyName];
+};
+
+const secretRotationHttpFn = async (
```
maidul98, line 81: Let's move these helper functions to a separate file because otherwise this file is very bloated

backend/src/ee/secretRotation/queue.ts
```diff
@@ -0,0 +1,444 @@
...
+    });
+  }
+  return data;
+};
+
+const getInterpolationValue = (variables: ISecretRotationData) => (key: string) => {
+  if (key.includes("|")) {
+    const [keyword, ...arg] = key.split("|").map((el) => el.trim());
+    switch (keyword) {
+      case "random": {
```
maidul98, line 69: Let's not hard code this and make this into a constant so that when we have more of these custom functions, we know where find them all

backend/src/ee/secretRotation/templates/mysql.ts
```diff
@@ -0,0 +1,84 @@
...
+        "outputs.db_username": {
+          assign: TAssignOp.Direct as const,
+          value: "${internal.username}"
+        },
+        "outputs.db_password": {
+          assign: TAssignOp.Direct as const,
+          value: "${internal.rotated_password}"
+        }
+      },
+      pre: {
```
maidul98, line 65: it's not too clear to me what `pre` is for

backend/src/ee/secretRotation/queue.ts
```diff
@@ -0,0 +1,444 @@
...
+import { SecretVersion } from "../models";
+import { eventPushSecrets } from "../../events";
+
+const REGEX = /\${([^}]+)}/g;
+const SLUG_ALPHABETS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+const nanoId = customAlphabet(SLUG_ALPHABETS, 10);
+
+const secretRotationQueue = new Queue("secret-rotation-service", process.env.REDIS_URL as string);
+
+const interpolate = (data: any, getValue: (key: string) => unknown) => {
```
maidul98, line 40: I think this can be just one function, can you please simplify so that it calls interpolate on itself? I think this will increase readability for others when adding on to the code base

pseudo code:

```
function interpolate(data: any, variables: ISecretRotationData): any {
  if (!data || typeof data === "number") return data;

  if (typeof data === "string") {
    return data.replace(REGEX, (_, key) => {
      if (key.includes("|")) {
        const [keyword, arg] = key.split("|").map((s: string) => s.trim());
        switch (keyword) {
          case "random":
            return nanoId(parseInt(arg, 10));
          default:
            throw new Error(`Interpolation key not found - ${key}`);
        }
      }

      // Handle references like "inputs.admin_username"
      const [type, keyName] = key.split(".").map((s: string) => s.trim());
      return variables[type as keyof ISecretRotationData][keyName] as string;
    });
  }

  if (typeof data === "object" && Array.isArray(data)) {
    return data.map(item => interpolate(item, variables));
  }

  if (typeof data === "object") {
    ...
  }

  if (typeof data === "object") {
    if ('ref' in data && typeof data.ref === 'string') {
      ...
    } else {
      ...
    }
  }

  return data;
}
```

backend/src/ee/secretRotation/queue.ts
```diff
@@ -0,0 +1,444 @@
...
+            ? variables.inputs.username2
+            : variables.inputs.username1;
+      } else {
+        newCredential.internal.username = lastCred
+          ? lastCred.internal.username
+          : variables.inputs.username1;
+      }
+    }
+    if (provider.template.functions.set?.pre) {
+      secretRotationPreSetFn(provider.template.functions.set.pre, newCredential);
```
maidul98, line 265: can't we just pass in `provider.template.functions.set?.pre` without the if?

backend/src/ee/secretRotation/queue.ts
```diff
@@ -0,0 +1,444 @@
...
+    }
+    await secretRotationSetFn(provider.template.functions.set, newCredential);
+    await secretRotationTestFn(provider.template.functions.test, newCredential);
+
+    if (variables.creds.length === 2) {
```
maidul98, line 270: i'm worried that we are going to have a lot of these one off cases to account for different providers. It would be best IMO if each provider had it's own implementation that was called but let's see how this goes

backend/src/ee/secretRotation/queue.ts
```diff
@@ -0,0 +1,444 @@
...
+          algorithm,
+          keyEncoding,
+          skipMultilineEncoding
+        });
+      })
+    });
+
+    // akhilmhdh: @tony need to do something about this as its depend on authData which is not possibile in here
+    // await EEAuditLogService.createAuditLog(
+    //   {actor:ActorType.Machine},
```
maidul98, line 374: I think setting the actor to something like `Machine` & `infisical rotation` will be good enough for time being

backend/src/ee/secretRotation/service.ts
```diff
@@ -0,0 +1,130 @@
+import { ISecretRotationEncData, TCreateSecretRotation, TGetProviderTemplates } from "./types";
+import { rotationTemplates } from "./templates";
+import { SecretRotation } from "./models";
+import { client, getEncryptionKey, getRootEncryptionKey } from "../../config";
+import { BadRequestError } from "../../utils/errors";
+import Ajv from "ajv";
```
maidul98, line 6: Is there a reason we did not already use the Zod we have for validation? What was the limitation in Zod that made you want to use AJV?

backend/src/controllers/v3/secretsController.ts
akhilmhdh: Its not used anymore so removed it.

backend/src/ee/secretRotation/templates/mysql.ts
akhilmhdh: Internal are for keeping details that are not needed in outputs. But needed in other stages. 

Let's take sendgrid as an example. After creating the token. the token details return both token data and token id. 

End users want token but the system needs token id to call sendgrid `test` api and delete api

backend/src/ee/secretRotation/templates/mysql.ts
akhilmhdh: pre is like pre in package.json. Something that gets triggered before main function that is setter

backend/src/ee/secretRotation/queue.ts
akhilmhdh: it should not be variables but rather get value because going forward we may introduce more sets like env or some secrets to be pulled etc. 

Then getValue is idea because they don't need to know the original value is

backend/src/ee/secretRotation/service.ts
akhilmhdh: AJV is a JSON schema validator and zod is ts based. We need to save the validation schema as some format

backend/src/ee/secretRotation/queue.ts
akhilmhdh: Nope. It should always be 2 and not more. Its the rotated one. Always it should be deleted on third time

backend/src/ee/secretRotation/queue.ts
akhilmhdh: We can and do a undefined check

maidul98 (approved):

PR #1126, opened 2023-10-27
Multipart Update to Authentication (ST V3, Modularization of Auth Validation Methods, SSO logic)

dangtony98:
# Description 📣

This PR contains various updates to the authentication system for the Infisical API:

1. Service Token V3 (ST V3): ST V3 has been revised to employ a refresh token system similar to OAuth 2.0. 

- Instead of using one static secret to authenticate with Infisical, users get a refresh token that can be exchanged for short-lived access tokens to authenticate with the API.
- To mitigate secret zero, the refresh token can be made optionally rotating. If **Rotating Refresh Token** option is enabled for a refresh token, then when a request is made to get a new access token, a new refresh token is also issued (rendering all old access and refresh tokens defunct). If an old access/refresh token is used against the API, Infisical will know that an unintended actor may be trying to gain access to secrets and raise alarm.
- The (optionally rotating) refresh token mechanism has nice interplay with an upcoming agent initiative that will facilitate token authentication lifecycle procedures for more complex infrastructure use-cases.

2. Modularization of validation functions for each authentication mode: With many ways to authenticate with the API (i.e. JWT, Service Token V2/V3, API Key V1/V2), some of which will be deprecated in the near future, the code has been refactored to more cleanly detect and validate each inbound authentication mode.

3. Modularization of SSO logic: With a few SSO methods to authenticate with Infisical (i.e. Google, GitHub, GitLab, SAML), each sharing similar provider logic, the code has been refactored to reduce logic duplication.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Code improvements

---


backend/src/ee/controllers/v3/serviceTokenDataController.ts
```diff
@@ -56,6 +58,99 @@ export const getServiceTokenDataKey = async (req: Request, res: Response) => {
     });
 }
 
+/**
+ * Return access and refresh token as per refresh operation
+ * @param req 
+ * @param res 
+ */
+ export const refreshToken = async (req: Request, res: Response) => {
```
maidul98, line 66: pretty clean implementation

backend/src/helpers/auth.ts
```diff
@@ -1,381 +1,13 @@
...
-
-import { ActorType } from "../ee/models";
-import { getUserAgentType } from "../utils/posthog";
-
-/**
- * 
- * @param {Object} obj
- * @param {Object} obj.headers - HTTP request headers object
- */
-export const validateAuthMode = ({
```
maidul98, line 46: hmm, so what are we using instead

backend/src/routes/v3/secrets.ts
```diff
@@ -7,15 +7,15 @@ import { AuthMode } from "../../variables";
 router.get(
   "/raw",
   requireAuth({
-    acceptedAuthModes: [AuthMode.JWT, AuthMode.API_KEY, AuthMode.SERVICE_TOKEN, AuthMode.SERVICE_TOKEN_V3]
+    acceptedAuthModes: [AuthMode.JWT, AuthMode.API_KEY, AuthMode.API_KEY_V2, AuthMode.SERVICE_TOKEN, AuthMode.SERVICE_ACCESS_TOKEN]
```
maidul98, line 10: wow this is quite a lot

maidul98:
Nice refactor, it made sense to me. Here are some of my comments mostly nits

1) I think `refreshToken` field should be just called `token`. Users need to know if its refresh or not similar to how they don't need to know if JWT_SECRET

```{
  "publicKey": ",
  "privateKey": "",
  "refreshToken": "" <---- i think this should be called just token
}
```

2) I think this toggle should be under a accordion titled `Advanced`. This is so that the average user doesn't have to think about whether they should toggle this on or off since it'll be hidden from sight. Also a tool tip would help here because i'm not sure if this should be on or off or what its implications are 

Screenshot 2023-11-01 at 11 28 19 PM

3) Instead of only downloading the keys file, i think it would beneficial to allow users to copy the json to clipboard as a single line like this  `{ "publicKey": ", "privateKey": "", "refreshToken": "" }`. This is because i assume majority of cases will require users to pass this entire json string as: flag to the cli, as environment variable, as a string for k8s. Very rarely will users point to this file from clients. What do you think?

4) I think by default, the service token you create should be active. When i first tried using v3 service token, i thought something was wrong only to find out later it was inactive after creating 

5) IP white list feature still has that bug from a while ago, i think we should hide that feature until we uncover what the issue was?

6) It's not super clear what `@dev - /` for a first time user. Maybe path should be it's own column?
Screenshot 2023-11-01 at 11 48 43 PM

7) Nit: For the folder named `authMode` with all the auth validators, maybe `auth-mode-validators` might be a better folder name

Side note for future: i think this could have been split into two PRs, one for service token related and one for SSO

PR #1138, opened 2023-10-29
Fix bug with copying secret to clipboard with an override

### LINKED ISSUE #1137: [ENG-215] copy override value
When you have an override value and you press COPY, you get the regular value instead of the override:

image

<sub>[ENG-215](https://linear.app/infisical/issue/ENG-215/copy-override-value)</sub>

Shraeyas:
# Description 📣

On clicking copy secret if an override is enabled, it copies the original value instead of the override value.
Fixes #1137 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

---


maidul98 (approved):
nice catch

vmatsiiako:
I think the if statement should be the other way around, no?

PR #1141, opened 2023-10-30
feat: backward compatiable enc key in webhook

akhilmhdh:
# Description 📣

## Type ✨

- [x] Bug fix

# Tests 🛠️

```sh
```

---


maidul98:
I think we will have to use different encrypt functions like we do here 

https://github.com/Infisical/infisical/blob/main/backend/src/helpers/bot.ts#L55

maidul98 (approved):
looks good to me

Issue #1120, opened 2023-10-25
cannot remove projects


WladyX:
With the latest helm 0.4.1 and latests image 0.40.2, if i try to remove a project i see in the UI:
```
Failed to delete project
```

in the logs i get:
```
2023-10-25T15:25:55.686Z [IFSC:backend-main] error: MongoTransactionError: Cannot call commitTransaction after calling abortTransaction
    at endTransaction (/backend/node_modules/mongodb/lib/sessions.js:442:22)
    at node:internal/util:364:7
    at new Promise (<anonymous>)
    at endTransaction (node:internal/util:350:12)
    at ClientSession.commitTransaction (/backend/node_modules/mongodb/lib/sessions.js:270:16)
    at /backend/build/helpers/workspace.js:224:27
    at Generator.next (<anonymous>)
    at fulfilled (/backend/build/helpers/workspace.js:5:58)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
The server encountered an error while processing the request
```

imrefazekas:
I have the same issue. Project can not be deleted. Same error.

Shraeyas:
Can I work on this issue?

maidul98:
@WladyX This is related to our move to using transactions on MongoDB. We are actively looking into this and will provide more details here as we further uncover the issue

WladyX:
Thanks for the update, appreciate it.

Shraeyas:
@maidul98 I tried to setup mongo instance with a replica set locally and this issue does not pop up.

I wanted to understand what's the expectation here. I should be able to make a PR with updated docker files to setup mongo locally with replica sets if thats what we are looking for.

maidul98:
@Shraeyas Yes, if you can make a PR to add replica set for docker compose, that would be great

Shraeyas:
@maidul98 I had a few observations related to error handling while deleting projects and have created a PR (#1142 ) for that.

Will have a look into the mongodb replica set as well.

PR #1142, opened 2023-10-30
Fix bug with error handling while deleting project

Shraeyas:
# Description 📣

Related to #1120 

There were 2 issues:

1. If theres an error in the try block in /backend/src/helpers/workspace.ts, the commitTransaction() as being executed after abortTransaction() causing the "Cannot call commitTransaction after calling abortTransaction" error.
2. After fixing the above issue, now the error shown in the API response was generic and missing any message which has also been fixed. Refer to following ss

Screenshot from 2023-10-31 00-19-58

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

---

### DIFF
--- a/backend/src/helpers/workspace.ts
+++ b/backend/src/helpers/workspace.ts
@@ -272,6 +272,10 @@ export const deleteWorkspace = async ({
 		}, {
 			session
 		});
+
+		if (!existingSession) {
+			await session.commitTransaction();
+		}
 		
 		return workspace;
 	} catch (err) {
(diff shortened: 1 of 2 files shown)


maidul98:
Thanks for the PR @Shraeyas. Just so i understand, it looks like you said the issue with deleting project still exists after this PR?

Shraeyas:
@maidul98 Yes, the issue is still there. This one is only to fix the error handling part.

maidul98:
We have removed the use of transactions for time being, something better is soon to be released!

PR #1150, opened 2023-11-02
fix: resolved error in org settings and integrations page alert hidde…

akhilmhdh:
# Description 📣

1. Changed org user membership find from manual iteration to api call from permission set\
2. Removed integrations alert on empty integrations

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


frontend/src/views/Settings/OrgSettingsPage/components/OrgDeleteSection/OrgDeleteSection.tsx
```diff
@@ -1,81 +1,70 @@
...
 import { navigateUserToOrg } from "@app/views/Login/Login.utils";
 
 export const OrgDeleteSection = () => {
-    const router = useRouter();
-    const { currentOrg } = useOrganization();
-    const { user } = useUser();
-    const { createNotification } = useNotificationContext();
-    const { data: members } = useGetOrgUsers(currentOrg?._id ?? "");
-    
-    const membershipOrg = members?.find((member) => member.user._id === user._id);
```
maidul98, line 23: I understand that this wasn't great and the new approach is better, but it doesn't fully explain the issue we see in prod. What do you think hte issue is? Is it the time it takes to find the member and before it does it errors out on frontend?

maidul98 (approved):

PR #1152, opened 2023-11-03
fix: changed 2 fold operation of member workspace to one api call

### LINKED ISSUE #1149: [ENG-222] Unable to access one project, but can access another
### Describe the bug

We are a small team with two projects, and I have been granted admin access. I can sign in just fine, but I can only access one of them. The other project, when I try to view env variables for an environment, just gets "stuck" and continues to load (or sometimes I get a 400 resp).

Using the `infisical` CLI in a terminal it's the same thing. I can login just fine with `infisical login`, but I get this response when I attempt to use `infisical run` on the same project:

```
error: unable to get your encrypted workspace key. [err=CallGetEncryptedWorkspaceKey: Unsuccessful response [GET https://app.infisical.com/api/v2/workspace/........./encrypted-key] [status-code=500]]
Could not fetch secrets
If you are using a service token to fetch secrets, please ensure it is valid
If this issue continues, get support at https://infisical.com/slack
```

I'm the only one in my team who has this issue. Everyone can access the project just fine.

### Expected behavior

I should be able to access this one project like I can the other one.

macOS terminal and Chrome

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-222](https://linear.app/infisical/issue/ENG-222/unable-to-access-one-project-but-can-access-another)</sub>

akhilmhdh:
# Description 📣

1. Changed 2 api calls to singular call with support of batch in adding users to workspace membership
2. Minor ui adjustment in sidebar top headers
3. This also close #1149 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/validation/membership.ts
+++ b/backend/src/validation/membership.ts
@@ -78,3 +78,19 @@ export const DenyMembershipPermissionV1 = z.object({
     permissions: z.object({}).array()
   })
 });
+
+export const AddUserToWorkspaceV2 = z.object({
+  params: z.object({
+    workspaceId: z.string().trim()
+  }),
+  body: z.object({
+    members: z
+      .object({
+        orgMembershipId: z.string().trim(),
+        workspaceEncryptedKey: z.string().trim(),
+        workspaceEncryptedNonce: z.string().trim()
+      })
+      .array()
+      .min(1)
+  })
+});
--- a/backend/src/controllers/v2/index.ts
+++ b/backend/src/controllers/v2/index.ts
@@ -9,17 +9,19 @@ import * as secretsController from "./secretsController";
 import * as serviceAccountsController from "./serviceAccountsController";
 import * as environmentController from "./environmentController";
 import * as tagController from "./tagController";
+import * as membershipController from "./membershipController";
 
 export {
-    authController,
-    signupController,
-    usersController,
-    organizationsController,
-    workspaceController,
-    serviceTokenDataController,
-    secretController,
-    secretsController,
-    serviceAccountsController,
-    environmentController,
-    tagController,
-}
+  authController,
+  signupController,
+  usersController,
+  organizationsController,
+  workspaceController,
+  serviceTokenDataController,
+  secretController,
+  secretsController,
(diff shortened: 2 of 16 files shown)


maidul98:
The code makes sense to me. @dangtony98 since you created v1 it might be good if you also have a look before we merge

backend/src/validation/membership.ts
```diff
@@ -78,3 +78,19 @@ export const DenyMembershipPermissionV1 = z.object({
...
 });
+
+export const AddUserToWorkspaceV2 = z.object({
+  params: z.object({
+    workspaceId: z.string().trim()
+  }),
+  body: z.object({
+    members: z
+      .object({
+        orgMembershipId: z.string().trim(),
```
dangtony98, line 89: I feel that it's strange that you need to know the `orgMembershipId` of the user you want to add to an organization.

This would work for dashboard purposes but in terms of exposing a REST API for users to in the future make these calls themselves, it makes it unnecessarily complicated; from a users POV it would make more sense to pass in the `userId` instead when making this call and have the backend perform required checks from there.

akhilmhdh: Correct: This is to add to a workspace not organization

Have to disagree with that statement of complication rather i would say it made it easier for user. I'll tell you why.

Let's take the instance in which a user adds to a workspace. They will definitely need to fetch org membership of the user even to get user id. if ur fetching a org membership then u need to pass that. 

There is no api to get all user ids in an organization its all org membership filled with userids.

Also  by hierarchy of our design, org -> org membership -> workspace not in between user id which will end up lot more bugs in future

maidul98: Adding users to a workspace won't be easy either way because users will still need to do encryption to pass along the workspace key to invitees

dangtony98: @maidul98 True — We will have to re-think this in the near future.

maidul98 (approved):

PR #1168, opened 2023-11-08
Add environment field to trustedIp

ali-master:
# Description 📣

The `environment` field has been added as optional to `TrustedIp` and `AuditLog`. Both schemas now allow for an optional `environment` field and were updated in their related files. This change was needed to allow the system to track the environment from which a trusted IP originated. This will increase the security as IPs are now attached to a specific environment. The front-end code was also adjusted to reflect these changes.
Also, it fixes a problem with the `onIPAllowlistModalSubmit` handler, adding a missing field. the 'environment' field added to it. This will ensure that the correct environment data is sent and received. This change covers most of the parts of the application code that deal with IP addresses.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,5 @@
 # backend
 node_modules
-.env
-.env.dev
-.env.gamma
-.env.prod
-.env.infisical
 
 *~
 *.swp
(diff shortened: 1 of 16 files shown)


maidul98:
This PR is now stale. The backend structure has changed completely

PR #1171, opened 2023-11-10
New onboarding experience

akhilmhdh:
# Description 📣

Don't merge, requires a discussion on when its oauth based signup and how to migrate existing users and also infisical

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

<details>

<summary>
Screenshots
</summary>

Screenshot 2023-11-12 at 11 08 15 PM

Screenshot 2023-11-12 at 11 27 07 PM

Screenshot 2023-11-12 at 11 27 15 PM

</details>

# Tests 🛠️

```sh
```

---


backend/src/bootstrap.ts
```diff
@@ -0,0 +1,43 @@
...
+import { logger } from "./utils/logging";
+import mongoose from "mongoose";
+import { redisClient } from "./services/RedisService";
+
+type BootstrapOpt = {
+  transporter: nodemailer.Transporter;
+};
+
+export const bootstrap = async ({ transporter }: BootstrapOpt) => {
+  const spinner = ora().start();
```
maidul98, line 13: I don't think this spinner is necessary here. Please only add dependencies for must have items so we can reduce # of dependency.  I think we can just use our logger for this instead

maidul98:
Screenshot 2023-11-13 at 7 03 45 PM

I think text should be aligned here. Also when sign ups are disabled please also add blockers for SSO. After that, should be good to merge

maidul98:
I think we can remove `INVITE_ONLY_SIGNUP` from docs now that we have server configs?

maidul98:
Also you mentioned there are issues with current backup PDF logic, can you please patch those when you get a chance?

backend/src/bootstrap.ts
akhilmhdh: The issue with logger is for each message there is a stack trace too from where it gets fired etc. 

So putting these messages there maynot give users a good experience when seeing it. This would give a elegant experience and i wanna add more things to it later. Also ora is a tiny library too maintained by a great dev, used by many company. 

If i replace this with logger it would look a bit awful on booting up for customers

akhilmhdh:
> Also you mentioned there are issues with current backup PDF logic, can you please patch those when you get a chance?

Ill raise a PR after this gets merged don't wanna increase scope

backend/src/index.ts
```diff
@@ -271,7 +278,22 @@ const main = async () => {
   app.use(requestErrorHandler);
 
   const server = app.listen(port, async () => {
-    logger.info(`Server started listening at port ${port}`);
+    if (!serverCfg.initialized) {
+      logger.info(`Welcome to Infisical
+
+Create your Infisical administrator account at:
+http://localhost:${port}/admin/signup
```
maidul98, line 285: These ports won't be useful because the port that the server is listening on will likley be different than the one that is sending traffic to it

maidul98 (approved):
lgtm

PR #1178, opened 2023-11-16
fix: resolved backup key generation in signup, removed owner check and logic race condition error

akhilmhdh:
# Description 📣

1. CB timing issue with back up key generation in signup (now same as admin page synced one)
2. Removed owner check in custom role as its not used anymore
3. Login client init synced instead promise and then async

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/components/utilities/attemptLogin.ts
+++ b/frontend/src/components/utilities/attemptLogin.ts
@@ -8,9 +8,6 @@ import Telemetry from "./telemetry/Telemetry";
 import { saveTokenToLocalStorage } from "./saveTokenToLocalStorage";
 import SecurityClient from "./SecurityClient";
 
-// eslint-disable-next-line new-cap
-const client = new jsrp.client();
-
 interface IsLoginSuccessful {
   mfaEnabled: boolean;
   success: boolean;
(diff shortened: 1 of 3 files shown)


frontend/src/components/utilities/attemptLogin.ts
```diff
@@ -22,118 +19,100 @@ interface IsLoginSuccessful {
...
-          }
-        } catch (err) {
-          reject(err);
-        }
-      }
-    );
+  // eslint-disable-next-line new-cap
+  const client = new jsrp.client();
+  await new Promise((resolve) => {
+    client.init({ username: email, password }, () => resolve(null));
```
maidul98, line 35: From the exampels https://github.com/alax/jsrp it shows that the client instances need to be in the call back but you have nothing as the call back. How does this change fix the race condition error?

maidul98:
Rest looks good

frontend/src/components/utilities/attemptLogin.ts
akhilmhdh: There are two async ways in nodejs. Callbacks was something before promises was introduced. 

Here the callbacks is used, to run async code or the async client when its ready. Instead we use promise so that the client init is done and we wait until its done making the client instance ready to use

The issue which is my still hypothesis is the mix of callback and async try catch  inside a callback giving unexpected behavoirs.

maidul98 (approved):

Issue #859, opened 2023-08-14
Dockerconfigjson template


WladyX:
Infisical should have a docker config secret template so that we could deploy image pull secrets as code.
Vault example: https://blog.kelbert.fr/posts/acces-private-docker-registry-with-vault-secret-operator-s-help/
External secrets example: https://external-secrets.io/latest/guides/common-k8s-secret-types/#dockerconfigjson-example

neto210206:
It seems that you are suggesting that Infisical should provide a Docker configuration secret template to facilitate the deployment of image recovery secrets as code. This could improve the process of accessing private Docker records. You provided examples using Vault and External Secrets for Kubernetes. This is indeed a valuable feature that improves security and automation in Kubernetes deployments.

By offering a Docker configuration secret model, Infisical could simplify the integration of access to private Docker records in Kubernetes deployments. Users could easily manage their Docker credentials and authentication information, promoting consistency and reducing manual configurations.

This functionality is aligned with modern DevOps practices, where infrastructure as code (IaC) is critical to maintaining consistent, reliable and version-controlled deployments.

If you are suggesting this improvement to the Infisical team, you can provide your examples and explain how this functionality could benefit users in terms of security, automation and ease of use. In addition, you can share the examples provided from Vault and External Secrets to illustrate how this approach has been successfully implemented in other tools.

maidul98:
The operator now supports `kubernetes.io/dockerconfigjson` which you can set via the `secretType` here https://infisical.com/docs/integrations/platforms/kubernetes. Ensure you have a key in your Infisical project called `.dockerconfigjson:` set to the plain text docker json config

WladyX:
Great news, thank you @maidul98 and @Allex1 !

Allex1:
@maidul98 can you pls also upload the latest helm chart version 0.3.2 containing this feature into the public helm repo so we can sub chart it ?
Thanks

maidul98:
@Allex1 The new version has been released. Are you able to install it?

Allex1:
yes, all good. Thanks

PR #1182, opened 2023-11-18
Make secret type field configurable

Allex1:
# Description 📣

Hi, I know this is not yet complete as it needs docs, examples updated but wanted to get an opinion on this intial draft.
Addresses https://github.com/Infisical/infisical/issues/859

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Rebuilt docker image ,started it in k8s and used it with a resource :

```sh
---
apiVersion: secrets.infisical.com/v1alpha1
kind: InfisicalSecret
metadata:
  name: docker-release
spec:
  hostAPI: https://infisical.xxx.com/api
  resyncInterval: 60
  authentication:
    serviceToken:
      serviceTokenSecretReference:
        secretName: infisical-servicetoken
        secretNamespace: infisical-operator
      secretsScope:
        envSlug: common
        secretsPath: /imagePullSecret
  managedSecretReference:
    secretName: docker-release
    secretNamespace: infisical-test
    secretType: "kubernetes.io/dockerconfigjson"
```

---


maidul98:
Thank you for the PR @Allex1. We will aim to get this reviewed before end of this week

maidul98:
Hey @Allex1 I reviewed the PR and saw that you added a secret type field. I don't think this will suffice as each secret type has it's own format and fields that need to be filled out accordingly. 

What we need is some Go Lang secret templating system that can format the secret from Infisical into the desired Kubernetes secret type. External secrets manager does like this: https://external-secrets.io/v0.4.3/guides-common-k8s-secret-types/. I will dive into this to see what we can do from our end

Allex1:
@maidul98 I may have been a bit too optimistic in my description 😊
Like you said this pr only allows to define the secret type but in our case that is enough to be able to use the infisical operator to generate a secret of type .dockerconfigjson which we need to pull images from our private repo. 

If I change the description and crd to only make the type configurable instead of the current hardcoded `Opaque` will you be willing to allow this functionality merged/released until you will address all of the secret types/formats ? 
In the current setup we really need this and would rather not fork/maintain our own version until you make it available .
Also `managedSecretReference` seemed like the obvious place to define the type but let me know if you have other recommendations.
Thanks .

maidul98:
The Kube secret type of `kubernetes.io/dockercfg` has a different structure than `Opaque`. Are you planning on simply adding a key called `.dockercfg` in Infisical for time being? Is it that you tried to already set your auth details to that key but now only missing the type of the secret created?

Allex1:
I'm actually talking about `kubernetes.io/dockerconfigjson`  which has the same secret structure as `Opaque`  as long as the  encoded string is defined as `.dockerconfigjson` and the type is `kubernetes.io/dockerconfigjson`. So all is needed it the `type` field to be configurable and [not hardcoded](https://github.com/Infisical/infisical/pull/1182/files#diff-e032d6c97bdcf5bd7f3b390f751eb1b7ab38ac3402513e35336e69fbcb85c2a1L145) to `Opaque`
Built the operator from this pr, tested and it's working fine. 
I'll update the pr with the correct description and field options.

Allex1:
@maidul98 updated the pr to only allow for the `type` field to be configurable which should be a good step forward in supporting other secret types and also allow users to use `.dockerconfigjson` . Also kept this change backwards compatible by having the field optional and defaulting to `Opaque` if not provided, basically keeping the existing behavior. 
Let me know if there is something else that needs addressing .
Thanks

maidul98 (approved):

PR #1186, opened 2023-11-21
feat(audit-log): improvement in loading time

akhilmhdh:
# Description 📣

1. Changed pagination to load more for audit log to avoid total document count scan
2. Added index to workspace field in audit log to make look up faster

Screenshot 2023-11-21 at 2 31 39 PM

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/ee/controllers/v1/workspaceController.ts
```diff
@@ -567,6 +567,83 @@ export const rollbackWorkspaceSecretSnapshot = async (req: Request, res: Respons
...
+    }
+
+    #swagger.parameters['limit'] = {
+      "description": "Maximum number of versions to return",
+      "required": false,
+      "type": "string"
+    }
+
+   #swagger.parameters['startDate'] = {
+      "description": "Filter logs from the provided date",
```
maidul98, line 599: When writing API specs, let's assume that users don't know much. In this case, I would not know what format the date should be. Is it 11/2/23 or is in some other format. Can you pease go through all of the specs and make sure there isn't any ambiguity about what format a data type should be in?

dangtony98: This is a very good point actually; it’s the little things in making the API usable

maidul98:
Rest looks good to me

PR #1198, opened 2023-11-28
(feat): Secret reminders

DanielHougaard:
# Description 📣

This PR introduces a new concept called "secret reminders". It utilizes a mix of cron expressions and Redis to achieve this, therefore Redis will be required if the user chooses to self-host and wants to use secret reminders. The docs has also been updated to reflect this. The user simply chooses how often they'd like to be reminded, and if they want to associate a "note" to the reminder. The reminder will then be sent to every user that has access to the project where the secret lives in.

CleanShot 2023-11-28 at 11 48 38
CleanShot 2023-11-28 at 11 49 12

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


DanielHougaard:
> nit: Instead of saying `How many days between`, i think it might be more clear to say `Reminder interval`? What do you think?

Yea that could be an idea. The thought process behind it was that I wanted to make it as clear as possible that the interval is counted in days

frontend/package.json
```diff
@@ -55,6 +55,7 @@
     "axios": "^0.27.2",
     "axios-auth-refresh": "^3.3.6",
     "base64-loader": "^1.0.0",
+    "chrono": "^1.0.5",
```
akhilmhdh, line 58: Are we using this lib?

frontend/src/views/SecretMainPage/components/SecretListView/CreateReminderForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+        subTitle={
+          <div>
+            Set up a reminder for when this secret should be rotated. Everyone with access to this
+            project will be notified when the reminder is triggered.
+          </div>
+        }
+      >
+        <form onSubmit={handleSubmit(handleFormSubmit)}>
+          <div className="space-y-2">
+            <div>
```
akhilmhdh, line 64: U can wrap this whole(input and below label) in controller component of react hook form to avoid using setValue and gettt

frontend/src/views/SecretMainPage/components/SecretListView/SecretListView.utils.ts
```diff
@@ -20,6 +20,15 @@ export const formSchema = z.object({
   overrideAction: z.string().trim().optional(),
   comment: z.string().trim().optional(),
   skipMultilineEncoding: z.boolean().optional(),
+
```
akhilmhdh, line 23: Not sure how this is happening there seems to be some weird spacing between object keys etc.

akhilmhdh (changes requested):

frontend/src/views/SecretMainPage/components/SecretListView/SecretListView.utils.ts
DanielHougaard: Probably a linting rule, but I think 1 space is good, it allows for separation. More than 1 space is not ideal though, which is how it is right now I believe

akhilmhdh (approved):

Issue #906, opened 2023-08-28
Allow secret set commands to use service token


maidul98:
Hey, can anyone tell me why INFISICAL_TOKEN is not being used here?
```export INFISICAL_TOKEN=***
export INFISICAL_API_URL=***
ip=$(curl -s <https://ipinfo.io/ip>)
infisical secrets set HOST_NAME=bitnami HOST_IP=$ip --env=dev --path=/github --domain=$INFISICAL_API_URL```
I am told to login when I run the script
`You must be logged in to run this command. To login, run [infisical login]`

[Slack Message](https://infisical-users.slack.com/archives/C04BSBMQAQ7/p1693223283282639?thread_ts=1693223283.282639&cid=C04BSBMQAQ7)

ThomasPelletier:
+1

Maple-Shade:
Are there still plans to merge this in? This is a feature I would like to see added.

maidul98:
I believe this is now possible using machine identities. @DanielHougaard can provide more details

Maple-Shade:
Per the documentation there is not yet support for the CLI with machine identities.

https://infisical.com/docs/documentation/platform/identities/machine-identities#workflow

> I believe this is now possible using machine identities. @DanielHougaard can provide more details

DanielHougaard:
All commands that previously supported Service Tokens, now also support Machine Identities. There are some commands that do not support Machine Identity auth yet. Any command that takes the `token` flag, also supports Machine Identity auth now.

john-craig:
> All commands that previously supported Service Tokens, now also support Machine Identities. There are some commands that do not support Machine Identity auth yet. Any command that takes the `token` flag, also supports Machine Identity auth now.

The set command does not support the token flag until my pull request.

akhilmhdh:
Closing as stale. Usage of machine identity is the recommended pattern

PR #1206, opened 2023-11-29
Add support for authentication via service token for 'secret set' command

john-craig:
Add Support for Authentication via Service Token for 'secret set' Command

# Description 📣
Allows a service token to be specified by either the `--token` flag or the `INFISICAL_TOKEN` environment variable in conjunction with the `infisical secret set ...` command. It will use that token for authentication, rather than requiring a login.

I tried to mimic the existing style of implementation done for the `secret get` command. This is to say I moved the bulk of the code out of the actual function for the `secret set` command and into a dedicated utils function. I then split the processing for the JWT and ServiceToken into two separate functions which are called according to the case.

This closes issue #906.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
I did basic unit tests to verify this change, e.g. running the secret set command against a local Infisical server instance using the service token to authenticate.

```sh
infisical secret set --token KJfs348KDF myvariable=mysecret
```

---

### DIFF
--- a/cli/packages/cmd/secrets.go
+++ b/cli/packages/cmd/secrets.go
@@ -4,22 +4,17 @@ Copyright (c) 2023 Infisical Inc.
 package cmd
 
 import (
-	"crypto/sha256"
-	"encoding/base64"
 	"fmt"
 	"regexp"
 	"sort"
 	"strings"
-	"unicode"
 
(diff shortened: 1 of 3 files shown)


john-craig:
What is blocking this from being merged? It seems there's people in the issue interested in using it.

PR #1228, opened 2023-12-10
fix: "Injecting..." status string can be omitted by log levels

### LINKED ISSUE #1091: [ENG-192] `infisical run` should print its messages to `stderr`
### Describe the bug

The `infisical run` command prints messages to stdout.

### To Reproduce

```
$ infisical run -- some command >/dev/null
```

I would expect it to print

```
Injecting 4 Infisical secrets into your application process
failed to wait for command termination: exit status 1
```

to stderr, but it ends up in stdout.

### Expected behavior

`infisical run` prints to stderr instead.

### Additional context

Infisical CLI v0.14.2

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-192](https://linear.app/infisical/issue/ENG-192/infisical-run-prints-to-stdout-makes-it-difficult-to-use-the-output-of)</sub>

rawkode:
When using `infisical run`, I am often running another command that needs to be processed or consumed by another; such as:

infisical run -- supabase status -o env

The Injecting string was being printed directly to stdout and stopping such scripting from being successful, without further adding tail -n+2.

This change defaults the output to the INFO logging level, which means the behaviour is the exact same for everything; however those who wish can omit this output with -l error|fatal

Fixes #1091 

# Description 📣

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

image

---

### DIFF
--- a/cli/packages/cmd/run.go
+++ b/cli/packages/cmd/run.go
@@ -204,7 +204,8 @@ func init() {
 func executeSingleCommandWithEnvs(args []string, secretsCount int, env []string) error {
 	command := args[0]
 	argsForCommand := args[1:]
-	color.Green("Injecting %v Infisical secrets into your application process", secretsCount)
+
+	log.Info().Msgf(color.GreenString("Injecting %v Infisical secrets into your application process", secretsCount))
 
 	cmd := exec.Command(command, argsForCommand...)
 	cmd.Stdin = os.Stdin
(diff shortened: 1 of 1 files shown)


maidul98 (approved):
looks good to me

PR #1236, opened 2023-12-12
feat: brought back secret indexing popup in overview page

akhilmhdh:
# Description 📣

1. Brought back secret blind indexing popup in secret overview page for user attention

## Type ✨

- chore

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -22,16 +22,16 @@
         "@hookform/resolvers": "^2.9.10",
         "@octokit/rest": "^19.0.7",
         "@radix-ui/react-accordion": "^1.1.2",
-        "@radix-ui/react-alert-dialog": "^1.0.2",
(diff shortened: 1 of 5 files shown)


maidul98:
Wow, why so many additions

maidul98:
The reason it was commented out was because members would the pop up, but only admins can take action on it. Also because members can't fetch the status, it defaults the showing the section anyways. What we can do is make it so that this pop is only shown to admins or shown to both admins + members but disable the buttom for members. Let me know what you think

akhilmhdh:
Roger that. Will make that change and the other addition is the fix for recursion bug caused by radix ui dependency collision

akhilmhdh:
@maidul98 only admins will get the popup now

maidul98 (approved):
tested on my end too, works as expected

PR #1266, opened 2023-12-24
ui and docs improvements

vmatsiiako:
# Description 📣

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

### DIFF
--- a/docs/documentation/guides/nextjs-vercel.mdx
+++ b/docs/documentation/guides/nextjs-vercel.mdx
@@ -241,7 +241,7 @@ At this stage, you know how to use the Infisical-Vercel integration to sync prod
         Check out the [security guide](/security/overview).
     </Accordion>
     <Accordion title="Is there way to retain end-to-end encryption for syncing production secrets to Vercel?">
-        Yes. You can also use the Infisical [Node SDK](https://github.com/Infisical/infisical-node) to fetch secrets back to your Next.js app
+        Yes. You can also use the Infisical [Node SDK](https://infisical.com/docs/sdks/languages/node) to fetch secrets back to your Next.js app
         in both development and production.
 
         Depending on how you use it, however, it may require certain pages to be server-side rendered.  
--- a/docs/documentation/guides/nextjs-vercel.mdx
+++ b/docs/documentation/guides/nextjs-vercel.mdx
@@ -240,12 +240,6 @@ At this stage, you know how to use the Infisical-Vercel integration to sync prod
         
         Check out the [security guide](/security/overview).
     </Accordion>
-    <Accordion title="Is there way to retain end-to-end encryption for syncing production secrets to Vercel?">
-        Yes. You can also use the Infisical [Node SDK](https://infisical.com/docs/sdks/languages/node) to fetch secrets back to your Next.js app
-        in both development and production.
-
-        Depending on how you use it, however, it may require certain pages to be server-side rendered.  
-    </Accordion>
 </AccordionGroup>
 
 See also:
--- a/docs/documentation/guides/node.mdx
+++ b/docs/documentation/guides/node.mdx
@@ -5,7 +5,7 @@ title: "Node"
 This guide demonstrates how to use Infisical to manage secrets for your Node stack from local development to production. It uses:
 
 - Infisical (you can use [Infisical Cloud](https://app.infisical.com) or a [self-hosted instance of Infisical](https://infisical.com/docs/self-hosting/overview)) to store your secrets.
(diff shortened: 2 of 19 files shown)


docs/documentation/guides/nextjs-vercel.mdx
```diff
@@ -241,7 +241,7 @@ At this stage, you know how to use the Infisical-Vercel integration to sync prod
         Check out the [security guide](/security/overview).
     </Accordion>
     <Accordion title="Is there way to retain end-to-end encryption for syncing production secrets to Vercel?">
-        Yes. You can also use the Infisical [Node SDK](https://github.com/Infisical/infisical-node) to fetch secrets back to your Next.js app
+        Yes. You can also use the Infisical [Node SDK](https://infisical.com/docs/sdks/languages/node) to fetch secrets back to your Next.js app
```
maidul98, line 244: New SDKs do not retain end-to-end encryption. We are moving away from e2ee to make it easier for customers to interact with api

maidul98: Maybe you can point to old sdks that do but say that this will soon be deprecated

docs/documentation/guides/python.mdx
```diff
@@ -38,7 +38,7 @@ python3 -m venv env
 source env/bin/activate
 ```
 
-Install Flask and [infisical-python](https://github.com/Infisical/infisical-python), the client Python SDK for Infisical.
+Install Flask and [infisical-python](https://infisical.com/docs/sdks/languages/python), the client Python SDK for Infisical.
```
maidul98, line 41: New SDKs do not use service tokens so this guide doesn't make sense any more since it starts with `Create an Infisical Token`. I think you should leave all the sdk links as is because guides will need to be updated accordingly. Simple link updates will make things more confusing

docs/documentation/platform/identities/overview.mdx
```diff
@@ -4,8 +4,8 @@ description: "Programmatically interact with Infisical"
 ---
 
 <Note>
-  Currently, identities can only be used to make authenticated requests to the Infisical API and do not work with any clients such as [Node SDK](https://github.com/Infisical/infisical-node)
-  , [Python SDK](https://github.com/Infisical/infisical-python), CLI, K8s operator, Terraform Provider, etc.
+  Currently, identities can only be used to make authenticated requests to the Infisical API and do not work with any clients such as [Node SDK](https://infisical.com/docs/sdks/languages/node)
```
maidul98, line 7: looks good

docs/documentation/platform/secret-reference.mdx
```diff
@@ -12,8 +12,8 @@ This means that updating the value of a base secret propagates directly to other
     Currently, the secret referencing feature is only supported by the 
     [Infisical CLI](/cli/overview) and [native integrations](/integrations/overview).
 
-    We intend to add support for it to the [Node SDK](https://github.com/Infisical/infisical-node)
-    and [Python SDK](https://github.com/Infisical/infisical-python) this quarter.
+    We intend to add support for it to the [Node SDK](https://infisical.com/docs/sdks/languages/node), 
+    [Python SDK](https://infisical.com/docs/sdks/languages/python), and [Java SDK](https://infisical.com/docs/sdks/languages/java) this quarter.
```
maidul98, line 16: looks good

maidul98 (commented):
Frontend changes look good, i like the docs button. Some SDK link changes need additional changes to make sense in the given context

vmatsiiako:
good catch! updated!

docs/documentation/guides/node.mdx
```diff
@@ -5,7 +5,7 @@ title: "Node"
 This guide demonstrates how to use Infisical to manage secrets for your Node stack from local development to production. It uses:
 
 - Infisical (you can use [Infisical Cloud](https://app.infisical.com) or a [self-hosted instance of Infisical](https://infisical.com/docs/self-hosting/overview)) to store your secrets.
-- The [infisical-node](https://github.com/Infisical/infisical-node) client SDK to fetch secrets back to your Node application on demand.
+- The [infisical-node](https://infisical.com/docs/sdks/languages/node) client SDK to fetch secrets back to your Node application on demand.
```
maidul98, line 8: We'll need to update the rest of the guide vlad. The rest of the guide follows service tokens which are not compatible with new sdks. 

At @DanielHougaard can update this after break so that it is more consistent with new sdk steps. If you want to merge this in, i would revert the link back to old sdk link to reduce confusion. What do you think?

vmatsiiako:
sounds good, fixed!

maidul98 (approved):

PR #1267, opened 2023-12-24
Sdk/docs update 2

DanielHougaard:
# Description 📣

Updated docs to reflect the new SDK caching functionality, and attaching to process env.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

---

### DIFF
--- a/docs/sdks/languages/java.mdx
+++ b/docs/sdks/languages/java.mdx
@@ -103,7 +103,7 @@ public class App {
 
 ### Caching
 
-The SDK caches every secret and updates it periodically based on the provided cacheTTL. For example, if cacheTTL of 300 is provided, then a secret will be refetched every 5 minutes. The SDK first attempts to return the cached secret, but if no secret is found it'll make a request to get the fresh secret. Every time the secret is fetched from the cache, the cached secret expiry is reset.
+To reduce the number of API requests, the SDK temporarily stores secrets it retrieves. By default, a secret remains cached for 5 minutes after it's first fetched. Each time it's fetched again, this 5-minute timer resets. You can adjust this caching duration by setting the "cacheTTL" option when creating the client.
 
 ## Working with Secrets
 
--- a/docs/sdks/languages/java.mdx
+++ b/docs/sdks/languages/java.mdx
@@ -19,6 +19,7 @@ public class Example {
         ClientSettings settings = new ClientSettings();
         settings.setClientID("MACHINE_IDENTITY_CLIENT_ID");
         settings.setClientSecret("MACHINE_IDENTITY_CLIENT_SECRET");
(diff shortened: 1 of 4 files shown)


docs/sdks/languages/java.mdx
```diff
@@ -88,13 +89,22 @@ public class App {
...
         <ParamField query="setSiteURL()" type="string" default="https://app.infisical.com" optional>
             Your self-hosted absolute site URL including the protocol (e.g. `https://app.infisical.com`)
         </ParamField>
     </Expandable>
 
 </ParamField>
 
+### Caching
+
+The SDK caches every secret and updates it periodically based on the provided cacheTTL. For example, if cacheTTL of 300 is provided, then a secret will be refetched every 5 minutes. The SDK first attempts to return the cached secret, but if no secret is found it'll make a request to get the fresh secret. Every time the secret is fetched from the cache, the cached secret expiry is reset.
```
maidul98, line 106: Is there actually a periodic check? Or do you mean that a expiry is set on the cached secrets and when a request to fetch secrets is made and it exceeds a cached secret ttl, fresh secrets will be fetched?

docs/sdks/overview.mdx
```diff
@@ -31,3 +31,36 @@ From local development to production, Infisical SDKs provide the easiest way for
...
+       <Accordion title="Can I attach the environment variables to my process environment?">
+        Yes you can! The client SDK provides a method to attach the secrets to your process environment.
+        When using the `listSecrets()` method, you can pass a `attachToProcessEnv` parameter, which tells the SDK to attach all the found secrets to your process environment.
+        
+        Note: The exact parameter name may differ depending on the language.
+    </Accordion>
+    <Accordion title="What if a request for a secret fails?">
+        The SDK caches every secret and falls back to the cached value if a request fails. If no cached value is found, and the request fails, then the SDK throws an error.
+    </Accordion>
+    <Accordion title="What's the point if I still have to manage a token for the SDK?">
```
maidul98, line 54: The new authentication method significantly lowers the risk of token exposure. Here's how it works:

- Limited-use client secrets: You can create client IDs and secrets that can only be used a certain number of times (e.g., once) or expire after a short time (e.g., 20 seconds).
- In-memory access tokens: The access token generated from these limited-use secrets is stored securely in memory, not saved or transmitted elsewhere.
- Expired client secrets: After the token is created, the original client secret becomes invalid, preventing any further use.

Also client secrets will typically be delivered via trusted brokers method ensuring client id and client secret only comes together at just-in-time-use.

Let's remove this question from the FAQ since this has a more nuanced answer now compared to when we had just plain service tokens

maidul98 (approved):

PR #1279, opened 2024-01-05
Contribution guide refactor & SDK contribution guide

DanielHougaard:
# Description 📣

This PR introduces a new section to the contribution guide for contributing to the SDK. The guide is not just a setup guide, but also acts a bit like a knowledge base, because there are quite a few things that require an in-depth explanation for it to make sense.

Additionally I also restructured the existing contribution guide so we can have different guides for different parts of our growing codebase. 

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

---


docs/contributing/sdk/developing.mdx
```diff
@@ -0,0 +1,408 @@
...
+pub struct CreateSecretResponse {
+  pub secret: Secret, // "Secret" is defined elsewhere.
+}
+````
+
+### Adding input types to the schema generator
+
+You will _only_ have to define outputs in our schema generator, then QuickType will take care of the rest behind the scenes. You can find the Rust crate that takes care of type generation here: `crates/sdk-schemas/src/main.rs`.
+
+Simply add the output _(also called response)_, to the write_schema_for_response! macro. This will let QuickType know that it should generate types for the given structs. The main function will look something like this:
```
maidul98, line 138: Does  `write_schema_for_response! macro` contain a typo? If it is code we should put it in `code block`

DanielHougaard: Good catch!

maidul98 (approved):
lgtm

PR #1284, opened 2024-01-07
Enable new integration auth credential for each new integration

dangtony98:
# Description 📣

Previously, every new integration of the same category (e.g. Railway) in a project would use the same integration authentication credentials to create that integration. With this PR, whenever you create a new integration, the user is directed to input a new set of authentication credentials to be used for that integration instead of reusing a past one. Put differently, this allows users to use multiple authentication credentials for each integration category instead of being limited to one.

Highlights:
- When an integration is deleted, if there are no other integrations using the same linked integration authentication credential, then the linked authentication credential is also deleted; this is important for backwards compatibility since currently multiple integrations may be linked to one authentication credential.
- When an integration is created, a new integration authorization credential is also prompted to be inputted for it.
- When "revoking an integration authorization," all integrations and integration authorizations for that integration category are deleted.

This is a current solution that will be overhauled in Integration V2 whenever that is undertaken.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/controllers/v1/integrationAuthController.ts
+++ b/backend/src/controllers/v1/integrationAuthController.ts
@@ -2,7 +2,7 @@ import { Request, Response } from "express";
 import { Types } from "mongoose";
 import { standardRequest } from "../../config/request";
 import { getApps, getTeams, revokeAccess } from "../../integrations";
-import { Bot, IntegrationAuth, Workspace } from "../../models";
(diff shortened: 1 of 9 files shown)


maidul98 (approved):
Good thing we separated auth from the integration model

Issue #1288, opened 2024-01-08
[Secret reminders] Multiple mails sent when set two or more days


xevolic:
### Describe the bug

At first I set 7-day repetition to check if mails are sent every 7 days.
That worked good.

Then I set 3-day repetition to check if mails will be sent every 3 days.

Unfortunately I got mails:
- every 7 days
- and every 3 days.

I also noticed that **mails were sent not only to me**, but also to other users in our organization, which is strange.

### To Reproduce

Steps to reproduce the behavior: I think above description is enough and I don't need to describe it again.

### Expected behavior

Only current settings should be taken into consideration.

### Screenshots

![Screenshot 2024-01-08 at 13-39-36 Poczta — Tomasz Kuter — Outlook,2](https://github.com/Infisical/infisical/assets/111504995/8e97dc00-f7b0-497e-8e2e-260ec29f4c90)

### Platform you are having the issue on:

Pro version.

### Additional context

We deleted our account in previous week.

maidul98:
cc: @DanielHougaard

DanielHougaard:
Hi @xevolic, you're right! There's indeed an issue with how we handle reminder updates. I've just raised a PR to resolve this. Thank you for bringing this to our attention!

PR #1301, opened 2024-01-11
(Fix): Secret reminder duplication and deletion

DanielHougaard:
# Description 📣

- As mentioned in #1288 sometimes repeatable jobs would be duplicated instead of updated. When a user would attempt to update a reminder, we would try and find the existing reminder with the **new** data, which it would then fail to find. This resulted in the existing reminder not being updated, and a new reminder being created. This led to "duplicate" reminders for the same secret.
   - The fix for this was to create a new `TUpdateSecretReminder` type, which takes both the new _and_ the old repeat value, so we can probably update the old reminder.

- Additionally I also noticed that we weren't handling reminder deletion properly when a secret gets deleted. That was also fixed in this PR. 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/helpers/reminder.ts
+++ b/backend/src/helpers/reminder.ts
@@ -1,9 +1,9 @@
-import { ISecret } from "../models";
 import {
   createRecurringSecretReminder,
   deleteRecurringSecretReminder,
   updateRecurringSecretReminder
 } from "../queues/reminders/sendSecretReminders";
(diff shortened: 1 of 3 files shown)


DanielHougaard:
@akhilmhdh Closing this PR as it's fixed in the new postgres backend.

PR #1304, opened 2024-01-11
add project-config-dir flag for run command

maidul98:
# Description 📣

Adds project-config-dir fag to CLI run command

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
Manually tested run, secrets and export command. Also tested with secret refs. All works

### DIFF
--- a/cli/packages/util/config.go
+++ b/cli/packages/util/config.go
@@ -99,8 +99,6 @@ func GetWorkSpaceFromFile() (models.WorkspaceConfigFile, error) {
 		return models.WorkspaceConfigFile{}, err
 	}
 
-	fmt.Println("cfgFile", cfgFile)
-
 	configFileAsBytes, err := os.ReadFile(cfgFile)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


cli/packages/util/config.go
```diff
@@ -99,6 +99,8 @@ func GetWorkSpaceFromFile() (models.WorkspaceConfigFile, error) {
 		return models.WorkspaceConfigFile{}, err
 	}
 
+	fmt.Println("cfgFile", cfgFile)
```
DanielHougaard, line 102: Log statement

DanielHougaard (dismissed):
Looks great! Left a single comment, but other than that it looks good -- Good work!

cli/packages/util/config.go
maidul98: good catch

DanielHougaard (approved):

PR #1326, opened 2024-01-24
feat: cli export allow filtering with tags

### LINKED ISSUE #705: [ENG-145] tags filter not working while exporting via cli
Trying to export secrets based on tags does not work via the cli tool

### To Reproduce

Set some tags on secrets and pass the --tags option to the cli as descibed in the --help documentation

` -t, --tags string         filter secrets by tag slugs`

### Expected behavior

To filter by tags

Linux amd64

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-145](https://linear.app/infisical/issue/ENG-145/tags-filter-not-working-while-exporting-via-cli)</sub>

quinton11:
## Description 📣
Closes #705

### DIFF
--- a/cli/packages/cmd/export.go
+++ b/cli/packages/cmd/export.go
@@ -87,16 +87,12 @@ var exportCmd = &cobra.Command{
 
 		var output string
 		if shouldExpandSecrets {
-			substitutions := util.ExpandSecrets(secrets, infisicalToken, "")
-			output, err = formatEnvs(substitutions, format)
-			if err != nil {
-				util.HandleError(err)
-			}
-		} else {
-			output, err = formatEnvs(secrets, format)
-			if err != nil {
-				util.HandleError(err)
-			}
+			secrets = util.ExpandSecrets(secrets, infisicalToken, "")
+		}
+		secrets = util.FilterSecretsByTag(secrets, tagSlugs)
+		output, err = formatEnvs(secrets, format)
+		if err != nil {
+			util.HandleError(err)
 		}
 
 		fmt.Print(output)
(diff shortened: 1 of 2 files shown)


Salman2301:
I saw that you are removing the secrets based on the for loop. Nice patch but
Would be better if we pass the secrets filtering to the api

After some drill down of this code... Found this commented piece
https://github.com/Infisical/infisical/blob/8893aec213b58dca4595300c51bbf546c25c67a9/cli/packages/util/secrets.go#L164

Basically need to implement filter by tag in the API (v3)
https://github.com/Infisical/infisical/blob/8893aec213b58dca4595300c51bbf546c25c67a9/backend/src/helpers/secrets.ts#L516-L524

quinton11:
@Salman2301  based on the nature of the project I believe decryption of secrets is handled on the client side via the private key, by extension that means filtering via key words would happen there as well(same happens on the frontend).  Filtering at the api side will mean the backend having access to the user's private key. That is of course if you have a way of filtering encrypted secrets. I may be wrong and I stand corrected

Salman2301:
Sorry for the confusion, I mean filter using the tags.... Infisical doesn't store the raw secret value... but the tags associated with secret which is what the pull request solving are stored as string[] ( foreign key to the tag database). We can just query tags using the Orm if the user provided. I believe I could be wrong here too.

quinton11:
@Salman2301 Right, must've misunderstood, that would work I believe, will take a look

maidul98:
For now client side filetering works fine. @Salman2301 filtering at API level can be added in the future

maidul98 (approved):

Issue #243, opened 2023-01-21
Mono-repo with multi-services support + local setup and scope config file improvement (.infisical.json)


Grraahaam:
# Feature description 🚀

Today, AKAIK there's no easy way to run Infisical's CLI in a monorepo with multiple services divided into multiple infisical projects, e.g. `infisical run -- docker-compose up -d` ...

Imagine a monorepo with three services (pretty common) controlled by `docker-compose` : 
- `frontend`
- `backend`
- `database`

I have a Infisical project for each of the above services (best practice?) and I can't easily find a way without refactoring the code (e.g. install infisical CLI in a specific Dockerfile stage and pass a env token as an argument) to inject my secrets within when running `docker-compose`

## Suggestions ✨

**Writing documentation about Infisical monorepo/multi-service best practices, support and maybe adaptation of the API/CLI (I presume, to generate tokens bound to multiple projects)**

An option in both the dashboard & the CLI to generate single access token to multiple projects, and a single config in each of them (otherwise it won't make sense 🙃 )

### Token structure integration 🎫

For the token structure, the token `xxxx-xxxxxxxxxxxxxxx-xxxxxxx` points to three different Infisical projects (`<project-name>:<config>`) :
- `project-one:dev`
- `project-three:dev`
- `project-four:prd`

When requesting the secrets with this token, I'd like to get all the secrets of the above projects (from their respective configs), making it mono-repo friendly by allowing the user to easily inject multiple infisical projects secrets into a single app (made of multiple services)

⚠️ **There might be a conflict between same env names, in that case we could either warn the user and block the token creation until he fixes the name duplications or overwrite the value with the latest variable of the same name**

### CLI integration 💻

For the CLI, here's two suggestions :

**Interactive setup**

```bash
# Allow multiple project selection (e.g. spacebar)
$ infisical init
"Space to select, Enter to confirm"
> project-one
  project-two
  project-three
> project-four

# Press <ENTER>

# Select a single config per project
$ infisical init
"Space to select, Enter to confirm"

"Select config for : project-one"
> dev
  stg
  prd
"Select config for : project-four"
  dev
> stg
  prd
```

And the `.infisical.json` repository setup file could look like the following (check if we want to keep using JSON instead of YAML?) : 

```json
{
  // "workspaceId":"63c7c37331ffe61437214e61" we could use the project name instead easier for the user experience (project pre-config)
  // and format the project names at the creation time, in the dashboard (enforce alpha+hyphen maximum)
[...]

PR #1329, opened 2024-01-25
feat: allow directory to path mapping for better mono repositories

rawkode:
# Description 📣

This change adds a new parameter to `.infisical.json` that allows the user to turn on `directoryFromConfigMapping: true` to allow the `--path` CLI argument to be auto confingured with the relative path from where infisical was executed from the `.infisical.json` config.

The default value is `false` which makes this behaviour backwards compatible.

Related #243 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

I couldn't see any relevant tests to add the new  use-case to. So here's some manual testing:

https://asciinema.org/a/41jXyufl6fzYsKwnWd7ImTMYq

---

### DIFF
--- a/cli/packages/cmd/export.go
+++ b/cli/packages/cmd/export.go
@@ -74,6 +74,13 @@ var exportCmd = &cobra.Command{
 			util.HandleError(err, "Unable to parse flag")
 		}
 
+		if !cmd.Flags().Changed("path") {
+			workspaceMappedPath := util.GetPathFromWorkspaceFile()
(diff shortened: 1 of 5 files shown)


rawkode:
I'd really love some feedback on this, please 😄 

cc @maidul98 @vmatsiiako

rawkode:
Please let me know if we can ever revisit this, it is an important monorepository use-case.

@maidul98 @vmatsiiako

PR #1352, opened 2024-01-30
Add customer portal redirect and patch free trial endpoint

dangtony98:
# Description 📣

This PR includes two updates having to do with the license server / customer portal logic:
- It adds back the redirect endpoint + logic to direct customers to manage their billing/plan information.
- It fixes the free trial endpoint so customers can start free trials in Infisical Cloud

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -24,7 +24,8 @@ import {
   TOrgPlanDTO,
   TOrgPlansTableDTO,
   TOrgPmtMethodsDTO,
-  TStartOrgTrailDTO,
+  TStartOrgTrialDTO,
+  TCreateOrgPortalSession,
   TUpdateOrgBillingDetailsDTO
 } from "./license-types";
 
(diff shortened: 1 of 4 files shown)


backend/src/ee/services/license/license-service.ts
```diff
@@ -222,13 +223,67 @@ export const licenseServiceFactory = ({
...
+    const organization = await orgDAL.findOrgById(orgId);
+    if (!organization) {
+      throw new BadRequestError({
+        message: "Failed to find organization"
+      });
+    }
+
+    const {
+      data: { pmtMethods }
+    } = await licenseServerCloudApi.request.get(
```
akhilmhdh, line 253: One comment would be it would be great for all external request we set the type instead of any. Eslint will catch this going forward.

That is request.get<{pmtMethods: <type>}>()

akhilmhdh (dismissed):

PR #1354, opened 2024-01-30
(Feat): Re-add missing V2 workspaces endpoint

DanielHougaard:
# Description 📣

We missed an endpoint during the postgres migration. This PR re-adds the missing get organization workspace endpoint as it's documented [here](https://infisical.com/docs/api-reference/endpoints/organizations/workspaces).

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/org/org-service.ts
+++ b/backend/src/services/org/org-service.ts
@@ -3,6 +3,7 @@ import slugify from "@sindresorhus/slugify";
 import jwt from "jsonwebtoken";
 
 import { OrgMembershipRole, OrgMembershipStatus } from "@app/db/schemas";
+import { TProjects } from "@app/db/schemas/projects";
 import { TLicenseServiceFactory } from "@app/ee/services/license/license-service";
 import {
   OrgPermissionActions,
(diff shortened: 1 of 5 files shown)


akhilmhdh (approved):

backend/src/services/org/org-service.ts
```diff
@@ -93,6 +98,37 @@ export const orgServiceFactory = ({
     const members = await orgDAL.findAllOrgMembers(orgId);
     return members;
   };
+
+  const findAllWorkspaces = async ({ actor, actorId, orgId }: TFindAllWorkspacesDTO) => {
+    const { permission } = await permissionService.getOrgPermission(actor, actorId, orgId);
+    ForbiddenError.from(permission).throwUnlessCan(
```
akhilmhdh, line 104: Why all projects then filter. Directly filter can be applied right

akhilmhdh (changes requested):

backend/src/services/org/org-service.ts
akhilmhdh: Nvm. Will need to change this behavoir in frontend

akhilmhdh (approved):

PR #1385, opened 2024-02-09
feat: changed docker compose to use init container pattern for running migration

akhilmhdh:
This also is to decouple the doc change and docker compose change made in PR #1381 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


.env.example
```diff
@@ -3,16 +3,18 @@
 # THIS IS A SAMPLE ENCRYPTION KEY AND SHOULD NEVER BE USED FOR PRODUCTION
 ENCRYPTION_KEY=6c1fe4e407b8911c104518103505b218
 
+# Required
+DB_CONNECTION_URI=postgres://infisical:infisical@db:5432/infisical
```
maidul98, line 7: Should `@db` be `@localhost` so we can run the schema migrations?

docker-compose.dev.yml
```diff
@@ -55,81 +87,60 @@ services:
...
-    restart: always
-    depends_on:
-      - redis
-    environment:
-      - REDIS_HOSTS=local:redis:6379
-    ports:
-      - "8085:8081"
-    networks:
-      - infisical-dev
+  # mongo:
```
maidul98, line 113: Since we don't need Mongo, let's delete these

maidul98 (commented):
Just some small comments. Looks good otherwise. I didn't spin it up locally since i assume you did thorough testing

.env.example
akhilmhdh: We can't do that because this .env.example is being used for docker compose example. 

The docker compose doesn't expose db port via localhost

docker-compose.dev.yml
akhilmhdh: Oke

maidul98 (approved):
app

PR #1389, opened 2024-02-09
Add admin invite only signup field

Salman2301:
# Description 📣
Allow admin to enable/disable invite only signup, Change from a switch between allow sign up true/false to a dropdown options between Anyone | Invite Only | Disabled

- Add `inviteOnlySignUp` column in the `SuperAdmin` table
- Add migrations script to add extra field in the server config
- Check the screenshot for the UI (modified some padding too)

Missing something?. feel free to let me know

Admin Dashboard _ Infisical

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

1. Start the server
2. Goto Admin dashboard using the sidebar > image > Admin dashboard
3. UI

---


Salman2301:
@akhilmhdh @maidul98

maidul98:
It would be really nice if users can specify they want to only allow users from `@domain.com` email. Do you think this can be easly added to this PR?

Salman2301:
Yeah we can show an input field when the user select "Anyone" option and ask for **only Allowed domain** ( Leave it blank for any domain )
And we can check, if the user whose about to signup use the same email handle.
We can do it in the same PR.

I also want to hide and show the oAuth options like Github / Google...
I can either do that in this same PR or create a new one right after

capture-2024-01-27-024759_720

> Ignore the old design.... It's moved to Admin dashboard as per the feedback i got from slack

Salman2301:
@maidul98 In the new commit
- Add input field for the user to enter the domain, when the user select **Anyone**
- Check if the SSO(Github, Google) email or signup before the OTP has the same domain as in db
- Update migrate script to add column `allowSpecificDomainSignUp`
- Use forms and Save button instead of updating immediately

Admin Dashboard _ Infisical · 10 27am · 02-10

Admin Dashboard _ Infisical · 10 27am · 02-10 (1)

backend/src/db/migrations/20240205200732_invite-only.ts
```diff
@@ -0,0 +1,27 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const isTablePresent = await knex.schema.hasTable(TableName.SuperAdmin);
+  if (isTablePresent) {
+    await knex.schema.alterTable(TableName.SuperAdmin, (t) => {
+      t.boolean("inviteOnlySignUp").defaultTo(false);
+      t.string("allowSpecificDomainSignUp");
```
maidul98, line 10: Different departments or branches of a large organization might use distinct domains so not all will be the same. Can you add option for multiple?

backend/src/server/routes/v1/sso-router.ts
```diff
@@ -91,7 +91,8 @@ export const registerSsoRouter = async (server: FastifyZodProvider) => {
               lastName: "",
               authMethod: AuthMethod.GITHUB,
               callbackPort: req.query.state as string,
-              isSignupAllowed: Boolean(serverCfg.allowSignUp)
+              serverCfg
```
maidul98, line 94: Instead of passing in serverCfg, you can just call it inside of the service

backend/src/server/routes/v3/signup-router.ts
```diff
@@ -23,8 +23,16 @@ export const registerSignupRouter = async (server: FastifyZodProvider) => {
...
     handler: async (req) => {
-      await server.services.signup.beginEmailSignupProcess(req.body.email);
-      return { message: `Sent an email verification code to ${req.body.email}` };
+      const { email } = req.body;
+      const config = await server.services.superAdmin.initServerCfg();
+
+      if (config?.allowSpecificDomainSignUp) {
+        const domain = email.split("@")[1];
+
+        if (domain !== config.allowSpecificDomainSignUp) throw new Error(`Unsupported email domain (${domain}).`);
```
maidul98, line 32: Instead of generic error, throw BadRequest

backend/src/db/migrations/20240205200732_invite-only.ts
Salman2301: - Do want it as a comma separated or Array of string in the database?
- Same input element or Input tags?

Salman2301: Added it as a comma separated for now.

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
```diff
@@ -1,22 +1,43 @@
...
+  Tabs
+} from "@app/components/v2";
 import { useOrganization, useServerConfig, useUser } from "@app/context";
 import { useUpdateServerConfig } from "@app/hooks/api";
 
 enum TabSections {
   Settings = "settings"
 }
 
+type SignUpMode = "disabled" | "invite-only" | "anyone";
```
akhilmhdh, line 26: Use Enum, strings are error prone

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
```diff
@@ -28,37 +49,106 @@ export const AdminDashboardPage = () => {
     }
   }, [isNotAllowed, isUserLoading]);
 
+  useEffect(() => {
```
akhilmhdh, line 52: You can just remove this `useEffect`. Set allowspecificDomain is useless state as the config already contains the state.

Just use tertiary operator inside select below

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
```diff
@@ -28,37 +49,106 @@ export const AdminDashboardPage = () => {
...
+    }
+
+    if (config.allowSpecificDomainSignUp) {
+      setAllowSpecificDomain(config.allowSpecificDomainSignUp);
+    }
+  }, [config]);
+
+  const handleSubmit = async () => {
+    try {
+      config.allowSignUp = signUpMode !== "disabled";
```
akhilmhdh, line 70: Avoid mutating the state directly. Create a new copy of config and mutate that. 

The updateServerConfig automatically fetches the latest config on change

https://dev.to/il3ven/common-error-accidentally-mutating-state-in-react-4ndg

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
```diff
@@ -28,37 +49,106 @@ export const AdminDashboardPage = () => {
...
+            <TabPanel value={TabSections.Settings}>
+              <div className="mb-6 rounded-lg border border-mineshaft-600 bg-mineshaft-900 p-4">
+                <div className="flex justify-between">
+                  <div className="mb-4 text-xl font-semibold text-mineshaft-100">
+                    Allow user to Sign Up
+                  </div>
+                  <Select
+                    className="w-72 bg-mineshaft-700"
+                    dropdownContainerClassName="bg-mineshaft-700"
+                    onValueChange={(state) => setSignUpMode(state as SignUpMode)}
```
akhilmhdh, line 116: Use react hook form for all form management inside the frontend please

frontend/src/pages/signupinvite.tsx
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 import crypto from "crypto";
 
-import { useState } from "react";
+import { useEffect,useState } from "react";
```
akhilmhdh, line 5: Lint the frontend code base

backend/src/db/migrations/20240205200732_invite-only.ts
```diff
@@ -0,0 +1,27 @@
...
+export async function down(knex: Knex): Promise<void> {
+  if (await knex.schema.hasColumn(TableName.SuperAdmin, "inviteOnlySignUp")) {
+    await knex.schema.alterTable(TableName.SuperAdmin, (t) => {
+      t.dropColumn("inviteOnlySignUp");
+    });
+  }
+
+  if (await knex.schema.hasColumn(TableName.SuperAdmin, "allowSpecificDomainSignUp")) {
+    await knex.schema.alterTable(TableName.SuperAdmin, (t) => {
+      t.dropColumn("allowSpecificDomainSignUp");
```
akhilmhdh, line 24: `allowSpecificDomainSignUp` this feels as if this is boolean. I feel its bettter to name it as `allowedSignupDomain`.

specific also doesn't make sense.

akhilmhdh: CC: @maidul98

backend/src/server/routes/v3/signup-router.ts
```diff
@@ -23,8 +25,20 @@ export const registerSignupRouter = async (server: FastifyZodProvider) => {
       }
     },
     handler: async (req) => {
-      await server.services.signup.beginEmailSignupProcess(req.body.email);
-      return { message: `Sent an email verification code to ${req.body.email}` };
+      const { email } = req.body;
+      const serverCfg = await getServerCfg();
+
+      if (serverCfg?.allowSpecificDomainSignUp) {
```
akhilmhdh, line 31: I think we are missing allowSignupCheck here altogether here. Its missing in main codebase also. Need to fix that immediately from our side. 

CC: @maidul98

akhilmhdh (changes requested):

backend/src/db/migrations/20240205200732_invite-only.ts
maidul98: Agree, `allowedSignupDomain` makes more sense

backend/src/server/routes/v3/signup-router.ts
Salman2301: I can add it in this PR

backend/src/db/migrations/20240205200732_invite-only.ts
Salman2301: Done

frontend/src/pages/signupinvite.tsx
Salman2301: Done

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
Salman2301: eslint is not liking that 
Screenshot 2024-02-12 at 4 04 47 AM

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
Salman2301: Not using enum just moved to yup

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
Salman2301: Nester ternary was little clean IMO
```
  const signUpStatus = config.allowSignUp
  ? config.inviteOnlySignUp && "invite-only"
  : "disabled";

  const signUpType = signUpStatus || "anyone";
  
  ```

File: frontend/src/views/admin/DashboardPage/DashboardPage.tsx (hunk 6 above)
Comment (Salman2301) [reply frontend/src/views/admin/DashboardPage/DashboardPage.tsx]: fixed

File: frontend/src/views/admin/DashboardPage/DashboardPage.tsx (hunk 7 above)
Comment (Salman2301) [reply frontend/src/views/admin/DashboardPage/DashboardPage.tsx]: fixed

**Comment by Salman2301** [comment]
It shows this error when the domain is not valid during the signup, but I want to show something similar for the SSO (Google / Github). Since the page get redirected back and forth not sure what will be the best method. so, Open to suggestion, right now it's just redirect back to signup page after the SSO attempt. expected behaviour should show a similar toast like below.
Sign Up _ Infisical

**Comment by maidul98** [comment]
> It shows this error when the domain is not valid during the signup, but I want to show something similar for the SSO (Google / Github). Since the page get redirected back and forth not sure what will be the best method. so, Open to suggestion, right now it's just redirect back to signup page after the SSO attempt. expected behaviour should show a similar toast like below.

If users see error message like the one you have posted, I think it will be fine

**Comment by Salman2301** [comment]
The current login flow doesn't allow that. We might need to build a better error flow to make it work.
But i would like to keep this PR simple. Will split into a different PR for better error handling on SSO and Better Authentication methods.
I already have a plan for that.

**Comment by akhilmhdh** [comment]
Hey @Salman2301 tomorrow ill be going through your PR. Sorry for the delay. Was stuck on couple of my work

**Comment by Salman2301** [comment]
All good @akhilmhdh

File: backend/src/server/routes/v3/signup-router.ts [hunk 11]
```diff
@@ -153,6 +187,13 @@ export const registerSignupRouter = async (server: FastifyZodProvider) => {
       if (!userAgent) throw new Error("user agent header is required");
       const appCfg = getConfig();
 
+      const serverCfg = await getServerCfg();
+      if (!serverCfg.allowSignUp) {
```
akhilmhdh, line 191: As i was going through the code base, the database field `inviteOnlySignup` is not used anywhere in business logic. It's get updated in config but we are not using that field anywhere.

Then realized we don't need to have that field. Because when signup is disabled we can still allow  nvited signup. So the ui gets simplified to toggle version. 

Meaning we can just show like old one to allow or disallow signup and then in allowed mode provide the filter by domain like now

akhilmhdh: Or am i missing something in where you have to introduce this db field

Salman2301: Yeah,  good question

> signup is disabled we can still allow nvited signup

If the sign up disabled but allowing invite can be confusing for a large Organizations.
If the super admin choose to disabled signup but one of the project owner invited someone can be problematic.

That's the only reason it's there.

It makes to remove  if you want to let someone invite to org even, if the signup is disabled.

akhilmhdh:
This should be good to go with a rebase. i can't rebase this because after rebase i need to do a force push which I can't as i don't have the required permission. 

@Salman2301 kindly rebase the code. Its an easy rebase. 

CC: @maidul98

maidul98 (approved):
great work!

PR #1400, opened 2024-02-13
(Feat): Interacting with projects programmatically

DanielHougaard:
## Description 📣

This PR introduces a new concept called ghost users. Due to the nature of Infisical's end-to-end encryption structure, it was previously not possible to interact with projects programmatically through the API and other consuming applications. By introducing ghost user's, we're slowly moving to a more industry-standard way of storing secrets securely, while at the same time improving the user experience for everyone.
To sum it up. The main purpose of the PR is to make the API more user-friendly, and to introduce a better approach as to how we securely store secrets. Existing projects will get an option to upgrade to the new ghost user logic. All new projects will be created with the ghost logic by default.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


frontend/src/views/SecretApprovalPage/components/SecretApprovalRequest/components/SecretApprovalRequestChanges.tsx
```diff
@@ -9,6 +9,7 @@ import {
   faXmarkCircle
 } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { cx } from "cva";
```
akhilmhdh, line 12: use twmerge for this.

frontend/src/pages/org/[id]/overview/index.tsx
```diff
@@ -497,61 +493,39 @@ const OrganizationPage = withPermission(
...
+          await addUsersToProject.mutateAsync({
+            emails: orgUsers
+              .map((member) => member.user.email)
+              .filter((email) => email !== user.email),
+            projectId: newProjectId
+          });
         }
-        createNotification({ text: "Workspace created", type: "success" });
+
+        // eslint-disable-next-line no-promise-executor-return -- We do this because the function returns too fast, which sometimes causes an error when the user is redirected.
```
Comment (akhilmhdh) [inline frontend/src/pages/org/[id]/overview/index.tsx:523]: Didn't understand.

frontend/src/components/v2/UpgradeProjectAlert/UpgradeProjectAlert.tsx
```diff
@@ -0,0 +1,113 @@
...
+      projectId: project.id,
+      privateKey: PRIVATE_KEY
+    });
+
+    await getLatestProjectStatus();
+
+    setTimeout(() => setIsUpgrading(false), 5_000);
+  }, []);
+
+  useEffect(() => {
```
akhilmhdh, line 54: You can just interval feature in react-query. You don't need to manually do this. Based on the return api u can output it directly. You don't need to use another derived state

akhilmhdh (changes requested):
One more question. As we are polling the system. Are we blocking user interactions. 


Because what will happen if user's does something while the indexing is happening behind the scene. Blind index didn't had this issue because those secrets were never accessible by end users. But this case is not

DanielHougaard:
> One more question. As we are polling the system. Are we blocking user interactions.
> 
> Because what will happen if user's does something while the indexing is happening behind the scene. Blind index didn't had this issue because those secrets were never accessible by end users. But this case is not

This is an interesting case yea. I agree, we should to some degree block secret mutations while an upgrade is happening. I've just pushed a change that blocks secret write operations if an upgrade is active. You can check it out [here](https://github.com/Infisical/infisical/pull/1400/commits/1a799e31ca57b3fb1cbb56461eb81eb00eb842c6).

backend/src/server/routes/v2/index.ts
```diff
@@ -21,6 +22,7 @@ export const registerV2Routes = async (server: FastifyZodProvider) => {
     async (projectServer) => {
       await projectServer.register(registerProjectRouter);
       await projectServer.register(registerIdentityProjectRouter);
+      await projectServer.register(registerProjectMembershipRouter);
```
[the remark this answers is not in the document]
DanielHougaard: Last I spoke to Tony, I was told that we'd be doing a "larger" move at a later time, where we'll properly start renaming things to the `project` namespace. The reason for going with the `v2` router, is because the `v1` router would have conflicting endpoints if we use the `workspace` namespace. I think if we're planning to do a larger renaming later, we should keep this on the `v2` router for this.

backend/src/services/org/org-service.ts
```diff
@@ -118,6 +132,54 @@ export const orgServiceFactory = ({
     return workspaces.filter((workspace) => organizationWorkspaceIds.has(workspace.id));
   };
 
+  const addGhostUser = async (orgId: string, tx?: Knex) => {
+    const email = `ghost@${nanoid(16)}-${orgId}.com`; // We add a nanoid because the email is unique. And we have to create a new ghost user each time, so we can have access to the private key.
```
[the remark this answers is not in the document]
DanielHougaard: > Use the alphaNum nanoId in lib. The above will generate symbols which would practically make the email invalid.

The only type of special characters that `nanoid` uses, is `_` and `-`, which is fine.

> This would ensure if there is any point of email verification being done its still a valid email. Another thing use @infisical because that would reduce scope of a domain attack.

Email verification won't be done, as it's not possible to log in as the ghost user at any point. However I've added `@infisical`.

backend/src/services/project-membership/project-membership-service.ts
```diff
@@ -55,64 +71,75 @@ export const projectMembershipServiceFactory = ({
...
     ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionActions.Create, ProjectPermissionSub.Member);
 
-    const invitee = await userDAL.findOne({ email });
-    if (!invitee || !invitee.isAccepted)
-      throw new BadRequestError({
-        message: "Faield to validate invitee",
-        name: "Invite user to project"
+    const invitees: TUsers[] = [];
+
+    for (const email of emails) {
```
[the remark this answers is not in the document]
DanielHougaard: The function was like this before my changes as well, and in fact this endpoint isn't actually used anywhere. We have a query for it in the frontend, but that query isn't actually used anywhere at all. However, I've updated the function now to address the mentioned issues.

backend/src/services/project/project-dal.ts
```diff
@@ -52,6 +54,32 @@ export const projectDALFactory = (db: TDbClient) => {
...
+        .select(selectAllTableCols(TableName.Users))
+        .where({ ghost: true })
+        .first();
+      return ghostUser;
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Find project top-level user" });
+    }
+  };
+
+  const setProjectUpgradeStatus = async (projectId: string, status: ProjectUpgradeStatus | null, tx?: Knex) => {
```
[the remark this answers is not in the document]
DanielHougaard: Yeah, but I'd like an explicit function for this for two reasons.

1. A custom error message that reflects exactly which DAL operation failed
2. For readability and reusability. 

I would prefer to keep it this way.

backend/src/ee/services/secret-approval-request/secret-approval-request-secret-dal.ts
```diff
@@ -11,6 +17,27 @@ export const secretApprovalRequestSecretDALFactory = (db: TDbClient) => {
   const secretApprovalRequestSecretOrm = ormify(db, TableName.SecretApprovalRequestSecret);
   const secretApprovalRequestSecretTagOrm = ormify(db, TableName.SecretApprovalRequestSecretTag);
 
+  const bulkUpdateNoVersionIncrement = async (
+    data: Array<{ filter: Partial<TSecretApprovalRequestsSecrets>; data: TSecretApprovalRequestsSecretsUpdate }>,
+    tx?: Knex
+  ) => {
+    try {
+      const secs = await Promise.all(
+        data.map(async ({ filter, data: updateData }) => {
```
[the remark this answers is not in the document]
DanielHougaard: We are actually only getting 100 versions per folder, the rest are discarded.

If I was going to use the `insert` method you mentioned, I'd need a full object instead of just a partial update object.

Do you think this will still be an issue, considering we're only updating 100 versions per folder?

backend/src/services/secret/secret-dal.ts
```diff
@@ -41,6 +45,24 @@ export const secretDALFactory = (db: TDbClient) => {
     }
   };
 
+  const bulkUpdateNoVersionIncrement = async (
+    data: Array<{ filter: Partial<TSecrets>; data: TSecretsUpdate }>,
+    tx?: Knex
+  ) => {
+    try {
+      const secs = await Promise.all(
```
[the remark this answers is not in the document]
DanielHougaard: Very good catch!

backend/src/services/secret/secret-version-dal.ts
```diff
@@ -36,6 +36,46 @@ export const secretVersionDALFactory = (db: TDbClient) => {
...
+    data: Array<{ filter: Partial<TSecretVersions>; data: TSecretVersionsUpdate }>,
+    tx?: Knex
+  ) => {
+    try {
+      const secs = await Promise.all(
+        data.map(async ({ filter, data: updateData }) => {
+          const [doc] = await (tx || db)(TableName.SecretVersion)
+            .where(filter)
+            .update(updateData)
+            // .increment("version", 1) // TODO: Is this really needed?
```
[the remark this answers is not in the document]
DanielHougaard: Very good catch!

frontend/src/pages/org/[id]/overview/index.tsx
Comment (DanielHougaard) [reply frontend/src/pages/org/[id]/overview/index.tsx:523]: Started a Slack conversation about this!

backend/src/services/org/org-service.ts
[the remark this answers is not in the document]
akhilmhdh: I am not talking about the login based email verification. Generally speaking if we run some migration and there existed an email validation checks its better the email above to valid ones rather than invalid emails

backend/src/ee/services/secret-approval-request/secret-approval-request-secret-dal.ts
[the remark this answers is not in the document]
akhilmhdh: 100 is still in my opinion could be a bit trouble someone. If it was insert i mean 1000 wouldn't have matter. Where as 100 its individual query. N query this is not one. Where as the other one is 1 single query

backend/src/services/project-membership/project-membership-service.ts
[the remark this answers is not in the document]
akhilmhdh: `The function was like this before my changes as well`. Yes it's same as before but there is a one major change.

`for (const email of emails)`. This one is a new addition. This changed the whole process. The above is used by our frontend to add users to workspace on creation. If you don't need this change then kindly roll back the changes made in this file. Else kindly make the above changes, if the for loop exist

backend/src/server/routes/v1/project-router.ts
```diff
@@ -268,14 +268,16 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
+      const { invitees, latestKey } = await server.services.projectMembership.inviteUserToProject({
         actorId: req.permission.id,
         actor: req.permission.type,
         actorOrgId: req.permission.orgId,
         projectId: req.params.workspaceId,
-        email: req.body.email
+        emails: [req.body.email]
       });
 
+      const invitee = invitees[0];
```
maidul98, line 279: Are you refering to invitee or inviter? Also is this relying on the frontend to always add the first user as the one who invites everyone?

backend/src/server/routes/v2/project-membership-router.ts
```diff
@@ -0,0 +1,95 @@
...
+import { ProjectMembershipsSchema } from "@app/db/schemas";
+import { EventType } from "@app/ee/services/audit-log/audit-log-types";
+import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
+import { AuthMode } from "@app/services/auth/auth-type";
+
+export const registerProjectMembershipRouter = async (server: FastifyZodProvider) => {
+  server.route({
+    method: "POST",
+    url: "/:projectId/memberships",
+    schema: {
```
maidul98, line 12: can you please add .describe on the zod parameters to briefly explain what the fields are? this will be helpful for api doc generation

backend/src/server/routes/v2/project-router.ts
```diff
@@ -17,7 +26,7 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
         }
       ],
       params: z.object({
-        workspaceId: z.string().trim()
+        projectId: z.string().trim()
```
maidul98, line 29: Looks like a breaking change?

backend/src/server/routes/v2/project-router.ts
```diff
@@ -52,4 +61,88 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
+  /* Create new project */
+  server.route({
+    method: "POST",
+    url: "/",
+    config: {
+      rateLimit: authRateLimit
+    },
+    schema: {
+      body: z.object({
+        projectName: z.string().trim(),
```
maidul98, line 127: what about optional project slug?

maidul98: Also this API looks very very similar to the other project create. I can't seem to tell the difference other than naming parameters

backend/src/services/auth/auth-login-service.ts
```diff
@@ -275,7 +275,7 @@ export const authLoginServiceFactory = ({ userDAL, tokenService, smtpService }:
     if (isOauthSignUpDisabled) throw new BadRequestError({ message: "User signup disabled", name: "Oauth 2 login" });
 
     if (!user) {
-      user = await userDAL.create({ email, firstName, lastName, authMethods: [authMethod] });
+      user = await userDAL.create({ email, firstName, lastName, authMethods: [authMethod], isGhost: false });
```
maidul98, line 278: nit: this is true by default `isGhost: false`

backend/src/services/auth/auth-signup-service.ts
```diff
@@ -50,7 +50,7 @@ export const authSignupServiceFactory = ({
       throw new Error("Failed to send verification code for complete account");
     }
     if (!user) {
-      user = await userDAL.create({ authMethods: [AuthMethod.EMAIL], email });
+      user = await userDAL.create({ authMethods: [AuthMethod.EMAIL], email, isGhost: false });
```
maidul98, line 53: nit: isGhost: false is true by default

backend/src/services/org/org-dal.ts
```diff
@@ -76,7 +76,41 @@ export const orgDALFactory = (db: TDbClient) => {
...
+      return members.map(({ email, firstName, lastName, userId, publicKey, ...data }) => ({
+        ...data,
+        user: { email, firstName, lastName, id: userId, publicKey }
+      }));
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Find all org members" });
+    }
+  };
+
+  const findOrgMembersByEmail = async (orgId: string, emails: string[]) => {
```
maidul98, line 90: filter out ghost user from this result too?

backend/src/services/org/org-service.ts
```diff
@@ -118,6 +131,54 @@ export const orgServiceFactory = ({
     return workspaces.filter((workspace) => organizationWorkspaceIds.has(workspace.id));
   };
 
+  const addGhostUser = async (orgId: string, tx?: Knex) => {
+    const email = `ghost-${alphaNumericNanoId(16)}-${orgId}@infisical.com`; // We add a nanoid because the email is unique. And we have to create a new ghost user each time, so we can have access to the private key.
```
maidul98, line 135: what do you think about calling it `sudo-${alphaNumericNanoId(16)}-${orgId}@infisical.com`. It might sound a bit less scary and when we discuss this in the context of debugging with customers. Instead of saying `ghost` user we can refer to it as `sudo` user. This is your call though, just a suggestion

backend/src/services/project/project-queue.ts
```diff
@@ -0,0 +1,462 @@
...
+            receiverId: ghostUser.user.id,
+            encryptedKey: newEncryptedProjectKey,
+            nonce: newEncryptedProjectKeyIv,
+            senderId: ghostUser.user.id
+          },
+          tx
+        );
+
+        // Create a membership for the ghost user
+        await projectMembershipDAL.create(
```
maidul98, line 212: I know there is a limit in terms of how many users someone can have on free tier, does this account for it?

backend/src/services/project/project-fns.ts
```diff
@@ -0,0 +1,50 @@
+import crypto from "crypto";
+
+import { decryptAsymmetric, encryptAsymmetric } from "@app/lib/crypto";
+
+import { AddUserToWsDTO } from "./project-types";
+
+export const createWsMembers = ({ members, decryptKey, userPrivateKey }: AddUserToWsDTO) => {
+  const key = decryptAsymmetric({
```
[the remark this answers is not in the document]
maidul98: nit: `assignWorkspaceKeysToMembers` might better reflect what is going on in the function

backend/src/services/project-bot/project-bot-service.ts
```diff
@@ -1,132 +1,128 @@
...
     ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionActions.Edit, ProjectPermissionSub.Integrations);
 
+    const project = await projectBotDAL.findProjectByBotId(botId);
+
+    if (!project) {
+      throw new BadRequestError({ message: "Failed to find project by bot ID" });
+    }
+
+    if (project.version === ProjectVersion.V2) {
+      throw new BadRequestError({ message: "Failed to set bot active, project has a default bot enabled" });
```
maidul98, line 123: Nit: maybe we can just say Bot is already active

backend/src/services/secret/secret-service.ts
```diff
@@ -776,6 +819,14 @@ export const secretServiceFactory = ({
       subject(ProjectPermissionSub.Secrets, { environment, secretPath: path })
     );
 
+    const isProjectBeingUpgraded = await projectDAL.isProjectBeingUpgraded(projectId);
+
+    if (isProjectBeingUpgraded) {
+      throw new BadRequestError({
+        message: "Project is currently being upgraded, and secrets cannot be written. Please try again"
+      });
+    }
```
maidul98, line 828: nit: looks like this check is being repeated a lot, maybe you can put it in a function

maidul98: also hmm i tried to edit secrets while migration was happening and i didn't get any errors

backend/src/services/smtp/templates/workspaceInvitation.handlebars
```diff
@@ -1,15 +1,15 @@
...
+  </head>
+  <body>
     <h2>Join your team on Infisical</h2>
-    <p>{{inviterFirstName}} ({{inviterEmail}}) has invited you to their Infisical project — {{workspaceName}}</p>
+    <p>You have been invited to a new Infisical project — {{workspaceName}}</p>
     <a href="{{callback_url}}">Join now</a>
     <h3>What is Infisical?</h3>
-    <p>Infisical is an easy-to-use end-to-end encrypted tool that enables developers to sync and manage their secrets and configs.</p>
-</body>
+    <p>Infisical is an easy-to-use end-to-end encrypted tool that enables developers to sync and manage their secrets
```
maidul98, line 12: rephrase e2ee

maidul98 (commented):
Looks real good! Just left some minor nits. Should be good to go after that

backend/src/ee/services/saml-config/saml-config-service.ts
```diff
@@ -338,7 +338,8 @@ export const samlConfigServiceFactory = ({
             email,
             firstName,
             lastName,
-            authMethods: [AuthMethod.EMAIL]
+            authMethods: [AuthMethod.EMAIL],
+            isGhost: false
```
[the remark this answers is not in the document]
DanielHougaard: Not setting the field will lead to type errors. I suspect this is because I set the field to `notNullable`. Could this be an issue with the type system?

cc @akhilmhdh

backend/src/lib/crypto/srp.ts
```diff
@@ -24,3 +32,97 @@ export const srpCheckClientProof = async (
   server.setClientPublicKey(clientPublicKey);
   return server.checkClientProof(clientProof);
 };
+
+// FOR GHOST USER STUFF
```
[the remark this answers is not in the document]
DanielHougaard: Whoops! This is a leftover, good catch

backend/src/services/org/org-service.ts
DanielHougaard: I agree!

backend/src/services/project-membership/project-membership-service.ts
```diff
@@ -223,11 +395,65 @@ export const projectMembershipServiceFactory = ({
     return membership;
   };
 
+  const deleteProjectMemberships = async ({
```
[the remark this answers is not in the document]
DanielHougaard:
```
const member = await userDAL.findUserByProjectMembershipId(membershipId);

    if (member?.isGhost) {
      throw new BadRequestError({
        message: "Unauthorized member delete",
        name: "Delete project membership"
      });
    }
```

Line 386-393

maidul98 (approved):

PR #1449, opened 2024-02-23
(Feat): V2 slugified project routes

DanielHougaard:
# Description 📣

This PR introduces new routes on the V2 api `(/api/v2/workspace)`, that use project slugs instead of project ID's. This is a prerequisite for adding project support to our Terraform provider. 

I've introduced a new project DAL operation called `getProjectByFilter`, which will make it easier to query for projects based on both slugs and ID's.

Additionally, this PR also fixes the v1 update project endpoint. The endpoint is currently only accessible with the JWT auth method. It's now also accessible by MI's.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/server/plugins/auth/inject-identity.ts
+++ b/backend/src/server/plugins/auth/inject-identity.ts
@@ -19,13 +19,14 @@ export type TAuthMode =
       orgId?: string;
       authMethod: AuthMethod;
     }
-  // | {
-  //     authMode: AuthMode.API_KEY;
-  //     actor: ActorType.USER;
-  //     userId: string;
-  //     user: TUsers;
-  //     orgId?: string;
-  //   }
+  | {
+      authMode: AuthMode.API_KEY;
+      authMethod: null;
+      actor: ActorType.USER;
+      userId: string;
+      user: TUsers;
+      orgId: string;
+    }
   | {
       authMode: AuthMode.SERVICE_TOKEN;
       serviceToken: TServiceTokens & { createdByEmail: string };
@@ -78,8 +79,8 @@ const extractAuth = async (req: FastifyRequest, jwtSecret: string) => {
         actor: ActorType.USER
       } as const;
     case AuthTokenType.API_KEY:
-      throw new Error("API Key auth is no longer supported.");
-    // return { authMode: AuthMode.API_KEY, token: decodedToken, actor: ActorType.USER } as const;
+      // throw new Error("API Key auth is no longer supported.");
+      return { authMode: AuthMode.API_KEY, token: decodedToken, actor: ActorType.USER } as const;
     case AuthTokenType.IDENTITY_ACCESS_TOKEN:
       return {
         authMode: AuthMode.IDENTITY_ACCESS_TOKEN,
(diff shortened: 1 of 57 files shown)


backend/src/server/plugins/auth/inject-identity.ts
```diff
@@ -97,36 +119,43 @@ export const injectIdentity = fp(async (server: FastifyZodProvider) => {
     if (!authMode) return;
 
     switch (authMode) {
+      // May or may not have an orgId. If it doesn't have an org ID, it's likely because the token is from an org that doesn't enforce org-level auth.
```
maidul98, line 122: Hmm, it would be nice if there was some consistency. At least for JWT and machine identity

maidul98 (commented):
Just some small comments

backend/src/services/service-token/service-token-service.ts
```diff
@@ -142,7 +146,10 @@ export const serviceTokenServiceFactory = ({
     const updatedToken = await serviceTokenDAL.updateById(serviceToken.id, {
       lastUsed: new Date()
     });
-    return { ...serviceToken, lastUsed: updatedToken.lastUsed };
+
+    const organization = await orgDAL.findOrgByProjectId(serviceToken.projectId);
+
```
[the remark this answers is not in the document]
DanielHougaard: Yes this is true. However I feel like we should try to standardize this as much as possible. If we don't make things consistent on the backend we'll end up having a lot of "know-how". 

Imagine someone creating a new endpoint, and they enable it for machine identities and service tokens. This endpoint depends on the `orgId` being provided from the auth middleware. When they go to test it with MI's it works as you would expect, but service tokens won't work for this endpoint now. 

This is still partially an issue, since we don't _(always)_ have an orgId on JWT, and API key auth. But moving forward I would say it makes sense to keep the data in the auth middleware as consistent as possible.

frontend/src/config/request.ts
```diff
@@ -19,19 +15,22 @@ apiRequest.interceptors.request.use((config) => {
...
-    // eslint-disable-next-line no-param-reassign
-    config.headers.Authorization = `Bearer ${mfaTempToken}`;
-  } else if (token && config.headers) {
-    // eslint-disable-next-line no-param-reassign
-    config.headers.Authorization = `Bearer ${token}`;
-  } else if(providerAuthToken && config.headers) {
-    // eslint-disable-next-line no-param-reassign
-    config.headers.Authorization = `Bearer ${providerAuthToken}`;
+
+  if (config.headers) {
```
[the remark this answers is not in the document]
DanielHougaard: We didn't actually. I was doing some work in this file which didn't make it into the PR, but I decided to leave the code cleanup. Now it just checks that `config.headers` is present _once_, instead of once for each if statement.

DanielHougaard:
NOTE: The breaking change check will fail, because we've changed the create project router to take an organization slug instead of an organization ID. This is a public API endpoint documented on our [API reference page](https://infisical.com/docs/api-reference/endpoints/workspaces/create-workspace). I've spoken to @maidul98 about this back when the change was made, and we decided to move forward with changing the endpoint.

backend/src/server/routes/v2/project-router.ts
```diff
@@ -127,29 +136,24 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
       }
     },
-    onRequest: verifyAuth([AuthMode.JWT, AuthMode.API_KEY, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
     handler: async (req) => {
       const project = await server.services.project.createProject({
         actorId: req.permission.id,
         actor: req.permission.type,
-        orgId: req.body.organizationId,
+        orgSlug: req.body.organizationSlug,
```
maidul98, line 156: I thought this would be for a new project end point and not the existing one

maidul98: Oh right, no one other than the web UI uses this end point because it was not possible to create projects and invite users programmatically

backend/src/services/project/project-service.ts
```diff
@@ -86,19 +91,41 @@ export const projectServiceFactory = ({
...
+  const createProject = async ({
+    orgId,
+    orgSlug,
+    actor,
+    actorId,
+    actorOrgId,
+    workspaceName,
+    slug
+  }: TCreateProjectDTO) => {
+    if (orgSlug && orgId) {
```
maidul98, line 103: Why do we still support both

DanielHougaard: V1 project router still uses the `orgId`, I'm unsure where this is actually used though?

DanielHougaard: I don't think we actually use it. Perhaps we should just delete the endpoint entirely? @maidul98

maidul98 (approved):

maidul98 (approved):

maidul98 (approved):

Issue #1433, opened 2024-02-21
Add Self Signed SSL Certificate for SMTP


KobeUyeda:
### Describe the bug
Currently when the email server is using a self signed SSL certificate SMTP will fail and not connect up to the server. Once we add the argument to the `tls : { rejectUnauthorized: false }` to the formatSMTPConfig function it fixed the issue after we built the docker file.

```ts
export const formatSmtpConfig = () => ({
  host: envCfg.SMTP_HOST,
  port: envCfg.SMTP_PORT,
  auth:
    envCfg.SMTP_USERNAME && envCfg.SMTP_PASSWORD
      ? { user: envCfg.SMTP_USERNAME, pass: envCfg.SMTP_PASSWORD }
      : undefined,
  secure: envCfg.SMTP_SECURE,
  tls : { rejectUnauthorized: false },
  from: `"${envCfg.SMTP_FROM_NAME}" <${envCfg.SMTP_FROM_ADDRESS}>`
});
```

### To Reproduce
1. Go to 'backend/src/lib/config/env.ts'
2. add `tls : { rejectUnauthorized: false }` to the config
3. Replace the `false` with a SMTP_SSL environment variable
4. Docker Build fixes Error for SMTP

### Expected behavior
When trying to connect up to the SMTP server we were receiving this error 

{"level":50,"time":1708545357278,"pid":18,"hostname":"097fb33e7de7","severity":"ERROR","err":{"type":"Error","message":"self-signed certificate","stack":"Error: self-signed certificate\n    at TLSSocte\n    at TLSSocket.onConnectSecure (node:_tls_wrap:1674:34)\n    at TLSSocket.emit (node:events:518:28)\n    at TLSSocket.emit (node:domain:488:12)\n    at TLSSocket._finishInit (node:_tls_wrap:1085:8)\n   kedone (node:_tls
 at ssl.onhandshakedone (node:_tls_wrap:871:12)","code":"ESOCKET","command":"CONN"},"msg":"self-signed certificate"}

After adding the argument for tls: rejectUnauthorized it fixed the issue.

### Additional context
Didn't know if this was suppose to be a bug or a feature needed to the .env

KobeUyeda:
https://github.com/Infisical/infisical/pull/1464

Created a merge request for this.

iAdtya:
@KobeUyeda Hii, let me take this one !!

PR #1464, opened 2024-02-24
Issue #1433: Smtp ssl self signed

KobeUyeda:
# Description 📣

In this code I added the ability for SMTP to accept self signed SSL certificates for SMTP servers. The .env variable SMTP_SSL is set to true by default, once set to false it can connect to self signed SSL certificates.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

To test out if this worked I built the back end docker file to see if it could connect up to the SMTP server. After I everything was built and I ran the docker file I was no longer receiving SMTP errors and it said it was able to connect.

---


maidul98 (commented):
Can you please provide details to ensure that this will continue to work for SMTP services such as AWS SES (non tls), Gmail SMTP, SendGrid?

PR #1471, opened 2024-02-26
Telemetry stats event for self hosted instance on midnight

akhilmhdh:
# Description 📣

1. For self hosted infisical a stats event will be captured at every midnight to understand our users and no other telemetry event will be captured. 
2. While making this change had to update server config to use redis and found out a major race condition that could happen

On fresh bootup in replication there could be two server cfg rows created, rare but possible. Thus we keep server cfg with an id of `00000` and then fetch that one always.

3. Fixed another one while working on instance id, that @DanielHougaard reported to once that the zod generated db schema asks for value even when default is provided. The change is instead of `zod.infer` -> `zod.input` for insert and update. That's the schema changes most

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98:
In the future, please try to separate into separate PRs. The shorter the PR, the quicker it is to review

akhilmhdh:
@maidul98 yah but this one was depending on the other change. All the changes are related to this PR. Except the schema file auto generated changes the PR is small

maidul98 (approved):
Noticed a few typos, I use https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker and it works great, you should use it too.

I also saw that you removed created/updated at fields from config api and it looks like these fields are not used by the frontend so it should be fine to be breaking. 

Otherwise looks good!

PR #1474, opened 2024-02-26
Fix: Add detailed decryption error logging

DanielHougaard:
# Description 📣

Added extra logging when symmetric decryption fails on the frontend. This will be used to better debug decryption issues when it's reported by users.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/components/utilities/cryptography/crypto.ts
+++ b/frontend/src/components/utilities/cryptography/crypto.ts
@@ -210,7 +210,14 @@ const decryptSymmetric = ({ ciphertext, iv, tag, key }: DecryptSymmetricProps): 
   try {
     plaintext = aes.decrypt({ ciphertext, iv, tag, secret: key });
   } catch (err) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/utilities/cryptography/crypto.ts
```diff
@@ -210,7 +210,14 @@ const decryptSymmetric = ({ ciphertext, iv, tag, key }: DecryptSymmetricProps):
   try {
     plaintext = aes.decrypt({ ciphertext, iv, tag, secret: key });
   } catch (err) {
-    console.log("Failed to perform decryption");
+    console.log("Failed to decrypt with the following parameters", {
```
maidul98, line 213: can you please use the logger and not console.log?

DanielHougaard: We don't have a logger on the frontend unfortunately

PR #1475, opened 2024-02-27
Add support for LDAP authentication + Aliases

dangtony98:
# Description 📣

This PR adds support for the LDAP authentication method.

With  this update, an admin can configure an organization in Infisical to connect to an LDAP server (e.g. Active Directory, JumpCloud LDAP, etc.). After doing so, users that log in via the LDAP method are created to exist within the corresponding organization in Infisical only and hence can only access that organization.

This includes:
- Endpoints to create/update LDAP configuration for an organization.
- LDAP authentication logic.
- New aliases concept.
- Permission + license logic for LDAP.
- Database migrations and UX updates to support and prioritize `username` field for users hereon instead of `email`.
- Documentation for how to set up the LDAP authentication method.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


backend/src/db/migrations/20240305165532_ldap-config.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import { TableName } from "../schemas";
+import { createOnUpdateTrigger, dropOnUpdateTrigger } from "../utils";
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasTable(TableName.LdapConfig))) {
+    await knex.schema.createTable(TableName.LdapConfig, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.uuid("orgId").notNullable().unique();
+      t.foreign("orgId").references("id").inTable(TableName.Organization);
```
akhilmhdh, line 11: I think @dangtony98  forgot to add cascade here

backend/src/db/seed-data.ts
```diff
@@ -21,6 +21,7 @@ export let userPublicKey: string | undefined;
 
 export const seedData1 = {
   id: "3dafd81d-4388-432b-a4c5-f735616868c1",
+  username: process.env.TEST_USER_USERNAME || "test@localhost.local",
```
akhilmhdh, line 24: Suggestion: how about seed contains a username like `test` instead of email.

backend/src/ee/services/ldap-config/ldap-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+      throw new BadRequestError({
+        message:
+          "Failed to create LDAP configuration due to plan restriction. Upgrade plan to create LDAP configuration."
+      });
+
+    const orgBot = await orgBotDAL.transaction(async (tx) => {
+      const doc = await orgBotDAL.findOne({ orgId }, tx);
+      if (doc) return doc;
+
+      const { privateKey, publicKey } = generateAsymmetricKeyPair();
```
akhilmhdh, line 81: Can we move the encryption to outside the transaction. Because the more time a transaction takes the more table lock in time.

backend/src/ee/services/ldap-config/ldap-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+  const getLdapCfgWithPermissionCheck = async ({ actor, actorId, orgId, actorOrgId }: TOrgPermission) => {
+    const { permission } = await permissionService.getOrgPermission(actor, actorId, orgId, actorOrgId);
+    ForbiddenError.from(permission).throwUnlessCan(OrgPermissionActions.Read, OrgPermissionSubjects.Ldap);
+    return getLdapCfg({
+      orgId
+    });
+  };
+
+  // eslint-disable-next-line
+  const getLdapPassportOpts = (req: FastifyRequest, done: any) => {
```
akhilmhdh, line 288: Please avoid putting server specific things inside the service layer. You can directly pass organization slug which is the dependent on organizationSlug and then do the operations.

akhilmhdh: So i would recommend keeping the boot logic here and the parent wrapper inside the passport router i think saml also follows the same pattern

backend/src/ee/services/ldap-config/ldap-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+          );
+        }
+      });
+    } else {
+      userAlias = await userDAL.transaction(async (tx) => {
+        const uniqueUsername = await normalizeUsername(username, userDAL);
+        const newUser = await userDAL.create(
+          {
+            username: uniqueUsername,
+            email: emails[0],
```
akhilmhdh, line 377: `isAccepted: true` is missing

frontend/src/views/Settings/OrgSettingsPage/components/OrgAuthTab/LDAPModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+    
+    const {
+        control,
+        handleSubmit,
+        reset,
+    } = useForm<AddLDAPFormData>({
+        resolver: yupResolver(schema)
+    });
+    
+    useEffect(() => {
```
akhilmhdh, line 58: Inside the `useForm` there is an argument called `value` which does the same as below useEffect more effectively.

frontend/src/views/Settings/OrgSettingsPage/components/OrgAuthTab/LDAPModal.tsx
```diff
@@ -0,0 +1,229 @@
+import { useEffect } from "react";
+import { Controller, useForm } from "react-hook-form"; 
+import { yupResolver } from "@hookform/resolvers/yup";
+import * as yup from "yup";
```
akhilmhdh, line 4: We are moving towards zod. To make the validation same as both frontend and backend. 

Can you please switch to zod and any new forms you have created

docker-compose.dev.yml
```diff
@@ -126,9 +126,36 @@ services:
     ports:
       - 1025:1025 # SMTP server
       - 8025:8025 # Web UI
+  
+  openldap: # note: more advanced configuration is available
```
akhilmhdh, line 130: Can we keep the ldap in another docker compose profile like `db-test` because the more dependencies the more time to get started with docker compose local development

akhilmhdh: Ref: https://docs.docker.com/compose/profiles/

akhilmhdh (changes requested):

backend/src/ee/services/ldap-config/ldap-config-service.ts
dangtony98: Great catch. I didn't even think of this actually.

backend/src/ee/services/ldap-config/ldap-config-service.ts
dangtony98: At this point actually there shouldn't be `isAccepted` (i.e. it is false) because the user will later complete the signup flow and only at that point this becomes `true`.

backend/src/ee/services/ldap-config/ldap-config-service.ts
dangtony98: Fixed in update that I will push out today.

backend/src/ee/services/ldap-config/ldap-config-service.ts
dangtony98: I was thinking this at first but I think the encryption logic is actually not expensive; I've taken this from somewhere else in the codebase where this pattern was implemented.

frontend/src/views/Settings/OrgSettingsPage/components/OrgAuthTab/LDAPModal.tsx
dangtony98: Updated in latest version of PR that I will push soon.

docker-compose.dev.yml
dangtony98: Done. Now uses docker compose profiles, so the `LDAP` related services will only start up when the profile is specified in the start command; I also put a shortcut in the `Makefile` that is `make up-dev-ldap`.

PR #1480, opened 2024-02-27
docs: improved secret rotation documentation with better understanding

akhilmhdh:
# Description 📣

As stated. above

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---


docs/documentation/platform/secret-rotation/mysql.mdx
```diff
@@ -3,35 +3,98 @@ title: "MySQL/MariaDB"
...
+To learn more about MySQL Permissioning, please follow this [documentation](https://dev.mysql.com/doc/refman/8.0/en/privileges-provided.html).
+
 ## Working
 
-1. User's has to create the two user's for Infisical to rotate and provide them required database access
-2. Infisical will connect with your database with admin access
-3. If last rotated one was username1, then username2 is chosen to be rotated
-5. Update it's password with random value
-6. After testing it gets saved to the provided secret mapping
+1. Infisical connects to your database using the provided `admin` user account.
```
maidul98, line 21: Please the steps component https://mintlify.com/docs/content/components/steps

vmatsiiako:
Can I merge this?

docs/documentation/platform/secret-rotation/mysql.mdx
akhilmhdh: @maidul98 Steps is better when user need to follow it one by one. The working is just a set of points on how the system does it. 

Also steps without a description is not a good thing. The above points are short and precise right

maidul98:
Also can you please add example screenshots for the steps. Here is a good example of this: https://infisical.com/docs/documentation/platform/identities/universal-auth

This would help users see what example values look like

akhilmhdh:
@maidul98 updated with images

docs/images/secret-rotation/postgres-step1.png
maidul98, line 1: same for here, let's show real examples so that users can follow along easily

docs/images/secret-rotation/mysql-step1.png
maidul98, line 1: same for here, let's show real examples so that users can follow along easily

maidul98 (approved):
@akhilmhdh I have revised the documentation and removed the images for now because the placement looked off. 

Feedback for future documentation you write: 
- Please run your documentation through something like ChatGPT so that it fixes typos. The prompt i use often is:
```
Context: This is a developer documentation 
Tone: formal
Action: Fix typos and grammer and rephrase where needed
Content:
<paste your docs>
```
- When you write documentation, your should should be to step into the shoes of a user who doesn't know the feature at all and provide them with all the details they would need to use it. Detailed documentation will increase the use of the products we make

PR #1481, opened 2024-02-27
feat(admin): disable sso authentication method 

Salman2301:
# Description 📣
This add ability for super admin to disable any SSO / email authentication methods available in Infisical across organizations. Even though this methods only work, if the authentication ENV is setup properly. I was thinking we can hide the button in the frontend for the user and also add some checks to protect login/register route. So, for a large organization this can be handy.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

Admin Dashboard _ Infisical · 1 13am · 02-28

## Test
1. Checkout PR
2. goto Admin panel
3. disable email
4. logout the current user
5. login the user via email. Should throw an error

### DIFF
--- a/backend/src/db/migrations/20240227141616_disable-auth.ts
+++ b/backend/src/db/migrations/20240227141616_disable-auth.ts
@@ -1,0 +1,17 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(TableName.SuperAdmin, (t) => {
+    t.specificType("disabledAuthMethods", "text[]");
+  });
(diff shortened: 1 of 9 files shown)


maidul98:
I think what would make more sense is to just hide all of the SSO that has not been configured on the sign up page. This is because when a SSO is configured (environment varables are set for it), we can assume it is meant to be used. 

What do you think?

Salman2301:
@maidul98 agree with you on it.
This makes thing simple.

In addition hiding unconfigured SSO, we need a way to disable Local authentication method. May be will add that instead of the current UI to hide all authentication method.

I will send you my reason via private slack.

PR #1482, opened 2024-02-27
Fix: Invite all members to project when there are no members to invite

DanielHougaard:
# Description 📣

Fixed a 500 status server error occurring when attempting to invite members to a project, when there's only 1 member in the organization. The issue was happening due to our mailing service. When we try to send emails, and the recipients array is empty, it will throw a `No recipients defined` error. 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---


maidul98 (approved):
nice!

PR #1488, opened 2024-02-28
fix postgresql volume path on docker-compose.prod.yml

Kiskadee-dev:
/data/db doesn't seems to exist, would never persist data otherwise

# Description 📣

The default location of postgresql data is /var/lib/postgresql/data, /data/db doesn't exists

## Type ✨

- [ x ] Bug fix

## Results

Now it persists data


maidul98 (approved):
You are correct! Thank you

PR #1497, opened 2024-02-29
Add postgres cert to migration knex command 

maidul98:
When running the migration via knex command, we did not have cert support in postgres connection. This PR adds cert support needed to run secure schema migrations


DanielHougaard (approved):
LGTM!

PR #1498, opened 2024-02-29
patch super user migration

maidul98:
When multiple super admin configs already exist and at least one has a value equal to `ADMIN_CONFIG_UUID` then this migration will fail because it will try to update one of the configs also to ADMIN_CONFIG_UUID. Since two ids cant be the same we get the following error during migration run in helm:

```
migration file "20240226094411_instance-id.ts" failed
migration failed with error: update "super_admin" set "id" = $1 where "id" is not null - duplicate key value violates unique constraint "super_admin_pkey"
update "super_admin" set "id" = $1 where "id" is not null - duplicate key value violates unique constraint "super_admin_pkey"
Key (id)=(00000000-0000-0000-0000-000000000000) already exists.
error: update "super_admin" set "id" = $1 where "id" is not null - duplicate key value violates unique constraint "super_admin_pkey"
```

### DIFF
--- a/backend/src/db/migrations/20240226094411_instance-id.ts
+++ b/backend/src/db/migrations/20240226094411_instance-id.ts
@@ -11,7 +11,11 @@ export async function up(knex: Knex): Promise<void> {
   // this is updated to avoid race condition on replication
   // eslint-disable-next-line
   // @ts-ignore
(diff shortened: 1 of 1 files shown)


maidul98:
cc: @akhilmhdh

DanielHougaard (approved):

PR #1505, opened 2024-03-01
Feat: Agent exec and custom polling interval

DanielHougaard:
# Description 📣

This PR introduces custom polling intervals and exec commands. Additionally, now instead of re-rendering the secrets every 5 minutes, we only re-render the secrets when a change has actually happened. When a secret change is detected, 2 things will happen:
   1. Secrets are re-rendered. 
   2. The exec command (if any) will be executed.

The polling interval will default to 5 minutes if not defined, and if it's defined the minimum value is 60 seconds.

The exec command has support for a custom timeout. This means that if a command fails to execute with X amount of seconds, the execution flow will fail and print an error to the terminal. If no timeout is provided, it will default to 30 seconds.

Both the exec command and polling interval is defined on a template-level. 

### New example agent config file
```
infisical:
    address: "http://localhost:8080"
auth:
    type: "universal-auth"
    config:
        client-id: "./agent-test/client-id"
        client-secret: "./agent-test/client-secret"
        remove_client_secret_on_read: false
sinks:
    - type: "file"
      config:
          path: "./agent-test/access-token-sink"
templates:
    - source-path: "./agent-test/dot-env-secret-template"
      destination-path: "./agent-test/.target-env"
      config:
          polling-interval: 60s
          exec:
              timeout: 60
              command: docker compose --help
```

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/cli/agent-config.yaml
+++ b/cli/agent-config.yaml
@@ -1,5 +1,5 @@
 infisical:
-  address: "http://localhost:8080"
+  address: "https://app.infisical.com/"
 auth:
   type: "universal-auth"
   config:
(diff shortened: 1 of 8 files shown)


maidul98 (approved):
Works perfect!

rhythmbhiwani:
@maidul98 I pulled the recent commit from main, but then cli's go code started to give error. There is a minor change required, I created in the PR #1509

PR #1517, opened 2024-03-02
Stop ability to rename a secret to empty name from frontend

rhythmbhiwani:
# Description 📣

Fixes a bug from frontend side, which previously allowed secrets to be renamed to empty string, thus breaking the system as they won't be able to rename or delete that secret again. Added a check to update the key only if it has value.

- TODO
Although this change is done from frontend, but this validation should also be in backend, which restricts the secret name to be sent as empty

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
Attached video for reference

https://github.com/Infisical/infisical/assets/47269183/d9c7dbd6-7daa-4826-833d-af3020aa6253

---

### DIFF
--- a/backend/src/services/secret/secret-service.ts
+++ b/backend/src/services/secret/secret-service.ts
@@ -375,6 +375,10 @@ export const secretServiceFactory = ({
 
     await projectDAL.checkProjectUpgradeStatus(projectId);
 
+    if (inputSecret.newSecretName === "") {
+      throw new BadRequestError({ message: "New secret name cannot be empty" });
+    }
+
     const folder = await folderDAL.findBySecretPath(projectId, environment, path);
     if (!folder) throw new BadRequestError({ message: "Folder not  found", name: "Create secret" });
     const folderId = folder.id;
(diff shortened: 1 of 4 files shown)


rhythmbhiwani:
Which one of these looks good to inform the user
CC: @Grraahaam 
Screenshot 2024-03-02 at 6 33 44 PM
Screenshot 2024-03-02 at 6 34 08 PM
Screenshot 2024-03-02 at 6 34 19 PM

rhythmbhiwani:
Screenshot 2024-03-02 at 7 44 12 PM
Updated to look like this when trying to submit

DanielHougaard (changes requested):
Really great find, this is great!

One small change I'd like to request though, is instead of allowing the user to click the save button, and displaying an error, we should do something a little different. 

Instead, I think we should fully just disable the save secret button when the key is empty. Additionally I think we should also add a check for this on the backend. We could use our zod schema validation to check this.

rhythmbhiwani:
https://github.com/Infisical/infisical/assets/47269183/e8b59bcc-3f2e-468b-818e-e3ddfa936ecc

Updated the frontend to disable save button if secret name is empty

DanielHougaard (approved):
I committed my backend change to also make it impossible to do this through the API. Really awesome work, approved!

PR #1518, opened 2024-03-02
Added Feature to Create Secrets and Folders in all envs from overview page

### LINKED ISSUE #1273: [ENG-380] Create secrets from the overview page
### Feature description
Currently you need to click to "Explore" a particular environment on the web dashboard if you want to create a new secret. Ideally, we want to have a way to create secrets right from the "Secret overview" screen. 

### Why would it be useful?
This would save a lot of time for users who need to create secrets on a constant basis. It has been requested many times by Infisical users. 

<sub>[ENG-380](https://linear.app/infisical/issue/ENG-380/create-secrets-from-the-overview-page)</sub>

rhythmbhiwani:
# Description 📣

Closes #1273 . Adds button and popup to add secrets and folders on overview page to add in all environments at once.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Attaching Video for reference
https://drive.google.com/file/d/1H8tAu4aCqJwzBEjQwu9s8ZxkUF5MbCfP/view?usp=sharing
---


maidul98:
Really nice demo, @akhilmhdh will review this PR

akhilmhdh:
Screenshot 2024-03-04 at 2 31 33 PM

I have made some changes in the create flow. The previous one was great but ended up having a bit unpredictable flow. 

Like the input is asked for an environment with a value. Users will have no idea whether its gets override or not. Also i don't think we need to ask users for all inputs in each environment. They can just create one, duplicate in multiple env then edit it as needed in overview page like before.

Now the form gives a clear insight whether the value gets overriden or not for an existing secret

akhilmhdh (approved):

akhilmhdh:
Ready to merge @maidul98 unless @vmatsiiako has some insights to share

rhythmbhiwani:
I Agree, this ux is more clear, and flexible. Looks great

rhythmbhiwani:
@akhilmhdh There are some minor overflows in the model, can we fix that before merging.
Screenshot 2024-03-04 at 7 21 23 PM
Screenshot 2024-03-04 at 7 21 34 PM

rhythmbhiwani:
@akhilmhdh There are minor overflows in the model, can we fix that also

Screenshot 2024-03-04 at 7 21 23 PM
Screenshot 2024-03-04 at 7 21 34 PM

vmatsiiako:
IMO i liked the original flow of adding different secret values from the modal much better

rhythmbhiwani:
If we keep the different values model, we can modify that to alert user when the secret would be overwritten and update the secret just how @akhilmhdh has implemented in this new modal

akhilmhdh:
@vmatsiiako there are users with 8 environments and more than 3 many. For them the modal would be really stretched. 

Asking users to input all of them will feel confusing even if it's optional.

PR #1521, opened 2024-03-02
Show organization name in `infisical init` command

### LINKED ISSUE #1339: Including Org name in cli usage of infisical init
### Feature description
Would be great to include the org name when using the cli when running the infisical init command right now it only shows the project name which can be a bit confusing if you have similar names in both your personal and org

### Why would it be useful?
Why would this feature be useful for Infisical users? Allow cli users to distinguish between where a project lives across multiple orgs / personal accounts

rhythmbhiwani:
# Description 📣

Closes #1339 . Right now when running `infisical init` command, it only gives list of project names. So if you have multiple projects in different organizations with same name, it would be confusing to select the right project.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Previously:

https://github.com/Infisical/infisical/assets/47269183/3b703226-ecfa-43af-b3fe-e9ff4b40ed75

After Update:

https://github.com/Infisical/infisical/assets/47269183/da9266e2-e92b-4afd-8bf6-89146fb2ccbf

```sh
go run main.go init
```

---

### DIFF
--- a/backend/src/services/project/project-service.ts
+++ b/backend/src/services/project/project-service.ts
@@ -17,7 +17,6 @@ import { TProjectPermission } from "@app/lib/types";
 import { ActorType } from "../auth/auth-type";
 import { TIdentityOrgDALFactory } from "../identity/identity-org-dal";
 import { TIdentityProjectDALFactory } from "../identity-project/identity-project-dal";
-import { TOrgDALFactory } from "../org/org-dal";
 import { TOrgServiceFactory } from "../org/org-service";
 import { TProjectBotDALFactory } from "../project-bot/project-bot-dal";
 import { TProjectEnvDALFactory } from "../project-env/project-env-dal";
@@ -47,7 +46,6 @@ type TProjectServiceFactoryDep = {
   projectDAL: TProjectDALFactory;
   projectQueue: TProjectQueueFactory;
   userDAL: TUserDALFactory;
-  orgDAL: TOrgDALFactory;
   folderDAL: TSecretFolderDALFactory;
   projectEnvDAL: Pick<TProjectEnvDALFactory, "insertMany" | "find">;
   identityOrgMembershipDAL: TIdentityOrgDALFactory;
@@ -66,7 +64,6 @@ export type TProjectServiceFactory = ReturnType<typeof projectServiceFactory>;
 export const projectServiceFactory = ({
   projectDAL,
   projectQueue,
-  orgDAL,
   projectKeyDAL,
   permissionService,
   userDAL,
@@ -309,19 +306,8 @@ export const projectServiceFactory = ({
     return deletedProject;
   };
 
-  const getProjects = async (actorId: string, populateOrgName?: boolean) => {
+  const getProjects = async (actorId: string) => {
     const workspaces = await projectDAL.findAllProjects(actorId);
-    if (populateOrgName) {
-      const orgs = await orgDAL.findAllOrgsByUserId(actorId);
-      return workspaces.map((workspace) => {
-        const orgName = orgs.find((org) => org.id === workspace.orgId)?.name || "";
-        return {
-          ...workspace,
-          orgName,
-          displayName: `${workspace.name} (${orgName})`
-        };
-      });
-    }
     return workspaces;
   };
 
--- a/backend/src/services/project/project-service.ts
+++ b/backend/src/services/project/project-service.ts
@@ -17,6 +17,7 @@ import { TProjectPermission } from "@app/lib/types";
 import { ActorType } from "../auth/auth-type";
 import { TIdentityOrgDALFactory } from "../identity/identity-org-dal";
 import { TIdentityProjectDALFactory } from "../identity-project/identity-project-dal";
+import { TOrgDALFactory } from "../org/org-dal";
 import { TOrgServiceFactory } from "../org/org-service";
 import { TProjectBotDALFactory } from "../project-bot/project-bot-dal";
 import { TProjectEnvDALFactory } from "../project-env/project-env-dal";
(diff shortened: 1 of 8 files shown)


cli/packages/api/api.go
```diff
@@ -161,6 +180,22 @@ func CallGetAllWorkSpacesUserBelongsTo(httpClient *resty.Client) (GetWorkSpacesR
 		return GetWorkSpacesResponse{}, fmt.Errorf("CallGetAllWorkSpacesUserBelongsTo: Unsuccessful response:  [response=%v]", response)
 	}
 
+	// Call the organization API
+	orgResponse, err := CallGetAllOrganizations(httpClient)
+	if err != nil {
+		return GetWorkSpacesResponse{}, err
+	}
+
+	// Update organization names in workspacesResponse
```
maidul98, line 189: While this works, I think a new backend API or existing one with org name attached will be better. Have you thought about this?

rhythmbhiwani: We can consider a new api. I didn't wanted to change existing api as it may be a breaking change

rhythmbhiwani:
@maidul98 Updated the logic to backend with a query param `populateOrgName` to the `/v1/workspace` api, which adds `orgName` and `displayName` property. Attached is a reference api response
Screenshot 2024-03-03 at 11 36 35 AM


Also updated the cli to use these values, but keep original behaviour for older apis for backward compatibility

backend/src/services/project/project-service.ts
```diff
@@ -306,8 +309,19 @@ export const projectServiceFactory = ({
...
+  const getProjects = async (actorId: string, populateOrgName?: boolean) => {
     const workspaces = await projectDAL.findAllProjects(actorId);
+    if (populateOrgName) {
+      const orgs = await orgDAL.findAllOrgsByUserId(actorId);
+      return workspaces.map((workspace) => {
+        const orgName = orgs.find((org) => org.id === workspace.orgId)?.name || "";
+        return {
+          ...workspace,
+          orgName,
+          displayName: `${workspace.name} (${orgName})`
```
maidul98, line 321: The response of the API should be generic but in this case it is templated specifically for CLI. We should try to make the API responses be generic so that it can be useful for many other cases not just for the CLI. Here is what i think we should do:

When users `init` using the CLI, it should ask them to select a org and then ask them to choose a project scoped from that org. Currenly in the cli, you will see all projects from all orgs as a long list but by first scoping by org it will be less overwhelming for users. What do you think?

rhythmbhiwani: We can do that, and in that case we can keep the api same as it is was previously, i'll just change the cli code

maidul98: Awesome

rhythmbhiwani:
https://github.com/Infisical/infisical/assets/47269183/aa841301-2642-4bcc-9e98-e0dbf96d6511

@maidul98 Here's the new demo

cli/packages/util/init.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/Infisical/infisical-merge/packages/api"
+	"github.com/Infisical/infisical-merge/packages/models"
+)
+
+func GetOrganizationsNameList(organizationResponse api.GetOrganizationsResponse) []string {
+	organizations := organizationResponse.Organizations
+
+	if len(organizations) == 0 {
+		message := fmt.Sprintf("You don't have any organization created in Infisical. You must first create a organization at %s", INFISICAL_TOKEN_NAME)
```
maidul98, line 14: Why is `INFISICAL_TOKEN_NAME` apart of the message?

cli/packages/cmd/init.go
```diff
@@ -52,34 +51,48 @@ var initCmd = &cobra.Command{
...
-		workspaces := workspaceResponse.Workspaces
-		if len(workspaces) == 0 {
-			message := fmt.Sprintf("You don't have any projects created in Infisical. You must first create a project at %s", util.INFISICAL_TOKEN_NAME)
-			util.PrintErrorMessageAndExit(message)
+		organizations := organizationResponse.Organizations
+
+		organizationNames := util.GetOrganizationsNameList(organizationResponse)
+
+		prompt := promptui.Select{
+			Label: "Which of your Infisical organization would you like to get projects from?",
```
maidul98, line 65: let's change to `Which Infisical organization would you like to select a project from?`

cli/packages/cmd/init.go
```diff
@@ -52,34 +51,48 @@ var initCmd = &cobra.Command{
...
+
+		organizationResponse, err := api.CallGetAllOrganizations(httpClient)
 		if err != nil {
-			util.HandleError(err, "Unable to pull projects that belong to you")
+			util.HandleError(err, "Unable to pull organizations that belong to you")
 		}
 
-		workspaces := workspaceResponse.Workspaces
-		if len(workspaces) == 0 {
-			message := fmt.Sprintf("You don't have any projects created in Infisical. You must first create a project at %s", util.INFISICAL_TOKEN_NAME)
```
rhythmbhiwani, line 62: @maidul98  I was also confused what INFISICAL_TOKEN_NAME is a part of this message, but it was already here. Doesn't seems right

cli/packages/util/init.go
rhythmbhiwani: Should it be the `INFISICAL_DEFAULT_URL`

rhythmbhiwani: Screenshot 2024-03-05 at 4 04 01 AM
I guess it was a mistake

cli/packages/api/api.go
```diff
@@ -145,12 +145,32 @@ func CallLogin2V2(httpClient *resty.Client, request GetLoginTwoV2Request) (GetLo
...
+	return orgResponse, nil
+}
+
 func CallGetAllWorkSpacesUserBelongsTo(httpClient *resty.Client) (GetWorkSpacesResponse, error) {
 	var workSpacesResponse GetWorkSpacesResponse
 	response, err := httpClient.
 		R().
 		SetResult(&workSpacesResponse).
 		SetHeader("User-Agent", USER_AGENT).
+		SetQueryParam("populateOrgName", "true").
```
maidul98, line 173: I think you forgot to remove this

maidul98 (approved):

maidul98 (approved):
Nice work!

Issue #1511, opened 2024-03-01
db-migration error


baek-sang:
### Describe the bug
postgresql variables not working

### To Reproduce
1. https://infisical.com/docs/self-hosting/deployment-options/docker-compose

```
$ cat .env
...
# Postgres creds
POSTGRES_PASSWORD=password-123
POSTGRES_USER=user-name-123
POSTGRES_DB=infisical
...
```
```
docker logs infisical-db-migration

> backend@1.0.0 migration:latest
> knex --knexfile ./src/db/knexfile.ts --client pg migrate:latest

Requiring external module ts-node/register
Working directory changed to /backend/src/db
[dotenv@16.4.1][DEBUG] No encoding is specified. UTF-8 is used by default
[dotenv@16.4.1][DEBUG] Failed to load /.env.migration ENOENT: no such file or directory, open '/.env.migration'
[dotenv@16.4.1][DEBUG] No encoding is specified. UTF-8 is used by default
[dotenv@16.4.1][DEBUG] Failed to load /.env ENOENT: no such file or directory, open '/.env'
Using environment: production
password authentication failed for user "infisical"
error: password authentication failed for user "infisical"
    at Parser.parseErrorMessage (/backend/node_modules/pg-protocol/src/parser.ts:369:69)
    at Parser.handlePacket (/backend/node_modules/pg-protocol/src/parser.ts:188:21)
    at Parser.parse (/backend/node_modules/pg-protocol/src/parser.ts:103:30)
    at Socket.<anonymous> (/backend/node_modules/pg-protocol/src/index.ts:7:48)
    at Socket.emit (node:events:518:28)
    at Socket.emit (node:domain:488:12)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Socket.Readable.push (node:internal/streams/readable:390:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
npm notice
npm notice New minor version of npm available! 10.2.4 -> 10.5.0
npm notice Changelog: <https://github.com/npm/cli/releases/tag/v10.5.0>
npm notice Run `npm install -g npm@10.5.0` to update!
npm notice
```

### Expected behavior
no error

### Platform you are having the issue on:
- fedora 39
- docker 25.0.3

akhilmhdh:
@baek-sang you have edited the postgres password and user. These env as used by postgres image.

[...]

PR #1526, opened 2024-03-04
chore: made db connection uri in example to get value from POSTGRES_DB, POSTGRES_USER, POSTGRES_PASSWORD

akhilmhdh:
# Description 📣

Some users have reported migration error or some auth failure when running in docker compose. Most of the time it seems inside .env.example users edit the `POSTGRES_` creds but forgot to update the `DB_CONNECTION_URI`

This resolves it by changing the .env.example `DB_CONNECTION_URI` to interpolate from the `POSTGRES_*` creds

#1511 

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,7 +4,7 @@
 ENCRYPTION_KEY=6c1fe4e407b8911c104518103505b218
 
 # Required
-DB_CONNECTION_URI=postgres://infisical:infisical@db:5432/infisical
+DB_CONNECTION_URI=postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:5432/${POSTGRES_DB}
 
 # JWT
 # Required secrets to sign JWT tokens


.env.example
```diff
@@ -4,7 +4,7 @@
 ENCRYPTION_KEY=6c1fe4e407b8911c104518103505b218
 
 # Required
-DB_CONNECTION_URI=postgres://infisical:infisical@db:5432/infisical
+DB_CONNECTION_URI=postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:5432/${POSTGRES_DB}
```
maidul98, line 7: Are you sure env interpolation works this way with docker compose? I just did a quick search and i see: https://stackoverflow.com/questions/39699288/interpolate-variables-in-environment-variables-definition-with-docker-compose-ym

akhilmhdh: https://docs.docker.com/compose/compose-file/12-interpolation/

Right in the documentation and tested in local

PR #1528, opened 2024-03-04
Fixed CLI issue of updating variables using `infisical secrets set`

rhythmbhiwani:
# Description 📣

As reported here https://infisical-users.slack.com/archives/C04BSBMQAQ7/p1709582727603379, when running `infisical secrets set` to update a variable, it was throwing error from the api
Captura de pantalla 2024-03-04 a la(s) 2 02 01 p m

When calling the update `CallUpdateSecretsV3` method, we were also passing the `SecretName` parameter with the request which we pass only if we want to rename the secret, but based on the functionality of cli, we are only updating the value of the secret, not the name. So sending `SecretName` to the api was considering if it want to update the secret name, but as the variable already exists, it throws error

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
infisical secrets set VAR1=value1 --env=dev // success

infisical secrets set VAR1=value10 --env=dev // now success
```
Screenshot 2024-03-05 at 3 52 59 AM

---

### DIFF
--- a/cli/packages/api/api.go
+++ b/cli/packages/api/api.go
@@ -365,14 +365,14 @@ func CallDeleteSecretsV3(httpClient *resty.Client, request DeleteSecretV3Request
 	return nil
 }
 
-func CallUpdateSecretsV3(httpClient *resty.Client, request UpdateSecretByNameV3Request) error {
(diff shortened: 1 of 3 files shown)


DanielHougaard (approved):
LGTM!

PR #1535, opened 2024-03-06
Fix: Edge case causing project upgrade to fail

DanielHougaard:
# Description 📣

- There's an edge case where project upgrades may fail. This happens when a user in a project no longer has an organization membership. The root cause of this is because we don't remove project memberships and project keys when an organization membership is deleted. I have raised a PR to fix the project memberships/keys, #1536.
- I've also added better logging to the project queue, so we'll be able to better identity what's going wrong through CloudWatch.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/project/project-queue.ts
+++ b/backend/src/services/project/project-queue.ts
@@ -102,8 +102,11 @@ export const projectQueueFactory = ({
 
       const oldProjectKey = await projectKeyDAL.findLatestProjectKey(data.startedByUserId, data.projectId);
 
-      if (!project || !oldProjectKey) {
-        throw new Error("Project or project key not found");
+      if (!project) {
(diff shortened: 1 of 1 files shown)


backend/src/services/project/project-queue.ts
```diff
@@ -515,6 +529,8 @@ export const projectQueueFactory = ({
           throw new Error("Parts of the upgrade failed. Some secrets were not updated");
         }
 
+        throw new Error("It worked! (This is a test error message)");
```
maidul98, line 532: I think you forgot to remove this?

maidul98 (approved):

PR #1536, opened 2024-03-06
Fix: Remove project keys & memberships when organization membership is deleted

DanielHougaard:
# Description 📣

- We currently aren't removing project memberships & keys when organization memberships are removed, this PR aims to fix that. This was causing an edge case in the project upgrade, so I thought it would make sense to also fix the root cause while also fixing the project upgrade edge-case. #1535 
- Fixed super admin database seeding. 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/org/org-service.ts
+++ b/backend/src/services/org/org-service.ts
@@ -22,6 +22,8 @@ import { ActorType, AuthMethod, AuthTokenType } from "../auth/auth-type";
 import { TAuthTokenServiceFactory } from "../auth-token/auth-token-service";
 import { TokenType } from "../auth-token/auth-token-types";
 import { TProjectDALFactory } from "../project/project-dal";
+import { TProjectKeyDALFactory } from "../project-key/project-key-dal";
+import { TProjectMembershipDALFactory } from "../project-membership/project-membership-dal";
 import { SmtpTemplates, TSmtpService } from "../smtp/smtp-service";
 import { TUserDALFactory } from "../user/user-dal";
 import { TIncidentContactsDALFactory } from "./incident-contacts-dal";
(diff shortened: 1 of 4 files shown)


backend/src/services/org/org-service.ts
```diff
@@ -503,10 +509,50 @@ export const orgServiceFactory = ({
     const { permission } = await permissionService.getUserOrgPermission(userId, orgId, actorOrgId);
     ForbiddenError.from(permission).throwUnlessCan(OrgPermissionActions.Delete, OrgPermissionSubjects.Member);
 
-    const membership = await orgDAL.deleteMembershipById(membershipId, orgId);
+    const deletedMembership = await orgDAL.transaction(async (tx) => {
+      const orgMembership = await orgDAL.deleteMembershipById(membershipId, orgId, tx);
 
-    await licenseService.updateSubscriptionOrgMemberCount(orgId);
-    return membership;
+      if (!orgMembership.userId) {
```
maidul98, line 515: when `deleteMembershipById` is successful, it returns what was deleted. I'm thinking `!orgMembership.userId` should be `orgMembership.userId` before you update the membership count. Let me know if i misunderstood

DanielHougaard: `userId` might be null. If the userId is null, then we can't run the rest of the function and remove project memberships & keys. This if statement just checks if a userId is present on the deleted organization membership. If no userId is present, it will update the subscription org member count and return the membership. This is the same logic as we had before.

Let me know if that makes sense!

maidul98 (approved):

PR #1574, opened 2024-03-15
feat: additional privilege for users and identity

akhilmhdh:
# Description 📣

Introduces additional privilege support to user and machine identities

Screenshot 2024-03-15 at 10 25 21 PM

Screenshot 2024-03-15 at 10 25 36 PM

Screenshot 2024-03-15 at 10 24 41 PM

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


vmatsiiako:
I think this UI/UX is a bit unintuitive. 

Here are some thoughts: 
1. Should we perhaps only limit additional privileges to secrets? I think this will make the experience so much simpler, and I this this is the most important aspect for our current task. Maybe you can just comment out the other ones for simplicity? 
2. I think it would be great to fit the additional privileges as well as role selection into a similar modal like the one below. This way people don't need to be redirected to many different screens without proper navigation.
Screenshot 2024-03-19 at 17 08 40

akhilmhdh:
@vmatsiiako hmm maybe we can. But i do think keeping both the role and privilege section into the same set of permission to make it uniform.

The issue with the above is. Additional privilege has a lot of operations . The above is a great one if u just want to select it. 
But what about adding deleting and editing all these privileges.

When these many operations its too much effort to the end user because each time they would have to click on modal

akhilmhdh:
Can you try to create a mock up @vmatsiiako with all the details. I'll look into it

vmatsiiako:
ok, I will aim to make the mock up tomorrow

vmatsiiako:
This is what I came up with

RBAC

akhilmhdh:
Screenshot 2024-03-28 at 7 18 33 PM

New ui for role and specific privilege

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+import { AuthMode } from "@app/services/auth/auth-type";
+
+export const registerIdentityProjectAdditionalPrivilegeRouter = async (server: FastifyZodProvider) => {
+  server.route({
+    url: "/",
+    method: "POST",
+    schema: {
+      body: z.union([
+        z.object({
+          identityId: z.string().min(1).describe(IDENTITY_ADDITIONAL_PRIVILEGE.CREATE.identityId),
```
maidul98, line 22: the reason we made this not identity slug is because we currently don't have that right?

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+          projectSlug: z.string().min(1).describe(IDENTITY_ADDITIONAL_PRIVILEGE.CREATE.projectSlug),
+          slug: z
+            .string()
+            .min(1)
+            .max(60)
+            .trim()
+            .default(`privilege-${slugify(alphaNumericNanoId(12))}`)
```
maidul98, line 29: Nit: what is the reason behind this? can we remove `privilege-` prefix?

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+            .refine((val) => val.toLowerCase() === val, "Must be lowercase")
+            .refine((v) => slugify(v) === v, {
+              message: "Slug must be a valid slug"
+            })
+            .describe(IDENTITY_ADDITIONAL_PRIVILEGE.CREATE.slug),
+          permissions: z.any().array().describe(IDENTITY_ADDITIONAL_PRIVILEGE.CREATE.permissions),
```
maidul98, line 35: I think we should include a example in the describe because i checked the CSAL link and i don't think people will be able to make sense of it quickly

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+          isPackedPermission: z
```
maidul98, line 36: What is this used for? Can we just allow un-packed for all so API users don't have to decide? I think having to choose which type will create confusion for api users

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+            .boolean()
+            .optional()
+            .default(false)
+            .describe(IDENTITY_ADDITIONAL_PRIVILEGE.CREATE.isPackPermission),
+          isTemporary: z.literal(false).default(false).describe(IDENTITY_ADDITIONAL_PRIVILEGE.CREATE.isTemporary)
+        }),
+        z.object({
```
maidul98, line 43: Is there two of these objects because one is for permanent and one is the temp role? i'm not fully understanding this

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+  server.route({
+    url: "/",
+    method: "PATCH",
+    schema: {
+      body: z.object({
+        // disallow empty string
+        slug: z.string().min(1).describe(IDENTITY_ADDITIONAL_PRIVILEGE.UPDATE.slug),
+        identityId: z.string().min(1).describe(IDENTITY_ADDITIONAL_PRIVILEGE.UPDATE.identityId),
+        projectSlug: z.string().min(1).describe(IDENTITY_ADDITIONAL_PRIVILEGE.UPDATE.projectSlug),
+        data: z
```
maidul98, line 108: Can we think of a better name for this param `data`? It seems a bit ambiguous to me

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+            permissions: z.any().array().describe(IDENTITY_ADDITIONAL_PRIVILEGE.UPDATE.permissions),
+            isTemporary: z.boolean().describe(IDENTITY_ADDITIONAL_PRIVILEGE.UPDATE.isTemporary),
+            temporaryMode: z
+              .nativeEnum(IdentityProjectAdditionalPrivilegeTemporaryMode)
+              .describe(IDENTITY_ADDITIONAL_PRIVILEGE.UPDATE.temporaryMode),
+            temporaryRange: z
+              .string()
+              .refine((val) => ms(val) > 0, "Temporary range must be a positive number")
+              .describe(IDENTITY_ADDITIONAL_PRIVILEGE.UPDATE.temporaryRange),
+            temporaryAccessStartTime: z
```
maidul98, line 134: Hmm so devs will need to import a package that outputs ISO to input it here?

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const { isPackedPermission, ...data } = req.body.data;
+      const privilege = await server.services.identityProjectAdditionalPrivilege.updateBySlug({
+        actorId: req.permission.id,
+        actor: req.permission.type,
+        actorOrgId: req.permission.orgId,
+        actorAuthMethod: req.permission.authMethod,
+        ...req.body,
+        data: {
```
maidul98, line 156: nit: Lot of data fields, i think they can be better named

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+      return { privilege };
+    }
+  });
+
+  server.route({
+    url: "/",
+    method: "DELETE",
+    schema: {
+      body: z.object({
+        slug: z.string().min(1).describe(IDENTITY_ADDITIONAL_PRIVILEGE.DELETE.slug),
```
maidul98, line 172: I see a few describe that are typos. Here it says `to create ` instead of `to delete`. Can you go through and make sure the texts are correct?

maidul98: nit:  i think this would be better named privilegeSlug

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,265 @@
...
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const privilege = await server.services.identityProjectAdditionalPrivilege.deleteBySlug({
+        actorId: req.permission.id,
+        actor: req.permission.type,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId,
+        ...req.body
```
maidul98, line 189: nit: not a huge fan of this pattern `...` into the service function. If feel like it would be more readable if we put it under a field name instead  because now there are tons of parameters for this function

backend/src/ee/routes/v1/user-additional-privilege-router.ts
```diff
@@ -0,0 +1,213 @@
...
+            .max(60)
+            .trim()
+            .default(`privilege-${slugify(alphaNumericNanoId(12))}`)
+            .refine((v) => v.toLowerCase() === v, "Slug must be lowercase")
+            .refine((v) => slugify(v) === v, {
+              message: "Slug must be a valid slug"
+            })
+            .describe(PROJECT_USER_ADDITIONAL_PRIVILEGE.CREATE.slug),
+          permissions: z.any().array().describe(PROJECT_USER_ADDITIONAL_PRIVILEGE.CREATE.permissions),
+          isTemporary: z.literal(false).default(false).describe(PROJECT_USER_ADDITIONAL_PRIVILEGE.CREATE.isTemporary)
```
maidul98, line 32: Not following this part:

```
          permissions: z.any().array().describe(PROJECT_USER_ADDITIONAL_PRIVILEGE.CREATE.permissions),
          isTemporary: z.literal(false).default(false).describe(PROJECT_USER_ADDITIONAL_PRIVILEGE.CREATE.isTemporary)

```

We are adding additional privilege to a user with a existing project membership so why do we need to add permissions and is isTemporary there? Isn't temp access details kept else where?

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
maidul98: For example:

```
[["read", "secrets", {environment: "dev", secretPath: {$glob: "/"}}]]


[["read", "secrets", {environment: "dev"}], ["create", "secrets", {environment: "dev"}]]


[["read", "secrets", {environment: "dev"}]]
```

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: Yah we don't have identitySlug

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: I just appened with privilege. We can append it with anything. Thhe append is made so that it feels a privilege slug

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: By default users don't have to think for that's why its false so that if they don't send it. They don't have to care about. 

Packed is casl packing permissions. It reduces the permission json object size by 50%. In server we save it as packed. That's why we set the option. In dashboard we already pack it from our side so server don't need to pack it

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: Yes because its the starting time and it needs to be in a standard format

akhilmhdh: I think every programming language has iso standard time function

backend/src/ee/routes/v1/user-additional-privilege-router.ts
akhilmhdh: The `isTemporary` is for the additional privilege itself. Each specific privilege has temporary or permanent way

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: Can you suggest one. I thought couple of ones but felt this is better because rest are required to identity the things and data means update information

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: Its for validation. It means if you set isTemporary true u need to provide rest of the details. Else don't need to

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: The issue with the other one is mapping.

You could accidentally or even someone else could end up mapping with wrong one when you have long ones.

eg: temporaryAccessEndTime req.body.temporaryAccessStartTime

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: Changed

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
akhilmhdh: Changed

backend/src/ee/services/permission/permission-dal.ts
```diff
@@ -69,9 +74,22 @@ export const permissionDALFactory = (db: TDbClient) => {
           db.ref("updatedAt").withSchema(TableName.ProjectMembership).as("membershipUpdatedAt"),
           db.ref("authEnforced").withSchema(TableName.Organization).as("orgAuthEnforced"),
           db.ref("orgId").withSchema(TableName.Project),
-          db.ref("slug").withSchema(TableName.ProjectRoles).as("customRoleSlug")
-        )
-        .select("permissions");
+          db.ref("slug").withSchema(TableName.ProjectRoles).as("customRoleSlug"),
+          db.ref("permissions").withSchema(TableName.ProjectRoles),
+          db.ref("id").withSchema(TableName.ProjectUserAdditionalPrivilege).as("userApId"),
```
maidul98, line 79: general nit: if you can use the long form instead of the abbreviation `ap` that would increase readability

backend/src/ee/services/permission/permission-dal.ts
```diff
@@ -102,15 +120,44 @@ export const permissionDALFactory = (db: TDbClient) => {
...
+              isTemporary: userApIsTemporary
+            })
           }
         ]
       });
+
+      if (!permission?.[0]) return undefined;
       // when introducting cron mode change it here
-      const activeRoles = permission?.[0]?.roles.filter(
+      const activeRoles = permission?.[0]?.roles?.filter(
```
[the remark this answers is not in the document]
akhilmhdh: Because it will be one document. The sql nest mapper function returns array always

backend/src/ee/services/permission/permission-dal.ts
akhilmhdh: The ap is short form because it is used as an alias and immediately changed to the right one in the next one.

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
```diff
@@ -0,0 +1,272 @@
...
+        isTemporary: true,
+        permissions: JSON.stringify(packRules(req.body.permissions))
+      });
+      return { privilege };
+    }
+  });
+
+  server.route({
+    url: "/",
+    method: "PATCH",
```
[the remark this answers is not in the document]
akhilmhdh: That will not work because we need to convert permanent to temporary. That will make the whole process more complex. The update is a patch and it needs only partial information unlike the other one

backend/src/ee/services/permission/permission-dal.ts
maidul98: yeah i can see that it gets renamed in later steps but it helps to quickly read because for a second i was like what is `ap` and i think others will think the same

backend/src/ee/routes/v1/identity-project-additional-privilege-router.ts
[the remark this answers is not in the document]
maidul98: Oh i see, okay. Let's leave it as is then

backend/src/ee/services/permission/permission-dal.ts
akhilmhdh: Hm will keep in mind. But its generally i didn't see the value as it gets immediately changed

maidul98 (approved):
lgtm

PR #1584, opened 2024-03-18
Feat integration page

Salman2301:
# Description 📣
This PR add environment, integration and sorting dropdown for the integration page.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

https://github.com/Infisical/infisical/assets/40920317/5f15994e-9df5-4f52-ad32-9b4fe948858e

---


Salman2301:
I would like to add the integration as an image rather than label. It saves horizontal space that need and aslo looks good imo
Screenshot 2024-03-18 at 5 25 23 PM

Salman2301:
Added tooltip
Screenshot 2024-03-18 at 8 59 17 PM



We can remove this https://github.com/Infisical/infisical/blob/c91789e6d05cc357b78fa777400e56b058618397/frontend/public/data/frequentConstants.ts#L5

Salman2301:
@Grraahaam agree with you on Github icon
Try half the size but it's too feels to small
Screenshot 2024-03-19 at 5 19 11 AM
Try min-8 (32px) looks better
Screenshot 2024-03-19 at 5 19 26 AM

Open to suggestion

Salman2301:
With white color
Screenshot 2024-03-19 at 5 30 44 AM

Salman2301:
@akhilmhdh or @maidul98 I want to work on integration logs after this PR... if it's get merged. I would like if anyone take a look.

Salman2301:
@Grraahaam Will take a look but it's probably out of scope for this PR. If you can let the team add it somewhere in the project milestone backlog for the light mode. I think that will be the best course of action for now.

akhilmhdh:
@Salman2301 can you kindly remove the images on left side. It doen't go with the ui at all. Kinda looks bad or mismatching

akhilmhdh (changes requested):
One more change would be top bar. The select is kinda weird or rather doesn't go with the ui

I would recommend keep the filter and sort as icon buttons to top right. Then  when clicked they could select the various options as needed. 

You can keep the filter like [Radix Menu Edit > Menu](https://www.radix-ui.com/primitives/docs/components/menubar)

Salman2301:
@akhilmhdh Do you want the filter and sort to be one icon?
```
Icon click ⌄
 > Filter
     > Development
         > dev
     > Integration
        > ...
 > Sort
     > Sort options 1?
```

---
Or two icons one for filter one for sort?

Salman2301:
I use the same ui as the audit log page...

LitoMore:
Hi, I'm the maintainer of the Simple Icons project.

Please note that the dark mode feature https://cdn.simpleiocns.org/github/black/white won't work in Safari.

See this: https://stackoverflow.com/questions/58493483/issue-with-embedded-svg-images-in-dark-mode

The `<object>` tag may have some security issues, so the better solution would be the `<picture>` tag. Here is an example:

```html
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://cdn.simpleicons.org/github/white">
  <source media="(prefers-color-scheme: light)" srcset="https://cdn.simpleicons.org/github/black">
  <img alt="GitHub" src="https://cdn.simpleicons.org/github/black">
</picture>
```

In the meantime, other CDN services exist, such as jsDelivr and unpkg. Here are their differences:

### cdn.simpleicons.org

It supports brand colors and SVG inline color schemes for dark mode. This service is hosted on Vercel. It does not support history versions. It always provides the latest version of Simple Icons. Which means an icon is subject to removal. See [removal of brands](https://github.com/simple-icons/simple-icons/blob/develop/DISCLAIMER.md#removal-of-brands).

### jsDelivr & unpkg

It's more professional for CDN, and it has a faster response speed. It supports history versions. But it does not have a fill color. The color is up to your HTML context.

**Note:** jsDelivr has `Access-Control-Allow-Origin: *` support, but unpkg does not. This means you cannot use the unpkg resource in CSS `background: url(...)`.

## Which solution would be better?

It depends on what you need about the image/icon.

### If you only need black and white

The jsDelivr would be the best solution. You can use CSS `invert()` to get black and white from a single resource.

### If you want the brand color

#### jsDelivr

You could bundle the [simple-icons.json](https://github.com/simple-icons/simple-icons/blob/develop/_data/simple-icons.json) file only. It has the full icon list with its slug and brand color value. And use [CSS mask](https://developer.mozilla.org/en-US/docs/Web/CSS/mask) to apply the brand color to your jsDelivr resource. Here is an example: https://github.com/LitoMore/simple-icons-figma/blob/a5c2fae7ddcb68985c54d7ac7a0f3a9c75cca34f/source/components/icon.tsx#L62-L68.

#### cdn.simpleicons.org

This is convenient but has a slower response than jsDelivr. And it has some limitations:

Consuming the Vercel resource of [LitoMore/simple-icons-cdn](https://github.com/LitoMore/simple-icons-cdn) rapidly may quickly reach the rate limit. We're using the Vercel free plan, so this solution is not friendly to some services with a large request volume.

---

Feel free to ping me if you have any questions.

maidul98:
I think this is useful feature as users sometimes have 100s of integrations. One feedback, i think the two filters being visible from the start makes it look like a lot is going on. It would be best if the user can click some filter icon and then they see those options. By default the filters should also be set to defaults such as `all integrations` and `all environments`. After this, i think we can merge this in. Does this make sense? 
Screenshot 2024-05-07 at 12 27 51 PM

maidul98:
Intgerations v2 has been rolled out by the Infisical team

Issue #1042, opened 2023-10-02
[ENG-142] Add labels/annotations support on generated Kubernetes Secrets


asoc:
### Feature description

I would like the ability to specify labels and annotations that should be added to the generated `Secret` resource.

### Why would it be useful?

Some applications use labels/annotations for management or tracking purposes. For a specific example, ArgoCD supports declaring repository credentials via Secrets but requires a specific label on the Secret for it to be usable. [https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#repository-credentials](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#repository-credentials)

Example from the link:

```
apiVersion: v1
kind: Secret
metadata:
  name: private-repo-creds
  namespace: argocd
  labels:
    argocd.argoproj.io/secret-type: repo-creds
stringData:
  type: git
  url: https://github.com/argoproj
  password: my-password
  username: my-username
```

<sub>From [SyncLinear.com](https://synclinear.com) | [ENG-142](https://linear.app/infisical/issue/ENG-142/add-labelsannotations-support-on-generated-kubernetes-secreds)</sub>

Brutus5000:
Damn came looking for a label solution exactly because of ArgoCD.

maidul98:
Labels from `InfisicalSecret` are now propagated to the manged secret that get created https://infisical.com/docs/integrations/platforms/kubernetes#propagating-labels-and-annotations

matti-kz:
This is confusing for me. 
Now one of the propagated labels is:
```labels:
    argocd.argoproj.io/instance: manifests```
however the InfisicalSecret doesn't set ownerReference on the created secret so ArgoCD assumes new secret resource is not part of the Application and immediately deletes it?
A workaround is to pass an annotation like this
```annotations:
    argocd.argoproj.io/sync-options: Prune=false
```
but this makes an app forever out of sync.

What am I missing here?
```apiVersion: secrets.infisical.com/v1alpha1
kind: InfisicalSecret
metadata:
  name: infisical-test-secret
  namespace: infisical
  annotations:
    argocd.argoproj.io/sync-options: Prune=false    <---- this is needed to make this work at all
spec:
  hostAPI: https://app.infisical.com/api
  resyncInterval: 60
  authentication:
    serviceToken:
      serviceTokenSecretReference:
        secretName: infisical-service-token
        secretNamespace: infisical
      secretsScope:
        envSlug: prod
[...]

PR #1605, opened 2024-03-19
add managed secret creation policy

maidul98:
# Description 📣

We allow users to have the managed kubernetes secret be created in any namespace by default. While this worked for most cases, it didn't for ArgoCD. ArgoCD would not know which resource the managed secret belonged to so it would auto prune it. 

To address the auto prune, we added a resource owner to the managed secret. This resolved the issue temporarily but left users who have managed secrets in namespaces not equal to the InfisicalSecret CRD in a error state https://github.com/Infisical/infisical/issues/1042#issuecomment-1862252370 . To Address both cases, this PR adds creation policy to the managed secret.

Creation polices: `Owner` and `Orphan` (default)

`Owner` will add owner reference to the managed secret 
`Orphan` will set no owner on the managed secret

```
apiVersion: secrets.infisical.com/v1alpha1
kind: InfisicalSecret
metadata:
  name: infisicalsecret-sample
  labels:
    label-to-be-passed-to-managed-secret: sample-value
  annotations:
    reflector.v1.k8s.emberstack.com/reflection-allowed: 'true'
spec:
  hostAPI: https://app.infisical.com/api
  resyncInterval: 10
  authentication:
    serviceToken:
      serviceTokenSecretReference:
        secretName: service-token
        secretNamespace: default
      secretsScope:
        envSlug: dev
        secretsPath: "/"
  managedSecretReference:
    secretName: managed-secret
    secretNamespace: default
    creationPolicy: "Orphan"

``` 

Other small improvements:
 - a small patch was made to prevent auto generated annotations to be propagated  to the managed secret from InfisicalSecret
- hard coded the operator version to prevent users from auto upgrading 
- fixed etags issue where the server doesn't send 304 so we use the etag directly to compare 
- updated helmify in make file so that developer doesn't need to install helmify on their own 

This change is breaking for users who expect owner reference to be set. To set owner reference, users must set managedSecretReference.creationPolicy to "Owner"

## Type ✨

- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/helm-charts/secrets-operator/templates/serviceaccount.yaml
+++ b/helm-charts/secrets-operator/templates/serviceaccount.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "secrets-operator.fullname" . }}-controller-manager
+  labels:
+    app.kubernetes.io/component: rbac
+    app.kubernetes.io/created-by: k8-operator
+    app.kubernetes.io/part-of: k8-operator
+  {{- include "secrets-operator.labels" . | nindent 4 }}
+  annotations:
+    {{- toYaml .Values.controllerManager.serviceAccount.annotations | nindent 4 }}
--- a/helm-charts/secrets-operator/values.yaml
+++ b/helm-charts/secrets-operator/values.yaml
@@ -1,5 +1,15 @@
 controllerManager:
   kubeRbacProxy:
+    args:
+    - --secure-listen-address=0.0.0.0:8443
+    - --upstream=http://127.0.0.1:8080/
+    - --logtostderr=true
+    - --v=0
+    containerSecurityContext:
+      allowPrivilegeEscalation: false
+      capabilities:
+        drop:
+        - ALL
     image:
       repository: gcr.io/kubebuilder/kube-rbac-proxy
       tag: v0.15.0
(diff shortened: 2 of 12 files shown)


helm-charts/secrets-operator/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
DanielHougaard, line 1: Did you mean to include this file?

helm-charts/secrets-operator/values.yaml
```diff
@@ -22,6 +41,8 @@ controllerManager:
         cpu: 10m
         memory: 64Mi
   replicas: 1
+  serviceAccount:
```
DanielHougaard, line 44: This doesn't seem to have anything to do with the PR, I assume we want to remove this part too?

k8-operator/packages/api/api.go
```diff
@@ -77,7 +77,7 @@ func CallGetSecretsV3(httpClient *resty.Client, request GetEncryptedSecretsV3Req
 		return GetEncryptedSecretsV3Response{}, fmt.Errorf("CallGetSecretsV3: Unsuccessful response. Please make sure your secret path, workspace and environment name are all correct [response=%s]", response)
 	}
 
-	if response.StatusCode() == 304 {
+	if response.Header().Get("etag") == request.ETag {
```
DanielHougaard, line 80: Just FYI this overlaps with my existing K8 PR

DanielHougaard (changes requested):
Awesome! I left a few comments

helm-charts/secrets-operator/templates/serviceaccount.yaml
maidul98: It is apart of the new helmify that was auto generated. It used to exist but now they made it standalone

helm-charts/secrets-operator/values.yaml
maidul98: Also helmify auto generated

k8-operator/packages/api/api.go
maidul98: Should be a tiny conflict

DanielHougaard (approved):

PR #1617, opened 2024-03-22
Feat: Recursively get all secrets from all folders in specified path

DanielHougaard:
# Description 📣

We have been getting a lot of requests for fetching secrets recursively, which we didn't have support for.

When the user specifies a path when making API calls, we will recursively get all folders that lives inside that path. We then get all the secrets from those paths and return them just like we used to. The only big change is that the `getSecrets` service now functions as a loop instead of just running a single time. Additionally the response now includes a `secretPath` field on each secret. 

This works on both the list secrets and list raw secrets endpoint. 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/server/routes/v3/secret-router.ts
+++ b/backend/src/server/routes/v3/secret-router.ts
@@ -171,11 +171,9 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
       response: {
         200: z.object({
           secrets: secretRawSchema
-            .merge(
-              z.object({
-                secretPath: z.string().optional()
-              })
-            )
+            .extend({
+              secretPath: z.string().optional()
+            })
             .array(),
           imports: z
             .object({
@@ -620,20 +618,18 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
       response: {
         200: z.object({
           secrets: SecretsSchema.omit({ secretBlindIndex: true })
-            .merge(
-              z.object({
-                _id: z.string(),
-                workspace: z.string(),
-                environment: z.string(),
-                secretPath: z.string().optional(),
-                tags: SecretTagsSchema.pick({
-                  id: true,
-                  slug: true,
-                  name: true,
-                  color: true
-                }).array()
-              })
-            )
+            .extend({
+              _id: z.string(),
+              workspace: z.string(),
+              environment: z.string(),
+              secretPath: z.string().optional(),
+              tags: SecretTagsSchema.pick({
+                id: true,
+                slug: true,
+                name: true,
+                color: true
+              }).array()
+            })
             .array(),
           imports: z
             .object({
--- a/backend/src/server/routes/v3/secret-router.ts
+++ b/backend/src/server/routes/v3/secret-router.ts
@@ -157,6 +157,11 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         workspaceSlug: z.string().trim().optional().describe(RAW_SECRETS.LIST.workspaceSlug),
         environment: z.string().trim().optional().describe(RAW_SECRETS.LIST.environment),
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.LIST.secretPath),
+        recursive: z
+          .enum(["true", "false"])
+          .default("false")
+          .transform((value) => value === "true")
+          .describe(RAW_SECRETS.LIST.recursive),
         include_imports: z
           .enum(["true", "false"])
           .default("false")
@@ -165,7 +170,13 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
       }),
       response: {
         200: z.object({
-          secrets: secretRawSchema.array(),
+          secrets: secretRawSchema
+            .merge(
+              z.object({
+                secretPath: z.string().optional()
+              })
+            )
+            .array(),
           imports: z
             .object({
               secretPath: z.string(),
@@ -218,7 +229,8 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         actorAuthMethod: req.permission.authMethod,
         projectId: workspaceId,
         path: secretPath,
-        includeImports: req.query.include_imports
+        includeImports: req.query.include_imports,
+        recursive: req.query.recursive
       });
 
       await server.services.auditLog.createAuditLog({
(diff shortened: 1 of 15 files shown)


backend/src/server/routes/v3/secret-router.ts
```diff
@@ -165,7 +170,13 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
       }),
       response: {
         200: z.object({
-          secrets: secretRawSchema.array(),
+          secrets: secretRawSchema
+            .merge(
```
akhilmhdh, line 174: Just an info merge is not needed anymore. Zod has now `.extend`

backend/src/server/routes/v3/secret-router.ts
```diff
@@ -157,6 +157,11 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         workspaceSlug: z.string().trim().optional().describe(RAW_SECRETS.LIST.workspaceSlug),
         environment: z.string().trim().optional().describe(RAW_SECRETS.LIST.environment),
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.LIST.secretPath),
+        deep: z
```
akhilmhdh, line 160: @dangtony98  @maidul98 is deep a good param. I felt something like recursive would be better right. Because deep feels like how much deep

backend/src/services/secret/secret-fns.ts
```diff
@@ -45,6 +51,132 @@ export const generateSecretBlindIndexBySalt = async (secretName: string, secretB
...
+interface FolderMap {
+  [parentId: string]: TSecretFolders[];
+}
+
+export const recursivelyGetSecretPaths = ({
+  folderDAL,
+  projectEnvDAL,
+  permissionService
+}: TRecursivelyFetchSecretsFromFoldersArg) => {
+  const buildHierarchy = (folders: TSecretFolders[]): FolderMap => {
```
akhilmhdh, line 83: Can we just keep these functions outside. Then we don't need to return another function instead just directly compute result by accepting the environment etc

akhilmhdh (changes requested):
If this is getting added to CLI @maidul98 @DanielHougaard  we should have show warning when multiple secrets with same name exist and have some conflict resolution too. 

Else issue i see in replication mode the cli may have two secrets with same key due to a non deterministic resolution on picking a value from multiple secret keys

akhilmhdh:
Other than these rest looks good to me. But confict resolution is really high priority we can't expect users to do this as its too error prone

backend/src/services/secret/secret-fns.ts
DanielHougaard: Yes we can do this instead. I was just trying to replicate the structure we have for other functions like the secret interpolation function.

DanielHougaard:
> If this is getting added to CLI @maidul98 @DanielHougaard we should have show warning when multiple secrets with same name exist and have some conflict resolution too.
> 
> Else issue i see in replication mode the cli may have two secret value with same key due to a non deterministic resolution on picking a value from multiple secret keys

Currently in the CLI it will select the "most nested" secret and use that. I spoke to Maidul about this, and we came to the conclusion that it's pretty unlikely that people will have duplicate secret keys in the first place.

So in short, conflicts are handled by selecting the deepest secret if there are any conflicts.

backend/src/services/secret/secret-dal.ts
```diff
@@ -171,6 +171,50 @@ export const secretDALFactory = (db: TDbClient) => {
     }
   };
 
+  const findByFolderIds = async (folderIds: string[], userId?: string, tx?: Knex) => {
+    try {
+      // check if not uui then userId id is null (corner case because service token's ID is not UUI in effort to keep backwards compatibility from mongo)
+      if (userId && !uuidValidate(userId)) {
```
maidul98, line 177: Shouldn't this register as an error? The user wanted to pass in a user id but the id they put is is not valid so i think they should be notifed. What do you think?

backend/src/services/secret/secret-fns.ts
```diff
@@ -45,6 +51,133 @@ export const generateSecretBlindIndexBySalt = async (secretName: string, secretB
...
+        path: p.path,
+        folderId: p.folderId
+      }))
+    );
+  });
+
+  return paths;
+};
+
+export const recursivelyGetSecretPaths = ({
```
maidul98, line 126: Hmm, this doesn't look like a service but it has a child function?

cli/packages/api/api.go
```diff
@@ -277,6 +277,10 @@ func CallGetSecretsV3(httpClient *resty.Client, request GetEncryptedSecretsV3Req
 		SetQueryParam("environment", request.Environment).
 		SetQueryParam("workspaceId", request.WorkspaceId)
 
+	if request.Recursive {
+		httpRequest.SetQueryParam("recursive", "true")
```
maidul98, line 281: You forgot to make this varable. It is hard coded right now

backend/src/services/secret/secret-fns.ts
```diff
@@ -45,6 +51,133 @@ export const generateSecretBlindIndexBySalt = async (secretName: string, secretB
...
+  folderDAL,
+  projectEnvDAL,
+  permissionService
+}: TRecursivelyFetchSecretsFromFoldersArg) => {
+  const getPaths = async ({ projectId, environment, currentPath, auth }: TGetPathsDTO) => {
+    const env = await projectEnvDAL.findOne({
+      projectId,
+      slug: environment
```
maidul98, line 134: What will cause this to change between the recursive calls?  or will it change?

backend/src/services/secret/secret-fns.ts
```diff
@@ -45,6 +51,133 @@ export const generateSecretBlindIndexBySalt = async (secretName: string, secretB
...
+    // Build the folder hierarchy map
+    const folderMap = buildHierarchy(folders);
+
+    // Generate the paths paths and normalize the root path to /
+    const paths = generatePaths(folderMap).map((p) => ({
+      path: p.path === "/" ? p.path : p.path.substring(1),
+      folderId: p.folderId
+    }));
+
+    const { permission } = await permissionService.getProjectPermission(
```
maidul98, line 155: Do we need to check this in the whole recsive stack call? Or can this be checked once at the very beginning

maidul98 (commented):
Left comments

backend/src/server/routes/v3/secret-router.ts
```diff
@@ -604,19 +618,18 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
...
-                tags: SecretTagsSchema.pick({
-                  id: true,
-                  slug: true,
-                  name: true,
-                  color: true
-                }).array()
-              })
-            )
+            .extend({
+              _id: z.string(),
```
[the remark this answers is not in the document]
DanielHougaard: Yeah it's for backwards compatibility unfortunately

backend/src/services/secret/secret-dal.ts
DanielHougaard: Hmm. We do it this way in other places, so either we change it everywhere, or we keep it consistent with the existing implementations I'd say

maidul98 (approved):

PR #1626, opened 2024-03-26
chore(ci): optimize Depot build workflows

jacobwgillespie:
# Description 📣

This PR optimizes the Depot build workflows to reduce their overall runtime, with two primary changes:

1. When building the release production images, the workflows now build for both `amd64` and `arm64` first, in the `load: true` step. While the arm64 image is not actually necessary for the integration tests to run, this allows the arm64 image to be built at the same time as the amd64 one, so after the tests pass and the `push: true` step executes, both platforms will be pre-cached and can immediately push.

2. The deployment pipeline workflow built the standalone Docker image twice unnecessarily, likely a copy-paste from another CI workflow. I've removed the unnecessary `load: true` step, since nothing actually happens to the loaded image.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Optimization

---

### DIFF
--- a/.github/workflows/build-staging-and-deploy-aws.yml
+++ b/.github/workflows/build-staging-and-deploy-aws.yml
@@ -27,15 +27,6 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Set up Depot CLI
         uses: depot/setup-action@v1
-      - name: 📦 Build backend and export to Docker
-        uses: depot/build-push-action@v1
-        with:
-          project: 64mmf0n610
-          token: ${{ secrets.DEPOT_PROJECT_TOKEN }}
-          load: true
-          context: .
-          file: Dockerfile.standalone-infisical
-          tags: infisical/infisical:test
       - name: 🏗️ Build backend and push to docker hub
         uses: depot/build-push-action@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-staging-and-deploy-aws.yml
```diff
@@ -27,15 +27,6 @@ jobs:
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Set up Depot CLI
         uses: depot/setup-action@v1
-      - name: 📦 Build backend and export to Docker
```
jacobwgillespie, line 30: While this step isn't necessary and can just be removed, one note on caching: 

You were previously seeing this image rebuild even though it seemed it it should have been cached - that was because the value of `INFISICAL_PLATFORM_VERSION` was different (it's not set here, and it is set below), which caused a re-run of the Next.js build. It would be possible to make the cache work by defining the same `INFISICAL_PLATFORM_VERSION` build arg here, if you did need to keep this step for some reason.

maidul98 (approved):

.github/workflows/build-staging-and-deploy-aws.yml
maidul98: Wow i didn't even see we had a duplicate

PR #1630, opened 2024-03-26
Feat: Machine Identity support for CLI commands

DanielHougaard:
# Description 📣

- Added support for machine Identity auth for CLI commands. There are some commands which still don't support Machine Identities. There are some commands that _only_ support JWT / user login, and those commands I haven't added MI support to. 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


cli/packages/cmd/export.go
```diff
@@ -111,6 +116,8 @@ func init() {
 	exportCmd.Flags().StringP("format", "f", "dotenv", "Set the format of the output file (dotenv, json, csv)")
 	exportCmd.Flags().Bool("secret-overriding", true, "Prioritizes personal secrets, if any, with the same name over shared secrets")
 	exportCmd.Flags().String("token", "", "Fetch secrets using the Infisical Token")
+	exportCmd.Flags().String("universal-auth-client-id", "", "Machine Identity universal auth client ID")
+	exportCmd.Flags().String("universal-auth-client-secret", "", "Machine Identity universal auth client secret")
 	exportCmd.Flags().StringP("tags", "t", "", "filter secrets by tag slugs")
```
maidul98, line 121: With this approach, users would be logging in each time but login operations are limited in most cases to reduce usage. We should separate using the token from token generation. For example if system A has access to the auth id and secret and system B needs to talk to infisical then system A can login and get a short lived token and pass it on to system B. This way, system B doesn't need to know the full credentials (least privilege).

`infisical login --method=universal-auth --client-id=<> --client-secret=<>`

default login method would be `user`. In the future we will have many more auth methods such as aws login, scim, etc. The token generated from this login can be passed in as `--token` like we already do.

I would also create a new sub command to show current token details and allow renew. For example:

`infisical token renew <token>` (renews the current token but throws error if it is service token type)
`infisical token summary` (show details like num used, time left to live)

cli/packages/cmd/secrets.go
```diff
@@ -38,12 +38,12 @@ var secretsCmd = &cobra.Command{
...
 		}
 
-		infisicalToken, err := util.GetInfisicalServiceToken(cmd)
-
+		token, err := util.GetInfisicalToken(cmd)
 		if err != nil {
 			util.HandleError(err, "Unable to parse flag")
 		}
 
+		projectId, err := cmd.Flags().GetString("projectId")
```
maidul98, line 46: use dash case `project-id`

cli/packages/cmd/secrets.go
```diff
@@ -449,6 +494,18 @@ func getSecretsByNames(cmd *cobra.Command, args []string) {
 		}
 	}
 
+	if shouldExpand {
```
maidul98, line 497: bug: local secret reference isn't working 
Screenshot 2024-04-08 at 5 26 51 PM
Screenshot 2024-04-08 at 5 28 19 PM

cli/packages/util/folders.go
```diff
@@ -44,11 +44,21 @@ func GetAllFolders(params models.GetAllFoldersParameters) ([]models.SingleFolder
 		folders, err := GetFoldersViaJTW(loggedInUserDetails.UserCredentials.JTWToken, workspaceFile.WorkspaceId, params.Environment, params.FoldersPath)
 		folderErr = err
 		foldersToReturn = folders
-	} else {
+	} else if params.InfisicalToken != "" {
```
maidul98, line 47: debugging message here would be good like we have above: `log.Debug().Msg("GetAllFolders: Trying to fetch folders using logged in details")`

cli/packages/cmd/run.go
```diff
@@ -149,7 +175,15 @@ var runCmd = &cobra.Command{
 
 		log.Debug().Msgf("injecting the following environment variables into shell: %v", env)
 
-		Telemetry.CaptureEvent("cli-command:run", posthog.NewProperties().Set("secretsCount", len(secrets)).Set("environment", environmentName).Set("isUsingServiceToken", infisicalToken != "").Set("single-command", strings.Join(args, " ")).Set("multi-command", cmd.Flag("command").Value.String()).Set("version", util.CLI_VERSION))
+		Telemetry.CaptureEvent("cli-command:run",
+			posthog.NewProperties().
+				Set("secretsCount", len(secrets)).
+				Set("environment", environmentName).
+				Set("isUsingServiceToken", token.Type == "service-token").
```
maidul98, line 182: when you run the `run` command, it will throw nill pointer error

maidul98:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x104dd1a20]

goroutine 1 [running]:
github.com/Infisical/infisical-merge/packages/cmd.glob..func16(0x1400017ad00?, {0x1400025db40, 0x1, 0x2})
        /Users/maidulislam/Desktop/infisical/cli/packages/cmd/run.go:183 +0xd30
github.com/spf13/cobra.(*Command).execute(0x1053e9ca0, {0x1400025db20, 0x2, 0x2})
        /Users/maidulislam/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:920 +0x654
github.com/spf13/cobra.(*Command).ExecuteC(0x1053e96e0)
        /Users/maidulislam/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044 +0x320
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/maidulislam/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/Infisical/infisical-merge/packages/cmd.Execute()
        /Users/maidulislam/Desktop/infisical/cli/packages/cmd/root.go:32 +0x24
main.main()
        /Users/maidulislam/Desktop/infisical/cli/main.go:16 +0x178
exit status 2
```

maidul98 (commented):
Mostly nits. Did some testing and found some bugs. Let's make sure to test all the existing commands to make sure they work just like they used to

cli/packages/cmd/folder.go
```diff
@@ -36,18 +36,33 @@ var getCmd = &cobra.Command{
 			}
 		}
 
-		infisicalToken, err := util.GetInfisicalServiceToken(cmd)
-
+		projectId, err := cmd.Flags().GetString("projectId")
```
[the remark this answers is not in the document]
DanielHougaard: We already use projectId's elsewhere, but I'm not opposed to using project slugs. 

However I think project slugs should be a part of another initiative

cli/packages/cmd/login.go
```diff
@@ -55,95 +55,148 @@ var loginCmd = &cobra.Command{
...
-		err = util.StoreUserCredsInKeyRing(&userCredentialsToBeStored)
-		if err != nil {
-			log.Error().Msgf("Unable to store your credentials in system vault [%s]")
-			log.Error().Msgf("\nTo trouble shoot further, read https://infisical.com/docs/cli/faq")
-			log.Debug().Err(err)
-			//return here
-			util.HandleError(err)
-		}
+			if clientId == "" {
+				clientId = os.Getenv(util.INFISICAL_UNIVERSAL_AUTH_CLIENT_ID_NAME)
```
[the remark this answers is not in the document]
DanielHougaard: Yeah we could, however I'm just trying to think ahead to when we eventually add more ways to authenticate. In the future we may have other types of client secret's as an example I think?

cli/packages/cmd/login.go
```diff
@@ -55,95 +55,148 @@ var loginCmd = &cobra.Command{
...
+				util.HandleError(err)
+			}
 
-		whilte := color.New(color.FgGreen)
-		boldWhite := whilte.Add(color.Bold)
-		time.Sleep(time.Second * 1)
-		boldWhite.Printf(">>>> Welcome to Infisical!")
-		boldWhite.Printf(" You are now logged in as %v <<<< \n", userCredentialsToBeStored.Email)
+			boldGreen := color.New(color.FgGreen).Add(color.Bold)
+			time.Sleep(time.Second * 1)
```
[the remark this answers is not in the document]
DanielHougaard: I saw we were sleeping in the regular login print part. I assumed it was to counteract a bug, so I just copied it to the UA login part. Let me know if you still think we should remove it

cli/packages/cmd/secrets.go
DanielHougaard: Yeah I agree, but we are using camelCase for the project ID elsewhere, so I wanted to keep things consistent by using camelCase elsewhere. Let me know if you still think we should use dash-case.

cli/packages/cmd/run.go
DanielHougaard: Strange, I'm unable to reproduce this. Can you include reproduction steps please?

cli/packages/cmd/secrets.go
DanielHougaard: Unable to reproduce, can you include reproduction steps please?

CleanShot 2024-04-12 at 13 48 52@2x

DanielHougaard: Great catch! This is fixed now in this commit: https://github.com/Infisical/infisical/pull/1630/commits/71575b1d2ee49fb4103c91dca7fe06fc599de705

cli/packages/cmd/folder.go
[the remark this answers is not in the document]
maidul98: okay, let's do that at a later time

maidul98:
> Now the issue is first i need to generate it and then i can't directly pipe it out too because of intermediate information like titles etc. Then need to manually copy it too. I feel like this will cause quite traction in usage

Added flag `--silent` which will allow pipe

maidul98 (approved):
lgtm

PR #1643, opened 2024-04-02
fix(server): added sync secret for imports and added check for avoid cyclic import

akhilmhdh:
# Description 📣

1. This will trigger secret sync when you have import and the imports changes
2. Additional check when creating and updating imports to avoid cyclic import

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


dangtony98:
Just checked this PR and found the following incomplete edge-cases:
- Secret sync is only triggered for imports at 1 layer deep (i.e. if A <- B and B changes then the secret sync for A will re-trigger but if A <- B <- C and C changes then currently secret sync for A does not re-trigger).
- Secret sync is not triggered immediately for existing secrets in an import upon creation of the import and only after subsequent values are updated from within the import.

Will update this PR accordingly to ideally handle the above cases.

PR #1652, opened 2024-04-03
Feat: Request Access

DanielHougaard:
# Description 📣

Added ability for users to directly request access to paths / environments defined in a request policy. This is built on top of additional privileges, and it's highly dependent on additional privileges. 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    const duplicateRequest = await accessApprovalRequestDAL.findOne({
+      policyId: policy.id,
+      requestedBy: membership.id,
+      permissions: JSON.stringify(requestedPermissions),
+      isTemporary
+    });
+
+    if (duplicateRequest) {
```
dangtony98, line 121: I don't think this check fully works.

I was able to create multiple of the same access requests with the same contents without hitting the error "You already have a pending access request with the same criteria."

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -0,0 +1,360 @@
...
+
+    const approvers = await accessApprovalPolicyApproverDAL.find({
+      policyId: policy.id
+    });
+
+    const approverUsers = await userDAL.findUsersByProjectMembershipIds(
+      approvers.map((approver) => approver.approverId)
+    );
+
+    if (approverUsers.length !== approvers.length) {
```
dangtony98, line 110: I don't understand why is this check needed? In what case would `approverUsers.length !== approvers.length"?

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -0,0 +1,360 @@
...
+          policyId: policy.id,
+          requestedBy: membership.id,
+          temporaryRange: temporaryRange || null,
+          permissions: JSON.stringify(requestedPermissions),
+          isTemporary
+        },
+        tx
+      );
+
+      await smtpService.sendMail({
```
dangtony98, line 155: This doesn't have to be in the transaction — We don't need transactions for this function imo.

You can create the access approval request without the transaction. Then, if it fails, it will throw an error anyways which will mean that it won't ever reach `.sendMail()`.

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -0,0 +1,360 @@
...
+      actorAuthMethod,
+      actorOrgId
+    );
+    if (!membership) throw new UnauthorizedError({ message: "You are not a member of this project" });
+
+    const policies = await accessApprovalPolicyDAL.find({ projectId: project.id });
+    let requests = await accessApprovalRequestDAL.findRequestsWithPrivilegeByPolicyIds(policies.map((p) => p.id));
+
+    if (authorProjectMembershipId) {
+      requests = requests.filter((request) => request.requestedBy === authorProjectMembershipId);
```
dangtony98, line 206: Nit: This won't scale when the number of requests for a particular project becomes larger; ideally, filtering should be done at an earlier step like with `.where()` in the database call.

OK for the purposes of this PR but keep in mind for future (can deal with the problem when it arises).

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -0,0 +1,360 @@
...
+
+  const listApprovalRequests = async ({
+    projectSlug,
+    authorProjectMembershipId,
+    envSlug,
+    actor,
+    actorOrgId,
+    actorId,
+    actorAuthMethod
+  }: TListApprovalRequestsDTO) => {
```
dangtony98, line 189: I think we should add authorization resource check to this endpoint function. I don't think everyone should be able to view the list of approval requests for a project.

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -0,0 +1,360 @@
...
+    const reviewerProjectMembership = await projectMembershipDAL.findById(membership.id);
+
+    await verifyApprovers({
+      projectId: accessApprovalRequest.projectId,
+      orgId: actorOrgId,
+      envSlug: accessApprovalRequest.environment,
+      secretPath: accessApprovalRequest.policy.secretPath!,
+      actorAuthMethod,
+      permissionService,
+      approverProjectMemberships: [reviewerProjectMembership]
```
dangtony98, line 255: This function could've been modified to accept `userIds` instead of project memberships. Doing so would've avoided this additional line since then you could've just passed in `membership.userId`:

```
const reviewerProjectMembership = await projectMembershipDAL.findById(membership.id);
```

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -0,0 +1,360 @@
...
+
+type TSecretApprovalRequestServiceFactoryDep = {
+  additionalPrivilegeService: TProjectUserAdditionalPrivilegeServiceFactory;
+  additionalPrivilegeDAL: TProjectUserAdditionalPrivilegeDALFactory;
+  permissionService: TPermissionServiceFactory;
+  accessApprovalPolicyApproverDAL: TAccessApprovalPolicyApproverDALFactory;
+  projectEnvDAL: Pick<TProjectEnvDALFactory, "findOne">;
+  projectDAL: Pick<TProjectDALFactory, "checkProjectUpgradeStatus" | "findProjectBySlug">;
+  accessApprovalRequestDAL: TAccessApprovalRequestDALFactory;
+  accessApprovalPolicyDAL: TAccessApprovalPolicyDALFactory;
```
dangtony98, line 40: Nit: Incomplete `Pick` for a few data access layers.

frontend/src/views/SecretApprovalPage/components/AccessApprovalPolicyList/components/AccessPolicyForm.tsx
```diff
@@ -0,0 +1,266 @@
...
+    message: "The number of approvals should be lower than the number of approvers."
+  });
+
+type TFormSchema = z.infer<typeof formSchema>;
+
+export const AccessPolicyForm = ({
+  isOpen,
+  onToggle,
+  members = [],
+  projectSlug,
```
dangtony98, line 58: Nit: `projectSlug` not needed as a param here because you can get it from `useWorkspace()`.

frontend/src/views/SecretApprovalPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
dangtony98, line 1: Nit: Long file. I would break it down into smaller components/chunks and import them in cuz it's quite difficult to read through otherwise.

For example, the `ReviewRequestModal` can be its own file.

frontend/src/views/SecretApprovalPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
```diff
@@ -0,0 +1,621 @@
...
+};
+
+export const AccessApprovalRequest = ({
+  projectSlug,
+  projectId
+}: {
+  projectSlug: string;
+  projectId: string;
+}) => {
+  const [selectedRequest, setSelectedRequest] = useState<
```
dangtony98, line 267: Nit: This can definitely be refactored to use `usePopUp`.

Instead of storing data in `selectedRequest`, you can pass in data as second argument into `handlePopUpOpen("reviewRequest");` which you can then access inside the pop-up.

dangtony98 (commented):
Mostly nits that can be optionally implemented or disregarded but there are a few changes to be made.

Happy to assist with quick fixes.

backend/src/db/schemas/access-approval-requests-reviewers.ts
```diff
@@ -0,0 +1,26 @@
...
+// To update. Just run npm run generate:schema
+// Written by akhilmhdh.
+
+import { z } from "zod";
+
+import { TImmutableDBKeys } from "./models";
+
+export const AccessApprovalRequestsReviewersSchema = z.object({
+  id: z.string().uuid(),
+  member: z.string().uuid(),
```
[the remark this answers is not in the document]
DanielHougaard: Taken from secret approval logic.

backend/src/ee/services/access-approval-policy/access-approval-policy-service.ts
```diff
@@ -0,0 +1,274 @@
...
+    );
+    // ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionActions.Read, ProjectPermissionSub.SecretApproval);
+
+    const accessApprovalPolicies = await accessApprovalPolicyDAL.find({ projectId: project.id });
+    return accessApprovalPolicies;
+  };
+
+  const updateAccessApprovalPolicy = async ({
+    policyId,
+    approvers,
```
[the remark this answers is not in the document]
DanielHougaard: Approvers is not a required field.

frontend/src/components/basic/Divider.tsx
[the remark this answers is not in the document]
DanielHougaard: This was a quick change, fixed!

frontend/src/hooks/api/accessApproval/mutation.tsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+  });
+};
+
+export const useUpdateAccessApprovalPolicy = () => {
+  const queryClient = useQueryClient();
+
+  return useMutation<{}, {}, TUpdateAccessPolicyDTO>({
+    mutationFn: async ({ id, approvers, approvals, name, secretPath }) => {
+      const { data } = await apiRequest.patch(`/api/v1/access-approvals/${id}`, {
```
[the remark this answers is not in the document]
DanielHougaard: I totally agree. This structure was taken from the existing Secret Approval logic. We should probably change this for the secret approvals as well at some point. For now I've gone ahead and changed router grouping to be more REST compliant as suggested!

frontend/src/views/SecretApprovalPage/components/AccessApprovalRequest/components/AccessApprovalPolicyRow.tsx
```diff
@@ -0,0 +1,146 @@
...
+                    evt.preventDefault();
+                    setSelectedApprovers((state) =>
+                      isChecked ? state.filter((el) => el !== id) : [...state, id]
+                    );
+                  }}
+                  key={`create-policy-members-${id}`}
+                  iconPos="right"
+                  icon={isChecked && <FontAwesomeIcon icon={faCheckCircle} />}
+                >
+                  {user.email}
```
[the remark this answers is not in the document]
DanielHougaard: Changed this everywhere we use user.email on the frontend now!

backend/src/ee/services/access-approval-policy/access-approval-policy-fns.ts
```diff
@@ -0,0 +1,37 @@
+import { ForbiddenError, subject } from "@casl/ability";
+
+import { BadRequestError } from "@app/lib/errors";
+import { ActorType } from "@app/services/auth/auth-type";
+
+import { ProjectPermissionActions, ProjectPermissionSub } from "../permission/project-permission";
+import { TVerifyApprovers } from "./access-approval-policy-types";
+
+export const verifyApprovers = async ({
+  approverProjectMemberships,
```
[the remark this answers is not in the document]
DanielHougaard: Fixed!

backend/src/ee/services/access-approval-policy/access-approval-policy-service.ts
```diff
@@ -0,0 +1,274 @@
...
+          policyId: doc.id
+        })),
+        tx
+      );
+      return doc;
+    });
+    return { ...accessApproval, environment: env, projectId: project.id };
+  };
+
+  const getAccessApprovalPolicyByProjectId = async ({
```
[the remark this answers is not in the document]
DanielHougaard: Fixed.

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
DanielHougaard: the `membership` object does not contain a userId, so we'll need to use a dal operation to retrieve it no matter what unfortunately.

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
DanielHougaard: Works for me. Let's connect about this tomorrow when we're at the office.

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
DanielHougaard: You can consider it a safeguard. The idea behind it was that if an approver deletes their account, it will fail because it will fail to find some of the approver users. But this is redundant because we cascade. If a user is deleted the project membership will be deleted, and if the project membership is deleted, the approver itself is deleted.
_(user -> project membership -> request approver)_

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
DanielHougaard: This should be a transaction I'd say. If sendMail fails, something underlying has gone wrong, and the request shouldn't be processed I'd say. It's a safeguard that I wouldn't mind being in place. 

Note: sendMail **won't** fail if it's self-hosted, and the user hasn't set up SMTP.

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
dangtony98: Found the edge-case to replicate:
- Suppose User A is requesting access and User B can approve.
- User A requests access to any environment/path.
- User B rejects the request from User A.
- User A can now create multiple of the same access requests to the environment/path.

DanielHougaard: Great catch! This is now fixed.

dangtony98 (approved):

dangtony98:
@DanielHougaard We'll merge this when you're back together with Maidul.

vmatsiiako:
When are we thinking to merge this?

maidul98:
Let's get this done as soon as possible next week otheriwse it will get too stale

maidul98 (approved):
note: merging without group support. Group support will come at a later time

PR #1656, opened 2024-04-05
Fix: Duplicate organization memberships

DanielHougaard:
# Description 📣

We currently have no unique constraint on the `userId` and `orgId` field, which has led to duplicate organization memberships. We have as a part of this PR removed all duplicate (and broken) organization memberships from the production database. 

After this has been deployed, we should be getting error logs when it tries to create a duplicate membership, which will help us narrow down the exact cause of this. 

A small note: If self-hosted customers have duplicate memberships, they will have trouble upgrading to the latest version as the migration will fail to apply if there are duplicate memberships. To fix this they would simply have to remove the duplicate memberships from their database. This is an unlikely case but worth mentioning. 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/db/migrations/20240405000045_org-memberships-unique-constraint.ts
+++ b/backend/src/db/migrations/20240405000045_org-memberships-unique-constraint.ts
@@ -1,0 +1,15 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(TableName.OrgMembership, (table) => {
+    table.unique(["userId", "orgId"]);
+  });
+}
+
+export async function down(knex: Knex): Promise<void> {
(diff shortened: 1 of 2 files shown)


dangtony98 (approved):

backend/src/db/migrations/20240405000045_org-memberships-unique-constraint.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(TableName.OrgMembership, (table) => {
+    table.unique(["userId", "orgId"]);
```
maidul98, line 7: I understand this prevents it but how could this happen in the first place? Did we root cause it?

PR #1658, opened 2024-04-05
Feat: Delete duplicate memberships migration

DanielHougaard:
# Description 📣

This is an addition to the existing migration (#1656) which adds a unique constraint to organization memberships to avoid duplicates. This migration will remove all existing duplicate organization memberships as a part of the migration.


maidul98 (approved):
lgtm

PR #1675, opened 2024-04-10
Environment variables with a false value should be properly set

alexbarcelo:
# Description 📣

By default, this template was eating the environment variables with a false value (and trying to quote them resulted in an undesirable double quote due to the ` | quote` that appears there).

Explicitly managing the falsy value should solve this issue.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---


alexbarcelo:
I realized that my tests were bogus. The bug remains, my fix is not fixing it (a dirty secret was giving me the illusion of having solved it).

It needs a different fix (either I am blind and is a trivial template fix, or there is some mismatch between types that I don't know how to solve).

PR #1678, opened 2024-04-11
FIx: Fetching workspaces with no environments

DanielHougaard:
# Description 📣

Currently there's a bug when trying to fetch workspaces with no environments, which results in the API throwing a 500 error. This PR aims to fix this issue. Additionally I've also disabled the creation of new integrations when no environments are present on the project (see screenshot below)

CleanShot 2024-04-12 at 01 46 31

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/project/project-dal.ts
+++ b/backend/src/services/project/project-dal.ts
@@ -126,13 +126,11 @@ export const projectDALFactory = (db: TDbClient) => {
 
   const findProjectById = async (id: string) => {
     try {
-      const workspaces = await db(TableName.ProjectMembership)
+      const workspaces = await db(TableName.Project)
         .where(`${TableName.Project}.id`, id)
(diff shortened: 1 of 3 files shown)


backend/src/services/project/project-dal.ts
```diff
@@ -126,13 +126,11 @@ export const projectDALFactory = (db: TDbClient) => {
 
   const findProjectById = async (id: string) => {
     try {
-      const workspaces = await db(TableName.ProjectMembership)
```
maidul98, line 129: strange we used to do it like this

DanielHougaard: Yeah I don't see why we're doing it like that. It makes sense to just get the project and joining environments

maidul98 (approved):
lgtm

PR #1683, opened 2024-04-14
fix: resolved slow audit log list

akhilmhdh:
# Description 📣

1. Added end date by default to reduce the dataset on sorting

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/routes/v1/project-router.ts
+++ b/backend/src/ee/routes/v1/project-router.ts
@@ -3,7 +3,7 @@ import { z } from "zod";
 import { AuditLogsSchema, SecretSnapshotsSchema } from "@app/db/schemas";
 import { EventType, UserAgentType } from "@app/ee/services/audit-log/audit-log-types";
 import { AUDIT_LOGS, PROJECTS } from "@app/lib/api-docs";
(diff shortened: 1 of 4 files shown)


maidul98 (approved):
makes sense to me. For others, the query looks like this under the hood:


```
  const find = async (
    { orgId, projectId, userAgentType, startDate, endDate, limit = 20, offset = 0, actor, eventType }: TFindQuery,
    tx?: Knex
  ) => {
    try {
      const sqlQuery = (tx || db)(TableName.AuditLog)
        .where(
          stripUndefinedInWhere({
            projectId,
            orgId,
            eventType,
            actor,
            userAgentType
          })
        )
        .limit(limit)
        .offset(offset)
        .orderBy("createdAt", "desc");
      if (startDate) {
        void sqlQuery.where("createdAt", ">=", startDate);
      }
      if (endDate) {
        void sqlQuery.where("createdAt", "<=", endDate);
      }
      const docs = await sqlQuery;
      return docs;
    } catch (error) {
      throw new DatabaseError({ error });
    }
  };
```

PR #1703, opened 2024-04-18
Feat: CLI Integration Tests (Phase I)

DanielHougaard:
# Description 📣

First phase of CLI tests. Due to the errors we've been having lately with some CLI releases, we have deemed it a priority to create release tests for the CLI.

Additionally I found some issues while writing the tests with imports & recursive mode for raw endpoints / machine identities. Those issues are also fixed in this PR.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


cli/packages/cmd/run.go
```diff
@@ -310,7 +311,10 @@ func execCmd(cmd *exec.Cmd) error {
 		return fmt.Errorf("failed to wait for command termination: %v", err)
 	}
 
-	waitStatus := cmd.ProcessState.Sys().(syscall.WaitStatus)
-	os.Exit(waitStatus.ExitStatus())
+	if !util.IS_TEST_MODE {
```
maidul98, line 314: i think this pattern wil be hard to maintain as the team/codebase grows as the test related logic gets marinated into app logic. Here is my proposal:

Use [Snapshot testing](https://jestjs.io/docs/snapshot-testing). It will take snapshots of outputs and do a match check between what is expected and what is being outputted. This way, we know the output is always what we expect and don't need to do any sort of output parsing. 

One tool you can use is this one https://github.com/bradleyjkemp/cupaloy but you can look into others as well.

Or you can go a more manaul apprach where the expected output is saved to file for each test and then check to see if they match up when you run the actual command. What do you think?

maidul98 (approved):
LGTM, let's check if envs are injected during the run command in the next PR

PR #1705, opened 2024-04-18
Groups Phase 2 (Pending Group Additions)

dangtony98:
# Description 📣

This PR updates the groups implementation to allow assigning users who have not finished completing their account to groups; previously, it was only possible to add completed users to a group. 

This PR also makes minor SCIM updates (to be further ironed out) to coordination of pushing groups over to Infisical.

More specifically:
- We achieve adding incomplete users to groups by adding them to the group only after they finish signing up.
- The new `PendingGroupAddition` table is used to keep track of who needs to be added to a group after signup.
- Upon finishing signing up, if pending group addition(s) exist for the user, then they are converted to group memberships.

I've added multiple abstractions/functions to make it easier to add any user to a group (you can pass in the `userId` of any user you wish to add to a group and it will automatically take care of either directly adding them to the group or adding pending group additions; whichever is appropriate depending on their `isAccepted` status).

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


DanielHougaard (changes requested):
Nice! Left a few comments, mainly about transactions

I also ran into an issue with connecting groups to projects where it gives a `bad public key size` error -- I followed up on this via. Slack!

DanielHougaard (approved):
lgtm!

PR #1721, opened 2024-04-23
Audit log streams

akhilmhdh:
# Description 📣

Implemented audit log streams where audit log gets send to external logging providers via http. 

Screenshot 2024-04-23 at 10 46 08 PM
Screenshot 2024-04-23 at 10 46 03 PM

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/ee/services/audit-log-stream/audit-log-stream-service.ts
```diff
@@ -0,0 +1,174 @@
...
+};
+
+export type TAuditLogStreamServiceFactory = ReturnType<typeof auditLogStreamServiceFactory>;
+
+export const auditLogStreamServiceFactory = ({
+  auditLogStreamDAL,
+  permissionService,
+  licenseService
+}: TAuditLogStreamServiceFactoryDep) => {
+  const create = async ({ url, actor, token, actorId, actorOrgId, actorAuthMethod }: TCreateAuditLogStreamDTO) => {
```
maidul98, line 33: let's add verification that the sends a sample log to the provided endpoint just to make sure what they have entered is valid. We just listen for 200

backend/src/lib/api-docs/constants.ts
```diff
@@ -614,3 +614,21 @@ export const INTEGRATION = {
     integrationId: "The ID of the integration object."
   }
 };
+
+export const AUDIT_LOG_STREAMS = {
+  CREATE: {
+    url: "The socket URL to push logs to.",
```
maidul98, line 620: I think there is some terminology mix up because we just send the logs via http, there is no socket or streams right?

backend/src/lib/config/env.ts
```diff
@@ -119,6 +119,7 @@ const envSchema = z
   })
   .transform((data) => ({
     ...data,
+    isCloud: Boolean(data.LICENSE_SERVER_KEY),
```
maidul98, line 122: where do we use this in this PR?

maidul98 (changes requested):
Looks good, just left some small comments. Let's add docs for it as we spoke into this PR

PR #1729, opened 2024-04-24
Groups Phase 3 (LDAP)

dangtony98:
# Description 📣

This PR adds the ability to sync LDAP groups to internal groups in Infisical. More specifically:
- It adds two fields that users can specify (Group Search Base / Group DN and Group Search Filter) for help Infisical query for an LDAP user's group memberships.
- It adds a new table `LdapGroupMap` to keep track of which group CN in the LDAP server should be mapped to which `groupId` in Infisical.
- Upon logging into Infisical via LDAP, Infisical will add/remove the user from internal group(s) based on their LDAP group membership.

For example, you can set up a map from an LDAP group with CN `Engineers` to an internal group in Infisical called `Engineers`. Upon completing their account, users part of `Engineers` in the LDAP server will be automatically added to the `Engineers` group and provisioned access to any project that the group is part of. If the user is removed from the LDAP group, then upon their next login into Infisical, they will be de-provisioned from the internal group.

This PR also updates the LDAP documentation to include instructions for configuring LDAP group mappings.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


backend/src/db/migrations/20240423023203_ldap-config-groups.ts
```diff
@@ -0,0 +1,34 @@
...
+      t.uuid("groupId").notNullable();
+      t.foreign("groupId").references("id").inTable(TableName.Groups).onDelete("CASCADE");
+      t.unique(["ldapGroupCN", "groupId", "ldapConfigId"]);
+    });
+  }
+
+  await createOnUpdateTrigger(knex, TableName.LdapGroupMap);
+
+  await knex.schema.alterTable(TableName.LdapConfig, (t) => {
+    t.string("groupSearchBase").notNullable().defaultTo("");
```
akhilmhdh, line 22: If its empty string as default its better to keep it nullable

backend/src/ee/services/ldap-config/ldap-fns.ts
```diff
@@ -0,0 +1,41 @@
...
+          const dn = entry.dn.toString();
+          const regex = /cn=([^,]+)/;
+          const match = dn.match(regex);
+          // parse the cn from the dn
+          const cn = (match && match[1]) as string;
+
+          groups.push({ dn, cn });
+        });
+        res.on("error", (error) => {
+          console.error(`error: ${error.message}`);
```
akhilmhdh, line 32: Use logger.error

backend/src/ee/services/ldap-config/ldap-config-service.ts
```diff
@@ -396,6 +452,78 @@ export const ldapConfigServiceFactory = ({
...
+          }
+        })
+      ).map((userGroupMembership) => userGroupMembership.groupId);
+
+      const userGroupMembershipGroupIdsSet = new Set(ldapGroupIdsCurrentlyPartOf);
+
+      for await (const group of groupsToBePartOf) {
+        if (!userGroupMembershipGroupIdsSet.has(group.id)) {
+          // add user to group that they should be part of
+          await addUsersToGroupByUserIds({
```
akhilmhdh, line 491: I think both the add and delete to user group must be one transaction process

backend/src/ee/routes/v1/ldap-router.ts
```diff
@@ -44,26 +46,94 @@ export const registerLdapRouter = async (server: FastifyZodProvider) => {
...
         try {
-          const { isUserCompleted, providerAuthToken } = await server.services.ldap.ldapLogin({
-            externalId: user.uidNumber,
-            username: user.uid,
-            firstName: user.givenName,
-            lastName: user.sn,
-            emails: user.mail ? [user.mail] : [],
-            relayState: ((req as unknown as FastifyRequest).body as { RelayState?: string }).RelayState,
-            orgId: (req as unknown as FastifyRequest).ldapConfig.organization
+          const ldapConfig = (req as unknown as FastifyRequest).ldapConfig as LDAPConfig;
```
akhilmhdh, line 68: Just a suggestion try to keep the async await everywhere and avoid mixing up async await with then catch.

It's a bit confusing to read as you have to switch between two mental models

akhilmhdh:
Other than some suggestions rest looks good to me

PR #1762, opened 2024-04-29
Groups Phase 2B (Trust external SAML/LDAP email option, email verification, SCIM user ID ref update)

dangtony98:
# Description 📣

This PR adds a variety of updates largely centered around not trusting the emails coming back from an external SAML/LDAP identity provider.

Problem Context: Given Infisical's multi-tenant approach, users can be part of multiple organizations in an instance of Infisical. The concern was that a bad actor could in the future try to spoof an email of an existing Infisical user in another organization by connecting to an untrusted external IdP (e.g. a self-hosted instance and optionally modified instance of KeyCloak) in their own organization. Once connected, they could log into the bad actor's organization and switch organizations to access the legitimate user's other organizations.

In practice, this is not yet problematic because Infisical imposes the additional master decryption key / password requirement atop existing authentication with the external IdP.

Solution: To further tighten security, we introduce the optional ability for operators to specify whether or not they want to trust SAML/LDAP emails in the admin panel / server configuration; by default, SAML/LDAP emails will **not** be trusted. As part of this, we re-configure the SAML/LDAP authentication flows to create new users with aliases with the untrusted email that do not conflict with any existing users with the same claimed email; upon their first signup/login, SAML/LDAP users are prompted to verify their email via verification code — If the email is verified and an existing user exists with the same verified email, then the new user account is merged with the existing account automatically, maintaining the desired multi-tenancy with merged identities. The end result is a user that is part of multiple organizations where they can be known under different aliases.

More specifically, this PR entails:
- Option for admins to specify whether or not to trust SAML/LDAP emails.
- User aliases implemented for SAML authentication.
- Email verification upon signing up / completing a user's account via SAML/LDAP for the first time (frontend/backend components).
- Switch from referring to users by their `userId` to `orgMembershipId` for SCIM (you may want to test this as well).

In the future, we will work towards allowing users to verify their emails whenever after they've signed up; it's just that the resource merging operation for that would be complex and out-of-scope for this PR.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


frontend/src/views/Signup/SignupSSO.tsx
```diff
@@ -27,7 +54,15 @@ export const SignupSSO = ({ providerAuthToken }: Props) => {
             providerAuthToken={providerAuthToken}
           />
         );
-      case 1:
+      // case 2:
```
DanielHougaard, line 57: Can we remove this now that we do automatic merging?

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -432,17 +468,19 @@ export const scimServiceFactory = ({
 
     if (!active) {
       // tx
-      await deleteOrgMembership({
+      await deleteOrgMembershipFn({
```
DanielHougaard, line 471: I think we should wrap this entire function in a transaction to ensure that no memberships are actually deleted if one part of the function fails

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -386,17 +420,19 @@ export const scimServiceFactory = ({
     });
 
     if (!active) {
-      await deleteOrgMembership({
+      await deleteOrgMembershipFn({
```
DanielHougaard, line 423: Should ideally also be a transaction

DanielHougaard (changes requested):
Left some smaller comments for you. I also tested it on my end, and it works great!

backend/src/ee/services/scim/scim-service.ts
dangtony98: The function actually already is a transaction; the transaction is inside of it.

backend/src/ee/services/scim/scim-service.ts
dangtony98: The function actually already is a transaction; the transaction is inside of it.

maidul98 (approved):
Reviewed the migration file, lgtm. I assume rest has been reviewed by @DanielHougaard

PR #1778, opened 2024-05-03
AWS IAM Authentication Method

dangtony98:
# Description 📣

This PR adds the new AWS IAM authentication method to Identities.

With this method, IAM principals / resources in AWS avoid secret zero by not  needing to store an additional token to authenticate with Infisical. Instead, they can use credentials/metadata already present in AWS to prove their identity to Infisical — this works by relaying a signed `GetCallerIdentity` query through Infisical to be cross-checked with AWS; if successful and certain allow conditions are met, then the entity is considered authenticated and an access token granting access to the Infisical API is returned.

Workflow:
1. The operator sets up the AWS IAM authentication method in Infisical. As part of the setup, they specify what **Allowed Principal ARNs** and **Allowed Account IDs** can be trusted. They also specify the a **STS Endpoint** for Infisical to use to send the verification query.
2. The operator programs the client application in AWS to send a signed `GetCallerIdentity` query to Infisical at the `/api/v1/auth/aws-iam-auth/login` endpoint containing the `identityId` that they are authenticating as and `iamRequestBody` and `iamRequestHeaders` as params (in practice this would likely be done via SDK or other client but it can also be done manually).
3. Infisical forwards the params to AWS for verification according to the **STS Endpoint** defined in the previous step.
4. If successful, then Infisical checks the returned identity `Arn` and `Account` properties against the `Allowed Principal ARN` and `Allowed Account IDs` defined in step 1.
5. If successful, then Infisical returns an identity access token (similar to in Universal Auth).

Note 1: The AWS IAM authentication method will require additional engineering work for clients like Infisical Agent and SDKs to support it.

Note 2: The AWS IAM authentication method requires a network connection from Infisical to AWS to verify the IAM principal.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


backend/src/server/routes/v1/identity-aws-iam-auth-router.ts
```diff
@@ -0,0 +1,269 @@
...
+
+export const registerIdentityAwsIamAuthRouter = async (server: FastifyZodProvider) => {
+  server.route({
+    method: "POST",
+    url: "/aws-iam-auth/login",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      description: "Login with Universal Auth",
```
maidul98, line 23: small typo here, i think you mean AWS IAM?

backend/src/db/migrations/20240502044531_identity-aws-iam-auth.ts
```diff
@@ -0,0 +1,29 @@
...
+    await knex.schema.createTable(TableName.IdentityAwsIamAuth, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.bigInteger("accessTokenTTL").defaultTo(7200).notNullable();
+      t.bigInteger("accessTokenMaxTTL").defaultTo(7200).notNullable();
+      t.bigInteger("accessTokenNumUsesLimit").defaultTo(0).notNullable();
+      t.jsonb("accessTokenTrustedIps").notNullable();
+      t.timestamps(true, true, true);
+      t.uuid("identityId").notNullable().unique();
+      t.foreign("identityId").references("id").inTable(TableName.Identity).onDelete("CASCADE");
+      t.string("stsEndpoint").notNullable();
```
maidul98, line 17: i thought we made it so that users select the region? Or did you figure out something better ?

maidul98 (commented):
two small comments, the rest looks the good to me

backend/src/db/migrations/20240502044531_identity-aws-iam-auth.ts
dangtony98: I re-evaluated and realized that it's actually appropriate to allow users to specify the `stsEndpoint` instead of `region` since there're a few unique FIPS-certified endpoints that have a slightly different format being `sts-fips.<region>.amazonaws.com`. Previously, there was an attack-vector consideration in mind but this concern has since been resolved.

It might just be easier to allow people to pass in their STS endpoint and do validation on the list of possible STS endpoints that can be accepted.

maidul98 (dismissed):
lgtm!

PR #1781, opened 2024-05-04
feat: added secret expand option in secrets get API

sheensantoscapadngan:
# Description 📣
- raw secrets can now be expanded in the server-side 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v3/secret-router.ts
+++ b/backend/src/server/routes/v3/secret-router.ts
@@ -166,7 +166,7 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         workspaceSlug: z.string().trim().optional().describe(RAW_SECRETS.LIST.workspaceSlug),
         environment: z.string().trim().optional().describe(RAW_SECRETS.LIST.environment),
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.LIST.secretPath),
-        expand: z
+        expandSecretReferences: z
           .enum(["true", "false"])
           .default("false")
           .transform((value) => value === "true")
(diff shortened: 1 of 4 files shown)


backend/src/server/routes/v3/secret-router.ts
```diff
@@ -166,6 +166,11 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         workspaceSlug: z.string().trim().optional().describe(RAW_SECRETS.LIST.workspaceSlug),
         environment: z.string().trim().optional().describe(RAW_SECRETS.LIST.environment),
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.LIST.secretPath),
+        expand: z
+          .enum(["true", "false"])
+          .default("false")
+          .transform((value) => value === "true")
+          .describe(RAW_SECRETS.LIST.expand),
```
sheensantoscapadngan, line 173: intentionally made this false so that we can control the migration towards this pattern of use. my main concern is in how this would affect overall server latency and resource utilization. I'm thinking for this one we can start slow via the Github action I'm currently working on

backend/src/server/routes/v3/secret-router.ts
```diff
@@ -166,6 +166,11 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         workspaceSlug: z.string().trim().optional().describe(RAW_SECRETS.LIST.workspaceSlug),
         environment: z.string().trim().optional().describe(RAW_SECRETS.LIST.environment),
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.LIST.secretPath),
+        expand: z
```
maidul98, line 169: I think it might be better if we call this `expandSecretReferences`. This way anyone looking at it won't have to be like expand what? What do you think?

maidul98 (changes requested):
Looks good to me. I also tested it on my end and it works as expected. Left one comment about field name

backend/src/server/routes/v3/secret-router.ts
maidul98: resolved it

maidul98 (approved):

PR #1804, opened 2024-05-08
Feature: add support for secret folder rename in the overview page

sheensantoscapadngan:
# Description 📣
- Added support for secret folder rename in the overview page
- Added validation to ensure that folder names remain unique

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v1/secret-folder-router.ts
+++ b/backend/src/server/routes/v1/secret-folder-router.ts
@@ -141,7 +141,7 @@ export const registerSecretFolderRouter = async (server: FastifyZodProvider) => 
         }
       ],
       body: z.object({
-        workspaceId: z.string().trim().describe(FOLDERS.UPDATE.workspaceId),
+        projectSlug: z.string().trim().describe(FOLDERS.UPDATE.projectSlug),
         folders: z
           .object({
             id: z.string().describe(FOLDERS.UPDATE.folderId),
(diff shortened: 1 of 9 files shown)


frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
```diff
@@ -236,6 +242,64 @@ export const SecretOverviewPage = () => {
...
+    } else if (processedUpdates === totalFolderMatchCount) {
+      createNotification({
+        type: "success",
+        text: "Successfully renamed folder across environments"
+      });
+    } else if (processedUpdates > 0) {
+      createNotification({
+        type: "warning",
+        text: "Partially renamed folder across environments"
+      });
```
DanielHougaard, line 292: I'm not sure if it's a good idea to allow partial updates. I feel like it would make more sense to have a "bulk update" endpoint like we have with secrets, which would be wrapped in a transaction. That way you can always expect all your data to be updated correctly, or it will throw an error and nothing will be updated

DanielHougaard (changes requested):

frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
maidul98: Yeah, i agree. All or nothing for rename at the overview level because the user is intending to change it for all of the environments

sheensantoscapadngan: Got this. But to clarify, if a rename can't be made in an environment because a folder with an existing name already exists, should it stop the renaming of all the other folders in other environments?

sheensantoscapadngan: resolved via DM, yes it should error out ^

backend/src/server/routes/v1/secret-folder-router.ts
```diff
@@ -127,6 +127,71 @@ export const registerSecretFolderRouter = async (server: FastifyZodProvider) =>
...
+    },
+    schema: {
+      description: "Update folders by batch",
+      security: [
+        {
+          bearerAuth: []
+        }
+      ],
+      body: z.object({
+        workspaceId: z.string().trim().describe(FOLDERS.UPDATE.workspaceId),
```
DanielHougaard, line 144: Can we make the endpoint use project slugs? We are trying to move away from ID-based endpoints and instead move to slugs

So instead of `workspaceId` you would name it `projectSlug`

Here's an example route that uses project slugs instead of project ID's: https://github.com/Infisical/infisical/blob/b4a2123fa3812eb1bb9d73333865f4d96480d58c/backend/src/ee/routes/v1/dynamic-secret-router.ts#L158

backend/src/server/routes/v1/secret-folder-router.ts
```diff
@@ -127,6 +127,71 @@ export const registerSecretFolderRouter = async (server: FastifyZodProvider) =>
...
+        folders: req.body.folders,
+        actorId: req.permission.id,
+        actor: req.permission.type,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId,
+        projectId: req.body.workspaceId
+      });
+
+      await Promise.all(
+        req.body.folders.map(async (folder, ind) => {
```
DanielHougaard, line 173: Nit: Naming `ind` `index` for clarity

DanielHougaard (changes requested):
Looks good! Left some additional comments

DanielHougaard (approved):

PR #1817, opened 2024-05-12
feat: cli env upload

### LINKED ISSUE #1612: .env upload via CLI
### Feature description
Would like to to upload the .env file with the [infisical cli](https://infisical.com/docs/cli/commands/secrets). Currently working on a tiny shell script to read the .env as an array and pass that to `infisical secrets set "$env_vars_str" --env new-prod` but it's hackish and it would be nice if there was an option to match the web console's drag and drop functionality

### Why would it be useful?
Server admins and devs don't spend a lot of time in the web console. Nice alternative to drag and drop GUI option.

### Additional context
Briefly spoke with Vlad on [Slack](https://infisical-users.slack.com/archives/C04BSBMQAQ7/p1710961685778639).

quinton11:
Closes #1612 
@maidul98 

@akhilmhdh wrt the commented out block in `secret-service.ts`, using the cli secrets set command where in this case the secret key already exists, instead of updating the existing secret with the new value it always gave this response
```
error: CallUpdateSecretsV3: Unsuccessful response. Please make sure your secret path, workspace and environment name are all correct [response={"statusCode":400,"message":"Secret already exist","error":"BadRequest"}]
Unable to process secret update request
```
Commenting out that block resolves it. Is it by design or its a bug?

### DIFF
--- a/backend/src/services/secret/secret-service.ts
+++ b/backend/src/services/secret/secret-service.ts
@@ -310,16 +310,16 @@ export const secretServiceFactory = ({
     }
 
     let newSecretNameBlindIndex: string | undefined;
-    if (inputSecret?.newSecretName) {
-      const { keyName2BlindIndex: kN2NewBlindIndex } = await fnSecretBlindIndexCheck({
-        inputSecrets: [{ secretName: inputSecret.newSecretName }],
-        folderId,
-        isNew: true,
-        blindIndexCfg,
-        secretDAL
-      });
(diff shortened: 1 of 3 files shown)


akhilmhdh:
Hey @quinton11 Its a valid block and it should not be commented out. Its a check to ensure the new key submitted doesn't  exist

quinton11:
@akhilmhdh pushed an update undoing the comment

Issue #1103, opened 2023-10-20
Possibility to export env-s with different quota character


newadbie:
### Feature description
At this moment if we would use CLI command `infisical export > .env` all env-s will be exported in this pattern
`ENV_NAME=`VALUE`` i think it is worth to have possibility to dynamicly change quota character it may makes some issues with `'` character

### Why would it be useful?
If the value is multiline string with `\n` single quota character will not parse it as multiline, to have parsed multiline value env have to be in double quotas, i got issue with it when i have passed PRIVATE KEY from google recaptcha, i had to use this function `s.replace(/\\n/g, '\n')` (in nodejs) to get multiline parsed valid env, otherwise google recaptcha throws me an error

### Additional context
Everything is above :D

akhilmhdh:
@newadbie Yah there is a `skipMultiEncoding` to address these kinda issues but sadly its not yet available for cli. Good issue

sp6370:
Hi @akhilmhdh!
I would like to pick this up. 

Could you expand upon using `skipMultiEncoding`?

lalitkumawat1m:
Hello Team I want to work on this issue please assign to me

jasonjho:
Hi can this be fixed?

navaneeth-dev:
We also need an option to export secrets without single quotes, docker's --env-file passes single quotes to applications and causes errors.

I think this is an urgent feature, could I work on it?

guillemso1er:
I don't see how this is not a high priority, most software does work with dotenv files with single quotes surrounding values

maidul98:
Hi @guillemso1er, you should be able to format the output via https://infisical.com/docs/cli/commands/export#infisical-export:template

guillemso1er:
> Hi [@guillemso1er](https://github.com/guillemso1er), you should be able to format the output via https://infisical.com/docs/cli/commands/export#infisical-export:template

Hi @maidul98. I am aware, but is very cumbersome to create a file just for this, it since it does not support and in-line string on the command for the template parameter. When it should be a default behavior that the values come without the single quotes.

puneetdixit200:
Hi maintainers, I’d like to work on this CLI enhancement. I can add configurable quote character support for env export while preserving backward compatibility and tests. Please assign this issue to me if available.

Rakesh1002:
[...]

PR #1822, opened 2024-05-13
feat: delimiter flag for export command

nettijoe96:
# Description 📣

Creates new flags `delimit` and `quota` for the `export` command. `delimit` is currently only for CSV format. `quota` is only for .env formats. There are 2 flags rather than 1 because default handling is clearer and it allows for the possibility to extend delimit flag to .env for changing `=` in the future. Flags are permissive -- if format is not CSV or .env and a flag is provided, the flag is ignored and the command doesn't fail
 
Resolves: https://github.com/Infisical/infisical/issues/1103

## Type ✨

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

See unit tests in `cli/cmd_test.go`

Manually:

Screenshot 2024-05-13 at 3 14 21 PM

---

### DIFF
--- a/cli/packages/cmd/agent.go
+++ b/cli/packages/cmd/agent.go
@@ -351,7 +351,7 @@ func dynamicSecretTemplateFunction(accessToken string, dynamicSecretManager *Dyn
 	return func(args ...string) (map[string]interface{}, error) {
 		argLength := len(args)
 		if argLength != 4 && argLength != 5 {
-			return nil, fmt.Errorf("Invalid arguments found for dynamic-secret function. Check template %i", templateId)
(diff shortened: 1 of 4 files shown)


maidul98:
Hey @nettijoe96 custom formatting outisde of the predefined ones can be achieved with templates https://infisical.com/docs/cli/commands/export#infisical-export:template. I believe you can achieve what you added for this PR with templates. Let me know if you think otherwise.

nettijoe96:
> Hey @nettijoe96 custom formatting outisde of the predefined ones can be achieved with templates https://infisical.com/docs/cli/commands/export#infisical-export:template. I believe you can achieve what you added for this PR with templates. Let me know if you think otherwise.

Agreed!

PR #1825, opened 2024-05-14
Secret reference and integration sync support

akhilmhdh:
# Description 📣

1. This enable integrations to work with secret reference

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/services/secret/secret-queue.ts
```diff
@@ -102,28 +103,35 @@ export const secretQueueFactory = ({
...
       delay: 1000,
       backoff: {
         type: "exponential",
         delay: 3000
       },
       removeOnComplete: true,
-      removeOnFail: {
-        count: 5 // keep the most recent  jobs
-      }
+      removeOnFail: true
```
maidul98, line 115: it's sometimes helpful to see what the issues were because this is where they are recorded

backend/src/services/secret/secret-queue.ts
```diff
@@ -102,28 +103,35 @@ export const secretQueueFactory = ({
...
     });
   };
 
-  const syncSecrets = async (dto: TGetSecrets & { depth?: number }) => {
+  const syncSecrets = async ({
+    deDupeQueue = {},
+    ...dto
+  }: TGetSecrets & { depth?: number; deDupeQueue?: Record<string, boolean> }) => {
+    const deDuplicationKey = uniqueIntegrationKey(dto.environment, dto.secretPath);
```
maidul98, line 123: question: is this because we have many integrations and some can have the same scope so we filter them out here?

maidul98: or is this something else?

backend/src/services/secret/secret-queue.ts
```diff
@@ -349,21 +357,68 @@ export const secretQueueFactory = ({
...
+        );
         await Promise.all(
           imports
             .filter(({ folderId }) => Boolean(foldersGroupedById[folderId][0].path))
-            .map(({ folderId }) => {
-              const syncDto = {
+            // filter out already synced ones
+            .filter(
+              ({ folderId }) =>
+                !deDupeQueue[
```
maidul98, line 369: Would like to better understand this, maybe a huddle would be good

maidul98 (dismissed):
just some questions

backend/src/services/secret/secret-queue.ts
akhilmhdh: I thought our log already contains the fail state reason

backend/src/services/secret/secret-queue.ts
akhilmhdh: It's to avoid triggering sync secrets in the same stack. 

This happens when users have a cyclic secret reference or import.

maidul98 (approved):

PR #1827, opened 2024-05-15
Kubernetes Native Authentication Method

dangtony98:
# Description 📣

This PR adds the new Kubernetes authentication method to Identities.

With this method, applications in Kubernetes avoid secret zero by not needing to store an additional token to authenticate with Infisical. Instead, they can use their service account credential (JWT token) present in the pod under the `/var/run/secrets/kubernetes.io/serviceaccount/token` path to prove their identity to Infisical — this works by relaying the JWT token through Infisical to be cross-checked with the Kubernetes API server at the [TokenReview API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/); if successful and certain allow conditions are met, then the entity is considered authenticated and an access token granting access to the Infisical API is returned.

Workflow:

1. The operator sets up the Kubernetes authentication method in Infisical. As part of the setup, they specify what **Allowed Service Account Names** and **Allowed Namespaces** can be trusted. They also specify the Kubernetes Host / Base URL, CA Cert, and long-lived JWT token for Infisical to use to authenticate with the Kubernetes API server.
2. The operator programs the client application in Kubernetes to retrieve its service account credential (JWT token) at the `/var/run/secrets/kubernetes.io/serviceaccount/token` path and send it to the `/api/v1/auth/kubernetes-auth/login` endpoint containing the `identityId` that they are authenticating as and the `jwt` as params (in practice this would likely be done via SDK or other client but it can also be done manually).
4. Infisical forwards the params to the Kubernetes API server at the [TokenReview API](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/) for verification.
5. If successful, then Infisical checks the returned identity name and namespace properties against **Allowed Service Account Names** and **Allowed Namespaces** defined in step 1.
6. If successful, then Infisical returns an identity access token (similar to in Universal Auth).

Note 1: The Kubernetes authentication method will require additional engineering work for clients like Infisical Agent and SDKs to support it.

Note 2: The Kubernetes authentication method requires a connection from Infisical to a target Kubernetes API server to verify an application/pod.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


docs/documentation/platform/identities/kubernetes-auth.mdx
```diff
@@ -0,0 +1,233 @@
...
+description: "Learn how to authenticate with Infisical in Kubernetes"
+---
+
+**Kubernetes Auth** is a Kubernetes-native authentication method for applications (e.g. pods) to access Infisical.
+
+## Diagram
+
+    The following sequence digram illustrates the Kubernetes Auth workflow for authenticating applications running in pods with Infisical.
+
+```mermaid
```
maidul98, line 12: Does this render for you?
Screenshot 2024-05-17 at 9 30 27 PM

maidul98 (approved):
looks good to me, made some small changes in the docs. no hard requests

docs/documentation/platform/identities/kubernetes-auth.mdx
dangtony98: Yup it rendered for me.

backend/src/services/identity-kubernetes-auth/identity-kubernetes-auth-service.ts
```diff
@@ -0,0 +1,515 @@
...
+          allowedNames,
+          allowedAudience,
+          accessTokenMaxTTL,
+          accessTokenTTL,
+          accessTokenNumUsesLimit,
+          accessTokenTrustedIps: JSON.stringify(reformattedAccessTokenTrustedIps)
+        },
+        tx
+      );
+      await identityDAL.updateById(
```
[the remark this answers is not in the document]
dangtony98: This is a great suggestion

PR #1833, opened 2024-05-15
Azure Native Authentication Method

dangtony98:
# Description 📣

This PR adds the new Azure authentication method to Identities.

With this method, Azure resources like Azure VMs avoid secret zero by not needing to store an additional token to authenticate with Infisical. Instead, they can use their [managed identity access tokens ](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-to-use-vm-token) to prove their identity to Infisical — this works by relaying the managed identity access token through Infisical to be verified against the public key at the [Azure AD OpenID configuration endpoint](https://learn.microsoft.com/en-us/answers/questions/793793/azure-ad-validate-access-token); if successful and certain allow conditions are met, then the entity is considered authenticated and an access token granting access to the Infisical API is returned.

Workflow:

1. The operator sets up the Azure authentication method in Infisical. As part of the setup, they specify what **Service Principal IDs** can be trusted. They also specify the Tenant ID and (optional) Audience param included in the managed identity access tokens as part of the authentication method configuration.
2. The operator programs the client application in Azure to retrieve its [managed identity access tokens ](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-to-use-vm-token) and send it to the `/api/v1/auth/azure-auth/login` endpoint containing the `identityId` that they are authenticating as and the `jwt` as params (in practice this would likely be done via SDK or other client but it can also be done manually).
5. Infisical verifies the JWT token using the public key obtained from the [Azure AD OpenID configuration endpoint](https://learn.microsoft.com/en-us/answers/questions/793793/azure-ad-validate-access-token).
6. If successful, then Infisical checks the returned identity service principal ID (referred to as `oid` field on the JWT payload) against **Allowed Service Principal IDs** defined in step 1.
7. If successful, then Infisical returns an identity access token (similar to in Universal Auth).

Note: The Azure authentication method will require additional engineering work for clients like Infisical Agent and SDKs to support it.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


backend/src/server/routes/v1/identity-azure-auth-router.ts
```diff
@@ -0,0 +1,262 @@
...
+import { IdentityAzureAuthsSchema } from "@app/db/schemas";
+import { EventType } from "@app/ee/services/audit-log/audit-log-types";
+import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
+import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
+import { AuthMode } from "@app/services/auth/auth-type";
+import { TIdentityTrustedIp } from "@app/services/identity/identity-types";
+import { validateAzureAuthField } from "@app/services/identity-azure-auth/identity-azure-auth-validators";
+
+export const registerIdentityAzureAuthRouter = async (server: FastifyZodProvider) => {
+  server.route({
```
DanielHougaard, line 12: Isn't this endpoint missing an `onRequest` handler?

backend/src/services/identity-access-token/identity-access-token-dal.ts
```diff
@@ -39,11 +39,18 @@ export const identityAccessTokenDALFactory = (db: TDbClient) => {
...
+            `${TableName.Identity}.id`,
+            `${TableName.IdentityAzureAuth}.identityId`
+          );
+        })
         .select(selectAllTableCols(TableName.IdentityAccessToken))
         .select(
           db.ref("accessTokenTrustedIps").withSchema(TableName.IdentityUniversalAuth).as("accessTokenTrustedIpsUa"),
           db.ref("accessTokenTrustedIps").withSchema(TableName.IdentityGcpAuth).as("accessTokenTrustedIpsGcp"),
           db.ref("accessTokenTrustedIps").withSchema(TableName.IdentityAwsAuth).as("accessTokenTrustedIpsAws"),
+          db.ref("accessTokenTrustedIps").withSchema(TableName.IdentityAwsAuth).as("accessTokenTrustedIpsAzure"),
```
DanielHougaard, line 53: Breaking: You're reading from `IdentityAwsAuth` into `accessTokenTrustedIpsAzure`, should be `TableName.IdentityAzureAuth`

backend/src/services/identity-azure-auth/identity-azure-auth-service.ts
```diff
@@ -0,0 +1,286 @@
...
+import { validateAzureIdentity } from "./identity-azure-auth-fns";
+import {
+  TAttachAzureAuthDTO,
+  TGetAzureAuthDTO,
+  TLoginAzureAuthDTO,
+  TUpdateAzureAuthDTO
+} from "./identity-azure-auth-types";
+
+type TIdentityAzureAuthServiceFactoryDep = {
+  identityAzureAuthDAL: TIdentityAzureAuthDALFactory; // TODO: Pick
```
DanielHougaard, line 27: Pick missing here

DanielHougaard (commented):
Looks great, I've left out any nits and only commented on important things.

backend/src/services/identity-access-token/identity-access-token-dal.ts
dangtony98: Great catch. I'm fixing this in the latest push.

docs/documentation/platform/identities/machine-identities.mdx
```diff
@@ -7,7 +7,12 @@ description: "Learn how to use Machine Identities to programmatically interact w
 
 An Infisical machine identity is an entity that represents a workload or application that require access to various resources in Infisical. This is conceptually similar to an IAM user in AWS or service account in Google Cloud Platform (GCP).
 
+<<<<<<< HEAD
 Each identity must authenticate with the Infisical API using a supported authentication method like [Universal Auth](/documentation/platform/identities/universal-auth), [AWS Auth](/documentation/platform/identities/aws-auth), [Azure Auth](/documentation/platform/identities/azure-auth), or [GCP Auth](/documentation/platform/identities/gcp-auth) to get back a short-lived access token to be used in subsequent requests.
+=======
+Each identity must authenticate with the Infisical API using a supported authentication method like [Universal Auth](/documentation/platform/identities/universal-auth), [Kubernetes Auth](/documentation/platform/identities/kubernetes-auth), [AWS Auth](/documentation/platform/identities/aws-auth), or [GCP Auth](/documentation/platform/identities/gcp-auth) to get back a short-lived access token to be used in subsequent requests.
+
+> > > > > > > origin
```
DanielHougaard: Something seems to have gone wrong with the rebasing here

DanielHougaard (changes requested):
One last comment!

DanielHougaard (dismissed):

PR #1834, opened 2024-05-16
patch project identity update

maidul98:


backend/.eslintrc.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
     "@typescript-eslint/no-empty-function": "off",
     "@typescript-eslint/no-unsafe-enum-comparison": "off",
     "no-void": "off",
+    "no-await-in-loop": "off",
```
DanielHougaard, line 41: I wouldn't turn off this rule. Instead I would write for loops like this:

```js
for await (const item of items) {
   await something(item);
}
```

This won't give lint errors

backend/src/services/identity-project/identity-project-service.ts
```diff
@@ -135,16 +136,18 @@ export const identityProjectServiceFactory = ({
...
-      ActorType.IDENTITY,
-      projectIdentity.identityId,
-      projectIdentity.projectId,
-      actorAuthMethod,
-      actorOrgId
-    );
-    const hasRequiredPriviledges = isAtLeastAsPrivileged(permission, identityRolePermission);
-    if (!hasRequiredPriviledges)
-      throw new ForbiddenRequestError({ message: "Failed to delete more privileged identity" });
+    for (const { role: requestedRoleChange } of roles) {
```
DanielHougaard, line 139: ```for await (const { role: requestedRoleChange } of roles)```

DanielHougaard (changes requested):
Looks good, requested a small change

backend/.eslintrc.js
maidul98: sweet, let me change it

DanielHougaard (approved):

PR #1838, opened 2024-05-16
fix: get all secrets from aws ssm

akhilmhdh:
# Description 📣

1. By default ssm get secrets paginate by 10
2. Now it fetches everything from it

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/integration-auth/integration-sync-secret.ts
+++ b/backend/src/services/integration-auth/integration-sync-secret.ts
@@ -479,10 +479,10 @@ const syncSecretsAWSParameterStore = async ({
       .promise();
 
     if (parameters.Parameters) {
-      parameters.Parameters.forEach((sec) => {
-        if (sec.Name) {
-          const secKey = sec.Name.substring((integration.path as string).length);
-          awsParameterStoreSecretsObj[secKey] = sec;
+      parameters.Parameters.forEach((parameter) => {
+        if (parameter.Name) {
+          const secKey = parameter.Name.substring((integration.path as string).length);
+          awsParameterStoreSecretsObj[secKey] = parameter;
         }
       });
     }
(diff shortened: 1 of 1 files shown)


backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -462,79 +462,86 @@ const syncSecretsAWSParameterStore = async ({
...
 
-  const params = {
-    Path: integration.path as string,
-    Recursive: false,
-    WithDecryption: true
-  };
-
-  const parameterList = (await ssm.getParametersByPath(params).promise()).Parameters;
+    if (parameters.Parameters) {
+      parameters.Parameters.forEach((sec) => {
```
maidul98: nit: full names would be better for reability. sec => secret

backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -462,79 +462,86 @@ const syncSecretsAWSParameterStore = async ({
   ssm.config.update(config);
 
   const metadata = z.record(z.any()).parse(integration.metadata || {});
+  const awsParameterStoreSecretsObj: Record<string, AWS.SSM.Parameter> = {};
+
+  // now fetch all aws parameter store secrets
+  let hasNext = true;
+  let nextToken: string | undefined;
+  while (hasNext) {
```
maidul98, line 470: i think we should add some sleep in between each loop so it doesn't rate limit us? what do you think

backend/src/services/integration-auth/integration-sync-secret.ts
maidul98: i  made the change

maidul98 (dismissed):

maidul98 (approved):

PR #1848, opened 2024-05-20
feat: added integration sync status

sheensantoscapadngan:
# Description 📣
- Add integration sync status

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/db/migrations/20240520064127_add-integration-sync-status.ts
+++ b/backend/src/db/migrations/20240520064127_add-integration-sync-status.ts
@@ -1,0 +1,43 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasIsSyncedColumn = await knex.schema.hasColumn(TableName.Integration, "isSynced");
+  const hasSyncMessageColumn = await knex.schema.hasColumn(TableName.Integration, "syncMessage");
+  const hasLastSyncJobId = await knex.schema.hasColumn(TableName.Integration, "lastSyncJobId");
+
+  await knex.schema.alterTable(TableName.Integration, (t) => {
+    if (!hasIsSyncedColumn) {
+      t.boolean("isSynced").nullable();
+    }
+
+    if (!hasSyncMessageColumn) {
+      t.text("syncMessage").nullable();
+    }
+
+    if (!hasLastSyncJobId) {
+      t.string("lastSyncJobId").nullable();
+    }
+  });
+}
+
+export async function down(knex: Knex): Promise<void> {
+  const hasIsSyncedColumn = await knex.schema.hasColumn(TableName.Integration, "isSynced");
+  const hasSyncMessageColumn = await knex.schema.hasColumn(TableName.Integration, "syncMessage");
+  const hasLastSyncJobId = await knex.schema.hasColumn(TableName.Integration, "lastSyncJobId");
+
+  await knex.schema.alterTable(TableName.Integration, (t) => {
+    if (hasIsSyncedColumn) {
+      t.dropColumn("isSynced");
+    }
+
+    if (hasSyncMessageColumn) {
+      t.dropColumn("syncMessage");
+    }
+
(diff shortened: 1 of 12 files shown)


akhilmhdh:
@sheensantoscapadngan Let's make it into a tag. That `sync status`. Where red color for failed and green on success.

Now when you hover over it. For success it will show last sync status, for failed it will show last status status and reason for failing too. 

Keep it in like 2x2 table with first column containing the title and second column containing data.

frontend/src/hooks/api/integrations/queries.tsx
```diff
@@ -110,3 +111,28 @@ export const useDeleteIntegration = () => {
     }
   });
 };
+
+export const useSyncIntegration = (pollingRef: MutableRefObject<NodeJS.Timeout | null>) => {
```
akhilmhdh, line 115: You can keep the pollingRef timer inside the function and the timer cleanup useEffect also here. 

Then return both the mutation and `isSyncing` variable to show the syncing status thus outside the function nobody needs to care about the timer

frontend/src/hooks/api/integrations/queries.tsx
```diff
@@ -110,3 +111,28 @@ export const useDeleteIntegration = () => {
...
+  const queryClient = useQueryClient();
+
+  return useMutation<{}, {}, { id: string; workspaceId: string; lastUsed: string }>({
+    mutationFn: ({ id }) => apiRequest.post(`/api/v1/integration/${id}/sync`),
```
akhilmhdh, line 119: Its better to remove this logic and use the `refreshInterval` option of react-query for the `get workspace integrations` hook and do a polling for 3.5s automatically. 

The issue with below approach is what if someone else triggers the integration sync in mean time thus changing the last used to something else. Thus the spinner will keep spinning. And i can see more edge cases with this. 

So its easier if we can just add the get integration hook as a polling one so even if anyone. else make changes it will also be seen

backend/src/db/migrations/20240520064127_add-integration-sync-status.ts
```diff
@@ -0,0 +1,43 @@
...
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasIsSyncedColumn = await knex.schema.hasColumn(TableName.Integration, "isSynced");
+  const hasSyncMessageColumn = await knex.schema.hasColumn(TableName.Integration, "syncMessage");
+  const hasLastSyncJobId = await knex.schema.hasColumn(TableName.Integration, "lastSyncJobId");
+
+  await knex.schema.alterTable(TableName.Integration, (t) => {
+    if (!hasIsSyncedColumn) {
+      t.boolean("isSynced").nullable();
```
akhilmhdh, line 12: I see this column has never used again. Is this expected?

akhilmhdh (changes requested):

backend/src/db/migrations/20240520064127_add-integration-sync-status.ts
sheensantoscapadngan: we're using this in the frontend. backend returns this property as part of getById of integration
image

frontend/src/hooks/api/integrations/queries.tsx
sheensantoscapadngan: agree. If we do it this way, what should the user-feedback be upon clicking the Manual sync button? should we just display a notification that says "Successfully triggered manual sync" instead?

akhilmhdh: Yah that would be fine.

backend/src/db/migrations/20240520064127_add-integration-sync-status.ts
akhilmhdh: My bad. Didn't find any reference when i did search for it. My mistake.

akhilmhdh (approved):

PR #1852, opened 2024-05-20
rate limit based on identity

maidul98:
In the past we rate limited based on IP but this was highly inaccurate and subject to tampering. This PR improves rate limits by linking limits to the identity that is making the call. This can be service tokens, machine identity, API keys, or dashboard users. For APIs that do not require auth, IP address will be used.

### DIFF
--- a/backend/src/server/routes/index.ts
+++ b/backend/src/server/routes/index.ts
@@ -1,3 +1,4 @@
+import ratelimiter, { FastifyRateLimitOptions } from "@fastify/rate-limit";
 import { Knex } from "knex";
 import { z } from "zod";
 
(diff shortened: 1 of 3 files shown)


backend/src/server/routes/index.ts
```diff
@@ -839,6 +840,11 @@ export const registerRoutes = async (
     user: userDAL
   });
 
+  // Rate limiters and security headers
+  if (appCfg.isProductionMode) {
+    await server.register<FastifyRateLimitOptions>(ratelimiter, globalRateLimiterCfg());
```
DanielHougaard, line 845: How come we decided to move this check?

DanielHougaard (changes requested):

backend/src/server/routes/index.ts
maidul98: Because before the rate limit was configured before we parsed the auth methods and thier types. Now, we parse the request for auth type before we try to rate limit it

DanielHougaard: This makes sense

DanielHougaard (approved):

PR #1853, opened 2024-05-21
Docs: Jenkins Plugin

DanielHougaard:
# Description 📣

As we now have a native integration for Jenkins, these docs has to be updated in order to reflect the correct approach to using Infisical with Jenkins. 

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

---


vmatsiiako (approved):
approving but maybe we should still include the docs for CLI with Machine Identities?

PR #1837, opened 2024-05-16
Daily cron for cleaning up expired tokens from db

akhilmhdh:
# Description 📣

1. Moved audit log pruning to resource clean up queue
2. Added token clean up logic
3. Fixed identity token issues found
4. Resolved client ip failing when using identity token when header is comma seperate ip strings

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
# Here's some code block to paste some code snippets
```

---

- [x] I have read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/getting-started/code-of-conduct). 📝

PR #1861, opened 2024-05-22
feat: secure mfa endpoints with improved rate limiting and account locking

sheensantoscapadngan:
# Description 📣
- added progressive delay as well as user account locking after failed MFA attempts
- users should be able to unlock their accounts via email interaction
- when account is locked user can also remove lock by resetting password
- finalized rate limiting for mfa endpoints

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/auth/auth-fns.ts
+++ b/backend/src/services/auth/auth-fns.ts
@@ -56,12 +56,15 @@ export const enforceUserLockStatus = (isLocked: boolean, temporaryLockDateEnd?: 
 
   if (temporaryLockDateEnd) {
     const timeDiff = new Date().getTime() - temporaryLockDateEnd.getTime();
-    if (timeDiff < 0)
+    if (timeDiff < 0) {
+      const secondsDiff = (-1 * timeDiff) / 1000;
+      const timeDisplay =
+        secondsDiff > 60 ? `${Math.ceil(secondsDiff / 60)} minutes` : `${Math.ceil(secondsDiff)} seconds`;
+
       throw new UnauthorizedError({
         name: "User Locked",
-        message: `User is locked due to multiple failed login attempts. Try again after ${Math.round(
-          (-1 * timeDiff) / 1000
-        )} seconds.`
+        message: `User is temporary locked due to multiple failed login attempts. Try again after ${timeDisplay}. You can also reset your password now to proceed.`
       });
+    }
   }
 };


backend/src/server/config/rateLimiter.ts
```diff
@@ -52,6 +52,15 @@ export const inviteUserRateLimit: RateLimitOptions = {
   keyGenerator: (req) => req.realIp
 };
 
+export const mfaRateLimit: RateLimitOptions = {
+  timeWindow: 60 * 1000,
+  max: 20,
+  keyGenerator: (req) => {
+    // fallback to IP to avoid global rate limiting when authorization is set to empty
+    return req.headers.authorization?.split(" ")[1] || req.realIp;
```
sheensantoscapadngan, line 60: ah hm.. will decode this and filter it based on the userId specifically

sheensantoscapadngan: I can't use the `req.mfa` here because for some reason, the rate limiter gets executed first

backend/src/db/schemas/users.ts
```diff
@@ -22,9 +22,15 @@ export const UsersSchema = z.object({
...
+  isEmailVerified: z.boolean().default(false).nullable().optional(),
+  consecutiveFailedMfaAttempts: z.number(),
+  isLocked: z.boolean(),
+  temporaryLockDateEnd: z.date().nullable().optional()
 });
 
 export type TUsers = z.infer<typeof UsersSchema>;
-export type TUsersInsert = Omit<z.input<typeof UsersSchema>, TImmutableDBKeys>;
+export type TUsersInsert = Omit<
+  z.input<typeof UsersSchema>,
```
akhilmhdh, line 33: The isLocked and consecutiveFailedMfaAttempts will get overwritten in next schema generate. Its better to remove it

backend/src/server/routes/v1/user-router.ts
```diff
@@ -25,4 +29,29 @@ export const registerUserRouter = async (server: FastifyZodProvider) => {
       return { user };
     }
   });
+
+  server.route({
+    method: "GET",
+    url: "/:userId/unlock-verify",
```
akhilmhdh, line 35: Suggestion: I would prefer just unlock

backend/src/services/auth/auth-fns.ts
```diff
@@ -44,3 +44,24 @@ export const validateSignUpAuthorization = (token: string, userId: string, valid
   if (decodedToken.authTokenType !== AuthTokenType.SIGNUP_TOKEN) throw new UnauthorizedError();
   if (decodedToken.userId !== userId) throw new UnauthorizedError();
 };
+
+export const enforceUserLockStatus = (isLocked: boolean, temporaryLockDateEnd?: Date | null) => {
```
akhilmhdh, line 48: What if the email is not send or for some reason failed. Is there a way for user to re-send it again?

frontend/src/views/Login/components/MFAStep/MFAStep.tsx
```diff
@@ -178,20 +165,31 @@ export const MFAStep = ({ email, password, providerAuthToken }: Props) => {
...
-        window.location.reload();
-      } else if (error?.response?.data?.context?.triesLeft) {
-        setTriesLeft(error?.response?.data?.context?.triesLeft);
-        if (error.response.data.context.triesLeft === 0) {
-          window.location.reload();
-        }
+      if (triesLeft) {
+        setTriesLeft((left) => {
+          if (triesLeft === 1) {
+            window.location.reload();
```
akhilmhdh, line 187: Reload seems to be a very bad weird ux. I think we can try something else right?

backend/src/services/user/user-fns.ts
```diff
@@ -19,3 +21,54 @@ export const normalizeUsername = async (username: string, userDAL: Pick<TUserDAL
     }
   }
 };
+
+export const processFailedMfaAttempt = async (userId: string, userDAL: Pick<TUserDALFactory, "transaction">) => {
+  try {
+    const updatedUser = await userDAL.transaction(async (tx) => {
+      const PROGRESSIVE_DELAY_INTERVAL = 3;
+      const [user] = await tx(TableName.Users)
```
akhilmhdh, line 29: Kindly use dal layer and not knex functions directly. To do increment, the update operation orm has support for it.

https://github.com/Infisical/infisical/pull/1837 but i think its still in review

backend/src/services/user/user-fns.ts
```diff
@@ -19,3 +21,54 @@ export const normalizeUsername = async (username: string, userDAL: Pick<TUserDAL
...
+      if (!user) {
+        throw new Error("User not found");
+      }
+
+      const progressiveDelaysInMins = [5, 30, 60];
+
+      // lock user when failed attempt exceeds threshold
+      if (user.consecutiveFailedMfaAttempts >= PROGRESSIVE_DELAY_INTERVAL * (progressiveDelaysInMins.length + 1)) {
+        return (
+          await tx(TableName.Users)
```
akhilmhdh, line 43: same here no direct knex function other than dal layer

backend/src/services/user/user-fns.ts
```diff
@@ -19,3 +21,54 @@ export const normalizeUsername = async (username: string, userDAL: Pick<TUserDAL
...
+            .update({
+              isLocked: true,
+              temporaryLockDateEnd: null
+            })
+            .returning("*")
+        )[0];
+      }
+
+      // delay user only when failed MFA attempts is a multiple of configured delay interval
+      if (user.consecutiveFailedMfaAttempts % PROGRESSIVE_DELAY_INTERVAL === 0) {
```
akhilmhdh, line 54: same issue here

backend/src/services/user/user-fns.ts
```diff
@@ -19,3 +21,54 @@ export const normalizeUsername = async (username: string, userDAL: Pick<TUserDAL
     }
   }
 };
+
+export const processFailedMfaAttempt = async (userId: string, userDAL: Pick<TUserDALFactory, "transaction">) => {
```
akhilmhdh, line 25: Also i think as this logic is just used in the service layer you can keep it there itself.

backend/src/services/auth/auth-login-service.ts
```diff
@@ -212,6 +213,9 @@ export const authLoginServiceFactory = ({
     });
     // send multi factor auth token if they it enabled
     if (userEnc.isMfaEnabled && userEnc.email) {
+      const user = await userDAL.findById(userEnc.userId);
```
akhilmhdh, line 216: I think you missed when a locked user is trying to do login. It should fail even before reaching the mfa state

akhilmhdh (changes requested):

akhilmhdh (changes requested):
Some doubts

backend/src/services/auth/auth-fns.ts
sheensantoscapadngan: none. In this case (which currently is configured to happen only after the 12th failed MFA attempt), users will have to contact support or their administrators if they don't receive any email

sheensantoscapadngan: Discussed this with @maidul98, we'll be allowing account unlock via password reset as well 😄 
image

cc: @dangtony98

backend/src/services/auth/auth-login-service.ts
sheensantoscapadngan: this one actually already errors out for the user upon clicking "Continue with Email"
image

backend/src/services/user/user-fns.ts
sheensantoscapadngan: for now, I created a DAL method for this 
image

frontend/src/views/Login/components/MFAStep/MFAStep.tsx
sheensantoscapadngan: hmmm. the intention here is to bring back the user to the main login screen upon 3 failing attempts to authenticate with MFA

backend/src/services/user/user-fns.ts
sheensantoscapadngan: done @maidul98 !

frontend/src/views/Login/components/MFAStep/MFAStep.tsx
akhilmhdh: Use nextjs router for that. Reloading is not a good experience

sheensantoscapadngan: resolved ^

akhilmhdh (approved):

PR #1865, opened 2024-05-23
feat: added one to one support for aws secret manager integration

sheensantoscapadngan:
# Description 📣
- add support for one to one mapping in AWS SM

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -570,134 +575,145 @@ const syncSecretsAWSSecretManager = async ({
...
+      secretAccessKey: accessToken
     }
+  });
 
-    if (!isEqual(awsSecretManagerSecretObj, secKeyVal)) {
-      await secretsManager.send(
-        new UpdateSecretCommand({
-          SecretId: integration.app as string,
-          SecretString: JSON.stringify(secKeyVal)
+  const processAwsSecret = async (secretId: string, keyValuePairs: Record<string, string | null | undefined>) => {
```
maidul98, line 590: question: this is the aws secretId right? because i'm also seeing integration.app being passed here too so curious

backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -570,134 +575,145 @@ const syncSecretsAWSSecretManager = async ({
...
-          Tags: metadata.secretAWSTag
-            ? metadata.secretAWSTag.map((tag: { key: string; value: string }) => ({ Key: tag.key, Value: tag.value }))
-            : []
-        })
-      );
+  };
+
+  if (metadata.mappingBehavior === IntegrationMappingBehavior.ONE_TO_ONE) {
+    for await (const [key, value] of Object.entries(secrets)) {
+      await processAwsSecret(key, {
```
maidul98, line 711: So you think this APi will get rate limited if we call it very quickly one after another for each secret? Imagine there is 200 secrets?

maidul98 (commented):
Left some commnets

backend/src/server/routes/v1/integration-router.ts
```diff
@@ -49,6 +49,7 @@ export const registerIntegrationRouter = async (server: FastifyZodProvider) => {
             secretPrefix: z.string().optional().describe(INTEGRATION.CREATE.metadata.secretPrefix),
             secretSuffix: z.string().optional().describe(INTEGRATION.CREATE.metadata.secretSuffix),
             initialSyncBehavior: z.string().optional().describe(INTEGRATION.CREATE.metadata.initialSyncBehavoir),
+            mappingBehavior: z.string().optional().describe(INTEGRATION.CREATE.metadata.mappingBehavior),
```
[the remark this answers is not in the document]
sheensantoscapadngan: oh yea nice catch! will edit accordingly

backend/src/services/integration-auth/integration-sync-secret.ts
sheensantoscapadngan: yea, so if the integration is set to "many to one" then we're actually saving the target secret name inside `integration.app`. this is the previous implementation.

With one to one, we don't need to use the value saved in the integration.app field because the secret keys serve as the AWS secret name

backend/src/services/integration-auth/integration-sync-secret.ts
sheensantoscapadngan: yup, just erring on the side of caution here. I read a discussion with Akhi on slack regarding AWS API rate limits

maidul98 (approved):
lgtm!

PR #1874, opened 2024-05-26
Updates api endpoints for project role and identity specfic privilege

akhilmhdh:
# Description 📣

1. Improved and added doc for custom rbac api endpoints for machine identity support
2. Added a new input field for identity specific privilege for permissions
3. The old is depreciated as the new one is better suited for external usage, the other one is not aligned with the ui or expectations as from customer feedback
4. Corresponding changes in UI
5. The RBAC is breaking change - This endpoints was only for dashboard so its oke for it to break

## Type ✨

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/project-role/project-role-service.ts
+++ b/backend/src/services/project-role/project-role-service.ts
@@ -1,25 +1,30 @@
-import { ForbiddenError } from "@casl/ability";
-import { packRules } from "@casl/ability/extra";
+import { ForbiddenError, MongoAbility, RawRuleOf } from "@casl/ability";
+import { PackRule, packRules, unpackRules } from "@casl/ability/extra";
 
-import { ProjectMembershipRole, TOrgRolesUpdate, TProjectRolesInsert } from "@app/db/schemas";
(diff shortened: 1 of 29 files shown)


backend/src/services/project-role/project-role-service.ts
```diff
@@ -125,16 +205,14 @@ export const projectRoleServiceFactory = ({
...
   };
 
-  const listRoles = async (
-    actor: ActorType,
-    actorId: string,
-    projectId: string,
-    actorAuthMethod: ActorAuthMethod,
-    actorOrgId: string | undefined
-  ) => {
+  const listRoles = async ({ projectSlug, actorOrgId, actorAuthMethod, actorId, actor }: TListRolesDTO) => {
```
maidul98, line 211: nice, i like how you updated them to be projectSlug instead of project id now before making them public

maidul98 (approved):
saw you added things by project slug, nice!

PR #1884, opened 2024-05-28
fix: resolved trailing slash issue with additional privileges

sheensantoscapadngan:
# Description 📣
Previously, fetching secrets for users with additional privilege does not work. Upon investigation, this is because we're saving the path condition with no sanitation or pre-processing (in this case, removal of trailing slash)

Admin view of level1 path:
image

Permission of user1:
image

View of user1 in level1 path
image

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/helpers/string.ts
+++ b/frontend/src/helpers/string.ts
@@ -1,0 +1,5 @@
+export const removeTrailingSlash = (str: string) => {
+  if (str === "/") return str;
+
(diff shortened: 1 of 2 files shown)


akhilmhdh (commented):
@sheensantoscapadngan instead of doing it in ui level. Do it on API level.

akhilmhdh (approved):

PR #1887, opened 2024-05-28
feat: added support for personal secrets creation via CLI

sheensantoscapadngan:
# Description 📣
- With this PR, devs can now create personal secrets using the CLI 
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/cli/packages/cmd/secrets.go
+++ b/cli/packages/cmd/secrets.go
@@ -170,6 +170,11 @@ var secretsSetCmd = &cobra.Command{
 			util.HandleError(err, "Unable to get your local config details")
 		}
 
+		secretType, err := cmd.Flags().GetString("type")
+		if err != nil || (secretType != util.SECRET_TYPE_SHARED && secretType != util.SECRET_TYPE_PERSONAL) {
+			util.HandleError(err, "Unable to parse secret type")
+		}
+
 		loggedInUserDetails, err := util.GetCurrentLoggedInUserDetails()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


DanielHougaard (changes requested):
From testing it seems to not care that I pass in `--type=shared`. 

CleanShot 2024-05-28 at 22 57 39

CleanShot 2024-05-28 at 22 58 25

cli/packages/cmd/secrets.go
```diff
@@ -251,7 +266,16 @@ var secretsSetCmd = &cobra.Command{
...
 
-			if existingSecret, ok := secretByKey[key]; ok {
+			var existingSecret models.SingleEnvironmentVariable
+			var doesSecretExist bool
+
+			if secretType == util.SECRET_TYPE_SHARED {
+				existingSecret, doesSecretExist = sharedSecretMapByName[key]
+			} else {
+				existingSecret, doesSecretExist = personalSecretMapByName[key]
+			}
```
DanielHougaard, line 276: I would probably do an `else if secret.Type == util.SECRET_TYPE_SHARED`. And then and `else` statement that throws an error if the secretType is not `shared` or `personal`, because it's possible that the user has unknowingly made a typo when they're typing the command.

DanielHougaard (changes requested):

DanielHougaard (approved):

PR #1889, opened 2024-05-28
fix: added account locked prompt to sso signin flow

sheensantoscapadngan:
# Description 📣
- Previously, the SSO login page won't display the "User locked error prompt". This problem is fixed in this PR

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/views/Login/components/PasswordStep/PasswordStep.tsx
+++ b/frontend/src/views/Login/components/PasswordStep/PasswordStep.tsx
@@ -31,7 +31,6 @@ export const PasswordStep = ({
   setPassword,
   setStep
 }: Props) => {
-  
   const [isLoading, setIsLoading] = useState(false);
   const { t } = useTranslation();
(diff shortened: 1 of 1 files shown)


frontend/src/views/Login/components/PasswordStep/PasswordStep.tsx
```diff
@@ -146,13 +145,24 @@ export const PasswordStep = ({
           }
         }
       }
-    } catch (err) {
+    } catch (err: any) {
       setIsLoading(false);
+      console.error(err);
+
+      if (err.response.data.error === "User Locked") {
```
maidul98, line 152: shouldn't we show the message anyways? Even if it is not "user locked" error?

maidul98 (approved):

PR #1898, opened 2024-05-29
fix: remove ghost users being added as seats in stripe

ShubhamPalriwala:
# Description 📣
Currently on Stripe we add all users (including Ghost users) which makes the pricing inconsistent with the actual seats. This PR adds a filter to ignore the ghost users in this counting.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
- Create a project
- invite a couple of teammates to the org
- Now see the `users` table and see 3 of your non-ghost users & each project's 1 ghost user
- Now call the `licenseDALFactory.countOfOrgMembers()` method which is used in billing as well and you'll see now only non-ghost user count is reported

---

### DIFF
--- a/backend/src/ee/services/license/license-dal.ts
+++ b/backend/src/ee/services/license/license-dal.ts
@@ -16,6 +16,8 @@ export const licenseDALFactory = (db: TDbClient) => {
             void bd.where({ orgId });
           }
         })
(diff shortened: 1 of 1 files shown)


maidul98 (approved):
lgtm

PR #1905, opened 2024-05-30
feat: secret sharing supports expiry on view count

ShubhamPalriwala:
# Description 📣
Secret Sharing Feature has the following changes:
- Got rid of Name (Key Value pairs are removed)
- Multi line input to share anything across securely
- Expire secrets either based on time or on number of views
- Docs open in a new tab
- Updated docs as per new options including new org level placement
- Public shared page shows left time/views based on the secret
- And some Prettier linting

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---


backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -43,9 +54,18 @@ export const secretSharingServiceFactory = ({
...
     }
+    if (sharedSecret.expiresAfterViews != null && sharedSecret.expiresAfterViews >= 0) {
+      if (sharedSecret.expiresAfterViews === 0) {
+        await secretSharingDAL.deleteById(sharedSecretId);
+        return;
+      }
+      await secretSharingDAL.updateById(sharedSecretId, {
+        expiresAfterViews: sharedSecret.expiresAfterViews - 1
+      });
+    }
```
sheensantoscapadngan, line 68: use $decr

docs/documentation/platform/secret-sharing.mdx
```diff
@@ -1,37 +1,42 @@
 ---
 title: "Secret Sharing"
 sidebarTitle: "Secret Sharing"
-description: "Learn how to share time-bound secrets securely with anyone on the internet."
+description: "Learn how to share time or view-count bound secrets securely with anyone on the internet."
```
sheensantoscapadngan, line 4: I think users should be able to use both restrictions?

frontend/src/views/ShareSecretPublicPage/ShareSecretPublicPage.tsx
```diff
@@ -46,18 +50,26 @@ export const ShareSecretPublicPage = () => {
...
+              (timeDifference % millisecondsPerDay) / millisecondsPerHour
+            );
+            const minutesRemaining = Math.floor(
+              (timeDifference % millisecondsPerHour) / millisecondsPerMinute
+            );
+            const secondsRemaining = Math.floor((timeDifference % millisecondsPerMinute) / 1000);
+            setTimeLeft(`${hoursRemaining}h ${minutesRemaining}m ${secondsRemaining}s`);
+          }
+        } else if (data.expiresAfterViews) {
+          setTimeLeft(`${data.expiresAfterViews - 1} more views`);
```
sheensantoscapadngan, line 72: hmm.. I think we don't need to show this prompt to the public user? it's better that we limit the stuff they know about the shared secret

frontend/src/views/ShareSecretPublicPage/ShareSecretPublicPage.tsx
```diff
@@ -46,18 +50,26 @@ export const ShareSecretPublicPage = () => {
...
+            setTimeLeft("Expired");
+          } else {
+            const hoursRemaining = Math.floor(
+              (timeDifference % millisecondsPerDay) / millisecondsPerHour
+            );
+            const minutesRemaining = Math.floor(
+              (timeDifference % millisecondsPerHour) / millisecondsPerMinute
+            );
+            const secondsRemaining = Math.floor((timeDifference % millisecondsPerMinute) / 1000);
+            setTimeLeft(`${hoursRemaining}h ${minutesRemaining}m ${secondsRemaining}s`);
```
sheensantoscapadngan, line 69: here as well

sheensantoscapadngan (commented):
did we already configure a cleanup job to delete expired shared secrets? or is that something that we don't want to do?

backend/src/services/secret-sharing/secret-sharing-service.ts
ShubhamPalriwala: thanks, missed this.

docs/documentation/platform/secret-sharing.mdx
ShubhamPalriwala: I found it weird to have to setup both hence felt the UX might be better off to just let them choose the option and set only that.

frontend/src/views/ShareSecretPublicPage/ShareSecretPublicPage.tsx
ShubhamPalriwala: aye nice idea! done for both 💪🏼

ShubhamPalriwala:
> did we already configure a cleanup job to delete expired shared secrets? or is that something that we don't want to do?

Yepz, we have a cleanup cronjob for time based expiries! And for view based, it happens on the last view

docs/documentation/platform/secret-sharing.mdx
sheensantoscapadngan: For reference, bitwarden does it like the following

sheensantoscapadngan (approved):

maidul98 (approved):

PR #1916, opened 2024-06-03
feat: add user identities & machine identities to stripe billing

ShubhamPalriwala:
# Description 📣
- Adds versioning in Organization schema as `billingVersion` where
    - `v0` is the current developer based pricing where only users are charged
    - `v1` is the new pricing where identities (users & machines) are charged
- Combined limit of 5 for both users as well as machine identities
- Updates & Validates identity limit check on creation & deletion of users as well as machine identities

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
Run with the license server locally and test the new limits + combined pricing

```sh
```

---


dangtony98:
Closed because the desired implementation should be such that the Infisical application itself is unaware of any pricing scheme / versioning-related information.

This logic should be fully centralized on the side of Stripe / license server so that, for example, if we needed to know who is on the so-called billing version `v2` then we can obtain that information in Stripe instead of querying the DB here.

PR #1924, opened 2024-06-05
feat: added setup for production observability (metrics via OTEL)

sheensantoscapadngan:
# Description 📣
This PR adds instrumentation with OpenTelemetry with support for exporting to prometheus or to an OTLP endpoint.

The dev setup includes the following components:
- Open Telemetry (for instrumentation and gathering of application traces and metrics)
- Open Telemetry collector (for batching of telemetry data from application servers and for exposing data for long-term storage)
- Prometheus (for storage of metrics in time-series format)
- Grafana (for visualization of time-series data from Prometheus)

Documentation update:
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---


.env.example
```diff
@@ -63,3 +63,6 @@ CLIENT_SECRET_GITHUB_LOGIN=
 
 CLIENT_ID_GITLAB_LOGIN=
 CLIENT_SECRET_GITLAB_LOGIN=
+
+OTEL_COLLECTOR_OTLP_URL=
+OTEL_TELEMETRY_COLLECTION_ENABLED=
```
maidul98, line 68: Is there a reason we explicitly want this enabled/disabled env? What if we determined if it is enabled by checking if  OTEL_COLLECTOR_OTLP_URL is set?

sheensantoscapadngan: so currently I added support for both OTLP export and prometheus-native export. If OTEL_TELEMETRY_COLLECTION_ENABLED is set to true but OTEL_COLLECTOR_OTLP_URL is not defined, it will default to the prometheus export else it uses OTLP

sheensantoscapadngan: it's better that we have support for multiple export methods so that we can test out which one is the most ideal in our setup (and for those who want to self-host, they can choose what's ideal for them as well)

maidul98: okay, i understand. 

nit: I would rename `OTEL_COLLECTOR_OTLP_URL` to `OTEL_EXPORTER_OTLP_ENDPOINT`

Instead defaulting to promethus, I think users should intentionally set the export type. What do you think about setting `OTEL_EXPORT_TYPE=otlp` `OTEL_EXPORT_TYPE=prometheus` and we do a switch over this? This way, if we want to add other export type, we can easily do that in the future too?

backend/src/lib/telemetry/instrumentation.ts
```diff
@@ -0,0 +1,46 @@
...
+    new Resource({
+      [SEMRESATTRS_SERVICE_NAME]: "infisical-server",
+      [SEMRESATTRS_SERVICE_VERSION]: "0.1.0"
+    })
+  );
+
+  const metricReaders = [];
+  if (otlpURL) {
+    const otlpExporter = new OTLPMetricExporter({
+      url: `${otlpURL}/v1/metrics`,
```
maidul98, line 21: I see that we only added the endpoint but i assume most of these end points are secured by bear tokens or other header options? Maybe we should add at least a header option such as 

```
export OTEL_EXPORTER_OTLP_HEADERS="Authorization=Bearer an_apm_secret_token"
```

maidul98: Also question, is it always `/v1/metrics`?

backend/src/server/plugins/api-metrics.ts
```diff
@@ -0,0 +1,21 @@
+import opentelemetry from "@opentelemetry/api";
+import fp from "fastify-plugin";
+
+export const apiMetrics = fp(async (fastify) => {
+  const apiMeter = opentelemetry.metrics.getMeter("API");
+  const latencyHistogram = apiMeter.createHistogram("API latency", {
```
maidul98, line 6: Curious: what is the impact of such metric collection on the API? Does it increase our API response times much?

backend/src/main.ts
```diff
@@ -13,6 +14,11 @@ dotenv.config();
 const run = async () => {
   const logger = await initLogger();
   const appCfg = initEnvConfig(logger);
+
+  if (appCfg.OTEL_TELEMETRY_COLLECTION_ENABLED) {
+    await initTelemetry({ otlpURL: appCfg.OTEL_COLLECTOR_OTLP_URL });
```
maidul98, line 19: nit: rename to initTelemetryInstrumentation

docker-compose.dev.observable.yml
```diff
@@ -0,0 +1,201 @@
...
+        condition: service_started
+      redis:
+        condition: service_started
+      db-migration:
+        condition: service_completed_successfully
+    env_file:
+      - .env
+    ports:
+      - 4000:4000
+      - 9464:9464 # for OTEL collection of Prometheus metrics
```
maidul98, line 88: question: so if we don't use the OTEL protocol for export and only use Prometheus then do we still need a otel collector?

maidul98 (commented):
Good setup, left some questions and comments

.env.example
sheensantoscapadngan: yea that makes more sense, will adjust accordingly

backend/src/lib/telemetry/instrumentation.ts
sheensantoscapadngan: yup! `/v1/metrics` for metrics. 

it's defined in the OTEL specs

backend/src/lib/telemetry/instrumentation.ts
```diff
@@ -0,0 +1,91 @@
...
+  otlpURL?: string;
+  otlpUser?: string;
+  otlpPassword?: string;
+  otlpPushInterval?: number;
+}) => {
+  diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
+
+  const resource = Resource.default().merge(
+    new Resource({
+      [ATTR_SERVICE_NAME]: "infisical-server",
```
maidul98, line 32: infisical-core

backend/src/lib/telemetry/instrumentation.ts
```diff
@@ -0,0 +1,91 @@
...
+  registerInstrumentations({
+    instrumentations: [getNodeAutoInstrumentations()]
+  });
+};
+
+const setupTelemetry = () => {
+  const appCfg = initEnvConfig();
+
+  if (appCfg.OTEL_TELEMETRY_COLLECTION_ENABLED) {
+    console.log("Initializing telemetry instrumentation");
```
maidul98, line 80: use logger

maidul98 (commented):
looks good, just two small nits. also please add a run buck for setting this up locally in a runbook

maidul98 (approved):

PR #1928, opened 2024-06-07
feat: added option for disabling github secret deletion

sheensantoscapadngan:
# Description 📣
- This PR will allow users to configure integration with Github in such a way that it does not delete existing records that do not exist on Infisical. 
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/integration-auth/integration-sync-secret.ts
+++ b/backend/src/services/integration-auth/integration-sync-secret.ts
@@ -1364,7 +1364,7 @@ const syncSecretsGitHub = async ({
   }
 
   const metadata = z.record(z.any()).parse(integration.metadata);
-  if (!metadata.shouldDisableDelete) {
+  if (metadata.shouldEnableDelete) {
     for await (const encryptedSecret of encryptedSecrets) {
       if (
         !(encryptedSecret.name in secrets) &&
--- a/backend/src/services/integration-auth/integration-sync-secret.ts
+++ b/backend/src/services/integration-auth/integration-sync-secret.ts
@@ -1363,38 +1363,41 @@ const syncSecretsGitHub = async ({
     }
   }
 
-  for await (const encryptedSecret of encryptedSecrets) {
-    if (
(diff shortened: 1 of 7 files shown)


backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -1363,38 +1363,41 @@ const syncSecretsGitHub = async ({
...
-          });
-          break;
-        }
-        case GithubScope.Env: {
-          await octokit.request(
-            "DELETE /repositories/{repository_id}/environments/{environment_name}/secrets/{secret_name}",
-            {
-              repository_id: Number(integration.appId),
-              environment_name: integration.targetEnvironmentId as string,
+  const metadata = z.record(z.any()).parse(integration.metadata);
```
DanielHougaard, line 1366: If we know how the metadata structure is gonna look like, I think we should create a real zod model instead of just using `zod.any()`. This will give us more clarity with better type-safety, and act as a failsafe if the metadata is somehow falty.

DanielHougaard (changes requested):

frontend/src/pages/integrations/github/create.tsx
```diff
@@ -546,6 +550,21 @@ export default function GitHubCreateIntegrationPage() {
...
+                  <Controller
+                    control={control}
+                    name="shouldEnableDelete"
+                    render={({ field: { onChange, value } }) => (
+                      <Switch
+                        id="delete-github-option"
+                        onCheckedChange={(isChecked) => onChange(isChecked)}
+                        isChecked={value}
+                      >
+                        Delete secrets not in Infisical
```
DanielHougaard, line 563: We can make this a bit more clear to the user by naming it "Delete secrets in Github that are not in Infisical"

DanielHougaard (changes requested):
Looks good! Left a few comments for you

Issue #1930, opened 2024-06-08
Managed kubernetes secret annotations reset during update


minuchi:
### Describe the bug
When the Kubernetes operator fetches secret changes from Infisical, the annotations on the managed secret are reset. This results in the loss of all existing annotations, except for the newly updated `ETag` annotation.

### To Reproduce
1. Deploy the Kubernetes operator.
2. Create a managed secret with some annotations.
3. Trigger an update to fetch secret changes from Infisical.
4. Observe that all annotations, except for the `ETag`, are reset.

### Expected behavior
The annotations on the managed secret should be preserved during the update process. Only the `ETag` annotation should be updated without resetting other existing annotations.

### Screenshots
**Before the update:**
<img width="1179" alt="before-the-fix-1" src="https://github.com/Infisical/infisical/assets/112928337/9b5d1f70-fceb-4327-a3bd-401eb050ca5f">

**After the update (annotations reset):**
<img width="806" alt="before-the-fix-2" src="https://github.com/Infisical/infisical/assets/112928337/dd4733ad-6104-4596-b9a8-e6d0e27cb840">

### Platform you are having the issue on:
- Kubernetes

### Additional context
The current behavior of resetting annotations can lead to the loss of important metadata. Preserving existing annotations ensures that all critical information is retained while updating the `ETag`.

PR #1931, opened 2024-06-08
fix: preserve existing annotations when updating managed secret

minuchi:
# Description 📣

This pull request fixes https://github.com/Infisical/infisical/issues/1930 where the annotations on a managed Kubernetes secret were being reset during the update process. With this change, we ensure that existing annotations are preserved and only the `ETag` annotation is updated.

- Fixed the bug where updating the managed secret would reset all annotations. Now, only the `ETag` annotation is updated, preserving existing annotations.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

1. Deploy the Kubernetes operator.
2. Check the annotations on the managed secret before updating.
3. Trigger an update.
4. Verify the annotations on the managed secret after the update.

### Before the fix:
- Before update (annotations present)
  before-the-fix-1
- After update before the fix (annotations reset)
  before-the-fix-2

### After the fix:
- After update with the fix (annotations preserved)
  after-the-fix-1

---

### DIFF
--- a/k8-operator/controllers/infisicalsecret_helper.go
+++ b/k8-operator/controllers/infisicalsecret_helper.go
@@ -232,7 +232,6 @@ func (r *InfisicalSecretReconciler) UpdateInfisicalManagedKubeSecret(ctx context
 	}
 
 	managedKubeSecret.Data = plainProcessedSecrets
-	managedKubeSecret.ObjectMeta.Annotations = map[string]string{}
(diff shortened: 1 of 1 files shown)


maidul98 (approved):
nice fix

maidul98:
When this is done https://github.com/Infisical/infisical/actions/runs/9455645624, you can update your values file for helm to use the new image tag.

The new image should be tagged `v0.5.3` here https://hub.docker.com/r/infisical/kubernetes-operator/tags

PR #1944, opened 2024-06-10
Removing Master password for Oauth/SSO/LDAP users.

akhilmhdh:
# Description 📣

First step in moving away from our SRP system to a more proficient automation platform. With SRP we will send password also and we save the password with bcrypt hashing in db.

To avoid asking for master password, we just fetch the private key on login. For oauth user we do a handoff that is the provider token act as password to get the access token for infisical.

This also means no more book keeping master password for third part authorized users like LDAP, Google Oauth etc.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/lib/crypto/srp.ts
```diff
@@ -108,17 +114,18 @@ export const getUserPrivateKey = async (password: string, user: TUserEncryptionK
     raw: true
   });
   if (!derivedKey) throw new Error("Failed to derive key from password");
-  const key = decryptSymmetric({
-    ciphertext: user.protectedKey!,
-    iv: user.protectedKeyIV!,
-    tag: user.protectedKeyTag!,
-    key: derivedKey.toString("base64")
+  const key = decryptSymmetric128BitHexKeyUTF8({
```
maidul98, line 117: is this key always 128 bit or can it be 256 bit too?

backend/src/services/auth/auth-login-service.ts
```diff
@@ -248,14 +253,29 @@ export const authLoginServiceFactory = ({
...
-
     await userDAL.updateById(userEnc.userId, {
       consecutiveFailedPasswordAttempts: 0
     });
+    // from password decrypt the private key
+    if (password) {
+      const privateKey = await getUserPrivateKey(password, userEnc);
+      const hashedPassword = await bcrypt.hash(password, cfg.BCRYPT_SALT_ROUND);
+      const { iv, tag, ciphertext, encoding } = infisicalSymmetricEncypt(privateKey);
+      await userDAL.updateUserEncryptionByUserId(userEnc.userId, {
```
maidul98, line 264: Looks like we are updating this with the same info each time the user logins. Is this intended? I would assume we only update this once if they don't already have `serverEncrypted*` fields defined

backend/src/services/auth/auth-login-service.ts
```diff
@@ -511,9 +537,9 @@ export const authLoginServiceFactory = ({
         isEmailVerified: user.isEmailVerified,
         firstName: user.firstName,
         lastName: user.lastName,
+        hasExchangedPrivateKey: Boolean(userEnc?.serverEncryptedPrivateKey),
```
maidul98, line 540: Oh nice, so as long as they go through the normal flow at least once, they will not have to enter in master password for oauth flow?

frontend/src/views/Signup/components/UserInfoSSOStep/UserInfoSSOStep.tsx
```diff
@@ -69,12 +54,13 @@ export const UserInfoSSOStep = ({
...
   const [organizationNameError, setOrganizationNameError] = useState(false);
   const [attributionSource, setAttributionSource] = useState("");
-  const [errors, setErrors] = useState<Errors>({});
   const [isLoading, setIsLoading] = useState(false);
   const { t } = useTranslation();
   const { mutateAsync: selectOrganization } = useSelectOrganization();
 
   useEffect(() => {
+    const randomPassword = crypto.randomBytes(32).toString("hex");
+    setPassword(randomPassword);
```
maidul98, line 63: Would like a bit more context on this part

maidul98 (commented):
Backend logic looks good to me. Also reviewed the frontend and most of it makes sense to me. Left comments where more context would be helpful.  In terms of testing here is what I did:

- Successfully created a new account using Google SSO (this process works well). I understand that we generate a password specifically for SSO-only signups. How does this interact with the password reset process? I followed the entire flow for the new account created via Google SSO and set a new password. However, it appears that the auto-generated password wasn't replaced since I couldn't log in with the email and new password.
- Account created with username and password and then linked with both Google and Github SSO. Works well, no password step needed.
- @dangtony98 Can you please try SCIM/LDAP for testing?
- Haven't yet tested password reset for account created with username/password yet

backend/src/lib/crypto/srp.ts
akhilmhdh: @dangtony98 named this. I don't thin we have 256 bit key.

CC: @dangtony98

backend/src/services/auth/auth-login-service.ts
akhilmhdh: We keep the value till last login. Because they could also do reset password and more

frontend/src/views/Signup/components/UserInfoSSOStep/UserInfoSSOStep.tsx
akhilmhdh: For signup of SSO user, this password doesn't matter. Its a random string. Soon in server we will be doing the same property.

backend/src/services/auth/auth-fns.ts
```diff
@@ -15,10 +15,10 @@ export const validateProviderAuthToken = (providerToken: string, username?: stri
   if (decodedToken.username !== username) throw new Error("Invalid auth credentials");
 
   if (decodedToken.organizationId) {
-    return { orgId: decodedToken.organizationId, authMethod: decodedToken.authMethod };
+    return { orgId: decodedToken.organizationId, authMethod: decodedToken.authMethod, userName: decodedToken.username };
   }
 
-  return { authMethod: decodedToken.authMethod, orgId: null };
+  return { authMethod: decodedToken.authMethod, orgId: null, userName: decodedToken.username };
```
dangtony98, line 21: Why do we return `userName: decodedToken.username` here? It doesn't seem to be used anywhere else.

dangtony98 (dismissed):

PR #1947, opened 2024-06-11
Patch multi line encoding when expandSecretRef is enabled

maidul98:
By default when you create a secret, it will have multi line encoding off but we actually treat this as true in the backend and UI. User’s aren’t expecting their multi line secrets to be double quoted by and made into a single line with/n, however we are doing it by default at the moment. This PR makes multi line encoding opt in and not opt out

### DIFF
--- a/backend/src/db/schemas/secrets.ts
+++ b/backend/src/db/schemas/secrets.ts
@@ -23,7 +23,7 @@ export const SecretsSchema = z.object({
   secretCommentTag: z.string().nullable().optional(),
   secretReminderNote: z.string().nullable().optional(),
   secretReminderRepeatDays: z.number().nullable().optional(),
-  skipMultilineEncoding: z.boolean().default(false).optional(),
+  skipMultilineEncoding: z.boolean().default(false).nullable().optional(),
   algorithm: z.string().default("aes-256-gcm"),
   keyEncoding: z.string().default("utf8"),
   metadata: z.unknown().nullable().optional(),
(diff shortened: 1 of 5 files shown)


backend/src/db/schemas/secrets.ts
```diff
@@ -23,7 +23,7 @@ export const SecretsSchema = z.object({
   secretCommentTag: z.string().nullable().optional(),
   secretReminderNote: z.string().nullable().optional(),
   secretReminderRepeatDays: z.number().nullable().optional(),
-  skipMultilineEncoding: z.boolean().default(false).nullable().optional(),
+  skipMultilineEncoding: z.boolean().default(false).optional(),
```
maidul98, line 26: The migration file has this as a bool not nullable so not sure why it was set to nullable here

backend/src/services/secret/secret-fns.ts
```diff
@@ -309,7 +309,7 @@ export const interpolateSecrets = ({ projectId, secretEncKey, secretDAL, folderD
   };
 
   const expandSecrets = async (
-    secrets: Record<string, { value: string; comment?: string; skipMultilineEncoding?: boolean }>
+    secrets: Record<string, { value: string; comment?: string; skipMultilineEncoding?: boolean | null }>
```
maidul98, line 312: Why does our schema set skipMultilineEncoding to be nullable when the migration has it as a bool? @akhilmhdh

akhilmhdh: Because it can be nullable. Default false doesn't mean it cannot be nullable.

akhilmhdh (approved):

PR #1956, opened 2024-06-12
feat: custom rate limit for self hosters

ShubhamPalriwala:
# Description 📣
Allow self hosted users to set their own rate limits in admin panel

Right now, self hosted users get hard coded rate limits. We need to make it so that they can configure this in the admin panel

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
Go to Admin Panel > Rate Limits tab > Set a Limit to 1 or 2 > Query that API.page and you should see the 429 Too many requests error after the count.

```sh
```

---


frontend/src/views/admin/DashboardPage/DashboardPage.tsx
```diff
@@ -56,7 +68,15 @@ export const AdminDashboardPage = () => {
       signUpMode: config.allowSignUp ? SignUpModes.Anyone : SignUpModes.Disabled,
       allowedSignUpDomain: config.allowedSignUpDomain,
       trustSamlEmails: config.trustSamlEmails,
-      trustLdapEmails: config.trustLdapEmails
+      trustLdapEmails: config.trustLdapEmails,
```
akhilmhdh, line 71: As its a tab and two independent form i would recommend to split this into two components.

Then you can put the form validation required for each tab in there component space independently. Issue i see here is what like one tab has an error the other tab form won't work.

frontend/src/hooks/api/rateLimit/queries.ts
```diff
@@ -0,0 +1,34 @@
+import { useQuery, UseQueryOptions } from "@tanstack/react-query";
+
+import { apiRequest } from "@app/config/request";
+
+import { TRateLimit } from "./types";
+
+export const rateLimitQueryKeys = {
```
akhilmhdh, line 7: Barrel export in index.ts so everywhere can be imported from `@app/hooks/api`

backend/src/server/config/rateLimiter.ts
```diff
@@ -74,3 +62,38 @@ export const publicEndpointLimit: RateLimitOptions = {
   max: 30,
   keyGenerator: (req) => req.realIp
 };
+
+async function fetchRateLimitsFromDb(db: Knex) {
+  try {
+    const rateLimitDAL = rateLimitDALFactory(db);
```
akhilmhdh, line 68: This pattern feels very intuitive. I would recommend instead load the data from the  outside and just pass the configuration thus avoiding all these in between things.

backend/src/server/config/rateLimiter.ts
```diff
@@ -74,3 +62,38 @@ export const publicEndpointLimit: RateLimitOptions = {
...
+
+    readLimit.max = rateLimits.readRateLimit;
+    publicEndpointLimit.max = rateLimits.publicEndpointLimit;
+    writeLimit.max = rateLimits.writeRateLimit;
+    secretsLimit.max = rateLimits.secretsRateLimit;
+    authRateLimit.max = rateLimits.authRateLimit;
+    inviteUserRateLimit.max = rateLimits.inviteUserRateLimit;
+    mfaRateLimit.max = rateLimits.mfaRateLimit;
+    creationLimit.max = rateLimits.creationLimit;
+  } catch (error) {
```
akhilmhdh, line 79: Use logger for consoling

backend/src/services/rate-limit/rate-limit-service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+type TRateLimitServiceFactoryDep = {
+  rateLimitDAL: TRateLimitDALFactory;
+};
+
+export type TRateLimitServiceFactory = ReturnType<typeof rateLimitServiceFactory>;
+
+export const rateLimitServiceFactory = ({ rateLimitDAL }: TRateLimitServiceFactoryDep) => {
+  const initRateLimits = async (): Promise<TRateLimit> => {
+    const rateLimit = await rateLimitDAL.create({});
```
akhilmhdh, line 14: Keep a uuid of `0000-000` like admin cfg. Else it will cause if by accident there is two configuration. Same goes for find. Find by uuid of 00000-000... if that is the only document that should exist.

backend/src/services/rate-limit/rate-limit-service.ts
```diff
@@ -0,0 +1,44 @@
...
+    return rateLimit;
+  };
+
+  const getRateLimits = async (): Promise<TRateLimit> => {
+    let rateLimit = (await rateLimitDAL.find({}))[0];
```
akhilmhdh, line 19: Use findOne

backend/src/server/routes/v1/rate-limit-router.ts
```diff
@@ -0,0 +1,58 @@
...
+      rateLimit: readLimit
+    },
+    onRequest: (req, res, done) => {
+      verifyAuth([AuthMode.JWT, AuthMode.API_KEY])(req, res, () => {
+        verifySuperAdmin(req, res, done);
+      });
+    },
+
+    schema: {
+      body: RateLimitSchema.omit({
```
akhilmhdh, line 42: Avoid omit and use pick for body. But i would recommend explicitly writing it out. 

Because why avoid omit is because it should always be `.pick` for response and request validation as if schema changes other devs won't accidentally leaking data.

Explicitly saying data also gives control.

backend/src/services/rate-limit/rate-limit-service.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+    return rateLimit;
+  };
+
+  const updateRateLimit = async (updates: TRateLimitUpdateDTO): Promise<TRateLimit> => {
+    const rateLimit = await rateLimitDAL.findOne({});
+    if (!rateLimit) throw new BadRequestError({ name: "Rate Limit Update", message: "Rate Limit does not exist yet" });
+
+    const updateData: Record<string, number> = {};
+    for (const [key, value] of Object.entries(updates)) {
```
akhilmhdh, line 31: I am a bit confused by the need for this one

backend/src/services/rate-limit/rate-limit-service.ts
```diff
@@ -0,0 +1,44 @@
...
+  const getRateLimits = async (): Promise<TRateLimit> => {
+    let rateLimit = (await rateLimitDAL.find({}))[0];
+    if (!rateLimit) {
+      rateLimit = await initRateLimits();
+    }
+    return rateLimit;
+  };
+
+  const updateRateLimit = async (updates: TRateLimitUpdateDTO): Promise<TRateLimit> => {
+    const rateLimit = await rateLimitDAL.findOne({});
```
akhilmhdh, line 27: uuid zero findOne

akhilmhdh (changes requested):

frontend/src/views/admin/DashboardPage/DashboardPage.tsx
```diff
@@ -233,6 +291,177 @@ export const AdminDashboardPage = () => {
...
+                onSubmit={handleSubmit(onRateLimitFormSubmit)}
+              >
+                <div className="mb-8 flex flex-col justify-start">
+                  <div className="mb-4 text-xl font-semibold text-mineshaft-100">
+                    Set Rate Limits for your Infisical Instance
+                  </div>
+                  <Controller
+                    control={control}
+                    name="readRateLimit"
+                    defaultValue={300}
```
sheensantoscapadngan, line 306: should only be made available for non-cloud

backend/src/server/routes/v1/rate-limit-router.ts
sheensantoscapadngan, line 42: add check to ensure that it can only be called in non-cloud instances (maybe a flag for now)

backend/src/db/migrations/20240611151327_custom-rate-limits-for-self-hosting.ts
```diff
@@ -0,0 +1,28 @@
...
+      t.integer("readRateLimit").defaultTo(600).notNullable();
+      t.integer("writeRateLimit").defaultTo(200).notNullable();
+      t.integer("secretsRateLimit").defaultTo(60).notNullable();
+      t.integer("authRateLimit").defaultTo(60).notNullable();
+      t.integer("inviteUserRateLimit").defaultTo(30).notNullable();
+      t.integer("mfaRateLimit").defaultTo(20).notNullable();
+      t.integer("creationLimit").defaultTo(30).notNullable();
+      t.integer("publicEndpointLimit").defaultTo(30).notNullable();
+      t.timestamps(true, true, true);
+    });
```
sheensantoscapadngan, line 19: we should at least be tracking which user last updated the rate limit configs

sheensantoscapadngan:
addressing comments now...

sheensantoscapadngan:
make setting of rate limits dynamic so that it does not require instance restart

akhilmhdh:
@sheensantoscapadngan can we have a short discussion on slack please?

PR #1965, opened 2024-06-13
feat: allow custom rate limits

sheensantoscapadngan:
# Description 📣
- this PR will allow admins to configure custom rate limits for their Infisical instances

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/services/rate-limit/rate-limit-service.ts
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  const updateRateLimit = async (updates: TRateLimitUpdateDTO): Promise<TRateLimit> => {
+    const appCfg = getConfig();
+    if (!appCfg.ALLOW_RATELIMIT_UPDATES) {
+      throw new ForbiddenRequestError({
+        name: "Rate limit Updates Disabled",
+        message: "Changes to rate limits are disabled"
+      });
+    }
```
sheensantoscapadngan, line 33: for infisical cloud

backend/src/services/rate-limit/rate-limit-service.ts
```diff
@@ -0,0 +1,68 @@
...
+        logger.error(`Error syncing rate limit configurations: %o`, error);
+      }
+    };
+
+    // sync rate limits configuration every 10 minutes
+    const job = new CronJob("*/10 * * * *", rateLimitSync);
+    job.start();
+
+    return job;
+  };
```
sheensantoscapadngan, line 61: an alternative I'm thinking of is using redis to store rate limit max values and then in the rate limit, just fetch the latest values from redis

sheensantoscapadngan: but my worry is that spam API requests can bog down redis, that's why I went with the local cache approach instead

backend/src/db/migrations/20240611151327_custom-rate-limits-for-self-hosting.ts
```diff
@@ -0,0 +1,31 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+import { createOnUpdateTrigger, dropOnUpdateTrigger } from "../utils";
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasTable(TableName.RateLimit))) {
+    await knex.schema.createTable(TableName.RateLimit, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo("00000000-0000-0000-0000-000000000000");
```
akhilmhdh, line 9: This defaultTo is not needed.

backend/src/db/migrations/20240611151327_custom-rate-limits-for-self-hosting.ts
```diff
@@ -0,0 +1,31 @@
...
+      t.integer("mfaRateLimit").defaultTo(20).notNullable();
+      t.integer("creationLimit").defaultTo(30).notNullable();
+      t.integer("publicEndpointLimit").defaultTo(30).notNullable();
+      t.timestamps(true, true, true);
+    });
+
+    await createOnUpdateTrigger(knex, TableName.RateLimit);
+
+    // create rate limit entry
+    await knex(TableName.RateLimit).insert({});
```
akhilmhdh, line 24: Its better to do this programmatically than here. Like user deleting the row means this would end up in a crash loop from program perspective

backend/src/services/rate-limit/rate-limit-service.ts
akhilmhdh: I don't think for infisical cloud we need to disable this. I mean its super admin only right.

CC: @maidul98

frontend/src/views/admin/DashboardPage/RateLimitPanel.tsx
```diff
@@ -0,0 +1,248 @@
...
+      });
+    } catch (e) {
+      console.error(e);
+      createNotification({
+        type: "error",
+        text: "Failed to update rate limiting setting."
+      });
+    }
+  };
+
```
akhilmhdh, line 80: Add a loader to wait for rate limit data to finish.

backend/src/server/app.ts
```diff
@@ -39,6 +42,7 @@ type TMain = {
 // Run the server!
 export const main = async ({ db, smtp, logger, queue, keyStore }: TMain) => {
   const appCfg = getConfig();
+  const cronJobs: CronJob[] = [];
```
akhilmhdh, line 45: Hmm i need to think more about this pattern. Feels kinda very weird singleton pattern.

You can start the cron jobs in `routes/index` like normal and then do a server.register for cleanup event. I ideally i am thinking what if we move rate limiting logic inside the `routes` section as its not part of app logic

akhilmhdh (changes requested):

backend/src/db/migrations/20240611151327_custom-rate-limits-for-self-hosting.ts
sheensantoscapadngan: ah yeah that makes sense, will address!

backend/src/services/rate-limit/rate-limit-service.ts
sheensantoscapadngan: ah, was confused about who the super admin is. will address

frontend/src/views/admin/DashboardPage/RateLimitPanel.tsx
sheensantoscapadngan: there's already a loader for this in the "Save" button. Is that what you mean? If you're referring to a loader showing when the actual rate limit changes would take effect, not sure how that would look like given that in the current implementation, it might take around 10 mins 

image

akhilmhdh: I meant the form value loading from api there will be a brief time.

backend/src/server/app.ts
sheensantoscapadngan: moved the cron setup in the routes section

frontend/src/views/admin/DashboardPage/RateLimitPanel.tsx
sheensantoscapadngan: oh okay got this!

sheensantoscapadngan: done @akhilmhdh

backend/src/server/routes/v1/rate-limit-router.ts
```diff
@@ -0,0 +1,70 @@
+import { z } from "zod";
+
+import { RateLimitSchema } from "@app/db/schemas";
+import { BadRequestError } from "@app/lib/errors";
+import { readLimit } from "@app/server/config/rateLimiter";
+import { verifySuperAdmin } from "@app/server/plugins/auth/superAdmin";
+import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
+import { AuthMode } from "@app/services/auth/auth-type";
+
+export const registerRateLimitRouter = async (server: FastifyZodProvider) => {
```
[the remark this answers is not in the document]
sheensantoscapadngan: oh.. fixed this

maidul98 (approved):
lgtm!

maidul98:
There are no breaking changes for this, we need to fix the GHA

Issue #1802, opened 2024-05-08
CLI Login and Set Process Does Not Work Per Documentation


seancallaway:
### Describe the bug
After logging in, running `infisical secrets set` generates a "You must be logged in to run this command. To login, run [infisical login]" error, though `infisical secrets get` (with a projectId) will return data.

### To Reproduce
1. Set INFISICAL_API_URL (`export INFISICAL_API_URL=https://<REDACTED_FQDN>/api`)
2. Login, setting token (`export INFISICAL_TOKEN=$(infisical login --method=universal-auth --client-id=<REDACTED_CLIENT_ID> --client-secret=<REDACTED_CLIENT_SECRET> --silent --plain`)
3. Try to set a secret (`infisical secrets set --env=staging TEST_SECRET=test`)
4. Get error message (`You must be logged in to run this command. To login, run [infisical login]`
5. Try to get a secret (`infisical secrets --projectId <REDACTED_PROJECT_ID> --env=staging get TEST_SECRET`)
6. Get table containing "*not found*" for the SECRET VALUE (expected, as it is unset). I am able to get a secret set via the web UI for the dev environment.

### Expected behavior
I'd expect step 3 to succeed, setting the secret properly, based on the [documentation](https://infisical.com/docs/cli/commands/secrets#infisical-secrets-set).

### Platform you are having the issue on:

**CLI**
- MacOS: infisical version 0.22.2
- Alpine (CI container): infisical version 0.22.2

**Self-Hosted Docker-Compose Setup**
Installed today using `:latest`.

### Additional context
Tested from my workstation and our GitLab CI/CD pipeline.

The machine identity in question has the following permission in the project:

![image](https://github.com/Infisical/infisical/assets/1229384/2f0a43eb-c0a1-4563-b590-265280495f88)

seancallaway:
Team has indicated that there is, in fact, an issue with Infisical via [Slack](https://infisical-users.slack.com/archives/C04BSBMQAQ7/p1715268345969999?thread_ts=1715192395.606039&cid=C04BSBMQAQ7).

seancallaway:
Turns out that this just isn't a supported operation. Documentation should be fixed to indicate this or, better yet, implement this.

DanielHougaard:
Hi @seancallaway, this has now been added in the latest CLI version. Please update your CLI to `>= 0.23.2`!

PR #1981, opened 2024-06-15
feat(cli): Unify CLI auth methods

DanielHougaard:
# Description 📣

We've previously only had support for token-based auth in the CLI for a specific set of operations _(mostly operations we deemed as important for CI/CD workflows)_. As Infisical grows, so does our users use-cases, and we need to allow all our CLI operations to be used via. token-based auth. This PR addresses the lack of a unified auth system in the CLI (#1802). 

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/cli/packages/cmd/folder.go
+++ b/cli/packages/cmd/folder.go
@@ -127,9 +127,6 @@ var createCmd = &cobra.Command{
 
 		if token != nil && (token.Type == util.SERVICE_TOKEN_IDENTIFIER || token.Type == util.UNIVERSAL_AUTH_TOKEN_IDENTIFIER) {
 			params.InfisicalToken = token.Token
-		} else {
-			util.RequireLogin()
-			util.RequireLocalWorkspaceFile()
 		}
 
 		_, err = util.CreateFolder(params)
@@ -198,9 +195,6 @@ var deleteCmd = &cobra.Command{
 
 		if token != nil && (token.Type == util.SERVICE_TOKEN_IDENTIFIER || token.Type == util.UNIVERSAL_AUTH_TOKEN_IDENTIFIER) {
 			params.InfisicalToken = token.Token
-		} else {
-			util.RequireLogin()
-			util.RequireLocalWorkspaceFile()
 		}
 
 		_, err = util.DeleteFolder(params)
--- a/cli/packages/cmd/folder.go
+++ b/cli/packages/cmd/folder.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/Infisical/infisical-merge/packages/models"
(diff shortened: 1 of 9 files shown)


cli/packages/cmd/folder.go
```diff
@@ -95,19 +102,31 @@ var createCmd = &cobra.Command{
...
+
 		params := models.CreateFolderParameters{
 			FolderName:  folderName,
-			WorkspaceId: workspaceFile.WorkspaceId,
 			Environment: environmentName,
 			FolderPath:  folderPath,
+			WorkspaceId: projectId,
+		}
+
+		if token != nil && (token.Type == util.SERVICE_TOKEN_IDENTIFIER || token.Type == util.UNIVERSAL_AUTH_TOKEN_IDENTIFIER) {
```
maidul98, line 128: I'm wondering if the names of the tokens should be uniform. I think that would make it easy for end users if the token is just set under INFISICAL_TOKEN regardless of auth type. What i'm thinking is the only difference between service token and machine identity tokens is that service tokens have a private key attached to the token but they can both be added as bearer token. What do you think?

cli/packages/cmd/secrets.go
```diff
@@ -160,211 +154,38 @@ var secretsSetCmd = &cobra.Command{
...
+			util.HandleError(err, "Unable to parse secret type")
 		}
 
-		encryptedWorkspaceKey, _ := base64.StdEncoding.DecodeString(workspaceKeyResponse.EncryptedKey)
-		encryptedWorkspaceKeySenderPublicKey, _ := base64.StdEncoding.DecodeString(workspaceKeyResponse.Sender.PublicKey)
-		encryptedWorkspaceKeyNonce, _ := base64.StdEncoding.DecodeString(workspaceKeyResponse.Nonce)
-		currentUsersPrivateKey, _ := base64.StdEncoding.DecodeString(loggedInUserDetails.UserCredentials.PrivateKey)
+		var secretOperations []models.SecretSetOperation
+		if token != nil && (token.Type == util.SERVICE_TOKEN_IDENTIFIER || token.Type == util.UNIVERSAL_AUTH_TOKEN_IDENTIFIER) {
+			secretOperations, err = util.SetRawSecrets(args, secretType, environmentName, secretsPath, projectId, token)
```
maidul98, line 179: Could it be that the person using service tokens haven't yet upgraded their project?

cli/packages/cmd/secrets.go
```diff
@@ -804,25 +646,17 @@ func init() {
 
 	secretsCmd.Flags().Bool("secret-overriding", true, "Prioritizes personal secrets, if any, with the same name over shared secrets")
 	secretsCmd.AddCommand(secretsSetCmd)
+	secretsSetCmd.Flags().String("token", "", "Fetch secrets using the Infisical Token")
+	secretsSetCmd.Flags().String("projectId", "", "manually set the projectId to fetch folders from for machine identity")
```
maidul98, line 650: I think you meant `when using machine identity based auth`?

maidul98 (commented):
Just left some small comments

cli/packages/cmd/folder.go
DanielHougaard: This is actually handled in the `GetInfisicalToken()` method. It checks the `INFISICAL_TOKEN` env variable

cli/packages/cmd/secrets.go
DanielHougaard: In the `secrets set` command, I decided to assume that the user attempting to use their service token for setting secrets, has already upgraded their project. I opted for this to reduce complexity to some degree

maidul98 (approved):

PR #1982, opened 2024-06-15
(helm)Fix: protocol for container port

gmemstr:
This can trip up strict spec validators like https://kubenix.org/

# Description 📣

Ensures the Helm chart adheres strictly to the Kubernetes pod schema.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---


gmemstr:
Worth noting that [the secrets operator already has this](https://github.com/Infisical/infisical/blob/main/helm-charts/secrets-operator/templates/deployment.yaml#L49-L52).

PR #1988, opened 2024-06-17
Fix: Silent integration errors

DanielHougaard:
# Description 📣

1. Fixed silent integration errors. This made it hard to figure out if your secrets were actually syncing or not, because it wouldn't be visible in the Integrations UI on the dashboard if it failed.
2. Smaller redundancies in the integration flows.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/integration-auth/integration-sync-secret.ts
+++ b/backend/src/services/integration-auth/integration-sync-secret.ts
@@ -18,7 +18,7 @@ import {
   UpdateSecretCommand
 } from "@aws-sdk/client-secrets-manager";
 import { Octokit } from "@octokit/rest";
-import AWS from "aws-sdk";
+import AWS, { AWSError } from "aws-sdk";
 import { AxiosError } from "axios";
 import sodium from "libsodium-wrappers";
 import isEqual from "lodash.isequal";
@@ -452,7 +452,9 @@ const syncSecretsAWSParameterStore = async ({
   accessId: string | null;
   accessToken: string;
 }) => {
-  if (!accessId) return;
+  let response: { isSynced: boolean; syncMessage: string } | null = null;
+
+  if (!accessId) return null;
 
   const config = new AWS.Config({
     region: integration.region as string,
@@ -556,8 +558,12 @@ const syncSecretsAWSParameterStore = async ({
               logger.error(
                 `AWS Parameter Store Error [integration=${integration.id}]: double check AWS account permissions (refer to the Infisical docs)`
               );
-              throw err;
             }
+
+            response = {
+              isSynced: false,
+              syncMessage: (err as AWSError)?.message || "Error syncing with AWS Parameter Store"
+            };
           }
         }
       }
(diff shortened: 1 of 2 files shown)


backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -557,6 +557,8 @@ const syncSecretsAWSParameterStore = async ({
                 `AWS Parameter Store Error [integration=${integration.id}]: double check AWS account permissions (refer to the Infisical docs)`
               );
             }
+
+            throw err;
```
sheensantoscapadngan, line 561: context: I previously introduced a new set of policies that have to be added for the AWS integration to work as intended. I explicitly did not throw an error here so that the flow does not break for users that have not yet adjusted their policies to include the new set of permissions. Maybe we could adjust this so that it only throws an error if its not an "AccessDeniedException"?

i should have added a comment

backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -862,10 +867,6 @@ const syncSecretsHeroku = async ({
       }
     }
   );
-
-  await integrationDAL.updateById(integration.id, {
-    lastUsed: new Date()
-  });
```
sheensantoscapadngan, line 868: wait why did we remove this?

backend/src/services/integration-auth/integration-sync-secret.ts
DanielHougaard: Oh gotcha, great catch in that case. I've updated the code to only throw if it's an access denied error!

sheensantoscapadngan: oh wait, we should only be throwing if it's not an access denied error

backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -452,7 +452,9 @@ const syncSecretsAWSParameterStore = async ({
   accessId: string | null;
   accessToken: string;
 }) => {
-  if (!accessId) return;
+  let response: { isSynced: boolean; syncMessage: string } | null = null;
+
+  if (!accessId) return null;
```
sheensantoscapadngan, line 457: let's throw an error here instead of returning null?

backend/src/services/secret/secret-queue.ts
```diff
@@ -568,13 +562,15 @@ export const secretQueueFactory = ({
         await integrationDAL.updateById(integration.id, {
           lastSyncJobId: job.id,
           lastUsed: new Date(),
-          syncMessage: "",
-          isSynced: true
+          syncMessage: response?.syncMessage || "",
+          isSynced: response?.isSynced || true
```
sheensantoscapadngan, line 566: i think this always evaluates to true? 

maybe `??`

sheensantoscapadngan (approved):

PR #1993, opened 2024-06-18
feat(k8-operator): reconcile on managed secret deletion

DanielHougaard:
# Description 📣

Currently when the managed secret is deleted it'll be populated again on the next reconcile. This can be problematic especially if your resync interval is high. By adding reconcile on deletion, the user has a way to instantly refresh their secrets if needed.

The implementation takes place directly in the Manager, with minimal structural changes.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/k8-operator/controllers/infisicalsecret_controller.go
+++ b/k8-operator/controllers/infisicalsecret_controller.go
@@ -75,6 +75,31 @@ func (r *InfisicalSecretReconciler) handleFinalizer(ctx context.Context, infisic
 	return nil
 }
 
+func (r *InfisicalSecretReconciler) handleManagedSecretDeletion(a client.Object) []ctrl.Request {
+	var requests []ctrl.Request
+	infisicalSecrets := &secretsv1alpha1.InfisicalSecretList{}
+	err := r.List(context.Background(), infisicalSecrets)
+	if err != nil {
+		fmt.Printf("unable to list Infisical Secrets from cluster because [err=%v]", err)
+		return requests
+	}
+
+	for _, infisicalSecret := range infisicalSecrets.Items {
+		if a.GetName() == infisicalSecret.Spec.ManagedSecretReference.SecretName &&
+			a.GetNamespace() == infisicalSecret.Spec.ManagedSecretReference.SecretNamespace {
+			requests = append(requests, ctrl.Request{
+				NamespacedName: client.ObjectKey{
+					Namespace: infisicalSecret.Namespace,
+					Name:      infisicalSecret.Name,
+				},
+			})
+			fmt.Printf("\nManaged secret deleted in resource %s: [name=%v] [namespace=%v]\n", infisicalSecret.Name, a.GetName(), a.GetNamespace())
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


k8-operator/controllers/infisicalsecret_controller.go
```diff
@@ -154,9 +185,18 @@ func (r *InfisicalSecretReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}, nil
 }
 
-// SetupWithManager sets up the controller with the Manager.
 func (r *InfisicalSecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&secretsv1alpha1.InfisicalSecret{}).
+		Watches(
+			&source.Kind{Type: &corev1.Secret{}},
```
maidul98, line 192: Can we check if the operator has permission to watch secret resources?

akhilmhdh (approved):

k8-operator/controllers/infisicalsecret_controller.go
```diff
@@ -69,6 +75,31 @@ func (r *InfisicalSecretReconciler) handleFinalizer(ctx context.Context, infisic
 	return nil
 }
 
+func (r *InfisicalSecretReconciler) handleManagedSecretDeletion(a client.Object) []ctrl.Request {
```
maidul98, line 78: nit: why is it called `a`? Isn't this the `Secret` that we got the event for?

k8-operator/controllers/infisicalsecret_controller.go
```diff
@@ -69,6 +75,31 @@ func (r *InfisicalSecretReconciler) handleFinalizer(ctx context.Context, infisic
...
+	var requests []ctrl.Request
+	infisicalSecrets := &secretsv1alpha1.InfisicalSecretList{}
+	err := r.List(context.Background(), infisicalSecrets)
+	if err != nil {
+		fmt.Printf("unable to list Infisical Secrets from cluster because [err=%v]", err)
+		return requests
+	}
+
+	for _, infisicalSecret := range infisicalSecrets.Items {
```
maidul98, line 87: Just so i'm on the same page. When the delete event is captured, we add details of where the managed secret was deleted so that our secret reconciler gets triggered for that name/namespace?

DanielHougaard: Yes, that's correct. When a delete event is captured, we add the details of where the managed secret was deleted. This ensures that our secret reconciler gets triggered for the corresponding name and namespace.

k8-operator/controllers/infisicalsecret_controller.go
DanielHougaard: Yep, it goes the required permissions,
//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;**watch**;create;update;delete

maidul98 (approved):

PR #1994, opened 2024-06-19
misc: added special pruning of non-root folders

sheensantoscapadngan:
# Description 📣
- Some snapshots are not being pruned correctly because they belong to a non-root folder that for some reason do not have a corresponding secret folder version

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/services/secret-snapshot/snapshot-dal.ts
+++ b/backend/src/ee/services/secret-snapshot/snapshot-dal.ts
@@ -331,8 +331,8 @@ export const snapshotDALFactory = (db: TDbClient) => {
    * Prunes excess snapshots from the database to ensure only a specified number of recent snapshots are retained for each folder.
    *
    * This function operates in three main steps:
-   * 1. Pruning snapshots from root/non-versioned folders.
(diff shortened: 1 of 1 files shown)


backend/src/ee/services/secret-snapshot/snapshot-dal.ts
```diff
@@ -382,12 +382,11 @@ export const snapshotDALFactory = (db: TDbClient) => {
               .join(TableName.Environment, `${TableName.Environment}.id`, `${TableName.SecretFolder}.envId`)
               .join(TableName.Project, `${TableName.Project}.id`, `${TableName.Environment}.projectId`)
               .join("snapshot_cte", "snapshot_cte.id", `${TableName.Snapshot}.id`)
-              .whereNull(`${TableName.SecretFolder}.parentId`)
```
sheensantoscapadngan, line 385: by removing this, we're able to handle cases when non-root folder is in secret folder but not in secret folder version

PR #2000, opened 2024-06-20
Feat: Default organization slug for LDAP/SAML 

DanielHougaard:
# Description 📣

We now have a way of setting the default slug as a part of the admin settings / server configuration. When the default org slug is set, a few things will happen:

1. When attempting to log in with LDAP, the org slug input field will be hidden from the user, and the value will be pre-set to the default organization slug.
2. When logging in with SAML and the default org slug is set, it will entirely skip the org slug input page, and take you straight to the SAML auth.
3. If the default org slug is set, you'll never be asked to "select an organization". It will assume you're trying to log into the default organization and automatically select the default organization.

Technical:
Instead of storing the organization slug as text in the database, we're storing a reference to the organization itself by ID. This will allow cascading to work, and I generally believe we should always be keeping references instead of just plain text when we're keeping track of resources.

Minor:
1. Fixed a type issue 

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/db/schemas/super-admin.ts
```diff
@@ -16,7 +16,8 @@ export const SuperAdminSchema = z.object({
   allowedSignUpDomain: z.string().nullable().optional(),
   instanceId: z.string().uuid().default("00000000-0000-0000-0000-000000000000"),
   trustSamlEmails: z.boolean().default(false).nullable().optional(),
-  trustLdapEmails: z.boolean().default(false).nullable().optional()
+  trustLdapEmails: z.boolean().default(false).nullable().optional(),
+  defaultOrgId: z.string().uuid().nullable().optional()
```
maidul98, line 20: what about `defaultAuthOrg`? Incase we have default org for something else in the future

akhilmhdh: This name is a bit misleading to me. What is the default org to is my question that keeps popping?

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -70,10 +76,33 @@ export const superAdminServiceFactory = ({
...
+        throw new BadRequestError({
+          name: "Update server config",
+          message: "Failed to find default organization"
+        });
+      }
+    }
+
+    const updatedServerCfg = await serverCfgDAL.updateById(ADMIN_CONFIG_DB_UUID, {
+      ...omit(data, ["defaultOrgSlug"]),
+      defaultOrgId: organization?.id || null
```
akhilmhdh, line 96: The issue is on patch if user didn't provide org id this would become null right. I think if its undefined then it must not do anything.

akhilmhdh (changes requested):

backend/src/services/super-admin/super-admin-service.ts
DanielHougaard: This is actually alright. When the user hits the "Save" button in the UI and they already have a default org set, it'll be included in the request.

It will only be set to null if the user sets the field to an empty value in the UI @akhilmhdh

backend/src/db/schemas/super-admin.ts
DanielHougaard: Resolved! 

`defaultOrgId` is now called `defaultAuthOrgId` and `defaultOrgSlug` is called `defaultAuthOrgSlug`.

backend/src/services/super-admin/super-admin-dal.ts
```diff
@@ -1,7 +1,30 @@
...
+      .where(`${TableName.SuperAdmin}.id`, id)
+      .leftJoin(TableName.Organization, `${TableName.SuperAdmin}.defaultAuthOrgId`, `${TableName.Organization}.id`)
+      .select(
+        db.ref("*").withSchema(TableName.SuperAdmin) as unknown as keyof TSuperAdmin,
+        db.ref("slug").withSchema(TableName.Organization).as("defaultAuthOrgSlug")
+      )
+      .first();
+
+    return {
+      defaultAuthOrgSlug: config?.defaultAuthOrgSlug || null,
```
maidul98, line 21: Okay, cool so you are getting both the defaultAuthOrgId and defaultAuthOrgSlug returned. Why do we need null? won't `config?.defaultAuthOrgSlug` just be undefined if it empty?

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -70,10 +76,33 @@ export const superAdminServiceFactory = ({
...
+      if (!organization) {
+        throw new BadRequestError({
+          name: "Update server config",
+          message: "Failed to find default organization"
+        });
+      }
+    }
+
+    const updatedServerCfg = await serverCfgDAL.updateById(ADMIN_CONFIG_DB_UUID, {
+      ...omit(data, ["defaultAuthOrgSlug"]),
```
maidul98, line 95: I think you can replace omit this by adding this on top  

```
delete data["defaultAuthOrgSlug"]
```

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -70,10 +76,33 @@ export const superAdminServiceFactory = ({
...
+      defaultAuthOrgId: organization?.id || null
+    });
+
+    const result = {
+      ...updatedServerCfg,
+      defaultAuthOrgSlug: organization?.slug || null
```
maidul98, line 101: noticing a few ` || null`, can we just leave this as undefined?

maidul98 (commented):
I see what you are doing. User enters slug, and you take the slug and get it's id and save the id and when we get the admin config, we returns the id and the associated slug. 

I just remembered, one of the friction points i observed watching customer do this is not knowing where the org slug is located at since it is their first time setting things up. I think it will be a better experience if they can just select a org from a drop down (we already have a API for getting users orgs) and then we send the id to the DB and save as is. When we fetch admin config, we can compute org slug as you do now. This way, saving is identical to before too

backend/src/services/super-admin/super-admin-service.ts
maidul98, line 95: I think you can replace omit this by adding this on top  

```
delete data["defaultAuthOrgSlug"]
```

maidul98 (changes requested):
FYI, tried to load things up with a fresh DB and the migration that creates the instance row in the DB doesn't get created so the APIs throw error when it calls get config

maidul98 (approved):
lgtm, let's definitely test in gamma since this not working will impact logins for enterprise

maidul98 (approved):

PR #2001, opened 2024-06-20
feat: resolved generate srp failing for user enc v1 users

akhilmhdh:
# Description 📣

We recently introduced master key removal and it caused users in enc v1 to fail on generateSrpKey. This will resolve that problem.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/lib/crypto/srp.ts
+++ b/backend/src/lib/crypto/srp.ts
@@ -101,33 +101,51 @@ export const getUserPrivateKey = async (
   password: string,
   user: Pick<
     TUserEncryptionKeys,
-    "protectedKeyTag" | "protectedKey" | "protectedKeyIV" | "encryptedPrivateKey" | "iv" | "salt" | "tag"
+    | "protectedKeyTag"
+    | "protectedKey"
+    | "protectedKeyIV"
+    | "encryptedPrivateKey"
+    | "iv"
+    | "salt"
+    | "tag"
+    | "encryptionVersion"
   >
 ) => {
-  const derivedKey = await argon2.hash(password, {
-    salt: Buffer.from(user.salt),
-    memoryCost: 65536,
-    timeCost: 3,
-    parallelism: 1,
-    hashLength: 32,
-    type: argon2.argon2id,
-    raw: true
-  });
-  if (!derivedKey) throw new Error("Failed to derive key from password");
-  const key = decryptSymmetric128BitHexKeyUTF8({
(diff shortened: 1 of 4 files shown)


backend/src/services/auth/auth-login-service.ts
```diff
@@ -258,7 +259,10 @@ export const authLoginServiceFactory = ({
     });
     // from password decrypt the private key
     if (password) {
-      const privateKey = await getUserPrivateKey(password, userEnc);
+      const privateKey = await getUserPrivateKey(password, userEnc).catch((err) => {
+        logger.error(err, `private key generation failed for user id - ${user.id} and email -${user.email}`);
```
maidul98, line 263: let's prefix with logs with the function they are being called from. this will help when we are querying

```
loginExchangeClientProof: private key generation failed for [userId=${user.id}] and [email=${user.email}] 
```

backend/src/lib/crypto/srp.ts
```diff
@@ -101,33 +101,51 @@ export const getUserPrivateKey = async (
...
-    iv: user.protectedKeyIV as string,
-    tag: user.protectedKeyTag as string,
-    key: derivedKey
-  });
+  if (user.encryptionVersion === 1) {
+    return decryptSymmetric128BitHexKeyUTF8({
+      ciphertext: user.encryptedPrivateKey,
+      iv: user.iv,
+      tag: user.tag,
+      key: password.slice(0, 32).padStart(32 + (password.slice(0, 32).length - new Blob([password]).size), "0")
```
maidul98, line 119: What was the train of thought of choosing this parsing?

akhilmhdh: I have no idea to be honest. 

This is what we used to do in frontend for encryption v1. I was not present at that time.
 
@tony may have some input over it

akhilmhdh: CC: @dangtony98

backend/src/services/auth/auth-login-service.ts
akhilmhdh: Done

backend/src/lib/crypto/srp.ts
maidul98: Oh i see it in the code, okay

maidul98 (approved):

PR #2004, opened 2024-06-21
feat(platform): Show imported/overwritten values in secret overview

DanielHougaard:
# Description 📣

1. Currently when a value is imported and the user attempts to view the value in the secret's overview. This PR aims to make the imported value visible to the user instead of just showing "IMPORTED". Additionally, we now also show _where_ the value is being imported from.
2. In the secrets overview you can't view your overwritten values. In this update the user can now view and update their overwritten values in the secret overview.
3. I realized we were using hardcoded values for `shared` and `personal` secret types across the frontend. I've gone ahead and added an explicit type for the secret type and implemented it across the frontend. 

#### Imported values
CleanShot 2024-06-21 at 15 36 43

#### Overwritten values

https://github.com/Infisical/infisical/assets/62331820/5f20c648-d132-462a-8701-b9a93f832e25

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/components/dashboard/DropZone.tsx
+++ b/frontend/src/components/dashboard/DropZone.tsx
@@ -6,6 +6,8 @@ import { faUpload } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { parseDocument, Scalar, YAMLMap } from "yaml";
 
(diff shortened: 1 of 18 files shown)


akhilmhdh (approved):
Looks good to me. 

If the tooltip that says the imported from a bit from longer it would be great.

PR #2006, opened 2024-06-21
feat(api): Expand single secret references

DanielHougaard:
# Description 📣

Currently we only support expansion when listing all secrets, but not when getting a singular secret. This PR adds support for secret expansion on a single secret. 

The implementation is borrowed from the existing secret expansion logic from the `getSecretsRaw` method. It has been tweaked to accept a single secret instead of an array.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/server/routes/v3/secret-router.ts
+++ b/backend/src/server/routes/v3/secret-router.ts
@@ -300,6 +300,11 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.GET.secretPath),
         version: z.coerce.number().optional().describe(RAW_SECRETS.GET.version),
         type: z.nativeEnum(SecretType).default(SecretType.Shared).describe(RAW_SECRETS.GET.type),
+        expandSecretReferences: z
+          .enum(["true", "false"])
+          .default("false")
+          .transform((value) => value === "true")
+          .describe(RAW_SECRETS.GET.expand),
         include_imports: z
           .enum(["true", "false"])
           .default("false")
@@ -344,6 +349,7 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         actor: req.permission.type,
         actorAuthMethod: req.permission.authMethod,
         actorOrgId: req.permission.orgId,
+        expandSecretReferences: req.query.expandSecretReferences,
         environment,
         projectId: workspaceId,
         projectSlug: workspaceSlug,
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -347,6 +347,7 @@ export const RAW_SECRETS = {
     tagIds: "The ID of the tags to be attached to the created secret."
   },
   GET: {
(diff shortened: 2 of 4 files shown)


backend/src/server/routes/v3/secret-router.ts
```diff
@@ -300,6 +300,11 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.GET.secretPath),
         version: z.coerce.number().optional().describe(RAW_SECRETS.GET.version),
         type: z.nativeEnum(SecretType).default(SecretType.Shared).describe(RAW_SECRETS.GET.type),
+        expandSecretReferences: z
+          .enum(["true", "false"])
+          .default("false")
+          .transform((value) => value === "true")
+          .describe(RAW_SECRETS.GET.expand),
```
sheensantoscapadngan, line 307: would it be more appropriate if this was singular?
since we're dealing with individual secrets here...

backend/src/server/routes/v3/secret-router.ts
```diff
@@ -300,6 +300,11 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
         secretPath: z.string().trim().default("/").transform(removeTrailingSlash).describe(RAW_SECRETS.GET.secretPath),
         version: z.coerce.number().optional().describe(RAW_SECRETS.GET.version),
         type: z.nativeEnum(SecretType).default(SecretType.Shared).describe(RAW_SECRETS.GET.type),
+        expandSecretReferences: z
+          .enum(["true", "false"])
+          .default("false")
```
sheensantoscapadngan, line 305: what do you think of defaulting to `true` instead? 

I think from a client POV, they ideally shouldn't have to worry about what secret references are and whether they should be expanded

DanielHougaard: @sheensantoscapadngan Yea I personally agree that we should default it to true. However to keep it inline with our existing implementation for the get all secrets endpoint, I opted to make it default to false. I feel like if we're going to make it default to true, it should also default to true on the list secrets endpoint

backend/src/server/routes/v3/secret-router.ts
DanielHougaard: I actually think it makes sense

A secret can contain multiple references, it's not **a** reference, it may be multiple references to other secrets

I.e: TEST-SECRET: ${dev.secret-1}<>${dev.secret-2}.${dev.other-secret}

sheensantoscapadngan: ahhh right, it can be viewed that way as well.. I was thinking of "secret reference" referring to the collective of references but that makes sense

backend/src/server/routes/v3/secret-router.ts
sheensantoscapadngan: yea... It becomes inconsistent if we start doing it this way which could be confusing. let's keep this as it is

sheensantoscapadngan (approved):

PR #2007, opened 2024-06-22
feat: added project setting for rebuilding secret indices

sheensantoscapadngan:
# Description 📣
- This will enable users to manually resolve secret indexing issues that they encounter upon secret modifications
Screenshot 2024-06-22 at 10 47 35 PM

Project page:
image

Note: the ideal scenario here is where blind indexing issues never occur in the first place. While these cases are under investigation, it would be great to have a way of letting users unblock themselves when necessary

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


akhilmhdh:
@maidul98 now we need to check whether there blind index is null or not before this

sheensantoscapadngan:
we've confirmed already that the user who reported this has secret indices for all his secrets @akhilmhdh

akhilmhdh (approved):

PR #2019, opened 2024-06-25
Postgres read replica support

akhilmhdh:
# Description 📣

This PR introduces the feature for Infisical to use postgres read replica. The write operation and transactions goes through the primary/master instance. The read operation go through read replica. 

If read replicas are not provided master instance is used.  

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/ee/services/dynamic-secret-lease/dynamic-secret-lease-dal.ts
```diff
@@ -12,7 +12,10 @@ export const dynamicSecretLeaseDALFactory = (db: TDbClient) => {
 
   const countLeasesForDynamicSecret = async (dynamicSecretId: string, tx?: Knex) => {
     try {
-      const doc = await (tx || db)(TableName.DynamicSecretLease).count("*").where({ dynamicSecretId }).first();
+      const doc = await (tx || db.replicaNode())(TableName.DynamicSecretLease)
```
maidul98, line 15: hmm, there is no better way to somehow do this dynamically without adding db.replicaNode() all through out the code base?

maidul98: Is this what drizzle does in their code base? I know there is a on going talk about it here on knex js https://github.com/knex/knex/issues/2253#issuecomment-551610832

akhilmhdh: Dynamically not possible as knexjs itself doesn't allow anything intermediate conversion. 

I have seen the above thread and many more. The biggest issue with all the issues is first its not reliable. People are reporting x or y not working. Transactions breaking etc. Second overriding something internal knexjs could also cause overhead maintenance in future if they change the structure. 

The `replicaNode()` need to be made only on read query that too custom written and not using orm.

akhilmhdh: https://github.com/drizzle-team/drizzle-orm/blob/4ecfe1fca3ec0bc8fdec02b6751e030b8832fb41/drizzle-orm/src/pg-core/db.ts#L619

backend/src/@types/knex.d.ts
```diff
@@ -1,4 +1,4 @@
-import { Knex } from "knex";
+import { Knex as KnexOriginal } from "knex";
```
maidul98, line 1: question: if KnexOriginal is the same as Knex, what does this do?

backend/src/db/instance.ts
```diff
@@ -1,8 +1,38 @@
...
+  knex.QueryBuilder.extend("primaryNode", () => {
+    return db;
+  });
+
+  // @ts-expect-error the querybuilder type is expected but our intension is to return  a knex instance
+  knex.QueryBuilder.extend("replicaNode", () => {
+    if (!readReplicaDbs.length) return db;
+
+    const selectedReplica = readReplicaDbs[Math.floor(Math.random() * readReplicaDbs.length)];
+    return selectedReplica;
```
maidul98, line 32: I updated to round robin so there is even distribution of queries to read dbs. Maybe in the future we can add this counter to Redis to make it better

maidul98 (approved):
Also tested on my end, works well

backend/src/@types/knex.d.ts
akhilmhdh: This was renamed because of this line

https://github.com/Infisical/infisical/pull/2019/files/5d59fe8810668afb205b3cff57a8396c4d902844#diff-ae28c47becba560585c4ced163583dbcbf8516e29b8d68a0282d1b870b5518b9R284

PR #2020, opened 2024-06-25
feat: enabled customization of project audit logs retention period

sheensantoscapadngan:
# Description 📣
- This PR will enable users to configure audit logs retention period per project. For org-level logs, retention is set to their plan's value

image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v1/project-router.ts
+++ b/backend/src/server/routes/v1/project-router.ts
@@ -383,7 +383,7 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
         workspaceSlug: z.string().trim()
       }),
       body: z.object({
-        auditLogsRetentionDays: z.number().min(1).max(100)
+        auditLogsRetentionDays: z.number().min(0)
       }),
       response: {
         200: z.object({
--- a/backend/src/server/routes/v1/project-router.ts
+++ b/backend/src/server/routes/v1/project-router.ts
@@ -404,7 +404,7 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
       });
 
       return {
-        message: "Successfull changed project audit log retention",
+        message: "Successfully updated project's audit logs retention period",
         workspace
       };
     }
--- a/backend/src/server/routes/v1/project-router.ts
+++ b/backend/src/server/routes/v1/project-router.ts
@@ -373,6 +373,44 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
   });
 
   server.route({
+    method: "PUT",
+    url: "/:workspaceSlug/audit-logs-retention",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      params: z.object({
+        workspaceSlug: z.string().trim()
+      }),
+      body: z.object({
+        auditLogsRetentionDays: z.number().min(1).max(100)
+      }),
+      response: {
+        200: z.object({
+          message: z.string(),
+          workspace: ProjectsSchema
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT]),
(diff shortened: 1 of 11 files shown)


backend/src/server/routes/v1/project-router.ts
```diff
@@ -372,6 +372,44 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
+    url: "/:workspaceSlug/audit-logs-retention",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      params: z.object({
+        workspaceSlug: z.string().trim()
+      }),
+      body: z.object({
+        auditLogsRetentionDays: z.number().min(1).max(100)
```
akhilmhdh, line 386: Let keep the max even smaller. 100 is more than 3 months a bit too much. 

@maidul98 Your opinion please?

backend/src/services/project/project-service.ts
```diff
@@ -446,6 +447,43 @@ export const projectServiceFactory = ({
...
+    actorId,
+    actorOrgId,
+    actorAuthMethod,
+    auditLogsRetentionDays,
+    workspaceSlug
+  }: TUpdateAuditLogsRetentionDTO) => {
+    const project = await projectDAL.findProjectBySlug(workspaceSlug, actorOrgId);
+    if (!project) {
+      throw new NotFoundError({
+        message: "Project not found."
```
akhilmhdh, line 461: Good job using `NotFoundError`. Good API convention

frontend/src/views/Settings/ProjectSettingsPage/components/AuditLogsRetentionSection/AuditLogsRetentionSection.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+import { createNotification } from "@app/components/notifications";
+import { Button, FormControl, Input, UpgradePlanModal } from "@app/components/v2";
+import { useProjectPermission, useSubscription, useWorkspace } from "@app/context";
+import { usePopUp } from "@app/hooks";
+import { ProjectMembershipRole } from "@app/hooks/api/roles/types";
+import { useUpdateWorkspaceAuditLogsRetention } from "@app/hooks/api/workspace/queries";
+
+const formSchema = z.object({
+  auditLogsRetentionDays: z.coerce.number().min(0)
```
akhilmhdh, line 13: Please add `max` here also

akhilmhdh (changes requested):

backend/src/server/routes/v1/project-router.ts
sheensantoscapadngan: actually, the max is dictated by the value of `auditLogsRetentionDays` in the project's subscriptio plan. We're also doing this check in the backend. What do you think? 
image

sheensantoscapadngan: i'm thinking some organizations might need more days, so to be flexible I didn't define a max in the schema but rather did a check against the subscription in the service layer instead

maidul98: Retention period to pricing is a seperate topic, we can open up a thread about it in slack @akhilmhdh . I think this setting should only really be shown to self hosted instances. On cloud, all the user could do is lower the retention days which i don't think they care to do

backend/src/services/project/project-service.ts
maidul98: nice! we should start to use this in other places too

akhilmhdh (approved):

akhilmhdh (approved):

PR #2026, opened 2024-06-26
feat: allowed toggling login options as admin

sheensantoscapadngan:
# Description 📣
- This PR will allow admins to adjust the allowed login methods from the Admin configuration page

image
image

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/views/admin/DashboardPage/AuthPanel.tsx
+++ b/frontend/src/views/admin/DashboardPage/AuthPanel.tsx
@@ -1,0 +1,251 @@
+import { Controller, useForm } from "react-hook-form";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { z } from "zod";
+
+import { createNotification } from "@app/components/notifications";
+import { Button, FormControl, Switch } from "@app/components/v2";
+import { useServerConfig } from "@app/context";
+import { useUpdateServerConfig } from "@app/hooks/api";
+import { LoginMethod } from "@app/hooks/api/admin/types";
+
+const formSchema = z.object({
+  isEmailEnabled: z.boolean(),
+  isGoogleEnabled: z.boolean(),
+  isGithubEnabled: z.boolean(),
+  isGitlabEnabled: z.boolean(),
+  isSamlEnabled: z.boolean(),
+  isLdapEnabled: z.boolean(),
+  isOidcEnabled: z.boolean()
+});
+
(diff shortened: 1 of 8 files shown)


frontend/src/views/admin/DashboardPage/AuthPanel.tsx
```diff
@@ -0,0 +1,252 @@
...
+          render={({ field, fieldState: { error } }) => {
+            return (
+              <FormControl isError={Boolean(error)} errorText={error?.message}>
+                <Switch
+                  id="email-enabled"
+                  onCheckedChange={(value) => field.onChange(value)}
+                  isChecked={field.value}
+                >
+                  <p className="w-24">Email</p>
+                </Switch>
```
maidul98, line 134: I think on the UI, we should show some warning when email is unchecked and tell them that they will be locked out if this is the only method of access the current user has. Or maybe we can do a check for that ourselves to make sure the user doesn't lock their selves out. What do you think?

maidul98:
I'm still seeing SCIM when the only thing i have on is email 
CleanShot 2024-06-27 at 20 32 15@2x

maidul98 (commented):
Code looks good, i also played aroud with it and it works well! Just left a small comment

maidul98:
CleanShot 2024-06-27 at 20 45 41@2x

I'm seeing a blank page when i click the sign up but i'm thinking it shouldn't show at all unless email auth or one of the auth SSO are enabled. What do you think? 

Also we'll have to think about for password reset link

frontend/src/views/admin/DashboardPage/AuthPanel.tsx
sheensantoscapadngan: yeah for this one I think we could just do a check against the admin user and then throw an error if the user has none of the remaining enabled login options

sheensantoscapadngan: image

sheensantoscapadngan: updated text

image

sheensantoscapadngan:
Sign up text is now hidden when none of the ff is enabled: email, github, gitlab, google
image

maidul98 (approved):
lgtm

PR #2027, opened 2024-06-26
AWS Secret Manager assume role based integration

akhilmhdh:
# Description 📣

PR introduces aws secret manager integration to use [assume role](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_third-party.html) feature to avoid sharing client's access key and secret key.

integration-aws-iam-assume-select

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/db/migrations/20240626111536_integration-auth-aws-assume-role.ts
```diff
@@ -0,0 +1,35 @@
...
+  );
+  const hasAwsAssumeRoleIV = await knex.schema.hasColumn(TableName.IntegrationAuth, "awsAssumeIamRoleArnIV");
+  const hasAwsAssumeRoleTag = await knex.schema.hasColumn(TableName.IntegrationAuth, "awsAssumeIamRoleArnTag");
+  if (await knex.schema.hasTable(TableName.IntegrationAuth)) {
+    await knex.schema.alterTable(TableName.IntegrationAuth, (t) => {
+      if (!hasAwsAssumeRoleCipherText) t.text("awsAssumeIamRoleArnCipherText");
+      if (!hasAwsAssumeRoleIV) t.text("awsAssumeIamRoleArnIV");
+      if (!hasAwsAssumeRoleTag) t.text("awsAssumeIamRoleArnTag");
+    });
+  }
```
sheensantoscapadngan, line 18: what do you think about adding a metadata/methods JSON field in the IntegrationAuth table instead so that we could add integration auth-method specific details there? 

It might make sense to prepare for the future by making a generic field that can be reused by other auth methods for other integrations.

backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -695,24 +698,61 @@ const syncSecretsAWSSecretManager = async ({
...
+  if (awsAssumeRoleArn) {
+    if (!appCfg.CLIENT_ID_AWS_INTEGRATION || !appCfg.CLIENT_SECRET_AWS_INTEGRATION)
+      throw new Error("Missing AWS Credentials for Assuming Role. Contact server admin.");
+
+    const client = new STSClient({
+      region: integration.region as string,
+      credentials: {
+        accessKeyId: appCfg.CLIENT_ID_AWS_INTEGRATION,
+        secretAccessKey: appCfg.CLIENT_SECRET_AWS_INTEGRATION
+      }
```
sheensantoscapadngan, line 731: for instances deployed to AWS, I think we should provide an option so that they don't have to explicitly define AWS credentials? 

Like, it should attempt to use the credentials provided in the AWS workload instance if not explicitly defined in the env.

docs/integrations/cloud/aws-secret-manager-assume-role.mdx
```diff
@@ -0,0 +1,144 @@
+---
+title: "AWS Secrets Manager Assume Role"
+description: "Learn how to sync secrets from Infisical to AWS Secrets Manager without sharing any user credentials."
+---
```
sheensantoscapadngan, line 4: what if we moved this to the **AWS Secret Manager** doc as a separate tab instead of having a completely separate page for it? 

Since this is also AWS's best practice recommendation, I think we could make it the first tab (and possibly add "We recommend this method for security purposes" somewhere)

image

docs/integrations/cloud/aws-secret-manager.mdx
```diff
@@ -51,13 +51,13 @@ Prerequisites:
...
+       ![Select AWS Secrets Manager](../../images/integrations.png)
 
-    ![integrations](../../images/integrations.png)
-
-    Press on the AWS Secrets Manager tile and input your AWS access key ID and secret access key from the previous step.
-
-    ![integration auth](../../images/integrations/aws/integrations-aws-secret-manager-auth.png)
+    3. Select the **Access Key** option for Authentication Mode.
+       ![Select Access Key](../../images/integrations/aws/integrations-aws-secret-manager-auth.png)
+    4. Provide the **access key** and **secret key** for the AWS Iam User.
```
sheensantoscapadngan, line 60: If in a different page, there should at least be a link here to forward them to the AWS Assume doc? Maybe in something like a <Tip>

frontend/src/pages/integrations/aws-secret-manager/authorize.tsx
```diff
@@ -1,54 +1,70 @@
...
+  z.object({
+    type: z.literal(AwsAuthType.AccessKey),
+    accessKey: z.string().min(1),
+    accessSecretKey: z.string().min(1)
+  }),
+  z.object({
+    type: z.literal(AwsAuthType.AssumeRole),
+    iamRoleArn: z.string().min(1)
+  })
+]);
```
sheensantoscapadngan, line 38: today I learned about this haha. Nice!

backend/src/db/migrations/20240626111536_integration-auth-aws-assume-role.ts
akhilmhdh: There is already something called metadata. But technically this is not a metadata rather actual data. The whole integration table design needs a revamped and has quite some flaws including issues like using it in API etc. 

Something for the future i guess

docs/integrations/cloud/aws-secret-manager-assume-role.mdx
akhilmhdh: Hmmm, good point.

backend/src/services/integration-auth/integration-sync-secret.ts
akhilmhdh: Yah i can do that. Just remove the check 😅 and that's all i have to do. AWS SDK tries to find all the various possible ways to load the cred by itself

docs/integrations/cloud/aws-secret-manager.mdx
akhilmhdh: Resolved by moving to tab

backend/src/services/integration-auth/integration-sync-secret.ts
akhilmhdh: Removed the explicit check. AWS SDK will automatically pick it up

backend/src/db/migrations/20240626111536_integration-auth-aws-assume-role.ts
sheensantoscapadngan: got this, yea that makes sense!

sheensantoscapadngan (approved):

akhilmhdh:
@maidul98 merge this only when you have enough time to update and check things in staging.

PR #2032, opened 2024-06-27
Resolved identity roleId not setting null for predefined role selection

akhilmhdh:
# Description 📣

This PR resolves a present bug in which when switching a role of an identity from a custom role to a predefined role like `developer`, `admin` the role id not setting it to null. This causes the role deletion failing due to tangling reference

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/identity/identity-service.ts
+++ b/backend/src/services/identity/identity-service.ts
@@ -115,7 +115,7 @@ export const identityServiceFactory = ({
           { identityId: id },
           {
             role: customRole ? OrgMembershipRole.Custom : role,
-            roleId: customRole?.id
+            roleId: customRole?.id || null
           },
           tx
         );


backend/src/services/identity/identity-service.ts
```diff
@@ -115,7 +115,7 @@ export const identityServiceFactory = ({
           { identityId: id },
           {
             role: customRole ? OrgMembershipRole.Custom : role,
-            roleId: customRole?.id
+            roleId: customRole?.id || null
```
maidul98, line 118: If customRole is not defined then it will be `undefined`, right? I would think Knex would treat `undefined` like null under the hood

maidul98 (approved):

backend/src/services/identity/identity-service.ts
akhilmhdh: The difference is undefined in update means there is nothing and knexjs ignore that field.

Where as null is something else in SQL. Null is empty in SQL world. So in update if there customRole is empty then we set that field as null. 

The issue was before consider user set a custom role and then switched back to a predefined the customRole would be undefined. So the `roleId` was undefined, which basically made knexjs just ignore it rather than unsetting it causing tangling foreign key

PR #2040, opened 2024-06-28
feat: allow org members to mark projects as favorites

sheensantoscapadngan:
# Description 📣
- This PR will enable org members to create project favorites which will be listed accordingly in the ff UIs

image
image
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/user/user-service.ts
+++ b/backend/src/services/user/user-service.ts
@@ -8,6 +8,7 @@ import { SmtpTemplates, TSmtpService } from "@app/services/smtp/smtp-service";
 import { TUserAliasDALFactory } from "@app/services/user-alias/user-alias-dal";
 
 import { AuthMethod } from "../auth/auth-type";
+import { TProjectMembershipDALFactory } from "../project-membership/project-membership-dal";
 import { TUserDALFactory } from "./user-dal";
 
 type TUserServiceFactoryDep = {
@@ -26,8 +27,9 @@ type TUserServiceFactoryDep = {
     | "delete"
   >;
   userAliasDAL: Pick<TUserAliasDALFactory, "find" | "insertMany">;
-  orgMembershipDAL: Pick<TOrgMembershipDALFactory, "find" | "insertMany">;
+  orgMembershipDAL: Pick<TOrgMembershipDALFactory, "find" | "insertMany" | "findOne" | "updateById">;
   tokenService: Pick<TAuthTokenServiceFactory, "createTokenForUser" | "validateTokenForUser">;
+  projectMembershipDAL: Pick<TProjectMembershipDALFactory, "find">;
   smtpService: Pick<TSmtpService, "sendMail">;
 };
 
@@ -37,6 +39,7 @@ export const userServiceFactory = ({
   userDAL,
   userAliasDAL,
   orgMembershipDAL,
+  projectMembershipDAL,
   tokenService,
   smtpService
 }: TUserServiceFactoryDep) => {
@@ -247,6 +250,49 @@ export const userServiceFactory = ({
     return privateKey;
   };
 
+  const getUserProjectFavorites = async (userId: string, orgId: string) => {
+    const orgMembership = await orgMembershipDAL.findOne({
+      userId,
+      orgId
+    });
+
+    if (!orgMembership) {
+      throw new BadRequestError({
+        message: "User does not belong in the organization."
+      });
+    }
+
+    return { projectFavorites: orgMembership.projectFavorites || [] };
+  };
+
+  const updateUserProjectFavorites = async (userId: string, orgId: string, projectIds: string[]) => {
+    const orgMembership = await orgMembershipDAL.findOne({
+      userId,
+      orgId
+    });
+
+    if (!orgMembership) {
+      throw new BadRequestError({
+        message: "User does not belong in the organization."
+      });
+    }
+
+    const memberProjectFavorites = (
+      await projectMembershipDAL.find({
+        userId,
+        $in: {
+          projectId: projectIds
+        }
+      })
+    ).map((projectMembership) => projectMembership.projectId);
+
+    return (
+      await orgMembershipDAL.updateById(orgMembership.id, {
+        projectFavorites: memberProjectFavorites
+      })
+    ).projectFavorites;
+  };
+
   return {
     sendEmailVerificationCode,
     verifyEmailVerificationCode,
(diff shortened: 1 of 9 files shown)


frontend/src/layouts/AppLayout/AppLayout.tsx
```diff
@@ -271,6 +280,30 @@ export const AppLayout = ({ children }: LayoutProps) => {
     }
   };
 
+  const addProjectToFavorites = useCallback(
```
akhilmhdh, line 283: A bit confused by this usecallback?

akhilmhdh: And the below one

frontend/src/layouts/AppLayout/AppLayout.tsx
```diff
@@ -451,19 +484,47 @@ export const AppLayout = ({ children }: LayoutProps) => {
...
+                                    {name}
+                                  </SelectItem>
+                                </div>
+                                <div className="col-span-1 flex items-center">
+                                  {projectFavorites?.includes(id) ? (
+                                    <FontAwesomeIcon
+                                      icon={faSolidStar}
+                                      className="text-sm text-mineshaft-300 hover:text-mineshaft-400"
+                                      onClick={(e) => {
+                                        e.stopPropagation();
```
akhilmhdh, line 512: Missing try catch and notification for both remove and add

frontend/src/pages/org/[id]/overview/index.tsx
```diff
@@ -569,52 +575,160 @@ const OrganizationPage = withPermission(
     const filteredWorkspaces = orgWorkspaces.filter((ws) =>
       ws?.name?.toLowerCase().includes(searchFilter.toLowerCase())
     );
+    const favoriteWorkspaces = filteredWorkspaces.filter((ws) => projectFavorites?.includes(ws.id));
+    const nonFavoriteWorkspaces = filteredWorkspaces.filter((ws) =>
+      favoriteWorkspaces.every((entry) => entry.id !== ws.id)
+    );
+
+    const addProjectToFavorites = useCallback(
```
Comment (akhilmhdh) [inline frontend/src/pages/org/[id]/overview/index.tsx:583]: Same doubt - The purpose of useCallback

frontend/src/pages/org/[id]/overview/index.tsx
```diff
@@ -630,29 +744,13 @@ const OrganizationPage = withPermission(
...
-              <div className="ml-5 truncate text-sm text-mineshaft-100">{workspace.name}</div>
-            </div>
-            <div className="flex items-center justify-end sm:col-span-3 lg:col-span-2">
-              <div className="text-center text-sm text-mineshaft-300">
-                {workspace.environments?.length || 0} environments
-              </div>
-            </div>
-          </div>
-        ))}
+        {[...favoriteWorkspaces, ...nonFavoriteWorkspaces].map((workspace, ind) =>
```
Comment (akhilmhdh) [inline frontend/src/pages/org/[id]/overview/index.tsx:747]: I think this is unnecessarily complex. Just do a useMemo(if the list could be big or computation is a lot). Then just sort the workspace list by favorite or not and add a computed field `isFavourite`

frontend/src/pages/org/[id]/overview/index.tsx
```diff
@@ -569,52 +575,160 @@ const OrganizationPage = withPermission(
...
-              <div className="group ml-auto w-max cursor-pointer rounded-full border border-mineshaft-600 bg-mineshaft-900 py-2 px-4 text-sm text-mineshaft-300 transition-all group-hover:border-primary-500/80 group-hover:bg-primary-800/20 group-hover:text-mineshaft-200">
-                Explore{" "}
-                <FontAwesomeIcon
-                  icon={faArrowRight}
-                  className="pl-1.5 pr-0.5 duration-200 group-hover:pl-2 group-hover:pr-0"
-                />
+          </>
+        )}
+        <div className="mt-4 grid w-full grid-cols-1 gap-4 lg:grid-cols-2 xl:grid-cols-3 2xl:grid-cols-4">
+          {isWorkspaceLoading &&
```
Comment (akhilmhdh) [inline frontend/src/pages/org/[id]/overview/index.tsx:712]: Wait for `isFavorite` api call to finish loading. Else the list would jump if the second api call is delayed. 

Because consider the workspace loading finished. You will render everything as non favorite as favorites are not loaded yet. Then favorites would load and again list gets modified causing a jump effect.

akhilmhdh (changes requested):

frontend/src/layouts/AppLayout/AppLayout.tsx
sheensantoscapadngan: ah right this isn't necessary. will remove

frontend/src/pages/org/[id]/overview/index.tsx
Comment (sheensantoscapadngan) [reply frontend/src/pages/org/[id]/overview/index.tsx]: made changes in this component. I kept the one in `AppLayout` as it is because it's much simpler doing it this way there. I also added useMemo

frontend/src/pages/org/[id]/overview/index.tsx
```diff
@@ -570,56 +577,191 @@ const OrganizationPage = withPermission(
       ws?.name?.toLowerCase().includes(searchFilter.toLowerCase())
     );
 
+    const workspacesWithFaveProp = useMemo(
```
Comment (akhilmhdh) [inline frontend/src/pages/org/[id]/overview/index.tsx:580]: Same here, why not one single array of fav first and later non fav.

This would also mean the below logic is  much more simple

akhilmhdh (changes requested):

frontend/src/pages/org/[id]/overview/index.tsx
Comment (sheensantoscapadngan) [reply frontend/src/pages/org/[id]/overview/index.tsx:580]: For this one, I need the favorite and nonfavorite workspaces as two separate arrays because I'm using both of them separately, unlike the one in the AppLayout component.
image

I think it's more simpler this way rather than have one list as you're suggesting and then doing filtering in its usage

Comment (akhilmhdh) [reply frontend/src/pages/org/[id]/overview/index.tsx:580]: Exactly, 😅 you don't need seperate list. 

What you need is a flag `isFavorite`

akhilmhdh (dismissed):

backend/src/services/user/user-service.ts
```diff
@@ -247,6 +250,49 @@ export const userServiceFactory = ({
...
+      await projectMembershipDAL.find({
+        userId,
+        $in: {
+          projectId: projectIds
+        }
+      })
+    ).map((projectMembership) => projectMembership.projectId);
+
+    return (
+      await orgMembershipDAL.updateById(orgMembership.id, {
```
akhilmhdh, line 290: Kindly seperate these into variable names please. This part is really hard to understand

akhilmhdh (approved):

PR #2041, opened 2024-06-28
feat: Infisical Binary

DanielHougaard:
# Description 📣

As per requested by users, we now support running Infisical as a standalone binary. Previously it wasn't possible to deploy Infisical anywhere without Docker and various build-steps, increasing the self-hosting complexity. With the binary, all the user needs to do is:

1. Set environment variables
2. Run the binary file (./binary-file)

This is achieved through the use of [vercel/pkg](https://github.com/vercel/pkg). One trade off however is that vercel/pkg doesn't support ESM. So we need to convert our ESM code to CJS. This is handled as a part of the build step. This however means that I also had to downgrade two of our internal packages (nanoid and slugify), as the newer versions are pure ESM.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,90 @@
...
+                  echo "::set-output name=version::$VERSION"
+
+            - name: Version extration test
+              run: |
+                  echo "The extracted version is ${{ steps.extract_version.outputs.version }}"
+
+            - name: Set up Node.js
+              uses: actions/setup-node@v3
+              with:
+                  node-version: 18
```
maidul98, line 55: Should we now set this to version 20?

maidul98 (dismissed):
lgtm, i think there is a package lock conflict. otherwise good to go

akhilmhdh (changes requested):
please change action to node 20.

Rest everything looks all good

.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,104 @@
...
+jobs:
+    build-and-deploy:
+        runs-on: ubuntu-20.04
+        strategy:
+            matrix:
+                arch: [x64, arm64]
+                os: [linux, win]
+                include:
+                    - os: linux
+                      target: node18-linux
```
maidul98, line 19: 18 or 20 here?

maidul98 (approved):
lgtm

maidul98 (approved):
lgtm!

vhsantos:
Hi,

Do you have any estimate when the binaries will be available ?? 
Thanks in advance.

DanielHougaard:
Hi @vhsantos,

I'm finishing up the deployment and documentation for the binaries. Currently only Debian/Ubuntu is supported. Which system are you looking to use the binaries on?

vhsantos:
Hi @DanielHougaard 

Actually I'm looking for RHEL/AlmaLinux 9.
Maybe if you share with me the deployment/documentation I can help with something (init scripts, testing, etc) on RHEL ??

DanielHougaard:
Hi @vhsantos,

The only thing that I see as a concern for RHEL, is [yao-pkg/pkg](https://github.com/yao-pkg/pkg)'s ability to package to that distro. If we're able to package nodejs to the RHEL distro, then it's very straight forward from there!

PR #2061, opened 2024-07-03
Secret approval with groups

akhilmhdh:
# Description 📣

This PR enables secret approval to work with users part of groups. Instead of project membership we now take user id in secret approval. The api now returns the corresponding user details for various user entities like statusChangeBy, committer etc

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

- [ x] I have read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/getting-started/code-of-conduct). 📝

### DIFF
--- a/backend/src/db/migrations/20240702131735_secret-approval-groups.ts
+++ b/backend/src/db/migrations/20240702131735_secret-approval-groups.ts
@@ -1,0 +1,188 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  // migrate secret approval policy approvers to user id
+  const hasApproverUserId = await knex.schema.hasColumn(TableName.SecretApprovalPolicyApprover, "approverUserId");
+  const hasApproverId = await knex.schema.hasColumn(TableName.SecretApprovalPolicyApprover, "approverId");
+  if (!hasApproverUserId) {
+    // add the new fields
+    await knex.schema.alterTable(TableName.SecretApprovalPolicyApprover, (tb) => {
+      // if (hasApproverId) tb.setNullable("approverId");
+      tb.uuid("approverUserId");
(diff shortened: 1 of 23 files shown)


akhilmhdh:
@maidul98 After review, merge when your also available as this is migration change

maidul98:
@sheensantoscapadngan ping me in slack when you are done with the review and i'll skim over it before deploy

sheensantoscapadngan (commented):
I checked these ones and noticed these are all `AuthMode.JWT`. It should be safe to break these right since it's all supposed to be used via Web?

backend/src/db/migrations/20240702131735_secret-approval-groups.ts
```diff
@@ -0,0 +1,188 @@
...
+      // @ts-ignore because generate schema happens after this
+      approverUserId: knex(TableName.ProjectMembership)
+        .select("userId")
+        .where("id", knex.raw("??", [`${TableName.SecretApprovalPolicyApprover}.approverId`]))
+    });
+    // drop the old field
+    await knex.schema.alterTable(TableName.SecretApprovalPolicyApprover, (tb) => {
+      if (hasApproverId) tb.dropColumn("approverId");
+      tb.uuid("approverUserId").notNullable().alter();
+    });
```
sheensantoscapadngan, line 29: hmmm should we move these out of the `if (!hasApproverUserId)` and put it into its own `hasApproverUserId` condition? 

It might occur that the update of SecretApprovalPolicyApprover fails but approverUserId is already created so the drop field never gets executed.

sheensantoscapadngan:
group members not shown in policy user selection 
image

resolved: another effort

sheensantoscapadngan (approved):

maidul98 (approved):
Good idea to go from project membership to users. We'll need to do something similar for access requests (not yet though)

PR #2086, opened 2024-07-08
feat: allow admins to delete users

sheensantoscapadngan:
# Description 📣
- This PR will allow admins to delete users from the Admin panel
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v1/admin-router.ts
+++ b/backend/src/server/routes/v1/admin-router.ts
@@ -84,6 +84,82 @@ export const registerAdminRouter = async (server: FastifyZodProvider) => {
   });
 
   server.route({
+    method: "GET",
+    url: "/user-management/users",
+    config: {
+      rateLimit: readLimit
+    },
+    schema: {
+      querystring: z.object({
+        searchTerm: z.string().default(""),
+        offset: z.coerce.number().default(0),
+        limit: z.coerce.number().default(20)
+      }),
+      response: {
+        200: z.object({
+          users: UsersSchema.pick({
+            username: true,
+            firstName: true,
+            lastName: true,
+            email: true,
+            id: true
+          }).array()
+        })
+      }
+    },
+    onRequest: (req, res, done) => {
+      verifyAuth([AuthMode.JWT])(req, res, () => {
+        verifySuperAdmin(req, res, done);
+      });
+    },
+    handler: async (req) => {
+      const users = await server.services.superAdmin.getUsers({
+        ...req.query
+      });
+
+      return {
+        users
+      };
+    }
+  });
+
+  server.route({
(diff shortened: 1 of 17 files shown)


backend/src/server/routes/v1/user-router.ts
```diff
@@ -91,6 +92,75 @@ export const registerUserRouter = async (server: FastifyZodProvider) => {
...
+      });
+    },
+    handler: async () => {
+      const users = await server.services.user.listUsers();
+
+      return {
+        users
+      };
+    }
+  });
```
sheensantoscapadngan, line 126: I dont think any pagination is needed for this one because the intended usage is for dedicated/self-hosted instances

akhilmhdh: Please add pagination as loading entire user list is a bad idea. Even though indention is self hosted if this page is opened in production this could turn up bad.

Another thing is please move this to admin route. All admin panel routing should go through admin service. This is to ensure as admin operations are peak security and should be kept different

sheensantoscapadngan:
Do not return ghost users

frontend/src/views/admin/DashboardPage/UserPanel.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+    handlePopUpClose("removeUser");
+  };
+
+  return (
+    <div className="mb-6 rounded-lg border border-mineshaft-600 bg-mineshaft-900 p-4">
+      <div className="mb-4">
+        <p className="text-xl font-semibold text-mineshaft-100">Users</p>
+      </div>
+      <Input
```
akhilmhdh, line 144: You can move this inside the `UserPanel` and the searchUserFilter state also

backend/src/server/routes/v1/admin-router.ts
```diff
@@ -83,6 +83,82 @@ export const registerAdminRouter = async (server: FastifyZodProvider) => {
...
+    method: "GET",
+    url: "/user-management/users",
+    config: {
+      rateLimit: readLimit
+    },
+    schema: {
+      querystring: z.object({
+        searchTerm: z.string().default(""),
+        offset: z.coerce.number().default(0),
+        limit: z.coerce.number().default(20)
```
akhilmhdh, line 96: A max on the API limit argument

akhilmhdh (changes requested):

akhilmhdh (approved):
Looks good to me. Cross check with db any stale data exist like orphan data that has missing onCascade etc

Issue #2098, opened 2024-07-10
Bad TTL in Machine Identity Secret Renders Web App Inoperable


alex0112:
### Describe the bug
Setting a high TTL on a machine identity secret renders the app inoperable

### To Reproduce
1. Create a new machine identity
2. Generate a secret associated with the machine identity
3. Set the TTL to 9999999999 or similar
4. Observe the error that pops up in a toast notification, and refresh the page to produce the screenshots below

### Expected behavior
Either a limit on the TTL 

### Screenshots

![image](https://github.com/Infisical/infisical/assets/7142972/f93d006d-b717-4e06-a97f-92d82915c2bc)

### Platform you are having the issue on:
Web App (Firefox). Self Hosted. 

### Additional context
Purely speculation on my part: but it seems that a bad value for the TTL gets written to the database and the front end app is unable to handle the high value when it tries to render the token again. At least that seems consistent with my observations.

DanielHougaard:
Hi @alex0112

Thank you for the report. We have this on our internal todo list currently, we'll get back to you as soon as a fix for this is live!

sheensantoscapadngan:
Hey @alex0112! This should be resolved now in the latest release

PR #2100, opened 2024-07-11
misc: add max checks for TTL values of identities

sheensantoscapadngan:
# Description 📣
- This PR adds validation for TTL values so that int overflow won't occur like in the issue reported here https://github.com/Infisical/infisical/issues/2098

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v1/identity-aws-iam-auth-router.ts
+++ b/backend/src/server/routes/v1/identity-aws-iam-auth-router.ts
@@ -100,6 +100,7 @@ export const registerIdentityAwsAuthRouter = async (server: FastifyZodProvider) 
           .number()
           .int()
           .min(1)
+          .max(31536000)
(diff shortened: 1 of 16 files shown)


backend/src/server/routes/v1/identity-aws-iam-auth-router.ts
```diff
@@ -100,6 +100,7 @@ export const registerIdentityAwsAuthRouter = async (server: FastifyZodProvider)
           .number()
           .int()
           .min(1)
+          .max(31536000)
```
maidul98, line 103: Only 11 months? I think we can handle 10 years without issues right?

maidul98 (approved):

PR #2123, opened 2024-07-14
Switched CLI and K8s operator to secret raw endpoint

akhilmhdh:
# Description 📣

This PR changes all the endpoints of using secrets for CLI and k8s operator to secrets raw endpoint. This is the first step to infisical transition towards the new architecture

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


cli/packages/cmd/secrets.go
```diff
@@ -189,10 +189,24 @@ var secretsSetCmd = &cobra.Command{
 		if token != nil && (token.Type == util.SERVICE_TOKEN_IDENTIFIER || token.Type == util.UNIVERSAL_AUTH_TOKEN_IDENTIFIER) {
 			secretOperations, err = util.SetRawSecrets(args, secretType, environmentName, secretsPath, projectId, token)
 		} else {
-			util.RequireLogin()
-			util.RequireLocalWorkspaceFile()
+			workspaceFile, err := util.GetWorkSpaceFromFile()
```
maidul98, line 192: Why did we remove?

```
			util.RequireLogin()
			util.RequireLocalWorkspaceFile()
```

cli/packages/api/api.go
```diff
@@ -225,25 +225,6 @@ func CallIsAuthenticated(httpClient *resty.Client) bool {
 	return true
 }
 
-func CallGetAccessibleEnvironments(httpClient *resty.Client, request GetAccessibleEnvironmentsRequest) (GetAccessibleEnvironmentsResponse, error) {
```
maidul98, line 228: Why are we deleting this?

cli/packages/util/secrets.go
```diff
@@ -1116,6 +729,7 @@ func SetRawSecrets(secretArgs []string, secretType string, environmentName strin
 			Environment: environmentName,
 		}
 
+		fmt.Println(projectId, environmentName)
```
maidul98, line 732: Needed?

k8-operator/packages/api/api.go
```diff
@@ -51,43 +31,6 @@ func CallGetServiceTokenDetailsV2(httpClient *resty.Client) (GetServiceTokenDeta
 	return tokenDetailsResponse, nil
 }
 
-func CallGetSecretsV3(httpClient *resty.Client, request GetEncryptedSecretsV3Request) (GetEncryptedSecretsV3Response, error) {
```
maidul98, line 54: Hmm so one issue we might have is if people update their clients without updating infisical-core

maidul98 (commented):
One issue i'm concerned about is people updating their CLI because there is a new update notification that pops up but they haven't updated infisical-core. Same for the k8s operator. Thoughts?

Also have you tested all the CLI actions and k8s change locally?

Also seeing CLI tests failing

cli/packages/util/secrets.go
akhilmhdh: Removed

cli/packages/cmd/secrets.go
akhilmhdh: Instead i am explicitly doing a check for that and getting the logged in credential

cli/packages/api/api.go
akhilmhdh: Not used

k8-operator/packages/api/api.go
akhilmhdh: Doesn't matter because infisical core supported raw a long time ago. I mean raw endpoint was introduced long time back right?

maidul98 (approved):
set command didn't work so i fixed it. Left comment in slack before we merge. Looks good otherwise

maidul98 (approved):

maidul98 (approved):

PR #2144, opened 2024-07-17
feat: added project data key

sheensantoscapadngan:
# Description 📣
- This PR adds the generation of project data key for secret encryption/decryption

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/kms/kms-service.ts
+++ b/backend/src/services/kms/kms-service.ts
@@ -1,7 +1,7 @@
 import slugify from "@sindresorhus/slugify";
 import { Knex } from "knex";
 
-import { TKeyStoreFactory } from "@app/keystore/keystore";
+import { KeyStorePrefixes, TKeyStoreFactory } from "@app/keystore/keystore";
 import { getConfig } from "@app/lib/config/env";
 import { randomSecureBytes } from "@app/lib/crypto";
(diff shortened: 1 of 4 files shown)


backend/src/services/kms/kms-service.ts
```diff
@@ -167,35 +167,120 @@ export const kmsServiceFactory = ({
...
-      if (!project) {
-        throw new BadRequestError({ message: "Project not found" });
+    let project = await projectDAL.findById(projectId);
+    if (!project) {
+      throw new BadRequestError({ message: "Project not found" });
+    }
+
+    if (!project.kmsSecretManagerKeyId) {
+      // create default kms key for certificate service
+      const lock = await keyStore
```
maidul98, line 177: We do this so that multiple requests to create the default KMS doesn't create conflict?

sheensantoscapadngan: yup exactly

sheensantoscapadngan: last thing we want is for data key to be encrypted by a stale KMS

maidul98 (approved):
lgtm

PR #2146, opened 2024-07-18
feat: enhance approval policies

aheruz:
# Description 📣

Enhance Approval Workflow:
- Merge Policies Tabs
- Implement policy's enforcements levels
- Implement by-pass for soft enforcement requests
- Adding visual counter to requests tab

Video demo: https://www.loom.com/share/e13d9ccc3cd64d9b99a432e3877f53a4?sid=5bc15a57-eb94-4ca8-873c-88cae3ad4ef6

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


dangtony98:
This looks very promising!

Two suggestions:
- Ellipses Icon (instead of Edit/Delete Icons): I would change the edit and delete buttons on each row in the policies to be an ellipses with a dropdown menu with edit and delete options; I've been working toward standardizing the UI this way — You can see an example of this in the [Identity Table](https://github.com/Infisical/infisical/blob/95a46617870f1e6d5675706be2e33e444f8269f4/frontend/src/views/Org/MembersPage/components/OrgIdentityTab/components/IdentitySection/IdentityTable.tsx#L39).
- Replace Soft Delete yellow warning with a question mark next to Enforcement Level where users can hover over and understand the difference between Hard/Soft Delete; this would seem to be a more standardized pattern that we can more easily apply elsewhere.

frontend/src/helpers/policies.ts
```diff
@@ -0,0 +1,12 @@
+import { PolicyType } from "@app/hooks/api/policies/enums";
+
```
akhilmhdh, line 2: We are trying to move away from all these helpers things because we want to follow co-location principle. 

Its better to keep it in there places that keeping it here

akhilmhdh (changes requested):
Please do a type check and linting on frontend and backend.

Another important point is @maidul98 i think we need a better terminology. The enforcement hard and soft is not at all clear for me. I had to see how it works to understand the difference. May be the description needs improvement. Can we use something simple

aheruz:
@dangtony98 Implemented ; )
Screenshot 2024-07-18 at 17 38 49
Screenshot 2024-07-18 at 17 39 15

frontend/src/helpers/policies.ts
aheruz: There is not a clear place where to define this shared enums. I think we should follow the current approach until we define a better one.

backend/src/db/migrations/20240717184929_add-enforcement-level-secrets-policies.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from "knex";
+
+import { EnforcementLevel } from "@app/lib/types";
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.table("secret_approval_policies", (table) => {
+    table.specificType("enforcementLevel", "VARCHAR(10)").notNullable().defaultTo(EnforcementLevel.Hard);
```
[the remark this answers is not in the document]
aheruz: Fixed!

akhilmhdh (dismissed):

maidul98 (approved):
Looks much much better than what we had before, grea work!
 
Two things we missed in this PR but can be done in a seperate PR is making users provide a reason for breaking glass via soft enforcement and notifying project admins about this event

PR #2152, opened 2024-07-19
feat(ENG-985): secret share within organization

aheruz:
# Description 📣

Adding accessType on secret sharing:
- add `accessType` column to `secret_sharing`
- expose `accessType`
- adding 'General Access' to select access type on secret sharing creation
- enforce access by organization on backend request

## Type ✨

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

N/A

---


backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -62,13 +67,14 @@ export const secretSharingServiceFactory = ({
...
-      orgId
+      orgId,
+      accessType
     });
     return { id: newSharedSecret.id };
   };
 
   const createPublicSharedSecret = async (createSharedSecretInput: TCreatePublicSharedSecretDTO) => {
-    const { encryptedValue, iv, tag, hashedHex, expiresAt, expiresAfterViews } = createSharedSecretInput;
+    const { encryptedValue, iv, tag, hashedHex, expiresAt, expiresAfterViews, accessType } = createSharedSecretInput;
```
maidul98, line 77: We can assume for secret shares that are created from the public page are always open to the public. Only secret shares that are created from within the org can choose access types

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -105,9 +112,21 @@ export const secretSharingServiceFactory = ({
...
   };
 
-  const getActiveSharedSecretByIdAndHashedHex = async (sharedSecretId: string, hashedHex: string) => {
+  const getActiveSharedSecretByIdAndHashedHex = async (sharedSecretId: string, hashedHex: string, orgId?: string) => {
     const sharedSecret = await secretSharingDAL.findOne({ id: sharedSecretId, hashedHex });
     if (!sharedSecret) return;
+
+    const orgName = sharedSecret.orgId ? (await orgDAL.findOrgById(sharedSecret.orgId))?.name : "";
+    // Support organization level access for secret sharing
+    if (sharedSecret.accessType === SecretSharingAccessType.Organization && orgId !== sharedSecret.orgId) {
```
maidul98, line 121: I think it might make more sense to throw an error here because the user is trying to view a secret share for a different org than they are logged in as?

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -118,7 +137,10 @@ export const secretSharingServiceFactory = ({
       }
       await secretSharingDAL.updateById(sharedSecretId, { $decr: { expiresAfterViews: 1 } });
     }
-    return sharedSecret;
+    if (sharedSecret.accessType === SecretSharingAccessType.Organization) {
```
maidul98, line 140: I think we should still check to make sure the org of the logged in user match up? `sharedSecret.accessType === SecretSharingAccessType.Organization && orgId === sharedSecret.orgId`

maidul98 (changes requested):
The UI looks great. Left some small comments for the backend

backend/src/db/migrations/20240718170955_add-access-secret-sharing.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { SecretSharingAccessType } from "@app/lib/types";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasColumn = await knex.schema.hasColumn(TableName.SecretSharing, "accessType");
+  if (!hasColumn) {
+    await knex.schema.table(TableName.SecretSharing, (table) => {
+      table.string("accessType", 20).notNullable().defaultTo(SecretSharingAccessType.Anyone);
```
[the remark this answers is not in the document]
aheruz: Do you think we'll need longer accessTypes in the future?
Generally is a good idea to set a length for fields that are going to store "enums" or predefined values. It's also use less storage and it brings slightly better performance for indexing.

backend/src/services/secret-sharing/secret-sharing-service.ts
aheruz: Agree, fixed!

backend/src/services/secret-sharing/secret-sharing-service.ts
aheruz: But it's just validation, right?
If an error is returned, we'll lost the orgName in the frontend.
The idea of this logic is expose the organization name to bring context to the user, so they know that they need to log in into XY organization to get access.

backend/src/db/migrations/20240718170955_add-access-secret-sharing.ts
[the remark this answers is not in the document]
aheruz: Applied!

backend/src/services/secret-sharing/secret-sharing-service.ts
aheruz: Redundancy added ; )

PR #2162, opened 2024-07-22
Organization Role Page

dangtony98:
# Description 📣

This PR contains a revamped `RolePage` for organization-level roles; it improves upon the UI/UX of the previous role page.

Part of initiative to improve UI/UX/DX across the platform; this will serve as an example for what should be done at the project-level eventually as well.

Screenshot 2024-07-22 at 7 03 10 PM

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


akhilmhdh (commented):
The only issue i see here is each tick is doing a full update right.

I would recommend making it into a save button for permissions change because on actual production the latency will be visible and may make the user confused.

To replicate just do a network throttling from the developer toolbox in the browser.

akhilmhdh (approved):

PR #2171, opened 2024-07-24
fix(binary): aarch64 binary native bindings fix

DanielHougaard:
# Description 📣

We were building the Binary on an amd64 architecture, which resulted in issues when dealing with libraries that use node native bindings (like our own SDK does, and like argon2 does). When using a package that uses native bindings, it uses "optional dependencies", and those will be installed when the package itself is installed via `npm install`.

The way it works is that it checks your OS/arch to figure out which native binding to download, and we were always building on am amd64 architecture. This means that PKG would always bundle the amd64 native bindings, instead of the arch-specific bindings.

The solution for this was creating a self-hosted runner, which will always run on `ubuntu 24.04, arm64`. I've updated the Github workflow to use the self-hosted runner when building for Linux arm64.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

---

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -14,7 +14,6 @@ defaults:
 
 jobs:
     build-and-deploy:
-        runs-on: ubuntu-20.04
         strategy:
             matrix:
                 arch: [x64, arm64]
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -24,6 +23,7 @@ jobs:
                       target: node20-linux
                     - os: win
                       target: node20-win
+        runs-on: ${{ (matrix.arch == 'arm64' && matrix.os == 'linux') && 'ubuntu24-arm64' || 'ubuntu-20.04' }}
```
maidul98, line 26: Why this specific version of ubuntu? Can it just be `matrix.arch == 'arm64' && matrix.os == 'linux'`?

maidul98 (approved):

PR #2172, opened 2024-07-24
feat(core): Group support for access requests

DanielHougaard:
# Description 📣

This PR introduces group support for access requests. Groups have a separate database column for memberships, making it impossible to directly reference it without having to make a new field specifically for the group membership. We've moved away from using the project membership ID, to instead use a user ID / project ID combination, as this will give the same result but with more flexibility.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


akhilmhdh (commented):
@DanielHougaard I don't see the logic to delete project user additional privileges as now they are user ids. On the even of leaving/removing from a project as there membership is not linked they won't get removed right?

DanielHougaard:
> @DanielHougaard I don't see the logic to delete project user additional privileges as now they are user ids. On the event of leaving/removing from a project as there membership is not linked they won't get removed right?

@akhilmhdh, you're right! Privileges should be cleaned up once the user loses access to the project or organization, and I've updated the PR to handle this. I've decided to not clean up access requests as I think it makes sense that the user is able to keep a record of all access requests that were made in the past, regardless of if the user has been removed from the project or not. 

https://github.com/Infisical/infisical/pull/2172/commits/b702ffb668eeeca2a8dedfc4df815ddbfb1a0139

akhilmhdh (approved):
Code wise looking good to me. Need to test it.

maidul98:
@DanielHougaard have you tested this thoroughly?

DanielHougaard:
> @DanielHougaard have you tested this thoroughly?

Yes, but I would like to test this additionally on Gamma prior to full release

PR #2184, opened 2024-07-26
feat(cli): persistant `file` vault passphrase

DanielHougaard:
# Description 📣

We've been seeing issues with using the CLI keyring vault on certain linux systems, due to a default keyring not being present. Our solution to this has been to point people towards our file-based keyring, which works great except for this:
1. If the keyring fails, and the user hasn't explicitly set the keyring to `file` type, the CLI will throw an error and exit.
2. The user will need to re-enter their passphrase often, unless an environment variable is set to specify it. 

This PR addresses both of these issues.

We've introduced a new command called `vault set [auto|file] passphrase` _(only works with `file` currently)_, which will persist the password entered indefinitely. 

When the keyring get's used, we now have an error check in place that will fallback to the `file` keyring. This means that if something is wrong while using the normal keyring, we'll use the file-based keyring instead. 

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/cli/packages/cmd/vault.go
+++ b/cli/packages/cmd/vault.go
@@ -4,6 +4,7 @@ Copyright (c) 2023 Infisical Inc.
 package cmd
 
 import (
+	"encoding/base64"
 	"fmt"
 	"strings"
 
@@ -16,10 +17,49 @@ import (
 var AvailableVaultsAndDescriptions = []string{"auto (automatically select native vault on system)", "file (encrypted file vault)"}
 var AvailableVaults = []string{"auto", "file"}
 
+var vaultSetPassphraseCmd = &cobra.Command{
+	Example:               `infisical vault set-passphrase [your-passphrase]`,
+	Use:                   "set-passphrase [your-passphrase]",
+	Short:                 "Used to set the passphrase for the file vault",
+	DisableFlagsInUseLine: true,
+	Args:                  cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) != 1 {
+			log.Error().Msgf("Please provide a passphrase to set for the file vault")
+			return
+		}
+
+		passphrase := args[0]
+
+		configFile, err := util.GetConfigFile()
+		if err != nil {
+			log.Error().Msgf("Unable to set passphrase for file vault because of [err=%s]", err)
+			return
+		}
+
+		if configFile.VaultBackendType != "file" {
+			log.Error().Msgf("You are not using file vault to store your login details. You can only set passphrase for file vault")
+			return
+		}
+
+		// encode with base64
+		encodedPassphrase := base64.StdEncoding.EncodeToString([]byte(passphrase))
+		configFile.VaultBackendPassphrase = encodedPassphrase
+
+		err = util.WriteConfigFile(&configFile)
+		if err != nil {
+			log.Error().Msgf("Unable to set passphrase for file vault because of [err=%s]", err)
+			return
+		}
+
+		fmt.Printf("\nSuccessfully, set passphrase for file vault. You can now store your login details securely at rest\n")
+	},
+}
+
 var vaultSetCmd = &cobra.Command{
-	Example:               `infisical vault set pass`,
-	Use:                   "set [vault-name]",
-	Short:                 "Used to set the vault backend to store your login details securely at rest",
+	Example:               `infisical vault set [file|auto]`,
+	Use:                   "set [file|auto]",
+	Short:                 "Used to set the type of vault backend to store your login details securely at rest",
 	DisableFlagsInUseLine: true,
 	Args:                  cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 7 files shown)


cli/packages/cmd/vault.go
```diff
@@ -4,62 +4,93 @@ Copyright (c) 2023 Infisical Inc.
...
+					}
+
+					passphrase, err := passphrasePrompt.Run()
+					if err != nil {
+						log.Error().Msgf("Unable to set passphrase for file vault because of [err=%s]", err)
+						return
+					}
+
+					if passphrase == "" || len(passphrase) < 8 {
+						log.Error().Msgf("Passphrase must be at least 8 characters long")
```
maidul98, line 68: i don't think a requirement is necessary

cli/packages/util/keyringwrapper.go
```diff
@@ -20,16 +27,32 @@ func SetValueInKeyring(key, value string) error {
 		PrintErrorAndExit(1, err, "Unable to get current vault. Tip: run [infisical rest] then try again")
 	}
 
-	return keyring.Set(currentVaultBackend, MAIN_KEYRING_SERVICE, key, value)
+	err = keyring.Set(currentVaultBackend, MAIN_KEYRING_SERVICE, key, value)
+
+	if err == keyring.ErrUnsupportedPlatform || keyringNotConfigured(err) {
```
maidul98, line 32: Looks like the default to file mode is not being set. You can reproduce it by logging in inside of codespaces
CleanShot 2024-07-26 at 19 22 07@2x

cli/packages/cmd/vault.go
```diff
@@ -4,62 +4,93 @@ Copyright (c) 2023 Infisical Inc.
...
+		Description: "encrypted file vault",
+	},
+}
 
 var vaultSetCmd = &cobra.Command{
-	Example:               `infisical vault set pass`,
-	Use:                   "set [vault-name]",
-	Short:                 "Used to set the vault backend to store your login details securely at rest",
+	Example:               `infisical vault set [file|auto] [option]`,
+	Use:                   "set [file|auto] [option]",
```
maidul98, line 36: I think option should actually be a flag? `infisical vault set file --passphrase`

cli/packages/cmd/vault.go
```diff
@@ -13,53 +14,56 @@ import (
...
-			Telemetry.CaptureEvent("cli-command:vault set", posthog.NewProperties().Set("currentVault", currentVaultBackend).Set("wantedVault", wantedVaultTypeName).Set("version", util.CLI_VERSION))
-		} else {
-			log.Error().Msgf("The requested vault type [%s] is not available on this system. Only the following vault backends are available for you system: %s", wantedVaultTypeName, strings.Join(AvailableVaults, ", "))
-		}
+		util.PrintWarning("This command has been deprecated. Please use 'infisical vault use [file|auto]' to select which vault to use.\n")
+		selectVaultTypeCmd(cmd, args)
 	},
 }
 
+var vaultUseCmd = &cobra.Command{
```
maidul98, line 58: I thought vault `use` would also have the `passphrase` flag since this is a replacement of vault use?

cli/packages/util/keyringwrapper.go
```diff
@@ -20,16 +23,51 @@ func SetValueInKeyring(key, value string) error {
...
+
+	if err != nil {
+		configFile, _ := GetConfigFile()
+		PrintWarning("Fallback file keyring is being used\n\nYou can persist your file passphrase by running the following command:\ninfisical vault set file --passphrase <your-passphrase>\n")
+
+		if configFile.VaultBackendPassphrase == "" {
+			passphrasePrompt := promptui.Prompt{
+				Label: "Enter the passphrase to use for keyring encryption: ",
+			}
+			passphrase, err := passphrasePrompt.Run()
```
maidul98, line 36: I'm getting some kind of massive lag when typing in the terminal

DanielHougaard: I haven't gotten this before no -- how big is your terminal size? I know our CLI generally gets very buggy when the terminal is small

maidul98: figured it out

maidul98 (approved):

cli/packages/util/keyringwrapper.go
```diff
@@ -20,16 +23,51 @@ func SetValueInKeyring(key, value string) error {
 		PrintErrorAndExit(1, err, "Unable to get current vault. Tip: run [infisical rest] then try again")
 	}
 
-	return keyring.Set(currentVaultBackend, MAIN_KEYRING_SERVICE, key, value)
+	err = keyring.Set(currentVaultBackend, MAIN_KEYRING_SERVICE, key, value)
+
+	if err != nil {
+		configFile, _ := GetConfigFile()
+		PrintWarning("System keyring could not be used, switching to `file` vault for local token storage\n\nYou can persist your file vault passphrase by running the following command:\ninfisical vault set file --passphrase <your-passphrase>\n")
```
maidul98, line 30: This message will continue to show up even if the first run set the vault to file and the pass phrase has been saved to file

maidul98 (approved):

PR #2186, opened 2024-07-27
feat: persist tab state on route change.

LemmyMwaura:
# Description 📣

This PR hoists the tab state to the URL and persists it when tab switching at both the org and project level and also across both.

## Type ✨

- [x] Bug fix
- [x] New feature

# Screenshot

https://github.com/user-attachments/assets/9bd1a3ee-5ae9-4b28-aef4-32711b69cb90

# Tests 🛠️

---


maidul98:
I don't remember seeing this text sestion before. Was this intentional? 

CleanShot 2024-07-28 at 13 03 07@2x

LemmyMwaura:
> I don't remember seeing this text sestion before. Was this intentional?
> 

No that was accidental. cleared it.

frontend/src/types/TabSections.ts
```diff
@@ -0,0 +1,13 @@
+enum TabSections {
+  Member = "members",
+  Roles = "roles",
+  Groups = "groups",
+  Identities = "identities",
+  ServiceTokens = "service-tokens"
+}
+
+const isTabSection = (value: any): value is TabSections => {
```
sheensantoscapadngan, line 9: let's use "string" here instead of any?

frontend/src/types/TabSections.ts
```diff
@@ -0,0 +1,13 @@
...
+  Member = "members",
+  Roles = "roles",
+  Groups = "groups",
+  Identities = "identities",
+  ServiceTokens = "service-tokens"
+}
+
+const isTabSection = (value: any): value is TabSections => {
+  return Object.values(TabSections).includes(value);
+}
```
sheensantoscapadngan, line 11: for uniformity with the existing structure, let's do export here right away

frontend/src/types/index.ts
```diff
@@ -0,0 +1,3 @@
+import { TabSections, isTabSection } from "./TabSections";
+
+export { TabSections, isTabSection };
```
sheensantoscapadngan, line 3: it might be better just scope this to the Org view by adding the types folder in the Org directory here 
image

Same for the project tab enum. 

we try to keep the files and folders as localized as possible

frontend/src/types/TabSections.ts
LemmyMwaura: updated

frontend/src/types/TabSections.ts
LemmyMwaura: alright

frontend/src/types/index.ts
LemmyMwaura: moved to the org and project level

maidul98 (approved):

PR #2194, opened 2024-07-30
fix(cli): make yaml exporting reliable and standardized

### LINKED ISSUE #1955: Multiline secrets aren't exported properly in YAML
### Describe the bug

I have a multiline secret that I want to export in YAML format. The secret looks like this

Screenshot 2024-06-12 at 09 28 51

A multiline string is valid for YAML if the format looks like this

```yaml
TEST: |
  A valid
  multiline secret
  for testing
```

This is also a valid YAML multiline string

```yaml
ANOTHER_TEST: |
  Another example
    but this time this one is indented
  and this one is not
```

However, this is not

```yaml
NOT_VALID: |
    The first line shouldn't be indented further
  than the other lines
  This is not valid
```

There are ways to allow this but for how a lot of people use YAML for configuration, it's safe to just support the first two examples:

* It starts with a `|`
* And then the string itself is indented and starts from the line after it

However, the current YAML export throws this out, which isn't valid YAML

```yaml
TEST: A valid
multiline secret
for testing
```

Now there are ways in YAML to support the third example, but for configuration purposes, I believe supporting the first two examples is what's expected.

### To Reproduce

1. Create a new secret called `TEST` on any environment
2. Paste this as the value

```
A valid
multiline secret
for testing
```

3. Export the secret in YAML format with `infisical export`
4. The result will be as I've described above

### Expected behavior

What I expected from this feature was to have it exported as valid multiline string in YAML like this:

```yaml
TEST: |
  A valid
  multiline secret
  for testing
```

### Screenshots

* The secret I used for testing this

image (1)

* The result I got from the export

image

I initially encountered this issue on Windows but since its an export related issue, it exists on all platforms

GLEF1X:
# Description 📣

This pull request updates the `formatAsYaml` function to improve `YAML` formatting for environment variables. The function now uses a map to store key-value pairs and employs the `yaml.Marshal` method for proper YAML serialization. This change ensures better handling of special characters and multi-line values in YAML output. The function's signature has been modified to return both a string and an error, allowing for more robust error handling. These improvements enhance the reliability and correctness of YAML output in the application.

It also fixes #1955

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Test suite for the new implementation of `formatAsYaml` is included in the PR

---

### DIFF
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -77,7 +77,6 @@ require (
 	github.com/muesli/termenv v0.15.2 // indirect
 	github.com/oklog/ulid v1.3.1 // indirect
 	github.com/pelletier/go-toml v1.9.3 // indirect
-	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/spf13/afero v1.6.0 // indirect
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -77,6 +77,7 @@ require (
 	github.com/muesli/termenv v0.15.2 // indirect
 	github.com/oklog/ulid v1.3.1 // indirect
 	github.com/pelletier/go-toml v1.9.3 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/spf13/afero v1.6.0 // indirect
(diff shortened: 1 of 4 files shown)


cli/go.mod
```diff
@@ -77,6 +77,7 @@ require (
 	github.com/muesli/termenv v0.15.2 // indirect
 	github.com/oklog/ulid v1.3.1 // indirect
 	github.com/pelletier/go-toml v1.9.3 // indirect
+	github.com/pkg/errors v0.9.1 // indirect
```
akhilmhdh, line 80: Please remove this package as its not needed anymore in golang. 

Go error management already has this

akhilmhdh (changes requested):

cli/go.mod
GLEF1X: Thank you for catching that! I completely missed that this functionality had been added to the standard library. I just updated the PR and removed the dependency

akhilmhdh (approved):

maidul98 (approved):

PR #2201, opened 2024-07-30
feat: Implement secrets-path glob testing tool 

LemmyMwaura:
# Description 📣

This PR implements the secrets path glob testing tool. 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Screenshot
Screenshot 2024-07-30 231601

# Tests 🛠️

---

### DIFF
--- a/frontend/public/locales/en/translations.json
+++ b/frontend/public/locales/en/translations.json
@@ -258,6 +258,9 @@
         "copy-service-token": "Copy your service token",
         "copy-service-token-description": "Once you close this popup, you will never see your service token again"
       }
+    },
+    "glob-tool": {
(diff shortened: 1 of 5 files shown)


akhilmhdh:
I don't think we need a separate page for this. What we need is a small modal so that users don't need to switch context at all. 

What i am envisioning is they see a small circular information button on clicking opens up a popup/modal that ask them to test it out. This way its more smoother flow. 

Instead of modifying the ui components at root level, we should create a new component that forms from the root component. 

```
<FormControl label={<PathPermissionHelper  />}/>
```

PR #2223, opened 2024-08-02
misc: moved to license-plan-based rate limits

sheensantoscapadngan:
# Description 📣
This PR will enable users in shared cloud to have different rate limits based on their organization's license plan

License Server PR:
https://github.com/Infisical/license-server/pull/21

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/services/license/licence-fns.ts
+++ b/backend/src/ee/services/license/licence-fns.ts
@@ -43,13 +43,8 @@ export const getDefaultOnPremFeatures = (): TFeatureSet => ({
   externalKms: false,
   rateLimits: {
     readLimit: 60,
-    publicEndpointLimit: 30,
     writeLimit: 200,
-    secretsLimit: 40,
-    authRateLimit: 60,
-    inviteUserRateLimit: 30,
-    mfaRateLimit: 20,
-    creationLimit: 30
+    secretsLimit: 40
   }
 });
 
--- a/backend/src/ee/services/license/licence-fns.ts
+++ b/backend/src/ee/services/license/licence-fns.ts
@@ -40,7 +40,17 @@ export const getDefaultOnPremFeatures = (): TFeatureSet => ({
   secretRotation: true,
   caCrl: false,
   instanceUserManagement: false,
-  externalKms: false
+  externalKms: false,
+  rateLimits: {
+    readLimit: 60,
+    publicEndpointLimit: 30,
+    writeLimit: 200,
+    secretsLimit: 60,
+    authRateLimit: 60,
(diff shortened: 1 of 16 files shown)


backend/src/server/plugins/inject-rate-limits.ts
```diff
@@ -0,0 +1,35 @@
...
+      return;
+    }
+
+    const plan = await server.services.license.getPlan(req.auth.orgId);
+    const { rateLimits } = plan;
+
+    if (plan.customRateLimits) {
+      req.rateLimits = defaultRateLimiterConfig;
+      return;
+    }
```
sheensantoscapadngan, line 21: this should only be the case when self-hosted/dedicated instance :D

backend/src/ee/services/license/licence-fns.ts
```diff
@@ -40,7 +40,17 @@ export const getDefaultOnPremFeatures = (): TFeatureSet => ({
   secretRotation: true,
   caCrl: false,
   instanceUserManagement: false,
-  externalKms: false
+  externalKms: false,
+  rateLimits: {
+    readLimit: 60,
+    publicEndpointLimit: 30,
+    writeLimit: 200,
```
maidul98, line 47: what is the difference between write vs creation end points?

maidul98 (commented):
Looks go to me. Although it looks like the whole org has a single rate limit now. Ideally it is per identity (user or machine) because right now let's say my org can have TPS of 10 for write but now every machine is subject to the same rate limit. That would mean that if i have machine A and machine B apart of the same org, their requests cannot go past 10 TPS individual

backend/src/server/config/rateLimiter.ts
```diff
@@ -22,40 +21,46 @@ export const globalRateLimiterCfg = (): RateLimitPluginOptions => {
 // GET endpoints
 export const readLimit: RateLimitOptions = {
   timeWindow: 60 * 1000,
-  max: () => getRateLimiterConfig().readLimit,
+  hook: "preValidation",
```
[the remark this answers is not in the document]
sheensantoscapadngan: exactly!

backend/src/ee/services/license/licence-fns.ts
sheensantoscapadngan: hmmm based on existing usage, creationLimit is for org, project, and machine identity creation while writeLimit is for everything else

maidul98 (approved):
lgtm

PR #2238, opened 2024-08-05
feat: added secrets deletion feature on integration removal

sheensantoscapadngan:
# Description 📣
- This PR will allow users to optionally delete secrets in the destination (for now, only Github) upon deletion of integration 
image
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/integration-auth/integration-delete-secret.ts
+++ b/backend/src/services/integration-auth/integration-delete-secret.ts
@@ -1,3 +1,4 @@
+import { retry } from "@octokit/plugin-retry";
 import { Octokit } from "@octokit/rest";
 
 import { TIntegrationAuths, TIntegrations } from "@app/db/schemas";
@@ -79,7 +80,7 @@ const getIntegrationSecretsV2 = async (
  * Return the secrets in a given [folderId] including secrets from
  * nested imported folders recursively.
  */
-const getIntegrationSecrets = async (
+const getIntegrationSecretsV1 = async (
   dto: {
     projectId: string;
     environment: string;
@@ -129,7 +130,7 @@ const getIntegrationSecrets = async (
     if (folder) {
       // get secrets contained in each imported folder by recursively calling
       // this function against the imported folder
-      const importedSecrets = await getIntegrationSecrets(
+      const importedSecrets = await getIntegrationSecretsV1(
         {
           environment: dto.environment,
           projectId: dto.projectId,
@@ -167,7 +168,8 @@ export const deleteGithubSecrets = async ({
     selected_repositories_url?: string | undefined;
   }
 
-  const octokit = new Octokit({
+  const OctokitWithRetry = Octokit.plugin(retry);
+  const octokit = new OctokitWithRetry({
     auth: accessToken
   });
 
@@ -177,11 +179,11 @@ export const deleteGithubSecrets = async ({
     Env = "github-env"
   }
 
-  let encryptedSecrets: GitHubSecret[];
+  let encryptedGithubSecrets: GitHubSecret[];
 
   switch (integration.scope) {
     case GithubScope.Org: {
-      encryptedSecrets = (
+      encryptedGithubSecrets = (
         await octokit.request("GET /orgs/{org}/actions/secrets", {
           org: integration.owner as string
         })
@@ -189,7 +191,7 @@ export const deleteGithubSecrets = async ({
       break;
     }
     case GithubScope.Env: {
-      encryptedSecrets = (
+      encryptedGithubSecrets = (
         await octokit.request("GET /repositories/{repository_id}/environments/{environment_name}/secrets", {
           repository_id: Number(integration.appId),
           environment_name: integration.targetEnvironmentId as string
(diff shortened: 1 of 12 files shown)


backend/src/services/integration-auth/integration-delete-secret.ts
```diff
@@ -0,0 +1,350 @@
...
+          repo: integration.app as string
+        })
+      ).data.secrets;
+      break;
+    }
+  }
+
+  for await (const encryptedSecret of encryptedSecrets) {
+    if (encryptedSecret.name in secrets) {
+      switch (integration.scope) {
```
sheensantoscapadngan, line 213: logic here is only delete secrets that were managed by the integration

backend/src/services/integration-auth/integration-delete-secret.ts
```diff
@@ -0,0 +1,350 @@
...
+  const content: Record<string, boolean> = {};
+  if (dto.depth > MAX_SYNC_SECRET_DEPTH) {
+    logger.info(
+      `getIntegrationSecrets: secret depth exceeded for [projectId=${dto.projectId}] [folderId=${dto.folderId}] [depth=${dto.depth}]`
+    );
+    return content;
+  }
+
+  // process secrets in current folder
+  const secrets = await secretV2BridgeDAL.findByFolderId(dto.folderId);
```
maidul98, line 47: How do you know that secrets for this project are in secrets engine v2? I was expecting something like this pattern:


https://github.com/search?q=repo%3AInfisical%2Finfisical%20shouldUseSecretV2Bridge&type=code

backend/src/services/integration-auth/integration-delete-secret.ts
```diff
@@ -0,0 +1,350 @@
...
+    }
+  }
+  return content;
+};
+
+/**
+ * Return the secrets in a given [folderId] including secrets from
+ * nested imported folders recursively.
+ */
+const getIntegrationSecrets = async (
```
maidul98, line 82: let's call this v1 since you are calling the other one v2?

backend/src/services/integration-auth/integration-delete-secret.ts
```diff
@@ -0,0 +1,350 @@
...
+    auth: accessToken
+  });
+
+  enum GithubScope {
+    Repo = "github-repo",
+    Org = "github-org",
+    Env = "github-env"
+  }
+
+  let encryptedSecrets: GitHubSecret[];
```
maidul98, line 180: nit: rename to encryptedSecretsInGithub

backend/src/services/integration-auth/integration-delete-secret.ts
```diff
@@ -0,0 +1,350 @@
...
+      suffixedSecrets[newKey] = secrets[key];
+    });
+  }
+
+  switch (integration.integration) {
+    case Integrations.GITHUB: {
+      await deleteGithubSecrets({
+        integration,
+        accessToken,
+        secrets: Object.keys(suffixedSecrets).length !== 0 ? suffixedSecrets : secrets
```
maidul98, line 341: Can you have both suffixed and non suffixed secrets or just one or the other?

backend/src/services/integration-auth/integration-delete-secret.ts
```diff
@@ -0,0 +1,350 @@
...
+    Repo = "github-repo",
+    Org = "github-org",
+    Env = "github-env"
+  }
+
+  let encryptedSecrets: GitHubSecret[];
+
+  switch (integration.scope) {
+    case GithubScope.Org: {
+      encryptedSecrets = (
```
maidul98, line 184: I don't see much error handling for the octokit calls. Maybe we can add some kind of retry logic if it exists for octokit

backend/src/services/integration-auth/integration-delete-secret.ts
```diff
@@ -0,0 +1,350 @@
...
+          owner: integration.owner as string,
+          repo: integration.app as string
+        })
+      ).data.secrets;
+      break;
+    }
+  }
+
+  for await (const encryptedSecret of encryptedSecrets) {
+    if (encryptedSecret.name in secrets) {
```
maidul98, line 212: This will make a new request right after the first one finishes. We should add some kind of delay between each call so eveything doesn't get called so quickly

maidul98 (approved):
Looks good, just left some small comments

backend/src/services/integration-auth/integration-delete-secret.ts
sheensantoscapadngan: for an integration, you can only have one or the other

backend/src/services/integration-auth/integration-delete-secret.ts
sheensantoscapadngan: added octokit retry plugin

backend/src/services/integration-auth/integration-delete-secret.ts
sheensantoscapadngan: agree with this. I added 50 milliseconds between each delete request

maidul98 (approved):

PR #2242, opened 2024-08-06
Tag based filtering for secret endpoint

### LINKED ISSUE #2229: Add tags filter support for secrets v3 API
### Feature description
Users should be able to send a list of tags to filter secrets by in the v3 secrets API. This will then be consued by clients such as CLI.

akhilmhdh:
# Description 📣

This PR brings back tag based filtering for secret raw endpoints for both v2 and v3 projects. The respective changes are made in CLI to do the same

Closes #2229 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -597,7 +597,7 @@ export const RAW_SECRETS = {
     environment: "The slug of the environment to list secrets from.",
     secretPath: "The secret path to list secrets from.",
     includeImports: "Weather to include imported secrets or not.",
-    tagSlugs: "The comma seperated tag slugs to filter secrets"
+    tagSlugs: "The comma separated tag slugs to filter secrets"
   },
   CREATE: {
     secretName: "The name of the secret to create.",
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -596,7 +596,8 @@ export const RAW_SECRETS = {
       "The slug of the project to list secrets from. This parameter is only applicable by machine identities.",
     environment: "The slug of the environment to list secrets from.",
     secretPath: "The secret path to list secrets from.",
-    includeImports: "Weather to include imported secrets or not."
+    includeImports: "Weather to include imported secrets or not.",
+    tagSlugs: "The comma seperated tag slugs to filter secrets"
   },
   CREATE: {
     secretName: "The name of the secret to create.",
--- a/backend/src/services/secret/secret-service.ts
+++ b/backend/src/services/secret/secret-service.ts
@@ -2086,7 +2086,7 @@ export const secretServiceFactory = ({
 
     return {
       ...updatedSecret[0],
-      tags: [...existingSecretTags, ...tags].map((t) => ({ id: t.id, slug: t.slug, name: t.name, color: t.color }))
+      tags: [...existingSecretTags, ...tags].map((t) => ({ id: t.id, slug: t.slug, color: t.color }))
     };
   };
 
--- a/backend/src/services/secret/secret-service.ts
+++ b/backend/src/services/secret/secret-service.ts
@@ -964,7 +964,8 @@ export const secretServiceFactory = ({
     environment,
     includeImports,
     expandSecretReferences,
-    recursive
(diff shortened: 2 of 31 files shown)


backend/src/lib/api-docs/constants.ts
```diff
@@ -596,7 +596,8 @@ export const RAW_SECRETS = {
       "The slug of the project to list secrets from. This parameter is only applicable by machine identities.",
     environment: "The slug of the environment to list secrets from.",
     secretPath: "The secret path to list secrets from.",
-    includeImports: "Weather to include imported secrets or not."
+    includeImports: "Weather to include imported secrets or not.",
+    tagSlugs: "The comma seperated tag slugs to filter secrets"
   },
```
maidul98, line 601: I think we should remove tag slugs in the schema and just make tag name be required to be a slug. Users don't yet use slugs in production clients yet since it wasn't supported before so i think we can safely make this change in the backend

backend/src/services/secret/secret-service.ts
```diff
@@ -998,6 +1000,9 @@ export const secretServiceFactory = ({
     });
 
     const decryptedSecrets = secrets.map((el) => decryptSecretRaw(el, botKey));
+    const filteredSecrets = tagSlugs.length
```
maidul98, line 1003: question: why not add this filter to the sql query?

maidul98 (commented):
Left comments

backend/src/services/secret/secret-service.ts
akhilmhdh: The reasoning was to filter by name and slug but i guess it's not needed anymore. 

But i guess with unification of this then it should be ok

akhilmhdh: Keeping it for now to reduce complexity as changing the filter in secret dal would cause a lot of changes in multiple places at the moment. Will address this on new secret service

backend/src/lib/api-docs/constants.ts
akhilmhdh: Removed tag names instead of tag slug

backend/src/services/secret/secret-service.ts
```diff
@@ -2081,7 +2086,7 @@ export const secretServiceFactory = ({
 
     return {
       ...updatedSecret[0],
-      tags: [...existingSecretTags, ...tags].map((t) => ({ id: t.id, slug: t.slug, name: t.name, color: t.color }))
+      tags: [...existingSecretTags, ...tags].map((t) => ({ id: t.id, slug: t.slug, color: t.color }))
```
maidul98, line 2089: Did we miss the name tag here?

backend/src/services/secret/secret-service.ts
```diff
@@ -1106,14 +1111,14 @@ export const secretServiceFactory = ({
       };
 
       // expand secrets
-      await batchSecretsExpand(decryptedSecrets);
+      await batchSecretsExpand(filteredSecrets);
```
maidul98, line 1114: i don't think the names of the tags are set to their slugs here from the response

maidul98 (commented):
Just left 2 comments. let's just double check we aren't missing the name field because that will break peoples terraform state

backend/src/services/secret/secret-service.ts
akhilmhdh: That is secret expansion. We do the mapping in `decryptSecretRaw`

https://github.com/Infisical/infisical/pull/2242/files/bb52e2beb4567eb2eef9972ac90512fad3acb1a1#diff-9ad04b946494acd4d22dd0a4b21441235594fae872d62838717f57fdda3087d8R414

As this is used everywhere as end result in raw operation we will never miss the conversion

backend/src/services/secret/secret-service.ts
akhilmhdh: Still not sure where this endpoint is used? May daniel has better idea over this one. But i think yah i missed it. 

Not used in terraform

maidul98 (approved):

PR #2244, opened 2024-08-06
feat: password protect secret share

### LINKED ISSUE #2231: Allow secret share to be password protected
### Feature description
Right now anyone with a secret share link can view the shared secret. It would be great to have the option to protect the share with a password as well. This way, you need both the password and the link to accesss the secret.

LemmyMwaura:
# Description 📣

You can now protect shared secrets with a password. When sharing a secret, you have the option to add a password. To access a password-protected secret, users need both the password and the link.

NB: password is optional, a user can still create the secret without password protection and it'll work the same as before.

fixes #2231 

## How It Works

1. **Initial API Call:**
   - When the secret-sharing page loads, an API call is made to retrieve the secret.
   - The handler performs these checks:
     - **Authorization:** Ensure the request is authorized. (`throw error if unauthorized`)
     - **Existence:** Verify the secret exists. (`throw error if does not exist`)
     - **Expiry:** Check if the secret is expired. (`throw error if expired`)
     - **Password Protection:**
       - If the secret is password-protected:
         - **Send Falsy Value:** Return a falsy value to the client.
       - If the secret is not password-protected:
         - **View Count:** Decrement the view count and return the secret.
         - **Protected In Trasit:** Send Encrypted Secret to client.

2. **UI Handling:**
   - The UI receives either a secret or a falsy value:
     - **Secret:** Decipher and display it to the user.
     - **Falsy Value:** Prompt the user for a password.

3. **Password Validation API Call:**
   - After the user provides a password, a second API call is made to validate it.
   - The handler performs these checks:
     - **Authorization:** Ensure the request is authorized. (`throw error if unauthorized`)
     - **Existence:** Verify the secret exists. (`throw error if does not exist`)
     - **Password Protection:**
       - if password does not matche:
         - **Send Falsy Value:** Return a falsy value to the client.
       - If the password matches:
         - **View Count:** Decrement the view count and return the secret.
         - **Protected In Trasit:** Send Encrypted Secret to client.

## Video

https://github.com/user-attachments/assets/ab9a55d2-e884-471a-bd8b-c652f767e696

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -62,6 +63,7 @@ export const secretSharingServiceFactory = ({
 
     const newSharedSecret = await secretSharingDAL.create({
       name,
+      password,
```
maidul98, line 66: I don't think this is a good idea to keep this in plain text

maidul98 (commented):
i noticed there were were no errors shown when the password i entered was incorrect. Let's add some better validation

backend/src/services/secret-sharing/secret-sharing-service.ts
LemmyMwaura: yeah, currently hashing it and adding a validate endpoint on the backend.

backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -92,6 +95,49 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
...
+        password: z.string().min(1),
+        hashedHex: z.string()
+      }),
+      response: {
+        200: z.object({
+          isValid: z.boolean()
+        })
+      }
+    },
+    handler: async (req) => {
```
akhilmhdh, line 118: Please keep this whole a seperate service function

frontend/src/views/ShareSecretPublicPage/components/ShareSecretForm.tsx
```diff
@@ -149,6 +151,19 @@ export const ShareSecretForm = ({ isPublic, value }: Props) => {
           </FormControl>
         )}
       />
+      <Controller
+        control={control}
+        name="password"
+        render={({ field, fieldState: { error } }) => (
+          <FormControl
+            label="Password (Optional)"
```
akhilmhdh, line 159: Use isOptional flag

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -103,6 +109,7 @@ export const secretSharingServiceFactory = ({
     }
 
     const newSharedSecret = await secretSharingDAL.create({
+      password,
```
akhilmhdh, line 112: Why is this password not hashed

frontend/src/views/ViewSecretPublicPage/components/PasswordContainer.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { createNotification } from "@app/components/notifications";
+
+type Props = {
+  secretId: string;
+  hashedHex: string;
+  handlePassMatch: (val: boolean) => void;
+};
+
+export const PasswordContainer = ({ secretId, hashedHex, handlePassMatch }: Props) => {
+  const [password, setPassword] = useState<string>('')
```
akhilmhdh, line 17: Use react hook form for all form management. You can avoid keeping isLoading and everything

backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -63,6 +64,7 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
       response: {
         200: SecretSharingSchema.pick({
           encryptedValue: true,
+          password: true,
```
akhilmhdh, line 67: I don't think password should be returned here in response. That defeats the whole purpose of password protection

akhilmhdh (changes requested):
From security stand point of view this doesn't change fact that somebody with the link can still see the secret.

Because if you see here this is just ui level change as user types password checks valid or not. But from API level i can just get the original secret value and decrypt it.

So to make it work in security the encrypted Value should not be written until the password is matched. Also password should not be writtened in response

backend/src/services/secret-sharing/secret-sharing-service.ts
LemmyMwaura: missed this. hashing it.

frontend/src/views/ViewSecretPublicPage/components/PasswordContainer.tsx
LemmyMwaura: Refactored to that. I had avoided it as state hooks require less code (in this case) as it's just one input.

backend/src/server/routes/v1/secret-sharing-router.ts
LemmyMwaura: true, fixed

LemmyMwaura:
> So to make it work in security the encrypted Value should not be written until the password is matched. Also password should not be writtened in response

True. I Wasn't quite sure how strict you wanted this. I've moved all the validation to the backend. We now only get a secret on initial load if it doesn't have a password and is still fresh. Otherwise, the user has to enter a valid password to retrieve the secret.

backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -74,12 +74,66 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
...
+      const sharedSecret = await req.server.services.secretSharing.getPasswordlessSecretByID({
         sharedSecretId: req.params.id,
         hashedHex: req.query.hashedHex,
         orgId: req.permission?.orgId
       });
+
+      if (!sharedSecret) return undefined;
+
+      return {
+        encryptedValue: sharedSecret.encryptedValue,
```
akhilmhdh, line 86: The encrypted value must not return if password is provided because if it does defeats the whole purpose of password protection

frontend/src/views/ViewSecretPublicPage/ViewSecretPublicPage.tsx
```diff
@@ -52,8 +62,19 @@ export const ViewSecretPublicPage = () => {
             </a>
           </p>
         </div>
-        {secret && key && <SecretContainer secret={secret} secretKey={key} />}
-        {error && <SecretErrorContainer />}
+        {!isLoading && (
+          <>
+            {!error && !secret && (
+              <PasswordContainer
```
akhilmhdh, line 68: I think you have made it much more complex than it should be given the use effect and state. 

What you can do to simplify is. Return a flag in response `isPasswordProtected`. If this flag is true encryptedValue will be null. 

Then after validation with password encryptedValue will be provided. And corresponding ui can be simplified much easier with that way. No need for useEffect

akhilmhdh (changes requested):
Our aim here is to not just block from a ui level rather API level too because a bad actor won't just stop on UI level

frontend/src/views/ShareSecretPublicPage/components/ShareSecretForm.tsx
```diff
@@ -149,6 +151,20 @@ export const ShareSecretForm = ({ isPublic, value }: Props) => {
...
       />
+      <Controller
+        control={control}
+        name="password"
+        render={({ field, fieldState: { error } }) => (
+          <FormControl
+            label="Password"
+            isError={Boolean(error)}
+            errorText={error?.message}
+            isOptional={true}
```
bluin4308, line 162: you can omit `={true}` here, It'll make your code cleaner

backend/src/server/routes/v1/secret-sharing-router.ts
LemmyMwaura: the encrypted value is never returned unless it's not password protected and we only return it so it can be deciphered on the client.

akhilmhdh: I am a bit confused. From the code i see the encrypted value is returned irrespective of password protected. Can you please point me where this is getting stripped off?

LemmyMwaura: @akhilmhdh 

The code highlighted above returns a passwordless encrypted secret to be deciphered on the client. The password validation happens later. 

At this point no password has been sent to the server. That's on initial load.

frontend/src/views/ViewSecretPublicPage/ViewSecretPublicPage.tsx
LemmyMwaura: The state and effect hooks are used to handle the callback situation as we need to pass the secret from the PasswordContainer child component. Adding a flag is still the same as we have to check if value is `falsy` (this was my initial approach). 

i get the gist of avoiding `effect` and `state` but we need them in this state as secret can be set twice 
- (1). on initial load (is passwordless) 
- (2). from child component if password protected

LemmyMwaura: The alternative is to avoid the callback situation and create a parent component for secretContainer, PasswordContainer and secretErrorContainer.

akhilmhdh:
Hey @LemmyMwaura Made some changes as needed to simplify the whole operation and ui. 

Good work @LemmyMwaura.

akhilmhdh (approved):

maidul98 (approved):

PR #2245, opened 2024-08-06
Adding an selectAll button in SecretOverviewPage

macci001:
# Description 📣

The PR adds a feature for selecting all the secrets/features.

## Type ✨

- [ ] New feature

# Tests 🛠️

https://github.com/user-attachments/assets/41e53b05-9092-4e56-9ff0-2e120c131530
---

- [Yes ] I have read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/getting-started/code-of-conduct). 📝

### DIFF
--- a/frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
+++ b/frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
@@ -135,6 +135,23 @@ export const SecretOverviewPage = () => {
     });
   }, []);
 
+  const selectAllEntries = useCallback(() => {
+    const folderRecord: Record<string, boolean> = {};
+    folderNames.forEach((folder: string) => {
+      folderRecord[folder] = true;
+    })
+
+    const secretRecord: Record<string, boolean> = {};
+    secKeys.forEach((secret: string) => {
+      secretRecord[secret] = true;
+    })
+
+    setSelectedEntries({
+      [EntryType.FOLDER]: folderRecord,
+      [EntryType.SECRET]: secretRecord
+    })
+  }, []);
+
   useEffect(() => {
     const handleParentTableWidthResize = () => {
       setExpandableTableWidth(parentTableRef.current?.clientWidth || 0);
@@ -649,6 +666,7 @@ export const SecretOverviewPage = () => {
           getFolderByNameAndEnv={getFolderByNameAndEnv}
           selectedEntries={selectedEntries}
           resetSelectedEntries={resetSelectedEntries}
+          selectAllEntries={selectAllEntries}
         />
         <div className="thin-scrollbar mt-4" ref={parentTableRef}>
           <TableContainer className="max-h-[calc(100vh-250px)] overflow-y-auto">
--- a/frontend/src/views/SecretOverviewPage/components/SelectionPanel/SelectionPanel.tsx
+++ b/frontend/src/views/SecretOverviewPage/components/SelectionPanel/SelectionPanel.tsx
@@ -1,5 +1,5 @@
 import { subject } from "@casl/ability";
-import { faMinusSquare, faTrash } from "@fortawesome/free-solid-svg-icons";
+import { faMinusSquare, faSquareCheck, faTrash } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { twMerge } from "tailwind-merge";
 
@@ -30,6 +30,7 @@ type Props = {
   getSecretByKey: (slug: string, key: string) => SecretV3RawSanitized | undefined;
   getFolderByNameAndEnv: (name: string, env: string) => TSecretFolder | undefined;
   resetSelectedEntries: () => void;
+  selectAllEntries: () => void;
   selectedEntries: {
     [EntryType.FOLDER]: Record<string, boolean>;
     [EntryType.SECRET]: Record<string, boolean>;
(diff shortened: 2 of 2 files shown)


maidul98 (commented):
Nice PR, here is some feedback: 
I think the double check box looks off, i recommend following this from AWS:
CleanShot 2024-08-06 at 09 50 09@2x

I haven't checked yet, but this same functionality should be available when you go into a environment

akhilmhdh (changes requested):

frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
```diff
@@ -656,11 +695,16 @@ export const SecretOverviewPage = () => {
               <THead>
                 <Tr className="sticky top-0 z-20 border-0">
                   <Th className="sticky left-0 z-20 min-w-[20rem] border-b-0 p-0">
-                    <div className="flex items-center border-b border-r border-mineshaft-600 px-5 pt-3.5 pb-3">
+                    <div className="flex items-center border-b border-r border-mineshaft-600 px-5 pt-3.5 pb-3 gap-x-2">
+                      <Checkbox
+                        id={`select-all-checkbox`}
+                        isChecked={checkAllEntriesSelected()}
```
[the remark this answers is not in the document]
macci001: Added the use of useMemo in c6d23f8fd33989a1001461db64ba9728fc0151fa commit

maidul98:
This change looks like it only applies to the overview page, can you do the same for within a environment?

macci001:
Added the selectAll feature to the environment page in 81593601bb8579b0dbc438529ad368cb7ff8165d.
I have some questions here:
* Why can we not select the folder in environment page?
* Wouldn't it be a useful feature if we can delete or move the folders across the env from the environment page just like an secret file?

akhilmhdh:
@macci001 Folder moving is temporary not executed due to how snapshots works. Until we have decision for snapshots folder moving is paused.

frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
```diff
@@ -183,6 +185,41 @@ export const SecretOverviewPage = () => {
...
+      secretRecord[secret] = true;
+    })
+
+    setSelectedEntries({
+      [EntryType.FOLDER]: folderRecord,
+      [EntryType.SECRET]: secretRecord
+    })
+  };
+
+  const allSelections = useMemo(() => {
```
akhilmhdh, line 205: Can't we just make it into one useMemo rather than 3

frontend/src/views/SecretMainPage/SecretMainPage.tsx
```diff
@@ -102,6 +104,28 @@ export const SecretMainPage = () => {
     }
   });
 
+  //selectAll functionality
+  const { selectAll: selectAllSecrets } = useSelectedSecretActions();
+  const selectedSecret = useSelectedSecrets();
+
+  const areAllSecretsSelected = useMemo(() => {
```
akhilmhdh, line 111: I don't see a reset when clicking on folder over select all

frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
macci001: There two areas to click on the folder row:
* anywhere except the icons: there will be an route change and on any route change, the store will have an reset. 
* icons (folder icon, edit icon, delete icon): there wouldn't be any reset.

In the above comment: Do you mean that when clicking on the folder icon, there is no reset?
I think no reset on the folder icon click signifies that the folder cannot be selected on the environment page.
Reset on folder icon click might confuse the user.

Please let me know if I misunderstood your comment 🫣

frontend/src/views/SecretMainPage/SecretMainPage.tsx
macci001: There two areas to click on the folder row:
* anywhere except the icons: there will be an route change and on any route change the store will have an reset. 
* icons (folder icon, edit icon, delete icon): there wouldn't be any reset.

In the above comment: Do you mean that when clicking on the folder icon, there is no reset?
I think no reset on the folder icon click signifies that the folder cannot be selected on the environment page.
Reset on folder icon click might confuse the user.

Please let me know if I misunderstood your comment 🫣

frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
macci001: Removed the use of multiple useMemo in 470d0827e5cada3facd0a322a90c3e066e9357a5

maidul98 (changes requested):
- User cannot unselect by toggling the checkbox. 
CleanShot 2024-08-20 at 22 59 49@2x

- Folder should also be selectable

macci001:
Follow-up question on following suggestion @maidul98 : 
* Folder should also be selectable
So for now, the folder selection is only allowed in OverviewPage and not allowed in SecretMainPage.
This PR supports the folder selection in overview page and not in main page following the current pattern.
Would you like to add folder selection in main page through this PR?
If yes, then in secret main page whenever a folder is selected the move option can be disabled as the folder moving is not supported as of now [comment reference](https://github.com/Infisical/infisical/pull/2245#issuecomment-2277532738)

macci001:
Update about the changes in c4c0750869737b00055dfef89c3609965dda8053
* Added the functionality to select folders in main page.
* Removed the de-select all button from both the main and overview page since the added selectAll checkbox in this PR can handle it.
* Also, updated the video clips in the PR description.
@maidul98 @akhilmhdh

PR #2248, opened 2024-08-06
misc: addressed reported flaws with CLI usage

sheensantoscapadngan:
# Description 📣

This PR addresses two issues in the CLI:
1. `secrets set` command requiring local workspace config file even when using auth tokens
2. "not-found" texts are being displayed during `secrets get` even with the `--plain` flag

The PR also adds an API for fetching project ID by slug or name which will make CLI usage easier to automate

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/cli/packages/cmd/secrets.go
+++ b/cli/packages/cmd/secrets.go
@@ -155,22 +155,24 @@ var secretsSetCmd = &cobra.Command{
 	DisableFlagsInUseLine: true,
 	Args:                  cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		util.RequireLocalWorkspaceFile()
-
-		environmentName, _ := cmd.Flags().GetString("env")
-		if !cmd.Flags().Changed("env") {
-			environmentFromWorkspace := util.GetEnvFromWorkspaceFile()
-			if environmentFromWorkspace != "" {
-				environmentName = environmentFromWorkspace
-			}
-		}
-
 		token, err := util.GetInfisicalToken(cmd)
 		if err != nil {
 			util.HandleError(err, "Unable to parse flag")
 		}
 
-		projectId, err := cmd.Flags().GetString("projectId")
+		if (token == nil) {
+			util.RequireLocalWorkspaceFile()
+		}
+
+		environmentName, _ := cmd.Flags().GetString("env")
+		if !cmd.Flags().Changed("env") {
+			environmentFromWorkspace := util.GetEnvFromWorkspaceFile()	
+			if environmentFromWorkspace != "" {
+				environmentName = environmentFromWorkspace
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


backend/src/server/routes/v1/project-router.ts
```diff
@@ -511,4 +512,52 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
+  server.route({
+    method: "GET",
+    url: "/id",
+    config: {
+      rateLimit: readLimit
+    },
+    schema: {
+      querystring: z.object({
+        slug: z.string().trim().optional().describe(PROJECTS.GET_ID.slug),
+        name: z.string().trim().optional().describe(PROJECTS.GET_ID.name)
```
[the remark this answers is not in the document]
akhilmhdh: Yah this will be a bad idea because name can be duplicate in an organization. Will an org have duplicate project name not sure how often. But a name can be duplicate but slug will never be.

backend/src/server/routes/v1/project-router.ts
```diff
@@ -511,4 +512,52 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
       return { serviceTokenData };
     }
   });
+
+  server.route({
+    method: "GET",
+    url: "/id",
```
akhilmhdh, line 518: This feels quite misleading @sheensantoscapadngan 

I would recommend a better name would be /workspace/slug/:slugName rather than as `/workspace/id` as i had very less idea what it meant

docs/api-reference/endpoints/workspaces/get-workspace-id.mdx
```diff
@@ -0,0 +1,4 @@
+---
+title: "Get Project ID"
```
akhilmhdh, line 2: May be its better we say `Get Project by Slug`

akhilmhdh (changes requested):
CLI operations seems to be working fine on my end now.

Only one catch is secrets set when doing it against without a workspace config file and without token doesn't throw any error. It just returns an empty table kinda weird because i didn't understand what happened

docs/api-reference/endpoints/workspaces/get-workspace-id.mdx
sheensantoscapadngan: hmmm main intention of this endpoint is to just fetch the ID of the project...  are you implying we should just make it fetch the entire project? so we support fetching project by ID, name, and slug?

backend/src/server/routes/v1/project-router.ts
sheensantoscapadngan: yeah this one was supposed to just get the ID of the project

akhilmhdh (approved):

akhilmhdh (approved):

PR #2254, opened 2024-08-08
Resolved keyring dataset too big by keeping only the encryption key

akhilmhdh:
# Description 📣

This PR fixes a major issue in CLI that got popped up due to recent release. Some of the users who started using  it suddenly ended up having the new fallback vault when doing normal commands like `run` etc. This caused a big hit on user experience. The fallback should not have triggered because else login would not have happened. 

The root issue was the switch from backup secret for offline access towards the keyring. According to docs keyring cannot sustain more than 3kb in mac. Thus total payload going more than that would fail. This is why we couldn't reproduce this on our end initially. To resolve this we moved back to old way but the encryption key is saved in keyring now.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/cli/packages/cmd/vault.go
+++ b/cli/packages/cmd/vault.go
@@ -31,37 +31,52 @@ var AvailableVaults = []VaultBackendType{
 }
 
 var vaultSetCmd = &cobra.Command{
-	Example:               `infisical vault set file --passphrase <your-passphrase>`,
-	Use:                   "set [file|auto] [flags]",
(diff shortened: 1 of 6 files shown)


maidul98 (approved):
Removed vault use command, and auto generated password because it would never be used by the user anyways. Both `infisical vault set file` and `keyringset` command use auto generted passphrase

PR #2258, opened 2024-08-09
feat(identities/machine): add ability to disable a machine identity

### LINKED ISSUE #2240: Ability to disable machine identities
### Feature description
To stop the use of a machine identity, you currently have to either delete the related client secret or the whole identity itself. However, this not ideal in some cases where you would like to temporarily pause usage. In order to solve this, we can make a feature to enable/disable a identity.

GLEF1X:
# Description 📣

This pull request introduces a new feature that allows the enabling or disabling of machine identities without the need to delete them or their related client secrets. This is particularly useful in scenarios where temporarily pausing the usage of an identity is necessary. The addition of this feature enhances the flexibility and control over machine identities within the system.

### Changes
* Database Update: Added a new `isDisabled` boolean column to the identities table to track the enabled/disabled state of each identity.
* Added the ability to toggle the `isDisabled` state on the identity detail page, allowing users to enable or disable an identity as needed.
* Added `Disabled` column to the identity data table.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

https://github.com/user-attachments/assets/40aa45b5-b06c-407b-89ed-dfb95cac7261

```sh
```

---

Resolves #2240

### DIFF
--- a/backend/src/services/identity-ua/identity-ua-service.ts
+++ b/backend/src/services/identity-ua/identity-ua-service.ts
@@ -57,6 +57,7 @@ export const identityUaServiceFactory = ({
     if (!identityUa) throw new UnauthorizedError({ message: "Invalid credentials" });
 
     const identityMembershipOrg = await identityOrgMembershipDAL.findOne({ identityId: identityUa.identityId });
+    if (identityMembershipOrg.identity.isDisabled) throw new UnauthorizedError({ message: "Identity is disabled" });
 
     checkIPAgainstBlocklist({
       ipAddress: ip,
--- a/backend/src/db/migrations/20240808151815_identity-add-is-disabled-field.ts
+++ b/backend/src/db/migrations/20240808151815_identity-add-is-disabled-field.ts
@@ -1,0 +1,19 @@
+import { Knex } from "knex";
+
+import { TableName } from "@app/db/schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasColumn(TableName.Identity, "isDisabled"))) {
+    await knex.schema.alterTable(TableName.Identity, (tb) => {
+      tb.boolean("isDisabled").notNullable().defaultTo(false);
+    });
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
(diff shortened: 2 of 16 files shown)


dangtony98:
Hey @GLEF1X!

Can you please adjust the UI side here to be inline with the UI/UX for suspending users on the platform? Except instead of suspended it would be **disabled**.

Here's a screenshot of what that would look like:

Screenshot 2024-08-09 at 7 38 18 AM

akhilmhdh:
@GLEF1X @dangtony98 i was thinking may be disable need not be in the edit modal rather just beside the identity details banner as the checkbox. Because disable in edit feels a bit unintuitive and users may not find it

GLEF1X:
@akhilmhdh Thanks for the suggestion! I considered moving the disable option out of the edit modal and placing it beside the identity details banner as a checkbox. However, I felt that following the AWS design approach might be more user-friendly, so I added a status row with text and an icon instead. @dangtony98 I've also made some adjustments to the UI/UX of the table.
 

### AWS examples:
CleanShot 2024-08-18 at 01 38 25@2x
CleanShot 2024-08-18 at 01 38 52@2x

### End result

https://github.com/user-attachments/assets/f9c41ca1-c4d0-4361-8bd8-98da9245e729

backend/src/services/identity-ua/identity-ua-service.ts
```diff
@@ -57,6 +57,7 @@ export const identityUaServiceFactory = ({
     if (!identityUa) throw new UnauthorizedError({ message: "Invalid credentials" });
 
     const identityMembershipOrg = await identityOrgMembershipDAL.findOne({ identityId: identityUa.identityId });
```
akhilmhdh, line 59: I think you neeed to add this in all the identity auth services

akhilmhdh: Rest looks good to me 

CC: @dangtony98

akhilmhdh (changes requested):

Devilla:
is this resolved?

PR #2276, opened 2024-08-12
Client secret cleanup on resource cleanup queue

akhilmhdh:
# Description 📣

This PR adds universal auth client secret cleanup to the resource clean up queue function. This removes all the expired and invalid client secrets. The removal is executed in a batch by batch process with a delay of 10ms for db breathe.

This also adds a new command in make file to ensure PR is ready by checking the UI and api lint and type errors. This is an idea copied from external-secrets operator. Fixed a small bug in audit log pruning also.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/services/audit-log/audit-log-dal.ts
+++ b/backend/src/ee/services/audit-log/audit-log-dal.ts
@@ -75,15 +75,19 @@ export const auditLogDALFactory = (db: TDbClient) => {
           .del()
           .returning("id");
         numberOfRetryOnFailure = 0; // reset
-        // eslint-disable-next-line no-await-in-loop
-        await new Promise((resolve) => {
-          setTimeout(resolve, 100); // time to breathe for db
-        });
       } catch (error) {
         numberOfRetryOnFailure += 1;
         logger.error(error, "Failed to delete audit log on pruning");
+      } finally {
+        // eslint-disable-next-line no-await-in-loop
+        await new Promise((resolve) => {
+          setTimeout(resolve, 10); // time to breathe for db
+        });
       }
-    } while (deletedAuditLogIds.length > 0 && numberOfRetryOnFailure < MAX_RETRY_ON_FAILURE);
+    } while (
+      deletedAuditLogIds.length > 0 ||
+      (numberOfRetryOnFailure > 0 && numberOfRetryOnFailure < MAX_RETRY_ON_FAILURE)
+    );
   };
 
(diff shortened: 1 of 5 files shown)


backend/src/ee/services/audit-log/audit-log-dal.ts
```diff
@@ -75,15 +75,19 @@ export const auditLogDALFactory = (db: TDbClient) => {
...
+      } finally {
+        // eslint-disable-next-line no-await-in-loop
+        await new Promise((resolve) => {
+          setTimeout(resolve, 10); // time to breathe for db
+        });
       }
-    } while (deletedAuditLogIds.length > 0 && numberOfRetryOnFailure < MAX_RETRY_ON_FAILURE);
+    } while (
+      deletedAuditLogIds.length > 0 ||
+      (numberOfRetryOnFailure > 0 && numberOfRetryOnFailure < MAX_RETRY_ON_FAILURE)
```
maidul98, line 89: Why do we need `numberOfRetryOnFailure > 0`, i feel like `numberOfRetryOnFailure < MAX_RETRY_ON_FAILURE` would be enough?

backend/src/services/identity-ua/identity-ua-client-secret-dal.ts
```diff
@@ -23,5 +24,58 @@ export const identityUaClientSecretDALFactory = (db: TDbClient) => {
...
+          .del()
+          .returning("id");
+        numberOfRetryOnFailure = 0; // reset
+      } catch (error) {
+        numberOfRetryOnFailure += 1;
+        logger.error(error, "Failed to delete client secret on pruning");
+      } finally {
+        // eslint-disable-next-line no-await-in-loop
+        await new Promise((resolve) => {
+          setTimeout(resolve, 10); // time to breathe for db
```
maidul98, line 71: I think 10 ms is too little. We should do more like 30 seconds

maidul98 (commented):
Looks good, just left two comments

backend/src/ee/services/audit-log/audit-log-dal.ts
akhilmhdh: We need it because each successful loop we reset the retry because that batch was successfully. Else this would not even retry

maidul98 (approved):

Issue #2277, opened 2024-08-12
Function CreateJunctionTable is not checking, if the table already exists or not


Ayush-Dutt-Sharma:
### Describe the bug

backend/src/db/utils.js 
createJunctionTable --> it is not checking if the table already exists or not, Table1PrimaryKey and Table2PrimaryKey can be dynamic, for bigger files optional Indexing can make querying faster.

dropOnUpdateTrigger, createOnUpdateTrigger, dropUpdatedAtTriggerFunction, createUpdateAtTriggerFunction --> needs better error handling

### To Reproduce
1. Go to backend/src/db/utils.js 

### Screenshots
<img width="858" alt="Screenshot 2024-08-12 at 5 43 13 PM" src="https://github.com/user-attachments/assets/9f3a54ba-c61e-4684-8ac2-aff268d3b5d0">

### Platform you are having the issue on:
Code

akhilmhdh:
Closing as resolved

PR #2278, opened 2024-08-12
CreateJunctionTable checks for table and optional foreign key and better error handling

Ayush-Dutt-Sharma:
# Description 📣 
Fixed --> Function CreateJunctionTable is not checking, if the table already exists or not #2277

Condition to check for existing table
Screenshot 2024-08-12 at 5 54 48 PM

Conditional Indexing on Foreign key
Screenshot 2024-08-12 at 5 55 18 PM

Added better Error Handling

## Type ✨

- [ ✅] Bug fix

- [ ✅] I have read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/getting-started/code-of-conduct). 📝

### DIFF
--- a/backend/src/db/utils.ts
+++ b/backend/src/db/utils.ts
@@ -20,18 +20,15 @@ export const createJunctionTable = async (
     indexForeignKeys = false
   } = options;
 
-  // Check if the table already exists
   const tableExists = await knex.schema.hasTable(tableName);
   
   if (!tableExists) {
     await knex.schema.createTable(tableName, (table) => {
       table.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
       
-      // Create foreign key columns
       table.uuid(`${table1Name}Id`).unsigned().notNullable();
       table.uuid(`${table2Name}Id`).unsigned().notNullable();
       
-      // Set up foreign key constraints
       table.foreign(`${table1Name}Id`)
         .references(table1PrimaryKey)
         .inTable(table1Name)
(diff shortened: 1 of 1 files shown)


backend/src/db/utils.ts
```diff
@@ -1,40 +1,132 @@
...
+  table2Name: TableName,
+  options: JunctionTableOptions = {}
+): Promise<void> => {
+  const {
+    table1PrimaryKey = 'id',
+    table2PrimaryKey = 'id',
+    indexForeignKeys = false
+  } = options;
+
+  try {
```
akhilmhdh, line 23: Please remove all try catch. The migration run in a transaction and if this fails the transaction should fail

backend/src/db/utils.ts
```diff
@@ -1,40 +1,132 @@
...
-FOR EACH ROW
-EXECUTE PROCEDURE on_update_timestamp();
-`);
-
-export const dropOnUpdateTrigger = (knex: Knex, tableName: string) =>
-  knex.raw(`DROP TRIGGER IF EXISTS "${tableName}_updatedAt" ON ${tableName}`);
+
+export const createUpdateAtTriggerFunction = async (knex: Knex): Promise<void> => {
+  try {
+    await knex.raw(`
```
akhilmhdh, line 71: Please remove all try catch

akhilmhdh (changes requested):

backend/src/db/utils.ts
Ayush-Dutt-Sharma: what if i don't remove try catch but throw the error, from the catch block

backend/src/db/utils.ts
Ayush-Dutt-Sharma: added throw error

backend/src/db/utils.ts
akhilmhdh: It doesn't make sense to catch and then throw an error. The error from knex already contains proper context. Please remove it.

Ayush-Dutt-Sharma: removed all try and catch.

akhilmhdh:
@Ayush-Dutt-Sharma please remove the consoles also. Its not needed as i told knex providers a pretty elaborate error message when failing

PR #2284, opened 2024-08-13
Certificate EST protocol (simpleenroll, simplereenroll, cacerts)

sheensantoscapadngan:
# Description 📣
- This PR will enable users to configure and manage EST for a certificate template 
- This PR also includes the implementation of the `/simpleenroll` EST endpoint with client certificate authentication and password-based auth

image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


sheensantoscapadngan:
- update any logic of chain validation to use peculiar/x509
- same with converting certs to pkcs7

backend/src/services/certificate-est/certificate-est-fns.ts
```diff
@@ -0,0 +1,22 @@
+import { Certificate, ContentInfo, EncapsulatedContentInfo, SignedData } from "pkijs";
+
+export const convertRawCertsToPkcs7 = (rawCertificate: ArrayBuffer[]) => {
```
dangtony98, line 3: Use methods from `@peculiar/x509` to reduce room for error; see example at the bottom of the repo [here](https://github.com/PeculiarVentures/x509).

Specifically:

const certs = new x509.X509Certificates([
  new x509.X509Certificate("..."),
  new x509.X509Certificate("..."),
]);

certs.export("base64") // <---

backend/src/services/certificate-template/certificate-template-service.ts
```diff
@@ -187,10 +203,203 @@ export const certificateTemplateServiceFactory = ({
     return certTemplate;
   };
 
+  const createEstConfiguration = async ({
+    certificateTemplateId,
+    caChain,
```
dangtony98, line 208: Add validation for the `caChain` field prior to encrypting and storing it; see `importCertToCa` method.

```
const certificates = certificateChain
      .match(/-----BEGIN CERTIFICATE-----[\s\S]+?-----END CERTIFICATE-----/g)
      ?.map((cert) => new x509.X509Certificate(cert));

    if (!certificates) throw new BadRequestError({ message: "Failed to parse certificate chain" });

    const chain = new x509.X509ChainBuilder({
      certificates
    });

    const chainItems = await chain.build(certObj);

    // chain.build() implicitly verifies the chain
    if (chainItems.length !== certificates.length + 1)
      throw new BadRequestError({ message: "Invalid certificate chain" });
```

backend/src/services/certificate-template/certificate-template-service.ts
```diff
@@ -187,10 +203,203 @@ export const certificateTemplateServiceFactory = ({
...
+    if (caChain) {
+      const kmsEncryptor = await kmsService.encryptWithKmsKey({
+        kmsId: certificateManagerKmsId
+      });
+
+      const { cipherTextBlob: encryptedCaChain } = await kmsEncryptor({
+        plainText: Buffer.from(caChain)
+      });
+
+      updatedData.encryptedCaChain = encryptedCaChain;
```
dangtony98, line 324: Similar to in `createEstConfiguration`, should validate the `caChain` before storing it.

backend/src/services/certificate/certificate-fns.ts
```diff
@@ -24,3 +24,26 @@ export const revocationReasonToCrlCode = (crlReason: CrlReason) => {
       return x509.X509CrlReason.unspecified;
   }
 };
+
+export const convertCertPemToRaw = (certPem: string) => {
+  return new x509.X509Certificate(certPem).rawData;
+};
+
+export const checkCertValidityAgainstChain = async (cert: x509.X509Certificate, chainCerts: x509.X509Certificate[]) => {
```
dangtony98, line 32: Replace method with methods built into `@peculiar/x509`.

Specifically:

```
// validate imported certificate and certificate chain
    const certificates = certificateChain
      .match(/-----BEGIN CERTIFICATE-----[\s\S]+?-----END CERTIFICATE-----/g)
      ?.map((cert) => new x509.X509Certificate(cert));

    if (!certificates) throw new BadRequestError({ message: "Failed to parse certificate chain" });

    const chain = new x509.X509ChainBuilder({
      certificates
    });

    const chainItems = await chain.build(certObj);

    // chain.build() implicitly verifies the chain
    if (chainItems.length !== certificates.length + 1)
      throw new BadRequestError({ message: "Invalid certificate chain" });
```

We can modularize this into a helper function actually to be used elsewhere in the codebase where we need cert-chain validation.

frontend/src/views/Project/CertificatesPage/components/CertificatesTab/components/CertificateTemplatesSection.tsx
dangtony98, line 1: Just noticed `ProjectPermission` on creating certificate templates here is listed for certificates; should be changed.

backend/src/services/certificate-authority/certificate-authority-service.ts
```diff
@@ -1373,6 +1377,8 @@ export const certificateAuthorityServiceFactory = ({
       notAfterDate = new Date(notAfter);
     } else if (ttl) {
       notAfterDate = new Date(new Date().getTime() + ms(ttl));
+    } else if (certificateTemplate?.ttl) {
```
dangtony98, line 1380: Missing in `issueCertFromCa` method. Also, in the certificate template modal it states this field as **Max TTL** but this isn't enforced in this check; this seems to be more like a default TTL if `ttl` is not specified.

backend/src/services/certificate-est/certificate-est-service.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    const leafCertificate = decodeURIComponent(sslClientCert).match(
+      /-----BEGIN CERTIFICATE-----[\s\S]+?-----END CERTIFICATE-----/g
+    )?.[0];
+
+    if (!sslClientCert || !leafCertificate) {
+      throw new UnauthorizedError({ message: "Missing client certificate" });
+    }
+
+    const clientCertBody = leafCertificate
```
dangtony98, line 67: Cert-chain validation can be achieved with `@peculiar/x509` instead as in other comments in this review; using this module would reduce room for error with manual implementation.

backend/src/services/certificate-est/certificate-est-service.ts
```diff
@@ -0,0 +1,271 @@
...
+      });
+    }
+
+    const { rawCertificate } = await certificateAuthorityService.signCertFromCa({
+      isInternal: true,
+      certificateTemplateId,
+      csr
+    });
+
+    return convertRawCertsToPkcs7([rawCertificate]);
```
dangtony98, line 159: `convertRawCertsToPkcs7` can be achieved with `@peculiar/x509` instead as in other comments in this review; using this module would reduce room for error with manual implementation.

backend/src/services/certificate-est/certificate-est-service.ts
```diff
@@ -0,0 +1,271 @@
...
+      certificateTemplateId
+    });
+
+    if (!estConfig.isEnabled) {
+      throw new BadRequestError({
+        message: "EST is disabled"
+      });
+    }
+
+    const leafCertificate = decodeURIComponent(sslClientCert).match(
```
dangtony98, line 185: Cert-chain validation can be achieved with `@peculiar/x509` instead as in other comments in this review; using this module would reduce room for error with manual implementation.

backend/src/services/certificate-est/certificate-est-service.ts
```diff
@@ -0,0 +1,271 @@
...
+      });
+    }
+
+    const { rawCertificate } = await certificateAuthorityService.signCertFromCa({
+      isInternal: true,
+      certificateTemplateId,
+      csr
+    });
+
+    return convertRawCertsToPkcs7([rawCertificate]);
```
dangtony98, line 232: `convertRawCertsToPkcs7` can be achieved with `@peculiar/x509` instead as in other comments in this review; using this module would reduce room for error with manual implementation.

backend/src/services/certificate-est/certificate-est-service.ts
```diff
@@ -0,0 +1,271 @@
...
+      projectDAL,
+      kmsService
+    });
+
+    const caCertRaw = convertCertPemToRaw(caCert);
+    const caParentsRaw = caCertChain
+      .match(/-----BEGIN CERTIFICATE-----[\s\S]+?-----END CERTIFICATE-----/g)
+      ?.map(convertCertPemToRaw);
+
+    return convertRawCertsToPkcs7([caCertRaw, ...(caParentsRaw ?? [])]);
```
dangtony98, line 263: `convertRawCertsToPkcs7` can be achieved with `@peculiar/x509` instead as in other comments in this review; using this module would reduce room for error with manual implementation.

dangtony98 (changes requested):

backend/src/server/routes/est/certificate-est-router.ts
```diff
@@ -0,0 +1,160 @@
...
+import { BadRequestError, UnauthorizedError } from "@app/lib/errors";
+import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
+
+export const registerCertificateEstRouter = async (server: FastifyZodProvider) => {
+  const appCfg = getConfig();
+
+  // add support for CSR bodies
+  server.addContentTypeParser("application/pkcs10", { parseAs: "string" }, (_, body, done) => {
+    try {
+      done(null, (body as string).replace(/\n/g, "").replace(/ /g, ""));
```
dangtony98, line 14: Why do we need `.replace(/\n/g, "").replace(/ /g, ""))` here? — This caused the PEM-encoded CSR to be malformed.

Removing it worked for me.

sheensantoscapadngan: BouncyCastle EST client sends the CSR with "/n" and spaces like the ff:
image

If we remove this, we get the following error during CSR parsing
image


How are you sending the CSR in your tests?

sheensantoscapadngan: As per the sample usage in the RFC, it's sending over the CSR without the begin and end tags
https://datatracker.ietf.org/doc/html/rfc7030#appendix-A.3

image

backend/src/services/certificate-est/certificate-est-fns.ts
sheensantoscapadngan: discussed via DM

sheensantoscapadngan:
Add to docs: 
- EST username

backend/src/services/certificate-authority/certificate-authority-service.ts
sheensantoscapadngan: it's enforced here Tony
image
image

sheensantoscapadngan: Also, `issueCertFromCa` does not have this because it cannot be called internally. 
image

It can only be invoked via API wherein ttl or notAfter is required

dangtony98 (approved):

PR #2295, opened 2024-08-15
fix: switched sync integration to have redis lock

akhilmhdh:
# Description 📣

This PR resolves a major issue that is causing a race condition in sync integration. 

The issues happens when
1. A waterfall or multiple envs having same secret replication
2. Then immediately mutate the ones that are getting replicated
3. Now what will happen is two sync integration will trigger and there will conflict like aws secret delete one throwing failure because the other one already deleted
4. There is another issue in which aws gives invalid result as in even after delete an immediate get actually gives old value as eventual consistency is taking place

To resolve this we put a redis lock over the integration so that only one for an projectid environment slug and secret path happens. The other one will wait until its available. 

Another thing we do is keep a time over last integration in a redis and if it happened just now, we take a pause for 2 second so that external systems can set there state in time.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -18,7 +18,7 @@ export const KeyStorePrefixes = {
 
   SyncSecretIntegrationLock: (projectId: string, environmentSlug: string, secretPath: string) =>
     `sync-integration-mutex-${projectId}-${environmentSlug}-${secretPath}` as const,
-  SetSyncSecretIntegrationLastRunTimestamp: (projectId: string, environmentSlug: string, secretPath: string) =>
+  SyncSecretIntegrationLastRunTimestamp: (projectId: string, environmentSlug: string, secretPath: string) =>
     `sync-integration-last-run-${projectId}-${environmentSlug}-${secretPath}` as const
 };
 
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -5,17 +5,26 @@ import { Redlock, Settings } from "@app/lib/red-lock";
 export type TKeyStoreFactory = ReturnType<typeof keyStoreFactory>;
 
 // all the key prefixes used must be set here to avoid conflict
-export enum KeyStorePrefixes {
-  SecretReplication = "secret-replication-import-lock",
-  KmsProjectDataKeyCreation = "kms-project-data-key-creation-lock",
-  KmsProjectKeyCreation = "kms-project-key-creation-lock",
-  WaitUntilReadyKmsProjectDataKeyCreation = "wait-until-ready-kms-project-data-key-creation-",
-  WaitUntilReadyKmsProjectKeyCreation = "wait-until-ready-kms-project-key-creation-",
-  KmsOrgKeyCreation = "kms-org-key-creation-lock",
-  KmsOrgDataKeyCreation = "kms-org-data-key-creation-lock",
-  WaitUntilReadyKmsOrgKeyCreation = "wait-until-ready-kms-org-key-creation-",
-  WaitUntilReadyKmsOrgDataKeyCreation = "wait-until-ready-kms-org-data-key-creation-"
-}
+export const KeyStorePrefixes = {
+  SecretReplication: "secret-replication-import-lock",
+  KmsProjectDataKeyCreation: "kms-project-data-key-creation-lock",
+  KmsProjectKeyCreation: "kms-project-key-creation-lock",
+  WaitUntilReadyKmsProjectDataKeyCreation: "wait-until-ready-kms-project-data-key-creation-",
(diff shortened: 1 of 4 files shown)


backend/src/keystore/keystore.ts
```diff
@@ -5,17 +5,26 @@ import { Redlock, Settings } from "@app/lib/red-lock";
...
+  WaitUntilReadyKmsProjectDataKeyCreation: "wait-until-ready-kms-project-data-key-creation-",
+  WaitUntilReadyKmsProjectKeyCreation: "wait-until-ready-kms-project-key-creation-",
+  KmsOrgKeyCreation: "kms-org-key-creation-lock",
+  KmsOrgDataKeyCreation: "kms-org-data-key-creation-lock",
+  WaitUntilReadyKmsOrgKeyCreation: "wait-until-ready-kms-org-key-creation-",
+  WaitUntilReadyKmsOrgDataKeyCreation: "wait-until-ready-kms-org-data-key-creation-",
+
+  SyncSecretIntegrationLock: (projectId: string, environmentSlug: string, secretPath: string) =>
+    `sync-integration-mutex-${projectId}-${environmentSlug}-${secretPath}` as const,
+  SetSyncSecretIntegrationLastRunTimestamp: (projectId: string, environmentSlug: string, secretPath: string) =>
```
maidul98, line 21: nit: I think we can omit `set` because this is just formatting a key. `SyncSecretIntegrationLastRunTimestamp` would be better

backend/src/services/secret/secret-queue.ts
```diff
@@ -641,111 +643,155 @@ export const secretQueueFactory = ({
...
-          iv: integrationAuth.awsAssumeIamRoleArnIV,
-          tag: integrationAuth.awsAssumeIamRoleArnTag,
-          key: botKey as string
-        });
+    const lastRunSyncIntegrationTimestamp = await keyStore.getItem(
+      KeyStorePrefixes.SetSyncSecretIntegrationLastRunTimestamp(projectId, environment, secretPath)
+    );
+
+    if (lastRunSyncIntegrationTimestamp) {
+      const isStaleSyncIntegration = new Date(job.timestamp) < new Date(lastRunSyncIntegrationTimestamp);
```
maidul98, line 661: will the job time stamp be when the job was triggered?

backend/src/services/secret/secret-queue.ts
```diff
@@ -641,111 +643,155 @@ export const secretQueueFactory = ({
...
-        Object.keys(secrets).forEach((key) => {
-          const prefix = metadata?.secretPrefix || "";
-          const suffix = metadata?.secretSuffix || "";
-          const newKey = prefix + key + suffix;
-          suffixedSecrets[newKey] = secrets[key];
+      const timeDifferenceWithLastIntegration = getTimeDifferenceInSeconds(
+        lockAcquiredTime.toISOString(),
+        lastRunSyncIntegrationTimestamp
+      );
+      if (timeDifferenceWithLastIntegration < 2000 && timeDifferenceWithLastIntegration > 0)
```
maidul98, line 673: nit: i think this 2000 should be a variable const

backend/src/services/secret/secret-queue.ts
```diff
@@ -641,111 +643,155 @@ export const secretQueueFactory = ({
...
+        await new Promise((resolve) => {
+          setTimeout(resolve, 2000 - timeDifferenceWithLastIntegration * 1000);
         });
-      }
+    }
+
```
maidul98, line 678: I think this will be better put in a function so it makes things easer to read and reusable if needed 

```
const delayIfNeeded = async (lockAcquiredTime, lastRunSyncTime) => {
  const timeSinceLastSync = getTimeDifferenceInSeconds(
    lockAcquiredTime.toISOString(),
    lastRunSyncTime
  );

  const minimumInterval = 2000; 
  const delayTime = Math.max(0, minimumInterval - timeSinceLastSync * 1000);

  if (delayTime > 0) {
    await new Promise(resolve => setTimeout(resolve, delayTime));
  }
};

await delayIfNeeded(lockAcquiredTime, lastRunSyncIntegrationTimestamp);

```

backend/src/services/secret/secret-queue.ts
akhilmhdh: yes

backend/src/services/secret/secret-queue.ts
akhilmhdh: Changed

backend/src/keystore/keystore.ts
akhilmhdh: Changed

maidul98 (approved):

PR #2302, opened 2024-08-17
feat: added pagination to secrets dashboard

### LINKED ISSUE #2224: Add pagination to secrets dashboard to reduce lag
### Feature description
Right now, all secrets are fetched in at a given secret path. However, when the number of secrets goes past 500, the UI will start o lag. In order to resolve this, we need to introduce pagination so that each page is limited to the number of secrets it can show. All search and download functions must work as expected as apart of this change.

Anonymous961:
Closes #2224

# Description 📣
Added pagination to the secrets dashboard. The number of secrets per page is currently fixed at 10. Additionally, a button can be added to allow users to change the number of items per page.

The page now looks like this:
image
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

### DIFF
--- a/frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
+++ b/frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
@@ -468,6 +468,29 @@ export const SecretOverviewPage = () => {
     filteredFolderNames?.length === 0 &&
     filteredDynamicSecrets?.length === 0;
 
+  const combinedItems = [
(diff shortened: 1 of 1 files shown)


Anonymous961:
Replaced custom pagination component with existing pagination component.
image
image

maidul98:
I'm only seeing frontend changes. Pagination should also happen on the backend.

Anonymous961:
I'm working on it. Would soon make a commit.

PR #2310, opened 2024-08-20
Feat: Azure DevOps Integration

DanielHougaard:
# Description 📣

This PR introduces a new Azure DevOps integration. This work is heavily based on @meetcshah19's work.

A few key changes:
1. Changed referencing from "Azure DevOps org ID" to "org name". There's a difference between organization ID's and organization names in Azure. We are actually using the Azure org name, not the org ID as it is referred to in Meet's PR.
2. Added unique parameter for azure devops org name (previously the teamId parameter was being reused)
3. Minor rephrasing in the create integration step.
4. Smaller overall improvements. 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


akhilmhdh (commented):
Asked @DanielHougaard to check for pagination in azure, to handle that case also as it may later bit us back

DanielHougaard:
Tested with 350 projects and 1000+ environments, no indication of pagination needed

PR #2311, opened 2024-08-20
Fix: Remove notification when unable to read secrets from environment

DanielHougaard:
# Description 📣

Removed a notification alert that would appear if a user doesn't have access to a certain environment. Since we load all environments on the load page, this notification would always appear if the user has restricted access (as in, they can't read production but can read all other environments as an example).

The fix for this is very straightforward, as we just filter out the "You are not allowed to read on secrets" error.

I've added a new `constants.ts` file which contains the error message as a variable. This is done to follow @dangtony98's file structure that he used when implementing certificates. Example can be [found here](https://github.com/Infisical/infisical/blob/main/frontend/src/hooks/api/certificates/constants.tsx). 

CleanShot 2024-08-20 at 21 49 37@2x

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/hooks/api/secrets/constants.ts
+++ b/frontend/src/hooks/api/secrets/constants.ts
@@ -1,0 +1,1 @@
+export const ERROR_NOT_ALLOWED_READ_SECRETS = "You are not allowed to read on secrets";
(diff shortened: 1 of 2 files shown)


maidul98 (approved):
Hot fix for now but i assume the real one would be to not call this api at all if the user doesn't have access to it

Issue #1582, opened 2024-03-17
Automatically validate the email verification code (no need to press <ENTER>)


Grraahaam:
### Feature description

Today, when I create an account, I'm receiving the verification code to sign in by email.  I then need to copy and paste it in the app, after what I need to press <kbd>ENTER</kbd> or the submit button.

1. We could validate the code directly when pasting it in the input, or once all the fields have been populated?
2. I've seen a topic somewhere here about replacing the verification code by a link to simplify this process

What do you think?

### Why would it be useful?

It'd improve the UX and simplify the user's verification process

### Additional context

Related issues:
- https://github.com/Infisical/infisical/issues/58
- https://github.com/Infisical/infisical/pull/683
- https://github.com/Infisical/infisical/pull/1529
- https://github.com/Infisical/infisical/issues/555

kavinda1995:
@Grraahaam I can work on this issue if anyone else is not working currently

DanielHougaard:
@Grraahaam I agree, that would be even better. I wouldn't start doing that just yet though. We need to make the emails nicer in general at some point, and I think such a change would fall pretty nicely into that scope. 

For now I think making it automatically continue efter entering the code is the right move!

DanielHougaard:
Oh wow I didn't see that one, awesome! React email is great, and this is actually what I had in mind for the mail refactor. Currently it's not a _huge_ priority, but it will defiantly become a more pressing topic in the future.

vishavsingla:
Is this issue resolved?

siddhant-agrawal01:
Is the issue resolved?

akshat99812:
I'll take this if it's available

rahuljagtapp:
@Grraahaam I can work on this issue  if issue is open

saloni0419:
@Grraahaam Can you please assign me this issue, if it is still open

Bhaveshshah777:
@DanielHougaard Here is a pull request to solve this bug , #2977 . Hope this works

Ananmayd:
@Grraahaam I would love to work on this issue if it is still open.

Siddharth-2382:
@DanielHougaard I've made the changes and created a PR. I'd appreciate it if someone could review it and get it merged.

Definisi:
Makes sense, no reason to require an extra click/enter when the code is already complete.

[...]

PR #2312, opened 2024-08-20
Automatically validate the email verification code

BKM14:
# Description 📣

Solves [#1582](https://github.com/Infisical/infisical/issues/1582)

Added another prop 'code' to the CodeInputStep component to keep track of the code entered so far. A useEffect is utilized to run whenever the code value changes. The incrementStep() function is only called when the length of the code is 6 and it is not equal to "123456".

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Used a local mailhog server to send dummy emails. Then generated the verification codes and tested it. An error message is displayed if the code isn't correct. If the code is correct, it re-directs to the next page, without having to press the enter key again.

---


BKM14:
@DanielHougaard, can you review you this please?
Thank you

PR #2317, opened 2024-08-21
feat: added identity and service token postgres update debounced

akhilmhdh:
# Description 📣

This PR improves infisical performance by reduction in SQL write for identity and service token lastUsed at. It achieves this debounce technique by using redis and our queue. 

Its first written to redis and then a queue job is pushed. Given its a same job id, duplicate queue pushed won't be considered. There will only be first one pushed and its delayed. After the delay its written towards postgres.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -169,1093 +169,19 @@
       }
     },
     "node_modules/@aws-crypto/crc32": {
-      "version": "3.0.0",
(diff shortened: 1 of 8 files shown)


maidul98 (approved):
makes sense, nice work

PR #2319, opened 2024-08-21
Feat: Redis support for dynamic secrets

DanielHougaard:
# Description 📣

I've added support for Redis dynamic secrets. The default statements work via. the Redis ACL for provisioning new users & deleting users. The renewal / revocation logic remains inline with our existing dynamic secrets.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


akhilmhdh (changes requested):
Screenshot 2024-08-22 at 3 46 41 PM

Everything working as expected just some tiny things
1. The default port should be 6379
2. Can you remove that space after the password field that is make the username and password field take the whole space
3. Please test it against aws elastic cache also

backend/src/lib/aws/elasticache.ts
```diff
@@ -0,0 +1,208 @@
+import { Sha256 } from "@aws-crypto/sha256-js";
+import {
+  CreateUserCommand,
```
akhilmhdh, line 3: @DanielHougaard as i have guessed its, it going way too much for a single provider. 

1. We need to move this logic to dynamic secret service. This should not be in lib as lib is for more reusable code.
2. Elastic cache seems to contain lot more logic than the redis provider considers. Its better we put this as another provider. Let's keep the aws change on hold or rather do it later. Because consider somebody requesting azure for the same the same provider would now handle more logic for that one too which doesn't make sense.

backend/src/ee/services/dynamic-secret/providers/models.ts
```diff
@@ -7,6 +7,52 @@ export enum SqlProviders {
...
+
+export const DynamicSecretRedisDBSchema = z
+  .object({
+    client: z.nativeEnum(RedisProviders),
+    host: z.string().trim().toLowerCase(),
+    port: z.number(),
+    username: z.string().trim(), // this is often "default".
+    password: z.string().trim().optional(),
+
+    elastiCacheIamUsername: z.string().trim().optional(),
```
akhilmhdh, line 23: Next issue for me here is schema, we need a precise schema. 

That is as client changes the schema should also change using `discriminatedUnion` because the validation in refine works but won't be shown to end user. This is the issue we are facing with `integration` api at the moment.

The docs won't show this data in the json schema as its missing. We need the schema to show it. That is its redis only these values are needed

backend/src/lib/aws/elasticache.ts
akhilmhdh: 3. AWS access could be in multiple ways too like using `assumeRole` too.

akhilmhdh (changes requested):

backend/src/ee/services/dynamic-secret/providers/aws-elasticache.ts
```diff
@@ -0,0 +1,105 @@
...
+const generateUsername = () => {
+  const charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-";
+  return `inf-${customAlphabet(charset, 32)()}`; // Username must start with an ascii letter, so we prepend the username with "inf-"
+};
+
+export const AwsElastiCacheDatabaseProvider = (): TDynamicProviderFns => {
+  const validateProviderInputs = async (inputs: unknown) => {
+    const providerInputs = DynamicSecretAwsElastiCacheSchema.parse(inputs);
+
+    JSON.parse(providerInputs.creationStatement);
```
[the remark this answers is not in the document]
akhilmhdh: Also can we please keep everything related to dynamic secret here. Keeping it lib is really confusing at the moment. I had to jump between two files to understand the logic and it seems quite hard to understand. 

I think if we keep those functions here it will much lean to understand?

docs/documentation/platform/dynamic-secrets/aws-elasticache.mdx
```diff
@@ -0,0 +1,145 @@
...
+	<ParamField path="Secret Access Key" type="string" required>
+		This is the secret access key of the AWS IAM user you created in the prerequisites. This will be used to provision and manage the dynamic secret leases.
+	</ParamField>
+
+	<ParamField path="CA(SSL)" type="string">
+		A CA may be required if your DB requires it for incoming connections. This is often the case when connecting to a managed service.
+	</ParamField>
+
+  </Step>
+  <Step title="(Optional) Modify Redis Statements">
```
akhilmhdh, line 97: I guess docs are pending to update?

backend/src/ee/services/dynamic-secret/providers/aws-elasticache.ts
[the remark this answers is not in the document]
DanielHougaard: > I don't see where we are using this 🤔 

We need to ensure the that the creation & revocation statements are valid and can be used to create and revoke users. We can't return the parsed statements here because we need to use the handlebars template to generate the username and password, before we can use the parsed statements

> Also can we please keep everything related to dynamic secret here. Keeping it lib is really confusing at the moment. I had to jump between two files to understand the logic and it seems quite hard to understand.
I think if we keep those functions here it will much lean to understand?

Done!

akhilmhdh (approved):

PR #2321, opened 2024-08-22
Feat: Select roles & projects when inviting members to organization

DanielHougaard:
# Description 📣

This PR aims to add support for inviting multiple members to organizations at once. Additionally, it also makes it possible to select which organization role the user should be granted once they accept the invite. 

The PR also introduces project invites with role assignment directly from the organization invite tab. This is especially useful if you have a large number of projects and a new member has joined the team.

#### Token metadata
Due to the nature of our current encryption logic, we can't add users to projects before they have fully set up their account _(An example of this would be to invite someone to your organization that has never used Infisical before)_. For this case I've introduced a new concept called "Token metadata", which is a JWT token that contains extra to flows like the signup flow. It's appended to the signup URL that the user is given.

The token metadata is **only** used if the user doesn't already have an Infisical account. Incase the user already has an account with Infisical, then the project memberships will be created as a part of the invite sending, and become usable after the user has accepted the organization invite.

#### Minor
The token metadata contains a list of project ID's, the user's ID, the roles to assign once added to the project, and the organization ID. The token metadata is used in the complete account step, where it will create the project memberships right after creating the user's encryption keys keypair, which is needed for project invites.

The logic needed inviting members has been refactored from living inside the project membership service, to now living inside the project membership fns file, where it can then be used independently from other parts of the codebase.

### Other small unrelated changes:
1. When logging into Infisical you are always prompted to select which organization you wish to log into. This PR contains a change that will skip the organization selection if you are only a part of one organization.
2. Removed `https://*.*.posthog.com` from the next config's content security policy, as per requested by @maidul98

## Screenshot of the new organization invite modal.
CleanShot 2024-08-23 at 02 22 44

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


backend/src/services/auth/auth-signup-service.ts
```diff
@@ -358,6 +370,45 @@ export const authSignupServiceFactory = ({
         tx
       );
 
+      if (metadata) {
```
akhilmhdh, line 373: May be we can call this something specific rather than a generic metadata. Because later when we read back we may not understand this at all.

backend/src/services/auth/auth-signup-service.ts
```diff
@@ -358,6 +370,45 @@ export const authSignupServiceFactory = ({
...
+        const metadataObj = jwt.verify(metadata, appCfg.AUTH_SECRET) as TTokenMetadata;
+
+        if (
+          metadataObj.payload.userId !== user.id ||
```
akhilmhdh, line 377: Better to add an `?` nullable coalescing check rather than dot.

backend/src/services/org/org-service.ts
```diff
@@ -450,98 +476,207 @@ export const orgServiceFactory = ({
...
+              }
+
+              if (!projectRoleSlug) {
+                throw new BadRequestError({
+                  message: "Selecting a project role is required to invite users to projects"
+                });
+              }
+
+              for await (const projectId of projectIds) {
+                await projectMembershipDAL.create(
```
akhilmhdh, line 550: I think you can move this to an insert many

akhilmhdh (changes requested):

akhilmhdh:
Tested it out. Working as expected. Good job @DanielHougaard. Just the metadata may be a bit misleading

akhilmhdh (approved):

PR #2326, opened 2024-08-22
Fix: Disallow org members to invite new members

DanielHougaard:
# Description 📣

Previously it was possible for organization members to invite new members, which isn't inline with the rest of our member management permissions on organization-level. This PR reduces the permission set for members so they're only able to read on member resources.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/services/permission/org-permission.ts
+++ b/backend/src/ee/services/permission/org-permission.ts
@@ -126,7 +126,7 @@ const buildMemberPermission = () => {
   can(OrgPermissionActions.Read, OrgPermissionSubjects.Workspace);
   can(OrgPermissionActions.Create, OrgPermissionSubjects.Workspace);
   can(OrgPermissionActions.Read, OrgPermissionSubjects.Member);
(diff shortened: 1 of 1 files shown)


backend/src/ee/services/permission/org-permission.ts
```diff
@@ -126,7 +126,7 @@ const buildMemberPermission = () => {
   can(OrgPermissionActions.Read, OrgPermissionSubjects.Workspace);
   can(OrgPermissionActions.Create, OrgPermissionSubjects.Workspace);
   can(OrgPermissionActions.Read, OrgPermissionSubjects.Member);
-  can(OrgPermissionActions.Create, OrgPermissionSubjects.Member);
+  can(OrgPermissionActions.Read, OrgPermissionSubjects.Member);
```
maidul98, line 129: These are now dups, we can delete this

maidul98 (approved):

PR #2331, opened 2024-08-24
Feat: Persistent self-hosting domains on `infisical login`

DanielHougaard:
# Description 📣

This PR implements persistent self-hosting domains for `infisical login`. Additionally, it also incldues a new flag on the login command, called `clear-domains`. In some cases users may want to wipe their saved domains, and this flag does exactly yhat.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


akhilmhdh (commented):
A small case missing. That is filtering of same domains. I can enter the same domain twice and the list will have two same domains.

akhilmhdh (approved):

PR #2333, opened 2024-08-26
Feat: Email notification on secret change requests

DanielHougaard:
# Description 📣

This PR aims to add support for email notifications for secret change requests. Whenever a new secret change request is created, the reviewers will be notified via. email that their attention is needed.

The email will include a direct link to the change request.

Additionally: This PR also resolves an issue with creation secret change policies.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/ee/services/secret-approval-request/secret-approval-request-fns.ts
+++ b/backend/src/ee/services/secret-approval-request/secret-approval-request-fns.ts
@@ -28,21 +28,19 @@ export const sendApprovalEmailsFn = async ({
 
   // now we need to go through each of the reviewers and print out all the commits that they need to approve
   for await (const reviewerUser of policy.userApprovers) {
-    if (reviewerUser) {
-      await smtpService.sendMail({
-        recipients: [reviewerUser?.email as string],
-        subjectLine: "Infisical Secret Change Request",
+    await smtpService.sendMail({
+      recipients: [reviewerUser?.email as string],
+      subjectLine: "Infisical Secret Change Request",
 
-        substitutions: {
-          firstName: reviewerUser.firstName,
-          projectName: project.name,
-          organizationName: project.organization.name,
-          approvalUrl: `${cfg.isDevelopmentMode ? "https" : "http"}://${cfg.SITE_URL}/project/${
-            project.id
-          }/approval?requestId=${secretApprovalRequest.id}`
-        },
-        template: SmtpTemplates.SecretApprovalRequestNeedsReview
-      });
-    }
+      substitutions: {
+        firstName: reviewerUser.firstName,
+        projectName: project.name,
+        organizationName: project.organization.name,
+        approvalUrl: `${cfg.isDevelopmentMode ? "https" : "http"}://${cfg.SITE_URL}/project/${
+          project.id
+        }/approval?requestId=${secretApprovalRequest.id}`
+      },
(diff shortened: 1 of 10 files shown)


backend/src/ee/services/secret-approval-request/secret-approval-request-fns.ts
```diff
@@ -0,0 +1,48 @@
...
+}: TSendApprovalEmails) => {
+  const cfg = getConfig();
+
+  const policy = await secretApprovalPolicyDAL.findById(secretApprovalRequest.policyId);
+
+  const project = await projectDAL.findProjectWithOrg(projectId);
+
+  // now we need to go through each of the reviewers and print out all the commits that they need to approve
+  for await (const reviewerUser of policy.userApprovers) {
+    if (reviewerUser) {
```
maidul98, line 31: Why would this be falsey? Won't this be a legit item

frontend/src/views/SecretApprovalPage/components/SecretApprovalRequest/SecretApprovalRequest.tsx
```diff
@@ -66,10 +67,19 @@ export const SecretApprovalRequest = () => {
   const { user: userSession } = useUser();
   const { permission } = useProjectPermission();
   const { data: members } = useGetWorkspaceUsers(workspaceId);
-  const isSecretApprovalScreen = Boolean(selectedApproval);
+  const isSecretApprovalScreen = Boolean(selectedApprovalId);
+  const queryParams = new URLSearchParams(window.location.search);
```
maidul98, line 71: You can use the https://nextjs.org/docs/app/api-reference/functions/use-search-params instead

maidul98 (commented):
Left some small comments

backend/src/ee/services/secret-approval-request/secret-approval-request-fns.ts
DanielHougaard: This is true! Sorry this check is a result of an old implementation I was playing around with.

frontend/src/views/SecretApprovalPage/components/SecretApprovalRequest/SecretApprovalRequest.tsx
DanielHougaard: @maidul98 I believe next/navigation was added in Next 13, we're still on version 12.3.

CleanShot 2024-08-26 at 20 52 36

akhilmhdh: Ohh actually useRouter has query component with it

DanielHougaard: @akhilmhdh I wonder why we use `URLSearchParams` everywhere else? I remember we discussed this a while back, I don't remember the exact reason for using `URLSearchParams` though 🤔

maidul98: nice catch, i see it here https://nextjs.org/docs/pages/api-reference/functions/use-router#router-object

akhilmhdh (approved):

PR #2341, opened 2024-08-27
feat(cli): `run` watch mode

DanielHougaard:
# Description 📣

This PR aims to add support for secret watching. The user will now be able to pass a `--watch` flag to the `infisical run` command. When the `watch` flag is passed, Infisical will automatically reload the command passed to `infisical run`.

### New flow example

1. User runs `infisical run --watch --comand="while true; do printenv; sleep 500000; done"`. _(Both --command and -- supports the new `watch` flag)_
   * At first this acts exactly like it always has, and prints the environment to the console and hangs until interrupted.
2. The user now changes a secret in Infisical.
3. At some point within the next 10 seconds (due to polling), the application will detect a change in the Infisical secrets, and then trigger a reload of the application.
4. The application reloads and the new secrets are loaded into the application. 

This is especially useful in local development setups, where the user usually would have to manually restart the entire process for the new secrets to load in. This implementation streamlines everything and in a sense acts like [Nodemon](https://github.com/remy/nodemon) acts when detecting changes in a Node.js project.

This can also be used in production environments where you'd want to reload an application when secret changes happen in Infisical. 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/cli/packages/cmd/run.go
+++ b/cli/packages/cmd/run.go
@@ -22,6 +22,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var INTERRUPT_ERR = fmt.Errorf("signal: interrupt")
+
 // runCmd represents the run command
 var runCmd = &cobra.Command{
 	Example: `
@@ -250,6 +252,10 @@ func executeSingleCommandWithEnvs(args []string, secretsCount int, env []string,
 
 	err := startCmd() // Initial command start, if no --watch flag is passed, it will work like in old versions of infisical CLI.
 	if err != nil {
+		if err.Error() == INTERRUPT_ERR.Error() {
+			log.Debug().Msg(color.HiMagentaString("Process was terminated manually by the user"))
+			os.Exit(1)
+		}
 		util.HandleError(err, "Failed to start command")
 	}
 
@@ -343,6 +349,10 @@ func executeMultipleCommandWithEnvs(fullCommand string, secretsCount int, env []
 
 	err := startCmd() // Initial command start, if no --watch flag is passed, it will work like in old versions of infisical CLI.
 	if err != nil {
+		if err.Error() == INTERRUPT_ERR.Error() {
+			log.Debug().Msg(color.HiMagentaString("Process was terminated manually by the user"))
+			os.Exit(1)
+		}
 		util.HandleError(err, "Failed to start command")
 	}
 
--- a/cli/packages/cmd/run.go
+++ b/cli/packages/cmd/run.go
@@ -218,86 +218,51 @@ func executeSingleCommandWithEnvs(args []string, secretsCount int, env []string,
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
 
 	if reloadParameters.Enabled {
-		log.Info().Msgf(color.YellowString("[HOT RELOAD] Watching for secret changes..."))
-		go func() {
-			<-sigChan
-			log.Info().Msg("Received termination signal. Cleaning up...")
-			cancelCtx()
-		}()
+		handleHotReloadCleanup(sigChan, cancelCtx)
 	}
 
-	var cmd *exec.Cmd
+	var currentCmd *exec.Cmd
 
 	startCmd := func() error {
+		if currentCmd != nil {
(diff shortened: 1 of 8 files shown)


cli/packages/util/helper.go
```diff
@@ -298,3 +299,16 @@ func GenerateRandomString(length int) string {
...
+	content := []byte{}
+
+	for _, secret := range sortedSecrets {
+		content = append(content, []byte(secret.Key)...)
+		content = append(content, []byte(secret.Value)...)
+	}
+
+	hash := sha256.Sum256(content)
+	return fmt.Sprintf(`"%s"`, hex.EncodeToString(hash[:]))
+}
```
sheensantoscapadngan, line 314: this is probably something we'll want to add to our backend response in the future. I can see this being useful for other use-cases

cli/packages/cmd/run.go
```diff
@@ -237,80 +345,95 @@ func filterReservedEnvVars(env map[string]models.SingleEnvironmentVariable) {
...
+		environmentVariables[k] = v.Value
+	}
 
-	if err := cmd.Start(); err != nil {
-		return err
+	// Create and sort the env slice using slices.SortFunc
+	env := make([]string, 0, len(environmentVariables))
+	for key, value := range environmentVariables {
+		env = append(env, key+"="+value)
 	}
```
sheensantoscapadngan, line 413: wait I don't see the sort anywhere?

cli/packages/cmd/run.go
```diff
@@ -116,108 +155,177 @@ var runCmd = &cobra.Command{
...
+				os.Exit(exitCode)
 			}
+		}()
+	}
 
-		} else {
-			err = executeSingleCommandWithEnvs(args, len(secretsByKey), env)
+	initialEnvironment, err := createInjectableEnvironment(request, projectConfigDir, secretOverriding, expandSecrets, token)
+	if err != nil {
+		util.HandleError(err, "[HOT RELOAD] Failed to fetch secrets")
```
sheensantoscapadngan, line 276: this one's not necessarily related to reload/watch-mode right?

cli/packages/cmd/run.go
```diff
@@ -116,108 +155,177 @@ var runCmd = &cobra.Command{
...
+	}
+	startProcess(initialEnvironment)
+	recheckSecretsChannel := make(chan bool, 1)
+
+	// this is the only logic strictly related to watch mode, the rest is shared with non-watch mode
+	if watchMode {
+		log.Info().Msg(color.HiMagentaString("[HOT RELOAD] Watching for secret changes..."))
+
+		// a simple goroutine that triggers the recheckSecretsChan every 5 seconds
```
sheensantoscapadngan, line 285: every watch interval

cli/packages/util/helper.go
DanielHougaard: We actually already do this on the serverside, but I recall the ETag being unreliable and not working. So my solution to get around it is to just compute the Etag on the client side

cli/packages/cmd/run.go
DanielHougaard: Sorry, leftover comment. We only need to sort the secrets, because if we don't sort it, we can't be sure that the ETag is reproducible

cli/packages/cmd/run.go
```diff
@@ -116,108 +155,177 @@ var runCmd = &cobra.Command{
...
+		secretsFetchedAt := time.Now()
+		if secretsFetchedAt.After(lastSecretsFetch) {
+			lastSecretsFetch = secretsFetchedAt
 		}
 
-		if shouldExpandSecrets {
+		shouldRestartProcess := cmd != nil
+		// terminate the old process before starting a new one
+		if shouldRestartProcess {
+			beingTerminated = true
```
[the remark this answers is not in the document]
DanielHougaard: Yes, we set `beingTerminated` to true at the beginning of the restart process flow. Once the restart flow is over, we set it to false. 

We use `beingTerminated` to filter out errors in a goroutine (see line 250-270)

sheensantoscapadngan (approved):

PR #2353, opened 2024-08-29
feat(cli): `run` command watch mode

DanielHougaard:
# Description 📣

This PR aims to add support for secret watching. The user will now be able to pass a `--watch` flag to the `infisical run` command. When the `watch` flag is passed, Infisical will automatically reload the command passed to `infisical run`.

### New flow example

1. User runs `infisical run --watch --comand="while true; do printenv; sleep 500000; done"`. _(Both --command and -- supports the new `watch` flag)_
   * At first this acts exactly like it always has, and prints the environment to the console and hangs until interrupted.
2. The user now changes a secret in Infisical.
3. At some point within the next 10 seconds (due to polling), the application will detect a change in the Infisical secrets, and then trigger a reload of the application.
4. The application reloads and the new secrets are loaded into the application. 

This is especially useful in local development setups, where the user usually would have to manually restart the entire process for the new secrets to load in. This implementation streamlines everything and in a sense acts like [Nodemon](https://github.com/remy/nodemon) acts when detecting changes in a Node.js project.

This can also be used in production environments where you'd want to reload an application when secret changes happen in Infisical. 

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/cli/packages/cmd/run.go
+++ b/cli/packages/cmd/run.go
@@ -243,7 +243,9 @@ func executeSpecifiedCommand(commandFlag string, args []string, watchMode bool, 
 
 		// start the process
 		log.Info().Msgf(color.GreenString("Injecting %v Infisical secrets into your application process", environment.SecretsCount))
-		cmd, err = util.RunCommand(commandFlag, args, environment.Variables)
+
+		shouldWaitForExit := !watchMode
+		cmd, err = util.RunCommand(commandFlag, args, environment.Variables, shouldWaitForExit)
 		if err != nil {
 			defer WaitGroup.Done()
 			util.HandleError(err)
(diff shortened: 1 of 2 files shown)


callumfrederiksen (approved):

maidul98 (changes requested):
Left feedback via slack

cli/packages/cmd/run.go
```diff
@@ -314,3 +286,226 @@ func execCmd(cmd *exec.Cmd) error {
...
+
+	initialEnvironmentVariables, err := fetchAndFormatSecretsForShell(request, projectConfigDir, secretOverriding, expandSecrets, token)
+	if err != nil {
+		util.HandleError(err, "Failed to fetch secrets")
+	}
+
+	runCommandWithWatcher(initialEnvironmentVariables)
+
+	// this is the only logic strictly related to watch mode, the rest is shared with non-watch mode
+	if watchMode {
```
maidul98, line 414: since this function IS for watch mode, do we need this any more?

maidul98 (commented):
Otherwise, the refactor lgtm

cli/packages/cmd/run.go
```diff
@@ -314,3 +286,217 @@ func execCmd(cmd *exec.Cmd) error {
...
+				os.Exit(exitCode)
+			}
+		}()
+	}
+
+	recheckSecretsChannel := make(chan bool, 1)
+	recheckSecretsChannel <- true
+
+	// a simple goroutine that triggers the recheckSecretsChan every watch interval (defaults to 10 seconds)
+	go func() {
```
maidul98, line 412: Instead of making a go routine, a more resource efficent way of doing this is using `time.Ticker`. Here is a example of this 

```
ticker := time.NewTicker(time.Duration(watchModeInterval) * time.Second)
defer ticker.Stop()

for {
  select {
  case <-ticker.C:
      watchMutex.Lock()
      // logic
      watchMutex.Unlock()
  }
}
```

maidul98 (commented):
One last comment and should be ready for merge

maidul98 (approved):

PR #2354, opened 2024-08-30
Fix: Security attributes for dynamic secret endpoints

DanielHougaard:
# Description 📣

Our new SDK requires the security attribute to be set on the API endpoints we want to be accessible. This issue wasn't spotted earlier because we never tested against endpoints that didn't have the `security` attribute set. This is happening because our new SDK's are based on our OpenAPI spec, and without the security / bearerAuth field set, it will fail to authenticate. 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -1,3 +1,12 @@
+export const DEFAULT_REQUEST_SCHEMA = {
+  // Add more default attributes here if needed
+  security: [
+    {
+      bearerAuth: []
+    }
+  ]
+};
+
 export const GROUPS = {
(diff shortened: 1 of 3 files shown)


backend/src/lib/api-docs/constants.ts
```diff
@@ -1,3 +1,12 @@
+export const DEFAULT_REQUEST_SCHEMA = {
```
maidul98, line 1: Instead of this, i think it is better to follow this https://fastify.dev/docs/latest/Reference/Validation-and-Serialization/#adding-a-shared-schema

PR #2360, opened 2024-08-30
feat(integrations): Add visibility support to Github Integration

DanielHougaard:
# Description 📣

Previously it wasn't possible to select the visibility type for the Github integration. It was hardcoded to `all`, which is actually very misleading. `all` means it will only expose secrets to all _public_ repositories. This means that there was no way to expose organization-scoped secrets to private repositories.

This PR introduces visibility support, so the user can choose between:
- All public repositories
- All private repositories
- Select their own repositores -- this gives more fine-grained control over which repositories will get access to the organization-level secrets.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

###


akhilmhdh (commented):
Code wise looks good. Need to test it out

akhilmhdh (approved):

PR #2368, opened 2024-09-03
Audit log queue looping

akhilmhdh:
# Description 📣

This PR fixes a bug in resource clean up that could cause a looping in audit log and token cleanup. Also adds log point to identity present status of pruning. 

Root cause of multiple clean up runs:
If there are no issues then `deletedClientSecret.length > 0`  is false but `numberOfRetryOnFailure < MAX_RETRY_ON_FAILURE` will always be true because numberOfRetryOnFailure will be zero which is always less than max. To fix this, we need to check for `isretrying`

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/resource-cleanup/resource-cleanup-queue.ts
+++ b/backend/src/services/resource-cleanup/resource-cleanup-queue.ts
@@ -37,13 +37,21 @@ export const dailyResourceCleanUpQueueServiceFactory = ({
 }: TDailyResourceCleanUpQueueServiceFactoryDep) => {
   queueService.start(QueueName.DailyResourceCleanUp, async () => {
     logger.info(`${QueueName.DailyResourceCleanUp}: queue task started`);
+    logger.info(`${QueueName.DailyResourceCleanUp}: audit log`);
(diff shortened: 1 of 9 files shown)


backend/src/services/resource-cleanup/resource-cleanup-queue.ts
```diff
@@ -37,13 +37,21 @@ export const dailyResourceCleanUpQueueServiceFactory = ({
 }: TDailyResourceCleanUpQueueServiceFactoryDep) => {
   queueService.start(QueueName.DailyResourceCleanUp, async () => {
     logger.info(`${QueueName.DailyResourceCleanUp}: queue task started`);
+    logger.info(`${QueueName.DailyResourceCleanUp}: audit log`);
     await auditLogDAL.pruneAuditLog();
+    logger.info(`${QueueName.DailyResourceCleanUp}: remove expired access token`);
```
maidul98, line 42: I think we should put these inside of the actual methods?

PR #2373, opened 2024-09-04
feat(dynamic-secrets): elastic search support

DanielHougaard:
# Description 📣

This PR introduces a new dynamic secret type, Elastic Search.

Elastic Search is a distributed, open-source search and analytics engine designed to handle large volumes of data quickly and in near real-time. The dynamic secret works by creating a new user within Elastic Search. The user can be used to query and perform operations on the Elastic Search cluster.

This PR also includes some minor fixes for AWS ElastiCache dynamic secret documentation.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


backend/src/ee/services/dynamic-secret/providers/elastic-search.ts
```diff
@@ -0,0 +1,176 @@
+import { Client as ElasticCacheClient } from "@elastic/elasticsearch";
```
akhilmhdh, line 1: ElasticSearch right?

backend/src/ee/services/dynamic-secret/providers/elastic-search.ts
```diff
@@ -0,0 +1,176 @@
...
+
+const parseJsonStatement = (str: string) => {
+  try {
+    return JSON.parse(str) as object;
+  } catch {
+    throw new BadRequestError({ message: "Failed to parse ElasticSearch statements" });
+  }
+};
+
+const CreateElasticSearchUserSchema = z
```
akhilmhdh, line 29: Let's avoid this. The creation, revocation string statement is majorly used for db query as the statement would be complex to provide as option. 

But for JSON structured API lets keep it as in API level itself like the atlas provider. Because the issue is we want better ux for end user. Putting a JSON like that would confuse the user to understand how it's working. 

Let's move it inside the original the schema and use that.

akhilmhdh: This would also allow us to be me more controlled

backend/src/ee/services/dynamic-secret/providers/models.ts
```diff
@@ -30,6 +30,29 @@ export const DynamicSecretAwsElastiCacheSchema = z.object({
...
 });
 
+export const DynamicSecretElasticSearchSchema = z.object({
+  host: z.string().trim().min(1),
+  port: z.number(),
+
+  // two auth types "user, apikey"
+  auth: z.discriminatedUnion("type", [
+    z.object({
+      type: z.literal("user"),
```
akhilmhdh, line 40: Please use Enum

akhilmhdh (changes requested):

akhilmhdh (approved):

PR #2377, opened 2024-09-05
fix(integrations/circle-ci): Refactored Circle CI integration

DanielHougaard:
The Circle CI integration seemingly never worked in the first place _(or broke because Circle CI pushed a breaking change)_ due to in-proper project slugs. This PR resolves it by refactoring the fetching of apps, and the secret syncing step.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/integration-auth/integration-sync-secret.ts
+++ b/backend/src/services/integration-auth/integration-sync-secret.ts
@@ -35,7 +35,7 @@ import { TCreateManySecretsRawFn, TUpdateManySecretsRawFn } from "@app/services/
 
 import { TIntegrationDALFactory } from "../integration/integration-dal";
 import { IntegrationMetadataSchema } from "../integration/integration-schema";
-import { TIntegrationsWithEnvironment } from "./integration-auth-types";
+import { CircleCiVcsType, TIntegrationsWithEnvironment } from "./integration-auth-types";
 import {
   IntegrationInitialSyncBehavior,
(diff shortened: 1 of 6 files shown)


backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -1938,9 +1938,9 @@ const syncSecretsCircleCI = async ({
     })
   ).data;
 
-  let orgSlug: string | null = null;
+  let projectSlug: string | null = null;
   if (!integration.owner) {
-    orgSlug = `${circleciOrganizationDetail[0].slug}/${integration.app}`;
+    projectSlug = `${circleciOrganizationDetail[0].slug}/${integration.app}`;
```
maidul98, line 1943: It looks like you are still choosing the first org name from the list of orgs that get fetched via the collaborations API. I don't think this will work because the slug is in the fortmat of `vcs-slug/org-name/repo-name`. We need to to know which org to scope it to. Here is a example of what the collaborations API retruns this. It already has has the vcs-slug, the org id + name all it is missing is the project name which we have. 

I think it will be better if we get the org of the project the user selected so that we can create this project slug. What do you think?


CleanShot 2024-09-05 at 08 41 36@2x

maidul98 (approved):
lgtm

PR #2384, opened 2024-09-06
Manager users without waiting for confirmation of mail

akhilmhdh:
# Description 📣

This PR addresses a painpoint in which organizations has to wait for user confirmation before getting started. This will allow organization to invite users and then start assigning them to projects already. Helps smoother the automation too.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/auth/auth-signup-service.ts
+++ b/backend/src/services/auth/auth-signup-service.ts
@@ -24,6 +24,7 @@ import { TProjectMembershipDALFactory } from "../project-membership/project-memb
 import { TProjectUserMembershipRoleDALFactory } from "../project-membership/project-user-membership-role-dal";
 import { SmtpTemplates, TSmtpService } from "../smtp/smtp-service";
 import { TUserDALFactory } from "../user/user-dal";
+import { UserEncryption } from "../user/user-types";
 import { TAuthDALFactory } from "./auth-dal";
 import { validateProviderAuthToken, validateSignUpAuthorization } from "./auth-fns";
 import { TCompleteAccountInviteDTO, TCompleteAccountSignupDTO } from "./auth-signup-type";
@@ -174,7 +175,7 @@ export const authSignupServiceFactory = ({
       encryptedPrivateKey,
       iv: encryptedPrivateKeyIV,
       tag: encryptedPrivateKeyTag,
-      encryptionVersion: 2
+      encryptionVersion: UserEncryption.V2
     });
     const { tag, encoding, ciphertext, iv } = infisicalSymmetricEncypt(privateKey);
     const updateduser = await authDAL.transaction(async (tx) => {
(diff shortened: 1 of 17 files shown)


backend/src/services/auth/auth-signup-service.ts
```diff
@@ -183,26 +180,82 @@ export const authSignupServiceFactory = ({
...
+        });
+        const encKeys = await generateUserSrpKeys(email, password, {
+          publicKey: systemGeneratedUserEncryptionKey.publicKey,
+          privateKey: serverGeneratedPrivateKey
+        });
+        // now reencrypt server generated key with user provided password
+        userEncKey = await userDAL.upsertUserEncryptionKey(
+          us.id,
+          {
+            encryptionVersion: 2,
```
DanielHougaard, line 217: How about we make the encryption version an enum? Like we did with project versions?

backend/src/services/auth/auth-signup-service.ts
```diff
@@ -183,26 +180,82 @@ export const authSignupServiceFactory = ({
...
+            serverEncryptedPrivateKeyIV: iv,
+            serverEncryptedPrivateKey: ciphertext
+          },
+          tx
+        );
+      } else {
+        userEncKey = await userDAL.upsertUserEncryptionKey(
+          us.id,
+          {
+            encryptionVersion: 2,
```
DanielHougaard, line 239: Same as above

backend/src/services/auth/auth-signup-service.ts
```diff
@@ -348,65 +400,76 @@ export const authSignupServiceFactory = ({
...
+        });
+        const encKeys = await generateUserSrpKeys(email, password, {
+          publicKey: systemGeneratedUserEncryptionKey.publicKey,
+          privateKey: serverGeneratedPrivateKey
+        });
+        // now reencrypt server generated key with user provided password
+        userEncKey = await userDAL.upsertUserEncryptionKey(
+          us.id,
+          {
+            encryptionVersion: 2,
```
DanielHougaard, line 433: Same as above

backend/src/services/auth/auth-signup-service.ts
```diff
@@ -348,65 +400,76 @@ export const authSignupServiceFactory = ({
...
+            serverEncryptedPrivateKeyIV: iv,
+            serverEncryptedPrivateKey: ciphertext
+          },
+          tx
+        );
+      } else {
+        userEncKey = await userDAL.upsertUserEncryptionKey(
+          us.id,
+          {
+            encryptionVersion: 2,
```
DanielHougaard, line 455: Same as above

DanielHougaard (changes requested):
This is super neat! Great idea to consolidate the invite logic into a single flow. Left a few smaller comments!

PR #2388, opened 2024-09-08
feat: user details page audit logs & groups visualization

DanielHougaard:
# Description 📣

This PR aims to add more clarity as to what permissions a user has across an organization, and to show organization-wide  audit logs for the user. This will help organization leaders gain insight into what their members are doing across all projects.

The implementation tries to use existing functionality, but extending it so it can support the new use-cases (an example of this is the audit log component / api handlers)

CleanShot 2024-09-08 at 20 33 31@2x

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


dangtony98:
@DanielHougaard It looks like when you flip to the User Page, this modal comes up because audit logs is a paid feature.

I think users should still be able to view and act on this page; perhaps on this page specifically it should show an empty table in this case saying that audit logs aren't available on this plan and that they should upgrade.

Screenshot 2024-09-09 at 8 16 03 PM

maidul98:
I think we should keep groups simple. We are in the users details, so it would make sense to only see which groups this user belongs to and remove the user from that group. I find seeing group-projects related info very confusing/out of place

CleanShot 2024-09-10 at 00 09 21@2x
CleanShot 2024-09-10 at 00 08 38@2x

frontend/src/views/Org/UserPage/components/UserProjectsSection/UserGroupsRow.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const UserGroupsRow = ({ group, handlePopUpOpen }: Props) => {
+  return (
+    <>
+      <Tr
+        className="group h-10 cursor-pointer transition-colors duration-100 hover:bg-mineshaft-700"
+        key={`user-project-membership-${group.id}`}
+      >
+        <Td>{group.name}</Td>
+        <Td>
+          {true && (
```
dangtony98, line 23: Why do we need `true && (`? Can't we just omit this?

frontend/src/hooks/api/auditLogs/queries.tsx
```diff
@@ -5,27 +5,29 @@ import { apiRequest } from "@app/config/request";
...
-        {
-          params: {
-            ...filters,
-            offset: pageParam,
-            startDate: filters?.startDate?.toISOString(),
-            endDate: filters?.endDate?.toISOString()
-          }
+      const auditLogEndpoint = workspaceId
+        ? `/api/v1/workspace/${workspaceId}/audit-logs`
+        : "/api/v1/organization/audit-logs";
```
dangtony98, line 22: Note: In the future as we move audit logs out of projects into the organization level, I would deprecate `/v1/workspace/:workspaceId/audit-logs` and add `projectId` as a query param on the corresponding organization level endpoint for audit logs.

No action needed; just comment.

backend/src/services/group-project/group-project-dal.ts
```diff
@@ -95,6 +95,116 @@ export const groupProjectDALFactory = (db: TDbClient) => {
...
+          db.ref("name").withSchema(TableName.ProjectRoles).as("customRoleName"),
+          db.ref("slug").withSchema(TableName.ProjectRoles).as("customRoleSlug"),
+          db.ref("temporaryMode").withSchema(TableName.GroupProjectMembershipRole),
+          db.ref("isTemporary").withSchema(TableName.GroupProjectMembershipRole),
+          db.ref("temporaryRange").withSchema(TableName.GroupProjectMembershipRole),
+          db.ref("temporaryAccessStartTime").withSchema(TableName.GroupProjectMembershipRole),
+          db.ref("temporaryAccessEndTime").withSchema(TableName.GroupProjectMembershipRole)
+        );
+
+      const groupsWithProjects = sqlNestRelationships({
```
dangtony98, line 146: All of this is no longer needed since we decided not to display project information associated with groups on the frontend.

Let's delete and simply return groups that user is part of.

It's not needed in general to return projects alongside groups for particular user. I think the API experience should be simple:

- You query for the groups of a user; that's all the endpoint should return.
- If you need to know what projects a group is part of, then you make a subsequent query for that.

This keeps the backend code more maintainable and reduce sprawl since I'm sure we will be introducing more data structures down the line.

backend/src/server/routes/v1/user-router.ts
```diff
@@ -134,4 +134,53 @@ export const registerUserRouter = async (server: FastifyZodProvider) => {
...
+        username: z.string().trim()
+      }),
+      response: {
+        200: z
+          .object({
+            id: z.string(),
+            name: z.string(),
+            slug: z.string(),
+            orgId: z.string(),
+            projectMemberships: z.array(
```
dangtony98, line 155: No longer needed to return all these additional projects for group; can delete.

dangtony98 (changes requested):

backend/src/ee/services/audit-log/audit-log-service.ts
```diff
@@ -22,7 +22,7 @@ export const auditLogServiceFactory = ({
   auditLogQueue,
   permissionService
 }: TAuditLogServiceFactoryDep) => {
-  const listProjectAuditLogs = async ({
+  const listAuditLogs = async ({
```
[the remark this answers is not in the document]
DanielHougaard: For now I've added a permission check on the `Member` organization subject, with the read action.

I feel like it doesn't make sense to create a new permission subject for Audit Logs yet. We can perhaps do that in the future, if we decide to make audit logs more standalone. Let me know what you think

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -6,14 +6,14 @@ import { PkiItemType } from "@app/services/pki-collection/pki-collection-types";
 
 export type TListProjectAuditLogDTO = {
   auditLogActor?: string;
-  projectId: string;
+  projectId: string | null;
```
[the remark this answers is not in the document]
DanielHougaard: Yeah and honestly, marking them as optional undefined parameters is a wrong design decision from the start, which we use in a lot of places in our codebase. Ideally when something is intended to be a nullish, you set it to be either null OR the value you want.

Anyways I've updated the code to match our existing type patterns.

backend/src/services/user/user-service.ts
```diff
@@ -295,6 +298,16 @@ export const userServiceFactory = ({
     return updatedOrgMembership.projectFavorites;
   };
 
+  const listUserGroups = async (username: string, orgId: string) => {
```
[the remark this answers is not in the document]
DanielHougaard: Not entirely correct. Our auth middleware prevents this. Users wouldn't be able to fetch group memberships from users **outside** their organization, as your comment seems to suggest. I've added a Member read permission check, so No Access org members can't use the endpoint

frontend/src/views/Org/UserPage/components/UserProjectsSection/UserGroupsRow.tsx
DanielHougaard: Oh wow good catch! That was a left-over from testing

dangtony98 (approved):

PR #2392, opened 2024-09-09
misc: throw SAML or SSO errors properly

sheensantoscapadngan:
# Description 📣
This PR ensures that errors thrown during SAML/SSO logins are displayed to the user to allow them to debug the setup issues they encounter.

Before this PR, when errors are thrown during the SAML login process, it redirects the user to the signup page with an undefined token:
image

This PR updates the flow so that users actually see the error like the ff:
image

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98 (approved):
good idea

PR #2396, opened 2024-09-09
fix: stale session after logging into CLI

DanielHougaard:
# Description 📣

This PR fixes a rare bug with sessions going stale after logging into the CLI. The issue is multi-part, but what was happening was:

1. After logging into the CLI, a new session is created based on user ID, user agent, and IP address. When logging in with the CLI, the user agent should always be `cli`. However, the user agent is always set to the browser's user agent, because we can't set the user agent on requests sent with the browser (unsafe header limitation).
   * The fix for this was adding a new "customUserAgent" parameter to the `/select-organization` endpoint. This allows us to specify user agents for sessions that deviate from the browsers user agent.
2. We were incrementing the session version whenever a user selected an organization.
   *  We are only supposed to increment the version when we want tokens to become invalid (such as after changing passwords or logging out. Bumping the session version is not relevant for switching organization.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/server/routes/v3/login-router.ts
+++ b/backend/src/server/routes/v3/login-router.ts
@@ -42,7 +42,8 @@ export const registerLoginRouter = async (server: FastifyZodProvider) => {
     },
     schema: {
       body: z.object({
-        organizationId: z.string().trim()
+        organizationId: z.string().trim(),
+        customUserAgent: z.enum(["cli"]).optional()
       }),
       response: {
         200: z.object({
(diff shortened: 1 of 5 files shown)


frontend/src/pages/login/select-organization.tsx
```diff
@@ -68,7 +68,10 @@ export default function LoginPage() {
         return;
       }
 
-      const { token } = await selectOrg.mutateAsync({ organizationId: organization.id });
+      const { token } = await selectOrg.mutateAsync({
+        organizationId: organization.id,
+        customUserAgent: callbackPort ? "cli" : undefined
```
maidul98, line 73: nit: `cli` should be a constant

backend/src/server/routes/v3/login-router.ts
```diff
@@ -42,7 +42,8 @@ export const registerLoginRouter = async (server: FastifyZodProvider) => {
     },
     schema: {
       body: z.object({
-        organizationId: z.string().trim()
+        organizationId: z.string().trim(),
+        customUserAgent: z.enum(["cli"]).optional()
```
maidul98, line 46: nit: might be better called just `userAgent`. If they are passing this in then we can assume it is custom

maidul98 (approved):
LGTM

PR #2403, opened 2024-09-10
Audit log batching

akhilmhdh:
# Description 📣

This PR is for performance improvement with Infisical. 

1. A batching strategy is introduced for reducing our database load in audit log generation. The system will first insert data to redis list and then after each 5 seconds or when the number of entries crosses a threshold its saved to database. The present cycle is 5 seconds. 
2. Removed FK from audit log with project id and org id. This removes the complete dependency of audit log with other tables. Thus given the shear size of audit log it becomes an append only log. Audit log clean up will handle removing orphan rows. This will also boost performance in other query and also no more locks will cause other operation to be in stuck.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/ee/services/audit-log/audit-log-queue.ts
```diff
@@ -42,88 +51,144 @@ export const auditLogQueueServiceFactory = ({
...
-      eventType: event.type,
-      expiresAt: new Date(Date.now() + ttl),
-      eventMetadata: event.metadata,
-      userAgentType
+  const startPeriodicAuditLogBatchWriteToDB = () =>
+    queueService.queue(QueueName.AuditLog, QueueJobs.AuditLogWriteToDB, undefined, {
+      removeOnFail: true,
+      removeOnComplete: true,
+      repeat: {
+        every: AUDIT_LOG_BATCH_WRITE_TO_DB_CYCLE_SPEED,
```
maidul98, line 181: So this will check for logs over filling every 5 seconds but there could be instances where the the RPS could be 1k per second. This would mean that by the time we try to do a check, there will already be 5k items or more under the redis key. 

Instead of doing a loop that keeps on running every 5 seconds (also could be wasted loops), we should trigger the batch write job after determining that we have met a threshold. What do you think?

maidul98 (commented):
Left comments and concerns

backend/src/ee/services/audit-log/audit-log-queue.ts
```diff
@@ -42,88 +51,144 @@ export const auditLogQueueServiceFactory = ({
...
+      );
     }
 
-    const plan = await licenseService.getPlan(orgId);
-    if (plan.auditLogsRetentionDays === 0) {
-      // skip inserting if audit log retention is 0 meaning its not supported
-      return;
+    // this is when audit log needs to write to db
+    if (job.name === QueueJobs.AuditLogWriteToDB) {
+      const lock = await keyStore.acquireLock([KeyStorePrefixes.AuditLogBatchLock], 5000, {
```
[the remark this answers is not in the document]
akhilmhdh: It will retry with jitter for 10 times, and fails it would do it next cycle

backend/src/ee/services/audit-log/audit-log-queue.ts
akhilmhdh: In those scenerio this will kick in 

https://github.com/Infisical/infisical/pull/2403/files/c7baec86841b17d7f1dba11a36e8b1a06c88e18c#diff-c2c2f65a2bc66e55f4b6f1623a270941d71e1d0f8fc58664d806c6010b3d27a5R163

That is there to ensure its below a threshold. 

Regarding waster cycles because it needed. We don't do batching until we reach a threshold. But what if the threshold is not met anytime soon. There could be documents that needs to inserted thus the periodic cycle kicks in.

backend/src/ee/services/audit-log/audit-log-queue.ts
```diff
@@ -42,88 +51,144 @@ export const auditLogQueueServiceFactory = ({
...
+        orgId,
+        eventType: event.type,
+        expiresAt: new Date(Date.now() + ttl),
+        eventMetadata: event.metadata,
+        userAgentType,
+        createdAt: new Date(job.timestamp),
+        updatedAt: new Date(job.timestamp)
+      };
+
+      await keyStore.listAppend(KeyStorePrefixes.AuditLogBatch, [JSON.stringify(auditLog)]);
```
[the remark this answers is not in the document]
akhilmhdh: Issue is complexity would scale up without any benefits. Consider we have multiple list for each org or project. Now we need to keep a check for all these list so that they don't go out of bounds. 

Next is irrespective of whoever is generating it, it will cause write in db. So it doesn't matter to split it up.

backend/src/ee/services/audit-log/audit-log-queue.ts
maidul98: I see, so at 500 logs, it will create a job to flush. Got it

akhilmhdh: Correction: 5K actually. I took average size of our audit log row and it was around 890 bytes.

PR #2418, opened 2024-09-12
Don't enforce max user and identity limits

meetcshah19:
# Description 📣
Stop enforcing user and machine Identity limits for Enterprise customers. Context : [ISSUE](https://linear.app/infisical/issue/ENG-1500/allow-feature-set-to-be-set-to-unlimited-number-of-uses)

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

### DIFF
--- a/backend/src/services/identity/identity-service.ts
+++ b/backend/src/services/identity/identity-service.ts
@@ -58,7 +58,8 @@ export const identityServiceFactory = ({
     if (!hasRequiredPriviledges) throw new BadRequestError({ message: "Failed to create a more privileged identity" });
 
     const plan = await licenseService.getPlan(orgId);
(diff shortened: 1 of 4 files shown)


maidul98:
Please make sure each of your PR descriptions provide enough context for reviewers (even if you told them privately)

dangtony98 (changes requested):
This PR falls short on two items:

1. These changes would lift member and identity limit restrictions for all users regardless of plan which should not be the case; the lift should only be applied to Enterprise plans.
2. The PR lifts limits on the backend but doesn't do it on the frontend; this means that users would still be blocked by the pop-up upgrade modal on the client-side before even reaching the server.

dangtony98 (approved):

PR #2426, opened 2024-09-13
Fix: Restricted Secret Environment UI Corrections

### LINKED ISSUE #2371: Missing secrets warning show to users who do not have access to the secrets of an enviornment
### Describe the bug
Users without access to the secrets of an environment appear as if they aren't set on the secrets overview page

### To Reproduce
1. Deny access to the secrets of an environment for a role/user
image
2. Go to the secrets overview page

### Expected behavior
Either the environment is hidden, or the missing secrets indicator is not shown, and the explore button is disabled.

### Screenshots
image
image

Clicking explore:
image

scott-ray-wilson:
# Description 📣

This PR hides environments the user does not have read access to on the secrets overview page. Also includes increased default pagination size for secrets and identities from 10 to 50 and temporary fix to prevent expanded secrets component width being set to 0.

image
(Example user without production env read permissions)

resolves #2371

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
+++ b/frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
@@ -85,7 +85,7 @@ enum RowType {
   Secret = "Secret"
 }
 
-const INIT_PER_PAGE = 10;
+const INIT_PER_PAGE = 50;
 
 export const SecretOverviewPage = () => {
   const { t } = useTranslation();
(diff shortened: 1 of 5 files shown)


frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
```diff
@@ -173,11 +173,31 @@ export const SecretOverviewPage = () => {
...
+        })
+      )
+    )
+  );
 
   useEffect(() => {
-    setVisibleEnvs(userAvailableEnvs);
-  }, [userAvailableEnvs]);
+    setVisibleEnvs(
+      userAvailableEnvs?.filter(({ slug }) =>
```
maidul98, line 190: Can we also hide it here?

CleanShot 2024-09-14 at 23 21 12@2x

maidul98 (commented):
Looks good to me, just left one comment

frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
scott-ray-wilson: Nice catch!
image

I noticed I missed it here as well, fixed:
image

maidul98 (approved):

PR #2429, opened 2024-09-15
feat: integration details page with logging

DanielHougaard:
# Description 📣

This PR's main objective is to introduce a new integrations details page, which we're doing to bring visibility to all the integration settings. Additionally this PR also introduces integration logs, which will log integration syncs. The logs are based on our audit log system and will therefore only work if the user is on a paid plan with audit logs enabled. This is a step towards expanding our paid product line.

The PR introduces a new actor called the "Platform" actor. The platform actor is useful for when we want to perform logging on automated actions that doesn't require any user intervention (such as integration syncs!). 

I've also refactored the audit logs quite a bit. You'll notice we have a new endpoint on the organization router for audit logs. This is done so we can start performing logs on an organization-level in the future. We still have the old audit log endpoint for backwards compatibility purposes, since the audit log endpoint is public.

We are still storing the integration details on the integration itself (such as lastUsed, isSynced, syncMessage). The reason for this is that we can't guarantee that the user has access to audit logs, so we can't always just compute those fields from the audit logs.

CleanShot 2024-09-15 at 20 08 35

CleanShot 2024-09-15 at 20 20 38

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/routes/v1/project-router.ts
+++ b/backend/src/ee/routes/v1/project-router.ts
@@ -146,12 +146,16 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
         actorId: req.permission.id,
         actorOrgId: req.permission.orgId,
         actorAuthMethod: req.permission.authMethod,
-        projectId: req.params.workspaceId,
-        ...req.query,
-        endDate: req.query.endDate,
-        startDate: req.query.startDate || getLastMidnightDateISO(),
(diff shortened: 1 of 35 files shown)


akhilmhdh (approved):
Perfecto

PR #2445, opened 2024-09-17
feat(cli/api): more descriptive api errors & CLI warning when using token auth while being logged in

DanielHougaard:
# Description 📣

#### API:
This PR adds more descriptive error messages for certain errors. Specifically we are now formatting JWT-related errors, so they're more understandable for the end users.

#### CLI:
We are now checking if the user is using user-based auth, and token based auth at the same time. If the user is using both token-based and user-based authentication, we will display a warning letting the user know that the token-based auth is being used.

This has been confusing users in the past, so this aims to add more clarity which authentication method is being used.

After talking to @maidul98, this seemed like something we'd want to keep on the API-end, instead handling directly in the CLI codebase.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/server/plugins/error-handler.ts
+++ b/backend/src/server/plugins/error-handler.ts
@@ -1,7 +1,9 @@
 import { ForbiddenError } from "@casl/ability";
 import fastifyPlugin from "fastify-plugin";
+import { JsonWebTokenError } from "jsonwebtoken";
 import { ZodError } from "zod";
 
+import { UserAgentType } from "@app/ee/services/audit-log/audit-log-types";
 import {
   BadRequestError,
   DatabaseError,
@@ -10,6 +12,12 @@ import {
   ScimRequestError,
   UnauthorizedError
 } from "@app/lib/errors";
+
+enum JWTErrors {
+  JwtExpired = "jwt expired",
+  JwtMalformed = "jwt malformed",
+  InvalidAlgorithm = "invalid algorithm"
+}
 
 export const fastifyErrHandler = fastifyPlugin(async (server: FastifyZodProvider) => {
   server.setErrorHandler((error, req, res) => {
@@ -35,6 +43,35 @@ export const fastifyErrHandler = fastifyPlugin(async (server: FastifyZodProvider
         schemas: error.schemas,
         status: error.status,
         detail: error.detail
+      });
+      // Handle JWT errors and make them more human-readable for the end-user.
+    } else if (error instanceof JsonWebTokenError) {
+      const isCliRequest = req.headers["user-agent"] === UserAgentType.CLI;
+
+      const message = (() => {
+        if (error.message === JWTErrors.JwtExpired) {
+          return "Your token has expired. Please re-authenticate.";
+        }
+        if (error.message === JWTErrors.JwtMalformed) {
+          if (isCliRequest) {
+            return "The access token is malformed. Are you sure the token you are using is correct? Check the INFISICAL_TOKEN environment variable, or the --token flag. If you are using user-login, please run [infisical login] to re-authenticate.";
+          }
+          return "The access token is malformed. Please ensure that the token is in the correct format and try again.";
+        }
+        if (error.message === JWTErrors.InvalidAlgorithm) {
+          if (isCliRequest) {
+            return "Invalid algorithm. Are you sure you are using the correct authentication method? Make sure to check that you don't have the INFISICAL_TOKEN variable set in your environment variables. If you are intentionally using the INFISICAL_TOKEN variable, make sure you are using the correct token." as const;
+          }
+          return "The access token is signed with an invalid algorithm. Please ensure that the token is in the correct format and try again. We recommend obtaining a new token.";
+        }
+
+        return error.message;
+      })();
+
+      void res.status(401).send({
+        statusCode: 401,
+        error: "TokenError",
+        message
       });
     } else {
       void res.send(error);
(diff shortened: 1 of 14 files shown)


backend/src/server/plugins/error-handler.ts
```diff
@@ -36,6 +44,36 @@ export const fastifyErrHandler = fastifyPlugin(async (server: FastifyZodProvider
...
+          if (isCliRequest) {
+            return "The access token is malformed. Are you sure the token you are using is correct? Check the INFISICAL_TOKEN environment variable, or the --token flag. If you are using user-login, please run [infisical login] to re-authenticate. Please ensure that the INFISICAL_TOKEN variable is not set, if you are not intentionally using it.";
+          }
+          return "The access token is malformed. Please ensure that the token is in the correct format and try again.";
+        }
+        if (error.message === JWTErrors.InvalidAlgorithm) {
+          if (isCliRequest) {
+            return "Invalid algorithm. Are you sure you are using the correct authentication method? Make sure to check that you don't have the INFISICAL_TOKEN variable set in your environment variables. If you are intentionally using the INFISICAL_TOKEN variable, make sure you are using the correct token.";
+          }
+          return "The access token is signed with an invalid algorithm. Please ensure that the token is in the correct format and try again. We recommend obtaining a new token.";
```
maidul98, line 66:
```
The access token is signed with an invalid algorithm. Please provide a valid token and try again.
```

backend/src/server/plugins/error-handler.ts
```diff
@@ -36,6 +44,36 @@ export const fastifyErrHandler = fastifyPlugin(async (server: FastifyZodProvider
...
+        }
+        if (error.message === JWTErrors.JwtMalformed) {
+          if (isCliRequest) {
+            return "The access token is malformed. Are you sure the token you are using is correct? Check the INFISICAL_TOKEN environment variable, or the --token flag. If you are using user-login, please run [infisical login] to re-authenticate. Please ensure that the INFISICAL_TOKEN variable is not set, if you are not intentionally using it.";
+          }
+          return "The access token is malformed. Please ensure that the token is in the correct format and try again.";
+        }
+        if (error.message === JWTErrors.InvalidAlgorithm) {
+          if (isCliRequest) {
+            return "Invalid algorithm. Are you sure you are using the correct authentication method? Make sure to check that you don't have the INFISICAL_TOKEN variable set in your environment variables. If you are intentionally using the INFISICAL_TOKEN variable, make sure you are using the correct token.";
```
maidul98, line 64: Instead of this, we can add a warning in the CLI when the access token is being passed from `INFISICAL_TOKEN`. This way the user knows that their current session is being over written

maidul98 (changes requested):
Let's remove the errors specific to CLI and keeps the generic ones

maidul98 (changes requested):
Backend changes look good but for CLI, a better approach is to make use of `PersistentPreRun` function in the root CMD. You can call a function there and it will get run by every child command. 

What we want to check for is if there is a token over ride either from --token flag or from environment. If we see them being set we can show the message. 

CleanShot 2024-09-17 at 21 44 05@2x
CleanShot 2024-09-17 at 21 44 55@2x

cli/packages/models/cli.go
```diff
@@ -63,8 +63,9 @@ type DynamicSecretLease struct {
 }
 
 type TokenDetails struct {
-	Type  string
-	Token string
+	Type         string
+	Token        string
+	PassedAsFlag bool
```
maidul98, line 68: This might be better called source

cli/packages/cmd/root.go
```diff
@@ -43,14 +44,33 @@ func init() {
...
+
+		if !silent && err == nil && loggedInDetails.IsUserLoggedIn && !loggedInDetails.LoginExpired {
+			token, err := util.GetInfisicalToken(cmd)
+
+			if err == nil && token != nil {
+				var usingFrom string
+				if token.PassedAsFlag {
+					usingFrom = "--token flag"
+				} else {
+					usingFrom = "INFISICAL_TOKEN environment variable"
```
maidul98, line 67: This says `INFISICAL_TOKEN` but in getInfisicalToken we also check for `INFISICAL_UNIVERSAL_AUTH_ACCESS_TOKEN_NAME`. It's better you use replace `PassedAsFlag` to something like `source`

cli/packages/cmd/run.go
```diff
@@ -439,9 +439,9 @@ func executeCommandWithWatchMode(commandFlag string, args []string, watchModeInt
 
 func fetchAndFormatSecretsForShell(request models.GetAllSecretsParameters, projectConfigDir string, secretOverriding bool, shouldExpandSecrets bool, token *models.TokenDetails) (models.InjectableEnvironmentResult, error) {
 
-	if token != nil && token.Type == util.SERVICE_TOKEN_IDENTIFIER {
+	if util.ShouldUseInfisicalToken(token, []string{util.SERVICE_TOKEN_IDENTIFIER}) {
```
maidul98, line 442: Also can you please remove the helper function logic? The previous logic read much clear and this seems un related to this PR. We should only make code changes that we 100% require

maidul98 (commented):
Made minor adjustment to the error warning message and removed the comments. Let’s only add comments for very nuanced logic

maidul98 (approved):
`source` would be better set as a enum of sorts because now the get token function is tightly coupled with the warning logic but this is fine for now

PR #2448, opened 2024-09-18
feat(audit-logs): moved audit logs to organization-level

DanielHougaard:
# Description 📣

This PR moves the Audit Logs to the organization-level, instead of having audit logs on each individual project. I've added a new filter, so users can filter by project still. The way the permissions are configured, will allow organizations **admins and members** to read audit logs, which falls inline with what we had for projects.

I've also updated the endpoint in our API reference to reflect the new audit logs endpoint. The new audit log endpoint comes with more advanced filtering and is more extensible. 

Additionally, I also improved the UI of the audit logs slightly (specifically the filter options).

#### Note
The old audit logs endpoint needs to be fully deprecated eventually, but currently it's exposed in our public API.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation

---


dangtony98:
Screenshot 2024-09-20 at 8 27 05 AM

Upon fresh installation of Infisical, there's a `0` that appears on the organization-level audit log page (see screenshot above).

backend/src/server/routes/v1/organization-router.ts
dangtony98, line 1: Filtering in the UI by **Source** with CLI and K8s operator set had no effect (the resulting list of events continued to be those associated with Web).

frontend/src/views/Org/AuditLogsPage/components/LogsTableRow.tsx
```diff
@@ -476,7 +478,47 @@ export const LogsTableRow = ({ auditLog, isOrgAuditLogs, showActorColumn }: Prop
...
         );
+
+      case EventType.GET_WORKSPACE_KEY:
+        return (
+          <Td>
+            <p>{`Key ID: ${event.metadata.keyId}`}</p>
+          </Td>
+        );
+
+      case EventType.LOGIN_IDENTITY_UNIVERSAL_AUTH:
```
dangtony98, line 489: Note: It would be great to spend a day (I can do it if needed) to make sure that all audit log events are being shown in this table (including other logins across auth methods).

No action item is needed for this review comment.

frontend/src/views/Org/AuditLogsPage/components/LogsTable.tsx
```diff
@@ -41,12 +41,15 @@ export const LogsTable = ({
 }: Props) => {
   const { currentWorkspace } = useWorkspace();
 
+  const filterProjectId =
```
dangtony98, line 44: Nit: Confusing condition; I would break it up somehow so it's more readable.

Feel free to dismiss or update.

dangtony98 (changes requested):

dangtony98 (approved):

PR #2451, opened 2024-09-19
Feature: Server-side Pagination for Secrets Overview and Main Pages

scott-ray-wilson:
# Description 📣

This PR migrates secret pagination, filtering and sorting to the server for both the Secrets Overview and Secrets Main pages. Additional filters have been added to both these pages.

https://github.com/user-attachments/assets/32a05cfe-7f02-4f57-bf67-6be9ed345d47

Projects older than V3 will need to upgrade to continue to use the secrets dashboard. Older projects will still be able to access their secrets via the API.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


frontend/src/views/SecretOverviewPage/components/SecretV2MigrationSection/SecretV2MigrationSection.tsx
scott-ray-wilson, line 1: Adding a reminder here to not merge until this project upgrade process has been verified

backend/src/server/routes/v3/dashboard-router.ts
```diff
@@ -0,0 +1,610 @@
+import { z } from "zod";
```
sheensantoscapadngan, line 1: I think the total count breaks for replicated imports? 
image

As you can see I only have 3 import entries but the total count is showing as 5 (at the bottom left)

sheensantoscapadngan (commented):
nice work @scott-ray-wilson !! 

all else seems fine in my testing

backend/src/ee/services/dynamic-secret/dynamic-secret-service.ts
```diff
@@ -328,7 +431,61 @@ export const dynamicSecretServiceFactory = ({
...
+      projectId = project.id;
+    }
+
+    const { permission } = await permissionService.getProjectPermission(
+      actor,
+      actorId,
+      projectId,
+      actorAuthMethod,
+      actorOrgId
+    );
```
sheensantoscapadngan, line 471: it might be worth optimizing the dashboard route logic so that we fetch the permission object once for all resources? At the moment, I think we do at most 8 permission object fetches. These are roundtrip calls to the db. though the difference is probably negligible

scott-ray-wilson: spoke with @sheensantoscapadngan - going to wait on this

frontend/src/views/SecretOverviewPage/components/SecretV2MigrationSection/SecretV2MigrationSection.tsx
scott-ray-wilson: Sheen was able to complete the flow - had to manually refresh the page after completion so I added a query invalidation

backend/src/server/routes/v3/dashboard-router.ts
scott-ray-wilson: excluding replicas in count now

backend/src/ee/services/dynamic-secret/dynamic-secret-service.ts
```diff
@@ -300,19 +301,121 @@ export const dynamicSecretServiceFactory = ({
...
+    actor,
+    projectSlug,
+    path,
+    environmentSlug,
+    search,
+    ...params
+  }: TListDynamicSecretsDTO) => {
+    let { projectId } = params;
+
+    if (!projectId) {
```
maidul98, line 366: I see we are doing this patten a lot and it seems like we never really pass in the project slug, it's mostly the porject id. If we aren't making a call with projectslug, the we don't need it

backend/src/server/routes/v3/dashboard-router.ts
```diff
@@ -0,0 +1,610 @@
...
+    schema: {
+      description: "List project secrets overview",
+      security: [
+        {
+          bearerAuth: []
+        }
+      ],
+      querystring: z.object({
+        projectId: z.string().trim().describe(DASHBOARD.SECRET_OVERVIEW_LIST.projectId),
+        environments: z
```
maidul98, line 34: Why did we choose this to be a comma separated list and not a array of env names?

backend/src/server/routes/v3/dashboard-router.ts
```diff
@@ -0,0 +1,610 @@
...
+            .optional(),
+          totalFolderCount: z.number().optional(),
+          totalDynamicSecretCount: z.number().optional(),
+          totalSecretCount: z.number().optional(),
+          totalCount: z.number()
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT]),
+    handler: async (req) => {
```
maidul98, line 100: question: what about imports?

backend/src/services/secret-import/secret-import-dal.ts
```diff
@@ -70,6 +97,28 @@ export const secretImportDALFactory = (db: TDbClient) => {
...
   };
 
+  const getProjectImportCount = async (
+    { search, ...filter }: Partial<TSecretImports & { projectId: string; search?: string }>,
+    tx?: Knex
+  ) => {
+    try {
+      const docs = await (tx || db.replicaNode())(TableName.SecretImport)
+        .where(filter)
+        .where("isReplication", false)
```
maidul98, line 107: question: what about the imports that replicate? Those are imports too

maidul98 (commented):
Just some small comments and questions. Also we should center this message 
CleanShot 2024-09-20 at 12 44 03@2x

backend/src/ee/services/dynamic-secret/dynamic-secret-service.ts
scott-ray-wilson: I'll remove the slug option on new service functions I added but keep it were it was already present (I had to add projectId to a few that previously only used slugs, want to preserve any current usage)

backend/src/server/routes/v3/dashboard-router.ts
scott-ray-wilson: will refactor to arrays

backend/src/services/secret-import/secret-import-dal.ts
scott-ray-wilson: @sheensantoscapadngan noticed imports that replicate count twice for a single import in pagination, so when getting the count getting just the non-replicas provides the correct amount

backend/src/server/routes/v3/dashboard-router.ts
scott-ray-wilson: Imports is a separate query for the overview page (not used in pagination for overview, just displaying if a secret is imported or not)

maidul98: oh okay

backend/src/server/routes/v3/dashboard-router.ts
maidul98: No need for now

maidul98 (approved):

PR #2455, opened 2024-09-19
feat(integrations): email when integration sync fails

DanielHougaard:
# Description 📣

This PR aims to add support for email notifications, when an integration fails to sync for whatever reason. A key difference is that: If the integration is manually triggered, only the actor who triggered it will get the email. If it's an automated sync, all project members will receive the notification.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/secret/secret-queue.ts
+++ b/backend/src/services/secret/secret-queue.ts
@@ -562,6 +562,30 @@ export const secretQueueFactory = ({
       throw new Error("Secret path not found");
     }
 
+    const sendIntegrationSyncFailedMail = async (integrationId: string, syncMessage: string | null) => {
+      const appCfg = getConfig();
+
+      // If smtp is not configured, we can return early without having to fetch the project members.
+      if (!appCfg.isSmtpConfigured) return;
+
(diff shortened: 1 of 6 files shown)


backend/src/services/secret/secret-queue.ts
```diff
@@ -562,6 +562,30 @@ export const secretQueueFactory = ({
       throw new Error("Secret path not found");
     }
 
+    const sendIntegrationSyncFailedMail = async (integrationId: string, syncMessage: string | null) => {
+      const appCfg = getConfig();
+
+      // If smtp is not configured, we can return early without having to fetch the project members.
+      if (!appCfg.isSmtpConfigured) return;
+
+      const projectMembers = await projectMembershipDAL.findAllProjectMembers(projectId);
```
akhilmhdh, line 571: Should all the members get it or just admins?

Also i think we may need a  debounced check to ensure a lot of mails are not send when all the secret sync fails. Because from dashboard each change triggers sync integration.

Now consider integration failing and I did a quitie some change, I would be bombarded with mail

akhilmhdh (approved):

Issue #2446, opened 2024-09-17
New options for Allow user signups - invite only


GameBurrow:
### Feature description
Currently **Allow user signups** has 2 options:
- `Anyone`
- `Disabled`

We need third option - `Invite only`. Only allow registration for users that actually got invited.

### Why would it be useful?
Currently the only way to limit user registration is by domain only. That's useful for corporations/companies/businesses where all people use same emails, but for more casual use, where people cane have different email domains (for example I have a kube cluster that is shared by me and my close friends), I have to temporarely enable registrations, so I can invite those people, which I consider a security risk.  Especially considering you can't fully delete users through UI in Free tier, if some random person starts user registration process, while I have it open, it's annoying.

### Additional context
Strange thing is looking at old issue reports and forum posts, it seems like you had this feature using `INVITE_ONLY_SIGNUP` env flag but it got removed at one point? https://github.com/Infisical/infisical/issues/1892

kasyap1234:
@maidul98 I would like to work on this issue .

sheensantoscapadngan:
Hey @GameBurrow! When the **Allow user signups** option is Disabled, your users should still be able to register to the organization by clicking the invite link sent via email upon org invitation. At the moment, they will have to go through the invitation link for registration to work as expected. Let me know if you have any additional questions! 

<img width="497" alt="image" src="https://github.com/user-attachments/assets/39334fd3-0600-48cb-baec-b0b03528739b">

sheensantoscapadngan:
This is indeed a bug in the signup flow. I've raised a PR to address this [here](https://github.com/Infisical/infisical/pull/2466). 

Thank you for reporting this and for the prompt response!

vwbusguy:
> Hey @GameBurrow! When the **Allow user signups** option is Disabled, your users should still be able to register to the organization by clicking the invite link sent via email upon org invitation. At the moment, they will have to go through the invitation link for registration to work as expected. Let me know if you have any additional questions!
[...]

PR #2466, opened 2024-09-22
misc: addressed invalid redirect condition in signup invite page

sheensantoscapadngan:
# Description 📣
This PR resolves the following issue: https://github.com/Infisical/infisical/issues/2446

Basically it removes a condition in the signupinvite page that blocked invited users from creating an account when instance allow signup option is set to "Disabled"

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/pages/signupinvite.tsx
+++ b/frontend/src/pages/signupinvite.tsx
@@ -70,12 +70,6 @@ export default function SignupInvite() {
 
   const { mutateAsync: selectOrganization } = useSelectOrganization();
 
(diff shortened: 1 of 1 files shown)


vwbusguy:
Looks good to me.  It doesn't make sense that an invited user should get this far and not be able to sign up without global sign up being set.

DanielHougaard (approved):

PR #2477, opened 2024-09-24
feat: allow users to replace auth methods

meetcshah19:
# Description 📣
Allow users to replace / overwrite auth methods for identities.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/views/Org/MembersPage/components/OrgIdentityTab/components/IdentitySection/IdentityAuthMethodModal.tsx
+++ b/frontend/src/views/Org/MembersPage/components/OrgIdentityTab/components/IdentitySection/IdentityAuthMethodModal.tsx
@@ -54,7 +54,7 @@ const identityAuthMethods = [
 
 const schema = yup
   .object({
-    authMethod: yup.string().required("Auth method is required")
+    authMethod: yup.mixed<IdentityAuthMethod>().oneOf(Object.values(IdentityAuthMethod)).required("Auth method is required")
   })
   .required();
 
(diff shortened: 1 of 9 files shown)


dangtony98:
Note: Instead of passing `initialAuthMethod` prop into each Form component, you can also get the it through `useGetIdentityById` in the form to get the auth method since it is specified on the identity.

I do think that this PR may conflict with the initiative to add multiple auth methods per identity @DanielHougaard

DanielHougaard:
We're leaving this PR as-is for now because we have another initiative that will take precedence over this.

frontend/src/views/Org/MembersPage/components/OrgIdentityTab/components/IdentitySection/IdentityAuthMethodModal.tsx
scott-ray-wilson, line 1: When you open this modal, change the selected auth method and close it again, and re-open it, it doesn't reset back to the currently used auth method, I think this should reset on close

scott-ray-wilson (changes requested):

frontend/src/views/Org/MembersPage/components/OrgIdentityTab/components/IdentitySection/IdentityAuthMethodModal.tsx
scott-ray-wilson, line 1: Also seems like the general logic of this whole flow isn't working?

image

scott-ray-wilson (changes requested):

frontend/src/views/Org/MembersPage/components/OrgIdentityTab/components/IdentitySection/IdentityAuthMethodModal.tsx
scott-ray-wilson: @meetcshah19 this error was due to having the confirm overwrite modal inside the form, the button click on the form triggered the submit on the form as well

scott-ray-wilson (approved):

maidul98 (approved):
tested it out, works as expected scott and meet

PR #2482, opened 2024-09-25
feat(secret-sharing): server-side encryption

DanielHougaard:
# Description 📣

This PR moves the secret sharing encryption logic to the server-side, where we'll now handle encryption/decryption with our KMS. This results in a greatly reduced URL length, as the key is no longer necessary. I opted to keep a shortened version of the hashed hex as a part of the URL. We're keeping the first 13 characters as a part of the secret sharing URL to increase entropy. This PR imposes a 50% decrease in the secret sharing URL length.

Previously created secrets remain backwards compatible with this change. Meaning, if you created secrets before this new change, they will remain working.

Examples:

Old secret sharing URL
`http://localhost:8080/shared/secret/aea90f5e-439a-48d6-938b-a666ade4f796?key=e79d58893e44e656b56833bacc61e2e2c104d044c95870e07f7d59ffd8c8e163-9d4fc46ff176326c142176eafb04b1b2`

New secret sharing URL
`http://localhost:8080/shared/secret/649a6964-f973-43f2-a24b-4082c3f92703-6b86b273ff34f`

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---


backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -118,7 +117,7 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
         ...req.body,
         accessType: SecretSharingAccessType.Anyone
       });
-      return { id: sharedSecret.id };
+      return { id: sharedSecret.id, hashedHex: sharedSecret.hashedHex };
```
dangtony98, line 120: Note: In a future fuller revamp of secret sharing, I'd like to keep `id` and `hashedHex` in one instead of returning separate params to keep API simple; this would be achieved by removing direct ID reference to the DB secret sharing ID and instead generating a new one with higher entropy with more compressed text representation than `hex`.

No action needed for this comment for now.

dangtony98 (changes requested):

dangtony98 (approved):
LGTM now!

Issue #2400, opened 2024-09-10
InfisicalSecret does not update already generated Secret on field change


tiagovtristao:
### Describe the bug

The `InfisicalSecret` doesn't seem propagate any new changes when you change the `.spec.authentication.universalAuth.secretsScope.envSlug` field (this could be true for other fields as well). I had to delete the generated secret and the pod using it, so that the new secrets (associated with the new `envSlug`) would be refetched and propagated onto the pod.

Partial resource below:

```yaml
apiVersion: secrets.infisical.com/v1alpha1
kind: InfisicalSecret
metadata:
  name: foobar
spec:
  authentication:
    universalAuth:
      credentialsRef:
        secretName: infisical-universal-auth-credentials
        secretNamespace: default
      secretsScope:
        envSlug: production-k8s
        projectSlug: foobar-8-aaa
        recursive: false
        secretsPath: /website
```

### To Reproduce

Edit the `.spec.authentication.universalAuth.secretsScope.envSlug` field to point to a different one.

### Expected behavior
The secret generated by `InfisicalSecret` should be updated with the new variables, and the deployment reloaded if `secrets.infisical.com/auto-reload` is set.

akhilmhdh:
CC: @DanielHougaard

DanielHougaard:
Hi @tiagovtristao, thank you for the report. I've made an internal item for this and we'll get to it as soon as possible.

tiagovtristao:
@akhilmhdh and @DanielHougaard, thanks for getting into it so quickly! Much appreciated :-)

DanielHougaard:
@tiagovtristao this has now been fixed in version `v0.7.2`. Thank you for letting us know!

PR #2484, opened 2024-09-25
fix(k8-operator): updating CRD does not reflect in operator

DanielHougaard:
# Description 📣

This PR resolves a bug in the operator which was causing the operator not to reflect changes to the Infisical CRD. This was due to a state issue which caused the old CRD details to remain, even after updating the CRD.

The full issue is mentioned in https://github.com/Infisical/infisical/issues/2400

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/k8-operator/controllers/infisicalsecret_controller.go
+++ b/k8-operator/controllers/infisicalsecret_controller.go
@@ -5,17 +5,13 @@ import (
 	"fmt"
 	"time"
 
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	controllerUtil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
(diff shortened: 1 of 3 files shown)


k8-operator/controllers/infisicalsecret_controller.go
```diff
@@ -187,12 +187,18 @@ func (r *InfisicalSecretReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 func (r *InfisicalSecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&secretsv1alpha1.InfisicalSecret{}).
+		// For the Infisical Secret CRD we specifically watch for the update event, because we need to delete the entry in the resourceVariablesMap, so it will refresh the variables used to fetch secrets.
+		For(&secretsv1alpha1.InfisicalSecret{}, builder.WithPredicates(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				delete(resourceVariablesMap, string(e.ObjectNew.GetUID()))
```
maidul98, line 193: So deleting this will trigger CRD update? (not a big fan of the finalizer pattern, i have seen customers run into issues where the delete is in a stuck state, but this is unrelated to the PR)

maidul98 (approved):

PR #2488, opened 2024-09-26
fix: dashboard not showing when root accessn not provided

akhilmhdh:
# Description 📣

Fix for dashboard and overview not able to access when root folder access not provided

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v3/dashboard-router.ts
+++ b/backend/src/server/routes/v3/dashboard-router.ts
@@ -1,3 +1,4 @@
+import { ForbiddenError } from "@casl/ability";
 import { z } from "zod";
 
 import { SecretFoldersSchema, SecretImportsSchema, SecretTagsSchema } from "@app/db/schemas";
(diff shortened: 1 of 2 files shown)


frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
```diff
@@ -191,20 +191,10 @@ export const SecretOverviewPage = () => {
 
   const userAvailableEnvs = currentWorkspace?.environments || [];
 
-  const readableEnvs = userAvailableEnvs?.filter(({ slug }) =>
-    permission.can(
-      ProjectPermissionActions.Read,
-      subject(ProjectPermissionSub.Secrets, {
-        environment: slug,
-        secretPath
```
maidul98, line 199: Wouldn't this be solved if we had the same custom glob pattern defined in the frontend too? It's not that this is in correct but more so that the CASAL is not 1:1 like the backend. Or did i miss something

maidul98 (commented):
Left a question

backend/src/server/routes/v3/dashboard-router.ts
```diff
@@ -173,108 +174,114 @@ export const registerDashboardRouter = async (server: FastifyZodProvider) => {
...
+                      ...req.auditLogInfo
+                    }
+                  });
+                }
               }
             }
           }
         }
+      } catch (error) {
+        if (!(error instanceof ForbiddenError)) {
```
maidul98, line 282: So if it is a instance of forbidden then it won’t throw error? How come

akhilmhdh: It will be an empty list

maidul98: Why not throw the error? It feels like something the user should be notifed of

akhilmhdh: It shouldn't. Think of this way. User coming to the root folder. They only have access to nested folder 

Now they have 4 folders and according to pagination it gets 20 per page. So the remaining 16 would be secrets. But the secrets will throw an error as permission doesn't allow thus the dashboard won't load and show the folders.

What happens the folder is not shown. It means they can't access the nested folders.

maidul98: Oh i see. It sounds like behavior similar to what we had before where if you didn't have access to secrets, it just souldn't show for that path

frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx
akhilmhdh: Both frontend and backend has all the same operator and there is glob support. But think of this way. 

The user permission is `/backend`. But if you check permission in `/` both backend and frontend will throw an error. That's folders need to shown irrespective of both

maidul98 (approved):

PR #2494, opened 2024-09-26
feat(api): better errors and documentation

DanielHougaard:
# Description 📣

This PR introduces better use of our internal errors. Currently we aren't truly following a standard with our API errors. We are greatly overusing Unauthorized errors and Bad Request errors. This PR introduces a wider range use of Forbidden errors and Not Found errors. 

This PR also introduces
1. Fixed Forbidden and Unauthorized errors status codes being swapped around in our error handler, [see here](https://github.com/Infisical/infisical/pull/2494/files#diff-f9f0d9293f47bd302107c64be06a2d477c681880ca2fac3f8f5dec10ae44a96e) for more information.
2. Added more descriptive errors and improved wording across the entire codebase.
3. Improved API docs that show all possible error combinations.
## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/server/routes/sanitizedSchemas.ts
+++ b/backend/src/server/routes/sanitizedSchemas.ts
@@ -27,6 +27,44 @@ export const integrationAuthPubSchema = IntegrationAuthsSchema.pick({
   updatedAt: true
 });
 
+export const DefaultResponseErrorsSchema = {
+  400: z.object({
+    statusCode: z.literal(400),
+    message: z.literal(
+      `<string>: Contains a human-readable message about what went wrong. 400 errors are bad request errors, and are will occur to happen when passing or unexpected bad data in the request`
+    ),
+    error: z.literal(`<string>: Contains a human-readable error name.`)
+  }),
+  404: z.object({
+    statusCode: z.literal(404),
+    message: z.literal(
+      `<string>: Contains a human-readable message about what went wrong. 404 errors are not found errors, and will occur when a resource is not found`
+    ),
+    error: z.literal(`<string>: Contains a human-readable error name.`)
+  }),
+  401: z.object({
+    statusCode: z.literal(401),
+    message: z.literal(
+      `<string>: Contains a human-readable message about what went wrong. 401 errors are unauthorized errors, and will occur when attempting to access a resource with invalid or missing credentials`
+    ),
+    error: z.literal(`<string>: Contains a human-readable error name.`)
+  }),
+  403: z.object({
+    statusCode: z.literal(403),
+    message: z.literal(
+      `<string>: Contains a human-readable message about what went wrong. 403 errors are forbidden errors, and will occur when attempting to access a resource without the necessary permissions`
+    ),
+    error: z.literal(`<string>: Contains a human-readable error name.`)
+  }),
+  500: z.object({
+    statusCode: z.literal(500),
+    message: z.literal(
+      `<string>: Contains a human-readable message about what went wrong. 500 errors are internal server errors, and will occur when an unexpected error occurs on the server`
+    ),
+    error: z.literal(`<string>: Contains a human-readable error name.`)
+  })
+};
+
 export const sapPubSchema = SecretApprovalPoliciesSchema.merge(
   z.object({
     environment: z.object({
--- a/backend/src/server/routes/sanitizedSchemas.ts
+++ b/backend/src/server/routes/sanitizedSchemas.ts
@@ -51,10 +51,10 @@ export const DefaultResponseErrorsSchema = {
   }),
   403: z.object({
     statusCode: z.literal(403),
-    message: z.literal(
-      `<string>: Contains a human-readable message about what went wrong. 403 errors are forbidden errors, and will occur when attempting to access a resource without the necessary permissions`
-    ),
-    error: z.literal(`<string>: Contains a human-readable error name.`)
+    message: z.any(),
+    error: z.literal(
+      `<string>: Contains a human-readable error name. The message can be both an object or a string. It will always be a string except if the error is happening due to a request validation error.`
+    )
   }),
   500: z.object({
     statusCode: z.literal(500),
--- a/backend/src/server/routes/sanitizedSchemas.ts
+++ b/backend/src/server/routes/sanitizedSchemas.ts
@@ -30,38 +30,28 @@ export const integrationAuthPubSchema = IntegrationAuthsSchema.pick({
 export const DefaultResponseErrorsSchema = {
   400: z.object({
     statusCode: z.literal(400),
-    message: z.literal(
-      `<string>: Contains a human-readable message about what went wrong. 400 errors are bad request errors, and are will occur to happen when passing or unexpected bad data in the request`
-    ),
-    error: z.literal(`<string>: Contains a human-readable error name.`)
+    message: z.string(),
+    error: z.string()
   }),
   404: z.object({
     statusCode: z.literal(404),
-    message: z.literal(
-      `<string>: Contains a human-readable message about what went wrong. 404 errors are not found errors, and will occur when a resource is not found`
-    ),
-    error: z.literal(`<string>: Contains a human-readable error name.`)
+    message: z.string(),
+    error: z.string()
   }),
   401: z.object({
     statusCode: z.literal(401),
-    message: z.literal(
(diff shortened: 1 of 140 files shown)


backend/src/server/routes/sanitizedSchemas.ts
```diff
@@ -27,6 +27,44 @@ export const integrationAuthPubSchema = IntegrationAuthsSchema.pick({
...
+  404: z.object({
+    statusCode: z.literal(404),
+    message: z.literal(
+      `<string>: Contains a human-readable message about what went wrong. 404 errors are not found errors, and will occur when a resource is not found`
+    ),
+    error: z.literal(`<string>: Contains a human-readable error name.`)
+  }),
+  401: z.object({
+    statusCode: z.literal(401),
+    message: z.literal(
```
akhilmhdh, line 47: https://github.com/Infisical/infisical/pull/2494/files#diff-f9f0d9293f47bd302107c64be06a2d477c681880ca2fac3f8f5dec10ae44a96eR34

@DanielHougaard The message may not be string, for validation error it will be JSON object. This would cause response schema internal server error

akhilmhdh: Also @DanielHougaard You don't need to manually do this. There is a hook called `onHook` in fastify. You can write a plugin to inject it for each route.

akhilmhdh: https://fastify.dev/docs/latest/Reference/Hooks/#onroute

docs/mint.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Infisical",
-  "openapi": "https://app.infisical.com/api/docs/json",
+  "openapi": "http://localhost:8080/api/docs/json",
```
akhilmhdh, line 3: I guess this needs to be reverted back.

akhilmhdh (changes requested):
Some small changes and you can avoid that manully adding errro response schema everywhere with a plugin.

Also the message let's keep it unknown rather than string. Because it could string, JSON object depending on situation. By accident if we return object and its a string - fastify would cause an internal server error as response didn't match

akhilmhdh:
Also it seems test are failing?

backend/src/server/routes/sanitizedSchemas.ts
maidul98: Yeah, this seems pretty redundant

backend/src/services/identity-access-token/identity-access-token-service.ts
```diff
@@ -172,7 +176,7 @@ export const identityAccessTokenServiceFactory = ({
     });
 
     if (!identityOrgMembership) {
-      throw new UnauthorizedError({ message: "Identity does not belong to any organization" });
+      throw new BadRequestError({ message: "Identity does not belong to any organization" });
```
maidul98, line 179: This would mean that it is a client side error?

akhilmhdh (approved):

backend/src/services/secret/secret-service.ts
```diff
@@ -981,7 +981,7 @@ export const secretServiceFactory = ({
     if (!shouldUseSecretV2Bridge)
       throw new BadRequestError({
         message: "Project version does not support pagination",
-        name: "pagination_not_supported"
```
maidul98, line 984: Are we sure this isn't being used in any one of our clients?

backend/src/services/secret-import/secret-import-service.ts
```diff
@@ -118,7 +118,7 @@ export const secretImportServiceFactory = ({
         importEnv: folder.environment.id,
         importPath: secretPath
       });
-      if (existingImport) throw new BadRequestError({ message: "Cyclic import not allowed" });
+      if (existingImport) throw new NotFoundError({ message: "Cyclic import not allowed" });
```
maidul98, line 121: This feels like a bad request from the user? They shouldn't be doing cyclic import

backend/src/services/secret-import/secret-import-service.ts
```diff
@@ -546,9 +546,9 @@ export const secretImportServiceFactory = ({
     }
 
     if (!botKey)
-      throw new BadRequestError({
+      throw new NotFoundError({
         message: "Project bot not found. Please upgrade your project.",
-        name: "bot_not_found_error"
```
maidul98, line 551: Are we sure no client is using this error name conditionally?

backend/src/services/service-token/service-token-service.ts
```diff
@@ -154,21 +154,21 @@ export const serviceTokenServiceFactory = ({
...
+    if (!serviceToken) throw new NotFoundError({ message: "Service token not found" });
     const project = await projectDAL.findById(serviceToken.projectId);
 
-    if (!project) throw new UnauthorizedError({ message: "Service token project not found" });
+    if (!project) throw new NotFoundError({ message: "Service token project not found" });
 
     if (serviceToken.expiresAt && new Date(serviceToken.expiresAt) < new Date()) {
       await serviceTokenDAL.deleteById(serviceToken.id);
-      throw new UnauthorizedError({ message: "failed to authenticate expired service token" });
+      throw new ForbiddenRequestError({ message: "Service token has expired" });
```
maidul98, line 167: The user is not authenticated at all here so we should be throwing UnauthorizedError. The credentials they provided are expired

backend/src/services/service-token/service-token-service.ts
```diff
@@ -154,21 +154,21 @@ export const serviceTokenServiceFactory = ({
...
     }
 
-    const isMatch = await bcrypt.compare(TOKEN_SECRET, serviceToken.secretHash);
-    if (!isMatch) throw new UnauthorizedError();
+    const isMatch = await bcrypt.compare(tokenSecret, serviceToken.secretHash);
+    if (!isMatch) throw new ForbiddenRequestError();
```
maidul98, line 171: I think this should also remain UnauthorizedError

backend/src/services/identity-access-token/identity-access-token-service.ts
```diff
@@ -39,7 +39,7 @@ export const identityAccessTokenServiceFactory = ({
 
     if (accessTokenNumUsesLimit > 0 && accessTokenNumUses > 0 && accessTokenNumUses >= accessTokenNumUsesLimit) {
       await identityAccessTokenDAL.deleteById(tokenId);
-      throw new BadRequestError({
+      throw new ForbiddenRequestError({
```
maidul98, line 42: Token it self is invalid and this isn't about a permission issue. It should be UnauthorizedError, let's doube check ones similar to these

backend/src/services/identity-access-token/identity-access-token-service.ts
```diff
@@ -55,7 +55,7 @@ export const identityAccessTokenServiceFactory = ({
 
         if (currentDate > expirationDate) {
           await identityAccessTokenDAL.deleteById(tokenId);
-          throw new UnauthorizedError({
+          throw new ForbiddenRequestError({
```
maidul98, line 58: Same here, the token itself is invalid

backend/src/services/identity-access-token/identity-access-token-service.ts
```diff
@@ -67,7 +67,7 @@ export const identityAccessTokenServiceFactory = ({
...
```
maidul98, line 70: Same here, the token itself is invalid

backend/src/services/identity-access-token/identity-access-token-service.ts
```diff
@@ -107,15 +109,15 @@ export const identityAccessTokenServiceFactory = ({
 
       if (currentDate > expirationDate) {
         await identityAccessTokenDAL.deleteById(identityAccessToken.id);
-        throw new UnauthorizedError({
+        throw new ForbiddenRequestError({
```
maidul98, line 112: 401 Unauthorized: when the request lacks valid authentication credentials or the provided credentials have become invalid (expired, etc)

403 Forbidden: when the client is authenticated but lacks the necessary permissions for the requested action

This should also be UnauthorizedError. If you can double check this file for this, that would be good

maidul98 (commented):
Left some comments and questions. I saw you merged it in so maybe you can revert this

PR #2498, opened 2024-09-27
Fix: Various UI Improvements, Fixes and Backend Refactoring

scott-ray-wilson:
# Description 📣

This PR adds AWS region slugs to all integration dropdowns, fixes dropdown overflow, adds explanatory tooltips to the secrets table resource counter, moves the dashboard router to v1, and improves handling of querystring boolean values for dashboard endpoints.

## Type ✨

- [x] Improvements
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v1/dashboard-router.ts
+++ b/backend/src/server/routes/v1/dashboard-router.ts
@@ -23,7 +23,7 @@ const booleanSchema = z
   .transform((value) => {
     if (typeof value === "string") {
       // ie if not empty, 0 or false, return true
-      return !!value && Number(value) !== 0 && value.toLowerCase() !== "false";
+      return Boolean(value) && Number(value) !== 0 && value.toLowerCase() !== "false";
     }
 
     return value;
--- a/backend/src/server/routes/v1/dashboard-router.ts
+++ b/backend/src/server/routes/v1/dashboard-router.ts
@@ -17,6 +17,20 @@ import { AuthMode } from "@app/services/auth/auth-type";
 import { SecretsOrderBy } from "@app/services/secret/secret-types";
 import { PostHogEventTypes } from "@app/services/telemetry/telemetry-types";
 
+// handle querystring boolean values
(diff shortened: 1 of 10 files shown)


scott-ray-wilson:
@akhilmhdh this is not a publically exposed endpoint so the breaking API change is fine correct?

akhilmhdh:
@scott-ray-wilson Yes it's fine

backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -17,6 +17,20 @@ import { AuthMode } from "@app/services/auth/auth-type";
 import { SecretsOrderBy } from "@app/services/secret/secret-types";
 import { PostHogEventTypes } from "@app/services/telemetry/telemetry-types";
 
+// handle querystring boolean values
+const booleanSchema = z
+  .union([z.boolean(), z.string().trim()])
+  .transform((value) => {
+    if (typeof value === "string") {
+      // ie if not empty, 0 or false, return true
+      return !!value && Number(value) !== 0 && value.toLowerCase() !== "false";
```
akhilmhdh, line 26: This should be an OR gate right?

Also `!!value` => Boolean(value)

akhilmhdh (changes requested):

akhilmhdh:
I'll test this with an integration today and get back to you interms of UI

backend/src/server/routes/v1/dashboard-router.ts
scott-ray-wilson: Will swap to `Boolean(value)` but why would should this be an OR gate? 

it's essentially saying return `true` only if not "" and "0" and "false",  which all of these should evaluate true for it to be `true`?

scott-ray-wilson:
@akhilmhdh bump

akhilmhdh (approved):

PR #2507, opened 2024-09-28
Fix: Integration Sync Retry on Error Patch

scott-ray-wilson:
# Description 📣

This PR fixes integration sync retries on failure.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/secret/secret-queue.ts
+++ b/backend/src/services/secret/secret-queue.ts
@@ -930,6 +930,11 @@ export const secretQueueFactory = ({
               }
             });
 
+            // re-throw error to re-run job unless final attempt, then log and send fail email
+            if (job.attemptsStarted !== job.opts.attempts) {
+              throw err;
+            }
+
             await integrationDAL.updateById(integration.id, {
               lastSyncJobId: job.id,
               syncMessage: message,


akhilmhdh (approved):

akhilmhdh (changes requested):
Small issue

backend/src/services/secret/secret-queue.ts
```diff
@@ -930,6 +930,11 @@ export const secretQueueFactory = ({
               }
             });
 
+            // re-throw error to re-run job unless final attempt, then log and send fail email
+            if (job.attemptsStarted !== job.opts.attempts) {
```
akhilmhdh, line 934: @scott-ray-wilson Don't this throw still triggers the failed integration email.

scott-ray-wilson: @akhilmhdh it will only send the email if it fails on the final attempt (ie when the error isn't re-thrown)

akhilmhdh: That's beacuse there sendFailedIntegration sync has a debounce. But it still triggered on first try.

Even if your throw from catch, finally will execute in JS event cycle

akhilmhdh (approved):

PR #2512, opened 2024-09-30
feat: enforce oidc sso

sheensantoscapadngan:
# Description 📣
- This PR will enable users to enforce OIDC SSO logins in their organizations

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/services/org/org-dal.ts
```diff
@@ -29,13 +29,37 @@ export const orgDALFactory = (db: TDbClient) => {
...
+        .select(selectAllTableCols(TableName.Organization))
+        .select(
+          db.raw(`
+            CASE 
+              WHEN ${TableName.SamlConfig}."orgId" IS NOT NULL THEN 'saml'
+              WHEN ${TableName.OidcConfig}."orgId" IS NOT NULL THEN 'oidc'
+              ELSE ''
+            END as "orgAuthMethod"
+        `)
+        )) as (TOrganizations & { orgAuthMethod: string })[];
```
dangtony98, line 61: Doesn't this logic imply that an organization can only have one organization-level auth method? If we decide in the future to allow organizations to support multiple organization-level auth methods, then won't we have to undothis?

I'm thinking for the sake of flexibility, to not impose this assumption — Can we do without it?

dangtony98 (changes requested):

backend/src/services/org/org-dal.ts
sheensantoscapadngan: The reason why we need this is so that we know where to redirect the user to during enforcement of SSO method. I don't think users would want to enforce both OIDC and SAML logins

image

dangtony98 (approved):

PR #2514, opened 2024-09-30
fix: allow creation of multiple project envs

DanielHougaard:
# Description 📣

Previously when creating multiple project environments at once, it was possible that our database would throw an error, due to the unique composite constraint on project ID / position. This would happen because the transactions would turn into race conditions of whichever would first finish wish the selected position.

The solution for this is to use the keystore to ensure that only one project environment update / create / delete operation will run at once.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/project-env/project-env-service.ts
+++ b/backend/src/services/project-env/project-env-service.ts
@@ -50,7 +50,7 @@ export const projectEnvServiceFactory = ({
     ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionActions.Create, ProjectPermissionSub.Environments);
 
     const lock = await keyStore
-      .acquireLock([KeyStorePrefixes.ProjectEnvironmentCreation, projectId], 5000)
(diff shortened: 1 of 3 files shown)


backend/src/services/project-env/project-env-service.ts
```diff
@@ -45,32 +49,56 @@ export const projectEnvServiceFactory = ({
...
 
-    const envs = await projectEnvDAL.find({ projectId });
-    const existingEnv = envs.find(({ slug: envSlug }) => envSlug === slug);
-    if (existingEnv)
-      throw new BadRequestError({
-        message: "Environment with slug already exist",
-        name: "CreateEnvironment"
-      });
+    const lock = await keyStore
+      .acquireLock([KeyStorePrefixes.ProjectEnvironmentCreate, projectId], 5000)
```
maidul98, line 53: Shouldn't we use a function for the key here?

maidul98 (approved):

PR #2529, opened 2024-10-03
Databricks integration

vmatsiiako:
# Description 📣

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


frontend/src/pages/integrations/databricks/create.tsx
```diff
@@ -0,0 +1,251 @@
...
+} from "../../../hooks/api/integrationAuth";
+import { useGetWorkspaceById } from "../../../hooks/api/workspace";
+
+export default function DatabricksCreateIntegrationPage() {
+  const router = useRouter();
+  const { mutateAsync } = useCreateIntegration();
+
+  const { integrationAuthId } = queryString.parse(router.asPath.split("?")[1]);
+
+  console.log("integrationAuthId", integrationAuthId)
```
scott-ray-wilson, line 39: remove console log

frontend/src/pages/integrations/databricks/create.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+  const { data: workspace } = useGetWorkspaceById(localStorage.getItem("projectData.id") ?? "");
+  const { data: integrationAuth, isLoading: isintegrationAuthLoading } = useGetIntegrationAuthById(
+    (integrationAuthId as string) ?? ""
+  );
+
+  console.log("integrationAuth", integrationAuth)
```
scott-ray-wilson, line 46: remove console log

frontend/src/pages/integrations/databricks/create.tsx
```diff
@@ -0,0 +1,251 @@
...
+  const { data: integrationAuthScopes, isLoading: isIntegrationAuthScopesLoading } =
+    useGetIntegrationAuthApps({
+      integrationAuthId: (integrationAuthId as string) ?? ""
+    });
+
+  console.log("integrationAuthScopes", integrationAuthScopes)
```
scott-ray-wilson, line 52: remove console log

backend/src/services/integration-auth/integration-app-list.ts
```diff
@@ -455,6 +455,28 @@ const getAppsCircleCI = async ({ accessToken }: { accessToken: string }) => {
   return apps;
 };
 
+/**
+ * Return list of projects for Databricks integration
+ */
+const getAppsDatabricks = async ({ accessToken }: { accessToken: string }) => {
+  const res = await request.get<{ scopes: { name: string; backend_type: string }[] }>(
+    `${IntegrationUrls.DATABRICKS_API_URL}/2.0/secrets/scopes/list`,
```
scott-ray-wilson, line 463: I think we should be passing in the databricks workspace url from the integration as it's unique to each instance

backend/src/services/integration-auth/integration-app-list.ts
```diff
@@ -455,6 +455,28 @@ const getAppsCircleCI = async ({ accessToken }: { accessToken: string }) => {
...
+        Authorization: `Bearer ${accessToken}`,
+        "Accept-Encoding": "application/json"
+      }
+    }
+  );
+
+  const scopes = res.data.scopes.map((a) => ({
+    name: a.name, // name maps to unique scope name in Databricks
+    backend_type: a.backend_type
+  }));
```
scott-ray-wilson, line 475: If a user doesn't have any scopes this throws an error - might be nice to return an empty array and notify the user they have no scopes

backend/src/services/integration-auth/integration-app-list.ts
```diff
@@ -1104,6 +1126,11 @@ export const getApps = async ({
         accessToken
       });
 
+    case Integrations.DATABRICKS:
+      return getAppsDatabricks({
+        accessToken
```
scott-ray-wilson, line 1131: pass integration databricks workspace url

backend/src/services/integration-auth/integration-list.ts
```diff
@@ -73,6 +74,7 @@ export enum IntegrationUrls {
   RAILWAY_API_URL = "https://backboard.railway.app/graphql/v2",
   FLYIO_API_URL = "https://api.fly.io/graphql",
   CIRCLECI_API_URL = "https://circleci.com/api",
+  DATABRICKS_API_URL = "https:/dbc-5e0185c0-20ac.cloud.databricks.com/api",
```
scott-ray-wilson, line 77: This is your personal databricks workspace url no? I had to overwrite this value with mine to get this integration to work - think we need to use their databricks workspace url from the integration form and pass it were pertinent instead of using a fixed value

backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -4021,6 +4095,14 @@ export const syncIntegrationSecrets = async ({
         accessToken
       });
       break;
+    case Integrations.DATABRICKS:
+      await syncSecretsDatabricks({
+        integration,
+        integrationAuth,
+        secrets,
+        accessToken
```
scott-ray-wilson, line 4103: pass databricks workspace url

frontend/src/pages/integrations/databricks/authorize.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { faArrowUpRightFromSquare, faBookOpen } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import { useSaveIntegrationAccessToken } from "@app/hooks/api";
+
+import { Button, Card, CardTitle, FormControl, Input } from "../../../components/v2";
+
+export default function DatabricksCreateIntegrationPage() {
+  const router = useRouter();
+  const { mutateAsync } = useSaveIntegrationAccessToken();
```
scott-ray-wilson, line 15: optionally can use `isLoading` from mutation hook instead of having to use manual state

frontend/src/pages/integrations/databricks/authorize.tsx
```diff
@@ -0,0 +1,118 @@
...
+        url: instanceURL,
+        accessToken: apiKey
+      });
+
+      setIsLoading(false);
+
+      router.push(`/integrations/databricks/create?integrationAuthId=${integrationAuth.id}`);
+    } catch (err) {
+      console.error(err);
+    }
```
scott-ray-wilson, line 50: `setIsLoading(false);` should probably be moved after the `catch` or in a `finally` so that it is set `false` whether it fails or succeeds. Alternatively use `isLoading` from mutation hook above to not have to manage this

frontend/src/pages/integrations/databricks/authorize.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { Button, Card, CardTitle, FormControl, Input } from "../../../components/v2";
+
+export default function DatabricksCreateIntegrationPage() {
+  const router = useRouter();
+  const { mutateAsync } = useSaveIntegrationAccessToken();
+
+  const [apiKey, setApiKey] = useState("");
+  const [instanceURL, setInstanceURL] = useState("");
+  const [apiKeyErrorText, setApiKeyErrorText] = useState("");
+  const [instanceURLErrorText, setInstanceURLErrorText] = useState("");
```
scott-ray-wilson, line 20: Are you familiar with `react-hook-form`? If not I'd be happy to run through it with you! We use it in other locations in the app -  It makes field/form management with errors a lot simpler - no change required just more a curiosity

frontend/src/pages/integrations/databricks/create.tsx
```diff
@@ -0,0 +1,251 @@
...
+        sourceEnvironment: selectedSourceEnvironment,
+        secretPath
+      });
+
+      setIsLoading(false);
+
+      router.push(`/integrations/${localStorage.getItem("projectData.id")}`);
+    } catch (err) {
+      console.error(err);
+    }
```
scott-ray-wilson, line 106: same as above - suggest moving `setIsLoading(false);` after `catch` or in `finally` or using the mutation hook `isLoading`

scott-ray-wilson (changes requested):
Syncing worked great once I realized I needed to override the hardcoded `DATABRICKS_API_URL` - see comments for more details but think we need to be using the workspace URL dynamically from the submitted integration form

backend/src/services/integration-auth/integration-sync-secret.ts
vmatsiiako: actually it's passed through integrationAuth here I think

frontend/src/pages/integrations/databricks/authorize.tsx
vmatsiiako: But shouldn't it be before the router.push though?

backend/src/services/integration-auth/integration-list.ts
vmatsiiako: Yes, indeed. This is actually not used anymore.

backend/src/services/integration-auth/integration-app-list.ts
vmatsiiako: great catch!

frontend/src/pages/integrations/databricks/authorize.tsx
DanielHougaard: Setting it before or after router.push doesn't matter in this case, because router.push is fast and will finish before the page reloads anyways, I would move `setIsLoading(false)` to the `finally` block like Scott mentioned

frontend/src/pages/integrations/databricks/authorize.tsx
```diff
@@ -0,0 +1,118 @@
...
+          className="px-6"
+        >
+          <Input placeholder="" value={apiKey} onChange={(e) => setApiKey(e.target.value)} type="password" />
+        </FormControl>
+        <Button
+          onClick={handleButtonClick}
+          colorSchema="primary"
+          variant="outline_bg"
+          className="mb-6 mt-2 ml-auto mr-6 w-min"
+          isLoading={isLoading}
```
scott-ray-wilson, line 109: Suggest disabling button while loading

frontend/src/pages/integrations/databricks/create.tsx
```diff
@@ -0,0 +1,251 @@
...
+            </Select>
+          </FormControl>
+        )}
+        <Button
+          onClick={handleButtonClick}
+          colorSchema="primary"
+          variant="outline_bg"
+          className="mb-6 mt-2 ml-auto mr-6 w-min"
+          isLoading={isLoading}
+          isDisabled={integrationAuthScopes.length === 0}
```
scott-ray-wilson, line 196: Suggest disabling button while loading as well

frontend/src/pages/integrations/databricks/authorize.tsx
```diff
@@ -0,0 +1,118 @@
...
+        setInstanceURLErrorText("Instance URL cannot be blank");
+        return;
+      }
+
+      setIsLoading(true);
+
+      const integrationAuth = await mutateAsync({
+        workspaceId: localStorage.getItem("projectData.id"),
+        integration: "databricks",
+        url: instanceURL,
```
scott-ray-wilson, line 41: QoL remove trailing slash if present or handle server-side

scott-ray-wilson (changes requested):

scott-ray-wilson (approved):

PR #2530, opened 2024-10-04
feat: terraform imports prerequisite / api improvements

DanielHougaard:
# Description 📣

This PR is a prerequisite for supporting Terraform Imports. It allows us to fetch secret imports purely by ID, and it exposes more information about secret folders, which is necessary for Terraform.

#### Minor
I also added a new type to our Pull request template called Improvement, which should be used when making improvements/adjustments to existing code, as suggested by @scott-ray-wilson.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
---

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -6,6 +6,7 @@
 
 - [ ] Bug fix
 - [ ] New feature
+- [ ] Improvement
 - [ ] Breaking change
 - [ ] Documentation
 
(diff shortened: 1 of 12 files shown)


akhilmhdh (approved):
Breaking changes in api if it's unrelated to ignore.

The folder dal can do a join with the environment to avoid that second environment call. But i guess I'll be making that change later.

PR #2573, opened 2024-10-11
feat: Process Envkey import in queue 

DanielHougaard:
# Description 📣

This PR implements Redis support for external imports. The reason behind this implementation is to support larger file uploads. Previously if EnvKey imports were larger than 2MB, the request would time-out. The first solution to this was to use `createManySecrets` instead of creating secrets 1-by-1. This wasn't enough for large imports (7MB and above). This Redis implementation allows for any import size. We notify users via. email when 1: the import starts, 2: the import fails or succeeds.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/ee/services/secret-snapshot/secret-snapshot-service.ts
+++ b/backend/src/ee/services/secret-snapshot/secret-snapshot-service.ts
@@ -1,9 +1,15 @@
 import { ForbiddenError, subject } from "@casl/ability";
 
-import { TableName, TSecretTagJunctionInsert, TSecretV2TagJunctionInsert } from "@app/db/schemas";
+import {
+  TableName,
+  TSecretSnapshotFolders,
+  TSecretSnapshotSecretsV2,
+  TSecretTagJunctionInsert,
+  TSecretV2TagJunctionInsert
+} from "@app/db/schemas";
 import { decryptSymmetric128BitHexKeyUTF8 } from "@app/lib/crypto";
 import { InternalServerError, NotFoundError } from "@app/lib/errors";
-import { groupBy } from "@app/lib/fn";
+import { chunkArray, groupBy } from "@app/lib/fn";
 import { logger } from "@app/lib/logger";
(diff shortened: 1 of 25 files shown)


backend/src/ee/services/secret-snapshot/secret-snapshot-service.ts
```diff
@@ -240,22 +246,37 @@ export const secretSnapshotServiceFactory = ({
...
-          const snapshotFolders = await snapshotFolderDAL.insertMany(
-            folderVersions.map(({ id }) => ({
-              folderVersionId: id,
-              envId: folder.environment.envId,
-              snapshotId: newSnapshot.id
-            })),
-            tx
-          );
+
+          const chunkedSnapshotSecrets = chunkArray(secretVersions, 2500);
```
akhilmhdh, line 250: Ohh your chunking like this. 

You can do this instead. Its in the ORM already.

https://github.com/Infisical/infisical/blob/09d28156f8e1a60cde9f31bd7db89502319c00b7/backend/src/lib/knex/index.ts#L144

backend/src/services/external-migration/external-migration-fns.ts
```diff
@@ -141,7 +141,7 @@ export const importDataIntoInfisicalFn = async ({
 
   // Import environments
   if (data.environments) {
```
akhilmhdh, line 143: The one major issue i see with this whole setup is consider it failed. It keeps the stale things.

So they can't do it again in same project right? Because consider createEnvironment, previous environment won't allow it. They would have to create a new project for this again.

Now same could get worse for secrets as well.  One secret failing could be quite troublesome

akhilmhdh (changes requested):

maidul98 (approved):
Doesn't work with 25mb file but works with 10. Mergeing in because we can't wait much longer

PR #2575, opened 2024-10-11
feat: added secret path to template and optional more arguments as js…

akhilmhdh:
# Description 📣

1. Added secret path to secrets function in template
2. Added optional argument as a json object for more option like recursive

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/cli/packages/cmd/agent.go
+++ b/cli/packages/cmd/agent.go
@@ -317,6 +317,8 @@ type secretArguments struct {
 }
 
 func secretTemplateFunction(accessToken string, existingEtag string, currentEtag *string) func(string, string, string, ...string) ([]models.SingleEnvironmentVariable, error) {
+	// ...string is because golang doesn't have optional arguments.
+	// thus we make it slice and pick it only first element
 	return func(projectID, envSlug, secretPath string, args ...string) ([]models.SingleEnvironmentVariable, error) {
 		var parsedArguments secretArguments
 		// to make it optional
--- a/cli/packages/cmd/agent.go
+++ b/cli/packages/cmd/agent.go
@@ -313,7 +313,15 @@ func ParseAgentConfig(configFile []byte) (*Config, error) {
 }
 
 type secretArguments struct {
-	IsRecursive bool `json:"recursive"`
+	IsRecursive                  bool  `json:"recursive"`
+	ShouldExpandSecretReferences *bool `json:"expandSecretReferences,omitempty"`
+}
+
+func (s *secretArguments) SetDefaults() {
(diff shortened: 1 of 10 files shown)


cli/packages/cmd/agent.go
```diff
@@ -311,9 +312,24 @@ func ParseAgentConfig(configFile []byte) (*Config, error) {
...
+}
+
+func secretTemplateFunction(accessToken string, existingEtag string, currentEtag *string) func(string, string, string, ...string) ([]models.SingleEnvironmentVariable, error) {
+	// ...string is because golang doesn't have optional arguments.
+	// thus we make it slice and pick it only first element
+	return func(projectID, envSlug, secretPath string, args ...string) ([]models.SingleEnvironmentVariable, error) {
+		var parsedArguments secretArguments
+		// to make it optional
+		if len(args) > 0 {
+			err := json.Unmarshal([]byte(args[0]), &parsedArguments)
```
maidul98, line 326: but what if there are more arguments? How will that work? Maybe we should assign each arg number to a field so it is clear how to add other args in the future?

cli/secret-render-template
```diff
@@ -1,5 +1,5 @@
-{{- with secret "6553ccb2b7da580d7f6e7260" "dev" "/" }}
+{{- with secret "ae06a663-2740-49c5-b998-bf4ceca2c120" "dev" "/" `{"recursive":true}` }}
```
maidul98, line 1: If we need to add other args before `{"recursive":true}` then would it be a issue?

maidul98 (commented):
While we are at it, we should also add support for secret reference expending. Also i don't see any docs update, we should update the docs so people can find out of these template updates https://infisical.com/docs/integrations/platforms/infisical-agent#available-secret-template-functions

cli/packages/cmd/agent.go
akhilmhdh: What do you mean other args it will be simply a json object.

{recursive: true, arg2: value ...}

cli/secret-render-template
akhilmhdh: Answered as above

cli/packages/cmd/agent.go
maidul98: I guess a better question is how will the unmarshal work when there are more than one arg? I see you doing arg[0]. Will we have a bunch of if statments to unmarshal each arg?

maidul98 (approved):

PR #2589, opened 2024-10-15
Feature: SCIM Group to Organization Role Mapping

scott-ray-wilson:
# Description 📣

This PR enables admins to configure SCIM group organization role mappings to specify which organization membership role newly provisioned users from that group should have.

CleanShot 2024-10-15 at 07 25 58@2x

CleanShot 2024-10-15 at 07 30 43@2x

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/db/schemas/external-group-org-role-mappings.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { z } from "zod";
+
+import { TImmutableDBKeys } from "./models";
+
+export const ExternalGroupOrgRoleMappingsSchema = z.object({
+  id: z.string().uuid(),
+  groupName: z.string(),
+  role: z.string(),
+  roleId: z.string().uuid().nullable().optional(),
```
sheensantoscapadngan, line 14: why is roleId here nullable and optional? it should be required right?

backend/src/server/routes/v1/external-group-org-role-mapping-router.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+      });
+
+      return mappings;
+    }
+  });
+
+  // update mappings for current org
+  server.route({
+    method: "POST", // using post since this endpoint creates, updates and deletes mappings
```
sheensantoscapadngan, line 41: nit: but I think PUT would be more appropriate because we're overwriting the entire mapping every request it seems like?

backend/src/services/external-group-org-role-mapping/external-group-org-role-mapping-dal.ts
```diff
@@ -0,0 +1,46 @@
...
+    const mappings = await externalGroupOrgRoleMappingOrm.transaction(async (tx) => {
+      await externalGroupOrgRoleMappingOrm.delete({ $in: { id: mappingsToDelete.map((mapping) => mapping.id) } }, tx);
+
+      const updatedMappings: TExternalGroupOrgRoleMappings[] = [];
+      for await (const { id, ...mappingData } of mappingsToUpdate) {
+        const updatedMapping = await externalGroupOrgRoleMappingOrm.update({ id }, mappingData, tx);
+        updatedMappings.push(updatedMapping[0]);
+      }
+
+      const insertedMappings = await externalGroupOrgRoleMappingOrm.insertMany(mappingsToInsert, tx);
```
sheensantoscapadngan, line 37: I think we can simplify this by:

1. deleting all org mappings
2. adding all org mappings

both done in a single transaction

backend/src/services/org/org-role-fns.ts
```diff
@@ -5,6 +5,9 @@ import { TOrgRoleDALFactory } from "@app/services/org/org-role-dal";
 
 const RESERVED_ORG_ROLE_SLUGS = Object.values(OrgMembershipRole).filter((role) => role !== "custom");
 
+export const isCustomOrgRole = (membershipSlug: string) =>
+  !RESERVED_ORG_ROLE_SLUGS.includes(membershipSlug as OrgMembershipRole);
```
sheensantoscapadngan, line 9: nit: I think roleSlug would be more appropriate as the parameter name?

backend/src/db/schemas/external-group-org-role-mappings.ts
scott-ray-wilson: `roleId` is only used if `role` is `custom` - if using a reserved role, ie `admin`, `member` etc. `roleId` is `null` (following `org_memberships` table convention)

backend/src/server/routes/v1/external-group-org-role-mapping-router.ts
scott-ray-wilson: ahhh yeah 100%!

backend/src/services/external-group-org-role-mapping/external-group-org-role-mapping-dal.ts
scott-ray-wilson: I actually did this originally but then I realized you lose context on `updatedAt` col... might not be a huge deal though - what are your thoughts?

sheensantoscapadngan (dismissed):

sheensantoscapadngan (approved):

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -692,6 +698,43 @@ export const scimServiceFactory = ({
...
+    // set new membership roles to group mapping value
+    await orgMembershipDAL.update(
+      {
+        $in: {
+          id: newOrgMemberships.map((membership) => membership.id)
+        }
+      },
+      {
+        role: externalGroupMapping.role,
+        roleId: externalGroupMapping.roleId
```
maidul98, line 733: Do we make sure that if a user tries to delete the custom role id, we prevent it until the role isn't assigned to a group mapping?

maidul98: What about non custom roles

maidul98 (approved):

PR #2595, opened 2024-10-16
feat: field type is not even used in schema so removed as some providers don't provide it

akhilmhdh:
# Description 📣

Some providers like okta don't provider type in emails section it seems. I just. removed it as we don't even use it also. So why make it required. We only care about primary flag in email.

On response we just make it fallback to work.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


dangtony98 (approved):
LGTM

PR #2603, opened 2024-10-17
feat: secret reference graph for understanding how its pulled

akhilmhdh:
# Description 📣

This PR introduces a secret reference details window in detail side bar. This helpers user to see the original value and also a graph view inference made to pull the reference. 

The function is attached to the original function so that both are kept in sync. The next thing is tracing is optional inside the function so that most of other the time expansion happens its not done

Screenshot 2024-10-17 at 4 29 45 PM

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


akhilmhdh:
Don't merge this until permission v2 is merged

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1636,6 +1638,77 @@ export const secretV2BridgeServiceFactory = ({
...
+    const tree = await getExpandedSecretStackTrace({
+      environment,
+      secretPath,
+      value: secretValue
+    });
+    const value = await expandSecretReferences({
+      environment,
+      secretPath,
+      value: secretValue
+    });
```
scott-ray-wilson, line 1708: Isn't this needlessly crawling references twice? it's split into two functions in `secret-v2-bridge-fns` but they both run the same function returning a different property. Maybe just making `stackTrace` an optional prop to `expandSecretReferences`?

frontend/src/views/SecretMainPage/components/SecretListView/SecretDetailSidebar.tsx
```diff
@@ -412,9 +417,38 @@ export const SecretDetailSidebar = ({
...
+                  {(isAllowed) => (
+                    <Modal>
+                      <ModalTrigger asChild>
+                        <Button
+                          variant="outline_bg"
+                          className="w-full px-2 py-1"
+                          leftIcon={<FontAwesomeIcon icon={faRoad} />}
+                          isDisabled={!isAllowed}
+                        >
+                          Reference tree
```
scott-ray-wilson, line 434: nit: `Reference Tree`

frontend/src/views/SecretMainPage/components/SecretListView/SecretDetailSidebar.tsx
```diff
@@ -412,9 +417,38 @@ export const SecretDetailSidebar = ({
...
+                        </Button>
+                      </ModalTrigger>
+                      <ModalContent
+                        title="Secret Reference Tree"
+                        subTitle="Shows the actual value and reference tree"
```
scott-ray-wilson, line 439: How do you feel about something like `Visual breakdown of secrets referenced by this secret`?

frontend/src/views/SecretMainPage/components/SecretListView/SecretDetailSidebar.tsx
```diff
@@ -412,9 +417,38 @@ export const SecretDetailSidebar = ({
...
+                  I={ProjectPermissionActions.Edit}
+                  a={subject(ProjectPermissionSub.Secrets, { environment, secretPath })}
+                >
+                  {(isAllowed) => (
+                    <Modal>
+                      <ModalTrigger asChild>
+                        <Button
+                          variant="outline_bg"
+                          className="w-full px-2 py-1"
+                          leftIcon={<FontAwesomeIcon icon={faRoad} />}
```
scott-ray-wilson, line 431: Thoughts on `faDiagramProject` instead?

frontend/src/views/SecretMainPage/components/SecretListView/SecretDetailSidebar.tsx
```diff
@@ -412,9 +417,38 @@ export const SecretDetailSidebar = ({
...
+                          isDisabled={!isAllowed}
```
scott-ray-wilson, line 432: I think we should also check if this is even a secret that has references in it and disable this button if not, with a tooltip saying `This secret has no references` or something like that

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
```diff
@@ -0,0 +1,124 @@
...
+        <div className="w-full border-b border-gray-800 py-1" />
+      </div>
+    )}
+    {node.children.length > 0 && (
+      <div className="border-l border-gray-600 pl-6 transition-all hover:border-primary-600">
+        {node.children.map((el, index) => (
+          <SecretReferenceNode node={el} key={`node-${node.key}-${index + 1}`} />
+        ))}
+      </div>
+    )}
```
scott-ray-wilson, line 91: I think children should be behind an expandable, so only the root references are shown initially giving the user the ability to expand each descendant instead of all being "expanded" initially

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
```diff
@@ -0,0 +1,124 @@
...
+    <div>
+      <FormControl label="Expanded value">
+        <SecretInput
+          key="value-overriden"
+          isReadOnly
+          value={secretValue}
+          containerClassName="text-bunker-300 hover:border-primary-400/50 border border-mineshaft-600 bg-bunker-800  px-2 py-1.5"
+        />
+      </FormControl>
+      <div className="max-h-80 overflow-auto">
```
scott-ray-wilson, line 119: I don't think we should cap height here, I'd let it fully expand to dialog max (90% vh) and then just scroll the whole content. Also adding a little margin between the value and the children might look cleaner

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
```diff
@@ -0,0 +1,124 @@
...
+          {node.key}
+          <Tooltip content={node.value}>
+            <span
+              className={twMerge(
+                "ml-2 rounded border border-gray-400 px-1 text-xs text-gray-400",
+                !node.value && "border-red-400 text-red-400"
+              )}
+            >
+              {node.value ? "value" : "empty"}
+            </span>
```
scott-ray-wilson, line 43: I think this should be a little more clear that if you hover it shows the value - maybe an eye icon or even `Show Value` or `Reveal Value` - it kind of looks like a badge to me right now - not something that a user would intuitively click/hover over

backend/src/services/secret-v2-bridge/secret-v2-bridge-fns.ts
```diff
@@ -415,14 +422,21 @@ export const expandSecretReferencesFactory = ({
     return secretCache[cacheKey][secretKey] || "";
   };
 
-  const recursivelyExpandSecret = async (dto: { value?: string; secretPath: string; environment: string }) => {
-    if (!dto.value) return "";
+  const recursivelyExpandSecret = async (dto: {
```
scott-ray-wilson, line 425: Not necessarily saying to do so here but don't we need at some level to check permissions for each reference within the secret? Otherwise can't people potentially see a secret value they shouldn't via an expanded value/reference tree?

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1636,6 +1638,77 @@ export const secretV2BridgeServiceFactory = ({
...
+      actorId,
+      projectId,
+      actorAuthMethod,
+      actorOrgId
+    );
+
+    ForbiddenError.from(permission).throwUnlessCan(
+      ProjectPermissionActions.Read,
+      subject(ProjectPermissionSub.Secrets, { environment, secretPath })
+    );
```
scott-ray-wilson, line 1662: Don't we need to check for permissions at each reference along the way in addition to the root secret?

backend/src/services/secret/secret-service.ts
```diff
@@ -1075,6 +1076,18 @@ export const secretServiceFactory = ({
     return secrets;
   };
 
+  const getSecretReferenceTree = async (dto: TGetSecretReferencesTreeDTO) => {
+    const { shouldUseSecretV2Bridge } = await projectBotService.getBotKey(dto.projectId);
+
+    if (!shouldUseSecretV2Bridge)
+      throw new BadRequestError({
+        message: "Project version doesn't not support secret reference tree",
```
scott-ray-wilson, line 1084: type `doesn't not` -> `does not`

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
```diff
@@ -0,0 +1,124 @@
...
+      </div>
+    )}
+  </div>
+);
+
+export const SecretReferenceTree = ({ secretPath, environment, secret }: Props) => {
+  const { currentWorkspace } = useWorkspace();
+  const projectId = currentWorkspace?.id || "";
+
+  const { data } = useGetSecretReferenceTree({
```
scott-ray-wilson, line 99: could disabled query if secret contains no reference values (regex check)

scott-ray-wilson (changes requested):
As a general comment on the PR - this new reference tree addition is awesome but I don't think users should have to navigate to this view to see the expanded secret value. 

This should be visible from the table itself (both overview and main pages), maybe a hover or button as mentioned [here](https://infisical-users.slack.com/archives/C06QSKDKU4S/p1727216814516129)

backend/package-lock.json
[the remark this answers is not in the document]
akhilmhdh: It is

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
akhilmhdh: As i said in PR description. We are seperating it into so that stack trace is not executed on normal operation.  This is avoid the overhead of stack trace in secret normal operations

frontend/src/views/SecretMainPage/components/SecretListView/SecretDetailSidebar.tsx
akhilmhdh: I don't think we really need to make it that much complex to be honest. Like its not like something compute heavy or anything like that.

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
akhilmhdh: I tried this but didn't like the experience. 

First most of references won't be in so many levels. So I don;t think there is much to expand. Second the space is already a bit smaller so i don't want to make interactions inside there

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
akhilmhdh: 90%VH modal is also bad. I think a max cap is fine to a good extend. May be ill increase some more but on a bigger screen 90%VH is awful experience

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
akhilmhdh: https://github.com/Infisical/infisical/pull/2603/files/39013cbeee4e1d013a4e13bc68b4903ce28bf299#diff-7d02b63b99e49c50295fdd39e013c8e9294316950497e86f31e56be915dfa8caR460

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
akhilmhdh: I actually forgot the load condition, I'll add that. Its ok for not disabling on regex condition.

akhilmhdh:
@scott-ray-wilson We don't want to show the entire tree on hover and next issue with hover is what about when they type, and all other cases it will popup. 

Second thing is i don't want to fill any more button on main page rows. It's already too much.

frontend/src/views/SecretMainPage/components/SecretListView/SecretReferenceDetails.tsx
akhilmhdh: Changed to Reveal Value

backend/src/services/secret/secret-service.ts
akhilmhdh: Changed

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
scott-ray-wilson: wouldn't passing a boolean to a single function solve this concern though instead of running it twice? you already have the boolean one layer down, why not just
```  
const expandSecret = async (inputSecret: {
    value?: string;
    skipMultilineEncoding?: boolean | null;
    secretPath: string;
    environment: string;
    shouldStackTrace?: boolean;
  }) => {
    if (!inputSecret.value) return inputSecret.value;

    const shouldExpand = Boolean(inputSecret.value?.match(INTERPOLATION_SYNTAX_REG));
    if (!shouldExpand) return inputSecret.value;

    const { expandedValue } = await recursivelyExpandSecret(inputSecret);

    return inputSecret.skipMultilineEncoding ? formatMultiValueEnv(expandedValue) : expandedValue;
  };```
  
  and only pass `shouldStackTrace` on new endpoint

akhilmhdh: The return signature would be the issue. I would need to make an object. I don't want that when it's not for tracing.

Also its not doing so much work in the second run everything is already cached

scott-ray-wilson: addressed

frontend/src/views/SecretMainPage/components/SecretListView/SecretDetailSidebar.tsx
scott-ray-wilson: I just feel like it might be confusing to to see a reference tree button when no references are present?

akhilmhdh: Ok then let's disable it

scott-ray-wilson: or at least say no references in the dialog when it opens, just feels ambiguous

akhilmhdh:
Update screenshot
Screenshot 2024-10-22 at 11 44 17 PM
Screenshot 2024-10-22 at 11 53 18 PM
Screenshot 2024-10-22 at 11 57 44 PM

scott-ray-wilson (approved):

PR #2606, opened 2024-10-17
feat: multiple auth methods for identities

DanielHougaard:
# Description 📣

This PR implements support for adding multiple auth methods to an identity. It only supports adding different auth methods, you cannot add two of the same auth methods.

This initiative brings no breaking changes.

#### Minor
1. Additionally I've made an improvement to the generate schema script. It now also runs eslint fix after generating the new schemas, so we don't have to manually fix the schemas after generation.

2. Fixed a typo that's been present in the backend for a while. `IdentityAuthMethods.Univarsal` had a typo and improper casing, it's now named `IdentityAuthMethods.UNIVERSAL_AUTH`

## Type ✨

- [ ] Bug fix
- [x] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/db/migrations/20241014084900_identity-multiple-auth-methods.ts
+++ b/backend/src/db/migrations/20241014084900_identity-multiple-auth-methods.ts
@@ -1,0 +1,55 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasAuthMethodColumnAccessToken = await knex.schema.hasColumn(TableName.IdentityAccessToken, "authMethod");
+
+  if (!hasAuthMethodColumnAccessToken) {
+    await knex.schema.alterTable(TableName.IdentityAccessToken, (t) => {
+      t.string("authMethod").nullable();
+    });
+
+    // Backfilling: Update the authMethod column in the IdentityAccessToken table to match the authMethod of the Identity
+    await knex(TableName.IdentityAccessToken).update({
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore because generate schema happens after this
+      authMethod: knex(TableName.Identity)
+        .select("authMethod")
+        .whereRaw(`${TableName.IdentityAccessToken}."identityId" = ${TableName.Identity}.id`)
+        .whereNotNull("authMethod")
+        .first()
+    });
+
+    // ! We delete all access tokens where the identity has no auth method set!
+    // ! Which means un-configured identities that for some reason have access tokens, will have their access tokens deleted.
+    await knex(TableName.IdentityAccessToken)
+      .whereNotExists((queryBuilder) => {
+        void queryBuilder
+          .select("id")
+          .from(TableName.Identity)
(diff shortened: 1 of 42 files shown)


backend/src/db/migrations/20241014084900_identity-multiple-auth-methods.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // ! We delete all access tokens where the identity has no auth method set!
+    // ! Which means un-configured identities that for some reason have access tokens, will have their access tokens deleted.
+    await knex(TableName.IdentityAccessToken)
+      .whereNotExists((queryBuilder) => {
+        void queryBuilder
+          .select("id")
+          .from(TableName.Identity)
+          .whereRaw(`${TableName.IdentityAccessToken}."identityId" = ${TableName.Identity}.id`)
+          .whereNotNull("authMethod");
```
akhilmhdh, line 32: I am bit confused by this `whereNotNull`. Because its with no auth method set right?

akhilmhdh: Also i don't think there will be access token with unconfigured identity. It will have an auth method right to get a token

backend/src/services/identity/identity-fns.ts
```diff
@@ -0,0 +1,29 @@
...
+  uaId?: string;
+  gcpId?: string;
+  awsId?: string;
+  kubernetesId?: string;
+  oidcId?: string;
+  azureId?: string;
+  tokenId?: string;
+}) => {
+  return [
+    ...(uaId ? [IdentityAuthMethod.UNIVERSAL_AUTH] : []),
```
akhilmhdh, line 21: Or you could just do it like this

```
[ uaId?IdentityAuthMethod.UNIVERSAL_AUTH:null, .rest of them ].filter(<as below>)
```

frontend/src/views/Org/MembersPage/components/OrgIdentityTab/components/IdentitySection/IdentityAuthMethodModal.tsx
```diff
@@ -86,187 +98,123 @@ export const IdentityAuthMethodModal = ({ popUp, handlePopUpOpen, handlePopUpTog
...
   };
 
+  const isSelectedAuthAlreadyConfigured =
+    identityAuthMethodData?.configuredAuthMethods?.includes(watchedAuthMethod);
+
   useEffect(() => {
-    if (identityAuthMethodData?.authMethod) {
-      setValue("authMethod", identityAuthMethodData.authMethod);
-      return;
+    if (popUp?.identityAuthMethod?.data?.authMethod) {
```
akhilmhdh, line 124: If you keep the modal outside. That is `<Modal>...` wrapping outside this will unmount the component. Thus you won't need this useEffect at all. Actually both useEffect can then be removed

akhilmhdh (changes requested):
Small concerns and changes. Need to application test it

akhilmhdh (approved):

maidul98 (approved):
lgtm

PR #2620, opened 2024-10-18
feat: automatic k8 operator token refreshing

DanielHougaard:
# Description 📣

This PR introduces automatic token refreshing, which is handled internally by our Go Lang SDK.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/k8-operator/controllers/infisicalsecret_helper.go
+++ b/k8-operator/controllers/infisicalsecret_helper.go
@@ -294,8 +294,9 @@ func (r *InfisicalSecretReconciler) GetResourceVariables(infisicalSecret v1alpha
 	if _, ok := resourceVariablesMap[string(infisicalSecret.UID)]; !ok {
 
 		client := infisicalSdk.NewInfisicalClient(infisicalSdk.Config{
-			SiteUrl:   api.API_HOST_URL,
-			UserAgent: api.USER_AGENT_NAME,
+			SilentMode: true,
+			SiteUrl:    api.API_HOST_URL,
+			UserAgent:  api.USER_AGENT_NAME,
 		})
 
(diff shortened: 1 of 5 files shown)


k8-operator/controllers/infisicalsecret_helper.go
```diff
@@ -294,8 +294,9 @@ func (r *InfisicalSecretReconciler) GetResourceVariables(infisicalSecret v1alpha
 	if _, ok := resourceVariablesMap[string(infisicalSecret.UID)]; !ok {
 
 		client := infisicalSdk.NewInfisicalClient(infisicalSdk.Config{
-			SiteUrl:   api.API_HOST_URL,
-			UserAgent: api.USER_AGENT_NAME,
+			SilentMode: true,
```
maidul98, line 297: Why silent? Wouldn't it be good to see these errors for the end human operator?

maidul98 (commented):
Feature seems to be working as expected but it keeps saying it wasn't able to renew periodically even though in the next iteration, it can. When we know the ttl will exceed the max ttl, we can just login instead of using the expired creds. 

CleanShot 2024-10-22 at 23 18 49@2x


Here some  other comments:
- I don't think we need to paint the logs yellow
- The logs say that it wasn't able to renew but i think it  does? The user is never made aware that it has renewed successfully after though 
- I think our approach with logging might be incorrect for the operator. When there are are more than 2 InfisicalSecret CRD, these logs aren't useful anymore because they don't contain any context for which CRD they are for. I would had some context to the SDK logs for now. I'll make a linear item to revamp our operator logs in a future PR. 
CleanShot 2024-10-22 at 23 08 41@2x

maidul98 (approved):
Note: ask updates applied. lgtm

PR #2652, opened 2024-10-28
feat: add support for custom ca in k8 operator

sheensantoscapadngan:
# Description 📣
- This PR adds support for using custom CA certificates in k8 operator which will allow connections to self-hosted Infisical instances with private/internal certificates

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


sheensantoscapadngan:
PR is dependent on https://github.com/Infisical/go-sdk/pull/18. need to update the sdk version here once we've published a new Go SDK version

k8-operator/controllers/infisicalsecret_controller.go
```diff
@@ -107,6 +107,18 @@ func (r *InfisicalSecretReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		api.API_HOST_URL = infisicalSecretCR.Spec.HostAPI
 	}
 
+	if infisicalSecretCR.Spec.TLS.CaRef.SecretName != "" {
+		api.API_CA_CERTIFICATE, err = r.GetInfisicalCaCertificateFromKubeSecret(ctx, infisicalSecretCR)
```
maidul98, line 111: How will this work if you have multiple infisical instances (staging, prod, etc) and you use the single cluster? It looks like this will be the same api.API_CA_CERTIFICATE for all Infisical CRDs?

maidul98 (approved):

k8-operator/controllers/infisicalsecret_controller.go
sheensantoscapadngan: if you have multiple infisical instances, you would need multiple `InfisicalSecret` CRDs with different configs for the `HOST_URL` and `CA`.

the reconcile loop runs independently between different InfisicalSecrets so there should be no issue here

maidul98: >  runs independently between different InfisicalSecrets

Hmm are you sure? This would then mean we can get rid of this pattern https://github.com/Infisical/infisical/blob/main/k8-operator/controllers/infisicalsecret_controller.go#L49

maidul98 (approved):
lgtm

PR #2670, opened 2024-10-31
feat: added key id column

sheensantoscapadngan:
# Description 📣
- This PR adds the Key ID column so that it can be extracted easily for use in API or in kubernetes KMS plugin (design inspo from aws KMS)

Before:
image

After:
image

This PR also adds kubernetes encryption to the docs
image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


scott-ray-wilson (changes requested):
Looks great; since this is for ease of extraction maybe we should add a copy-to-clipboard button next to it?

scott-ray-wilson (approved):
LGTM!

PR #2671, opened 2024-10-31
feat: hardware security module's support

DanielHougaard:
# Description 📣

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/lib/config/env.ts
+++ b/backend/src/lib/config/env.ts
@@ -175,7 +175,7 @@ const envSchema = z
         .string()
         .optional()
         .transform((val) => {
-          if (process.env.NODE_ENV === "development") return "/usr/local/lib/softhsm/libsofthsm2.so";
+          // if (process.env.NODE_ENV === "development") return "/usr/local/lib/softhsm/libsofthsm2.so";
           return val;
         })
     ),
@@ -200,6 +200,11 @@ const envSchema = z
     HSM_SLOT: z.coerce.number().optional().default(0),
     HSM_MECHANISM: zpStr(z.string().optional().default("AES_GCM"))
   })
+  // To ensure that basic encryption is always possible.
+  .refine(
+    (data) => data.ENCRYPTION_KEY != null || data.ROOT_ENCRYPTION_KEY != null,
+    "Either ENCRYPTION_KEY or ROOT_ENCRYPTION_KEY must be defined."
+  )
   .transform((data) => ({
     ...data,
 
--- a/backend/src/lib/config/env.ts
+++ b/backend/src/lib/config/env.ts
@@ -162,7 +162,38 @@ const envSchema = z
     DISABLE_AUDIT_LOG_GENERATION: zodStrBool.default("false"),
     SSL_CLIENT_CERTIFICATE_HEADER_KEY: zpStr(z.string().optional()).default("x-ssl-client-cert"),
     WORKFLOW_SLACK_CLIENT_ID: zpStr(z.string().optional()),
-    WORKFLOW_SLACK_CLIENT_SECRET: zpStr(z.string().optional())
+    WORKFLOW_SLACK_CLIENT_SECRET: zpStr(z.string().optional()),
+
+    // HSM
+    HSM_LIB_PATH: zpStr(
+      z
+        .string()
+        .optional()
+        .transform((val) => {
+          if (process.env.NODE_ENV === "development") return "/usr/local/lib/softhsm/libsofthsm2.so";
+          return val;
+        })
+    ),
+    HSM_PIN: zpStr(
+      z
+        .string()
+        .optional()
+        .transform((val) => {
+          if (process.env.NODE_ENV === "development") return "1234";
+          return val;
+        })
+    ),
+    HSM_KEY_LABEL: zpStr(
+      z
+        .string()
+        .optional()
+        .transform((val) => {
+          if (process.env.NODE_ENV === "development") return "auth-app";
+          return val;
+        })
+    ),
+    HSM_SLOT: z.coerce.number().optional().default(0),
+    HSM_MECHANISM: zpStr(z.string().optional().default("AES_GCM"))
   })
   .transform((data) => ({
     ...data,
@@ -174,10 +205,18 @@ const envSchema = z
     isRedisConfigured: Boolean(data.REDIS_URL),
     isDevelopmentMode: data.NODE_ENV === "development",
     isProductionMode: data.NODE_ENV === "production" || IS_PACKAGED,
+
     isSecretScanningConfigured:
       Boolean(data.SECRET_SCANNING_GIT_APP_ID) &&
       Boolean(data.SECRET_SCANNING_PRIVATE_KEY) &&
       Boolean(data.SECRET_SCANNING_WEBHOOK_SECRET),
+    isHsmConfigured:
+      Boolean(data.HSM_LIB_PATH) &&
+      Boolean(data.HSM_PIN) &&
+      Boolean(data.HSM_KEY_LABEL) &&
+      Boolean(data.HSM_MECHANISM) &&
+      data.HSM_SLOT !== undefined,
+
     samlDefaultOrgSlug: data.DEFAULT_SAML_ORG_SLUG,
     SECRET_SCANNING_ORG_WHITELIST: data.SECRET_SCANNING_ORG_WHITELIST?.split(",")
   }));
--- a/backend/src/lib/config/env.ts
+++ b/backend/src/lib/config/env.ts
@@ -1,6 +1,8 @@
 import { Logger } from "pino";
 import { z } from "zod";
 
+import { RootKeyEncryptionStrategy } from "@app/services/kms/kms-types";
+
 import { removeTrailingSlash } from "../fn";
 import { zpStr } from "../zod";
 
@@ -164,6 +166,9 @@ const envSchema = z
     WORKFLOW_SLACK_CLIENT_ID: zpStr(z.string().optional()),
     WORKFLOW_SLACK_CLIENT_SECRET: zpStr(z.string().optional()),
 
+    // KMS ENCRYPTION
+    ROOT_KEY_ENCRYPTION_STRATEGY: z.nativeEnum(RootKeyEncryptionStrategy).default(RootKeyEncryptionStrategy.Basic),
+
     // HSM
     HSM_LIB_PATH: zpStr(
       z
@@ -197,6 +202,9 @@ const envSchema = z
   })
   .transform((data) => ({
     ...data,
+
+    // ROOT_KEY_ENCRYPTION_STRATEGY: "HSM",
+
     DB_READ_REPLICAS: data.DB_READ_REPLICAS
       ? databaseReadReplicaSchema.parse(JSON.parse(data.DB_READ_REPLICAS))
       : undefined,
--- a/backend/src/main.ts
+++ b/backend/src/main.ts
@@ -9,6 +9,7 @@ import { initLogger } from "./lib/logger";
 import { queueServiceFactory } from "./queue";
 import { main } from "./server/app";
 import { bootstrapCheck } from "./server/boot-strap-check";
+import { initializePkcs11Module } from "./services/hsm/hsm-fns";
 import { smtpServiceFactory } from "./services/smtp/smtp-service";
 
 dotenv.config();
@@ -53,13 +54,17 @@ const run = async () => {
   const queue = queueServiceFactory(appCfg.REDIS_URL);
   const keyStore = keyStoreFactory(appCfg.REDIS_URL);
 
-  const server = await main({ db, auditLogDb, smtp, logger, queue, keyStore });
+  const pkcs11Module = initializePkcs11Module();
+  pkcs11Module.initialize();
+
+  const server = await main({ db, auditLogDb, hsmModule: pkcs11Module.getModule(), smtp, logger, queue, keyStore });
   const bootstrap = await bootstrapCheck({ db });
 
   // eslint-disable-next-line
   process.on("SIGINT", async () => {
     await server.close();
     await db.destroy();
+    pkcs11Module.finalize();
     process.exit(0);
   });
 
@@ -67,6 +72,7 @@ const run = async () => {
   process.on("SIGTERM", async () => {
     await server.close();
     await db.destroy();
+    pkcs11Module.finalize();
     process.exit(0);
   });
 
--- a/backend/src/services/hsm/hsm-fns.ts
+++ b/backend/src/services/hsm/hsm-fns.ts
@@ -1,0 +1,40 @@
+import * as grapheneLib from "graphene-pk11";
+
+import { getConfig } from "@app/lib/config/env";
+import { logger } from "@app/lib/logger";
+
+export type HsmModule = {
+  module: grapheneLib.Module | null;
+  graphene: typeof grapheneLib;
+};
+
+export const initializePkcs11Module = () => {
+  const appCfg = getConfig();
+
+  let module: grapheneLib.Module | null = null;
+
+  const initialize = () => {
+    if (!appCfg.isHsmConfigured) {
+      return;
+    }
+
+    module = grapheneLib.Module.load(appCfg.HSM_LIB_PATH!, "SoftHSM");
+    module.initialize();
+    logger.info("PKCS#11 module initialized");
+  };
+
+  const finalize = () => {
+    if (module) {
+      module.finalize();
+      logger.info("PKCS#11 module finalized");
+    }
+  };
+
+  const getModule = (): HsmModule => ({ module, graphene: grapheneLib });
+
+  return {
+    initialize,
+    finalize,
+    getModule
+  };
+};
--- a/backend/src/services/hsm/hsm-service.ts
+++ b/backend/src/services/hsm/hsm-service.ts
@@ -152,7 +152,7 @@ export const hsmServiceFactory = ({ pkcs11Module: { module, graphene } }: THsmSe
     }
   };
 
-  const isActive = async () => {
+  const isActive = () => {
     if (!module || !appCfg.isHsmConfigured || !sessionManager) {
       return false;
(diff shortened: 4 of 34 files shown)


backend/src/lib/config/env.ts
```diff
@@ -162,10 +162,47 @@ const envSchema = z
...
+          if (process.env.NODE_ENV === "development") return "auth-app";
+          return val;
+        })
+    ),
+    HSM_SLOT: z.coerce.number().optional().default(0),
+    HSM_MECHANISM: zpStr(z.string().optional().default("AES_GCM"))
   })
+  // To ensure that basic encryption is always possible.
+  .refine(
+    (data) => data.ENCRYPTION_KEY != null || data.ROOT_ENCRYPTION_KEY != null,
```
akhilmhdh, line 200: It could be also undefined right?

backend/src/main.ts
```diff
@@ -53,20 +54,25 @@ const run = async () => {
   const queue = queueServiceFactory(appCfg.REDIS_URL);
   const keyStore = keyStoreFactory(appCfg.REDIS_URL);
 
-  const server = await main({ db, auditLogDb, smtp, logger, queue, keyStore });
+  const pkcs11Module = initializePkcs11Module();
```
akhilmhdh, line 57: If its used and passed around as hsmModule why not call it like that? Felt confusing

backend/src/main.ts
```diff
@@ -53,20 +54,25 @@ const run = async () => {
...
+  pkcs11Module.initialize();
+
+  const server = await main({ db, auditLogDb, hsmModule: pkcs11Module.getModule(), smtp, logger, queue, keyStore });
   const bootstrap = await bootstrapCheck({ db });
```
akhilmhdh, line 61: Can we also do the HSM is working check in bootstrap as well.

Like if HSM is not configured it could say skipping HSM check due to not configured and if configured does a connection check.

backend/src/services/hsm/hsm-fns.ts
```diff
@@ -0,0 +1,40 @@
+import * as grapheneLib from "graphene-pk11";
+
+import { getConfig } from "@app/lib/config/env";
+import { logger } from "@app/lib/logger";
+
+export type HsmModule = {
+  module: grapheneLib.Module | null;
```
akhilmhdh, line 7: I think this should be moved out of hsm-fns and keep it in ``hsm-module` or something. 

I was searching for this file as hsm-module 😅

backend/src/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,379 @@
...
+const USER_ALREADY_LOGGED_IN_ERROR = "CKR_USER_ALREADY_LOGGED_IN";
+
+export type THsmServiceFactory = ReturnType<typeof hsmServiceFactory>;
+
+enum RequiredMechanisms {
+  AesGcm = "AES_GCM",
+  AesKeyWrap = "AES_KEY_WRAP"
+}
+
+class HsmSessionManager {
```
akhilmhdh, line 21: Why did we switch to class all of a sudden?

backend/src/services/kms/kms-service.ts
```diff
@@ -813,31 +862,74 @@ export const kmsServiceFactory = ({
...
     if (kmsRootConfig) {
       if (lock) await lock.release();
       logger.info("KMS: Encrypted ROOT Key found from DB. Decrypting.");
-      const decryptedRootKey = cipher.decrypt(kmsRootConfig.encryptedRootKey, encryptionKeyBuffer);
-      // set the flag so that other instancen nodes can start
+
+      const decryptedRootKey = await $decryptRootKey(kmsRootConfig).catch((err) => {
+        logger.error(err, `KMS: Failed to decrypt ROOT Key [strategy=${kmsRootConfig.encryptionStrategy}]`);
+        // We do not want to throw on startup. If the HSM has issues, this will throw an error, causing the entire API to shut down.
+        // If the API shuts down, the user will have no way to do recovery by importing their backup decryption key and rolling back to basic encryption.
```
akhilmhdh, line 874: @DanielHougaard I think this is not needed anymore right? As we don't plan to do backup of HSM at the moment

backend/src/services/kms/kms-service.ts
```diff
@@ -813,31 +862,74 @@ export const kmsServiceFactory = ({
...
+      logger.error("KMS: Failed to re-encrypt ROOT Key with selected strategy");
+      throw new Error("Failed to re-encrypt ROOT Key with selected strategy");
+    }
+
+    await kmsRootConfigDAL.updateById(KMS_ROOT_CONFIG_UUID, {
+      encryptedRootKey,
+      encryptionStrategy: strategy
+    });
+
+    ROOT_ENCRYPTION_KEY = decryptedRootKey;
```
akhilmhdh, line 930: What about other instance on horizontal scaling?

akhilmhdh: Also another question is: I see when the user picks the strategy its switched right?

But I assume there is some problems with our assumption:
1. The database does get the actual root encryption once with know encryption key once. Something to be considered good or bad?
2. Second thing is this in turn is recovery also right? They could save the key in some other place in DB for the time being then switch back to HSM?

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -288,12 +293,72 @@ export const superAdminServiceFactory = ({
...
+    }
+
+    const selectedStrategy = kmsRootCfg.encryptionStrategy;
+    const enabledStrategies: { enabled: boolean; strategy: RootKeyEncryptionStrategy; name: string }[] = [];
+
+    if (appCfg.ROOT_ENCRYPTION_KEY || appCfg.ENCRYPTION_KEY) {
+      const basicStrategy = RootKeyEncryptionStrategy.Basic;
+
+      enabledStrategies.push({
+        name: "Regular Encryption",
```
akhilmhdh, line 312: Better to say something like SoftKMS

frontend/src/hooks/useFileDownload.tsx
```diff
@@ -0,0 +1,13 @@
+import { useCallback } from "react";
+
+export const useFileDownload = () => {
```
akhilmhdh, line 3: We have a lib for this called `file-saver` something like this. This is because file downloading is a bit complex due to browser compatibility

package.json
```diff
@@ -24,6 +24,7 @@
 		"husky": "^8.0.3"
 	},
 	"dependencies": {
-		"@radix-ui/react-radio-group": "^1.1.3"
+		"@radix-ui/react-radio-group": "^1.1.3",
+		"secrets.js-grempe": "^2.0.0"
```
akhilmhdh, line 28: Are we using this?

akhilmhdh (changes requested):
Pending: Application testing and need to read about HSM slot mechanism more

backend/Dockerfile
```diff
@@ -3,6 +3,17 @@ FROM node:20-alpine AS build
 
 WORKDIR /app
 
+RUN apk --update add \
```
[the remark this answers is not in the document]
akhilmhdh: Actually i don't think use thing in production right?

frontend/src/hooks/useFileDownload.tsx
DanielHougaard: Ah this thing is not used at all anymore, it was a leftover from the removal of recovery

backend/src/main.ts
DanielHougaard: We do this in the `startService` handler for the HSM. I think this makes sense as-is, because if the startService function is run, we want to make sure all the cryptographic functions are available before attempting to use the HSM in the KMS `startService` handler

backend/Dockerfile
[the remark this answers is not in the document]
DanielHougaard: @akhilmhdh They are all required for the `pkcs11js` library, which is used by `graphene-pk11`

backend/src/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,349 @@
...
+
+      if (!session) {
+        throw new Error("Session not initialized");
+      }
+
+      const randomData = session.generateRandom(256);
+      const encryptedData = await encrypt(Buffer.from(randomData), session);
+      const decryptedData = await decrypt(encryptedData, session);
+
+      if (Buffer.from(randomData).toString("hex") !== Buffer.from(decryptedData).toString("hex")) {
```
dangtony98, line 262: Nit: I think can use `Buffer.equals()` to compare the buffers instead of needing to do it via conversion to hex.

backend/src/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,349 @@
...
+          });
+          logger.info(`Master key created successfully with label: ${appCfg.HSM_KEY_LABEL}`);
+        }
+
+        // Verify HSM supports required mechanisms
+        const mechs = session.slot.getMechanisms();
+        const mechNames: string[] = [];
+
+        // eslint-disable-next-line no-plusplus
+        for (let i = 0; i < mechs.length; i++) {
```
dangtony98, line 316: Nit: Could've used `.map()` on `mechNames` would be cleaner.

backend/src/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,349 @@
...
+    }
+  }
+};
+
+// eslint-disable-next-line no-empty-pattern
+export const hsmServiceFactory = ({ hsmModule: { module, graphene } }: THsmServiceFactoryDep) => {
+  const appCfg = getConfig();
+
+  // Constants for buffer structure
+  const IV_LENGTH = 16;
```
dangtony98, line 91: Can you double check if `IV_LENGTH=12` would work here? I believe NIST recommends 96-bit IV / nonces for AES-256-GCM.

backend/src/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,349 @@
...
+    const key = session.find(template).items(0);
+
+    if (!key) {
+      throw new Error("Failed to find master key");
+    }
+
+    return key;
+  };
+
+  const $generateAndWrapKey = (session: grapheneLib.Session) => {
```
dangtony98, line 111: I would add more comments across this file (perhaps above each function) to describe all the wrapping logic; it's not easily understandable for future developers who may have to work with this.

backend/src/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,349 @@
...
+  const decrypt: {
+    (encryptedBlob: Buffer, providedSession: grapheneLib.Session): Promise<Buffer>;
+    (encryptedBlob: Buffer): Promise<Buffer>;
+  } = async (encryptedBlob: Buffer, providedSession?: grapheneLib.Session) => {
+    if (!module) {
+      throw new Error("HSM service not initialized");
+    }
+
+    const $performDecryption = (s: grapheneLib.Session) => {
+      const wrappedKey = encryptedBlob.subarray(0, WRAPPED_KEY_LENGTH);
```
dangtony98, line 220: Just to clarify my understanding, is `wrappedKey` included together in `encryptedBlob` because they are stored together in the database? (i.e. when encrypting root key with HSM then the resulting ciphertext contains the encrypted root key, IV, and wrapped key so that when decrypting it we unwrap it with the master key in HSM and use that to decrypt the root key).

Is that correct? 

Also I'm curious about the need for this wrapping key — Can you point me to some resources discussing this as a best practice etc.?

backend/src/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    // Create new session
+    const slot = module.getSlots(appCfg.HSM_SLOT);
+    // eslint-disable-next-line no-bitwise
+    if (!(slot.flags & graphene.SlotFlag.TOKEN_PRESENT)) {
+      throw new Error("Slot is not initialized");
+    }
+
+    for (let i = 0; i < 10; i += 1) {
```
dangtony98, line 39: Why is this for-loop and subsequent promise / sleep needed again? I can't imagine there's no cleaner way to do this?

dangtony98 (changes requested):

backend/src/ee/services/license/license-fns.ts
```diff
@@ -29,6 +29,7 @@ export const getDefaultOnPremFeatures = (): TFeatureSet => ({
   auditLogStreams: false,
   auditLogStreamLimit: 3,
   samlSSO: false,
+  hsm: false,
```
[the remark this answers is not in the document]
DanielHougaard: I think `hsm` is fine for now, since we'll probably add more features to it later, so it's not just seal

backend/src/services/hsm/hsm-service.ts
DanielHougaard: No. `MechanismCollection` doesn't implement an iterator.

backend/src/services/hsm/hsm-service.ts
DanielHougaard: There can only be one session open per slot. Let's say that the user uses horizontal scaling for their instance, and 10 instances start up at the same time. They will all attempt to open a session at the same time, which will throw an error for most of those instances, because a session will already be open.

To fix this, we wait and try to get the session, I'm open for suggestions on how you'd improve this, since there's no native way to "listen" for open sessions.

backend/src/services/hsm/hsm-service.ts
DanielHougaard: Your understand is correct. To unwrap the wrapped key, you need the HSM master key. This is to limit exposure / usage of the master HSM key as much as possible.

https://cryptography.io/en/latest/hazmat/primitives/keywrap/
https://security.stackexchange.com/questions/270194/what-is-the-point-of-the-aes-key-wrap-algorithm

You can also think of the wrapping mechanism as a form of envelope encryption.

backend/src/ee/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,472 @@
...
+        };
+
+        pkcs11.C_EncryptInit(sessionHandle, encryptMechanism, aesKey);
+
+        // Calculate max buffer size (input length + potential full block of padding)
+        const maxEncryptedLength = Math.ceil(data.length / BLOCK_SIZE) * BLOCK_SIZE + BLOCK_SIZE;
+        const tempBuffer = Buffer.alloc(maxEncryptedLength);
+
+        // First call to get the actual length
+        const encryptedLength = pkcs11.C_Encrypt(sessionHandle, data, tempBuffer);
```
dangtony98, line 221: I'm not sure I understand this line.

Specifically, I don't think the variable should be called `encryptedLength` if the result of `pkcs11.C_Encrypt()` is actually the encrypted data?

backend/src/ee/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,472 @@
...
+
+        // Create a copy of the encrypted data using the actual length
+        const encryptedData = Buffer.from(tempBuffer.subarray(0, encryptedLength.length || 16));
```
dangtony98, line 224: Building off my previous comment, why is this line necessary? Isn't the encrypted data just the result of `pkcs11.C_Encrypt()`?

Put differently, why do you need to create a copy of the data like so with the subarray? If a copy is needed, why not just do `const encryptedDataCopy = Buffer.from(encryptedData);`?

backend/src/ee/services/hsm/hsm-service.ts
```diff
@@ -0,0 +1,472 @@
...
+      }
+
+      if (appCfg.HSM_SLOT >= slots.length) {
+        throw new Error(`HSM slot ${appCfg.HSM_SLOT} not found or not initialized`);
+      }
+
+      const slotId = slots[appCfg.HSM_SLOT];
+
+      const startTime = Date.now();
+      while (Date.now() - startTime < MAX_TIMEOUT) {
```
dangtony98, line 70: I thought we discussed some kind of locking mechanism here with Redis potentially to avoid relying on such `MAX_TIMEOUT` which can be faulty?

We probably want something that always works instead of works in most cases. Have you discussed this with @akhilmhdh?

dangtony98 (changes requested):

backend/src/ee/services/hsm/hsm-service.ts
DanielHougaard: Spoke to Akhi about it. There's no point doing the Redis lock in this case

backend/src/services/kms/kms-service.ts
```diff
@@ -610,6 +612,65 @@ export const kmsServiceFactory = ({
     }
   };
 
+  const $getBasicEncryptionKey = () => {
+    const appCfg = getConfig();
+
+    const encryptionKey = appCfg.ENCRYPTION_KEY || appCfg.ROOT_ENCRYPTION_KEY;
+    const isBase64 = !appCfg.ENCRYPTION_KEY;
```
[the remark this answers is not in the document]
DanielHougaard: I didn't create this logic, I just moved it around a bit. For now I'll leave this as-is.

backend/src/ee/services/hsm/hsm-service.ts
DanielHougaard: You're right, this was a leftover from trying to fix a invalid data issue. Resolved

backend/src/ee/services/hsm/hsm-service.ts
DanielHougaard: In the raw C implementation, the encrypted data is copied to the buffer through a reference. This also works in javascript, and I followed the docs very closely when building this implementation. The more clear approach would be to just use the returned value directly. Good catch, resolved!

dangtony98 (dismissed):

akhilmhdh (dismissed):

PR #2696, opened 2024-11-06
feat: added queue level debounce for secret sync and removed stale check

akhilmhdh:
# Description 📣

This PR resolves a race condition in stale check that was happening. We now have a debounce in queue level with 3second buffer to handle this corner case. I have also made get integration list after the lock acquire to work with consistent data.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98 (approved):
lgtm

PR #2697, opened 2024-11-06
Enhance types in the SecretOverviewPage.tsx file

asifrahaman13:
# Description 📣

The SecretOverPage.tsx file with the following path

```frontend/src/views/SecretOverviewPage/SecretOverviewPage.tsx```

Has several type hints missing. Added types to different parts of the code to make it better readable in terms of typescript type checks

## Type ✨

Improvement


asifrahaman13:
@maidul98 

Can you look into this PR

PR #2721, opened 2024-11-11
feat: move folders

DanielHougaard:
# Description 📣

This PR aims to add API support for moving folders around. Previously it was only possible to update the folder name, but not the location of the folder itself. We are simply changing the parent ID of a folder, not creating new folders. This means all the data inside the folder (such as secrets, or other folders) will remain intact.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---


akhilmhdh:
@DanielHougaard What about rollback? Did you check that.

PR #2726, opened 2024-11-12
Feat: Paste Secrets for Upload

scott-ray-wilson:
# Description 📣

This PR enables users to paste secret values for upload instead of requiring them to select a file. Pasting supports .json, .env and .yml format.

Also includes fix for conflicting secret keys preventing upload if not present on current page.

## Type ✨

- [x] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -840,4 +840,91 @@ export const registerDashboardRouter = async (server: FastifyZodProvider) => {
...
+      security: [
+        {
+          bearerAuth: []
+        }
+      ],
+      querystring: z.object({
+        projectId: z.string().trim(),
+        environment: z.string().trim(),
+        secretPath: z.string().trim().default("/").transform(removeTrailingSlash),
+        keys: z.string().trim().transform(decodeURIComponent)
```
DanielHougaard, line 860: Is the `decodeURIComponent` transformer really necessary? You'd think this is something Axios or Fastify would handle automatically

frontend/src/components/utilities/parseDotEnv.ts
```diff
@@ -6,7 +6,7 @@ const LINE =
  * @param {ArrayBuffer} src - source buffer
  * @returns {String} text - text of buffer
  */
-export function parseDotEnv(src: ArrayBuffer) {
```
DanielHougaard, line 9: I would move the parseJson function into the same file as the parseDotEnv function, since their use-case is very much alike. 

I suggest creating a single filed called `parseEnv.ts`.

frontend/src/views/SecretMainPage/components/SecretDropzone/PasteSecretEnvModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+          >
+            {(isAllowed) => (
+              <Button
+                leftIcon={<FontAwesomeIcon icon={faPaste} />}
+                onClick={() => onToggle(true)}
+                isDisabled={!isAllowed}
+                variant="star"
+                size={isSmaller ? "xs" : "sm"}
+              >
+                Paste Secret Values
```
DanielHougaard, line 114: They aren't pasting just secret values, they're also pasting secret keys. How about we say something like `Paste New Secrets`, or something along the lines of that?

frontend/src/components/utilities/parseJson.ts
```diff
@@ -0,0 +1,11 @@
+export const parseJson = (src: ArrayBuffer | string) => {
+  const file = src.toString();
+  const formatedData: Record<string, string> = JSON.parse(file);
+  const env: Record<string, { value: string; comments: string[] }> = {};
+  Object.keys(formatedData).forEach((key) => {
+    if (typeof formatedData[key] === "string") {
+      env[key] = { value: formatedData[key], comments: [] };
```
DanielHougaard, line 7: Any reason we don't support comments? We could add support for it by allowing the following input:

```json
{
   "KEY": "value" // We already support this
}
```

And

```json
{
   "KEY": {
      "value": "the value",
      "comment": "some comment"
   }
}
```

This would make the .yml / .json functionality more uniform.

DanielHougaard (changes requested):
This works very nicely! I left a few comments but mostly smaller stuff.

I think this would be great to document. At first I was questioning the format of the JSON  format, until I looked at the code

frontend/src/views/SecretMainPage/components/SecretDropzone/PasteSecretEnvModal.tsx
scott-ray-wilson: good point!

frontend/src/components/utilities/parseJson.ts
scott-ray-wilson: json itself actually doesn't support comments; it's data only - I didn't no this for awhile actually. Javascript objects deceive you into thinking it can haha

frontend/src/views/SecretMainPage/components/SecretDropzone/PasteSecretEnvModal.tsx
scott-ray-wilson: going with simply `Paste Secrets` since it can have overwriting values

frontend/src/components/utilities/parseJson.ts
scott-ray-wilson: 2nd option is viable but I'd say lets wait to see if someone requests that?

backend/src/server/routes/v1/dashboard-router.ts
scott-ray-wilson: Just double checked but yeah; I think if they did it by default they could unintentionally mutate data that wasn't encoded but had `%20` for example

frontend/src/components/utilities/parseJson.ts
DanielHougaard: > json itself actually doesn't support comments; it's data only - I didn't know this for awhile, Javascript objects deceive you into thinking it can haha

Yeah it was just for illustration I added that


> 2nd option is viable but I'd say lets wait to see if someone requests that?
Sure, we can do that. Works for now

DanielHougaard (approved):

DanielHougaard (approved):

PR #2742, opened 2024-11-15
feat: TOTP dynamic secret provider

sheensantoscapadngan:
# Description 📣
This adds support for new dynamic secret provider - TOTP + documentation

Tested with Bitbucket and our very own TOTP

image

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---


scott-ray-wilson (changes requested):
Only one minor comment on the current code but a couple of thoughts:

1) as a user, if I leased a TOTP dynamic secret I would expect the code to be valid for the duration of the lease; I think it's confusing I can have a valid lease but invalid code - maybe we could disable max/default TTL for TOTP and disable renew/revoke, and configure based of the providers delta and use the `authenticator.timeRemaining` to display expiry?

2) I had to use a QR code scanner to get the URL - I'm not sure the requested use case for this feature but I think most users would expect to be able to enter the key displayed on 2fa setup pages such as bitbucket or google as an alternative?

backend/src/ee/services/dynamic-secret/providers/totp.ts
```diff
@@ -4,20 +4,12 @@ import { HashAlgorithms } from "otplib/core";
...
     const providerInputs = await DynamicSecretTotpSchema.parseAsync(inputs);
 
-    const urlObj = new URL(providerInputs.url);
-    const secret = urlObj.searchParams.get("secret");
-    if (!secret) {
-      throw new BadRequestError({
-        message: "TOTP secret is missing from URL"
-      });
-    }
-
```
scott-ray-wilson: Can't we keep this validation for URL type based on the config type and add validation for the manual type? Would be helpful to catch these for users

backend/src/ee/services/dynamic-secret/providers/models.ts
```diff
@@ -221,9 +232,29 @@ export const LdapSchema = z.union([
...
+      .refine((val) => {
+        const urlObj = new URL(val);
+        const secret = urlObj.searchParams.get("secret");
+
+        return Boolean(secret);
+      }, "OTP URL must contain secret field")
+  }),
+  z.object({
+    configType: z.literal(TotpConfigType.MANUAL),
+    secret: z.string().min(1),
```
scott-ray-wilson, line 252: Not sure if ideal to do here but somewhere use `transform` to `replaceAll(" ", "")` to not require users to manually remove spaces from the key?

frontend/src/views/SecretMainPage/components/ActionBar/CreateDynamicSecretForm/TotpInputForm.tsx
```diff
@@ -113,19 +153,142 @@ export const TotpInputForm = ({
...
+                        label="Secret Key"
+                        className="flex-grow"
+                        isError={Boolean(error?.message)}
+                        errorText={error?.message}
+                      >
+                        <Input {...field} />
+                      </FormControl>
+                    )}
+                  />
+                  <div className="flex flex-row">
```
scott-ray-wilson, line 217: I think we should add some space between these fields with `gap-3` or some amount

frontend/src/views/SecretMainPage/components/DynamicSecretListView/EditDynamicSecretForm/EditDynamicSecretTotpForm.tsx
```diff
@@ -126,19 +157,142 @@ export const EditDynamicSecretTotpForm = ({
...
+                        label="Secret Key"
+                        className="flex-grow"
+                        isError={Boolean(error?.message)}
+                        errorText={error?.message}
+                      >
+                        <Input {...field} />
+                      </FormControl>
+                    )}
+                  />
+                  <div className="flex flex-row">
```
scott-ray-wilson, line 221: same here; `gap-3` or some amount

docs/documentation/platform/dynamic-secrets/totp.mdx
```diff
@@ -22,23 +22,34 @@ The Infisical TOTP dynamic secret allows you to generate time-based one-time pas
...
+    <ParamField path="Secret Key" type="string">
+      Base32 encoded secret used to generate TOTP codes.
+    </ParamField>
+    <ParamField path="Period" type="number">
+      Time interval in seconds between generating new TOTP codes.
+    </ParamField>
+    <ParamField path="Digits" type="number" required>
+      Number of digits to generate in each TOTP code.
+    </ParamField>
+    <ParamField path="Algorithm" type="string" required>
```
scott-ray-wilson, line 47: Based on the schema above I don't think digits and algorithm should be `required`?

scott-ray-wilson (changes requested):
1) I think we should either hide or disable these buttons with a tooltip for TOTP and also probably not showing the expiry since it's provider determined - I know it's a small window but we know this value is incorrect so I think it would be better not to display it?

CleanShot 2024-11-20 at 09 10 19@2x

2) I think we should have helper text on the create/update form for TOTP explaining that these default values should work with most providers and should only be configured if the provider explicitly provides different values

CleanShot 2024-11-20 at 09 19 18@2x

3) Prereq should include option for key in addition to url

CleanShot 2024-11-20 at 09 28 35@2x

backend/src/ee/services/dynamic-secret/providers/totp.ts
sheensantoscapadngan: I removed this because we're now doing the validation in the API layer during post/patch requests

backend/src/ee/services/dynamic-secret/providers/totp.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    if (algorithm) {
+      authenticatorInstance.options = { algorithm };
+    }
+
+    if (period) {
+      authenticatorInstance.options = { step: period };
+    }
+
+    return { entityId, data: { TOTP: authenticatorInstance.generate(secret) } };
```
scott-ray-wilson, line 68: Oh one other thing I think we should do; when we generate the TOTP I think we should check the remaining time on it and if it's less than some value (maybe 5 seconds) wait for the next code and send them the newer one? Since we don't refresh the code like traditional authenticators I think we should make sure the code has a big enough window for execution? Imagining they get a code at 1 second remaining and then it doesn't work

scott-ray-wilson (changes requested):

scott-ray-wilson (dismissed):

scott-ray-wilson (approved):
LGTM! Great work Sheen!

PR #2748, opened 2024-11-18
misc: address role slug issue in invite user endpoint

sheensantoscapadngan:
# Description 📣
This PR ensures that if roleSlugs is defined during user project invite, the roles are actually assigned correctly

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


DanielHougaard (changes requested):

DanielHougaard:
Actually this in theory a breaking change. If someone is passing in an empty array for the roles, their workflow will break. As I see it, this is not how the endpoint is intended to be used in the first place though. Either the user should pass in an array with items, or pass nothing at all.

Question is, are we okay with adding this potentially breaking change? I am personally okay with it, as the risk appears low

DanielHougaard (approved):

PR #2751, opened 2024-11-18
feat(dynamic-secrets): SAP ASE

DanielHougaard:
# Description 📣

This PR introduces SAP ASE support for dynamic secrets. SAP ASE uses the TDS protocol (v5.0) for connections. The package we use for establishing the TDS connection is called `odbc`, this package (almost) natively supports TDS v5 connections out of the box. I opted to use FreeTDS as our TDS driver, as this is very widely used and well documented. To make the TDS driver work, we need to install new packages to our Docker image, and add a odbc configuration file that contains information about our TDS driver _(`odbcinst.ini`, see backend docker files and standalone docker files)_. 

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


Dockerfile.fips.standalone-infisical
```diff
@@ -69,13 +69,21 @@ RUN groupadd -r -g 1001 nodejs && useradd -r -u 1001 -g nodejs non-root-user
 
 WORKDIR /app
 
-# Required for pkcs11js
+# Required for pkcs11js and ODBC
 RUN apt-get update && apt-get install -y \
     python3 \
     make \
     g++ \
+    unixodbc \
```
akhilmhdh, line 77: Hmmm so does the linux package also need to install these dependency?

backend/src/ee/services/dynamic-secret/providers/sap-ase.ts
```diff
@@ -0,0 +1,145 @@
...
+export const SapAseProvider = (): TDynamicProviderFns => {
+  const validateProviderInputs = async (inputs: unknown) => {
+    const providerInputs = await DynamicSecretSapAseSchema.parseAsync(inputs);
+
+    verifyHostInputValidity(providerInputs.host);
+    return providerInputs;
+  };
+
+  const $getClient = async (providerInputs: z.infer<typeof DynamicSecretSapAseSchema>, useMaster?: boolean) => {
+    const connectionString =
```
akhilmhdh, line 35: I think we need a localhost validation here like PG because odbc is able to connect to postgres as its generic. It won't be simple direct connection but its possible.

akhilmhdh (commented):
Pending application testing

Dockerfile.fips.standalone-infisical
DanielHougaard: Yeah it's needed for the nodegyp install

backend/src/ee/services/dynamic-secret/providers/sap-ase.ts
DanielHougaard: Right! Let me look into this

akhilmhdh (dismissed):

akhilmhdh (approved):

PR #2758, opened 2024-11-19
improvement: composite index for identityId:authMethod on access tokens table

DanielHougaard:
This PR adds a composite index on `identityId` and `authMethod`, in the identity access tokens table. The reason for this is to improve deletion speeds of access tokens where we query for identityId and authMethod.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---


maidul98:
Will cause down time since this migration will freeze access tokens table (count updates) and therefore all MI requests

PR #2760, opened 2024-11-19
feat: request ID support

DanielHougaard:
# Description 📣

This PR introduces request ID's on API error responses. This will make it easier to narrow down user-reported erorrs.
Additionally, we've also changed the format of request ID's, to have more entropy. This is done to avoid collisions in our logs.

**Minor**
1. I found a bug in the "Create Client Secret" modal, where it would fully crash the application if the user got a ValidationError/zod error from the API. This is also fixed in this PR.
2. Fixed mismatch between 403/401 errors on the frontend api error handler.
3. Improved API error handler to be more simplistic.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---


sheensantoscapadngan (commented):
add reqId to log points

backend/src/server/app.ts
```diff
@@ -104,6 +108,13 @@ export const main = async ({ db, hsmModule, auditLogDb, smtp, logger, queue, key
 
     await server.register(maintenanceMode);
 
+    await server.register(fastifyRequestContext, {
+      defaultStoreValues: (request) => ({
+        requestId: request.id,
+        log: request.log.child({ requestId: request.id })
```
sheensantoscapadngan, line 114: why do we need this part?

sheensantoscapadngan (approved):

sheensantoscapadngan (approved):

PR #2776, opened 2024-11-22
misc: unbinded scim from saml

sheensantoscapadngan:
# Description 📣
- This PR will separate SCIM from SAML so that it can also be used with OIDC SSO

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/ee/services/scim/scim-service.ts
```diff
@@ -302,13 +303,24 @@ export const scimServiceFactory = ({
...
+        detail: "Neither SAML or OIDC SSO is configured",
+        status: 400
+      });
+    }
+
     const appCfg = getConfig();
     const serverCfg = await getServerCfg();
 
+    const aliasType = org.orgAuthMethod === OrgAuthMethod.OIDC ? UserAliasType.OIDC : UserAliasType.SAML;
+    const trustScimEmails =
```
maidul98, line 317: would `autoVerifyScimEmails` be more fitting?

backend/src/services/org/org-dal.ts
```diff
@@ -14,20 +14,87 @@ import { DatabaseError } from "@app/lib/errors";
...
+        )
+        .first()) as TOrganizations & { orgAuthMethod?: string };
+
       return org;
     } catch (error) {
       throw new DatabaseError({ error, name: "Find org by id" });
     }
   };
 
+  const findOrgBySlug = async (orgSlug: string) => {
```
maidul98, line 61: Why don't we do find by id instead of slug or is there a reason it has to be be by slug when going off the redirect link?

backend/src/services/org/org-service.ts
```diff
@@ -187,6 +187,15 @@ export const orgServiceFactory = ({
     return members;
   };
 
+  const findOrgBySlug = async (slug: string) => {
```
maidul98, line 190: no need for permissions check here?

docs/documentation/platform/scim/overview.mdx
```diff
@@ -24,9 +28,14 @@ SCIM providers:
...
-    Infisical's SCIM implementation accounts for retaining the end-to-end encrypted architecture of Infisical because we decouple the **authentication** and **decryption** steps in the platform. 
-    
-    For this reason, SCIM-provisioned users are initialized but must finish setting up their account when logging in the first time by creating a master encryption/decryption key. With this implementation, IdPs and SCIM providers cannot and will not have access to the decryption key needed to decrypt your secrets.
-</Accordion>
-</AccordionGroup>
\ No newline at end of file
+  <Accordion title="Why do SCIM-provisioned users have to finish setting up their account?">
+    Infisical's SCIM implementation accounts for retaining the end-to-end
+    encrypted architecture of Infisical because we decouple the
+    **authentication** and **decryption** steps in the platform. For this
```
maidul98, line 34: auth and decryption is no longer decoupled, this is old info. sso works without user having to enter password any more

backend/src/services/org/org-dal.ts
sheensantoscapadngan: we do slug because we're expecting the user to input an identifier of the organization during the sign-in flow
image

maidul98 (approved):
lgtm

PR #2791, opened 2024-11-25
feat: Add new tag when creating secret

McPizza0:
# Description 📣

You can now create a new tag when adding a new secret to a project

CleanShot 2024-11-25 at 15 51 19@2x

This PR adds in a new `<CreatableSelect />` component which will be replaced once codebase refactor to `react-select` is complete

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [X] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

manually tested

---

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -102,7 +102,7 @@
     "react-mailchimp-subscribe": "^2.1.3",
     "react-markdown": "^8.0.3",
     "react-redux": "^8.0.2",
-    "react-select": "^5.8.1",
+    "react-select": "^5.8.3",
     "react-table": "^7.8.0",
     "react-toastify": "^9.1.3",
     "sanitize-html": "^2.12.1",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -102,7 +102,7 @@
     "react-mailchimp-subscribe": "^2.1.3",
     "react-markdown": "^8.0.3",
     "react-redux": "^8.0.2",
-    "react-select": "^5.8.3",
+    "react-select": "^5.8.1",
     "react-table": "^7.8.0",
     "react-toastify": "^9.1.3",
     "sanitize-html": "^2.12.1",
--- a/frontend/src/components/v2/CreatableSelect/CreatableSelect.tsx
+++ b/frontend/src/components/v2/CreatableSelect/CreatableSelect.tsx
@@ -1,0 +1,68 @@
+import { GroupBase } from "react-select";
+import ReactSelectCreatable, { CreatableProps } from "react-select/creatable";
+import { twMerge } from "tailwind-merge";
+
+import { ClearIndicator, DropdownIndicator, MultiValueRemove, Option } from "../Select/components";
+
+export const CreatableSelect = <T,>({
+  isMulti,
+  closeMenuOnSelect,
+  ...props
+}: CreatableProps<T, boolean, GroupBase<T>>) => {
+  return (
+    <ReactSelectCreatable
+      isMulti={isMulti}
+      closeMenuOnSelect={closeMenuOnSelect ?? !isMulti}
+      hideSelectedOptions={false}
+      unstyled
+      styles={{
+        input: (base) => ({
+          ...base,
+          "input:focus": {
+            boxShadow: "none"
+          }
+        }),
+        multiValueLabel: (base) => ({
+          ...base,
+          whiteSpace: "normal",
+          overflow: "visible"
+        }),
+        control: (base) => ({
+          ...base,
+          transition: "none"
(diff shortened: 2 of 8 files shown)


frontend/package.json
```diff
@@ -102,7 +102,7 @@
     "react-mailchimp-subscribe": "^2.1.3",
     "react-markdown": "^8.0.3",
     "react-redux": "^8.0.2",
-    "react-select": "^5.8.1",
+    "react-select": "^5.8.3",
```
DanielHougaard, line 105: Question: Why was patch version upgrade necessary?

frontend/src/components/v2/Select/components/index.tsx
```diff
@@ -0,0 +1,45 @@
+import {
```
DanielHougaard, line 1: I think the folder structure might be a little confusing here. Inside the `Select` folder we already have an index.tsx which exports from `Select.tsx`. I would probably move this file into `components/v2/CreatableSelect/components/SelectComponents.tsx`, because these functions are purely related to the CreatableSelect component

frontend/src/views/SecretMainPage/components/CreateSecretForm/CreateSecretForm.tsx
```diff
@@ -50,12 +51,39 @@ export const CreateSecretForm = ({
...
+    try {
+      const parsedSlug = slugSchema.parse(slug);
+      await createWsTag.mutateAsync({
+        workspaceID: workspaceId,
+        tagSlug: parsedSlug,
+        tagColor: ""
+      });
+    } catch (error) {
+      if (error instanceof z.ZodError) {
+        createNotification({
```
DanielHougaard, line 74: Mutation errors are automatically handled in `/src/reactQuery.tsx`, so this would result in multiple error notifications if the error is API related.

I would remove the `if (error instanceof z.ZodError) {` handler entirely

frontend/src/components/v2/CreatableSelect/CreatableSelect.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { ClearIndicator, DropdownIndicator, MultiValueRemove, Option } from "../Select/components";
+
+export const CreatableSelect = <T,>({
+  isMulti,
+  closeMenuOnSelect,
+  ...props
+}: CreatableProps<T, boolean, GroupBase<T>>) => {
+  return (
+    <ReactSelectCreatable
```
DanielHougaard, line 13: Currently it allows for empty space as valid input. I would treat an empty space like no input, so just show "No options" if the only input is an empty string (" ")

CleanShot 2024-11-26 at 18 53 41

frontend/src/views/SecretMainPage/components/CreateSecretForm/CreateSecretForm.tsx
```diff
@@ -148,16 +176,17 @@ export const CreateSecretForm = ({
               )
             }
           >
-            <FilterableSelect
+            <CreatableSelect
```
DanielHougaard, line 179: This only works from inside an environment. We should also add this to the overview create secret modal. 

CleanShot 2024-11-26 at 18 56 18

DanielHougaard (changes requested):
good work, left some comments

frontend/package.json
McPizza0: you're right, it wasnt needed 🤔

frontend/src/components/v2/Select/components/index.tsx
McPizza0: the select field child components are also used by `FilterableSelect` component, so I moved it to shared `Select` folder with the idea to later replace them all with the same `react-select` components
currently `Select` field is still using tailwind/radix dropdown as base

frontend/src/views/SecretMainPage/components/CreateSecretForm/CreateSecretForm.tsx
McPizza0: thought I implemented this
good catch!

frontend/src/views/SecretMainPage/components/CreateSecretForm/CreateSecretForm.tsx
```diff
@@ -70,17 +70,10 @@ export const CreateSecretForm = ({
         tagColor: ""
       });
     } catch (error) {
```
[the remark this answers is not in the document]
McPizza0: F

DanielHougaard (approved):

PR #2801, opened 2024-11-26
Fix(Azure Key Vault): Ignore disabled secrets

McPizza0:
# Description 📣

This fix ensures we only attempt to sync Azure Key Vault secrets that are not disabled in Azure.

**Now:**
CleanShot 2024-11-26 at 23 27 47@2x
CleanShot 2024-11-26 at 23 28 38@2x

**Previously** fetching the disabled key would throw an error, and block remaining sync functions
image

## Type ✨

- [X] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Manually tested by creating disabled secrets in Azure and manually triggering sync to replicate issue and create fix
---


maidul98 (approved):
Not a fan of using global vars in other functions but this will do

PR #2815, opened 2024-11-28
feat(k8-operator): push secrets

DanielHougaard:
# Description 📣

This PR introduces support for pushing secrets **from** the K8 operator **into** Infisical. We are using policies for the replacement of existing secrets. This mean the user can opt-in to overwrite existing/conflicting secrets in Infisical, to treat the K8 operator as the source of truth.

The operator will automatically update the serverside secret values if it detects a drift has happened, even if the key of a secret changes. If a secret key changes, the secret will be deleted and re-created with the correct secret key. We do this by keeping a map of ID/secret-key, that we are verifying on each reconcile.

If the deletionPolicy is set to Delete, all secrets managed by the operator will be deleted once the InfisicalPushSecret CRD is deleted.

Additionally, we are now using a proper logger for all our CRD's. No more printf or println statements.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---


akhilmhdh (changes requested):
Need to application test. But some questions from my side

1. Was this library something we were using before `go-logr` because I feel like it's quite unused one and uber zap is the most prominent one in this space.
2. Nit: Some messages starts with sentence case while others in small case

helm-charts/secrets-operator/Chart.yaml
```diff
@@ -13,9 +13,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: v0.7.5
+version: v0.7.6
```
maidul98, line 16: note: make sure to keep this updated before merge

helm-charts/secrets-operator/values.yaml
```diff
@@ -32,7 +32,7 @@ controllerManager:
         - ALL
     image:
       repository: infisical/kubernetes-operator
-      tag: v0.7.5
+      tag: v0.7.6
```
maidul98, line 35: note: you'll need to re-run  `make helm` command

k8-operator/controllers/infisicalpushsecret/infisicalpushsecret_controller.go
```diff
@@ -0,0 +1,249 @@
...
+		} else {
+			logger.Error(err, "Unable to fetch Infisical Secret CRD from cluster")
+			return ctrl.Result{
+				RequeueAfter: requeueTime,
+			}, nil
+		}
+	}
+
+	// Add finalizer if it doesn't exist
+	if !controllerutil.ContainsFinalizer(&infisicalPushSecretCRD, constants.INFISICAL_PUSH_SECRET_FINALIZER_NAME) {
```
maidul98, line 80: i thought we moved away from the finalizer pattern?

maidul98:
CleanShot 2024-12-17 at 15 54 28@2x
Something is wrong, it keeps updating secrets even though there were no secrets updated

maidul98 (commented):
Issue: The secrets that are created by the operator keep being updated (so the number of secret versions keep going up). If there aren't any changes, we shouldn't make a update. 

For future PRs, i recommend doing a separate PR to do folder restructure first and get it merged. This helps to reduce the surface area of pr testing and allows us to merge prs quicker.

Testing i did:
- auto secrets auto redeploy 
- delete managed secret 
- add new key and have it synced 

Need to do more testing on secret push after the update bug is patched

maidul98:
When `deletionPolicy: Delete` and you delete the kube secret, this is what i see:

CleanShot 2024-12-17 at 20 18 12@2x

k8-operator/controllers/infisicalpushsecret/infisicalpushsecret_controller.go
DanielHougaard: Finalizer is necessary in this case, or it will break when you delete the CRD or managed secret

maidul98 (approved):

PR #2825, opened 2024-12-01
feat(KMS): New external KMS support for Google GCP KMS

McPizza0:
# Description 📣

This PR adds Google's GCP KMS as a new External KMS integration.
image
image

This integration differs from AWS in the way its set up.
Detailed docs have been written to assist users in creating the necessary items within google.

## Type ✨

- [ ] Bug fix
- [X] New feature
- [ ] Improvement
- [ ] Breaking change
- [X] Documentation

# Tests 🛠️

Manually tested the full docs flow and tested adding a AWS KMS to ensure no issues arrise
---


backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -286,4 +291,89 @@ export const registerExternalKmsRouter = async (server: FastifyZodProvider) => {
...
+        gcpRegion: region,
+        keyName: ""
+      });
+      if (!results) {
+        throw new Error("Failed to validate GCP credential");
+      }
+
+      return results;
+    }
+  });
```
sheensantoscapadngan, line 328: best if we can somehow unify the paths for provider-specific endpoints. like it would be better if the path for this was something like `/gcp/keys`

sheensantoscapadngan: why is there a credential suffix in the url?

sheensantoscapadngan: let's also avoid having verbs in the url. let's try to stick to RESTful conventions when possible

backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -286,4 +291,89 @@ export const registerExternalKmsRouter = async (server: FastifyZodProvider) => {
...
+        actorOrgId: req.permission.orgId,
+        id: kmsId
+      });
+
+      if (!externalKms || externalKms.external.provider !== KmsProviders.GCP) {
+        throw new Error("KMS not found or not of type GCP");
+      }
+
+      const credentialJson = externalKms.external.providerInput.credential as TExternalKmsGcpCredentialSchema;
+
```
sheensantoscapadngan, line 366: same issues here

backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -286,4 +291,89 @@ export const registerExternalKmsRouter = async (server: FastifyZodProvider) => {
...
+      const { region, credential } = req.body;
+
+      const results = await server.services.externalKms.fetchGcpKeys({
+        credential,
+        gcpRegion: region,
+        keyName: ""
+      });
+      if (!results) {
+        throw new Error("Failed to validate GCP credential");
+      }
```
sheensantoscapadngan, line 324: hmm how would this happen if we're always returning an object

**
image
**

backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -286,4 +291,89 @@ export const registerExternalKmsRouter = async (server: FastifyZodProvider) => {
...
+      const credentialJson = externalKms.external.providerInput.credential as TExternalKmsGcpCredentialSchema;
+
+      const results = await server.services.externalKms.fetchGcpKeys({
+        credential: credentialJson,
+        gcpRegion: region,
+        keyName: ""
+      });
+      if (!results) {
+        throw new Error("Failed to validate GCP credential");
+      }
```
sheensantoscapadngan, line 374: same question here 
image

backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -286,4 +291,89 @@ export const registerExternalKmsRouter = async (server: FastifyZodProvider) => {
...
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const { region, credential } = req.body;
+
+      const results = await server.services.externalKms.fetchGcpKeys({
+        credential,
+        gcpRegion: region,
+        keyName: ""
```
sheensantoscapadngan, line 320: hmmm this seems to always be empty for all its usage

backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -286,4 +291,89 @@ export const registerExternalKmsRouter = async (server: FastifyZodProvider) => {
...
+    schema: {
+      params: z.object({
+        kmsId: z.string().trim().min(1)
+      }),
+      body: z.object({
+        region: z.string().trim().min(1)
+      }),
+      response: {
+        200: z.object({
+          keys: z.array(z.string())
```
sheensantoscapadngan, line 344: let's use `z.string().array()` for consistency

backend/src/ee/services/external-kms/providers/gcp-kms.ts
```diff
@@ -0,0 +1,111 @@
...
+    gcpKmsClient,
+    locationName
+  };
+};
+
+type GcpKmsProviderArgs = {
+  inputs: unknown;
+};
+type TGcpKmsProviderFactoryReturn = TExternalKmsProviderFns & {
+  // generateInputKmsKey: () => Promise<TExternalKmsGcpSchema>;
```
sheensantoscapadngan, line 24: remove if not needed

backend/src/ee/services/external-kms/providers/gcp-kms.ts
```diff
@@ -0,0 +1,111 @@
...
+    try {
+      const [keyRings] = await gcpKmsClient.listKeyRings({
+        parent: locationName
+      });
+
+      const validKeyRings = keyRings
+        .filter(
+          (keyRing): keyRing is { name: string } =>
+            keyRing !== null && typeof keyRing === "object" && "name" in keyRing && typeof keyRing.name === "string"
+        )
```
sheensantoscapadngan, line 59: when does it return a different value?

backend/src/ee/services/external-kms/providers/gcp-kms.ts
```diff
@@ -0,0 +1,111 @@
...
+        gcpKmsClient
+          .listCryptoKeys({
+            parent: keyRingName
+          })
+          .then(([cryptoKeys]) =>
+            cryptoKeys
+              .filter(
+                (key): key is { name: string } =>
+                  key !== null && typeof key === "object" && "name" in key && typeof key.name === "string"
+              )
```
sheensantoscapadngan, line 72: same question here: why do we have to do this?

backend/src/ee/services/external-kms/providers/gcp-kms.ts
```diff
@@ -0,0 +1,111 @@
...
+          )
+      );
+
+      const cryptoKeyLists = await Promise.all(keyListPromises);
+      keyList.push(...cryptoKeyLists.flat());
+      return { keys: keyList };
+    } catch {
+      throw new BadRequestError({
+        message: "Could not validate GCP KMS connection and credentials"
+      });
```
sheensantoscapadngan, line 83: why we don't we propagate the actual message of the error being thrown?

backend/src/ee/services/external-kms/providers/model.ts
```diff
@@ -42,14 +47,44 @@ export const ExternalKmsAwsSchema = z.object({
...
+export const ExternalKmsGcpClientSchema = ExternalKmsGcpSchema.pick({ gcpRegion: true }).extend({
+  credential: ExternalKmsGcpCredentialSchema
+});
+export type TExternalKmsGcpClientSchema = z.infer<typeof ExternalKmsGcpClientSchema>;
+
 // The root schema of the JSON
 export const ExternalKmsInputSchema = z.discriminatedUnion("type", [
-  z.object({ type: z.literal(KmsProviders.Aws), inputs: ExternalKmsAwsSchema })
+  z.object({ type: z.literal(KmsProviders.Aws), inputs: ExternalKmsAwsSchema }),
+  z.object({ type: z.literal(KmsProviders.GCP), inputs: ExternalKmsGcpSchema })
```
sheensantoscapadngan, line 81: nit but let's match the casing of Aws

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+  const fetchGCPKeys = async () => {
+    resetField("keyObject");
+    const credentialJson = kms ? undefined : await getCredentialFileJson();
+    if (!kms && !credentialJson) {
+      return;
+    }
+    const gcpRegion = getValues("gcpRegion").value;
+    if (!gcpRegion.length) {
+      setError("gcpRegion", {
+        message: "Please select a GCP region to fetch GPC Keys."
```
sheensantoscapadngan, line 203: typo

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+                field.onChange(e);
+                fetchGCPKeys();
+              }}
+              formatOptionLabel={formatOptionLabel}
+            />
+          </FormControl>
+        )}
+      />
+
+      {kms ? null : (
```
sheensantoscapadngan, line 289: are there no issues with displaying `null` ?

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+              value={field.value}
+              onChange={field.onChange}
+            />
+          </FormControl>
+        )}
+      />
+      {kms && (
+        <span className="text-xs text-mineshaft-300">
+          To change your GCP credentials, create a new External KMS server and assign it to project
+          you want to use it with.{" "}
```
sheensantoscapadngan, line 336: what is an external KMS server?

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+        message: "Invalid Service Account credential JSON."
+      });
+      return undefined;
+    }
+    clearErrors("credentialFile");
+    return parsedJson.data;
+  }
+
+  // handles the form submission
+  const handleAddGcpKms = async (data: AddExternalKmsGcpFormSchemaType) => {
```
sheensantoscapadngan, line 141: let's rename the method to just handleSave or something like that. might be misleading to label it as `add` when it can also `update`

frontend/src/hooks/api/kms/mutations.tsx
```diff
@@ -96,3 +101,37 @@ export const useLoadProjectKmsBackup = (projectId: string) => {
...
+        ? `/api/v1/external-kms/fetch-gcp-keys/${kmsId}`
+        : "/api/v1/external-kms/fetch-gcp-keys/credential";
+
+      const { data } = await apiRequest.post(apiUrl, {
+        region: gcpRegion,
+        ...rest
+      });
+
+      return data;
+    },
```
sheensantoscapadngan, line 132: why do we use two different routes for these and not just one?

frontend/src/hooks/api/kms/types.ts
```diff
@@ -100,3 +131,65 @@ export const AddExternalKmsSchema = z.object({
...
+    .describe("GCP key name"),
+  gcpRegion: z.object({ label: z.string().trim(), value: z.string().trim() }).describe("GCP Region")
+});
+
+export const AddExternalKmsGcpFormSchema = z.discriminatedUnion("formType", [
+  z
+    .object({
+      formType: z.literal("newGcpKms"),
+      credentialFile:
+        typeof window === "undefined"
```
sheensantoscapadngan, line 173: what's this condition for?

frontend/src/hooks/api/kms/types.ts
```diff
@@ -100,3 +131,65 @@ export const AddExternalKmsSchema = z.object({
...
+
+export type UpdateExternalKmsType = z.infer<typeof UpdateExternalKmsSchema>;
+
+const GCP_CREDENTIAL_MAX_FILE_SIZE = 8 * 1024; // 8KB
+const GCP_CREDENTIAL_ACCEPTED_FILE_TYPES = ["application/json"];
+
+const AddExternalKmsGcpFormSchemaStandardInputs = z.object({
+  keyObject: z
+    .object({ label: z.string().trim(), value: z.string().trim() })
+    .describe("GCP key name"),
```
sheensantoscapadngan, line 164: can we just register the `value` itself instead of the object?

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+  const { mutateAsync: updateGcpExternalKms } = useUpdateExternalKms(currentOrg?.id!);
+  const { mutateAsync: fetchGcpKeys, isLoading: isFetchGcpKeysLoading } =
+    useExternalKmsFetchGcpKeys(currentOrg?.id!);
+
+  // transforms the credential file into a JSON object
+  async function getCredentialFileJson(): Promise<ExternalKmsGcpCredentialSchemaType | undefined> {
+    const files = getValues("credentialFile");
+    if (!files || !files.length) {
+      return undefined;
+    }
```
sheensantoscapadngan, line 120: can we `setError` here too?

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+    const file = files[0];
+    if (file.type !== "application/json") {
+      setError("credentialFile", {
+        message: "Only .json files are accepted."
+      });
+      return undefined;
```
sheensantoscapadngan, line 126: also I think there's no need to explicitly return `undefined`. it can just be `return` afaik

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
McPizza0: not that I have seen or aware of; but yes can switch to `!kms &&` instead

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
McPizza0: Ah, makes sense!
Was following convention of AWS form - will update both

frontend/src/hooks/api/kms/types.ts
McPizza0: Nextjs server side rendering causes issues with FileType
so we check if its on the server, and set different types

frontend/src/hooks/api/kms/types.ts
McPizza0: Object is needed for `<FilterableSelect />` component

McPizza0:
> the last dropdown option gets cut off image

This is interesting
might need to do some Z-index work on the `<FilterableSelect />` component 🤔

frontend/src/hooks/api/kms/mutations.tsx
DanielHougaard: Yeah I think this will cause confusion since the function name remains the same

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
DanielHougaard: Nit: I would rather have it return `null` than undefined, it's more specific

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
DanielHougaard: @McPizza0 `fetch GPC Keys.`, GPC should be GCP

DanielHougaard (changes requested):

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
McPizza0: Error here would trigger when a user sets a region before uploading credential file
this is a catch to ensure user has actually uploaded something before displaying errors on the upload field

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
McPizza0: CleanShot 2024-12-02 at 18 29 25@2x
Cant return nothing - switched to null as per Daniels recommendation

backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -44,7 +48,8 @@ const sanitizedExternalSchemaForGetById = KmsKeysSchema.extend({
     statusDetails: true,
     provider: true
   }).extend({
-    providerInput: ExternalKmsAwsSchema
```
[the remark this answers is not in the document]
McPizza0: yeah, agree that makes sense
happy to move once a location has been made - dont see a good folder for it

docs/documentation/platform/kms-configuration/gcp-kms.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+      2. In the KMS section, click **Create Key Ring**.
+      ![GCP Create Key Ring](/images/platform/kms/gcp/keyring-create.png)
+
+      3. Give the key ring a **Name** and select a **Region**, then click **Create**.
+      <Info>
+        We don't currently support multi-region key rings.
+      </Info>
+
+      4. On the "Create Key" page, give the key a **Name** and ensure **Protection Level** is set to **Software**, then click **Continue**.
```
sheensantoscapadngan, line 53: do they really have to set Protection Level to "Software"? I think it generally shouldn't matter right? Orgs have different compliance requirements and some  may have to use "HSM" or "EXTERNAL" to connect to their own external key manager

backend/src/ee/routes/v1/external-kms-router.ts
```diff
@@ -286,4 +291,65 @@ export const registerExternalKmsRouter = async (server: FastifyZodProvider) => {
...
+        const externalKms = await server.services.externalKms.findById({
+          actor: req.permission.type,
+          actorId: req.permission.id,
+          actorAuthMethod: req.permission.authMethod,
+          actorOrgId: req.permission.orgId,
+          id: req.body.kmsId
+        });
+
+        if (!externalKms || externalKms.external.provider !== KmsProviders.Gcp) {
+          throw new Error("KMS not found or not of type GCP");
```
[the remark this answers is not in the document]
McPizza0: now I know we have them 😂

frontend/src/views/Settings/OrgSettingsPage/components/OrgEncryptionTab/GcpKmsForm.tsx
```diff
@@ -0,0 +1,349 @@
...
+              onChange={field.onChange}
+            />
+          </FormControl>
+        )}
+      />
+      {kms && (
+        <span className="text-xs text-mineshaft-300">
+          To change your GCP credentials, create a new KMS server and assign it to project you want
+          to use it with.{" "}
+        </span>
```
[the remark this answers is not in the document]
McPizza0: sorry, my bad
changed to "... new External KMS..."

docs/documentation/platform/kms-configuration/gcp-kms.mdx
McPizza0: I dont have any way to test if that needs any changes on our side
do you think there shouldnt be any differences?

McPizza0: i dont think this should block the PR
we can test and open separate issue to address docs

sheensantoscapadngan: yeah I was able to test it with `Protection Level` set to HSM and everything went as expected. It's an internal thing that GCP handles, so the `Protection Level` shouldn't really affect our integration. It's about how GCP handles the key material under the hood. If it's set to external, then GCP attempts to use the key material referenced from the external key store. 

I say we rephrase it to be "set Protection level based on your requirements"

DanielHougaard (approved):

PR #2828, opened 2024-12-02
fix: Schema validation errors correctly returned as 422

McPizza0:
# Description 📣

When input schema validation failed, the server would return a HTTP code 401 which could in some cases break the client.
This PR changes the schema validation HTTP error code to be 422 (Unprocessable Content) to be inline with standards.

CleanShot 2024-12-02 at 23 51 41@2x

## Type ✨

- [X] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Manual Testing

---

### DIFF
--- a/backend/src/server/routes/sanitizedSchemas.ts
+++ b/backend/src/server/routes/sanitizedSchemas.ts
@@ -32,7 +32,7 @@ export const DefaultResponseErrorsSchema = {
   400: z.object({
     requestId: z.string(),
     statusCode: z.literal(400),
-    message: z.any(),
+    message: z.string(),
     error: z.string()
   }),
   404: z.object({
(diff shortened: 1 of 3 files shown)


backend/src/server/routes/sanitizedSchemas.ts
```diff
@@ -32,7 +32,7 @@ export const DefaultResponseErrorsSchema = {
   400: z.object({
     requestId: z.string(),
     statusCode: z.literal(400),
-    message: z.string(),
```
DanielHougaard, line 35: 1. Revert this back to `z.string()`. 
2. Set the `401` message to `z.string()` instead of any, since we now know 401 message will be a string. 
3. Create a new item in the object for 422 errors so we can get proper API documentation for it.

```ts
  422: z.object({
    requestId: z.string(),
    statusCode: z.literal(422),
    message: z.any(),
    error: z.string()
  }),

DanielHougaard (changes requested):

backend/src/server/routes/sanitizedSchemas.ts
McPizza0: ah good point!

DanielHougaard (approved):

PR #2834, opened 2024-12-03
fix(Approval Workflows): Workflows keep approval history after deletion

McPizza0:
# Description 📣

Before:
When you delete an approval workflow, all related requests were also deleted. This can be an issue for auditing and logging purposes

https://github.com/user-attachments/assets/1c158eed-9842-4ba3-b70d-cfaf58570e3b

Now:
When you delete a workflow; the following happens:
- workflow has a `deletedAt` date set
- all open Secret and Access requests are closed
- any previously generated additionalUserPrivileges are deleted

## Type ✨

- [X] Bug fix
- [ ] New feature
- [X] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Tested manually as per videos above
---

### DIFF
--- a/backend/src/ee/services/access-approval-request/access-approval-request-dal.ts
+++ b/backend/src/ee/services/access-approval-request/access-approval-request-dal.ts
@@ -61,7 +61,8 @@ export const accessApprovalRequestDALFactory = (db: TDbClient) => {
           db.ref("approvals").withSchema(TableName.AccessApprovalPolicy).as("policyApprovals"),
           db.ref("secretPath").withSchema(TableName.AccessApprovalPolicy).as("policySecretPath"),
           db.ref("enforcementLevel").withSchema(TableName.AccessApprovalPolicy).as("policyEnforcementLevel"),
-          db.ref("envId").withSchema(TableName.AccessApprovalPolicy).as("policyEnvId")
+          db.ref("envId").withSchema(TableName.AccessApprovalPolicy).as("policyEnvId"),
+          db.ref("deletedAt").withSchema(TableName.AccessApprovalPolicy).as("policyDeletedAt")
         )
 
         .select(db.ref("approverUserId").withSchema(TableName.AccessApprovalPolicyApprover))
@@ -118,7 +119,8 @@ export const accessApprovalRequestDALFactory = (db: TDbClient) => {
             approvals: doc.policyApprovals,
             secretPath: doc.policySecretPath,
             enforcementLevel: doc.policyEnforcementLevel,
-            envId: doc.policyEnvId
+            envId: doc.policyEnvId,
+            deletedAt: doc.policyDeletedAt
           },
           requestedByUser: {
             userId: doc.requestedByUserId,
@@ -141,7 +143,7 @@ export const accessApprovalRequestDALFactory = (db: TDbClient) => {
               }
             : null,
 
-          isApproved: !!doc.privilegeId
+          isApproved: !!doc.policyDeletedAt || !!doc.privilegeId
         }),
         childrenMapper: [
           {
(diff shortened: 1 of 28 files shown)


backend/src/ee/services/access-approval-request/access-approval-request-dal.ts
```diff
@@ -141,7 +143,7 @@ export const accessApprovalRequestDALFactory = (db: TDbClient) => {
               }
             : null,
 
-          isApproved: !!doc.privilegeId
+          isApproved: !!doc.policyDeletedAt || !!doc.privilegeId
```
DanielHougaard, line 146: This will resolve to `true` if the policy is not deleted. We are basically saying: isApproved = true IF policy not deleted, OR privilege ID present

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -130,6 +130,9 @@ export const accessApprovalRequestServiceFactory = ({
         message: `No policy in environment with slug '${environment.slug}' and with secret path '${secretPath}' was found.`
       });
     }
+    if (policy.deletedAt) {
+      throw new BadRequestError({ message: "The policy linked to this request has been deleted" });
```
DanielHougaard, line 134: Error would make more sense if it was `You can't create access requests for a policy that has been deleted`

backend/src/ee/services/secret-approval-policy/secret-approval-policy-dal.ts
```diff
@@ -177,5 +177,11 @@ export const secretApprovalPolicyDALFactory = (db: TDbClient) => {
     }
   };
 
-  return { ...secretApprovalPolicyOrm, findById, find };
+  const softDeleteById = async (policyId: string, tx?: Knex) => {
+    const date = new Date();
+    const softDeletedPolicy = await secretApprovalPolicyOrm.updateById(policyId, { deletedAt: date }, tx);
```
DanielHougaard, line 182: Nit: Make this one line, `deletedAt: new Date()`

backend/src/ee/services/secret-approval-policy/secret-approval-policy-service.ts
```diff
@@ -301,7 +305,14 @@ export const secretApprovalPolicyServiceFactory = ({
...
 
-    await secretApprovalPolicyDAL.deleteById(secretPolicyId);
+    await secretApprovalPolicyDAL.transaction(async (tx) => {
+      await secretApprovalRequestDAL.update(
+        { policyId: secretPolicyId, status: RequestState.Open },
+        { status: RequestState.Closed },
+        tx
+      );
+      await secretApprovalPolicyDAL.softDeleteById(secretPolicyId, tx);
+    });
```
DanielHougaard, line 315: Nit: Currently we are doing `return sapPolicy`, which was okay before because `sapPolicy` wouldn't change. Now we are actually updating the policy, so we should return the updated policy instead.

You can do `return await secretApprovalPolicyDAL.softDeleteById(secretPolicyId, tx);` inside the transaction. And do `updatedPolicy = await secretApprovalPolicyDAL.transaction(async (tx) => { .....`

DanielHougaard (changes requested):
Left a few comments, testing on my end now

DanielHougaard:
When an access request policy is deleted, existing access requests remain active.

I think we should be deleting additional privileges associated to requests that are owned by the policy that gets deleted.

backend/src/ee/services/access-approval-request/access-approval-request-dal.ts
McPizza0: `!!doc.policyDeletedAt` will resolve to false if the value is null and true if the value is set
CleanShot 2024-12-09 at 12 44 31@2x

We use this value to show the difference on the UI and block changes from happening
if we do not return as "isApproved" it will show in the UI that its an outstanding request and offer the user to approve etc.

We can make separate front end changes to avoid this if we need

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
McPizza0: "You can't create access requests for a policy that has been deleted"
this makes it seem like they already know (or should know) the policy has been deleted

"The policy linked to this request has been deleted"
this informs them that the policy has been deleted, and since an error will show, they would assume its why they cant create a new request/approval

2 different approaches

McPizza0:
> When an access request policy is deleted, existing access requests remain active.
> 
> I think we should be deleting additional privileges associated to requests that are owned by the policy that gets deleted.

we do this in deleteAccessApprovalPolicy

DanielHougaard (approved):
Approved, lets thoroughly test this in Gamma

PR #2835, opened 2024-12-03
Improvement: Integrations Table and UI Improvements

scott-ray-wilson:
# Description 📣

This PR improves management of integrations by adding an integration table with filter, search and pagination. Various elements of high-level integrations UI have been adjusted as well. Integrations missing environment variables in self-hosted instances will now notify users and provide a link to the pertinent docs.

CleanShot 2024-12-03 at 14 54 52@2x

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


frontend/src/views/IntegrationsPage/components/IntegrationsSection/components/IntegrationsTable.tsx
```diff
@@ -0,0 +1,415 @@
+import { useMemo, useState } from "react";
```
sheensantoscapadngan, line 1: what do you think about making the search bar filter against the destination fields as well? I was under the impression that I can use the search bar to filter baesd on the destination values

image

frontend/src/views/IntegrationsPage/components/IntegrationsSection/components/IntegrationRow.tsx
```diff
@@ -0,0 +1,190 @@
+import { useMemo } from "react";
```
sheensantoscapadngan, line 1: thoughts on sorting this based on the integration by default?

frontend/src/views/IntegrationsPage/components/IntegrationsSection/components/IntegrationDetails.tsx
```diff
@@ -0,0 +1,137 @@
...
+          <FormLabel label="Category" />
+          <div className={FIELD_CLASSNAME}>{integration.targetService}</div>
+        </div>
+      )}
+      {(integration.integration === "checkly" || integration.integration === "github") && (
+        <div>
+          <FormLabel label="Secret Suffix" />
+          <div className={FIELD_CLASSNAME}>{integration?.metadata?.secretSuffix || "-"}</div>
+        </div>
+      )}
```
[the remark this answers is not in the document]
scott-ray-wilson: yeah I'll just hide the field entirely if not set

frontend/src/views/IntegrationsPage/components/IntegrationsSection/components/IntegrationRow.tsx
scott-ray-wilson: I originally had this but my concern is that if an integration failed to sync and it's not on the current page you may miss it; I've pushed a change that will default to ordering by integration unless there are 1 or more failed integrations; then it will sort by status - let me know your thoughts on this approach. Also how do you feel about this styling change for failed integrations?

CleanShot 2024-12-04 at 17 21 19@2x

sheensantoscapadngan: looks good!!!

sheensantoscapadngan (approved):

sheensantoscapadngan (approved):

PR #2855, opened 2024-12-09
feat: added endpoint to update integration auth

akhilmhdh:
# Description 📣

This PR adds endpoint to update integration auth id. Used by terraform.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/server/routes/v1/integration-auth-router.ts
+++ b/backend/src/server/routes/v1/integration-auth-router.ts
@@ -83,6 +83,67 @@ export const registerIntegrationAuthRouter = async (server: FastifyZodProvider) 
   });
 
   server.route({
+    method: "PATCH",
+    url: "/:integrationAuthId",
+    config: {
+      rateLimit: writeLimit
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    schema: {
+      description: "Update the integration authentication object required for syncing secrets.",
+      security: [
+        {
+          bearerAuth: []
+        }
+      ],
+      querystring: z.object({
+        integrationAuthId: z.string().trim().describe(INTEGRATION_AUTH.UPDATE_BY_ID.integrationAuthId)
+      }),
+      body: z.object({
+        integration: z.string().trim().optional().describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.integration),
+        accessId: z.string().trim().optional().describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.accessId),
+        accessToken: z.string().trim().optional().describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.accessToken),
+        awsAssumeIamRoleArn: z
+          .string()
+          .url()
+          .trim()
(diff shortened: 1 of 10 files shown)


backend/src/server/routes/v1/integration-auth-router.ts
```diff
@@ -82,6 +82,67 @@ export const registerIntegrationAuthRouter = async (server: FastifyZodProvider)
...
+        awsAssumeIamRoleArn: z
+          .string()
+          .url()
+          .trim()
+          .optional()
+          .describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.awsAssumeIamRoleArn),
+        url: z.string().url().trim().optional().describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.url),
+        namespace: z.string().trim().optional().describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.namespace),
+        refreshToken: z.string().trim().optional().describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.refreshToken)
+      }),
```
sheensantoscapadngan, line 115: I think better if we used a separate constant dedicated for `Update Integration Auth` (I'm referring to the values in describe)

backend/src/server/routes/v1/integration-auth-router.ts
```diff
@@ -82,6 +82,67 @@ export const registerIntegrationAuthRouter = async (server: FastifyZodProvider)
...
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        projectId: integrationAuth.projectId,
+        event: {
+          type: EventType.AUTHORIZE_INTEGRATION,
+          metadata: {
+            integration: integrationAuth.integration
+          }
+        }
+      });
```
sheensantoscapadngan, line 141: should we emit a different event where we could include the updated details as metadata?

backend/src/server/routes/v1/integration-auth-router.ts
```diff
@@ -82,6 +82,67 @@ export const registerIntegrationAuthRouter = async (server: FastifyZodProvider)
...
+      security: [
+        {
+          bearerAuth: []
+        }
+      ],
+      querystring: z.object({
+        integrationAuthId: z.string().trim().describe(INTEGRATION_AUTH.UPDATE_BY_ID.integrationAuthId)
+      }),
+      body: z.object({
+        integration: z.string().trim().optional().describe(INTEGRATION_AUTH.CREATE_ACCESS_TOKEN.integration),
```
sheensantoscapadngan, line 103: can we declare this as a nativeEnum instead so that we no longer have to do this check?
image

backend/src/server/routes/v1/integration-auth-router.ts
akhilmhdh: @sheensantoscapadngan It's ok! 

1. Both are generic same text
2. This endpoint won't used anymore after this as scott is working on integration v2

sheensantoscapadngan (approved):

PR #2858, opened 2024-12-09
feat(azure-app-integration): label & reference support

DanielHougaard:
# Description 📣

This PR adds support for environment labeling and references as secret values.

The labeling is necessary incase there are multiple integrations for multiple environments, and the keys are overlapping. Azure supports multiple secrets with the same key, with different labels.

Reference support allows for users to set the secret value to an azure reference. If this happens, we set the content type of the azure key, to `application/vnd.microsoft.appconfig.keyvaultref+json;charset=utf-8`, which indicates that the secret should be treated as a reference in Azure App Config. 

## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/pages/integrations/azure-app-configuration/create.tsx
+++ b/frontend/src/pages/integrations/azure-app-configuration/create.tsx
@@ -4,7 +4,11 @@ import Head from "next/head";
 import Image from "next/image";
 import Link from "next/link";
 import { useRouter } from "next/router";
-import { faArrowUpRightFromSquare, faBookOpen } from "@fortawesome/free-solid-svg-icons";
+import {
+  faArrowUpRightFromSquare,
+  faBookOpen,
+  faQuestionCircle
+} from "@fortawesome/free-solid-svg-icons";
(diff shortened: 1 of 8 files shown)


frontend/src/pages/integrations/azure-app-configuration/create.tsx
```diff
@@ -155,35 +174,70 @@ export default function AzureAppConfigurationCreateIntegration() {
...
+                render={({ field: { onChange, value } }) => (
+                  <Switch
+                    id="use-environment-labels"
+                    onCheckedChange={(isChecked) => onChange(isChecked)}
+                    isChecked={value}
+                  >
+                    <FormLabel label="Use Labels" />
+                  </Switch>
+                )}
+              />
```
sheensantoscapadngan, line 221: do we still need this? I mean... if they don't add any label then I think that would just mean that no label should be used

sheensantoscapadngan (approved):

PR #2859, opened 2024-12-10
fix(dashboard): pasting secrets into create secret modal

DanielHougaard:
# Description 📣

Fixed pasting secrets into the create secret modal. Previously the paste behavior was very buggy, and caused the value to be wiped if you pasted again. It would also wipe the secret key if you pasted a new value. All in all it felt very unnatural to use the secret paste feature.

Additionally I also fixed the splitter, so the value becomes everything after the FIRST splitter (such as `:` or `=`), instead of the last.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/helpers/parseEnvVar.ts
+++ b/frontend/src/helpers/parseEnvVar.ts
@@ -15,7 +15,10 @@ export const getKeyValue = (pastedContent: string, delimiters: string[]) => {
     }
   });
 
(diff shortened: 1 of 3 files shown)


scott-ray-wilson (approved):
LGTM!

PR #2862, opened 2024-12-10
fix(dashboard): improved project creation speed

DanielHougaard:
# Description 📣

Removed the previously necessary 2 second delay when creating projects, and instead added a forceful refetch of projects. We are already invalidating the get all projects query after creating a new project, so this seems redundant right? No. When we invalidate queries, it doesn't wait for them to invalidate. It starts the invalidation, and then redirects to the project page, which causes the UI to glitch out. To completely fix this, we do a manual refetch and await it.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---


McPizza0 (approved):
lgtm

PR #2865, opened 2024-12-11
fix(secret-reminders): proper cleanup on deleted resources

DanielHougaard:
# Description 📣

This PR fixes secret reminders not getting deleted when secrets are deleted. Additionally, I've also added coverage for project and organization deletions. This means if a project or organization is deleted, all secrets within that have reminders, will have their reminder job removed, just like on a normal secret deletion.

**Misc**
1. While testing the deletion of reminders on org-level, I realized that after a user deletes an organization, they're completely unable to create a new organization without logging out. This PR also addresses this issue. I also found a lot of UI bugs with the `/org/none` page, this PR fixes those as well to make the entire flow run more smoothly.
2. Added new `$notNull` filter for the `find` ORM.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/services/auth-token/auth-token-service.ts
+++ b/backend/src/services/auth-token/auth-token-service.ts
@@ -1,6 +1,7 @@
 import crypto from "node:crypto";
 
 import bcrypt from "bcrypt";
+import { Knex } from "knex";
 
 import { TAuthTokens, TAuthTokenSessions } from "@app/db/schemas";
 import { getConfig } from "@app/lib/config/env";
@@ -123,14 +124,13 @@ export const tokenServiceFactory = ({ tokenDAL, userDAL, orgMembershipDAL }: TAu
     return deletedToken?.[0];
   };
 
-  const getUserTokenSession = async ({
-    userId,
-    ip,
-    userAgent
-  }: TIssueAuthTokenDTO): Promise<TAuthTokenSessions | undefined> => {
+  const getUserTokenSession = async (
+    { userId, ip, userAgent }: TIssueAuthTokenDTO,
+    tx?: Knex
+  ): Promise<TAuthTokenSessions | undefined> => {
     let session = await tokenDAL.findOneTokenSession({ userId, ip, userAgent });
     if (!session) {
-      session = await tokenDAL.insertTokenSession(userId, ip, userAgent);
+      session = await tokenDAL.insertTokenSession(userId, ip, userAgent, tx);
     }
     return session;
   };
--- a/backend/src/services/auth-token/auth-token-service.ts
+++ b/backend/src/services/auth-token/auth-token-service.ts
@@ -128,7 +128,7 @@ export const tokenServiceFactory = ({ tokenDAL, userDAL, orgMembershipDAL }: TAu
     { userId, ip, userAgent }: TIssueAuthTokenDTO,
     tx?: Knex
   ): Promise<TAuthTokenSessions | undefined> => {
-    let session = await tokenDAL.findOneTokenSession({ userId, ip, userAgent });
+    let session = await tokenDAL.findOneTokenSession({ userId, ip, userAgent }, tx);
     if (!session) {
       session = await tokenDAL.insertTokenSession(userId, ip, userAgent, tx);
     }
--- a/backend/src/services/org/org-service.ts
+++ b/backend/src/services/org/org-service.ts
@@ -31,11 +31,13 @@ import { BadRequestError, ForbiddenRequestError, NotFoundError, UnauthorizedErro
 import { groupBy } from "@app/lib/fn";
 import { alphaNumericNanoId } from "@app/lib/nanoid";
 import { isDisposableEmail } from "@app/lib/validator";
+import { TQueueServiceFactory } from "@app/queue";
 import { getDefaultOrgMembershipRoleForUpdateOrg } from "@app/services/org/org-role-fns";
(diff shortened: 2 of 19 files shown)


backend/src/services/auth-token/auth-token-service.ts
```diff
@@ -123,14 +124,13 @@ export const tokenServiceFactory = ({ tokenDAL, userDAL, orgMembershipDAL }: TAu
...
-  const getUserTokenSession = async ({
-    userId,
-    ip,
-    userAgent
-  }: TIssueAuthTokenDTO): Promise<TAuthTokenSessions | undefined> => {
+  const getUserTokenSession = async (
+    { userId, ip, userAgent }: TIssueAuthTokenDTO,
+    tx?: Knex
+  ): Promise<TAuthTokenSessions | undefined> => {
     let session = await tokenDAL.findOneTokenSession({ userId, ip, userAgent });
```
akhilmhdh, line 131: I think you should pass this also as transaction then

maidul98:
Can you please do a check in our daily clean up to check for secret remidners where the secret id doesn't exist. if it doesn't exist, we should remove that job

maidul98:
We should also see why we aren't cleaning up bulmq jobs after they finish 
CleanShot 2024-12-17 at 00 34 19@2x

backend/src/services/org/org-service.ts
```diff
@@ -419,24 +435,88 @@ export const orgServiceFactory = ({
   /*
    * Delete organization by id
    * */
-  const deleteOrganizationById = async (
-    userId: string,
-    orgId: string,
-    actorAuthMethod: ActorAuthMethod,
-    actorOrgId: string | undefined
-  ) => {
+  const deleteOrganizationById = async ({
```
maidul98, line 438: nit: it is better to keep prs focued on one objective and open seperate ones if focus changes. the shorter a pr is the quicker it can be merged

backend/src/services/secret/secret-dal.ts
```diff
@@ -339,6 +341,94 @@ export const secretDALFactory = (db: TDbClient) => {
...
+        } catch (error) {
+          numberOfRetryOnFailure += 1;
+          logger.error(error, `Failed to process batch at offset ${offset}`);
+
+          if (numberOfRetryOnFailure >= MAX_RETRY_ON_FAILURE) {
+            break;
+          }
+
+          // Retry the current batch
+          offset -= REMINDER_PRUNE_BATCH_SIZE;
```
maidul98, line 413: This might cause an infinite loop. 

case:

batch size = 5000
offset = 0 [0, 5000]
success 
offset = 5000 [5000, 10k] 
fails (so we do offset -= REMINDER_PRUNE_BATCH_SIZE)
offset = 0 (we keep doing this over and over again)


i think it is better to use a while loop in this case

backend/src/services/secret/secret-fns.ts
```diff
@@ -1138,3 +1142,49 @@ export const decryptSecretWithBot = (
...
+  queueService: Pick<TQueueServiceFactory, "stopRepeatableJob">;
+  projectBotService: Pick<TProjectBotServiceFactory, "getBotKey">;
+  folderDAL: Pick<TSecretFolderDALFactory, "findByProjectId">;
+};
+
+export const fnDeleteProjectSecretReminders = async (
+  projectId: string,
+  { secretDAL, secretV2BridgeDAL, queueService, projectBotService, folderDAL }: TFnDeleteProjectSecretReminders
+) => {
+  const projectFolders = await folderDAL.findByProjectId(projectId);
```
maidul98, line 1158: question: couldn't this return a large payload like 10k folders?

backend/src/services/secret/secret-fns.ts
```diff
@@ -1138,3 +1142,49 @@ export const decryptSecretWithBot = (
...
+      : (secret as { secretReminderRepeatDays: number }).secretReminderRepeatDays;
+
+    // We're using the queue service directly to get around conflicting imports.
+    if (repeatDays) {
+      await queueService.stopRepeatableJob(
+        QueueName.SecretReminder,
+        QueueJobs.SecretReminder,
+        {
+          // on prod it this will be in days, in development this will be second
+          every: appCfg.NODE_ENV === "development" ? secondsToMillis(repeatDays) : daysToMillisecond(repeatDays)
```
maidul98, line 1184: comment: this is a good use of comments

backend/src/services/org/org-service.ts
```diff
@@ -419,24 +435,88 @@ export const orgServiceFactory = ({
...
-    return organization;
+
+    if (!userAgentHeader) {
+      throw new BadRequestError({ name: "User agent not set on request." });
+    }
+
+    const cfg = getConfig();
+    const authToken = authorizationHeader.replace("Bearer ", "");
+
+    const decodedToken = jwt.verify(authToken, cfg.AUTH_SECRET) as AuthModeJwtTokenPayload;
```
maidul98, line 474: comment: this feels like something that would happen in the auth verify middleware?

maidul98 (commented):
Left a few small comments

backend/src/services/org/org-service.ts
DanielHougaard: @maidul98 We need the details from the token, to generate a new almost identical token, without the org ID

Issue #2845, opened 2024-12-05
`blur` results in lost input data


felixhummel:
### Describe the bug
UI: `blur` results in lost input data

### To Reproduce
- secrets > add
- painstakingly add values
- miss-click in the background (for example to get rid of browser auto-completion for "KEY" field)

### Expected behavior
nothing

### actual
Form data is lost. My frustration increases.

PR #2892, opened 2024-12-17
improv ui: don't close "Create Secrets" modal when clicking outside it

Pranav2612000:
# Description 📣

- As mentioned in #2845, it can be frustrating for the user to see their data being lost when they mistakenly click outside
the modal while adding secrets.
This PR provides a probable fix for this by ensuring that the modal remains open.
**Note:** This is just one of the ways of fixing this problem. We need to decide an uniform way of handling modals throughout the product ( should modals close when clicked outside? ). This PR serves as a starting point for starting this discussion.
- This PR also includes a commit to enable the `modal` mode for radix dialogs. This ensures that interaction with outside elements is disabled and only dialog content is visible to screen readers. ( [Ref](https://www.radix-ui.com/primitives/docs/components/dialog) )

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️
- Tested manually by opening the "Create Secrets" modal and ensuring that it doesn't close after this change

## Possible Improvements
- Abstract away the `onPointerDownOutside` prop to decide whether a modal closes on clicking outside

---

### DIFF
--- a/frontend/src/components/v2/Modal/Modal.tsx
+++ b/frontend/src/components/v2/Modal/Modal.tsx
@@ -72,7 +72,7 @@ ModalContent.displayName = "ModalContent";
 
 export type ModalProps = Omit<DialogPrimitive.DialogProps, "open"> & { isOpen?: boolean };
 export const Modal = ({ isOpen, ...props }: ModalProps) => (
-  <DialogPrimitive.Root open={isOpen} {...props} />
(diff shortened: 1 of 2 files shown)


akhilmhdh:
Cheers 🍻

Issue #431, opened 2023-03-12
Other Vercel Integrations getting overwritten by Infisical


Makisuo:
### Describe the bug
When Infisical is updated, it deletes all environment variables for a different integration I use (sentry)
 
### To Reproduce
1. Add Sentry Integration to Project
2. Everything should be there
3. Add Infisical Integration
4. Sentry Enviornment Variables are gone

### Expected behavior
Shouldnt overried them

### Platform you are having the issue on:
Vercel

mv-turtle:
Right, this is a very valid use case! @dangtony98 was looking into this some time ago

Thank you for flagging :)

polypixeldev:
I've just encountered the same issue.

dangtony98:
Hey @polypixeldev @Makisuo!

I've added this to our roadmap as sync prioritization/behavior (will likely get to it in April). The idea is we'll give users the ability to select which type of sync behavior you want such as:

- Overwriting everything in the integration in favor of the environment variables in Infisical.
- Overwriting everything in Infisical in favor of the environment variables currently in the integration.
- Replacing environment variables in the integration with updated values in Infisical without deleting other variables.

...

This is still an early train of thought but we'll have this fleshed-out in the coming month.

polypixeldev:
Is this still being worked on?

sheensantoscapadngan:
I can work on this one @dangtony98

sheensantoscapadngan:
@vmatsiiako

delavegar25:
hi, is this still being worked on ?

nealchandra:
+1, this is quite frustrating. Other integrations often completely automate the setting of their relevant env vars, overriding them essentially creates a bunch of work to transcribe these values to Infisical by hand and increases the likelihood of misconfiguration.

I don't think this fix needs all the controls discussed above -- as a starting point simply not unsetting env vars that are already set in Vercel and not referenced in Infisical would be significantly preferable to the current behavior of silently nuking them.

maidul98:
@nealchandra I think the current approach tries to keep Infisical as a single source of truth. This is why it attempts to create a one to one mapping of secrets in Infisical (this can mean deleting ones that do not exist in the thrid party service). 

It sounds like you would instead like to sync secrets over to the thrid party service by: update them if they exist, create them if they don't exist and leave eveything else as is? @Makisuo Would you also agree with this flow?

nealchandra:
[...]

PR #2895, opened 2024-12-19
fix(vercel-integration): vercel integration sync logic

DanielHougaard:
# Description 📣

This PR addresses https://github.com/Infisical/infisical/issues/431. Currently the Vercel integration will override all secrets in Vercel once a sync from Infisical happens. This means if you have any existing variables outside of Infisical, they will be removed from your Vercel environment variables. This is dangerous, because it can potentially destroy the entire flow of a user's existing setup.

I've added support for two sync behaviors. Overwrite target, and prefer source. As discussed with @maidul98, adding these two is ideal for the time being.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/pages/integrations/vercel/create.tsx
+++ b/frontend/src/pages/integrations/vercel/create.tsx
@@ -13,6 +13,7 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import queryString from "query-string";
 
 import { useCreateIntegration } from "@app/hooks/api";
+import { IntegrationSyncBehavior } from "@app/hooks/api/integrations/types";
 
 import {
   Button,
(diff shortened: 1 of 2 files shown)


frontend/src/pages/integrations/vercel/create.tsx
```diff
@@ -36,12 +37,26 @@ const vercelEnvironments = [
...
+
 export default function VercelCreateIntegrationPage() {
   const router = useRouter();
   const { mutateAsync } = useCreateIntegration();
 
   const [selectedSourceEnvironment, setSelectedSourceEnvironment] = useState("");
   const [secretPath, setSecretPath] = useState("/");
+  const [initialSyncBehavior, setInitialSyncBehavior] = useState<IntegrationSyncBehavior>(
+    IntegrationSyncBehavior.OVERWRITE_TARGET
+  );
```
sheensantoscapadngan, line 59: maybe let's default to `prefer source` so total overwrite is somehow a more explicit choice?

frontend/src/pages/integrations/vercel/create.tsx
```diff
@@ -36,12 +37,26 @@ const vercelEnvironments = [
...
+];
+
 export default function VercelCreateIntegrationPage() {
   const router = useRouter();
   const { mutateAsync } = useCreateIntegration();
 
   const [selectedSourceEnvironment, setSelectedSourceEnvironment] = useState("");
   const [secretPath, setSecretPath] = useState("/");
+  const [initialSyncBehavior, setInitialSyncBehavior] = useState<IntegrationSyncBehavior>(
+    IntegrationSyncBehavior.OVERWRITE_TARGET
```
maidul98, line 58: i thought default would be PREFER_SOURCE?

sheensantoscapadngan (approved):

PR #2956, opened 2025-01-08
fix: k8's manifest out of sync

DanielHougaard:
# Description 📣

This PR updates the k8's manifest to be in sync with the operator.

Generated with
```
./bin/kustomize build ./config/default > ./kubectl-install/install-secrets-operator.yaml
```

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/k8-operator/kubectl-install/install-secrets-operator.yaml
+++ b/k8-operator/kubectl-install/install-secrets-operator.yaml
@@ -79,7 +79,6 @@ spec:
                     - identityId
                     type: object
                   kubernetesAuth:
-                    description: Rest of your types should be defined similarly...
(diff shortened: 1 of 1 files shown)


maidul98 (commented):
i don't think this includes things like tolerations https://github.com/Infisical/infisical/blob/main/helm-charts/secrets-operator/templates/deployment.yaml#L91 along with other settings available in helm. Unsure if this defit will cause issues

maidul98 (approved):
People have full control over the manifest, so they can add tolerations, etc on their own

PR #2972, opened 2025-01-10
misc: add support for array values in OIDC aud field

sheensantoscapadngan:
# Description 📣
This PR ensures that the OIDC auth method can be used with Kubernetes setups by enabling the aud field to support an array input (which some systems do)

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/identity-oidc-auth/identity-oidc-auth-fns.ts
+++ b/backend/src/services/identity-oidc-auth/identity-oidc-auth-fns.ts
@@ -2,3 +2,11 @@ import picomatch from "picomatch";
 
 export const doesFieldValueMatchOidcPolicy = (fieldValue: string, policyValue: string) =>
   policyValue === fieldValue || picomatch.isMatch(fieldValue, policyValue);
+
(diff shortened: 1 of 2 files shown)


backend/src/services/identity-oidc-auth/identity-oidc-auth-fns.ts
```diff
@@ -2,3 +2,11 @@ import picomatch from "picomatch";
 
 export const doesFieldValueMatchOidcPolicy = (fieldValue: string, policyValue: string) =>
   policyValue === fieldValue || picomatch.isMatch(fieldValue, policyValue);
+
+export const doesAudValueMatchOidcPolicy = (fieldValue: string | string[], policyValue: string) => {
+  if (Array.isArray(fieldValue)) {
+    return fieldValue.some((entry) => entry === policyValue || picomatch.isMatch(entry, policyValue));
```
maidul98, line 8: Q: why do we need pico match here when we already do `entry === policyValue`?

maidul98 (approved):

PR #2974, opened 2025-01-10
feat(audit-logs): shared secrets audit logs

DanielHougaard:
# Description 📣

This PR introduces support for secret sharing audit logs. The 3 new events are:

1. Shared secret created
2. Shared secret deleted
3. Shared secret read

The created/deleted events will only happen if the secret is created within an organization. And the read event will be logged regardless if it's a public user, or an authenticated organization member reading the secret. This means we now have a new actor type called `Unknown User`. The Unknown user actor type will be used when the actor isn't authenticated.

**Smaller:**
I also added `isClearable` to the project selector in the audit logs tab, and removed the useEffect that made it default to a project on load.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/ee/services/audit-log/audit-log-types.ts
+++ b/backend/src/ee/services/audit-log/audit-log-types.ts
@@ -31,7 +31,7 @@ export type TListProjectAuditLogDTO = {
 
 export type TCreateAuditLogDTO = {
   event: Event;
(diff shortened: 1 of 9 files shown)


scott-ray-wilson (approved):
LGTM!

PR #2975, opened 2025-01-10
feat(api): custom cors settings

DanielHougaard:
# Description 📣

This PR introduces support for custom CORS settings, defined in the environment variables of the instance. Currently we allow for customization of the allowed origins, and allowed headers.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/lib/config/env.ts
+++ b/backend/src/lib/config/env.ts
@@ -199,7 +199,29 @@ const envSchema = z
     INF_APP_CONNECTION_GITHUB_APP_CLIENT_SECRET: zpStr(z.string().optional()),
     INF_APP_CONNECTION_GITHUB_APP_PRIVATE_KEY: zpStr(z.string().optional()),
     INF_APP_CONNECTION_GITHUB_APP_SLUG: zpStr(z.string().optional()),
-    INF_APP_CONNECTION_GITHUB_APP_ID: zpStr(z.string().optional())
+    INF_APP_CONNECTION_GITHUB_APP_ID: zpStr(z.string().optional()),
+
+    /* CORS ----------------------------------------------------------------------------- */
+
+    CORS_ALLOWED_ORIGINS: zpStr(
+      z
+        .string()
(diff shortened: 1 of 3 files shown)


backend/src/server/app.ts
```diff
@@ -87,7 +87,16 @@ export const main = async ({ db, hsmModule, auditLogDb, smtp, logger, queue, key
 
     await server.register<FastifyCorsOptions>(cors, {
       credentials: true,
-      origin: appCfg.SITE_URL || true
+      ...(appCfg.CORS_ALLOWED_ORIGINS?.length
+        ? {
+            origin: appCfg.CORS_ALLOWED_ORIGINS
+          }
+        : {
+            origin: appCfg.SITE_URL || true
```
maidul98, line 95: the site url should always be apart of the list og origins even if CORS_ALLOWED_ORIGINS is specified? Otherwise the frontend wouldn't work. We want that to always work

maidul98 (commented):
just one comment

backend/src/lib/config/env.ts
```diff
@@ -199,7 +199,29 @@ const envSchema = z
...
+
+    /* CORS ----------------------------------------------------------------------------- */
+
+    CORS_ALLOWED_ORIGINS: zpStr(
+      z
+        .string()
+        .optional()
+        .transform((val) => {
+          if (!val) return undefined;
+          return JSON.parse(val) as string[];
```
maidul98, line 212: would this be a problem because val could be anything? what if it isn't a array, then what?

DanielHougaard: It'll throw an error if it's a incorrectly formatted, which will then cause the instance to not start up. I think this behavior is acceptable, because the user will know something is wrong right away. What do you propose?

maidul98 (approved):

PR #2979, opened 2025-01-13
Refactored permission service for project to have project operation type validation

akhilmhdh:
# Description 📣

Previously, we used to check a project operation is correctly for the respective project type using an additional `ForbidOnInvalidProject` function from permission service. The issue with this approach is folks may forget to add it. Thus refactoring the permission functiion to accept an object with all keys and operation type key. `Global` key means an operation that can be made against any type.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/services/access-approval-policy/access-approval-policy-service.ts
+++ b/backend/src/ee/services/access-approval-policy/access-approval-policy-service.ts
@@ -1,6 +1,6 @@
 import { ForbiddenError } from "@casl/ability";
 
-import { ProjectOperationType } from "@app/db/schemas";
+import { ActionProjectType } from "@app/db/schemas";
 import { TPermissionServiceFactory } from "@app/ee/services/permission/permission-service";
 import { ProjectPermissionActions, ProjectPermissionSub } from "@app/ee/services/permission/project-permission";
 import { BadRequestError, ForbiddenRequestError, NotFoundError } from "@app/lib/errors";
@@ -93,7 +93,7 @@ export const accessApprovalPolicyServiceFactory = ({
       projectId: project.id,
       actorAuthMethod,
       actorOrgId,
-      projectOperationType: ProjectOperationType.SecretManager
+      projectOperationType: ActionProjectType.SecretManager
     });
 
     ForbiddenError.from(permission).throwUnlessCan(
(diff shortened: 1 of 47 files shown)


backend/src/services/project/project-service.ts
```diff
@@ -841,14 +863,14 @@ export const projectServiceFactory = ({
...
+    const { permission } = await permissionService.getProjectPermission({
       actor,
       actorId,
       projectId,
       actorAuthMethod,
-      actorOrgId
-    );
-    ForbidOnInvalidProjectType(ProjectType.CertificateManager);
+      actorOrgId,
+      projectOperationType: ProjectOperationType.Global
```
maidul98, line 872: why does this say global when before it used to be for cert manager? some for the ones below

maidul98 (commented):
Some minor comments

backend/src/services/project/project-service.ts
akhilmhdh: Good catch. Forgot tony had cert and ssh in project service. Fixxed it

backend/src/ee/services/access-approval-policy/access-approval-policy-service.ts
```diff
@@ -193,7 +193,14 @@ export const accessApprovalPolicyServiceFactory = ({
...
 
     // Anyone in the project should be able to get the policies.
-    await permissionService.getProjectPermission(actor, actorId, project.id, actorAuthMethod, actorOrgId);
+    await permissionService.getProjectPermission({
+      actor,
+      actorId,
+      projectId: project.id,
+      actorAuthMethod,
+      actorOrgId,
+      projectOperationType: ProjectOperationType.SecretManager
```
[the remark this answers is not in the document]
akhilmhdh: Updated

backend/src/ee/services/access-approval-request/access-approval-request-service.ts
```diff
@@ -106,7 +106,7 @@ export const accessApprovalRequestServiceFactory = ({
       projectId: project.id,
       actorAuthMethod,
       actorOrgId,
-      projectOperationType: ProjectOperationType.SecretManager
+      projectOperationType: ActionProjectType.SecretManager
```
maidul98, line 109: nit: you might wan to update the param too: `projectOperationType` => projectType?

maidul98 (approved):

PR #2986, opened 2025-01-14
fix: addressed cloudflare pages error 304

sheensantoscapadngan:
# Description 📣
This PR ensures that error 304s for the redeploy process will be treated as a successful sync attempt. This also ensures that the auto redeploy option only applies for production syncs as per the documentation [here](https://developers.cloudflare.com/api/resources/pages/subresources/projects/subresources/deployments/methods/create/)

## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/integration-auth/integration-sync-secret.ts
+++ b/backend/src/services/integration-auth/integration-sync-secret.ts
@@ -3764,6 +3764,7 @@ const syncSecretsCloudflarePages = async ({
       )
       .catch((error) => {
         if (error instanceof AxiosError && error.response?.status === 304) {
+          logger.info("CF pages redeployment returned status code 304");
           return;
         }
 
--- a/backend/src/services/integration-auth/integration-sync-secret.ts
+++ b/backend/src/services/integration-auth/integration-sync-secret.ts
@@ -3711,11 +3711,10 @@ const syncSecretsCloudflarePages = async ({
   const getSecretsRes = (
     await request.get<{
       result: { deployment_configs: Record<string, { env_vars: Record<string, unknown> }> };
-    }>(`${IntegrationUrls.CLOUDFLARE_PAGES_API_URL}/client/v4/accounts/${accessId}/pages/projects/${integration.app}?_=${Date.now()}`, {
+    }>(`${IntegrationUrls.CLOUDFLARE_PAGES_API_URL}/client/v4/accounts/${accessId}/pages/projects/${integration.app}`, {
       headers: {
         Authorization: `Bearer ${accessToken}`,
-        Accept: "application/json",
-        "Cache-Control": "no-store"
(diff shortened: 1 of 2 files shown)


backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -3751,17 +3750,25 @@ const syncSecretsCloudflarePages = async ({
...
   const metadata = z.record(z.any()).parse(integration.metadata);
-  if (metadata.shouldAutoRedeploy) {
-    await request.post(
-      `${IntegrationUrls.CLOUDFLARE_PAGES_API_URL}/client/v4/accounts/${accessId}/pages/projects/${integration.app}/deployments`,
-      {},
-      {
-        headers: {
-          Authorization: `Bearer ${accessToken}`,
-          Accept: "application/json"
+  if (metadata.shouldAutoRedeploy && integration.targetEnvironment === "production") {
```
maidul98, line 3753: why does it have to be prod env? can't they choose to sync dev secrets too

backend/src/services/integration-auth/integration-sync-secret.ts
```diff
@@ -3751,17 +3750,25 @@ const syncSecretsCloudflarePages = async ({
...
+            Authorization: `Bearer ${accessToken}`,
+            Accept: "application/json"
+          }
         }
-      }
-    );
+      )
+      .catch((error) => {
+        if (error instanceof AxiosError && error.response?.status === 304) {
+          return;
```
maidul98, line 3767: I think we should add a log for this case so we can track it. add useful metadata to the log so we can track it

backend/src/services/integration-auth/integration-sync-secret.ts
sheensantoscapadngan: the way we do auto-deployment right now is via:
`post /accounts/{account_id}/pages/projects/{project_name}/deployment`

which is only for triggering production deployments. if integration sync is supposed to be targetting "preview" environment then it shouldn't trigger a production deployment. 

image

sheensantoscapadngan: this also somehow contributed to the error 304 - when you have one integration synced to preview env and one integration synced to production env in CF, then they would both attempt to redeploy even though only the prod integration would be valid

backend/src/services/integration-auth/integration-sync-secret.ts
maidul98:
```
logger.info("syncSecretsCloudflarePages: CF pages redeployment returned status code 304 for integration [id=${integration}]")
```

PR #2997, opened 2025-01-16
feat(groups): unique names

DanielHougaard:
# Description 📣

This PR adds unique names to groups within organizations. The same group name can be used in different organizations, but one organization can only have one group with the same name.

The migration for this factors in the possibility that there already exists groups with duplicate names, which we need to resolve before adding the new unique constraint. We do this by suffixing `(<number>)` at the end of the duplicate group names.

This PR serves as a prerequisite to the Terraform PR, https://github.com/Infisical/terraform-provider-infisical/pull/88

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/ee/services/group/group-service.ts
+++ b/backend/src/ee/services/group/group-service.ts
@@ -32,7 +32,7 @@ type TGroupServiceFactoryDep = {
   userDAL: Pick<TUserDALFactory, "find" | "findUserEncKeyByUserIdsBatch" | "transaction" | "findOne">;
   groupDAL: Pick<
     TGroupDALFactory,
-    "create" | "findOne" | "update" | "delete" | "findAllGroupPossibleMembers" | "findById"
+    "create" | "findOne" | "update" | "delete" | "findAllGroupPossibleMembers" | "findById" | "transaction"
   >;
   groupProjectDAL: Pick<TGroupProjectDALFactory, "find">;
   orgDAL: Pick<TOrgDALFactory, "findMembership" | "countAllOrgMembers">;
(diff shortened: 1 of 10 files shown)


backend/src/ee/services/group/group-service.ts
```diff
@@ -88,6 +88,13 @@ export const groupServiceFactory = ({
...
+    const existingGroup = await groupDAL.findOne({ orgId: actorOrgId, name });
+    if (existingGroup) {
+      throw new BadRequestError({
+        message: `Failed to create group with name '${name}'. Group with the same name already exists`
+      });
+    }
+
     const group = await groupDAL.create({
       name,
       slug: slug || slugify(`${name}-${alphaNumericNanoId(4)}`),
```
sheensantoscapadngan, line 100: let's put this as well as the `create group` below into a transaction block so that the DB read request is always served by the primary node and not by replicas

backend/src/ee/services/group/group-service.ts
```diff
@@ -145,6 +152,16 @@ export const groupServiceFactory = ({
...
 
+    if (name) {
+      const existingGroup = await groupDAL.findOne({ orgId: actorOrgId, name });
+
+      if (existingGroup && existingGroup.id !== id) {
+        throw new BadRequestError({
+          message: `Failed to update group with name '${name}'. Group with the same name already exists`
+        });
+      }
+    }
```
sheensantoscapadngan, line 163: same for this one here

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -790,6 +790,18 @@ export const scimServiceFactory = ({
       });
 
     const newGroup = await groupDAL.transaction(async (tx) => {
+      const conflictingGroup = await groupDAL.findOne({
+        name: displayName,
+        orgId
+      });
```
sheensantoscapadngan, line 796: add tx

backend/src/server/routes/v2/group-project-router.ts
```diff
@@ -30,7 +30,7 @@ export const registerGroupProjectRouter = async (server: FastifyZodProvider) =>
       ],
       params: z.object({
         projectId: z.string().trim().describe(PROJECTS.ADD_GROUP_TO_PROJECT.projectId),
-        groupId: z.string().trim().describe(PROJECTS.ADD_GROUP_TO_PROJECT.groupId)
+        groupIdOrName: z.string().trim().describe(PROJECTS.ADD_GROUP_TO_PROJECT.groupId)
```
sheensantoscapadngan, line 33: let's update the description correspondingly?

backend/src/services/group-project/group-project-service.ts
```diff
@@ -128,7 +138,7 @@ export const groupProjectServiceFactory = ({
     const projectGroup = await groupProjectDAL.transaction(async (tx) => {
       const groupProjectMembership = await groupProjectDAL.create(
         {
-          groupId: group.id,
+          groupId: group!.id,
```
sheensantoscapadngan, line 141: is `!` here still necessary? looks like we're already doing a null/undefined check above

backend/src/services/group-project/group-project-service.ts
```diff
@@ -163,7 +173,7 @@ export const groupProjectServiceFactory = ({
       // share project key with users in group that have not
       // individually been added to the project and that are not part of
       // other groups that are in the project
-      const groupMembers = await userGroupMembershipDAL.findGroupMembersNotInProject(group.id, project.id, tx);
+      const groupMembers = await userGroupMembershipDAL.findGroupMembersNotInProject(group!.id, project.id, tx);
```
sheensantoscapadngan, line 176: same here

sheensantoscapadngan (approved):

PR #3007, opened 2025-01-18
Base64 decode in operator

akhilmhdh:
# Description 📣

This PR provides a new template function for infisical operator. This is used to convert base64 encoded string to a binary file content in k8s operator.

Without this function scenario:
1. Customer has a binary file (file1)
2. They encode it to base64 to store in Infisical (1st base64 encoding)
3. When this gets stored in Kubernetes secrets, it gets base64 encoded again (2nd base64 encoding)

The problem:
1. When Kubernetes reads the secret, it decodes the outer base64 layer
2. But you're left with the content still base64 encoded (from the Infisical step)
3. Customer doesn't want to have to add an extra step in their pods to decode this remaining base64 layer to consume it as binary data

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation


maidul98 (changes requested):
The docs were unclear so i went ahead and improved it. Remember to put as much care into the docs need we do to the code. Other notes:

- You forgot to add the base64 function to the create managed secret function. I went ahead and did that. Make sure to test your PR before you raise it because this is something I think you would have caught during testing 

- The CRD currently only reconciles when the etag of the secrets change. We should update the logic so that it reconciles when we make template changes too in the CRD

maidul98 (approved):

Issue #3047, opened 2025-01-27
Tooltips are not directly above the icons in a secret.


isaiahmartin847:
### Describe the Bug  
This is a UI bug specific to each secret. The tooltips are off-center to the right. Additionally, the 'Copy Key' icon should always be visible, rather than appearing only on hover.

### To Reproduce  
1. Go to `secret-manager/[project-id]/overview`.  
2. Click on a secret.  
3. Hover over the copy name, and you will notice that the tooltips are not centered.  
4. You will see that they are not properly aligned.

### Expected Behavior  
The tooltips should be directly above the icons.

### Screenshots  

![Image](https://github.com/user-attachments/assets/fd88a311-fb32-4c69-b6ca-3c4b680d7280)

![Image](https://github.com/user-attachments/assets/893f35e5-30a2-4d9b-b87f-c85f23daa0ca)

![Image](https://github.com/user-attachments/assets/2fdbb530-368b-4a11-8d04-13ee33d5d2ac)

### Platform  
This is on the Infisical web app.

### Additional Context  
I think I can take this one. Please assign it to me. I should have a PR ready by the end of the day.

PR #3048, opened 2025-01-27
Improve Visibility and Alignment of Tooltips and Copy Secret Key Icon

isaiahmartin847:
# Description 📣

issue: #3047 

In this PR, I updated two things. The first is to make the `Copy Secret Key` icon visible at all times instead of only on hover. I think this will improve usability, being a user myself, I didn't know this would happen until I accidentally hovered over it. In addition, I noticed that the tooltips were not inline with the Copy icon, and when you change the secret name, the Save and Cancel icon tooltips were also not aligned.

### Changes made in `SecretRenameRow.tsx`
In line 198, I simply changed the opacity to 100 and removed the hover property.
I addressed the alignment issue by wrapping each Tooltip with a `div` that has the `relative` property, which fixed the offset bug.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

I tested my changes by running the dev environment and making sure that the tooltips were centered over the icons. I also resized the window width to ensure that they didn't revert back to being offset.

## Note
If you believe that the `Copy Secret Key` icon should only show up on hover, please inform me, and I can change it back to how it was previously.
	
### DIFF
--- a/frontend/src/pages/secret-manager/OverviewPage/components/SecretOverviewTableRow/SecretRenameRow.tsx
+++ b/frontend/src/pages/secret-manager/OverviewPage/components/SecretOverviewTableRow/SecretRenameRow.tsx
@@ -195,7 +195,7 @@ function SecretRenameRow({ environments, getSecretByKey, secretKey, secretPath }
                       ariaLabel="copy-value"
                       variant="plain"
                       size="sm"
-                      className="p-0 opacity-0 group-hover:opacity-100"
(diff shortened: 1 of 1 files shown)


akhilmhdh (approved):

akhilmhdh:
Thank you @isaiahmartin847

Issue #3034, opened 2025-01-23
The workspace limit does not dynamically update when projects are created or deleted


moizsheraz:
### Describe the bug
The workspace limit does not dynamically update when projects are created or deleted causing  a wrong Popup showing when user clicks on `Create New Project` btn

### Expected behavior
On Creation or deletion of a project a Workspace limit should be updated dynamically without reloading the page .. 

### Screenshots
I think this recording is enough to show what is going wrong
https://github.com/user-attachments/assets/56554b2a-6dfa-404b-aac8-1e21876fe02c

isaiahmartin847:
Interesting! I also found that if you delete your third project and then try to create another, you'll encounter a paywall even though you only have two projects. I plan to dig into the code tomorrow and find a fix.

moizsheraz:
@isaiahmartin847 sure go ahead

isaiahmartin847:
@vmatsiiako, could you please assign this to me? Thank you.

PR #3051, opened 2025-01-28
Fixed the frontend subscription data not syncing with the backend

isaiahmartin847:
# Description 📣

Issue: #3034 

This pull request addresses the bug mentioned in the issue, but it also resolves additional bugs that were previously unnoticed. Let me explain:

  

On the first load, the client fetches the subscription data. However, as you create new projects and navigate within the app, it never fetches the subscription data again. For example, if you created your last project under the free tier and added some secrets, then returned to the organization overview, you would still be able to open the dialog to create a new project—even if you had already hit your limit. This occurs because the front end becomes out of sync with the back end.

  

This fix also resolves a range of other unknown issues related to the subscription. For instance, consider a scenario where **Admin1** and **Admin2** are logged in simultaneously. If **Admin2** adds the member limit to a project, **Admin1** would still be able to open the "Add New Member" dialog, fill it out, and submit it—even though the back end recognizes that the project has reached its member limit. This would result in an error.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

If you have any questions, please let me know. I would be happy to address them.


frontend/src/context/SubscriptionContext/SubscriptionContext.tsx
```diff
@@ -12,7 +12,7 @@ export const useSubscription = () => {
   const { data: subscription } = useSuspenseQuery({
     queryKey: subscriptionQueryKeys.getOrgSubsription(organizationId),
     queryFn: () => fetchOrgSubscription(organizationId),
-    staleTime: Infinity
+    staleTime: 0
```
maidul98, line 15: instead of setting it to zero, we can invalidate it at project create/delete

isaiahmartin847: Thanks for the input! I can make that change. But what about all the other possible issues? Do you think they are unlikely to occur?

akhilmhdh (approved):
Looking good to me. Approved! Good job @isaiahmartin847

olekgolus11:
hey, is the issue going to be resolved?

PR #3052, opened 2025-01-28
Feature: Setup Password

scott-ray-wilson:
# Description 📣

This PR adds support to setup a password for users who signed up via methods other than email/password. Also updates reset password from backup key to store hashed password.

PR also includes adjustments to Secret Syncs layout to conform to nav restructure.

CleanShot 2025-01-27 at 18 52 41@2x

CleanShot 2025-01-27 at 18 52 01@2x

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/auth/auth-password-service.ts
+++ b/backend/src/services/auth/auth-password-service.ts
@@ -4,6 +4,8 @@ import jwt from "jsonwebtoken";
 import { SecretEncryptionAlgo, SecretKeyEncoding } from "@app/db/schemas";
 import { getConfig } from "@app/lib/config/env";
 import { generateSrpServerKey, srpCheckClientProof } from "@app/lib/crypto";
+import { BadRequestError } from "@app/lib/errors";
+import { OrgServiceActor } from "@app/lib/types";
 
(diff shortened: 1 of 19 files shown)


backend/src/services/auth/auth-password-service.ts
```diff
@@ -267,13 +280,115 @@ export const authPaswordServiceFactory = ({
...
+
+      if (!user) throw new BadRequestError({ message: `Could not find user with ID ${actor.id}` });
+
+      if (!user.isAccepted || !user.authMethods)
+        throw new BadRequestError({ message: `You must complete signup to set a password` });
+
+      await userDAL.updateById(
+        actor.id,
+        {
+          authMethods: [...user.authMethods, AuthMethod.EMAIL]
```
akhilmhdh, line 351: What if there is already an `AuthMethod.Email`

akhilmhdh (changes requested):
I think you may also need to tell users they must be logged in do this- 

Because felt it quite confusing, when it's was just back to login page.

akhilmhdh (approved):

PR #3060, opened 2025-01-28
feat(api): list secrets filter by metadata

DanielHougaard:
# Description 📣

This PR adds support for filtering secrets by their metadata. If multiple metadata fields are passed in, the operation will work as an `AND` operator, which means the secrets returned will match all the metadata fields.

The format for the `secretMetadata` query parameter is `key1:value1,key2:value2`.

The metadata filter is limited to max 10 metadata key-value pairs. This is to avoid overloading the backend, and to avoid bad actors passing into a list with millions of key-value pairs, which could result in a DoS attack.
```typescript
if (metadata.length > 10) {
                ctx.addIssue({
                  code: z.ZodIssueCode.custom,
                  message: "You can only filter by up to 10 metadata fields"
                });
              }
```

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -688,7 +688,9 @@ export const RAW_SECRETS = {
     environment: "The slug of the environment to list secrets from.",
     secretPath: "The secret path to list secrets from.",
     includeImports: "Weather to include imported secrets or not.",
-    tagSlugs: "The comma separated tag slugs to filter secrets."
+    tagSlugs: "The comma separated tag slugs to filter secrets.",
+    secretMetadata:
+      "The secret metadata key-value pairs to filter secrets by. When querying for multiple metadata pairs, the query is treated as an AND operation. Secret metadata format is key1:value1,key2:value2."
   },
   CREATE: {
     secretName: "The name of the secret to create.",
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -689,8 +689,8 @@ export const RAW_SECRETS = {
     secretPath: "The secret path to list secrets from.",
     includeImports: "Weather to include imported secrets or not.",
     tagSlugs: "The comma separated tag slugs to filter secrets.",
-    secretMetadata:
-      "The secret metadata key-value pairs to filter secrets by. When querying for multiple metadata pairs, the query is treated as an AND operation. Secret metadata format is key1:value1,key2:value2."
+    metadataFilter:
+      "The secret metadata key-value pairs to filter secrets by. When querying for multiple metadata pairs, the query is treated as an AND operation. Secret metadata format is key=value1,value=value2|key=value3,value=value4."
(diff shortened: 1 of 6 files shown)


backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -414,6 +414,22 @@ export const secretV2BridgeDALFactory = (db: TDbClient) => {
           `${TableName.SecretTag}.id`
         )
         .leftJoin(TableName.ResourceMetadata, `${TableName.SecretV2}.id`, `${TableName.ResourceMetadata}.secretId`)
+        .where((bd) => {
+          if (filters?.secretMetadata && filters.secretMetadata?.length > 0) {
+            filters.secretMetadata.forEach((meta) => {
+              void bd.whereExists((qb) => {
```
akhilmhdh, line 420: I don't think you need do this.

`.where((bd)=>)` => `and ()` in knexjs

Then `.where( (bd)=> bd.where((deeperBd)=>).orWhere((deeperbd)=>) )` => .and( |first metadata field| or |second metadata field| )

You don't need exist here and secret id here. As it's already part of left join above.

akhilmhdh (changes requested):

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
DanielHougaard: Right! This is also more performant. Implemented

akhilmhdh (approved):

akhilmhdh (approved):
All good. 

One small thought process was I felt `OR` would be more appropriate than `AND` when multiple combo of `key=key1,value=value1:key=key2,value=value2`

I felt in above `[ key1 and value1 ] or [ key2 and value2 ]` than `[ key1 and value1 ] and [ key2 and value2 ]`. But it may be subjective.

CC: @DanielHougaard @maidul98

backend/src/lib/api-docs/constants.ts
```diff
@@ -688,7 +688,9 @@ export const RAW_SECRETS = {
     environment: "The slug of the environment to list secrets from.",
     secretPath: "The secret path to list secrets from.",
     includeImports: "Weather to include imported secrets or not.",
-    tagSlugs: "The comma separated tag slugs to filter secrets."
+    tagSlugs: "The comma separated tag slugs to filter secrets.",
+    secretMetadata:
+      "The secret metadata key-value pairs to filter secrets by. When querying for multiple metadata pairs, the query is treated as an AND operation. Secret metadata format is key1:value1,key2:value2."
```
maidul98, line 693: docs are not up dated with code: `metadataFilter=key=value1,value=value2|key=value3,value=value4` not `:`

maidul98 (approved):

PR #3069, opened 2025-01-30
Feature: OIDC Group Membership Mapping

scott-ray-wilson:
# Description 📣

This PR adds support for optionally syncing OIDC group memberships to Infisical. When a user logs in they will be added or removed from Infisical groups that match group names present in their token groups claim. Includes docs for Keycloak setup. 

CleanShot 2025-01-30 at 09 23 27@2x

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [x] Documentation

# Tests 🛠️

```sh
```

---


frontend/src/pages/organization/GroupDetailsByIDPage/components/AddGroupMemberModal.tsx
```diff
@@ -82,9 +82,9 @@ export const AddGroupMembersModal = ({ popUp, handlePopUpToggle }: Props) => {
         text: "Successfully assigned user to the group",
         type: "success"
       });
-    } catch {
+    } catch (error) {
       createNotification({
-        text: "Failed to assign user to the group",
+        text: (error as Error)?.message ?? "Failed to assign user to the group",
```
sheensantoscapadngan, line 87: should we be getting the message from the axios response instead? (need to test on my side)

frontend/src/pages/organization/SettingsPage/components/OrgAuthTab/OrgOIDCSection.tsx
```diff
@@ -148,6 +174,63 @@ export const OrgOIDCSection = (): JSX.Element => {
...
+                    href="https://infisical.com/docs/documentation/platform/sso/overview"
+                  >
+                    See your OIDC provider docs for details.
+                  </a>
+                  <p className="mt-4 text-yellow">
+                    <FontAwesomeIcon className="mr-1" icon={faWarning} />
+                    Group membership changes in the OIDC provider only sync with Infisical when a
+                    user logs in. For example, if you remove a user from a group in the OIDC
+                    provider, this change will not be reflected in Infisical until their next login.
+                  </p>
```
sheensantoscapadngan, line 208: hmmm it's kind of tricky because if the user logs in via email, they would still somehow retain the old groups right? 

I wonder if we should make that clearer

sheensantoscapadngan:
for the tooltips, thoughts on explicitly guiding users to manage groups via IdP instead of prompting them to cancel?

backend/src/db/migrations/20250129214629_oidc-configs-manage-group-memberships-col.ts
```diff
@@ -0,0 +1,23 @@
+import { Knex } from "knex";
```
sheensantoscapadngan, line 1: (ignore the file placement of the comment)

let's add audit logs?

frontend/src/pages/organization/GroupDetailsByIDPage/components/AddGroupMemberModal.tsx
scott-ray-wilson: Seems to work as is?

CleanShot 2025-01-31 at 10 09 23@2x

scott-ray-wilson: just removing this change as it shows the error message from the query client

scott-ray-wilson: CleanShot 2025-01-31 at 10 11 47@2x

sheensantoscapadngan (approved):

PR #3079, opened 2025-02-03
(WIP) feat: kmip client management

sheensantoscapadngan:
# Description 📣

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


backend/src/ee/routes/v1/kmip-operation-router.ts
```diff
@@ -0,0 +1,477 @@
...
+    }
+
+    req.kmipUser = {
+      projectId,
+      clientId,
+      name: kmipClient.name
+    };
+  });
+
+  server.route({
```
akhilmhdh, line 68: Regarding the URLs, Is this a kmip convention. If not feels very weird.

backend/src/ee/services/kmip/kmip-org-config-dal.ts
```diff
@@ -0,0 +1,12 @@
+import { TDbClient } from "@app/db";
+import { TableName } from "@app/db/schemas";
+import { ormify } from "@app/lib/knex";
+
+export type TKmipOrgConfigDALFactory = ReturnType<typeof kmipOrgConfigDALFactory>;
+
+export const kmipOrgConfigDALFactory = (db: TDbClient) => {
+  const kmipOrgConfigOrm = ormify(db, TableName.KmipOrgConfig);
+  return {
+    ...kmipOrgConfigOrm
```
akhilmhdh, line 10: Nit: Spread is unnecessary. Like wise for others

backend/src/ee/services/kmip/kmip-service.ts
```diff
@@ -0,0 +1,819 @@
...
+      decryptor({ cipherTextBlob: kmipConfig.encryptedServerIntermediateCaCertificate })
+    );
+
+    const clientIntermediateCaCert = new x509.X509Certificate(
+      decryptor({ cipherTextBlob: kmipConfig.encryptedClientIntermediateCaCertificate })
+    );
+
+    return {
+      id: kmipConfig.id,
+      serverCertificateChain: `${serverIntermediateCaCert.toString("pem")}\n${rootCaCert.toString("pem")}`.trim(),
```
akhilmhdh, line 590: As this is done in multiple places, even in gateway, may be we can make this into a simple reusable function that will be take multiple certificate type arguments and join it like in order.

I know the library already providers a chain builder - but we can have our own simple version

backend/src/ee/services/kmip/kmip-service.ts
```diff
@@ -0,0 +1,819 @@
...
+
+    const altNamesArray: {
+      type: "email" | "dns" | "ip";
+      value: string;
+    }[] = altNames
+      .split(",")
+      .map((name) => name.trim())
+      .map((altName) => {
+        // check if the altName is a valid hostname
+        if (hostnameRegex.test(altName)) {
```
akhilmhdh, line 669: Instead of exposing the regex this would be better like the `isValidIp` - `isValidHostName` function or some more specific on the regex text happening.

Like what kinda of hostname we are looking for.

backend/src/services/kms/kms-key-dal.ts
```diff
@@ -93,6 +93,34 @@ export const kmskeyDALFactory = (db: TDbClient) => {
...
+        .select(
+          db.ref("encryptedKey").withSchema(TableName.InternalKms).as("internalKmsEncryptedKey"),
+          db.ref("encryptionAlgorithm").withSchema(TableName.InternalKms).as("internalKmsEncryptionAlgorithm"),
+          db.ref("version").withSchema(TableName.InternalKms).as("internalKmsVersion")
+        );
+
+      return result.map((entry) => ({
+        ...KmsKeysSchema.parse(entry),
+        isActive: !entry.isDisabled,
+        encryptedKey: entry.internalKmsEncryptedKey,
```
akhilmhdh, line 115: I see this is used in the service layer of `kmip-operator` and in router - but the encrypted keys are not at all used.

Can we avoid passing any sensitive data of kms-dal to other than kms-service? This is because kms service encrypted keys should be very carefully given to others as it the root encryption key of rest.

cli/packages/cmd/kmip.go
```diff
@@ -0,0 +1,103 @@
...
+	Example:               `infisical kmip start`,
+	Short:                 "Used to start a KMIP server",
+	Use:                   "start",
+	DisableFlagsInUseLine: true,
+	Args:                  cobra.NoArgs,
+	Run:                   startKmipServer,
+}
+
+func startKmipServer(cmd *cobra.Command, args []string) {
+	addr, err := cmd.Flags().GetString("addr")
```
akhilmhdh, line 33: A bit more elaborate argument would be helpful. Something popped in my mind was - what was the address for? Server or listening to infisical etc

cli/packages/cmd/kmip.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		util.HandleError(err, "Unable to parse flag")
+	}
+
+	identityAuthMethod, err := cmd.Flags().GetString("identity-auth-method")
```
akhilmhdh, line 38: Why not just directly accept the token like other commands? So the token retrieval  is something the `login` command handles?

frontend/src/layouts/ProjectLayout/ProjectLayout.tsx
```diff
@@ -105,6 +105,20 @@ export const ProjectLayout = () => {
...
                       )}
+                      {isCmek && (
+                        <Link
+                          to={`/${ProjectType.KMS}/$projectId/kmip` as const}
+                          params={{
+                            projectId: currentWorkspace.id
+                          }}
+                        >
+                          {({ isActive }) => (
+                            <MenuItem isSelected={isActive} icon="lock-closed">
```
akhilmhdh, line 116: New icon i guess? Let me know if you have any concept and I'll make one for you quickly

akhilmhdh (changes requested):

backend/src/ee/routes/v1/kmip-operation-router.ts
sheensantoscapadngan: it's not a kmip convention - I made it this way so that it maps one to one to the KMIP operations supported by our KMIP server. Other implementations have the KMIP server and the KMS logic all in one process but this isn't possible for us so we have to do it this way. Basically, we're implementing the logic side of KMIP in Infisical and the KMIP server is just a translation layer

akhilmhdh (approved):
Approved on code wise

PR #3125, opened 2025-02-17
Fix failing redirect to create new organization page on no organizations

dangtony98:
# Description 📣

This PR fixes a bug where users get stuck on an infinitely-loading select organization page when they have deleted all their organizations (or are no longer part of any); the user now gets correctly redirected to the page to create a new organization.

Update: This PR now also fixes another edge-case that is an existing user already logged into a deleted organization getting stuck on an error/spinner screen; in such case, the user token is no longer valid so currently I have them re-login.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

## Summary by CodeRabbit

- **New Features**
  - Introduced a streamlined navigation path for users without an associated organization.
  
- **Bug Fixes**
  - Improved error handling during authentication, logout, and invitation processes to ensure smoother session management.
  
- **Style**
  - Updated the layout and presentation of the organization-less page for enhanced user experience.
  
- **Refactor**
  - Consolidated token validation and session management logic, and simplified routing configuration for more reliable navigation.
  
- **Chores**
  - Enhanced session clearing logic and modularized logout functionality for improved maintainability.


backend/package-lock.json
```diff
@@ -21,7 +21,7 @@
         "@fastify/etag": "^5.1.0",
         "@fastify/formbody": "^7.4.0",
         "@fastify/helmet": "^11.1.1",
-        "@fastify/multipart": "^8.3.1",
+        "@fastify/multipart": "8.3.1",
```
scott-ray-wilson, line 24: curious if this was intentional and if so why?

scott-ray-wilson (changes requested):
Code looks good, was able to create a new org 👍 one potential improvement would be handling this situation:

https://github.com/user-attachments/assets/6483a1d3-4497-467a-9242-ae95a8714d99

And minor nit but I think you might not have your editor configured to use or prettier rules as commits have single quotation and no semicolons

backend/package-lock.json
dangtony98: Unsure; I think this was added after running `npm install` — can bring it back.

akhilmhdh (approved):

scott-ray-wilson (approved):
LGTM!

PR #3174, opened 2025-03-04
Add descriptions to secret folders

carlosmonastyrski:
# Description 📣

This PR introduces a new feature allowing folder-level descriptions to be added to the secret_folder entity. A new optional string field, description, has been added with a maximum length of 255 characters. This field provides a way to document the context of the secrets within a folder, making it easier for teams to organize and navigate secrets effectively.

The UI has been updated to include a new text area for the description field. The input is limited to 255 characters and supports up to 10 lines. This description will be displayed as a tooltip when a user hovers over a folder's name within the environment UI.

Additionally, when using the overview modal to edit a folder, if a custom description was set for a specific environment, the description field will overwrite the environment's description.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

I tested this new flow locally, using Postman to verify the folder-related endpoints, and ensured the corresponding changes were documented accurately for these updated endpoints.

https://www.loom.com/share/65096f224f284c99a23f1bba4deced37?sid=c4c06782-cc8d-465e-a559-9e64ff31d1ee

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - Introduced an optional description field for folders, enhancing metadata management.
  - Updated database structures to support storing folder descriptions.
  - Enhanced API endpoints and frontend components to enable adding, updating, and displaying folder descriptions, including improved tooltips and form interactions.
  - Added support for folder descriptions in audit logs for better tracking.

### DIFF
--- a/frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/FolderForm.tsx
+++ b/frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/FolderForm.tsx
@@ -1,15 +1,21 @@
+import { useRef } from "react";
 import { Controller, useForm } from "react-hook-form";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { z } from "zod";
 
 import { Button, FormControl, Input, ModalClose } from "@app/components/v2";
+import { TextArea } from "@app/components/v2/TextArea/TextArea";
 
 type Props = {
-  onCreateFolder?: (folderName: string) => Promise<void>;
-  onUpdateFolder?: (folderName: string) => Promise<void>;
+  onCreateFolder?: (folderName: string, description: string | undefined) => Promise<void>;
+  onUpdateFolder?: (folderName: string, description: string | undefined) => Promise<void>;
   isEdit?: boolean;
   defaultFolderName?: string;
+  defaultDescription?: string;
+  showDescriptionOverwriteWarning?: boolean;
 };
+
+const descriptionOverwriteWarningMessage = "Warning: Any changes made here will overwrite any custom edits in individual environment folders."
 
 const formSchema = z.object({
   name: z
@@ -18,15 +24,20 @@ const formSchema = z.object({
     .regex(
       /^[a-zA-Z0-9-_]+$/,
       "Folder name can only contain letters, numbers, dashes, and underscores"
-    )
+    ),
+  description: z
+  .string()
+  .optional()
 });
 type TFormData = z.infer<typeof formSchema>;
 
 export const FolderForm = ({
   isEdit,
   defaultFolderName,
+  defaultDescription,
   onCreateFolder,
-  onUpdateFolder
+  onUpdateFolder,
+  showDescriptionOverwriteWarning = false
 }: Props): JSX.Element => {
   const {
     control,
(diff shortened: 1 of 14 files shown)


frontend/src/pages/secret-manager/SecretDashboardPage/components/FolderListView/FolderListView.tsx
```diff
@@ -115,8 +117,14 @@ export const FolderListView = ({
             }}
             onClick={() => handleFolderClick(name)}
           >
-            {name}
-          </div>
+            <Tooltip
```
sheensantoscapadngan, line 120: btw I think having some sort of "i" icon would be preferable here because otherwise it's not really obvious to users that you can actually hover over it to view the description? 

image

Maybe something like the following:
image

frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/FolderForm.tsx
```diff
@@ -61,6 +89,31 @@ export const FolderForm = ({
...
+                {...field}
+                rows={3}
+                ref={descriptionRef}
+                onInput={handleInput}
+                className="thin-scrollbar w-full !resize-none bg-mineshaft-900"
+                maxLength={255}
+              />
+            </FormControl>
+          )}
+        />
```
sheensantoscapadngan, line 116: if a folder already has a description and I intend to remove the description by setting it to "", it would retain the description instead of remove it 
image

image

frontend/src/pages/secret-manager/SecretDashboardPage/components/FolderListView/FolderListView.tsx
carlosmonastyrski: Good point! I'll update it to include an "i" icon so it's to make it clearer that users can hover over it for the description.

frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/FolderForm.tsx
```diff
@@ -36,15 +47,32 @@ export const FolderForm = ({
...
+      const maxDescriptionLines = 10;
+  
+      if (lines.length > maxDescriptionLines) {
+        textarea.value = lines.slice(0, maxDescriptionLines).join("\n");
+      }
+    }
+  };
+
+  const onSubmit = async ({ name, description }: TFormData) => {
+    const descriptionShaped = description && description.trim() !== "" ? description : null;
```
sheensantoscapadngan, line 70: curious, what's the reason why we have to explicitly make it null when string value is ""?

I think an alternative would be to just pass in the "" string to the API which would clear the description value

carlosmonastyrski: That's a valid option too. I set it to null to ensure we only store a description when there's an actual value present. This way, we can clearly distinguish between folders with no description and those that have one. For example, if we use description: "", it could lead to misunderstandings when users see responses like { ... description: "", ... } in Postman, as it might appear as if the description is intentionally empty rather than absent.

sheensantoscapadngan: sounds good!

sheensantoscapadngan (approved):

PR #3192, opened 2025-03-06
Fix: Handle Audit Log Query Duration

scott-ray-wilson:
# Description 📣

This PR increases the audit log endpoint and nginx config timeout to handle query max duration of 2 minutes which was previously timing out at 30 seconds due to server config. Also fixes empty audit log error notifcation.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - The audit logs functionality now supports extended processing times for long-running queries.
  - The audit logs interface has been updated to provide a more accurate loading indicator during data fetching.
  - A new endpoint for audit logs has been introduced, enhancing server API capabilities.
  
- **Bug Fixes**
  - Improved loading state indication for the button in the audit logs interface.

### DIFF
--- a/backend/src/server/routes/v1/organization-router.ts
+++ b/backend/src/server/routes/v1/organization-router.ts
@@ -14,6 +14,7 @@ import { AUDIT_LOGS, ORGANIZATIONS } from "@app/lib/api-docs";
 import { getLastMidnightDateISO, removeTrailingSlash } from "@app/lib/fn";
 import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
 import { slugSchema } from "@app/server/lib/schemas";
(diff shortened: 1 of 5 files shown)


backend/src/server/routes/v1/organization-router.ts
```diff
@@ -108,6 +109,7 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
     config: {
       rateLimit: readLimit
     },
+    preHandler: extendTimeout(QUERY_TIMEOUT),
```
akhilmhdh, line 112: I think we should find out the query that causing the latency rather than increasing the timeout.

PR #3215, opened 2025-03-10
fix: change dd-mm-yy to mm-dd-yy

mahyarmirrashed:
# Description 📣

Changed the date format from `DD-MM-YYYYY` to `MM-DD-YYYY` on Audit Logs page.

image

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

No tests needed.

## Summary by CodeRabbit

- **Style**
	- Updated the date display format on the Audit Logs page to show month-first ordering for improved clarity.
	- Changed the header for the timestamp in the logs table to specify the format as "Timestamp (MM/DD/YYYY)" for clearer user understanding.

### DIFF
--- a/frontend/src/pages/organization/AuditLogsPage/components/LogsTable.tsx
+++ b/frontend/src/pages/organization/AuditLogsPage/components/LogsTable.tsx
@@ -28,7 +28,7 @@ type Props = {
 
 const AUDIT_LOG_LIMIT = 15;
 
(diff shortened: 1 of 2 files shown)


mahyarmirrashed:
> are we not adding the small label to show them what the format is? i think we should. this would resolve all confusion across region

Updated the picture to match with the updated changes.

PR #3226, opened 2025-03-11
Add proper support for systemd 

maidul98:
There wasn't a great way to start the gateway with systemd so that it can run in the background and be managed by systemd. This pr addeds a install sub command that decouples install from running. The goal was so you can run something like this in your IaC:

```infisical gateway install --token=<> --domain=<> && systemctl start infisical-gateway```

## Summary by CodeRabbit

- **New Features**
  - Introduced a new CLI subcommand for installing and enabling the Infisical gateway as a system service, complete with token and domain configuration for seamless connection. Enhanced error handling ensures proper usage on supported platforms.
  - Added comprehensive documentation for the Infisical gateway, detailing operation, management, and installation as a systemd service.

- **Improvements**
  - Updated the service configuration to locate the gateway executable using the system’s PATH, simplifying setup and management.
  - Added a new command to the CLI documentation, expanding the functionality of the CLI.

### CODE REVIEW

File: cli/packages/cmd/gateway.go [hunk 1]
```diff
@@ -110,6 +104,50 @@ var gatewayCmd = &cobra.Command{
...
+			util.HandleError(fmt.Errorf("systemd service installation is only supported on Linux"))
+		}
+
+		if os.Geteuid() != 0 {
+			util.HandleError(fmt.Errorf("systemd service installation requires root/sudo privileges"))
+		}
+
+		token, err := util.GetInfisicalToken(cmd)
+		if err != nil {
+			util.HandleError(err, "Unable to parse flag")
```
DanielHougaard, line 125: Nit: The error returned from this function will always be related to parsing the token cobra flag. So we should change `Unable to parse flag` to `Unable to parse token flag`

cli/packages/gateway/systemd.go
```diff
@@ -17,7 +17,7 @@ After=network.target
 [Service]
 Type=simple
 EnvironmentFile=/etc/infisical/gateway.conf
-ExecStart=/usr/local/bin/infisical gateway
+ExecStart=infisical gateway
```
DanielHougaard, line 20: Shouldn't this contain an absolute path? `/usr/bin/env infisical gateway`

DanielHougaard (changes requested):

cli/packages/gateway/systemd.go
maidul98: no, we tell people to use the cli script to install which will always install in the bin of the machine. so the cli should ideally be in the path already

cli/packages/cmd/gateway.go
maidul98: addressed

DanielHougaard (approved):

PR #3257, opened 2025-03-17
Add file option to secret set CLI command to retrieve secrets from .env or .yaml files

carlosmonastyrski:
# Description 📣

Updated the CLI secret set command to support retrieving secrets from both .env and .yaml files. Also, improved the documentation to reflect these changes.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - Introduced a `--file` option for the `infisical secrets set` command, enabling users to load secrets from external files.
  - Enhanced secret management with support for both YAML and `.env` formats, including validation of secret keys.

- **Bug Fixes**
  - Improved error handling for issues related to the "file" flag during command execution.

- **Documentation**
  - Updated documentation to include details on the `--file` flag, usage examples, and supported file formats for setting secrets.

### DIFF
--- a/cli/packages/cmd/secrets.go
+++ b/cli/packages/cmd/secrets.go
@@ -143,7 +143,12 @@ var secretsSetCmd = &cobra.Command{
 	Short:                 "Used set secrets",
 	Use:                   "set [secrets]",
 	DisableFlagsInUseLine: true,
-	Args:                  cobra.MinimumNArgs(1),
+	Args: func(cmd *cobra.Command, args []string) error {
+		if cmd.Flags().Changed("file") {
+			return cobra.MinimumNArgs(0)(cmd, args)
+		}
+		return cobra.MinimumNArgs(1)(cmd, args)
+	},
 	Run: func(cmd *cobra.Command, args []string) {
 		token, err := util.GetInfisicalToken(cmd)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cli/packages/cmd/secrets.go
```diff
@@ -691,6 +701,7 @@ func init() {
 	secretsSetCmd.Flags().String("projectId", "", "manually set the project ID to for setting secrets when using machine identity based auth")
 	secretsSetCmd.Flags().String("path", "/", "set secrets within a folder path")
 	secretsSetCmd.Flags().String("type", util.SECRET_TYPE_SHARED, "the type of secret to create: personal or shared")
+	secretsSetCmd.Flags().String("file", "", "Path to a file with key=value secret pairs (comments with # or // are ignored)")
```
DanielHougaard, line 704: This should mention that the file types supported are yaml and `.env` files

cli/packages/util/secrets.go
```diff
@@ -563,7 +564,99 @@ func GetPlainTextWorkspaceKey(authenticationToken string, receiverPrivateKey str
...
+	}
+
+	return secrets, nil
+}
+
+func validateSecretKey(key string) error {
+	if key == "" {
+		return errors.New("secret keys cannot be empty")
+	}
+	if unicode.IsNumber(rune(key[0])) {
```
DanielHougaard, line 621: Why can't secret keys start with numbers?

cli/packages/cmd/secrets.go
```diff
@@ -143,7 +143,12 @@ var secretsSetCmd = &cobra.Command{
 	Short:                 "Used set secrets",
 	Use:                   "set [secrets]",
 	DisableFlagsInUseLine: true,
-	Args:                  cobra.MinimumNArgs(1),
+	Args: func(cmd *cobra.Command, args []string) error {
+		if cmd.Flags().Changed("file") {
+			return cobra.MinimumNArgs(0)(cmd, args)
+		}
```
DanielHougaard, line 149: Currently this allows people to use both command-line arguments **and** file at the same time. I think we should make it explicit so the user can only use file **or** command-line arguments

DanielHougaard (changes requested):

cli/packages/util/secrets.go
carlosmonastyrski: That validation was added to match the existing logic a few lines below, where a failure would occur if a secret key started with a number.

CleanShot 2025-03-18 at 08 20 20@2x

cli/packages/cmd/secrets.go
```diff
@@ -143,7 +143,12 @@ var secretsSetCmd = &cobra.Command{
 	Short:                 "Used set secrets",
 	Use:                   "set [secrets]",
 	DisableFlagsInUseLine: true,
-	Args:                  cobra.MinimumNArgs(1),
+	Args: func(cmd *cobra.Command, args []string) error {
+		if cmd.Flags().Changed("file") {
+			return cobra.ExactArgs(0)(cmd, args)
```
DanielHougaard, line 148: Please throw an error if the user is trying to pass command-line arguments and the `file` flag at the same time, just so it's a little more clear why it's failing

DanielHougaard (changes requested):

DanielHougaard (approved):

PR #3264, opened 2025-03-17
Show when folders have imports in secret overview page

carlosmonastyrski:
# Description 📣

Change to display secret imports on the secret overview page.
CleanShot 2025-03-17 at 19 14 22

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - Enhanced the Secret Manager Overview to display and filter imported secrets.
  - Introduced a file import icon and a dropdown toggle to control the visibility of secret imports.
  - Updated the dashboard to show total counts and unique details for imported secrets.
  - Improved support for environment-specific secret imports, delivering a more comprehensive secret management experience.
  - Added an optional `environment` property to secret import types for better context.


backend/src/services/secret-import/secret-import-service.ts
```diff
@@ -469,6 +469,43 @@ export const secretImportServiceFactory = ({
...
+  }: Omit<TGetSecretImportsDTO, "environment"> & { environments: string[] }) => {
+    const { permission } = await permissionService.getProjectPermission({
+      actor,
+      actorId,
+      projectId,
+      actorAuthMethod,
+      actorOrgId,
+      actionProjectType: ActionProjectType.SecretManager
+    });
+    for (const environment of environments) {
```
akhilmhdh, line 490: The issue with throwing forbidden here is: If one of them fails, everything else also fails. I think you need to show the non accessible ones as zero

backend/src/services/secret-import/secret-import-service.ts
```diff
@@ -688,6 +725,51 @@ export const secretImportServiceFactory = ({
...
+      ForbiddenError.from(permission).throwUnlessCan(
```
akhilmhdh, line 749: Same here, I think @scott-ray-wilson did some internal flag check like in secret service to avoid throwing error

akhilmhdh (changes requested):
The overview page should work, even when you don't have access to another path.

akhilmhdh:
Good job @carlosmonastyrski

PR #3270, opened 2025-03-19
feat(helm): custom volume support

DanielHougaard:
# Description 📣

We currently don't support specifying your own volumes for k8 deployments, which is needed in order to use an HSM client with k8's, otherwise users will need to build their own docker image with the client files pre-packaged.

Instead with this PR, people will be able to persistent volumes like so: 
```yaml
  extraVolumeMounts:
    - name: hsm-data
      mountPath: /hsm-client
      subPath: ./hsm-client

  extraVolumes:
    - name: hsm-data
      persistentVolumeClaim:
        claimName: infisical-data-pvc # volume created which is not a part of the deployment itself
```

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

## Summary by CodeRabbit

- **Chores**
  - Updated the release version to 1.4.1.

- **New Features**
  - Introduced additional configuration options to enable flexible customization of volume mounts and volumes during deployments.
  - Added support for supplying extra volume mounts and volumes via `infisical.extraVolumeMounts` and `infisical.extraVolumes`.

### DIFF
--- a/helm-charts/infisical-standalone-postgres/templates/infisical.yaml
+++ b/helm-charts/infisical-standalone-postgres/templates/infisical.yaml
@@ -77,6 +77,14 @@ spec:
         {{- if  $infisicalValues.resources }}
         resources: {{- toYaml $infisicalValues.resources | nindent 12 }}
         {{- end }}
+        {{- with $infisicalValues.extraVolumeMounts }}
+        volumeMounts:
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
+      {{- with $infisicalValues.extraVolumes }}
+      volumes:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
 ---
 
 apiVersion: v1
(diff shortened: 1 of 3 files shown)


helm-charts/infisical-standalone-postgres/templates/infisical.yaml
```diff
@@ -77,6 +77,14 @@ spec:
         {{- if  $infisicalValues.resources }}
         resources: {{- toYaml $infisicalValues.resources | nindent 12 }}
         {{- end }}
+        {{- with $infisicalValues.extraVolumeMounts }}
+        volumeMounts:
+          {{- toYaml . | nindent 10 }}
```
maidul98, line 82: q: columeMounts and resources should be at the same level yet they have different indents. does k8s process this as vaild?

maidul98: Daniel said this is valid

DanielHougaard: Yes this is valid, I've been testing with this locally the today, and I ran it through a linter

maidul98 (approved):

PR #3274, opened 2025-03-19
Bulk Delete Secret Incorrectly says lacking permission

carlosmonastyrski:
# Description 📣

**Issue**
When attempting to bulk delete secrets across multiple environments on the secrets overview page, users received a permission error, even though they had the necessary permissions.

**Root Cause**
The problem was in the handleBulkDelete reducer logic. Specifically, `const entry = secretRecord[env.slug];` could return undefined. When that happened, the permissions check (permission.can) failed, and the reducer loop would break, causing only the first selected secret to be processed and skipping the rest.

**Solution**
Added a conditional check to skip entries where entry is undefined by simply returning the accumulator. This prevents unnecessary permission checks for invalid entries and ensures the reducer continues processing all selected secrets correctly.

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the bulk deletion process in the Secret Manager by ensuring that only valid secret entries are processed. This update prevents potential errors when handling missing or invalid entries, resulting in a more reliable and stable deletion experience.


scott-ray-wilson (approved):
LGTM!

PR #3303, opened 2025-03-24
Rework of secret imports on the overview page

carlosmonastyrski:
# Description 📣

Rework to add more clarity to overview page for secret imports. Move from folder row to secrets rows.
CleanShot 2025-03-24 at 20 15 26

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - Secret details now include the originating environment for enhanced context.

- **Refactor**
  - Unifies local and imported secret data in the management overview.
  - Removes a redundant secret lookup and legacy import display component.
  - Updates secret renaming behavior by disabling editing when no valid secrets are available.


frontend/src/pages/secret-manager/OverviewPage/components/SecretOverviewTableRow/SecretRenameRow.tsx
```diff
@@ -162,7 +162,7 @@ function SecretRenameRow({ environments, getSecretByKey, secretKey, secretPath }
           render={({ field, fieldState: { error } }) => (
             <Input
               autoComplete="off"
-              isReadOnly={isReadOnly}
+              isReadOnly={isReadOnly || secrets.filter(Boolean).length === 0}
```
DanielHougaard, line 165: In what case would `secrets.filter(Boolean).length === 0` be true?

DanielHougaard (changes requested):

frontend/src/pages/secret-manager/OverviewPage/components/SecretOverviewTableRow/SecretRenameRow.tsx
carlosmonastyrski: That would happen when there’s no custom secret defined in any environment, and only imported secrets are present for the given secret key.
It comes from this line:`const secrets = environments.map((env) => getSecretByKey(env.slug, secretKey));` — so if none of those environments have a secret that is not an import, secrets.filter(Boolean).length will be 0

DanielHougaard (approved):

PR #3347, opened 2025-04-01
Infisical SSH - V2

dangtony98:
# Description 📣

This PR adds:
- Infisical SSH V2 :)
- Documentation for how to configure host key signing as part of Infisical SSH; this enables a SSH client to verify the target remote host.
- Support for using external key pairs as part of creating an SSH CA.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [x] Documentation

---

## Summary by CodeRabbit

- **New Features**
	- SSH certificate validation now accepts a maximum TTL equal to the TTL.
	- Expanded SSH host management with capabilities to list, create, update, and delete SSH hosts.
	- New CLI commands let users connect to and register SSH hosts.
	- Introduced a new SSH Certificate Authority (CA) management interface with improved form handling.
	- New animations enhancing the certificate management experience.
	- New user interface components for managing SSH hosts and certificate authorities.
	- Added support for SSH host actions in project permissions, including Read, Create, Edit, Delete, and IssueHostCert.
	- New documentation file providing detailed guidance on SSH credential generation and management.

- **Documentation**
	- Updated Infisical SSH guides with clearer, streamlined setup instructions and a dedicated user guide.

- **Chores**
	- Upgraded dependencies and adjusted rate limits to bolster SSH service reliability.


backend/src/ee/services/ssh-host/ssh-host-service.ts
```diff
@@ -0,0 +1,708 @@
...
+          hostname
+        },
+        tx
+      );
+
+      if (existingHost) {
+        throw new BadRequestError({
+          message: `SSH host with hostname ${hostname} already exists`
+        });
+      }
```
sheensantoscapadngan, line 203: oh btw we have a new convention. Instead of doing this, we should add the unique constraint in the DB

backend/src/db/migrations/20250405185753_ssh-mgmt-v2.ts
```diff
@@ -0,0 +1,91 @@
...
+    await createOnUpdateTrigger(knex, TableName.SshHost);
+  }
+
+  if (!(await knex.schema.hasTable(TableName.SshHostLoginUser))) {
+    await knex.schema.createTable(TableName.SshHostLoginUser, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.timestamps(true, true, true);
+      t.uuid("sshHostId").notNullable();
+      t.foreign("sshHostId").references("id").inTable(TableName.SshHost).onDelete("CASCADE");
+      t.string("loginUser").notNullable(); // e.g. ubuntu, root, ec2-user, ...
```
sheensantoscapadngan, line 31: let's add a unique constraint here for both sshHostId and loginUser?

backend/src/db/migrations/20250405185753_ssh-mgmt-v2.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  if (!(await knex.schema.hasTable(TableName.SshHostLoginUserMapping))) {
+    await knex.schema.createTable(TableName.SshHostLoginUserMapping, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.timestamps(true, true, true);
+      t.uuid("sshHostLoginUserId").notNullable();
+      t.foreign("sshHostLoginUserId").references("id").inTable(TableName.SshHostLoginUser).onDelete("CASCADE");
+      t.uuid("userId").nullable();
+      t.foreign("userId").references("id").inTable(TableName.Users).onDelete("CASCADE");
```
sheensantoscapadngan, line 43: here as well

backend/src/ee/services/ssh/ssh-certificate-authority-fns.ts
```diff
@@ -388,3 +518,88 @@ export const createSshCert = async ({
...
+  if (actor !== ActorType.USER) {
+    throw new BadRequestError({
+      message: "Failed to convert actor to principals due to unsupported actor type"
+    });
+  }
+
+  const user = await userDAL.findById(actorId);
+
+  return [user.username];
+};
```
sheensantoscapadngan, line 605: hmmm I think we should also return user.email? 

note that usernames could be this randomly generated string when options like `trustSamlEmail` is enabled

backend/src/ee/services/ssh-host/ssh-host-dal.ts
```diff
@@ -0,0 +1,193 @@
...
+      const grouped = groupBy(rows, (r) => r.sshHostId);
+      return Object.values(grouped).map((hostRows) => {
+        const { sshHostId, hostname, userCertTtl, hostCertTtl, userSshCaId, hostSshCaId, projectId } = hostRows[0];
+
+        const loginMappingGrouped = groupBy(hostRows, (r) => r.loginUser);
+
+        const loginMappings = Object.entries(loginMappingGrouped).map(([loginUser]) => ({
+          loginUser,
+          allowedPrincipals: {
+            usernames: [user.username]
```
sheensantoscapadngan, line 55: it might be more appropriate to use email here

backend/src/ee/services/ssh-host/ssh-host-service.ts
```diff
@@ -0,0 +1,708 @@
...
+    const decryptedCaPrivateKey = secretManagerDecryptor({
+      cipherTextBlob: sshCaSecret.encryptedPrivateKey
+    });
+
+    // (dangtony98): will support more algorithms in the future
+    const keyAlgorithm = SshCertKeyAlgorithm.ED25519;
+    const { publicKey, privateKey } = await createSshKeyPair(keyAlgorithm);
+
+    // (dangtony98): include the loginUser as a principal on the issued certificate
+    const principals = [...internalPrincipals, loginUser];
```
sheensantoscapadngan, line 503: why do we need to?

backend/src/ee/services/ssh-host/ssh-host-service.ts
```diff
@@ -0,0 +1,708 @@
...
+      caPrivateKey: decryptedCaPrivateKey.toString("utf8"),
+      clientPublicKey: publicKey,
+      keyId,
+      principals,
+      requestedTtl: host.userCertTtl,
+      certType: SshCertType.USER
+    });
+
+    const { encryptor: secretManagerEncryptor } = await kmsService.createCipherPairWithDataKey({
+      type: KmsDataKey.SecretManager,
```
sheensantoscapadngan, line 515: I think this v2 would be a great opportunity for us to do this part correctly (as in like use SSH KMS data key)... thoughts?

backend/src/ee/services/ssh-host/ssh-host-service.ts
```diff
@@ -0,0 +1,708 @@
...
+        const projectHosts = await sshHostDAL.findUserAccessibleSshHosts([project.id], actorId);
+
+        allowedHosts.push(...projectHosts);
+      } catch {
+        // intentionally ignore projects where user lacks access
+      }
+    }
+
+    return allowedHosts;
+  };
```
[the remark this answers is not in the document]
dangtony98: This is intentional actually because this endpoint is used for a user to list all the SSH hosts that only they can connect to (the filter also ensures that they cannot see any other allowed principals on the login users that they can access as except for themselves); it does not return any SSH hosts that they cannot connect to.

The `Read` permission applies to the get SSH hosts in project endpoint in `/v2/projects`.

backend/src/ee/services/ssh-host/ssh-host-service.ts
dangtony98: Oh actually I did put a unique constraint in the migration file; I guess we can remove this then?

backend/src/ee/services/ssh/ssh-certificate-authority-fns.ts
dangtony98: I would stick to `username` because it doesn't necessarily have to be an email. Most users on Infisical have usernames being their email anyways; also, I don't trust email because that's not verified - `username` should be used as source of identity.

backend/src/ee/services/ssh-host/ssh-host-dal.ts
dangtony98: Same reasoning for not trusting emails and taking usernames over it; I think `username` should be used as source of trusted identity.

backend/src/ee/services/ssh-host/ssh-host-service.ts
dangtony98: It won't work otherwise against an EC2 instance; it expects the `loginUser` to be included as a principal.

backend/src/ee/services/ssh-host/ssh-host-service.ts
dangtony98: Asked @akhilmhdh before about this actually; since each project has only one type anyways then it doesn't matter (it would matter else if each project supported multiple different lines). I wouldn't change it cuz we have other SSH material being encrypted under this key logic atm.

backend/src/db/migrations/20250405185753_ssh-mgmt-v2.ts
dangtony98: Done.

backend/src/db/migrations/20250405185753_ssh-mgmt-v2.ts
dangtony98: Done.

backend/src/ee/services/ssh/ssh-certificate-authority-fns.ts
sheensantoscapadngan: From a terraform perspective, email would simplify it a lot because they won't have to fetch the user beforehand or try to find out what the username is from the email when there's a difference

username != email is an edge-case which could only really happen when they toggled the `trust...` instance config after initial registration/scim. This would be generally okay though so I guess we could keep it as is

backend/src/ee/services/ssh-host/ssh-host-service.ts
sheensantoscapadngan: yup, let's remove

backend/src/ee/services/ssh-host/ssh-host-service.ts
sheensantoscapadngan: oh I see... because we currently don't have the AuthorizedPrincipals file to properly map the principals to the login user

cli/packages/cmd/ssh.go
```diff
@@ -595,6 +609,347 @@ func signKey(cmd *cobra.Command, args []string) {
...
+func sshConnect(cmd *cobra.Command, args []string) {
+	token, err := util.GetInfisicalToken(cmd)
+	if err != nil {
+		util.HandleError(err, "Unable to parse token")
+	}
+
+	var infisicalToken string
+	if token != nil && (token.Type == util.SERVICE_TOKEN_IDENTIFIER || token.Type == util.UNIVERSAL_AUTH_TOKEN_IDENTIFIER) {
+		infisicalToken = token.Token
+	} else {
```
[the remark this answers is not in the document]
dangtony98: Yes this is for actual users only.

cli/packages/cmd/ssh.go
```diff
@@ -48,6 +50,18 @@ var sshSignKeyCmd = &cobra.Command{
...
+	Use:   "connect",
+	Short: "Connect to an SSH host using issued credentials",
+	Run:   sshConnect,
+}
+
+var sshAddHostCmd = &cobra.Command{
+	Use:   "add-host",
+	Short: "Register a new SSH host with Infisical",
+	Run:   sshAddHost,
+}
```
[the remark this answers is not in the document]
dangtony98: Decided to go with `add-host` instead because the core action itself is adding a host; the other writing operations and potentially spinning up agent etc. is dictated by a flag.

cli/packages/cmd/ssh.go
```diff
@@ -617,5 +972,20 @@ func init() {
 	sshIssueCredentialsCmd.Flags().String("outFilePath", "", "The path to write the SSH credentials to such as ~/.ssh, ./some_folder, ./some_folder/id_rsa-cert.pub. If not provided, the credentials will be saved to the current working directory")
 	sshIssueCredentialsCmd.Flags().Bool("addToAgent", false, "Whether to add issued SSH credentials to the SSH agent")
 	sshCmd.AddCommand(sshIssueCredentialsCmd)
+
+	sshConnectCmd.Flags().String("token", "", "Use a machine identity access token")
```
sheensantoscapadngan, line 976: should we remove this?

frontend/src/layouts/ProjectLayout/ProjectLayout.tsx
```diff
@@ -134,18 +134,48 @@ export const ProjectLayout = () => {
...
+                              <MenuItem
+                                isSelected={isActive}
+                                icon="certificate-authority"
+                                iconMode="reverse"
+                              >
+                                Certificate Authorities
+                              </MenuItem>
+                            )}
+                          </Link> */}
+                        </>
```
sheensantoscapadngan, line 178: sayonara!

sheensantoscapadngan: so are we keeping this hidden? I see that you made some changes to the SSH CAs page so just confirming

cli/packages/cmd/ssh.go
```diff
@@ -595,6 +609,347 @@ func signKey(cmd *cobra.Command, args []string) {
...
+		if err != nil {
+			util.HandleError(err, "Unable to authenticate")
+		}
+		if loggedInUserDetails.LoginExpired {
+			util.PrintErrorMessageAndExit("Your login session has expired, please run [infisical login]")
+		}
+		infisicalToken = loggedInUserDetails.UserCredentials.JTWToken
+	}
+
+	projectId, err := cmd.Flags().GetString("projectId")
```
sheensantoscapadngan, line 751: I think we'd also want to add a CLI command to pull a host configuration from Infisical (which already includes user mapping, etc.) to bootstrap a host - that could be in a separate effort. 

it would be super convenient if after you add an SSH host from UI, you can just copy a CLI command from this view to bootstrap a host after the fact
image

backend/src/ee/services/ssh-host/ssh-host-service.ts
sheensantoscapadngan: resolution: AuthorizedPrincipals to follow

cli/packages/cmd/ssh.go
dangtony98: Agree! We can do that.

frontend/src/layouts/ProjectLayout/ProjectLayout.tsx
dangtony98: Yeah I'm keeping it hidden for now; there's a chance that we'll have to bring some parts of it back depending on actual usage.

I foresee it's possible that users will want to create and connect their own SSH CAs to hosts registered in Infisical.

cli/packages/cmd/ssh.go
```diff
@@ -595,6 +609,347 @@ func signKey(cmd *cobra.Command, args []string) {
...
+	})
+	if err != nil {
+		util.HandleError(err, "Failed to issue SSH credentials")
+	}
+
+	// Load credentials into SSH agent
+	err = addCredentialsToAgent(creds.PrivateKey, creds.SignedKey)
+	if err != nil {
+		util.HandleError(err, "Failed to add credentials to SSH agent")
+	}
```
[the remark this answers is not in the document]
dangtony98: Updated. Added a `writeHostCaToFile` flag which defaults to `true` to update this file accordingly with the instance that you are connecting to.

sheensantoscapadngan (approved):

dangtony98:
Hey @matzeeable so sorry for this issue caused by our latest improvement to Infisical SSH.

We'll be bringing back the CA tab this afternoon shortly and recommend some steps to migrate from the old version of Infisical SSH into this new version; we've improved the experience significantly with this update so I think it'll be really pleasant for you to use!

Any changes to get to the new version shouldn't be too difficult and I'd be happy to hop on a call to help and get feedback as well.

Apologies again and we'll bring it back shortly!

matzeeable:
@dangtony98 there is no reason to apologize 🙂 ok, is there already a short step -by-step instroduction how to migrate?

matzeeable:
I tried to figure out how the [new mechanism](https://infisical.com/docs/documentation/platform/ssh) works, but I ran into some issues:

- When I try to create a new host for an existing (prior version of Infisical) SSH project, I get the following error:

image

Creating a new SSH project and adding a host works. Affected lines:

https://github.com/Infisical/infisical/blob/862ed4f4e74d626b09d626dec17dfbfd0b1f9682/backend/src/ee/services/ssh-host/ssh-host-service.ts#L148-L160

- How can I use [`issue-credentials`](https://infisical.com/docs/cli/commands/ssh#infisical-ssh-issue-credentials) with the new mechanism? I want to connect to a remote host in our CI in non-interactive mode and therefore `connect` does not work.

PR #3355, opened 2025-04-02
Fetch self hosted license feature set without needing to redeploy the instance

carlosmonastyrski:
# Description 📣

This PR implements automatic polling for fetching the latest feature set from the license server, eliminating the need for customers to redeploy their self-hosted instances of Infisical to receive updates.

A background job now runs every 10 minutes, retrieving the latest feature set and updating onPremFeatures accordingly. This ensures that instances stay up to date with any changes made on the license server without requiring manual intervention.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - Introduced an automated background process that regularly refreshes licensing details every 10 minutes when a valid license key is provided, ensuring your licensed features remain current without manual effort.
  - Enhanced the initialization of the license service to include background synchronization capabilities.

- **Style**
  - Reformatted the `lastSecretModified` property for improved readability in the source code.

### DIFF
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -87,11 +87,13 @@ export const licenseServiceFactory = ({
   );
 
   const syncLicenseKeyOnPremFeatures = async (shouldThrow: boolean = false) => {
+    logger.info("Start syncing license key features");
     try {
       const {
         data: { currentPlan }
       } = await licenseServerOnPremApi.request.get<{ currentPlan: TFeatureSet }>("/api/license/v1/plan");
       onPremFeatures = currentPlan;
+      logger.info("Successfully synchronized license key features");
     } catch (error) {
       logger.error(error, "Failed to synchronize license key features");
       if (shouldThrow) throw error;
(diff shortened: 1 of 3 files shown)


backend/src/ee/services/license/license-service.ts
```diff
@@ -98,10 +110,7 @@ export const licenseServiceFactory = ({
       if (appCfg.LICENSE_KEY) {
         const token = await licenseServerOnPremApi.refreshLicense();
         if (token) {
-          const {
-            data: { currentPlan }
-          } = await licenseServerOnPremApi.request.get<{ currentPlan: TFeatureSet }>("/api/license/v1/plan");
-          onPremFeatures = currentPlan;
+          await syncLicenseKeyOnPremFeatures();
```
sheensantoscapadngan, line 113: For this one here, I think it's ideal to throw the error if there's any so that the feedback is quick... like customers know right away that something's wrong with the LICENSE_KEY

I think for the cron job, it's okay to fail silently there but for the initial bootstrap, we should throw accoridngly

PR #3422, opened 2025-04-15
Add project delete protection

carlosmonastyrski:
# Description 📣

To delete a project, you must first disable the project deletion protection feature. This works similarly to AWS's delete protection mechanism, providing an additional safeguard against accidental deletion.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - Introduced a workspace setting to enable or disable project delete protection, preventing accidental deletion when enabled.
  - Added a toggle control for delete protection in the project settings UI.
  - Users with appropriate permissions can update the delete protection status and receive notifications on success or error.
  - Added API support for toggling delete protection on projects and workspaces.
  - Enhanced project update functionality to include delete protection status.

- **Bug Fixes**
  - The "Delete Project" button is now disabled when delete protection is active for a workspace.


backend/src/server/routes/v1/project-router.ts
```diff
@@ -390,6 +390,43 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
     }
   });
 
+  server.route({
+    method: "POST",
```
scott-ray-wilson, line 394: Is there a reason for `POST` over `PATCH` in this scenario?

backend/src/services/project/project-service.ts
```diff
@@ -648,6 +655,29 @@ export const projectServiceFactory = ({
...
 
+  const toggleDeleteProtection = async ({
+    projectId,
+    actor,
+    actorId,
+    actorOrgId,
+    actorAuthMethod,
+    hasDeleteProtection
+  }: TToggleProjectDeleteProtectionDTO) => {
+    const { permission } = await permissionService.getProjectPermission({
```
scott-ray-wilson, line 666: might be good to make sure the project exists before trying to update so we can throw a not found error

scott-ray-wilson (changes requested):
Two comments on the existing code but also have a general question:

Was there a reason for making this a new endpoint/service function over adding this property as an optional parameter to the existing create/update project endpoints? This would be my initial thinking but I might be missing something.

If there is a reason for this pattern, right now this isn't exposed to the public so the only way a user can disable this is via the UI. If we keep it this way we should add API docs so users can access it. 

We should also probably add a note about delete protection to https://infisical.com/docs/documentation/platform/project

PR #3506, opened 2025-04-28
feat: strict find filter

DanielHougaard:
# Description 📣

This PR adds a new and improved find filter function, which will always prefix the table name to avoid ambiguous field postgres errors. This also fixes the bug with not being able to update personal secret overrides.
## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

## Summary by CodeRabbit

- **Refactor**
	- Enhanced data query filters for stricter and more precise results by adding table name prefixes and key exclusions.
	- Improved query construction for multi-table contexts and joins.

- **Chores**
	- Updated internal logic to prevent specific fields from unintentionally affecting search results.

### DIFF
--- a/backend/src/lib/knex/index.ts
+++ b/backend/src/lib/knex/index.ts
@@ -2,6 +2,8 @@
 import { Knex } from "knex";
 import { Tables } from "knex/types/tables";
 
+import { TableName } from "@app/db/schemas";
+
 import { DatabaseError } from "../errors";
 import { buildDynamicKnexQuery, TKnexDynamicOperator } from "./dynamic";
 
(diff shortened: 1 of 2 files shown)


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/b6961593-3a5c-41e2-9ad8-49c0e1eb8256?refs=a1bbd50c0bfdf8f4ce73f6da231f3152b7638693&source=prChecksComment)

backend/src/lib/knex/index.ts
```diff
@@ -25,6 +27,51 @@ export type TFindFilter<R extends object = object> = Partial<R> & {
...
+      });
+    }
+    if ($complex) {
+      return buildDynamicKnexQuery(bd, $complex);
+    }
+    return bd;
+  };
+
+/**
+ * @deprecated Use `buildStrictFindFilter` instead
```
carlosmonastyrski, line 73: are we sure we want to deprecate the old `buildFindFilter `? maybe we can just leave it or rename it like `NotStrict` or something like that, as there could be cases where we could need to use filter by different names of different tables? (of course considering not to trigger any ambiguous filter name)

carlosmonastyrski (dismissed):
LGTM, just left one small question there, but works perfectly on my end, great job Daniel 👍

backend/src/lib/knex/index.ts
maidul98: it may also make sense to combine them because they have a lot of overlap. we can add optional param object that can keep track of table name and excluded keys on `buildFindFilter` method. what do you think?

carlosmonastyrski: Yeah, that would be a great improvement. Maybe we can combine them and just have an external function to get the correct key based on the case:
```
function getFindFilterKey({ tableName?: TableName | null, key: string }) {
    return tableName ? [`${tableName}.${key}`] as never : key as never;
}
```
We'd probably need to update the function calls in a lot of places though.
CleanShot 2025-04-28 at 20 54 44@2x

maidul98: well, it would be optional, so we wouldn't need to define that param

carlosmonastyrski (approved):

PR #3535, opened 2025-05-02
feat(ssh-hosts): Add groups to ssh hosts allowed principals

carlosmonastyrski:
# Description 📣

Add groups to SSH Hosts allowed principals settings, now it can be set to grant access to individual users and groups.
CleanShot 2025-05-02 at 11 13 49
CleanShot 2025-05-02 at 11 12 45

Tested using the CLI with a host with only a group set on the allowed principals.
CleanShot 2025-05-02 at 11 17 28@2x

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

## Summary by CodeRabbit

- **New Features**
  - Added support for assigning SSH host login access to groups, not just individual users.
  - SSH host login mappings can now specify both usernames and group slugs as authorized principals.
  - UI for SSH host and SSH host group management updated to allow selection and display of both users and groups as principals.
  - Group-based SSH access is enforced and validated with permission checks.
  - Added ability to check group permissions on projects.

- **Documentation**
  - Updated API documentation to reflect that login mappings accept both usernames and group slugs as principals.

- **Bug Fixes**
  - Improved validation to prevent duplicate principals and ensure only valid groups with appropriate permissions can be assigned.


dangtony98:
LGTM except this is missing a filter considering group memberships in the `listProjectSshHosts` function in `v2/project-router.ts`.

Specifically, the below code block currently ensures that only hosts that a user is allowed to `Read` (with the condition on `hostname`) are rendered.

```javascript
for (const host of hosts) {
      try {
        ForbiddenError.from(permission).throwUnlessCan(
          ProjectPermissionSshHostActions.Read,
          subject(ProjectPermissionSub.SshHosts, {
            hostname: host.hostname
          })
        );

        allowedHosts.push(host);
      } catch {
        // intentionally ignore projects where user lacks access
      }
    }
```

This must be updated to consider groups because currently if a group is given a role that has `Read` access to SSH hosts with a specific `hostname` then the user that is part of that group does not see any hosts since this check does not consider group memberships.

dangtony98 (approved):

PR #3598, opened 2025-05-14
feat(secret-sync): Secret Key Schema

x032205:
# Description 📣

Adds secret key schema

## Type ✨

- [ ] Bug fix
- [x] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation


docs/integrations/cloud/heroku.mdx
scott-ray-wilson, line 1: this is a native integration not secret sync so we should roll this one back

backend/src/services/secret-sync/aws-secrets-manager/aws-secrets-manager-sync-fns.ts
```diff
@@ -399,6 +400,9 @@ export const AwsSecretsManagerSyncFns = {
       if (syncOptions.disableSecretDeletion) return;
 
       for await (const secretKey of Object.keys(awsSecretsRecord)) {
+        // eslint-disable-next-line no-continue
+        if (!matchesSchema(secretKey, syncOptions.keySchema)) continue;
```
scott-ray-wilson, line 404: nit: I think it would be more optimal to do this check after checking if the key isn't present first; for some secrets we would be running this check even though we're not deleting it (applies to all/most syncs not just aws)

docs/integrations/secret-syncs/aws-parameter-store.mdx
```diff
@@ -40,6 +40,7 @@ description: "Learn how to configure an AWS Parameter Store Sync for Infisical."
                 - **Overwrite Destination Secrets**: Removes any secrets at the destination endpoint not present in Infisical.
                 - **Import Secrets (Prioritize Infisical)**: Imports secrets from the destination endpoint before syncing, prioritizing values from Infisical over Parameter Store when keys conflict.
                 - **Import Secrets (Prioritize AWS Parameter Store)**: Imports secrets from the destination endpoint before syncing, prioritizing values from Parameter Store over Infisical when keys conflict.
+            - **Key Schema**: Template that determines how secret names are transformed when syncing, using `{{secretKey}}` as a placeholder for the original secret name.
```
scott-ray-wilson, line 43: lets add a `<Note />` block stating that Infisical strongly recommends using a key schema to enable Infisical to determine which secrets it is responsible for (play with wording) with a 'click here to learn more' with a link to a diagram showing how the behavior differs between using a schema/without (let's add a section with the diagram explaining this in more in depth to https://infisical.com/docs/integrations/secret-syncs/overview at the bottom with an anchor so we can link directly to it)

Let's add this note too all sync pages at the key schema 😓 (sorry)

frontend/src/components/secret-syncs/forms/SecretSyncOptionsFields/SecretSyncOptionsFields.tsx
```diff
@@ -122,6 +122,23 @@ export const SecretSyncOptionsFields = ({ hideInitialSync }: Props) => {
...
+      <Controller
+        render={({ field: { value, onChange }, fieldState: { error } }) => (
+          <FormControl
+            tooltipClassName="max-w-md"
+            tooltipText="When a secret is synced, its key will be injected into the key schema before it reaches the destination. This is useful for organization."
+            isError={Boolean(error)}
+            isOptional
+            errorText={error?.message}
+            label="Key Schema"
+            helperText="Setting a Key Schema is recommended. This helps separate synced secrets from regular secrets within a destination."
```
scott-ray-wilson, line 134: Let's add a link to the overview doc addition here too and make it sound a little more critical vs just a recommendation (play with wording):

```
helperText={
              <Tooltip
                className="max-w-md"
                content="{{explanation why and with link to docs}}"
              >
                <div>
                  <span>Infisical strongly advises setting a Key Schema</span>{" "}
                  <FontAwesomeIcon icon={faCircleInfo} className="text-mineshaft-400" />
                </div>
              </Tooltip>
            }

scott-ray-wilson (changes requested):
Functionally looking good! Few nits and document/UI requests

scott-ray-wilson (approved):
LGTM!

PR #3612, opened 2025-05-16
fix(auth): cli auth bug

DanielHougaard:
# Description 📣

At last, we fixed the CLI login bug with callback port not being used correctly.

The issue was that on the login page we had a useEffect that would redirect to select org, and within select org we'd do SAML auth flow or redirect to the organization page. But at the same time, it would also load the `InitialStep` component because the step would always default to 1 on page load. Inside the InitialStep component we also handle SAML auth. This created a race condition leading to strange behavior for SAML auth handling.

The solution is to first wait for the useEffect to finish before rendering the rest of the login page.

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/frontend/src/pages/auth/LoginPage/LoginPage.tsx
+++ b/frontend/src/pages/auth/LoginPage/LoginPage.tsx
@@ -10,7 +10,7 @@ import { useNavigateToSelectOrganization } from "./Login.utils";
 
 export const LoginPage = ({ isAdmin }: { isAdmin?: boolean }) => {
   const { t } = useTranslation();
-  const [step, setStep] = useState(0);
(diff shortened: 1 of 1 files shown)


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/3fab10df-4a89-49ed-a35d-a3342c9d60fc?refs=8cc2e08f24bf6efc6d8b90647757dd1c46a1d11c&source=prChecksComment)

PR #3683, opened 2025-05-30
Add support for offline lottie 

maidul98:
In air gapped, lotties won't load because the WASM player is fetched from CDN. This PR bundles the player so we can fetch it directly from file system

Note: When using @lottiefiles/dotlottie-web directly, we need to make sure its version matches the one required by @lottiefiles/dotlottie-react. You can double check this with:

`npm list @lottiefiles/dotlottie-web`

The version should match in both places:

@lottiefiles/dotlottie-web@0.38.2  <-- This version
@lottiefiles/dotlottie-web@0.38.2    <-- Should match this version

If they don't match, the Lottie animations will fail to load because the WASM binary version needs to be compatible with the React components.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -25,7 +25,7 @@
         "@hookform/resolvers": "^3.9.1",
         "@lexical/react": "^0.29.0",
         "@lottiefiles/dotlottie-react": "^0.12.0",
-        "@lottiefiles/dotlottie-web": "^0.44.0",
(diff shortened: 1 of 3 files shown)


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/537cfbde-7313-4a68-be61-58e34eb1b12e?refs=6cbeb4ddf95322233bb22d16197bbac9831e47b7&source=prChecksComment)

PR #3691, opened 2025-05-30
feat(access-request): fix issue for deleted custom privileges reopening old closed access requests

carlosmonastyrski:
# Description 📣

Added two new columns on `access_approval_requests` to track the status of the requests as the `privilegeId` can be set to NULL by a Delete Cascade and that should not reopen the closed access request. Added an `status` field to track the real state of the request and a `privilegeDeletedAt` field to track when the privilege attached to the request was deleted.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


x032205 (commented):
Not sure if this is intentional or not, but after deleting a policy and re-creating it, the request counter no longer keeps track of the requests from the deleted policy.

image

x032205 (approved):

PR #3818, opened 2025-06-18
feat: added auto-bootstrap support to helm

sheensantoscapadngan:
# Description 📣
This PR adds auto-bootstrap support to CLI and helm

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/e784360b-a8b6-4010-a6e8-79167790a57a?refs=0eb36d7e35ded29a233046b21902bce39af09f04&source=prChecksComment)

cli/packages/cmd/bootstrap.go
```diff
@@ -4,16 +4,148 @@ Copyright (c) 2023 Infisical Inc.
...
 	"github.com/Infisical/infisical-merge/packages/api"
 	"github.com/Infisical/infisical-merge/packages/util"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 )
 
+// handleK8SecretOutput processes the k8-secret output type by creating a Kubernetes secret
+func handleK8SecretOutput(bootstrapResponse api.BootstrapInstanceResponse, k8SecretTemplate, k8SecretName, k8SecretNamespace string) error {
+	// Read Kubernetes service account credentials from the pod
+	k8sToken, err := os.ReadFile(util.KUBERNETES_SERVICE_ACCOUNT_TOKEN_PATH)
```
maidul98: instead of doing this manually, we should use the `client-go` k8s sdk which handles all of this for us (less error prone)

docs: https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/#using-official-client-libraries

example: https://github.com/kubernetes/client-go/blob/master/examples/in-cluster-client-configuration/main.go#L39

we can create secrets and etc in the module `corev1 "k8s.io/api/core/v1"`:

example:
```
secret := &corev1.Secret{
        ObjectMeta: metav1.ObjectMeta{
            Name:      secretName,
            Namespace: namespace,
        },
        Type: corev1.SecretTypeOpaque,
        Data: data,
    }
```

what do you think?

cli/packages/cmd/bootstrap.go
```diff
@@ -69,11 +201,56 @@ var bootstrapCmd = &cobra.Command{
...
+			log.Error().Msgf("Failed to get k8-secret-name: %v", err)
+		}
+
+		k8SecretNamespace, err := cmd.Flags().GetString("k8-secret-namespace")
+		if err != nil {
+			log.Error().Msgf("Failed to get k8-secret-namespace: %v", err)
+		}
+
+		if outputType == "k8-secret" {
+			if k8SecretTemplate == "" {
```
maidul98, line 205: maybe there should be a default? where the keys of the response just map over to their value? and if they want to template the keys and the values, they can do that?

cli/packages/cmd/bootstrap.go
```diff
@@ -4,16 +4,148 @@ Copyright (c) 2023 Infisical Inc.
...
+	k8sPort := os.Getenv(util.KUBERNETES_SERVICE_PORT_HTTPS_ENV_NAME)
+	if k8sHost == "" || k8sPort == "" {
+		return fmt.Errorf("failed to get Kubernetes API server address from environment variables")
+	}
+
+	k8sApiUrl := fmt.Sprintf("https://%s:%s", k8sHost, k8sPort)
+
+	// Parse and execute the template to render only the data/stringData section
+	tmpl, err := template.New("k8-secret-template").Funcs(template.FuncMap{
+		"b64enc": func(s string) string {
```
maidul98: nit: we should rename to something more legible: encodeBase64

cli/packages/cmd/bootstrap.go
```diff
@@ -102,6 +289,10 @@ func init() {
 	bootstrapCmd.Flags().String("email", "", "The desired email address of the instance admin")
 	bootstrapCmd.Flags().String("password", "", "The desired password of the instance admin")
 	bootstrapCmd.Flags().String("organization", "", "The name of the organization to create for the instance")
-
+	bootstrapCmd.Flags().String("output", "", "The type of output to use for the bootstrap command (json or k8-secret)")
+	bootstrapCmd.Flags().Bool("ignore-if-bootstrapped", false, "Whether to continue on error if the instance has already been bootstrapped")
+	bootstrapCmd.Flags().String("k8-secret-template", "", "The template to use for rendering the Kubernetes secret (entire secret JSON)")
+	bootstrapCmd.Flags().String("k8-secret-namespace", "", "The namespace to use for the Kubernetes secret")
```
maidul98: isn't this more like what namespace to create the secret in?

docs/cli/commands/bootstrap.mdx
```diff
@@ -75,8 +75,90 @@ This flag is required.
...
+This flag is optional and defaults to `false`.
+
+</Accordion>
+
+<Accordion title="--output">
+  The type of output format for the bootstrap command. Supports `k8-secret` for Kubernetes secret integration. This flag is optional and defaults to "".
+
+```bash
+# Kubernetes secret output
+infisical bootstrap --output=k8-secret --k8-secret-template='{"data":{"token":"{{.Identity.Credentials.Token}}"}}' --k8-secret-name=infisical-bootstrap --k8-secret-namespace=default
```
maidul98, line 95: maybe we should mention that the command assumes it is running in a pod?

docs/cli/commands/bootstrap.mdx
```diff
@@ -105,6 +187,47 @@ The command returns a JSON response with details about the created user, organiz
 }
 ```
 
+### Kubernetes Secret Output
```
maidul98, line 190: we should mention when this output is useful. like in what scenario

docs/self-hosting/guides/automated-bootstrapping.mdx
```diff
@@ -60,9 +62,95 @@ You can bootstrap an Infisical instance using either the API or the CLI.
     ```
 
     This example command pipes the output through `jq` to extract only the machine identity token, making it easy to capture and use directly in automation scripts or export as an environment variable for tools like Terraform.
+
   </Tab>
 </Tabs>
 
+## Helm Chart Auto Bootstrap
+
+When deploying Infisical using the official Helm chart, you can enable automatic bootstrapping that runs as part of the deployment process. This eliminates the need to manually bootstrap the instance after deployment.
```
maidul98: should we mention that it will also output the token for use by other tools or process?

helm-charts/infisical-standalone-postgres/values.yaml
```diff
@@ -13,6 +13,31 @@ infisical:
   # -- Automatically migrates new database schema when deploying
   autoDatabaseSchemaMigration: true
 
+  autoBootstrap:
+    # -- Enable auto-bootstrap of the Infisical instance
+    enabled: false
```
maidul98, line 18: i think this should be false to not introduce any new changes? make it more opt in

cli/packages/cmd/bootstrap.go
sheensantoscapadngan: yeah, good point about using the client library. I weighed that option but felt the manual approach was better here since we're only doing basic secret creation and wanted to keep CLI dependencies minimal since we don't use it anywhere else. do you think we should use the client library here anyway?

helm-charts/infisical-standalone-postgres/values.yaml
sheensantoscapadngan: it's already `false` xD

docs/cli/commands/bootstrap.mdx
sheensantoscapadngan: addressed
image

cli/packages/cmd/bootstrap.go
sheensantoscapadngan: addressed

cli/packages/cmd/bootstrap.go
```diff
@@ -4,16 +4,127 @@ Copyright (c) 2023 Infisical Inc.
...
+		if dataMap, ok := data.(map[string]interface{}); ok {
+			for key, value := range dataMap {
+				if strValue, ok := value.(string); ok {
+					secretData[key] = []byte(strValue)
+				}
+			}
+		}
+	}
+
+	if stringData, exists := dataSection["stringData"]; exists {
```
maidul98, line 78: q: why would there string or stringdata? i assume it i what we set?

maidul98 (approved):

PR #3923, opened 2025-07-03
fix(approval-policies): improve policies handling

DanielHougaard:
# Description 📣

This PR improves how we handle approval policies. We have some discrepancies in how paths are handled in change/access policies.

**Change policy:** Allows for an empty path, and it will result in a environment-wide policy
**Access policy**: Doesn't allow for an empty path, it'll break if an empty path is used.

For access policies I've added extra validation to ensure policies are never created/updated with an empty path. 

For _both_ access and change policies, I've added extra checks to ensure there can never be duplicate policies, because this would mess up the UI when creating access requests; and there's really no logical reason to have duplicate policies for a path by our current approach for approval policies
## Type ✨

- [x] Bug fix
- [ ] New feature
- [x] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/db/migrations/20250710115149_required-path-on-approval-policies.ts
+++ b/backend/src/db/migrations/20250710115149_required-path-on-approval-policies.ts
@@ -1,0 +1,55 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const existingSecretApprovalPolicies = await knex(TableName.SecretApprovalPolicy)
+    .whereNull("secretPath")
+    .orWhere("secretPath", "");
+
+  const existingAccessApprovalPolicies = await knex(TableName.AccessApprovalPolicy)
+    .whereNull("secretPath")
+    .orWhere("secretPath", "");
+
+  // update all the secret approval policies secretPath to be "/**"
+  if (existingSecretApprovalPolicies.length) {
+    await knex(TableName.SecretApprovalPolicy)
+      .whereIn(
+        "id",
+        existingSecretApprovalPolicies.map((el) => el.id)
+      )
+      .update({
+        secretPath: "/**"
+      });
+  }
+
+  // update all the access approval policies secretPath to be "/**"
+  if (existingAccessApprovalPolicies.length) {
+    await knex(TableName.AccessApprovalPolicy)
+      .whereIn(
+        "id",
+        existingAccessApprovalPolicies.map((el) => el.id)
+      )
+      .update({
+        secretPath: "/**"
+      });
+  }
+
+  await knex.schema.alterTable(TableName.SecretApprovalPolicy, (table) => {
+    table.string("secretPath").notNullable().alter();
+  });
+
+  await knex.schema.alterTable(TableName.AccessApprovalPolicy, (table) => {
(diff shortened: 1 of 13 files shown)


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/b13978a5-15ac-44df-89ba-59cb23ab9ffa?refs=97c96acea5511291172785592456004e25daebc8&source=prChecksComment)

scott-ray-wilson (changes requested):
I think we should be consistent with requirements on policies (we've even loosely talked about policies being combined into a single entity with different subjects instead of the current split we have) - it feels inconsistent that `secretPath` can be undefined for change policies but not access; I think we should make it required for both and run a migration to updated any null values on either to be `*` - thoughts?

backend/src/db/migrations/20250710115149_required-path-on-approval-policies.ts
scott-ray-wilson, line 1: shouldn't we check for table/col existence?

backend/src/ee/routes/v1/access-approval-policy-router.ts
scott-ray-wilson, line 1: should we remove trailing slash on these paths too?

backend/src/ee/services/secret-approval-policy/secret-approval-policy-service.ts
```diff
@@ -55,6 +55,27 @@ export const secretApprovalPolicyServiceFactory = ({
...
+  }: {
+    envId: string;
+    secretPath?: string | null;
+    policyId?: string;
+  }) => {
+    const policy = await secretApprovalPolicyDAL
+      .findOne({
+        envId,
+        // For environment-wide policies, we store the path as an empty string, even though the column is nullable; for that reason we check for an empty string.
+        ...(secretPath ? { secretPath } : { secretPath: "" }),
```
scott-ray-wilson: won't this always be defined now?

backend/src/ee/services/access-approval-policy/access-approval-policy-service.ts
```diff
@@ -290,9 +320,24 @@ export const accessApprovalPolicyServiceFactory = ({
       throw new BadRequestError({ message: "Approvals cannot be greater than approvers" });
     }
 
-    if (!accessApprovalPolicy) {
-      throw new NotFoundError({ message: `Secret approval policy with ID '${policyId}' not found` });
+    // Case: Previously we allowed secret path to be null, but now we don't.
+    // This check ensures that we have a secret path to match with for finding conflicting policies.
+    if (!secretPath && !accessApprovalPolicy.secretPath) {
+      throw new BadRequestError({ message: "Secret path is required to update the policy" });
```
scott-ray-wilson: wouldn't this always be false since we are setting it in the migration?

backend/src/db/migrations/20250710115149_required-path-on-approval-policies.ts
varonix0: Doesn't really matter in this case. When the migration reaches this the state of the database will always have those tables/cols

scott-ray-wilson (approved):
LGTM!

PR #3794, opened 2025-06-13
feat: refactoring ts to avoid infer each time

akhilmhdh:
# Description 📣

First batch of ts refactoring

TS Instantiations 11910135 ->  11390865

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
# Here's some code block to paste some code snippets
```

---

- [ ] I have read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/getting-started/code-of-conduct). 📝

PR #3926, opened 2025-07-04
feat(secret-reminders): rework secret reminders logic

carlosmonastyrski:
# Description 📣

This PR introduces several improvements and refactors to the reminders system:

- One-time Secret Reminders: Secret reminders can now be configured to run only once.
- Improved UI Display: The UI now shows how many days remain until a reminder is sent, instead of displaying the reminder frequency. 
- Reminder System Refactor:
   - New Reminders Table: Reminders are now stored in a dedicated table, enabling support for other resource types and making the logic reusable across the system
   - Decoupled from Secret Change Policies: Since reminders are tracked separately, they are no longer affected or blocked by secret change policies.
   - Daily Job Processing: Replaces the previous queue-based event approach with a daily job, giving us greater control and visibility over when reminders are triggered.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/2069871b-5593-472b-b3d9-3cadc0bda994?refs=f490ca22acc2e12917a2153a539859a1e785b824&source=prChecksComment)

backend/src/db/schemas/models.ts
```diff
@@ -172,7 +172,11 @@ export enum TableName {
   SecretScanningResource = "secret_scanning_resources",
   SecretScanningScan = "secret_scanning_scans",
   SecretScanningFinding = "secret_scanning_findings",
-  SecretScanningConfig = "secret_scanning_configs"
+  SecretScanningConfig = "secret_scanning_configs",
+
```
varonix0: nit: remove empty space from list of table names

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -458,7 +458,11 @@ export enum EventType {
 
   CREATE_PROJECT = "create-project",
   UPDATE_PROJECT = "update-project",
-  DELETE_PROJECT = "delete-project"
+  DELETE_PROJECT = "delete-project",
+
+  CREATE_SECRET_REMINDER = "create-secret-reminder",
```
varonix0, line 473: For some reason I Feel like creating/updating/deleting secret reminders should fall under secret events?

If you delete a secret reminder = you update a secret
If you create a secret reminder = you update a secret

Curious what you think

backend/src/server/routes/v1/reminder-router.ts
```diff
@@ -0,0 +1,163 @@
+import { z } from "zod";
```
varonix0: Reminder router should be secret reminder router I think

backend/src/server/routes/v1/index.ts
```diff
@@ -84,6 +85,7 @@ export const registerV1Routes = async (server: FastifyZodProvider) => {
   await server.register(registerInviteOrgRouter, { prefix: "/invite-org" });
   await server.register(registerUserActionRouter, { prefix: "/user-action" });
   await server.register(registerSecretImportRouter, { prefix: "/secret-imports" });
+  await server.register(registerReminderRouter, { prefix: "/reminders" });
```
varonix0: `/secret-reminders`

backend/src/services/reminder/reminder-dal.ts
```diff
@@ -0,0 +1,123 @@
+import { Knex } from "knex";
```
varonix0, line 1: Please rename all namings from `reminder` -> `secret-reminder`

backend/src/services/reminder/reminder-service.ts
```diff
@@ -0,0 +1,353 @@
...
+  reminderRecipientDAL: TReminderRecipientDALFactory;
+  smtpService: TSmtpService;
+  projectMembershipDAL: Pick<TProjectMembershipDALFactory, "findAllProjectMembers">;
+  permissionService: Pick<TPermissionServiceFactory, "getProjectPermission">;
+  secretV2BridgeDAL: Pick<TSecretV2BridgeDALFactory, "invalidateSecretCacheByProjectId">;
+};
+
+export type TReminderServiceFactory = ReturnType<typeof reminderServiceFactory>;
+
+export const reminderServiceFactory = ({
```
varonix0, line 28: Should be `secretReminderServiceFactory`

backend/src/services/reminder/reminder-dal.ts
```diff
@@ -0,0 +1,123 @@
...
+import { RemindersSchema } from "@app/db/schemas/reminders";
+import { ormify, selectAllTableCols, sqlNestRelationships } from "@app/lib/knex";
+
+export type TReminderDALFactory = ReturnType<typeof reminderDALFactory>;
+
+export const reminderDALFactory = (db: TDbClient) => {
+  const reminderOrm = ormify(db, TableName.Reminder);
+
+  const getTodayDateRange = () => {
+    const now = new Date();
```
varonix0: nit: a little hard to read at first glance. I would split things up a bit like so:

```js
const getTodayDateRange = () => {
  const today = new Date();
  const year = today.getUTCFullYear();
  const month = today.getUTCMonth();
  const date = today.getUTCDate();

  // Start of day: 00:00:00.000 UTC
  const startOfDay = new Date(Date.UTC(year, month, date, 0, 0, 0, 0));
  
  // End of day: 23:59:59.999 UTC  
  const endOfDay = new Date(Date.UTC(year, month, date, 23, 59, 59, 999));

  return { 
    startOfDay, 
    endOfDay 
  };
};
```

backend/src/services/reminder/reminder-service.ts
```diff
@@ -0,0 +1,353 @@
...
+type TReminderServiceFactoryDep = {
+  reminderDAL: TReminderDALFactory;
+  reminderRecipientDAL: TReminderRecipientDALFactory;
+  smtpService: TSmtpService;
+  projectMembershipDAL: Pick<TProjectMembershipDALFactory, "findAllProjectMembers">;
+  permissionService: Pick<TPermissionServiceFactory, "getProjectPermission">;
+  secretV2BridgeDAL: Pick<TSecretV2BridgeDALFactory, "invalidateSecretCacheByProjectId">;
+};
+
+export type TReminderServiceFactory = ReturnType<typeof reminderServiceFactory>;
```
varonix0: Didn't we move away from this sort of type inference?

https://github.com/Infisical/infisical/pull/3794

backend/src/services/secret/secret-queue.ts
```diff
@@ -1111,62 +1044,9 @@ export const secretQueueFactory = ({
     }
   });
 
+  // TODO: remove this queue (needed for queue initialization and perform the migration)
```
varonix0: tiniest nit: I would add your name to do the todo `TODO(Carlos):` so it's easier to glance at the code and see who is responsible for this part

backend/src/services/secret/secret-types.ts
```diff
@@ -310,6 +310,7 @@ export type TUpdateManySecretRawDTO = Omit<TProjectPermission, "projectId"> & {
     secretMetadata?: ResourceMetadataDTO;
     secretReminderRepeatDays?: number | null;
     secretReminderNote?: string | null;
+    secretPath?: string;
```
varonix0, line 313: Question: Why did we add this?

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -513,18 +513,15 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
...
-          `${TableName.SecretReminderRecipients}.secretId`
-        )
-        .leftJoin(TableName.Users, `${TableName.SecretReminderRecipients}.userId`, `${TableName.Users}.id`)
         .leftJoin(TableName.ResourceMetadata, `${TableName.SecretV2}.id`, `${TableName.ResourceMetadata}.secretId`)
         .leftJoin(
           TableName.SecretRotationV2SecretMapping,
           `${TableName.SecretV2}.id`,
           `${TableName.SecretRotationV2SecretMapping}.secretId`
         )
+        .leftJoin(TableName.Reminder, `${TableName.SecretV2}.id`, `${TableName.Reminder}.secretId`)
```
varonix0, line 523: Note: This query has a lot of left joins which might cause performance issues if we add more. For now I think it's okay, but we should keep an eye on the performance here.

Maybe you should add a comment to this DAL operation mentioning the amount of left joins and the impact on performance it can have? 

Related PR: https://github.com/Infisical/infisical/pull/3882/files

backend/src/db/schemas/models.ts
```diff
@@ -172,7 +172,11 @@ export enum TableName {
...
+  // reminders
+  Reminder = "reminders",
```
varonix0, line 177: Please rename this table to `secret_reminders`

backend/src/db/schemas/models.ts
```diff
@@ -172,7 +172,11 @@ export enum TableName {
...
+  ReminderRecipient = "reminders_recipients"
```
varonix0, line 178: I think it's strange that we have both `secret_reminder_recipients` and `reminders_recipients`. I think we should either name the new table `reminders_recipients` -> `secret_reminder_recipients_v2`, or keep using the existing `secret_reminder_recipients` table.

I'd probably lean towards renaming the table to `secret_reminder_recipients_v2`

backend/src/services/secret-v2-bridge/secret-v2-bridge-fns.ts
```diff
@@ -411,7 +393,10 @@ export const fnSecretBulkDelete = async ({
     deletedSecrets
```
varonix0, line 407: Are we sure we want to use `Promise.allSettled()` here? If one of the promises inside fails, `Promises.allSettled` **will fail silently**.

Perhaps we should use `Promise.all()` instead, which will throw an error if any of the promises fail

frontend/src/hooks/api/reminders/queries.tsx
```diff
@@ -0,0 +1,61 @@
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
```
varonix0, line 1: Please rename these as well to `secretReminder` instead of `reminder`

backend/src/ee/services/audit-log/audit-log-types.ts
carlosmonastyrski: That makes sense for the current approach. However, for the scope of this change, I've moved that functionality to a separate flow for the following reasons:

- Decoupled tracking: Reminders are not tracked as part of commit changes for secrets so from that perspective it's already a separate thing from the secret events we record with that.
- Resource-agnostic: We're expanding reminder functionality to other resources in the future, so it's better to make the implementation as independent from any specific resource as possible.
- Removal of blocking dependencies: Secret approvals currently block reminders until approval is granted, and we're moving away from this pattern.

backend/src/server/routes/v1/reminder-routers/secret-reminder-router.ts
```diff
@@ -6,18 +6,20 @@ import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
...
     config: {
       rateLimit: writeLimit
     },
     schema: {
       params: z.object({
-        projectId: z.string().uuid(),
         secretId: z.string().uuid()
       }),
+      querystring: z.object({
+        projectId: z.string().uuid()
```
varonix0: Question: Why do users need to provide the project ID? Can't we get this from the secret ID? Same applies to the other routes

backend/src/services/resource-cleanup/resource-cleanup-queue.ts
```diff
@@ -42,7 +43,6 @@ export const dailyResourceCleanUpQueueServiceFactory = ({
 }: TDailyResourceCleanUpQueueServiceFactoryDep) => {
   queueService.start(QueueName.DailyResourceCleanUp, async () => {
     logger.info(`${QueueName.DailyResourceCleanUp}: queue task started`);
-    await auditLogDAL.pruneAuditLog();
```
varonix0: Did you mean to remove audit log cleanup?

backend/src/services/reminder/reminder-service.ts
```diff
@@ -0,0 +1,362 @@
...
+      actorAuthMethod,
+      actorOrgId
+    });
+
+    ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionSecretActions.Edit, ProjectPermissionSub.Secrets);
+    await reminderDAL.delete({ secretId });
+    await secretV2BridgeDAL.invalidateSecretCacheByProjectId(projectId);
+  };
+
+  const removeReminderRecipients: TReminderServiceFactory["removeReminderRecipients"] = async (
```
varonix0: I don't actually see this function being used anywhere, can you double-check and if not needed remove it?

DanielHougaard (changes requested):

backend/src/db/migrations/20250701202824_add-reminder-table.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (!(await knex.schema.hasTable(TableName.Reminder))) {
+    await knex.schema.createTable(TableName.Reminder, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.uuid("secretId").nullable();
+      t.foreign("secretId").references("id").inTable(TableName.SecretV2).onDelete("CASCADE");
+      t.string("message", 1024).nullable();
+      t.integer("repeatDays").checkPositive().nullable();
+      t.timestamp("nextReminderDate").notNullable();
+      t.timestamps(true, true, true);
+      t.index("secretId");
```
akhilmhdh: You don't need another index - if unique index is applied

backend/src/db/migrations/20250701202824_add-reminder-table.ts
```diff
@@ -0,0 +1,44 @@
...
+    await knex.schema.createTable(TableName.ReminderRecipient, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.uuid("reminderId").notNullable();
+      t.foreign("reminderId").references("id").inTable(TableName.Reminder).onDelete("CASCADE");
+      t.uuid("userId").notNullable();
+      t.foreign("userId").references("id").inTable(TableName.Users).onDelete("CASCADE");
+      t.timestamps(true, true, true);
+      t.index("reminderId");
+      t.index("userId");
+      t.unique(["reminderId", "userId"]);
```
akhilmhdh, line 30: I would recommend to do an `EXPLAIN` over the non id based reminder dal operation and ensure these are needed

backend/src/services/reminder/reminder-queue.ts
```diff
@@ -0,0 +1,196 @@
...
+      delay: 5000,
+      jobId: QueueName.DailyReminders,
+      repeat: { pattern: "0 0 * * *", utc: true }
+    });
+  };
+
+  // TODO: remove once all the old reminders in queues are migrated
+  const startSecretReminderMigrationJob = async () => {
+    // clear previous job
+    await queueService.stopRepeatableJob(
```
akhilmhdh, line 170: Should we run this every hour? I am thinking may be each day? Because after first proper migration this would be useless

backend/src/services/resource-cleanup/resource-cleanup-queue.ts
carlosmonastyrski: seems like it was moved to the last row when I merged main
CleanShot 2025-07-14 at 10 28 28@2x

backend/src/services/reminder/reminder-queue.ts
carlosmonastyrski: Yes, absolutely. However, for the first few hours after rolling this out, I think the best approach would be to use a shorter interval. That way, if any reminders are created while we're updating the tasks, they'll be migrated to the new logic shortly after—minimizing the risk of any side effects for the user.

DanielHougaard (approved):

PR #3973, opened 2025-07-07
feat(fips): native FIPS support

DanielHougaard:
# Description 📣

## Native FIPS support

This PR introduces FIPS validation support (enterprise-only) for plain Infisical installations. There is no way to migrate from non-FIPS -> FIPS, so things doesn't need to be backwards compatible with each other.

**Core changes:**
1. We've bundled in OpenSSL 3.1.2 which is FIPS validated, and we've toggled on FIPS mode for it
2. We are forcing FIPS mode Node.js during startup. This means that Node.js won't allow OpenSSL to execute non-fips validated algorithms
3. Introduced crypto.js usage. This is **only** intended to be used in cases where we absolutely have to use non-fips validated algorithms. We use their md5 implementation for MongoDB digest auth for our dynamic secrets.
4. We're moved all crypto usage into a crypto provider (`cryptographyFactory`), which will handle all exports of crypto. This way we can be very explicit about what crypto we are using.
5. TweetNacl has been replaced with native node crypto. We replicate tweetnacl's asymmetric key-sharing encryption by using the Diffie-Hellman key exchange algorithm; which is implemented in native node.
6. `jsonwebtoken` is replaced with our own JWT implementation using native Node.
7. BCrypt has been replaced with pkdf2. We enforce a minimum of 1.000.000 iterations to keep the compute intensity high to avoid brute-force attacks.

To test things out, you can run `make up-dev`. Development node will now always run in FIPS-mode to ensure that we aren't developing features that won't be compatible with FIPS.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/a2eebaa5-cc62-41e3-8b6f-4b72a570c2b8?refs=b621225706e9130533f46cc691d980554f21260d&source=prChecksComment)

backend/src/lib/crypto/cryptography/crypto.ts
```diff
@@ -0,0 +1,450 @@
...
+      if (keySize === SymmetricKeySize.Bits128) {
+        iv = crypto.randomBytes(BLOCK_SIZE_BYTES_16);
+        cipher = crypto.createCipheriv(SecretEncryptionAlgo.AES_256_GCM, key, iv);
+      } else {
+        iv = crypto.randomBytes(IV_BYTES_SIZE);
+        cipher = crypto.createCipheriv(SecretEncryptionAlgo.AES_256_GCM, crypto.createSecretKey(key, "base64"), iv);
+      }
+
+      let ciphertext = cipher.update(plaintext, "utf8", "base64");
+      ciphertext += cipher.final("base64");
```
sheensantoscapadngan: if we want this to be reusable throughout the app, the algorithm to use here should ideally be passed in to the methods right?

Dockerfile.fips.standalone-infisical
```diff
@@ -173,13 +188,21 @@ ENV STANDALONE_MODE true
...
+ENV NODE_OPTIONS=--force-fips
+ENV FIPS_ENABLED=true
+  
+
 WORKDIR /backend
 
 ENV TELEMETRY_ENABLED true
 
 EXPOSE 8080
+EXPOSE 80
```
sheensantoscapadngan: what is this for??

backend/src/ee/services/dynamic-secret/providers/aws-elasticache.ts
```diff
@@ -39,8 +41,11 @@ type TDeleteElastiCacheUserInput = z.infer<typeof DeleteElasticCacheUserSchema>;
 const ElastiCacheUserManager = (credentials: TBasicAWSCredentials, region: string) => {
   const elastiCache = new ElastiCache({
     region,
+    useFipsEndpoint: crypto.isFipsModeEnabled(),
+    sha256: CustomAWSHasher,
```
sheensantoscapadngan, line 45: do we want a custom sha256 hasher even for non FIPS?

backend/src/lib/crypto/cache.ts
```diff
@@ -1,10 +1,4 @@
...
 
 export const generateCacheKeyFromData = (data: unknown) =>
-  crypto
-    .createHash("md5")
-    .update(JSON.stringify(data))
-    .digest("base64")
-    .replace(/\+/g, "-")
-    .replace(/\//g, "_")
-    .replace(/=/g, "");
+  crypto.rawCrypto.createHash("sha256").update(JSON.stringify(data)).digest("base64");
```
sheensantoscapadngan: why'd we remove the cleanup for `+`, `/`, and `=` here? are these not needed anymore?

backend/src/lib/crypto/cryptography/asymmetric-fips.ts
```diff
@@ -0,0 +1,124 @@
+import crypto from "crypto";
+
+import { SecretEncryptionAlgo } from "@app/db/schemas";
+import { CryptographyError } from "@app/lib/errors";
+
+export const asymmetricFipsValidated = () => {
+  const generateKeyPair = () => {
+    const { publicKey, privateKey } = crypto.generateKeyPairSync("x25519");
```
sheensantoscapadngan: is there no async equivalent?

backend/src/lib/crypto/cryptography/crypto.ts
```diff
@@ -0,0 +1,450 @@
...
+      compareHash
+    };
+  };
+  const jwt = () => {
+    $checkIsInitialized();
+
+    const sign = (payload: JWTPayload, secretOrKey: JWTSecretOrKey, options: JWTSignOptions = {}) => {
+      if (isFipsModeEnabled()) {
+        return jwtFipsValidated().sign(payload, secretOrKey, options);
+      }
```
sheensantoscapadngan: let's just use `jsonwebtoken`

backend/src/lib/crypto/cryptography/jwt-fips.ts
```diff
@@ -0,0 +1,327 @@
+import crypto from "crypto";
```
sheensantoscapadngan: remove for now

backend/src/ee/services/dynamic-secret/providers/aws-elasticache.ts
varonix0: Yes this is fine, it's just hashing. I don't see a reason to introduce extra conditional complexity for this

frontend/src/hooks/api/cmeks/mutations.tsx
```diff
@@ -66,6 +66,7 @@ export const useDeleteCmek = () => {
 export const useCmekEncrypt = () => {
   return useMutation({
     mutationFn: async ({ keyId, plaintext, isBase64Encoded }: TCmekEncrypt) => {
+      console.log("CALLING CMEK ENCRYPT");
```
sheensantoscapadngan: remove

frontend/src/pages/admin/SignUpPage/SignUpPage.tsx
```diff
@@ -49,7 +49,20 @@ export const SignUpPage = () => {
...
     try {
-      const { privateKey, ...userPass } = await generateUserPassKey(email, password);
+      console.log(
+        "Creating admin user...",
+        JSON.stringify({
+          email,
+          password,
+          fipsEnabled: config.fipsEnabled
+        })
+      );
```
sheensantoscapadngan: remove

frontend/src/pages/auth/SignUpSsoPage/components/UserInfoSSOStep/UserInfoSSOStep.tsx
```diff
@@ -94,12 +94,7 @@ export const UserInfoSSOStep = ({
 
     if (!errorCheck) {
       // Generate a random pair of a public and a private key
-      const pair = nacl.box.keyPair();
-      const secretKeyUint8Array = pair.secretKey;
-      const publicKeyUint8Array = pair.publicKey;
-      const privateKey = encodeBase64(secretKeyUint8Array);
-      const publicKey = encodeBase64(publicKeyUint8Array);
-      localStorage.setItem("PRIVATE_KEY", privateKey);
+      const { publicKey, privateKey } = await generateKeyPair(config.fipsEnabled);
```
sheensantoscapadngan, line 97: we're not setting PRIVATE_KEY anymore?

frontend/src/pages/secret-manager/SettingsPage/components/RebuildSecretIndicesSection/RebuildSecretIndicesSection.tsx
```diff
@@ -1,93 +0,0 @@
-import { createNotification } from "@app/components/notifications";
```
sheensantoscapadngan, line 1: why did we remove this?

frontend/src/pages/auth/SignUpSsoPage/components/UserInfoSSOStep/UserInfoSSOStep.tsx
varonix0: Great catch Sheen!

frontend/src/pages/secret-manager/SettingsPage/components/RebuildSecretIndicesSection/RebuildSecretIndicesSection.tsx
varonix0: It's not needed anymore, also ran it by Akhi. I wanted to get rid of all unused crypto in the frontend, and removing this would allow us to get rid of almost everything

sheensantoscapadngan (dismissed):

varonix0:
Merging at a later time as this is high-impact

backend/src/db/migrations/utils/env-config.ts
```diff
@@ -49,5 +51,19 @@ export const getMigrationEnvConfig = () => {
     process.exit(-1);
   }
 
-  return Object.freeze(parsedEnv.data);
+  let envCfg = Object.freeze(parsedEnv.data);
+
+  const fipsEnabled = await crypto.initialize(superAdminDAL);
+
+  if (fipsEnabled) {
```
akhilmhdh, line 63: Can you please put a detailed comment about logic flow made here? Also about one in environment config file on top of Root Encryption key

backend/src/db/seeds/5-machine-identity.ts
```diff
@@ -54,7 +55,9 @@ export async function seed(knex: Knex): Promise<void> {
       }
     ])
     .returning("*");
-  const clientSecretHash = await bcrypt.hash(seedData1.machineIdentity.clientCredentials.secret, 10);
+
+  const clientSecretHash = await crypto.hashing().createHash(seedData1.machineIdentity.clientCredentials.secret, 10);
```
akhilmhdh, line 59: I like the convention followed here similiar to SDK

backend/src/ee/services/group/group-fns.ts
```diff
@@ -94,26 +94,25 @@ const addAcceptedUsersToGroup = async ({
         });
       }
 
-      const botPrivateKey = infisicalSymmetricDecrypt({
+      const botPrivateKey = crypto.encryption().decryptWithRootEncryptionKey({
```
akhilmhdh: just my feedback but switch from -symmetric and then with decryptWithRootEncryptionKey feels weird

backend/package.json
```diff
@@ -188,6 +190,7 @@
     "cassandra-driver": "^4.7.2",
     "connect-redis": "^7.1.1",
     "cron": "^3.1.7",
+    "crypto-js": "^4.2.0",
```
akhilmhdh: I am confused by this libraries purpose

backend/src/lib/crypto/cryptography/asymmetric-fips.ts
```diff
@@ -0,0 +1,138 @@
+import crypto, { KeyObject } from "crypto";
```
akhilmhdh: Would recommend to prefix it as `node:crypto`

backend/src/lib/crypto/cryptography/crypto.ts
```diff
@@ -0,0 +1,431 @@
...
+    case DigestType.Base64:
+      encoder = cryptoJs.enc.Base64;
+      break;
+    default:
+      throw new CryptographyError({
+        message: `Invalid digest type: ${digest as string}`
+      });
+  }
+
+  return cryptoJs.MD5(message).toString(encoder);
```
akhilmhdh: This can be done with ndoejs crypto as well

akhilmhdh: On FIPS this will be disabled in nodeJS

backend/src/db/seeds/5-machine-identity.ts
varonix0: Thank you! I figured we may as well follow our SDK conventions

backend/src/ee/services/group/group-fns.ts
varonix0: But `infisicalSymmetricDecrypt` is decrypting with root key 🤔

akhilmhdh: My thinking was - the other ones follows more one purpose utility function. Can we replace this with may be`.symmetricDecrypt`?

backend/package.json
```diff
@@ -84,7 +84,9 @@
     "@babel/plugin-syntax-import-attributes": "^7.24.7",
     "@babel/preset-env": "^7.18.10",
     "@babel/preset-react": "^7.24.7",
+    "@smithy/types": "^4.3.1",
     "@types/bcrypt": "^5.0.2",
+    "@types/crypto-js": "^4.2.2",
```
maidul98: i see that this has been discontinued, sounds like there is a even more native version? https://www.npmjs.com/package/crypto-js

backend/src/db/migrations/20250602155451_fix-secret-versions.ts
```diff
@@ -3,11 +3,12 @@ import { Knex } from "knex";
 
 import { chunkArray } from "@app/lib/fn";
 import { selectAllTableCols } from "@app/lib/knex";
-import { logger } from "@app/lib/logger";
+import { initLogger, logger } from "@app/lib/logger";
 
 import { SecretType, TableName } from "../schemas";
 
 export async function up(knex: Knex): Promise<void> {
+  initLogger();
```
maidul98: we don't need to add initLogger(); everywhere since it is added to the knexfile https://github.com/Infisical/infisical/pull/3985/files

backend/src/db/migrations/utils/env-config.ts
```diff
@@ -49,5 +51,24 @@ export const getMigrationEnvConfig = () => {
...
+
+  // Fix for 128-bit entropy encryption key expansion issue:
+  // In FIPS it is not ideal to expand a 128-bit key into 256-bit. We solved this issue in the past by creating the ROOT_ENCRYPTION_KEY.
+  // If FIPS mode is enabled, we set the value of ROOT_ENCRYPTION_KEY to the value of ENCRYPTION_KEY.
+  // ROOT_ENCRYPTION_KEY is expected to be a 256-bit base64-encoded key, unlike the 32-byte key of ENCRYPTION_KEY.
+  // When ROOT_ENCRYPTION_KEY is set, our cryptography will always use a 256-bit entropy encryption key. So for the sake of FIPS we should just roll over the value of ENCRYPTION_KEY to ROOT_ENCRYPTION_KEY.
+  if (fipsEnabled) {
+    const newEnvCfg = {
+      ...envCfg,
+      ROOT_ENCRYPTION_KEY: envCfg.ENCRYPTION_KEY
```
maidul98, line 66: we don't really mention use of `ROOT_ENCRYPTION_KEY` in our docs too much. the reason for this is because we didn't want to create more confusion. i would leave it as ENCRYPTION_KEY but for people setting up fips, just mention to generate a 256 bit to start with. what do you think?

backend/src/ee/services/gateway/gateway-service.ts
```diff
@@ -149,9 +148,9 @@ export const gatewayServiceFactory = ({
 
       const alg = keyAlgorithmToAlgCfg(CertKeyAlgorithm.RSA_2048);
       // generate root CA
-      const rootCaKeys = await crypto.subtle.generateKey(alg, true, ["sign", "verify"]);
+      const rootCaKeys = await crypto.rawCrypto.subtle.generateKey(alg, true, ["sign", "verify"]);
```
maidul98: nit: instead of rawCrypto, i think `nativeCrypto` might better describe the source

backend/src/lib/crypto/cryptography/crypto.ts
```diff
@@ -0,0 +1,444 @@
...
+    $isInitialized = true;
+  };
+
+  const initialize = async (superAdminDAL: TSuperAdminDALFactory) => {
+    if ($isInitialized) {
+      return isFipsModeEnabled();
+    }
+
+    if (process.env.FIPS_ENABLED !== "true") {
+      logger.info("[FIPS]: Instance is running in non-FIPS mode.");
```
maidul98: I think this would raise questions for existing people. we should only say something when it is enabled

docker-compose.dev.yml
```diff
@@ -60,7 +60,7 @@ services:
     container_name: infisical-dev-api
     build:
       context: ./backend
-      dockerfile: Dockerfile.dev
+      dockerfile: Dockerfile.dev.fips
```
maidul98, line 63: not sure if my github is glitching but i saw no changes in our stand alone docker. is it safe to assume there is zero change there?

backend/src/lib/crypto/cryptography/crypto.ts
```diff
@@ -0,0 +1,444 @@
...
+    const serverCfg = await superAdminDAL.findById(ADMIN_CONFIG_DB_UUID).catch(() => null);
+
+    // if fips mode is enabled, we need to check if the deployment is a new deployment or an old one.
+    if (serverCfg) {
+      if (serverCfg.fipsEnabled) {
+        logger.info("[FIPS]: Instance is configured for FIPS mode of operation. Continuing startup with FIPS enabled.");
+        $setFipsModeEnabled(true);
+        return true;
+      }
+      logger.info("[FIPS]: Instance age predates FIPS mode inception date. Continuing without FIPS.");
```
maidul98, line 158: same for here

maidul98 (commented):
implementation looks good, just left some minor comments

backend/src/db/migrations/20250602155451_fix-secret-versions.ts
varonix0: Right, I added this before the patch was rolled out

backend/package.json
varonix0: We are locking it to a specific version. Crypto-js is super light weight. We considered node forge, but it's almost 2MB compared to crypto js which is around 400kB. We are _only_ using it for a single call to md5. I spoke to Akhi about this as well and I decided that crypto-js would be fine in this case

backend/src/lib/crypto/cryptography/crypto.ts
varonix0: I think this one makes sense. This will only be relevant for deployments on `infisical/infisical-fips` mode that predates the introduction of FIPS

docker-compose.dev.yml
varonix0: That's right, no changes are needed for the standalone docker image!

backend/src/lib/crypto/cryptography/crypto.ts
varonix0: I'll log out a generic message instead, `Cryptography module initialized in normal operation mode.`. Some indication on startup would be nice for logging and debugging purposes

backend/src/db/migrations/utils/env-config.ts
varonix0: Followed up internally in regards to this

backend/src/lib/crypto/cryptography/crypto.ts
maidul98: okay

backend/package.json
maidul98: but it seems like the native crypto has the same functionality and doesn't need a package?

varonix0: Removing support for MongoDB atlas as a resolution

PR #4078, opened 2025-07-12
feat: Checkly app connection and secrets sync

sidwebworks:
# Description 📣

This PR adds support for checklyhq.com app connection and secrets sync.

## Type ✨

- [ ] Bug fix
- [x] New feature
- [] Improvement
- [ ] Breaking change
- [x] Documentation

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/050e97df-f076-4b9c-8209-acaa65870170?refs=01d4277bbd50440a955072a6e496de2d809b0770&source=prChecksComment)

backend/src/services/app-connection/checkly/checkly-connection-schemas.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const ChecklyConnectionSchema = BaseChecklyConnectionSchema.extend({
+  method: ChecklyConnectionMethodSchema,
+  credentials: ChecklyConnectionAccessTokenCredentialsSchema
+});
+
+export const SanitizedChecklyConnectionSchema = z.discriminatedUnion("method", [
+  BaseChecklyConnectionSchema.extend({
+    method: ChecklyConnectionMethodSchema,
+    credentials: ChecklyConnectionAccessTokenCredentialsSchema.pick({})
```
scott-ray-wilson, line 33: question: any reason in particular you're thinking not to return `accountId`?

docs/docs.json
```diff
@@ -1328,6 +1315,20 @@
...
+                    "pages": [
+                      "api-reference/endpoints/app-connections/checkly/list",
+                      "api-reference/endpoints/app-connections/checkly/get-by-id",
+                      "api-reference/endpoints/app-connections/checkly/get-by-name",
+                      "api-reference/endpoints/app-connections/checkly/create",
+                      "api-reference/endpoints/app-connections/checkly/update",
+                      "api-reference/endpoints/app-connections/checkly/delete",
+                      "api-reference/endpoints/app-connections/checkly/sync-secrets",
+                      "api-reference/endpoints/app-connections/checkly/import-secrets",
+                      "api-reference/endpoints/app-connections/checkly/remove-secrets"
```
scott-ray-wilson: Think these 3 might of been a copy/paste error, these are sync endpoints

docs/integrations/secret-syncs/checkly.mdx
```diff
@@ -0,0 +1,161 @@
...
+                    "secretPath": "/my-secrets",
+                    "isEnabled": true,
+                    "syncOptions": {
+                        "initialSyncBehavior": "overwrite-destination",
+                        "autoSyncEnabled": true,
+                        "disableSecretDeletion": false
+                    },
+                    "destinationConfig": {
+                        "accountId": "3c90c3cc-0d44-4b50-8888-8dd25736052a",
+                        "accountName": "user@company.com"
```
scott-ray-wilson: nit: it seems like account names aren't email addresses for checkly? maybe we could make this "Example Company" or something? (same for response)

backend/src/services/secret-sync/checkly/checkly-sync-schemas.ts
```diff
@@ -0,0 +1,43 @@
...
+import {
+  BaseSecretSyncSchema,
+  GenericCreateSecretSyncFieldsSchema,
+  GenericUpdateSecretSyncFieldsSchema
+} from "@app/services/secret-sync/secret-sync-schemas";
+import { TSyncOptionsConfig } from "@app/services/secret-sync/secret-sync-types";
+
+const ChecklySyncDestinationConfigSchema = z.object({
+  accountId: z.string().min(1, "Account ID is required"),
+  accountName: z.string().min(1, "Account Name is required")
```
scott-ray-wilson: let's add a max length since json isn't constrained like native `text` columns (I'm guilty of not doing this previously 😅 )

frontend/src/pages/secret-manager/IntegrationsListPage/components/SecretSyncsTab/SecretSyncTable/helpers/index.ts
```diff
@@ -164,6 +164,10 @@ export const getSecretSyncDestinationColValues = (secretSync: TSecretSync) => {
       primaryText = "Railway Project";
       secondaryText = destinationConfig.projectName;
       break;
+    case SecretSync.Checkly:
+      primaryText = "Checkly Account";
+      secondaryText = destinationConfig.accountName;
```
scott-ray-wilson: thoughts on inverting these (ie primary is the account name)?

docs/integrations/app-connections/checkly.mdx
scott-ray-wilson, line 1: I think we should add a `<Note/>` block about the minimum role required for each feature behind a tab section (only secret syncs now) - not sure if `Read & Write` is sufficient or a user needs `Admin` to sync secrets

backend/src/services/app-connection/checkly/checkly-connection-public-client.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  async upsertVariable(connection: TChecklyConnectionConfig, variable: TChecklyVariable) {
+    const res = await this.getVariable(connection, variable);
+
+    if (!res) {
+      return this.createVariable(connection, variable);
+    }
+
+    return res;
+  }
```
scott-ray-wilson: this doesn't handle updating values?

backend/src/services/app-connection/checkly/checkly-connection-public-client.ts
```diff
@@ -0,0 +1,172 @@
...
+        return null;
+      }
+
+      throw error;
+    }
+  }
+
+  async getChecklyAccounts(connection: TChecklyConnectionConfig) {
+    // This endpoint is in beta and might be subject to changes
+    // Refer: https://developers.checklyhq.com/reference/getv1accounts
```
scott-ray-wilson, line 176: is there a non-beta endpoint we could use to avoid any potential breaking changes?

backend/src/services/secret-sync/checkly/checkly-sync-schemas.ts
scott-ray-wilson: thinking out loud; I'm not sure we need a config for this? If the scope of the token is against an ID it won't work to select a different one?

Otherwise maybe we should remove accountId from the connection so that it can be used with multiple accounts - is there a way we can validate the token without an account ID?

backend/src/services/app-connection/checkly/checkly-connection-schemas.ts
sidwebworks: Not really and the fact that someone might need it when using the API makes total sense. 

Let's add it.

backend/src/services/secret-sync/checkly/checkly-sync-schemas.ts
sidwebworks: Oh yes, there are 2 cases we wanted to support;  basically when using an account token - 
1. we always have the default account id which comes from the token itself so the user can manage it's own account.
2. if the user as access to any other accounts we show it to them in a select list incase they want to change the destination to one of the accounts they're managing.

Inside the public client you will see, we always use the one from the destination config (we basically override it)

> there a way we can validate the token without an account ID?

None that they have documented tbh - all the requests are scoped by the Account ID header

docs/docs.json
sidwebworks:
```suggestion
```

docs/integrations/secret-syncs/checkly.mdx
sidwebworks: Yeah so the default account it makes uses your email address so that's how it came into the response - but yes makes sense to chang this

frontend/src/pages/secret-manager/IntegrationsListPage/components/SecretSyncsTab/SecretSyncTable/helpers/index.ts
sidwebworks: Sure - anything for consistency :)

backend/src/services/app-connection/checkly/checkly-connection-public-client.ts
sidwebworks: Nope - did inform Maidul about this, TLDR; incase they pull the rug on us - everything will still work the same way except the select input won't list out all their accounts on create. Which we can replace with a stable endpoint incase this breaks.

Added the comment for future reference

docs/integrations/app-connections/checkly.mdx
sidwebworks: Role in Infisical or third party?

scott-ray-wilson: Checkly role

docs/integrations/app-connections/checkly.mdx
```diff
@@ -97,9 +99,12 @@ Infisical supports the use of [API Keys](https://app.checklyhq.com/settings/user
           "credentialsHash": "7c2d371dec195f82a6a0d5b41c970a229cfcaf88e894a5b6395e2dbd0280661f",
           "app": "checkly",
           "method": "api-key",
-          "credentials": {}
+          "credentials": {
+              "accountId": "[ACCOUNT ID]"
```
scott-ray-wilson: don't think we need this now since we removed `accountId` from connection

also need to remove references to `accountId` from this page (post example above), UI form text, etc.

backend/src/services/app-connection/checkly/checkly-connection-public-client.ts
```diff
@@ -130,19 +137,26 @@ class ChecklyPublicClient {
...
-    const res = await this.getVariable(connection, variable);
+  async upsertVariable(connection: TChecklyConnectionConfig, accountId: string, variable: TChecklyVariable) {
+    const res = await this.getVariable(connection, accountId, variable);
 
     if (!res) {
-      return this.createVariable(connection, variable);
+      return this.createVariable(connection, accountId, variable);
     }
 
+    await this.updateVariable(connection, accountId, variable);
```
scott-ray-wilson, line 147: I'm getting a 404 from this update when  I change a value in Infisical

scott-ray-wilson (changes requested):
two more comments

PR #4251, opened 2025-07-28
Separate Azure OAuth env vars to different env variables for each app connection

carlosmonastyrski:
# Description 📣

Previously, a single app connection was used to handle all Azure App Connection requests, which resulted in unnecessarily broad permission requests across all possible connection types. This change separates Azure App Connections into dedicated environment variables, with each connection isolated to its specific use case, reducing the permission scope for each connection to only what's actually needed. The original unified connection is maintained as a fallback to ensure backward compatibility and prevent breaking existing configurations, providing a seamless transition path while improving security through the principle of least privilege.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/bb2861ae-4da2-4e97-949d-68d6cd286256?refs=f265fa6d374f4b3736e74e1626b5286be930049c&source=prChecksComment)

backend/src/lib/config/env.ts
```diff
@@ -264,6 +264,14 @@ const envSchema = z
     // azure app
     INF_APP_CONNECTION_AZURE_CLIENT_ID: zpStr(z.string().optional()),
     INF_APP_CONNECTION_AZURE_CLIENT_SECRET: zpStr(z.string().optional()),
+    INF_APP_CONNECTION_AZURE_APP_CONFIGURATION_CLIENT_ID: zpStr(z.string().optional()),
```
varonix0, line 269: Uf this is a lot. Can we split this up with comments?

// Azure App Configuration App Connection

// Azure Key Vault App Connection

etc

backend/src/lib/config/env.ts
```diff
@@ -461,6 +469,38 @@ export const overwriteSchema: {
       {
         key: "INF_APP_CONNECTION_AZURE_CLIENT_SECRET",
         description: "The Client Secret of your Azure application."
+      },
+      {
+        key: "INF_APP_CONNECTION_AZURE_APP_CONFIGURATION_CLIENT_ID",
+        description: "The Application (Client) ID of your Azure application."
```
varonix0, line 491: I think we should split this up into different categories. One for each type of app connection. Right now it's all under "Azure"

backend/src/services/app-connection/azure-app-configuration/azure-app-configuration-connection-fns.ts
```diff
@@ -29,9 +29,19 @@ export const validateAzureAppConfigurationConnectionCredentials = async (
...
+    INF_APP_CONNECTION_AZURE_CLIENT_SECRET,
+    INF_APP_CONNECTION_AZURE_APP_CONFIGURATION_CLIENT_ID,
+    INF_APP_CONNECTION_AZURE_APP_CONFIGURATION_CLIENT_SECRET,
+    SITE_URL
+  } = getConfig();
 
-  if (!INF_APP_CONNECTION_AZURE_CLIENT_ID || !INF_APP_CONNECTION_AZURE_CLIENT_SECRET) {
+  const azureClientId = INF_APP_CONNECTION_AZURE_APP_CONFIGURATION_CLIENT_ID || INF_APP_CONNECTION_AZURE_CLIENT_ID;
+  const azureClientSecret =
+    INF_APP_CONNECTION_AZURE_APP_CONFIGURATION_CLIENT_SECRET || INF_APP_CONNECTION_AZURE_CLIENT_SECRET;
```
varonix0: Instead of doing these `OR` operators everywhere, I think we should do it in the envSchema transform, so we have:

```typescript
{
APP_CONNECTION_APP_CONFIGURATION_CLIENT_ID: INF_APP_CONNECTION_AZURE_CLIENT_SECRET || INF_APP_CONNECTION_AZURE_APP_CONFIGURATION_CLIENT_SECRET
// ....
}
```

That way we have separate variables, and they all default to the "shared" `INF_APP_CONNECTION_AZURE_CLIENT_SECRET` variable as an example. Then we don't need to do OR operators all over the backend

.env.example
```diff
@@ -126,6 +126,18 @@ INF_APP_CONNECTION_GCP_SERVICE_ACCOUNT_CREDENTIAL=
 INF_APP_CONNECTION_AZURE_CLIENT_ID=
```
varonix0: Let's remove `INF_APP_CONNECTION_AZURE_CLIENT_ID` and `INF_APP_CONNECTION_AZURE_CLIENT_SECRET` from the env example

DanielHougaard (approved):

PR #4344, opened 2025-08-08
Fix SAML duplicate accounts when signing in the first time on an existing account

carlosmonastyrski:
# Description 📣

Fixed an issue with SAML connections where duplicate accounts were created when the same email address already existed before SAML setup. Previously, if a user had an existing account with an email address and then a SAML connection was created for that same email, the system would create a second account for the SAML provider, resulting in duplicate accounts. This fix resolves the situation by linking new SAML connections to existing accounts with matching email addresses, while requiring email verification for the SAML authentication method before allowing sign-in.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/7e02db9f-566a-4609-9bbe-977fd77bf59c?refs=34a83016176a57675cfda3d5b514c974fc990327&source=prChecksComment)

backend/src/services/user/user-service.ts
```diff
@@ -106,6 +126,8 @@ export const userServiceFactory = ({
 
     const userEmails = user?.email ? await userDAL.find({ email: user.email }) : [];
 
+    await userAliasDAL.update({ userId: user.id }, { isEmailVerified: true });
```
akhilmhdh: Actually same suggestion: Instead of bulk let's do it directly. If alias is in token let's do

userAliasDAL.updateById(aliasId, {})

backend/src/services/user/user-service.ts
```diff
@@ -48,17 +51,32 @@ export const userServiceFactory = ({
...
+      authType: string;
+      aliasId: string;
+      username: string;
+    };
     // akhilmhdh: case sensitive email resolution
     const users = await userDAL.findUserByUsername(username);
     const user = users?.length > 1 ? users.find((el) => el.username === username) : users?.[0];
     if (!user) throw new NotFoundError({ name: `User with username '${username}' not found` });
+    let { isEmailVerified } = user;
+    const userAlias = await userAliasDAL.findOne({ userId: user.id, aliasType: authType, id: aliasId });
```
akhilmhdh: Small correction: If alias id provided in token and userAlias not found throw an error as well

backend/src/services/user/user-service.ts
```diff
@@ -95,7 +113,9 @@ export const userServiceFactory = ({
     if (!user) throw new NotFoundError({ name: `User with username '${username}' not found` });
     if (!user.email)
       throw new BadRequestError({ name: "Failed to verify email verification code due to no email on user" });
-    if (user.isEmailVerified)
+
+    const userAliases = await userAliasDAL.find({ userId: user.id });
+    if (user.isEmailVerified && userAliases?.every((alias) => alias.isEmailVerified))
```
akhilmhdh: I would recommend instead of this - is alias check verification for that else user isEmailVerified one

backend/src/services/user/user-service.ts
```diff
@@ -48,17 +51,32 @@ export const userServiceFactory = ({
   groupProjectDAL,
   tokenService,
   smtpService,
-  permissionService
+  permissionService,
+  userAliasDAL
 }: TUserServiceFactoryDep) => {
-  const sendEmailVerificationCode = async (username: string) => {
+  const sendEmailVerificationCode = async (referer: string) => {
```
akhilmhdh: instead of referer you can send the token from the frontend directly

akhilmhdh (changes requested):

akhilmhdh:
Things tested
- Existing SAML user working
- Having a non SAML account and using the email with same from a SAML
- Two organization having different saml with same email
- Switching organization with SAML configured
- Tested the migration
- Code quality check

PR #4389, opened 2025-08-18
fix(cli): failing tests

varonix0:
# Description 📣

The CLI tests are failing on `init`. This happens because when we call the `init` function, we are checking if the user is already logged in. If the user isn't logged in, we'll trigger the login flow. This is a new addition, and in the past it would just throw an error if you tried to do init without being logged in.

The core issue is that when you do interactive login, it doesn't ask you to select an organization. But when we check if the user is logged in during the `init` command, we call the `/checkAuth` endpoint. The `/checkAuth` endpoint will throw an error if no organization is set on the user's access token.

The fix is to allow the `/checkAuth` api call without having an organization selected.

**Why this worked in the past:**
In the past we used to force the user to select an organization during login (see screenshot attached), but in the latest versions we don't ask users to select an organization during interactive login. This means when you log in with interactive mode and then try to use `init`, the authentication check will fail because no organization is selected.

CleanShot 2025-08-18 at 10 19 29@2x

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

---

### DIFF
--- a/backend/src/server/routes/v1/auth-router.ts
+++ b/backend/src/server/routes/v1/auth-router.ts
@@ -67,7 +67,7 @@ export const registerAuthRoutes = async (server: FastifyZodProvider) => {
         })
       }
     },
-    onRequest: verifyAuth([AuthMode.JWT]),
+    onRequest: verifyAuth([AuthMode.JWT], { requireOrg: false }),
     handler: () => ({ message: "Authenticated" as const })
   });
 


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/3dda5cb1-c228-4b4e-8aed-748833a3f34a?refs=5aeb823c9e073d0fd203d4b5c01c8ddcd104e44a&source=prChecksComment)

PR #4435, opened 2025-08-28
Improve Audit Logs permissions

carlosmonastyrski:
# Description 📣

Removed unused permission options for Audit Logs (only Read is currently supported). Leaving Create/Update/Delete visible could cause confusion, so these have been removed.

Also fixed the LogsSection component: it previously only checked for Org-level permissions when displaying Project-level audit logs. As a result, even if a user had the correct Project-level permissions, logs were not shown if the Org-level audit log permission was missing. This has now been corrected.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/ee/services/audit-log/audit-log-service.ts
+++ b/backend/src/ee/services/audit-log/audit-log-service.ts
@@ -6,9 +6,9 @@ import { getConfig } from "@app/lib/config/env";
 import { BadRequestError } from "@app/lib/errors";
 import { ActorType } from "@app/services/auth/auth-type";
 
-import { OrgPermissionActions, OrgPermissionSubjects } from "../permission/org-permission";
+import { OrgPermissionAuditLogsActions, OrgPermissionSubjects } from "../permission/org-permission";
 import { TPermissionServiceFactory } from "../permission/permission-service-types";
-import { ProjectPermissionActions, ProjectPermissionSub } from "../permission/project-permission";
+import { ProjectPermissionAuditLogsActions, ProjectPermissionSub } from "../permission/project-permission";
 import { TAuditLogDALFactory } from "./audit-log-dal";
 import { TAuditLogQueueServiceFactory } from "./audit-log-queue";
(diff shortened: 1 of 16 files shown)


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/3f540353-2049-4985-b69d-6f0093556961?refs=dc23abdb8623f81719b9f583d0262f37e94502b7&source=prChecksComment)

frontend/src/pages/organization/RoleByIDPage/components/RolePermissionsSection/OrgPermissionAuditLogsRow.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+enum Permission {
+  NoAccess = "no-access",
+  Custom = "custom"
+}
+
+const PERMISSION_ACTIONS = [
+  {
+    action: OrgPermissionAuditLogsActions.Read,
+    label: "View"
```
scott-ray-wilson: nit: we call it "View" at org-level but "Read" at project

CleanShot 2025-08-29 at 12 49 13@2x
CleanShot 2025-08-29 at 12 48 58@2x

backend/src/ee/services/audit-log/audit-log-service.ts
scott-ray-wilson, line 1: Ignore file placement:

Not sure if we should address this in this PR or need to address at all but I think we may want to filter project-level audit logs from the org-view if you don't have project-level audit log read and are not an org-admin. I think it makes sense for org admins to see all, but maybe members should only see audit logs of projects they have access to even at the org-level?

scott-ray-wilson (changes requested):
works great; one comment and a thought

PR #4505, opened 2025-09-09
Stop blocking secret references with no matching reference and improve UI edit secret behavior to better highlight this

carlosmonastyrski:
# Description 📣

Improvement on the secret reference feature to avoid blocking users from setting secret values with no matching references as that can also be a valid secret value. But also improve the UI to mark those with a red color when editing the secret so it's clear no secret matches the reference.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/62066a58-173e-49c5-9285-9da3e99c95a4?refs=47da8d306386459774f00fd8f12f3988fe8e8816&source=prChecksComment)

frontend/src/components/v2/InfisicalSecretInput/InfisicalSecretInput.tsx
scott-ray-wilson, line 1: If I have a secret with values not present in other environments I get this error when trying to expand the row (guessing it doesn't handle undefined)
CleanShot 2025-09-09 at 10 35 15@2x
CleanShot 2025-09-09 at 10 34 58@2x

scott-ray-wilson, line 1: the error display doesn't seem to persist in a few scenarios
https://github.com/user-attachments/assets/a7873d8b-1412-4cd5-a607-97082529aa21

scott-ray-wilson: @carlosmonastyrski still seeing this one btw

scott-ray-wilson: this still seems to be the case too? (should be gray)
CleanShot 2025-09-10 at 17 14 00@2x

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -177,58 +177,71 @@ export const secretV2BridgeServiceFactory = ({
     );
 
     const referencesFolderGroupByPath = groupBy(referredFolders.filter(Boolean), (i) => `${i?.envId}-${i?.path}`);
+
```
scott-ray-wilson, line 177: realized up above still throws an error if I do ${not.valid} because the environment isn't found - think this should be handled?

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
scott-ray-wilson, line 1: ignore fileplacement:

the secret value resolves correctly but for some reason the expansion prepends an environment and path - also feel like we should probably render this differently?

CleanShot 2025-09-10 at 17 33 26@2x

PR #4531, opened 2025-09-15
Infisical Version Upgrade Tool

carlosmonastyrski:
# Description 📣

Introduced a new tool at `/upgrade-path` to simplify version upgrades. It keeps track of known version issues and guides users on the correct upgrade path from version X to Y, including any required intermediate steps for fixes or adaptations.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/c9c59829-1d53-43e8-8981-21a4e68752e0?refs=fd02746579504b8abf673068341dd98c644c0f72&source=prChecksComment)

backend/src/services/upgrade-path/github-client.ts
```diff
@@ -0,0 +1,187 @@
...
+      throw error;
+    }
+
+    const data = (await response.json()) as T;
+    return { data, rateLimit };
+  } catch (error) {
+    clearTimeout(timeout);
+
+    if (error instanceof Error && error.name === "AbortError") {
+      throw new Error(`Request timeout after ${config.timeout}ms`);
```
varonix0, line 171: If it times out, shouldn't we retry it?

backend/src/server/routes/v1/upgrade-path-router.ts
```diff
@@ -0,0 +1,122 @@
...
+        })
+      }
+    },
+    handler: async (req) => {
+      try {
+        const { fromVersion, toVersion } = req.body;
+
+        const result = await req.server.services.upgradePath.calculateUpgradePath(fromVersion, toVersion);
+
+        req.log.info(
```
varonix0: Question: Why not use logger.info() directly?

backend/src/server/routes/v1/upgrade-path-router.ts
```diff
@@ -0,0 +1,122 @@
...
+          { pathLength: result.path.length, hasBreaking: result.breakingChanges.length > 0 },
+          "Upgrade path calculated"
+        );
+
+        return result;
+      } catch (error) {
+        req.log.error(error, "Failed to calculate upgrade path");
```
varonix0: Same as above

backend/src/services/upgrade-path/upgrade-path-service.ts
```diff
@@ -0,0 +1,274 @@
...
+
+export const upgradePathServiceFactory = ({ keyStore }: TUpgradePathServiceFactory) => {
+  const getGitHubReleases = async (): Promise<FormattedRelease[]> => {
+    const cacheKey = "upgrade-path:releases";
+
+    try {
+      const cached = await keyStore.getItem(cacheKey);
+      if (cached) return JSON.parse(cached) as FormattedRelease[];
+    } catch (error) {
+      // Cache miss, continue to fetch from source
```
varonix0: Would be good to have a logger.error() here I think. Doesn't the keystore return null on cache miss?

backend/src/services/upgrade-path/upgrade-path-service.ts
```diff
@@ -0,0 +1,274 @@
...
+  };
+
+  const getUpgradePathConfig = async (): Promise<Record<string, VersionConfig>> => {
+    const cacheKey = "upgrade-path:config";
+
+    try {
+      const cached = await keyStore.getItem(cacheKey);
+      if (cached) return JSON.parse(cached) as Record<string, VersionConfig>;
+    } catch (error) {
+      // Cache miss, continue to fetch from source
```
varonix0: Same as above

backend/src/services/upgrade-path/upgrade-path-service.ts
```diff
@@ -0,0 +1,274 @@
...
+        throw new Error("Invalid configuration file path");
+      }
+
+      const yamlContent = await readFile(yamlPath, "utf8");
+
+      if (yamlContent.length > 1024 * 1024) {
+        throw new Error("Config file too large");
+      }
+
+      const config = yaml.load(yamlContent, { schema: yaml.FAILSAFE_SCHEMA }) as UpgradePathConfig;
```
varonix0, line 78: I think we should validate the config against a zod schema. I would create a schema for the upgrade path config structure and parse it with `schema.parse(config)` instead of doing a typescript cast like `as UpgradePathConfig`

backend/upgrade-path.yaml
```diff
@@ -0,0 +1,26 @@
+# Upgrade Path Configuration File
```
varonix0, line 1:
```yaml
versions:
  v0.147.1:
    breaking_changes:
      - title: "Test test test"
        description: "Test"
        action: "Test"
    db_schema_changes: "Test"
    notes: "Test"
  v0.63.1:
    breaking_changes:
      - title: "Major update: Moved database system From MongoDB to PostgreSQL"
        description: "This update includes a major database schema change. Please review the migration guide and test thoroughly before deploying to production."
        action: "Review the migration guide and test thoroughly before deploying to production. Upgrade guide: https://infisical.com/docs/self-hosting/guides/mongo-to-postgres"
    db_schema_changes: "Major schema restructuring with table reorganization. Extended migration time: 3 minutes."
    notes: "Critical update requiring maintenance window. Test thoroughly before production deployment."
```


This doesn't seem to work. It still gives me an upgrade path without any warnings if I try to upgrade from `v0.147.0` to `v0.147.1`

DanielHougaard (changes requested):
Few comments; I also think we should also add a CI test to validate the `upgrade-path.yaml` file when a PR is raised that makes changes to the `upgrade-path.yaml` file

.github/workflows/validate-upgrade-path.yml
```diff
@@ -0,0 +1,212 @@
...
+      - name: Validate upgrade-path.yaml format
+        if: steps.check-changes.outputs.changed == 'true'
+        run: |
+          echo "Running upgrade-path.yaml validation..."
+
+          # Debug: Check if file exists
+          ls -la backend/upgrade-path.yaml || echo "File not found"
+
+          # Debug: Show last few lines
+          echo "Last 5 lines of file:"
```
varonix0: Is this debug still needed?

.github/workflows/validate-upgrade-path.yml
```diff
@@ -0,0 +1,212 @@
+name: "Validate Upgrade Path Configuration"
```
varonix0, line 1: Why did we decide to use Python? Couldn't we just use Node.js and validate it with zod against the same zod schema we use in the backend? I feel like it would be best to test it directly against the schema being used in the backend


```typescript
const breakingChangeSchema = z.object({
  title: z.string().min(1).max(200),
  description: z.string().min(1).max(1000),
  action: z.string().min(1).max(500)
});

const versionConfigSchema = z.object({
  breaking_changes: z.array(breakingChangeSchema).optional(),
  db_schema_changes: z.string().max(1000).optional(),
  notes: z.string().max(2000).optional()
});
```

backend/src/services/upgrade-path/upgrade-path-service.ts
```diff
@@ -20,25 +21,44 @@ const versionSchema = z
...
-      if (cached) return JSON.parse(cached) as FormattedRelease[];
+      if (cached) {
+        const cachedReleases = JSON.parse(cached) as FormattedRelease[];
+        if (cachedReleases.length > 0) {
+          return cachedReleases;
+        }
+      }
     } catch (error) {
-      // Cache miss, continue to fetch from source
+      logger.error(error, "Failed to retrieve releases from cache");
```
varonix0, line 40: Please add the error to the log

backend/src/services/upgrade-path/upgrade-path-service.ts
```diff
@@ -104,23 +122,28 @@ export const upgradePathServiceFactory = ({ keyStore }: TUpgradePathServiceFacto
...
+      version.tagName.replace(new RE2(/^infisical\/v?/), "").replace(new RE2(/-[a-zA-Z]+$/), ""),
+      `v${cleanVersionNumber}`,
+      cleanVersionNumber
     ];
 
     for (const key of possibleKeys) {
       const versionConfig = config[key];
       if (versionConfig?.breaking_changes?.length) {
-        return versionConfig.breaking_changes;
+        return versionConfig.breaking_changes as BreakingChange[];
```
varonix0: How come this cast is needed? If zod schema matches the expected type it shouldn't be necessary right?

DanielHougaard (changes requested):
Seems like we also didn't add the upgrade path link here:

CleanShot 2025-09-24 at 04 32 28@2x

.github/workflows/validate-upgrade-path.yml
```diff
@@ -0,0 +1,181 @@
...
+      - name: Check for changes in upgrade-path.yaml
+        id: check-changes
+        run: |
+          # For local testing with act, always run validation
+          if [ "${ACT:-false}" = "true" ]; then
+            echo "changed=true" >> $GITHUB_OUTPUT
+            echo "Running validation (local act mode)"
+          else
+            # Check if upgrade-path.yaml was modified in this PR
+            if git diff --name-only HEAD^ HEAD | grep -q "backend/upgrade-path.yaml"; then
```
varonix0: Is this needed? Doesn't the job already only run when changes to the upgrade-path.yaml file is detected?

```yaml
pull_request:
    types: [opened, synchronize]
    paths:
      - "backend/upgrade-path.yaml"
```

.github/workflows/validate-upgrade-path.yml
```diff
@@ -0,0 +1,181 @@
...
+      - name: Setup Node.js for YAML validation
+        if: steps.check-changes.outputs.changed == 'true'
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Create lightweight validation script
+        if: steps.check-changes.outputs.changed == 'true'
+        run: |
+          # Create a temporary package.json with only the required dependencies
```
varonix0: This can be improved by adding a script in the `backend/scripts` folder, we can call it `validate-upgrade-path-file.ts`. Then we just add a new script run in the package.json called something like `validate-upgrade-path`. Then the job itself would jsut run `npm run validate-upgrade-path` If it throws, the test should fail.

This way we can also import the existing `versionSchema`, `breakingChangeSchema`, and `versionConfigSchema` directly in the script.

backend/src/server/routes/v1/upgrade-path-router.ts
```diff
@@ -0,0 +1,123 @@
+import RE2 from "re2";
+import { z } from "zod";
+
+import { BadRequestError } from "@app/lib/errors";
+import { logger } from "@app/lib/logger";
+import { publicEndpointLimit } from "@app/server/config/rateLimiter";
+
+const versionSchema = z
```
varonix0: This is a duplicate, it's already defined in `upgrade-path-service.ts`

backend/package.json
```diff
@@ -87,6 +87,7 @@
     "@smithy/types": "^4.3.1",
     "@types/bcrypt": "^5.0.2",
     "@types/jmespath": "^0.15.2",
+    "@types/js-yaml": "^4.0.9",
```
varonix0, line 91: q: Why did we add `js-yaml` as a dev dependency? It's already installed as a normal dependency

DanielHougaard (changes requested):
few more comments

package.json
```diff
@@ -25,6 +25,7 @@
   },
   "dependencies": {
     "@radix-ui/react-radio-group": "^1.1.3",
+    "js-yaml": "^4.1.0",
```
varonix0: No dependencies in root package.json release

backend/scripts/validate-upgrade-path-file.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    console.log("upgrade-path.yaml format is valid");
+  } catch (error) {
+    console.error(`Validation failed: ${error instanceof Error ? error.message : "Unknown error"}`);
+    process.exit(1);
+  }
+}
+
+if (require.main === module) {
```
varonix0: q: why is this needed?

.github/workflows/validate-upgrade-path.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   validate-upgrade-path:
     name: Validate upgrade-path.yaml
     runs-on: ubuntu-latest
-    timeout-minutes: 3
+    timeout-minutes: 15
```
varonix0, line 17: q: this takes 15 minutes to run?

DanielHougaard (changes requested):

PR #4565, opened 2025-09-20
Improve 2FA flow

carlosmonastyrski:
# Description 📣

Improved 2FA flow to be more intuitive, separated the recovery tokens from the normal 2FA token (previously check both on the same flow). Also force users to download the recovery codes in order to minimize the possibility of locked users.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/components/auth/Mfa.tsx
+++ b/frontend/src/components/auth/Mfa.tsx
@@ -6,9 +6,9 @@ import { t } from "i18next";
 import Error from "@app/components/basic/Error";
 import TotpRegistration from "@app/components/mfa/TotpRegistration";
 import SecurityClient from "@app/components/utilities/SecurityClient";
-import { Button, Input } from "@app/components/v2";
+import { Button } from "@app/components/v2";
 import { useSendMfaToken } from "@app/hooks/api";
-import { checkUserTotpMfa, verifyMfaToken } from "@app/hooks/api/auth/queries";
+import { checkUserTotpMfa, verifyMfaToken, verifyRecoveryCode } from "@app/hooks/api/auth/queries";
 import { MfaMethod } from "@app/hooks/api/auth/types";
 
 // The style for the verification code input
@@ -17,10 +17,10 @@ const codeInputProps = {
     fontFamily: "monospace",
     margin: "4px",
     MozAppearance: "textfield",
-    width: "48px",
+    width: "55px",
     borderRadius: "5px",
     fontSize: "24px",
-    height: "48px",
+    height: "55px",
     paddingLeft: "7",
     backgroundColor: "#0d1117",
     color: "white",
@@ -60,6 +60,7 @@ type Props = {
 
 export const Mfa = ({ successCallback, closeMfa, hideLogo, email, method }: Props) => {
   const [mfaCode, setMfaCode] = useState("");
+  const [showRecoveryCodeInput, setShowRecoveryCodeInput] = useState(false);
   const navigate = useNavigate();
   const [isLoading, setIsLoading] = useState(false);
   const [isLoadingResend, setIsLoadingResend] = useState(false);
@@ -82,16 +83,24 @@ export const Mfa = ({ successCallback, closeMfa, hideLogo, email, method }: Prop
   const verifyMfa = async (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault();
 
+    if (!mfaCode.trim()) return;
+
     setIsLoading(true);
     try {
-      const { token } = await verifyMfaToken({
-        email,
-        mfaCode,
-        mfaMethod: method
-      });
+      let result;
+
+      if (method === MfaMethod.TOTP && showRecoveryCodeInput) {
+        result = await verifyRecoveryCode(mfaCode.trim());
+      } else {
+        result = await verifyMfaToken({
+          email,
+          mfaCode: mfaCode.trim(),
+          mfaMethod: method
+        });
+      }
 
       SecurityClient.setMfaToken("");
-      SecurityClient.setToken(token);
+      SecurityClient.setToken(result.token);
 
       await successCallback();
       if (closeMfa) {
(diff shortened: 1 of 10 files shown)


maidul98:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/maidul98/pr-checks/f913db01-3398-40aa-9ba0-4ac6f6648d25?refs=49d9bb99bcefe452cffb4dbcb25f1ed7e9630c1c&source=prChecksComment)

frontend/src/components/mfa/RecoveryCodesDownload.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const content = [
+      "Infisical Two-Factor Authentication Recovery Codes",
+      `Generated on: ${new Date().toLocaleString()}`,
+      "",
+      "Important: Store these codes in a safe place. Each code can only be used once.",
+      "If you lose access to your mobile authenticator, you can use these codes to regain access to your account.",
+      "",
+      "Recovery Codes:",
+      ...recoveryCodes.map((code, index) => `${index + 1}. ${code}`)
+    ].join("\n");
```
scott-ray-wilson: thought: I don't know if we should include Infisical; if someone finds this file we're giving them context to what these recovery codes apply to (same with name of file) - maybe we should leave it to the user to do this so they take responsibility of that - I don't think others add this context (brex, stripe, etc.)

opinion: I kind of feel like in the download we shouldn't add the number prefix? Not sure why but it feels off to me, don't think I've seen others do this (could be wrong though)

CleanShot 2025-09-22 at 10 47 26@2x

frontend/src/components/mfa/RecoveryCodesDownload.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+          <div className="rounded-lg border border-mineshaft-600 bg-mineshaft-900 p-4">
+            <div className="grid grid-cols-2 gap-x-6 gap-y-2 font-mono text-sm">
+              {recoveryCodes.map((code, index) => (
+                <div key={code} className="flex items-center text-mineshaft-200">
+                  <span className="w-6 text-mineshaft-400">{index + 1}.</span>
+                  <span>{code}</span>
+                </div>
+              ))}
+            </div>
```
scott-ray-wilson, line 73: nit: the spacing between 10 and the code isn't the same alignment as the other single digit numbers:

CleanShot 2025-09-22 at 10 48 07@2x

frontend/src/components/auth/Mfa.tsx
```diff
@@ -162,76 +171,106 @@ export const Mfa = ({ successCallback, closeMfa, hideLogo, email, method }: Prop
...
+            size="md"
+            type="submit"
+            isFullWidth
+            className="h-11 rounded-lg font-medium shadow-sm transition-all duration-200 hover:shadow-md"
+            colorSchema="primary"
+            variant="outline_bg"
+            isLoading={isLoading}
+          >
+            {String(t("mfa.verify"))}
+          </Button>
```
scott-ray-wilson, line 274: improvement: maybe this button shouldn't be clickable unless the code field is fully populated? I can waste a retry if I accidentally click it when it's empty or not fully populated

CleanShot 2025-09-22 at 10 57 51@2x

frontend/src/components/auth/Mfa.tsx
```diff
@@ -162,76 +171,106 @@ export const Mfa = ({ successCallback, closeMfa, hideLogo, email, method }: Prop
...
       </form>
       {method === MfaMethod.TOTP && (
-        <div className="mt-2 flex flex-row justify-center text-sm text-bunker-400">
+        <div className="mt-6 flex flex-col items-center gap-4 text-sm">
+          <button
+            type="button"
+            onClick={() => {
+              setShowRecoveryCodeInput(!showRecoveryCodeInput);
+              setMfaCode("");
+              setTriesLeft(undefined);
```
scott-ray-wilson: UI discrepancy, ignore what I said about backend I looked and saw we do increment: 

https://github.com/user-attachments/assets/af705ab8-1b46-497e-ba6b-4452df652d89

scott-ray-wilson (changes requested):
much better! few comments

frontend/src/components/auth/Mfa.tsx
```diff
@@ -162,79 +183,134 @@ export const Mfa = ({ successCallback, closeMfa, hideLogo, email, method }: Prop
...
+        <div className="mx-auto mt-6 flex w-full max-w-sm flex-col items-center justify-center text-center">
+          <Button
+            size="md"
+            type="submit"
+            isFullWidth
+            className="h-11 rounded-lg font-medium shadow-sm transition-all duration-200 hover:shadow-md"
+            colorSchema="primary"
+            variant="outline_bg"
+            isLoading={isLoading}
+            isDisabled={!isCodeComplete}
```
scott-ray-wilson: let's also disable this if `triesLeft` is defined and less than or equal to zero (can spam it negatively)

frontend/src/components/auth/Mfa.tsx
```diff
@@ -79,35 +82,53 @@ export const Mfa = ({ successCallback, closeMfa, hideLogo, email, method }: Prop
...
-
-            SecurityClient.setMfaToken("");
-            SecurityClient.setToken("");
-          }
-          return (left as number) - 1;
-        });
+      if (typeof triesLeft === "number") {
+        const newTriesLeft = triesLeft - 1;
+        setTriesLeft(newTriesLeft);
+        if (newTriesLeft <= 0) {
```
scott-ray-wilson, line 124: let's add a error toast saying you've been temporarily locked out and to try again later (or with the exact time if we get that in the response)

PR #4591, opened 2025-09-29
Add warning on same destination secret sync

carlosmonastyrski:
# Description 📣

Secret Syncs allow users to set different syncs using the same destination (same Azure Key Vault, same app on Supabase, etc), which is correct but if done without context that is being used on other syncs can lead to undesired behaviors. This PR address these situations by showing a warning when trying to create/update a new sync with the same destination as one or more existing connections.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/secret-sync/secret-sync-maps.ts
+++ b/backend/src/services/secret-sync/secret-sync-maps.ts
@@ -171,6 +171,14 @@ export const DESTINATION_DUPLICATE_CHECK_MAP: Record<SecretSync, DestinationDupl
     const newTargetEnv = newConfig.targetEnvironment as string | undefined;
 
     const wildcardValues = ["*", ""];
+
+    if (
+      (newConfig.scope as string) === "group"
+        ? existingConfig.groupId !== newConfig.groupId
+        : existingConfig.projectId !== newConfig.projectId
+    )
+      return false;
+
     // If either has wildcard, it conflicts with any targetEnvironment
     if (
       !existingTargetEnv ||
@@ -181,12 +189,7 @@ export const DESTINATION_DUPLICATE_CHECK_MAP: Record<SecretSync, DestinationDupl
       return true;
     }
 
-    return (
-      existingTargetEnv === newTargetEnv &&
-      ((newConfig.scope as string) === "group"
-        ? existingConfig.groupId === newConfig.groupId
-        : existingConfig.projectId === newConfig.projectId)
-    );
+    return existingTargetEnv === newTargetEnv;
   },
   [SecretSync.CloudflarePages]: defaultDuplicateCheck,
   [SecretSync.CloudflareWorkers]: defaultDuplicateCheck,


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/8e8efacd-fc92-4879-b776-c697df49bf7a?refs=eb1ae216a61140c44517b47da7b24ccf4ee21501&source=prChecksComment) |

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -696,6 +698,51 @@ export const secretSyncServiceFactory = ({
...
+    );
+
+    if (!destinationConfig || typeof destinationConfig !== "object") {
+      return { hasDuplicate: false };
+    }
+
+    try {
+      const existingSyncs = await secretSyncDAL.find({
+        destination
+      });
```
scott-ray-wilson: this is comparing against all syncs; we should probably at least limit it to organization syncs, otherwise for certain syncs this could be checking thousands at some point - probably safe to assume most conflicting syncs would fall within the same organization

backend/src/lib/fn/object.ts
scott-ray-wilson, line 1: ignore placement:

I don't think deep equality check is going to be sufficient for this check - some secret syncs can actual conflict without having equal destination configs (optional fields, behavioral config options, etc.). We may need to resolve this per sync type and only compare the relevant fields.

We could just implement this for the high traffic syncs initially and expand as needed 

(video sent in slack because it's too big :/)

frontend/src/components/secret-syncs/forms/EditSecretSyncForm.tsx
```diff
@@ -60,6 +79,73 @@ export const EditSecretSyncForm = ({ secretSync, fields, onComplete }: Props) =>
     }
   };
 
+  useEffect(() => {
```
scott-ray-wilson, line 86: seems like there may be some circumstances when this doesn't retrigger when it should (see video)

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -698,10 +698,46 @@ export const secretSyncServiceFactory = ({
...
+      case SecretSync.AzureAppConfiguration:
+        return ["label"];
+      case SecretSync.AzureDevOps:
+        return ["devopsProjectName"];
+      case SecretSync.Checkly:
+        return ["groupName", "accountName"];
+      case SecretSync.DigitalOceanAppPlatform:
+        return ["appName"];
+      case SecretSync.GitLab:
+        return ["projectName", "shouldProtectSecrets", "shouldMaskSecrets", "shouldHideSecrets"];
```
scott-ray-wilson: think this should have groupName too, also target environment '*' might overwrite other explicit?

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -698,10 +698,46 @@ export const secretSyncServiceFactory = ({
     return updatedSecretSync as TSecretSync;
   };
 
+  const getSkipFieldsForDestination = (destination: SecretSync): string[] => {
+    switch (destination) {
```
scott-ray-wilson: github organization scope can conflict with each other by visibilty, but we couldn't do a compare, would need to check if one had 'all'; private / selected could technically overlap too but it'd be impossible to check unless we made a query

backend/src/services/secret-sync/secret-sync-service.ts
carlosmonastyrski: `groupName` is used on the Groups Scope sync and it's required (changes the group to sync secrets too), agree on the second one 👍

scott-ray-wilson (approved):
LGTM, let's come back and refactor to make this more clear for developers and consolidate the two checks

DexterWernJr:
> # Description 📣
> 
> 
> 
> Secret Syncs allow users to set different syncs using the same destination (same Azure Key Vault, same app on Supabase, etc), which is correct but if done without context that is being used on other syncs can lead to undesired behaviors. This PR address these situations by showing a warning when trying to create/update a new sync with the same destination as one or more existing connections.
> 
> 
> 
> 
> 
> 
> 
> ## Type ✨
> 
> 
> 
> - [ ] Bug fix
> 
> - [ ] New feature
> 
> - [ ] Improvement
> 
> - [ ] Breaking change
> 
> - [ ] Documentation
> 
> 
> 
> # Tests 🛠️
> 
> 
> 
> 
> 
> 
> 
> ```sh
> 
> # Here's some code block to paste some code snippets
> 
> ```
> 
> 
> 
> ---
> 
> 
> 
> - [ ] I have read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/getting-started/code-of-conduct). 📝 
> ****
> ****
> ****
>

mf4147100-art:
Hi

mf4147100-art:
```javascript
```

mf4147100-art:
``

PR #4637, opened 2025-10-09
Pki revamp

carlosmonastyrski:
# Description 📣

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/frontend/src/pages/cert-manager/CertificatesPage/components/CertificateIssuanceModal.tsx
+++ b/frontend/src/pages/cert-manager/CertificatesPage/components/CertificateIssuanceModal.tsx
@@ -597,8 +597,8 @@ export const CertificateIssuanceModal = ({ popUp, handlePopUpToggle, profileId }
                     >
                       <div className="space-y-2">
                         {value.map((san, index) => (
-                          // eslint-disable-next-line react/no-array-index-key
                           <div
+                            // eslint-disable-next-line react/no-array-index-key
                             key={`subject-alt-name-${index}`}
                             className="flex items-center gap-2"
                           >
@@ -746,80 +746,80 @@ export const CertificateIssuanceModal = ({ popUp, handlePopUpToggle, profileId }
                   <AccordionItem value="key-usages">
                     <AccordionTrigger>Key Usages</AccordionTrigger>
                     <AccordionContent>
-                        <div className="grid grid-cols-2 gap-2 pl-2">
-                          {filteredKeyUsages.map(({ label, value }) => {
-                            const isRequired = requiredKeyUsages.includes(value);
-                            return (
-                              <Controller
-                                key={label}
-                                control={control}
-                                name={`keyUsages.${value}` as any}
-                                render={({ field }) => (
-                                  <div className="flex items-center space-x-3">
-                                    <Checkbox
-                                      id={`key-usage-${value}`}
-                                      isChecked={field.value || false}
-                                      onCheckedChange={(checked) => {
+                      <div className="grid grid-cols-2 gap-2 pl-2">
+                        {filteredKeyUsages.map(({ label, value }) => {
+                          const isRequired = requiredKeyUsages.includes(value);
+                          return (
+                            <Controller
+                              key={label}
+                              control={control}
+                              name={`keyUsages.${value}` as any}
+                              render={({ field }) => (
+                                <div className="flex items-center space-x-3">
+                                  <Checkbox
+                                    id={`key-usage-${value}`}
+                                    isChecked={field.value || false}
+                                    onCheckedChange={(checked) => {
                                       if (!isRequired) {
-                                          field.onChange(checked);
-                                        }
-                                      }}
+                                        field.onChange(checked);
+                                      }
+                                    }}
                                     isDisabled={isRequired}
+                                  />
+                                  <div className="flex items-center gap-2">
+                                    <FormLabel
+                                      id={`key-usage-${value}`}
+                                      className={`text-sm ${isRequired ? "text-mineshaft-200" : "cursor-pointer text-mineshaft-300"}`}
+                                      label={label}
                                     />
-                                    <div className="flex items-center gap-2">
-                                      <FormLabel
-                                        id={`key-usage-${value}`}
-                                        className={`text-sm ${isRequired ? "text-mineshaft-200" : "cursor-pointer text-mineshaft-300"}`}
-                                        label={label}
-                                      />
-                                      {isRequired && <span className="text-xs">(Required)</span>}
-                                    </div>
+                                    {isRequired && <span className="text-xs">(Required)</span>}
                                   </div>
-                                )}
-                              />
-                            );
-                          })}
-                        </div>
+                                </div>
+                              )}
+                            />
+                          );
+                        })}
+                      </div>
                     </AccordionContent>
                   </AccordionItem>
 
                   <AccordionItem value="extended-key-usages">
                     <AccordionTrigger>Extended Key Usages</AccordionTrigger>
                     <AccordionContent>
-                        <div className="grid grid-cols-2 gap-2 pl-2">
-                          {filteredExtendedKeyUsages.map(({ label, value }) => {
-                            const isRequired = requiredExtendedKeyUsages.includes(value);
-                            return (
-                              <Controller
-                                key={label}
-                                control={control}
-                                name={`extendedKeyUsages.${value}` as any}
-                                render={({ field }) => (
-                                  <div className="flex items-center space-x-3">
-                                    <Checkbox
-                                      id={`ext-key-usage-${value}`}
-                                      isChecked={field.value || false}
-                                      onCheckedChange={(checked) => {
+                      <div className="grid grid-cols-2 gap-2 pl-2">
+                        {filteredExtendedKeyUsages.map(({ label, value }) => {
+                          const isRequired = requiredExtendedKeyUsages.includes(value);
+                          return (
+                            <Controller
+                              key={label}
+                              control={control}
+                              name={`extendedKeyUsages.${value}` as any}
+                              render={({ field }) => (
+                                <div className="flex items-center space-x-3">
+                                  <Checkbox
+                                    id={`ext-key-usage-${value}`}
+                                    isChecked={field.value || false}
+                                    onCheckedChange={(checked) => {
                                       if (!isRequired) {
-                                          field.onChange(checked);
-                                        }
-                                      }}
+                                        field.onChange(checked);
+                                      }
+                                    }}
                                     isDisabled={isRequired}
+                                  />
+                                  <div className="flex items-center gap-2">
+                                    <FormLabel
+                                      id={`ext-key-usage-${value}`}
+                                      className={`text-sm ${isRequired ? "text-mineshaft-200" : "cursor-pointer text-mineshaft-300"}`}
+                                      label={label}
                                     />
-                                    <div className="flex items-center gap-2">
-                                      <FormLabel
-                                        id={`ext-key-usage-${value}`}
-                                        className={`text-sm ${isRequired ? "text-mineshaft-200" : "cursor-pointer text-mineshaft-300"}`}
-                                        label={label}
-                                      />
-                                      {isRequired && <span className="text-xs">(Required)</span>}
-                                    </div>
+                                    {isRequired && <span className="text-xs">(Required)</span>}
                                   </div>
-                                )}
-                              />
-                            );
-                          })}
-                        </div>
+                                </div>
+                              )}
+                            />
+                          );
+                        })}
+                      </div>
                     </AccordionContent>
                   </AccordionItem>
                 </Accordion>
--- a/frontend/src/pages/cert-manager/PoliciesPage/PoliciesPage.tsx
+++ b/frontend/src/pages/cert-manager/PoliciesPage/PoliciesPage.tsx
@@ -29,6 +29,7 @@ export const PoliciesPage = () => {
       </Helmet>
       <div className="mx-auto mb-6 w-full max-w-7xl">
         <PageHeader
+          scope="project"
           title="Certificate Policies"
           description="Manage certificate templates and profiles for unified certificate issuance"
         />
--- a/frontend/src/pages/cert-manager/PoliciesPage/components/CertificateTemplatesV2Tab/shared/certificate-constants.ts
+++ b/frontend/src/pages/cert-manager/PoliciesPage/components/CertificateTemplatesV2Tab/shared/certificate-constants.ts
@@ -148,7 +148,6 @@ export const formatSanEffect = (effect: CertSanEffect): string => {
   }
 };
 
-
 export const SAN_TYPE_OPTIONS = Object.values(CertSubjectAlternativeNameType);
 export const KEY_USAGE_OPTIONS = Object.values(CertKeyUsageType);
 export const EXTENDED_KEY_USAGE_OPTIONS = Object.values(CertExtendedKeyUsageType);
@@ -165,7 +164,7 @@ export const USAGE_STATES = {
   OPTIONAL: "optional"
 } as const;
 
-export type UsageState = typeof USAGE_STATES[keyof typeof USAGE_STATES] | undefined;
+export type UsageState = (typeof USAGE_STATES)[keyof typeof USAGE_STATES] | undefined;
 
 export const TEMPLATE_SIGNATURE_ALGORITHMS = [
   "SHA256-RSA",
@@ -225,4 +224,3 @@ export const mapTemplateKeyAlgorithmToApi = (templateFormat: string): string => 
   };
   return mapping[templateFormat] || templateFormat;
 };
-
--- a/frontend/src/pages/cert-manager/PoliciesPage/components/CertificateTemplatesV2Tab/shared/schemas.ts
+++ b/frontend/src/pages/cert-manager/PoliciesPage/components/CertificateTemplatesV2Tab/shared/schemas.ts
@@ -6,8 +6,8 @@ import {
   CertKeyUsageType,
   CertSubjectAlternativeNameType,
   CertSubjectAttributeType,
-  SUBJECT_ATTRIBUTE_INCLUDE_OPTIONS,
-  SAN_INCLUDE_OPTIONS
+  SAN_INCLUDE_OPTIONS,
+  SUBJECT_ATTRIBUTE_INCLUDE_OPTIONS
 } from "./certificate-constants";
 
 export const uiAttributeSchema = z.object({


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/5af9fb58-e72a-49ed-9829-6c6146ba5ac1?refs=87a91ce4238ccd729d32f130a6db5df70df4be31&source=prChecksComment) |

backend/src/@types/knex.d.ts
```diff
@@ -656,6 +668,26 @@ declare module "knex/types/tables" {
...
+    [TableName.EstEnrollmentConfig]: KnexOriginal.CompositeTableType<
+      TEstEnrollmentConfigs,
+      TEstEnrollmentConfigsInsert,
+      TEstEnrollmentConfigsUpdate
+    >;
+    [TableName.ApiEnrollmentConfig]: KnexOriginal.CompositeTableType<
+      TApiEnrollmentConfigs,
+      TApiEnrollmentConfigsInsert,
+      TApiEnrollmentConfigsUpdate
+    >;
```
sheensantoscapadngan: I think we should be prefixing these with PKI so there's less ambiguity with the rest of the platform

backend/src/db/migrations/20251007133321_pki-v3-tables.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  if (!(await knex.schema.hasTable(TableName.CertificateProfile))) {
+    await knex.schema.createTable(TableName.CertificateProfile, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.string("projectId").notNullable();
+      t.foreign("projectId").references("id").inTable(TableName.Project).onDelete("CASCADE");
+
+      t.uuid("caId").notNullable();
+      t.foreign("caId").references("id").inTable(TableName.CertificateAuthority).onDelete("CASCADE");
```
sheensantoscapadngan: should we be cascading deletions of CAs?

backend/src/db/migrations/20251007133321_pki-v3-tables.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      t.uuid("certificateTemplateId").notNullable();
+      t.foreign("certificateTemplateId").references("id").inTable(TableName.CertificateTemplateV2).onDelete("CASCADE");
```
sheensantoscapadngan: same question here - I don't think we should allow deletion of a certificate template if a profile is using it?

backend/src/db/migrations/20251007133321_pki-v3-tables.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      t.uuid("estConfigId");
+      t.foreign("estConfigId").references("id").inTable(TableName.EstEnrollmentConfig).onDelete("SET NULL");
+
+      t.uuid("apiConfigId");
+      t.foreign("apiConfigId").references("id").inTable(TableName.ApiEnrollmentConfig).onDelete("SET NULL");
+
+      t.timestamps(true, true, true);
+
+      t.unique(["slug", "projectId"], { indexName: "certificate_profiles_slug_project_id_unique" });
```
sheensantoscapadngan: curious, why do we explicitly define the index name?

backend/src/db/schemas/index.ts
```diff
@@ -3,6 +3,7 @@ export * from "./access-approval-policies-approvers";
 export * from "./access-approval-policies-bypassers";
 export * from "./access-approval-requests";
 export * from "./access-approval-requests-reviewers";
+export * from "./api-enrollment-configs";
```
sheensantoscapadngan: same comment here about the the filename. I think we want to be explicit as possible given that we're multi-product

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -355,6 +355,19 @@ export enum EventType {
   CREATE_CERTIFICATE_TEMPLATE_EST_CONFIG = "create-certificate-template-est-config",
   UPDATE_CERTIFICATE_TEMPLATE_EST_CONFIG = "update-certificate-template-est-config",
   GET_CERTIFICATE_TEMPLATE_EST_CONFIG = "get-certificate-template-est-config",
+  CREATE_CERTIFICATE_TEMPLATE_V2 = "create-certificate-template-v2",
+  UPDATE_CERTIFICATE_TEMPLATE_V2 = "update-certificate-template-v2",
+  DELETE_CERTIFICATE_TEMPLATE_V2 = "delete-certificate-template-v2",
+  GET_CERTIFICATE_TEMPLATE_V2 = "get-certificate-template-v2",
+  LIST_CERTIFICATE_TEMPLATES_V2 = "list-certificate-templates-v2",
```
sheensantoscapadngan: I wonder if we could just keep this "create-certificate-template"? at least just the value, name can be kept as is

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2598,6 +2611,109 @@ interface GetCertificateTemplateEstConfig {
...
+    certificateTemplateId: string;
+  };
+}
+
+interface GetCertificateTemplateV2 {
+  type: EventType.GET_CERTIFICATE_TEMPLATE_V2;
+  metadata: {
+    certificateTemplateId: string;
+  };
+}
```
sheensantoscapadngan: add name or any recognizable identifier to the metadata other than IDs - from the audit log perspective, there's not much value in just the IDs itself

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2598,6 +2611,109 @@ interface GetCertificateTemplateEstConfig {
...
+    certificateProfileId: string;
+  };
+}
+
+interface GetCertificateProfile {
+  type: EventType.GET_CERTIFICATE_PROFILE;
+  metadata: {
+    certificateProfileId: string;
+  };
+}
```
sheensantoscapadngan, line 2664: same here - I guess with the others as well

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -454,7 +466,15 @@ const buildMemberPermissionRules = () => {
     ProjectPermissionSub.Certificates
   );
 
-  can([ProjectPermissionPkiTemplateActions.Read], ProjectPermissionSub.CertificateTemplates);
```
sheensantoscapadngan, line 457: no need for read certificate templates anymore? is this intentional?

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,387 @@
...
+  server.route({
+    method: "PATCH",
+    url: "/:id",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      hide: false,
+      tags: [ApiDocsTags.PkiCertificateProfiles],
+      params: getCertificateProfileByIdSchema,
```
sheensantoscapadngan: why are we using getVertificateProfileByIdSchema here ?

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,387 @@
...
+  server.route({
+    method: "GET",
+    url: "/:id/metrics",
+    config: {
+      rateLimit: readLimit
+    },
+    schema: {
+      hide: false,
+      tags: [ApiDocsTags.PkiCertificateProfiles],
+      params: getCertificateProfileByIdSchema,
```
sheensantoscapadngan: curious are we reusing these anywhere? I wonder if we can just define these here directly so there's much less files to seek.. now I have to click into it just so I can visualize the shape of the objects of an API endpoint

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,324 @@
...
+      body: z.object({
+        profileId: z.string().uuid(),
+        identifiers: z
+          .array(
+            z.object({
+              type: z.enum(["dns", "ip"]),
+              value: z.string()
+            })
+          )
+          .min(1),
```
sheensantoscapadngan: why are these called identifiers?

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
+
+interface OrderCertificateFromProfile {
+  type: EventType.ORDER_CERTIFICATE_FROM_PROFILE;
+  metadata: {
+    certificateProfileId: string;
+    orderId: string;
+    subjectAlternativeNames: string[];
   };
 }
```
sheensantoscapadngan, line 2700: same remark here about having a clear dentifier

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4051,13 +4125,22 @@ export type Event =
...
   | OrgAdminBypassSSOEvent
-  | CreateCertificateTemplate
-  | UpdateCertificateTemplate
-  | GetCertificateTemplate
-  | DeleteCertificateTemplate
   | CreateCertificateTemplateEstConfig
   | UpdateCertificateTemplateEstConfig
   | GetCertificateTemplateEstConfig
+  | CreateCertificateTemplate
+  | UpdateCertificateTemplate
```
sheensantoscapadngan, line 4159: IGNORE FILE PLACEMENT

failing lint 
image

backend/src/ee/services/license/__mocks__/license-fns.ts
```diff
@@ -33,7 +33,8 @@ export const getDefaultOnPremFeatures = () => {
     enterpriseSecretSyncs: false,
     enterpriseCertificateSyncs: false,
     enterpriseAppConnections: true,
-    machineIdentityAuthTemplates: false
+    machineIdentityAuthTemplates: false,
+    pkiLegacyTemplates: false
```
sheensantoscapadngan, line 37: Somehow I think `pkiLegacyTemplates` should be more for overriding the isEmpty(subscriber/template) check... Like if you set the value of this property to true then it will always just show the legacy section in the UI 
 
image

backend/src/services/certificate-profile/certificate-profile-service.ts
```diff
@@ -0,0 +1,679 @@
...
+  }
+};
+
+const validateEnrollmentConfigForUpdate = async (data: {
+  enrollmentType: EnrollmentType;
+  estConfigId?: string | null;
+  apiConfigId?: string | null;
+}): Promise<void> => {
+  if (data.enrollmentType === EnrollmentType.EST) {
+    if (!data.estConfigId) {
```
sheensantoscapadngan: aren't these more fitting to be in the zod validation layer in the API routes itself?

backend/src/services/certificate-profile/certificate-profile-service.ts
```diff
@@ -0,0 +1,679 @@
...
+
+    // Create the profile with the created config IDs
+    const { estConfig, apiConfig, ...profileData } = data;
+    const profile = await certificateProfileDAL.create({
+      ...profileData,
+      projectId,
+      estConfigId,
+      apiConfigId
+    });
+
```
sheensantoscapadngan: create these under one transaction so that changes are reverted if any of the creation fails

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+  server.route({
+    method: "POST",
+    url: "/",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      hide: false,
+      tags: [ApiDocsTags.PkiCertificateProfiles],
+      body: createCertificateProfileSchema,
```
sheensantoscapadngan: looks like we're still not defining these inline? please check with the other files

backend/src/server/routes/v1/certificate-template-router.ts
```diff
@@ -116,12 +117,8 @@ export const registerCertificateTemplateRouter = async (server: FastifyZodProvid
...
           metadata: {
             certificateTemplateId: certificateTemplate.id,
-            caId: certificateTemplate.caId,
-            pkiCollectionId: certificateTemplate.pkiCollectionId as string,
             name: certificateTemplate.name,
-            commonName: certificateTemplate.commonName,
-            subjectAlternativeName: certificateTemplate.subjectAlternativeName,
-            ttl: certificateTemplate.ttl
+            projectId: certificateTemplate.projectId
           }
```
sheensantoscapadngan, line 127: for the old certificate templates, I think we should still emit the same metadata? Maybe you can have it such that the metadata property could take two shapes? The one for the old cert template and the one for the new?

backend/src/server/routes/v2/certificate-templates-v2-router.ts
```diff
@@ -0,0 +1,267 @@
...
+      body: updateCertificateTemplateV2Schema,
+      response: {
+        200: z.object({
+          certificateTemplate: CertificateTemplatesV2Schema
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const certificateTemplate = await server.services.certificateTemplateV2.updateTemplateV2({
```
sheensantoscapadngan, line 288: nit - perhaps having the v2 in the service name is enough in this case? instead of defining it in both the method name and the service name. This applies to the other methods as well though maybe i'm missing something?

backend/src/server/routes/v2/certificate-templates-v2-router.ts
```diff
@@ -0,0 +1,267 @@
...
+        req.params.id,
+        req.body.request
+      );
+
+      return {
+        valid: result.isValid,
+        errors: result.errors.length > 0 ? result.errors : undefined
+      };
+    }
+  });
```
sheensantoscapadngan: it seems like we're not using this anywhere?

backend/src/services/certificate-template-v2/certificate-template-v2-service.ts
```diff
@@ -0,0 +1,650 @@
...
+    request: TCertificateRequest
+  ): Promise<TTemplateValidationResult> => {
+    const template = await certificateTemplateV2DAL.findById(templateId);
+    if (!template) {
+      throw new NotFoundError({ message: "Certificate template not found" });
+    }
+
+    return validateRequestAgainstPolicy(template, request);
+  };
+
```
sheensantoscapadngan, line 900: looks like we're not doing any user access validation here? so anyone can deduce the policies of any certificate template via the certificate template v2 endpoint

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+    url: "/issue-certificate",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      hide: false,
+      tags: [ApiDocsTags.PkiCertificates],
+      body: z.object({
+        profileId: z.string().uuid(),
+        commonName: validateTemplateRegexField,
```
sheensantoscapadngan: I think this validator was specifically built for cert templates - can we make a new one, probably a more generic one, for this use-case?

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+        ttl: z.string().refine((val) => ms(val) > 0, "TTL must be a positive number"),
+        keyUsages: z.nativeEnum(CertKeyUsage).array().optional(),
+        extendedKeyUsages: z.nativeEnum(CertExtendedKeyUsage).array().optional(),
+        notBefore: validateCaDateField.optional(),
+        notAfter: validateCaDateField.optional(),
+        altNames: validateAltNamesField.optional(),
+        signatureAlgorithm: z.string().optional(),
+        keyAlgorithm: z.string().optional()
```
sheensantoscapadngan: we don't have enums for these that we can validate here ?

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+      tags: [ApiDocsTags.PkiCertificates],
+      body: z.object({
+        profileId: z.string().uuid(),
+        commonName: validateTemplateRegexField,
+        ttl: z.string().refine((val) => ms(val) > 0, "TTL must be a positive number"),
+        keyUsages: z.nativeEnum(CertKeyUsage).array().optional(),
+        extendedKeyUsages: z.nativeEnum(CertExtendedKeyUsage).array().optional(),
+        notBefore: validateCaDateField.optional(),
+        notAfter: validateCaDateField.optional(),
+        altNames: validateAltNamesField.optional(),
```
sheensantoscapadngan: could it make more sense to refer to this as "subjectAltNames" instead? just trying to see if we can also finalize the property naming as part of this initiative since we're building out the new APIs

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+                  ip: "ip_address",
+                  email: "email",
+                  url: "uri"
+                } as const;
+                return {
+                  type: typeMapping[mappedType.type] as "dns_name" | "ip_address" | "email" | "uri",
+                  value: mappedType.value
+                };
+              })
+              .filter((item): item is NonNullable<typeof item> => item !== null)
```
sheensantoscapadngan: let's use an enum here for SAN types?

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+          : undefined,
+        validity: {
+          ttl: req.body.ttl
+        },
+        notBefore: req.body.notBefore ? new Date(req.body.notBefore) : undefined,
+        notAfter: req.body.notAfter ? new Date(req.body.notAfter) : undefined,
```
sheensantoscapadngan, line 116: what happens by the way if a user specifies both ttl and notBefore + notAfter? maybe we can add validation so that they only pass one or the other ?

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+        notAfter: req.body.notAfter ? new Date(req.body.notAfter) : undefined
+      });
+
+      const profile = await server.services.certificateProfile.getProfileById({
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId,
+        profileId: req.body.profileId
+      });
```
sheensantoscapadngan: hmm why do we do this here? can't we just return projectId from the `signCertificateFromProfile` ?

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+      body: z.object({
+        profileId: z.string().uuid(),
+        subjectAlternativeNames: z
+          .array(
+            z.object({
+              type: z.enum(["dns", "ip"]),
+              value: z.string()
+            })
+          )
+          .min(1),
```
sheensantoscapadngan: yeah I think we should definitely standardize the naming of the properties... in the above it's referred to as altNames but here its subjectAlternativeNames. We can just pick one and keep it consistent throughout

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+        ttl: z.string().refine((val) => ms(val) > 0, "TTL must be a positive number"),
+        keyUsages: z.nativeEnum(CertKeyUsage).array().optional(),
+        extendedKeyUsages: z.nativeEnum(CertExtendedKeyUsage).array().optional(),
+        notBefore: validateCaDateField.optional(),
+        notAfter: validateCaDateField.optional(),
+        commonName: validateTemplateRegexField.optional(),
+        signatureAlgorithm: z.string().optional(),
+        keyAlgorithm: z.string().optional()
```
sheensantoscapadngan: same comment here about enum and validation

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -0,0 +1,297 @@
...
+      }),
+      response: {
+        200: z.object({
+          orderId: z.string(),
+          status: z.enum(["pending", "processing", "valid", "invalid"]),
```
sheensantoscapadngan: yeah we should also be using enums here, together with the rest

backend/src/services/certificate-authority/internal/internal-certificate-authority-service.ts
```diff
@@ -1277,13 +1280,21 @@ export const internalCertificateAuthorityServiceFactory = ({
       throw new BadRequestError({ message: "notAfter date is after CA certificate's notAfter date" });
     }
 
-    const alg = keyAlgorithmToAlgCfg(ca.internalCa.keyAlgorithm as CertKeyAlgorithm);
-    const leafKeys = await crypto.nativeCrypto.subtle.generateKey(alg, true, ["sign", "verify"]);
+    // Use provided keyAlgorithm if available, otherwise fall back to CA's algorithm
+    const effectiveKeyAlgorithm =
+      (keyAlgorithm as CertKeyAlgorithm) || (ca.internalCa.keyAlgorithm as CertKeyAlgorithm);
```
sheensantoscapadngan: hmmm what happens though if the certificate template doesn't allow the key algorithm of the internal CA and that no key algorithm is passed in? Wouldn't that mean that a certificate can get issued that doesn't necessarily conform to the profile's certificate template?

perhaps you can also double check with the other parts where we're defaulting to whatever value is in the certificate authority

backend/src/services/certificate-common/certificate-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  if (!templateAttributes || templateAttributes.length === 0) {
+    Object.entries(attributeMap).forEach(([templateKey, requestKey]) => {
+      const value = request[requestKey];
+      if (value && typeof value === "string") {
+        subject[templateKey] = value;
+      }
+    });
+    return subject;
+  }
```
sheensantoscapadngan, line 67: why are we allowing the values automatically if no template attribute is defined? shouldn't we be erroring instead? The idea here is that the template should determine the fixed shape of output certificates. If there's anything that's passed into the issue endpoint for example that's not defined in the certificate template then we shouldn't be allowing it 

or.. maybe i'm misunderstanding this xD

backend/src/services/certificate-common/certificate-utils.ts
```diff
@@ -0,0 +1,118 @@
...
+  }>
+): string => {
+  if (!request.subjectAlternativeNames || request.subjectAlternativeNames.length === 0) {
+    return "";
+  }
+
+  if (!templateSans || templateSans.length === 0) {
+    return request.subjectAlternativeNames.map((san) => san.value).join(",");
+  }
+
```
sheensantoscapadngan, line 97: same question here

backend/src/services/certificate-template-v2/certificate-template-v2-service.ts
```diff
@@ -0,0 +1,650 @@
...
+      throw new Error("Template policy is required");
+    }
+
+    if (!policy.attributes || policy.attributes.length === 0) {
+      throw new Error("Template policy must include attributes array");
+    }
+
+    if (!policy.keyUsages || !policy.keyUsages.requiredUsages || !policy.keyUsages.optionalUsages) {
+      throw new Error("Template policy must include valid key usages configuration");
+    }
```
sheensantoscapadngan: why is `optionalUsages` and `requiredUsages` required here? note that you can have valid certificates without any key usage extsensions

backend/src/services/certificate-template-v2/certificate-template-v2-service.ts
```diff
@@ -0,0 +1,650 @@
...
+
+    if (policy.signatureAlgorithm) {
+      if (!policy.signatureAlgorithm.allowedAlgorithms.includes(policy.signatureAlgorithm.defaultAlgorithm)) {
+        throw new Error("Default signature algorithm must be in allowed algorithms list");
+      }
+    }
+
```
sheensantoscapadngan, line 73: are we missing subjectAlternativeNames and extendedKeyUsages here ?

note that for EKUs, the same logic applies - you can have a valid certificate with no EKU extensions

backend/src/services/certificate-template-v2/certificate-template-v2-service.ts
```diff
@@ -0,0 +1,650 @@
...
+          template.validity.minDuration.unit
+        );
+        if (requestDuration < minDuration) {
+          errors.push(`Requested validity period is below minimum required duration`);
+        }
+      }
+    }
+
+    return {
+      isValid: errors.length === 0,
```
sheensantoscapadngan: looks like we're missing validation for the `notBefore` and `notAfter` fields if that's what's defined?

backend/src/services/certificate-est-v3/certificate-est-v3-service.ts
```diff
@@ -0,0 +1,294 @@
+import * as x509 from "@peculiar/x509";
```
sheensantoscapadngan, line 1: looks like we're missing profile enrollment method validation checks here? like the one in the issue cert endpoint

backend/src/services/enrollment-config/est-enrollment-config-dal.ts
```diff
@@ -0,0 +1,76 @@
...
+      const profileCount = await (tx || db)(TableName.CertificateProfile)
+        .where({ estConfigId: configId })
+        .count("* as count")
+        .first();
+
+      return parseInt(profileCount || "0", 10) > 0;
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Check if EST enrollment config is in use" });
+    }
+  };
```
sheensantoscapadngan: it doesn't look like we're using this anywhere?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -0,0 +1,387 @@
...
+        challenges: [
+          {
+            type: "internal-validation",
+            status: "valid" as const,
+            url: `/api/v3/certificates/orders/${orderId}/internal`,
+            token: "internal-ca-validation"
+          }
+        ]
+      }));
+
```
sheensantoscapadngan: what is this part for?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -0,0 +1,387 @@
...
+      return {
+        orderId,
+        status: "valid",
+        subjectAlternativeNames,
+        authorizations,
+        finalize: `/api/v3/certificates/orders/${orderId}/finalize`,
```
sheensantoscapadngan: we don't have this yet right?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -0,0 +1,387 @@
...
+        certificate: certificateResult.certificate
+      };
+    }
+
+    if (caType === CaType.ACME) {
+      throw new BadRequestError({
+        message:
+          "ACME certificate ordering via profiles is not yet implemented. Use direct certificate issuance for ACME CAs."
+      });
+    }
```
sheensantoscapadngan, line 465: what do you mean here by direct certificate issuance for ACME CAs?

backend/src/services/certificate-template-v2/certificate-template-v2-schemas.ts
```diff
@@ -0,0 +1,167 @@
...
+  keyAlgorithm: templateV2KeyAlgorithmSchema.optional()
+});
+
+export const updateCertificateTemplateV2Schema = z.object({
+  slug: z
+    .string()
+    .min(1)
+    .max(255)
+    .regex(new RE2("^[a-z0-9-]+$"), "Slug must contain only lowercase letters, numbers, and hyphens")
+    .optional(),
```
sheensantoscapadngan: I think you should use the slugSchema method that we have

backend/src/services/certificate-template-v2/certificate-template-v2-schemas.ts
```diff
@@ -0,0 +1,167 @@
...
+    .max(255)
+    .regex(new RE2("^[a-z0-9-]+$"), "Slug must contain only lowercase letters, numbers, and hyphens"),
+  description: z.string().max(1000).optional(),
+  attributes: z.array(templateV2AttributeSchema).optional(),
+  keyUsages: templateV2KeyUsagesSchema.optional(),
+  extendedKeyUsages: templateV2ExtendedKeyUsagesSchema.optional(),
+  subjectAlternativeNames: z.array(templateV2SanSchema).optional(),
+  validity: templateV2ValiditySchema.optional(),
+  signatureAlgorithm: templateV2SignatureAlgorithmSchema.optional(),
+  keyAlgorithm: templateV2KeyAlgorithmSchema.optional()
```
sheensantoscapadngan: why make this optional here and check at the service layer?

backend/src/services/certificate-template-v2/certificate-template-v2-schemas.ts
```diff
@@ -0,0 +1,167 @@
...
+  defaultKeyType: z.string()
+});
+
+export const createCertificateTemplateV2Schema = z.object({
+  projectId: z.string().min(1),
+  slug: z
+    .string()
+    .min(1)
+    .max(255)
+    .regex(new RE2("^[a-z0-9-]+$"), "Slug must contain only lowercase letters, numbers, and hyphens"),
```
sheensantoscapadngan: same comment here re: using slugSchema

backend/src/services/certificate-profile/certificate-profile-service.ts
```diff
@@ -0,0 +1,679 @@
...
+    }
+
+    if (hasEnrollmentConfigChanges(data)) {
+      const mergedData = { ...existingProfile, ...data };
+      await validateEnrollmentConfigForUpdate({
+        enrollmentType: mergedData.enrollmentType as EnrollmentType,
+        estConfigId: mergedData.estConfigId,
+        apiConfigId: mergedData.apiConfigId
+      });
+    }
```
sheensantoscapadngan: I would put the mutations here inside a txn as well for consistency in error cases

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -352,9 +352,18 @@ export enum EventType {
   UPDATE_CERTIFICATE_TEMPLATE = "update-certificate-template",
   DELETE_CERTIFICATE_TEMPLATE = "delete-certificate-template",
   GET_CERTIFICATE_TEMPLATE = "get-certificate-template",
+  LIST_CERTIFICATE_TEMPLATES = "list-certificate-templates",
```
sheensantoscapadngan, line 355: IGNORE PLACEMENT OF COMMENT

Looks like we still allow people to create old certificate templates? I was expecting that we'd be throwing a descriptive error asking them to move to certificate profiles? Or maybe even disabling the "Create" button with a tooltip? I would think the same goes for subscribers

image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -352,9 +352,18 @@ export enum EventType {
...
   CREATE_CERTIFICATE_TEMPLATE_EST_CONFIG = "create-certificate-template-est-config",
   UPDATE_CERTIFICATE_TEMPLATE_EST_CONFIG = "update-certificate-template-est-config",
   GET_CERTIFICATE_TEMPLATE_EST_CONFIG = "get-certificate-template-est-config",
+  CREATE_CERTIFICATE_PROFILE = "create-certificate-profile",
+  UPDATE_CERTIFICATE_PROFILE = "update-certificate-profile",
+  DELETE_CERTIFICATE_PROFILE = "delete-certificate-profile",
+  GET_CERTIFICATE_PROFILE = "get-certificate-profile",
```
sheensantoscapadngan, line 362: IGNORE PLACEMENT OF COMMENT

Let's add a proper loading animation here? Same for the certificate templates tab

image

image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
   metadata: Record<string, string>; // no metadata yet
 }
 
-interface CreateCertificateTemplate {
-  type: EventType.CREATE_CERTIFICATE_TEMPLATE;
+interface OrgAdminAccessProjectEvent {
+  type: EventType.ORG_ADMIN_ACCESS_PROJECT;
```
sheensantoscapadngan, line 2525: IGNORE PLACEMENT OF COMMENT

I get this generic error when trying to create a certificate template with no details provided for some of the fields
image
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
-interface GetCertificateTemplate {
-  type: EventType.GET_CERTIFICATE_TEMPLATE;
+interface UpdateCertificateTemplateEstConfig {
+  type: EventType.UPDATE_CERTIFICATE_TEMPLATE_EST_CONFIG;
   metadata: {
     certificateTemplateId: string;
+    isEnabled: boolean;
+  };
+}
+
```
sheensantoscapadngan, line 2562: IGNORE PLACEMENT OF COMMENT

If a field is mandatory, maybe they shouldn't be able to submit the form if no value is inputted?
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
+interface GetCertificateTemplateEstConfig {
+  type: EventType.GET_CERTIFICATE_TEMPLATE_EST_CONFIG;
+  metadata: {
+    certificateTemplateId: string;
+  };
+}
+
```
sheensantoscapadngan, line 2569: IGNORE PLACEMENT OF COMMENT

Looks like the Required and Optional selections here still aren't converted to the checkbox model as discussed on Slack? 
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
-    subjectAlternativeName: string;
-    ttl: string;
   };
 }
 
 interface DeleteCertificateTemplate {
   type: EventType.DELETE_CERTIFICATE_TEMPLATE;
   metadata: {
     certificateTemplateId: string;
+    name: string;
```
sheensantoscapadngan, line 2613: IGNORE PLACEMENT OF COMMENT

Switching from API enrollment type to EST and then back to API causes issues where I couldn't submit the form..
image
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
     projectId: string;
-  }; // no metadata yet
+  };
 }
 
-interface OrgAdminBypassSSOEvent {
-  type: EventType.ORG_ADMIN_BYPASS_SSO;
-  metadata: Record<string, string>; // no metadata yet
+interface CreateCertificateProfile {
+  type: EventType.CREATE_CERTIFICATE_PROFILE;
```
sheensantoscapadngan, line 2633: IGNORE PLACEMENT OF COMMENT

From a UX perspective, I think I should be able to issue certificates from here?
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
+    name: string;
+    projectId: string;
+    enrollmentType: string;
+  };
 }
 
-interface CreateCertificateTemplateEstConfig {
-  type: EventType.CREATE_CERTIFICATE_TEMPLATE_EST_CONFIG;
+interface UpdateCertificateProfile {
+  type: EventType.UPDATE_CERTIFICATE_PROFILE;
```
sheensantoscapadngan, line 2643: IGNORE PLACEMENT OF COMMENT

I think we should be hiding the fields that you can't set on a certificate profile? 
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
 }
 
-interface GetAzureAdCsTemplatesEvent {
-  type: EventType.GET_AZURE_AD_TEMPLATES;
+interface DeleteCertificateProfile {
+  type: EventType.DELETE_CERTIFICATE_PROFILE;
   metadata: {
-    caId: string;
-    amount: number;
+    certificateProfileId: string;
```
sheensantoscapadngan, line 2653: IGNORE PLACEMENT OF COMMENT 

Remove the V2 here please, along with the other relevant places
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
+}
+
+interface IssueCertificateFromProfile {
+  type: EventType.ISSUE_CERTIFICATE_FROM_PROFILE;
+  metadata: {
+    certificateProfileId: string;
+    certificateId: string;
+    commonName: string;
+  };
+}
```
sheensantoscapadngan, line 2681: IGNORE PLACEMENT OF COMMENT

why do we show this Mandatory|Optional|Prohibited field here in the issue certificate flow? 

image
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
-interface GetCertificateTemplateEstConfig {
-  type: EventType.GET_CERTIFICATE_TEMPLATE_EST_CONFIG;
+interface ListCertificateProfiles {
+  type: EventType.LIST_CERTIFICATE_PROFILES;
   metadata: {
-    certificateTemplateId: string;
+    projectId: string;
+  };
+}
+
```
sheensantoscapadngan, line 2672: IGNORE PLACEMENT OF COMMENT

If I set two common name attributes, it doesn't embed the second value. We shouldn't be allowing multiple values for some of these fields right? I think it's worth double-checking or finalizing that

image
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4051,13 +4125,22 @@ export type Event =
...
   | GetCertificateTemplateEstConfig
+  | CreateCertificateTemplate
+  | UpdateCertificateTemplate
+  | DeleteCertificateTemplate
+  | GetCertificateTemplate
+  | ListCertificateTemplates
+  | CreateCertificateProfile
+  | UpdateCertificateProfile
+  | DeleteCertificateProfile
+  | GetCertificateProfile
```
sheensantoscapadngan, line 4166: IGNORE PLACEMENT OF COMMENT

Maybe we could hide the certificate-related fields until AFTER a user selects a certificate profile? While testing, I keep forgetting to select a profile and then having to re-input the values because the profile selection overwrites some values

image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+              })
+              .optional()
+          }).array(),
+          totalCount: z.number()
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const { profiles, totalCount } = await server.services.certificateProfile.listProfiles({
```
sheensantoscapadngan, line 147: IGNORE PLACEMENT OF COMMENT

Even though there's no SAN value defined in the profile's template, I was able to define one in the issued certificate which shouldn't be the case right? 

image

image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+            enrollmentType: certificateProfile.enrollmentType
+          }
+        }
+      });
+
+      return { certificateProfile };
+    }
+  });
+
+  server.route({
```
sheensantoscapadngan, line 108: IGNORE PLACEMENT OF COMMENT

Do the other platforms have a friendly name property for the issued certificates as well? Previously we were looking into removing this field altogether which is why we don't show it anywhere in the UI now 
image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+      tags: [ApiDocsTags.PkiCertificateProfiles],
+      querystring: listCertificateProfilesSchema.extend({
+        includeMetrics: z.coerce.boolean().optional().default(false),
+        expiringDays: z.coerce.number().min(1).max(365).optional().default(7)
+      }),
+      response: {
+        200: z.object({
+          certificateProfiles: CertificateProfilesSchema.extend({
+            metrics: z
+              .object({
```
sheensantoscapadngan, line 131: IGNORE PLACEMENT OF COMMENT


Looks like the TTL value in the issue certificate flow doesn't get prefilled by the value in the template? 
image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+        includeMetrics: z.coerce.boolean().optional().default(false),
+        expiringDays: z.coerce.number().min(1).max(365).optional().default(7)
+      }),
+      response: {
+        200: z.object({
+          certificateProfile: CertificateProfilesSchema.extend({
+            certificateAuthority: z
+              .object({
+                id: z.string(),
+                projectId: z.string(),
```
sheensantoscapadngan, line 192: IGNORE PLACEMENT OF COMMENT

Trying to delete a certificate template which is attached to an existing certificate profile throws a generic error
image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+      const certificateProfile = await server.services.certificateProfile.createProfile({
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId,
+        projectId: req.body.projectId,
+        data: req.body
+      });
+
+      await server.services.auditLog.createAuditLog({
```
sheensantoscapadngan, line 90: IGNORE PLACEMENT OF COMMENT


It looks like the signature algorithm I inputted during certificate issuance didn't apply properly? Or am I missing something? 

image

image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+                revokedCertificates: z.number()
+              })
+              .optional()
+          })
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const certificateProfile = await server.services.certificateProfile.getProfileByIdWithConfigs({
```
sheensantoscapadngan, line 236: IGNORE PLACEMENT OF COMMENT

Clicking into a certificate manager project opens up the Subscriber page even though I don't have it because it's a new project
image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+        });
+        result = { ...certificateProfile, metrics };
+      }
+
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        projectId: certificateProfile.projectId,
+        event: {
+          type: EventType.GET_CERTIFICATE_PROFILE,
+          metadata: {
```
sheensantoscapadngan, line 263: IGNORE PLACEMENT OF COMMENT

The dropdown of certificate authority looks off when no CA exists in the project. Please check with the other dropdowns as well 

image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
-    certificateTemplateId: string;
-    isEnabled: boolean;
+    certificateProfileId: string;
+    name: string;
   };
 }
 
-interface GetAzureAdCsTemplatesEvent {
-  type: EventType.GET_AZURE_AD_TEMPLATES;
+interface DeleteCertificateProfile {
```
sheensantoscapadngan, line 2650: IGNORE PLACEMENT OF COMMENT

For those with legacy entities, shouldn't we also have a "Certificates" section in the legacy side so they can go through the issuance flow with their old templates?
image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
...
+            certificateProfileId: certificateProfile.id
+          }
+        }
+      });
+
+      return { certificateProfile: result };
+    }
+  });
```
sheensantoscapadngan, line 272: IGNORE PLACEMENT OF COMMENT

Mentioning here so we don't forget - let's mark the old APIs as "legacy" in the docs as part of this PR

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2512,89 +2521,154 @@ interface LoadProjectKmsBackupEvent {
...
   metadata: {
-    certificateTemplateId: string;
-    isEnabled: boolean;
+    certificateProfileId: string;
   };
 }
 
-interface GetCertificateTemplateEstConfig {
-  type: EventType.GET_CERTIFICATE_TEMPLATE_EST_CONFIG;
+interface ListCertificateProfiles {
```
sheensantoscapadngan, line 2666: IGNORE PLACEMENT OF COMMENT

I can't seem to find the profile ID for the sign certificate endpoint. We should have a copy button here right? 
image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4051,13 +4125,22 @@ export type Event =
...
+  | GetCertificateTemplate
+  | ListCertificateTemplates
+  | CreateCertificateProfile
+  | UpdateCertificateProfile
+  | DeleteCertificateProfile
+  | GetCertificateProfile
+  | ListCertificateProfiles
+  | IssueCertificateFromProfile
+  | SignCertificateFromProfile
+  | OrderCertificateFromProfile
```
sheensantoscapadngan, line 4170: IGNORE PLACEMENT OF COMMENT

I was able to issue a certificate through the sign-certificate endpoint with state and country values even when it's not in the subject attributes of the template

image

image

backend/src/server/routes/v1/certificate-profiles-router.ts
```diff
@@ -0,0 +1,393 @@
+import { z } from "zod";
+
+import { CertificateProfilesSchema } from "@app/db/schemas";
+import { EventType } from "@app/ee/services/audit-log/audit-log-types";
```
sheensantoscapadngan, line 5: IGNORE PLACEMENT OF COMMENT

There's something wrong with the issuing CA certificate response here 
image

backend/src/ee/services/audit-log/audit-log-types.ts
carlosmonastyrski: Let's don't change the old flow, there is a feature flag in place to keep the old flow allowed on these cases. It could be weird to support a new flow for a deprecated feature.

carlosmonastyrski:
@greptile review this PR

PR #4655, opened 2025-10-11
[ENG-3913] Add mysql pam

fangpenlin:
# Description 📣

ref: https://linear.app/infisical/issue/ENG-3913/add-support-for-mysql-in-pam-product
depends on https://github.com/Infisical/cli/pull/28

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

This feature depends on the CLI PR 28:

https://github.com/Infisical/cli/pull/28

Be sure to check out that locally as well. To test this feature:

1. Setup a `relay` server then a `gateway` like usual for PAM with the changes in CLI PR 28. Keep them up and running locally
2. Obviously, you need a MySQL server to test. Here's the docker compose file for the non-SSL version:

```yaml
services:
  db:
    image: mysql
    restart: always
    environment:
      MYSQL_ROOT_PASSWORD: example
    ports:
      - 3306:3306
```

And then the one with SSL version:

```yaml
services:
  db:
    image: mysql
    restart: always
    environment:
      MYSQL_ROOT_PASSWORD: example
    command:
      - --ssl-ca=/var/lib/mysql-certs/ca.pem
      - --ssl-cert=/var/lib/mysql-certs/server-cert.pem
      - --ssl-key=/var/lib/mysql-certs/server-key.pem
      - --require-secure-transport=ON
    ports:
      - 3306:3306
    volumes:
      - ./mysql-certs:/var/lib/mysql-certs
```

You can create the self-signed certs with the following cmds:

```
mkdir -p ./mysql-certs
cd ./mysql-certs

openssl genrsa -out ca-key.pem 2048
openssl req -x509 -new -nodes -key ca-key.pem -days 365 -out ca.pem -subj "/CN=MySQL_CA"

openssl genrsa -out server-key.pem 2048
openssl req -new -key server-key.pem -out server.csr -subj "/CN=localhost"
openssl x509 -req -in server.csr -CA ca.pem -CAkey ca-key.pem -CAcreateserial -out server-cert.pem -days 365
```

3. Create a mysql resource with `host=localhost`, `port=3306` and selecting the local dev gateway
4. Create a mysql account with `username=root` and `password=example` with the new mysql resource
5. Go the account page, click "access" for the new mysql account
6. Run the `infisical pam db access-account <session id> --duration 4h` cmd as provided in the text field with the CLI PR
7. You should be able to connect to the localhost with the provided port, run simple query like `SELECT 1` to see how it works

### DIFF
--- a/backend/src/ee/services/pam-resource/mysql/mysql-resource-schemas.ts
+++ b/backend/src/ee/services/pam-resource/mysql/mysql-resource-schemas.ts
@@ -59,9 +59,7 @@ export const MySQLAccountSchema = BasePamAccountSchema.extend({
 });
 
 export const CreateMySQLAccountSchema = BaseCreatePamAccountSchema.extend({
-  credentials: MySQLAccountCredentialsSchema,
-  // We don't support pwd rotation for mysql yet
-  rotationEnabled: z.boolean().default(false)
+  credentials: MySQLAccountCredentialsSchema
 });
 
 export const UpdateMySQLAccountSchema = BaseUpdatePamAccountSchema.extend({
--- a/frontend/src/pages/pam/PamAccountsPage/components/PamAccountForm/MySQLAccountForm.tsx
+++ b/frontend/src/pages/pam/PamAccountsPage/components/PamAccountForm/MySQLAccountForm.tsx
@@ -18,7 +18,10 @@ type Props = {
 };
 
 const formSchema = genericAccountFieldsSchema.extend({
-  credentials: BaseSqlAccountSchema
+  credentials: BaseSqlAccountSchema,
+  // We don't support rotation for now, just feed a false value to
+  // make the schema happy
+  rotationEnabled: z.boolean().default(false)
 });
 
 type FormData = z.infer<typeof formSchema>;


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/859e4779-30b6-4a3e-bd0e-b3fb64cfb807?refs=dd713fbe612efccc4223e2f50cbf9b4faf175ad0&source=prChecksComment) |

frontend/src/pages/pam/PamResourcesPage/components/PamResourceForm/shared/SqlResourceFields.tsx
```diff
@@ -110,7 +110,7 @@ export const SqlResourceFields = ({ setSelectedTabIndex, selectedTabIndex }: Pro
                 errorText={error?.message}
                 isError={Boolean(error?.message)}
                 className={sslEnabled ? "" : "opacity-50"}
-                label="SSL Certificate"
+                label="Trusted CA SSL Certificate"
```
fangpenlin, line 113: During development, I was confused by this field. I thought it might be SSL cert pining, but turns out to be CA. Sometimes people do things like cert pining. To make it more clear, I think maybe it makes more sense to call this "Trust CA cert" instead

backend/package.json
```diff
@@ -221,7 +221,7 @@
     "lodash.isequal": "^4.5.0",
     "mongodb": "^6.8.1",
     "ms": "^2.1.3",
-    "mysql2": "^3.9.8",
+    "mysql2": "^3.15.2",
```
sheensantoscapadngan: were you able to test dynamic secrets and secret rotation after this version bump?

backend/src/ee/services/pam-resource/shared/sql/sql-resource-factory.ts
```diff
@@ -15,30 +16,134 @@ const EXTERNAL_REQUEST_TIMEOUT = 10 * 1000;
...
+              //       2. modify relay to add a new endpoint for returning if the target host is healthy or not
+              //          (like being able to do an auth handshake regardless pass or not)
+              if (
+                connectOnly &&
+                (error.message === `password authentication failed for user "${TEST_CONNECTION_USERNAME}"` ||
+                  error.message.includes("no pg_hba.conf entry for host"))
+              ) {
+                return;
+              }
+            }
```
sheensantoscapadngan: curious, why is the condition checking for BadRequestError? `client.raw` wouldn't throw that right or am I missing something?

backend/src/ee/services/pam-resource/shared/sql/sql-resource-factory.ts
```diff
@@ -15,30 +16,134 @@ const EXTERNAL_REQUEST_TIMEOUT = 10 * 1000;
...
+              user: actualUsername, // Use provided username or fallback
+              password: actualPassword, // Use provided password or fallback
+              database: connectionDetails.database,
+              ssl: sslEnabled
+                ? {
+                    rejectUnauthorized: sslRejectUnauthorized,
+                    ca: sslCertificate
+                  }
+                : undefined
+            });
```
sheensantoscapadngan, line 154: we don't need to overwrite the checkServerIdentity callback here ?

frontend/src/pages/pam/PamResourcesPage/components/PamResourceForm/shared/SqlResourceFields.tsx
sheensantoscapadngan: good call!

backend/src/ee/services/pam-resource/shared/sql/sql-resource-factory.ts
sheensantoscapadngan: Yeah I think this broke the addition of new Postgres resources. Please test it out on your side 
image

backend/src/ee/services/pam-resource/shared/sql/sql-resource-factory.ts
fangpenlin: mysql2 unfortunately doesn't support providing a custom `checkServerIdentity`. They have the callback func hardwired here:

https://github.com/sidorares/node-mysql2/blob/2543272a2ada8d8a07f74582549d7dd3fe948e2d/lib/base/connection.js#L358-L362

there's no easy way to workaround this, other than forking mysql2 or monkey patch some of these functions. In these line, it seems like checking the and see if the certificate matches the host name

https://github.com/sidorares/node-mysql2/blob/2543272a2ada8d8a07f74582549d7dd3fe948e2d/lib/base/connection.js#L370-L382

similar to what we did with what we did with PostgreSQL's `checkServerIdentity`. But I am not sure if this is exactly like that. May need to generate a certificate that has a different common name to test it out.

Regardless, if the mysql2 built-in logic rejects the cert because of the different common name, we may need to find a way to workaround it. That would be another ticket I guess, otherwise the scope of this PR will be too big.

backend/package.json
fangpenlin: The version bump is mostly for new line change introduced in mysql2 lib, where the `TLSSocket` is exposed under the client:

https://github.com/sidorares/node-mysql2/blob/2543272a2ada8d8a07f74582549d7dd3fe948e2d/lib/base/connection.js#L397

The older version doesn't. That's just for what I hope to workaround the mysql2 lib doesn't support `checkServerIdentity`, by disabling SSL cert check and does it manually. Currently that part is not done nor fully tested yet. So maybe we can rollback this for now.

However, we should have more automatic tests cover the existing features, otherwise any new changes introduced could break the existing features. The testing costs add up quickly.

fangpenlin: okay, just revert the mysql2 version

backend/src/ee/services/pam-resource/shared/sql/sql-resource-factory.ts
fangpenlin: done. the psql PAM res type is fixed. will test again before you review it to ensure it works

fangpenlin:
There's a really strange error on CI:

```
/app/src/ee/services/pam-resource/shared/sql/sql-resource-factory.ts:3
import { DatabaseError } from "pg";
         ^

SyntaxError: The requested module 'pg' does not provide an export named 'DatabaseError'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:213:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:320:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:606:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)
  ```
  
I am pretty certain the `DatabaseError` exists in pg from long time ago:
  
https://github.com/brianc/node-postgres/blob/a84ebb3fe85edafb1b358965bdb5e3de58b12c5d/packages/pg/lib/index.js#L7

It also works on my machine. But somehow on CI, it just failed to find it. Will look into it tomorrow. Otherwise the PR is pretty much ready for review.

PR #4728, opened 2025-10-23
PKI: Add support for certificate auto-renewal and manual renewal workflows

carlosmonastyrski:
# Description 📣

Added support for both automatic and manual renewal of certificates issued from profiles that use API enrollment. This enables certificates under these profiles to be renewed either automatically based on configured settings or manually triggered by users when needed.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
```

---

### DIFF
--- a/backend/src/services/certificate-v3/certificate-v3-queue.ts
+++ b/backend/src/services/certificate-v3/certificate-v3-queue.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-await-in-loop */
 import { EventType, TAuditLogServiceFactory } from "@app/ee/services/audit-log/audit-log-types";
+import { getConfig } from "@app/lib/config/env";
 import { logger } from "@app/lib/logger";
 import { QueueJobs, QueueName, TQueueServiceFactory } from "@app/queue";
 
@@ -21,130 +22,142 @@ export const certificateV3QueueServiceFactory = ({
   certificateV3Service,
   auditLogService
 }: TCertificateV3QueueServiceFactoryDep) => {
-  queueService.start(QueueName.CertificateV3AutoRenewal, async (job) => {
-    if (job.name === QueueJobs.CertificateV3DailyAutoRenewal) {
-      logger.info(`${QueueJobs.CertificateV3DailyAutoRenewal}: queue task started`);
+  const appCfg = getConfig();
 
-      const { QUEUE_BATCH_SIZE } = CERTIFICATE_RENEWAL_CONFIG;
-      let offset = 0;
-      let hasMore = true;
-      let totalCertificatesFound = 0;
-      let totalCertificatesRenewed = 0;
-
-      while (hasMore) {
-        const certificates = await certificateDAL.findCertificatesEligibleForRenewal({
-          limit: QUEUE_BATCH_SIZE,
-          offset
-        });
-
-        if (certificates.length === 0) {
-          hasMore = false;
-          break;
-        }
-
-        totalCertificatesFound += certificates.length;
-        logger.info(
-          `${QueueJobs.CertificateV3DailyAutoRenewal}: found ${certificates.length} certificates eligible for renewal (batch ${Math.floor(offset / QUEUE_BATCH_SIZE) + 1}, total found so far: ${totalCertificatesFound})`
-        );
-
-        for (const certificate of certificates) {
-          try {
-            if (certificate.renewBeforeDays) {
-              const { MIN_RENEW_BEFORE_DAYS, MAX_RENEW_BEFORE_DAYS } = CERTIFICATE_RENEWAL_CONFIG;
-              if (
-                certificate.renewBeforeDays < MIN_RENEW_BEFORE_DAYS ||
-                certificate.renewBeforeDays > MAX_RENEW_BEFORE_DAYS
-              ) {
-                // eslint-disable-next-line no-continue
-                continue;
-              }
-            }
-
-            await certificateV3Service.renewCertificate({
-              actor: ActorType.PLATFORM,
-              actorId: "",
-              actorAuthMethod: null,
-              actorOrgId: "",
-              certificateId: certificate.id,
-              internal: true
-            });
-
-            totalCertificatesRenewed += 1;
-
-            await auditLogService.createAuditLog({
-              projectId: certificate.projectId,
-              actor: {
-                type: ActorType.PLATFORM,
-                metadata: {}
-              },
-              event: {
-                type: EventType.AUTOMATED_RENEW_CERTIFICATE,
-                metadata: {
-                  certificateId: certificate.id,
-                  commonName: certificate.commonName || "",
-                  profileId: certificate.profileId!,
-                  renewBeforeDays: certificate.renewBeforeDays?.toString() || "",
-                  profileName: certificate.profileName || ""
-                }
-              }
-            });
-          } catch (error) {
-            const errorMessage = error instanceof Error ? error.message : String(error);
-            logger.error(error, `Failed to renew certificate ${certificate.id}: ${errorMessage}`);
-            await auditLogService.createAuditLog({
-              projectId: certificate.projectId,
-              actor: {
-                type: ActorType.PLATFORM,
-                metadata: {}
-              },
-              event: {
-                type: EventType.AUTOMATED_RENEW_CERTIFICATE_FAILED,
-                metadata: {
-                  certificateId: certificate.id,
-                  commonName: certificate.commonName || "",
-                  profileId: certificate.profileId || "",
-                  renewBeforeDays: certificate.renewBeforeDays?.toString() || "",
-                  profileName: certificate.profileName || "",
-                  error: errorMessage
-                }
-              }
-            });
-          }
-        }
-
-        offset += QUEUE_BATCH_SIZE;
-      }
-
-      logger.info(
-        `${QueueJobs.CertificateV3DailyAutoRenewal}: queue task completed. Renewed ${totalCertificatesRenewed} certificates out of ${totalCertificatesFound}`
-      );
+  const init = async () => {
+    if (appCfg.isSecondaryInstance) {
+      return;
     }
-  });
-
-  const startDailyAutoRenewalJob = async () => {
-    const { DAILY_CRON_SCHEDULE, QUEUE_START_DELAY_MS } = CERTIFICATE_RENEWAL_CONFIG;
 
     await queueService.stopRepeatableJob(
       QueueName.CertificateV3AutoRenewal,
       QueueJobs.CertificateV3DailyAutoRenewal,
-      { pattern: DAILY_CRON_SCHEDULE, utc: true },
+      { pattern: CERTIFICATE_RENEWAL_CONFIG.DAILY_CRON_SCHEDULE, utc: true },
       QueueName.CertificateV3AutoRenewal
     );
 
-    await queueService.queue(QueueName.CertificateV3AutoRenewal, QueueJobs.CertificateV3DailyAutoRenewal, undefined, {
-      delay: QUEUE_START_DELAY_MS,
-      jobId: QueueName.CertificateV3AutoRenewal,
-      repeat: { pattern: DAILY_CRON_SCHEDULE, utc: true }
-    });
+    await queueService.startPg<QueueName.CertificateV3AutoRenewal>(
+      QueueJobs.CertificateV3DailyAutoRenewal,
+      async () => {
+        try {
+          logger.info(`${QueueJobs.CertificateV3DailyAutoRenewal}: queue task started`);
+
+          const { QUEUE_BATCH_SIZE } = CERTIFICATE_RENEWAL_CONFIG;
+          let offset = 0;
+          let hasMore = true;
+          let totalCertificatesFound = 0;
+          let totalCertificatesRenewed = 0;
+
+          while (hasMore) {
+            const certificates = await certificateDAL.findCertificatesEligibleForRenewal({
+              limit: QUEUE_BATCH_SIZE,
+              offset
+            });
+
+            if (certificates.length === 0) {
+              hasMore = false;
+              break;
+            }
+
+            totalCertificatesFound += certificates.length;
+            logger.info(
+              `${QueueJobs.CertificateV3DailyAutoRenewal}: found ${certificates.length} certificates eligible for renewal (batch ${Math.floor(offset / QUEUE_BATCH_SIZE) + 1}, total found so far: ${totalCertificatesFound})`
+            );
+
+            for (const certificate of certificates) {
+              try {
+                if (certificate.renewBeforeDays) {
+                  const { MIN_RENEW_BEFORE_DAYS, MAX_RENEW_BEFORE_DAYS } = CERTIFICATE_RENEWAL_CONFIG;
+                  if (
+                    certificate.renewBeforeDays < MIN_RENEW_BEFORE_DAYS ||
+                    certificate.renewBeforeDays > MAX_RENEW_BEFORE_DAYS
+                  ) {
+                    // eslint-disable-next-line no-continue
+                    continue;
+                  }
+                }
+
+                await certificateV3Service.renewCertificate({
+                  actor: ActorType.PLATFORM,
+                  actorId: "",
+                  actorAuthMethod: null,
+                  actorOrgId: "",
+                  certificateId: certificate.id,
+                  internal: true
+                });
+
+                totalCertificatesRenewed += 1;
+
+                await auditLogService.createAuditLog({
+                  projectId: certificate.projectId,
+                  actor: {
+                    type: ActorType.PLATFORM,
+                    metadata: {}
+                  },
+                  event: {
+                    type: EventType.AUTOMATED_RENEW_CERTIFICATE,
+                    metadata: {
+                      certificateId: certificate.id,
+                      commonName: certificate.commonName || "",
+                      profileId: certificate.profileId!,
+                      renewBeforeDays: certificate.renewBeforeDays?.toString() || "",
+                      profileName: certificate.profileName || ""
+                    }
+                  }
+                });
+              } catch (error) {
+                const errorMessage = error instanceof Error ? error.message : String(error);
+                logger.error(error, `Failed to renew certificate ${certificate.id}: ${errorMessage}`);
+                await auditLogService.createAuditLog({
+                  projectId: certificate.projectId,
+                  actor: {
+                    type: ActorType.PLATFORM,
+                    metadata: {}
+                  },
+                  event: {
+                    type: EventType.AUTOMATED_RENEW_CERTIFICATE_FAILED,
+                    metadata: {
+                      certificateId: certificate.id,
+                      commonName: certificate.commonName || "",
+                      profileId: certificate.profileId || "",
+                      renewBeforeDays: certificate.renewBeforeDays?.toString() || "",
+                      profileName: certificate.profileName || "",
+                      error: errorMessage
+                    }
+                  }
+                });
+              }
+            }
+
+            offset += QUEUE_BATCH_SIZE;
+          }
+
+          logger.info(
+            `${QueueJobs.CertificateV3DailyAutoRenewal}: queue task completed. Renewed ${totalCertificatesRenewed} certificates out of ${totalCertificatesFound}`
+          );
+        } catch (error) {
+          logger.error(error, `${QueueJobs.CertificateV3DailyAutoRenewal}: certificate renewal failed`);
+          throw error;
+        }
+      },
+      {
+        batchSize: 1,
+        workerCount: 1,
+        pollingIntervalSeconds: 60
+      }
+    );
+
+    await queueService.schedulePg(
+      QueueJobs.CertificateV3DailyAutoRenewal,
+      CERTIFICATE_RENEWAL_CONFIG.DAILY_CRON_SCHEDULE,
+      undefined,
+      { tz: "UTC" }
+    );
   };
 
-  queueService.listen(QueueName.CertificateV3AutoRenewal, "failed", (_, err) => {
-    logger.error(err, `${QueueName.CertificateV3AutoRenewal}: failed`);
-  });
-
   return {
-    startDailyAutoRenewalJob
+    init
   };
 };
 
-export type TCertificateV3QueueFactory = ReturnType<typeof certificateV3QueueServiceFactory>;
+export type TCertificateV3QueueServiceFactory = ReturnType<typeof certificateV3QueueServiceFactory>;
--- a/backend/src/server/routes/index.ts
+++ b/backend/src/server/routes/index.ts
@@ -2339,7 +2339,7 @@ export const registerRoutes = async (
   await dailyReminderQueueService.startSecretReminderMigrationJob();
   await dailyExpiringPkiItemAlert.startSendingAlerts();
   await pkiSubscriberQueue.startDailyAutoRenewalJob();
-  await certificateV3Queue.startDailyAutoRenewalJob();
+  await certificateV3Queue.init();
   await kmsService.startService(hsmStatus);
   await microsoftTeamsService.start();
   await dynamicSecretQueueService.init();


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/ca53444a-5a43-4655-bccb-7fcfe96acf58?refs=8961e5d83d1cecbe1d665646a3a86c91e9fc0e99&source=prChecksComment) |

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateRenewalModal.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+      createNotification({
+        text: "Certificate renewed successfully",
+        type: "success"
+      });
+
+      handlePopUpToggle("renewCertificate", false);
+    } catch (err) {
+      console.error(err);
+    }
```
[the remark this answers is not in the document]
carlosmonastyrski: We already show the error message from the query error

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  if (await knex.schema.hasColumn(TableName.PkiApiEnrollmentConfig, "autoRenewDays")) {
+    await knex.schema.alterTable(TableName.PkiApiEnrollmentConfig, (t) => {
+      t.renameColumn("autoRenewDays", "renewBeforeDays");
+    });
+  }
```
sheensantoscapadngan, line 10: renames are generally avoid but this should be ok because nobody's using this yet I assume xD

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  if (!(await knex.schema.hasColumn(TableName.Certificate, "renewBeforeDays"))) {
+    await knex.schema.alterTable(TableName.Certificate, (t) => {
+      t.integer("renewBeforeDays").nullable();
+      t.uuid("renewedFromId").nullable();
+      t.uuid("renewedById").nullable();
```
sheensantoscapadngan: 1. what's the difference between these two fields? 

2. can we make the naming here more explicit? maybe `renewedFromCertificateId` and `renewedByCertificateId`?

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4009,6 +4044,21 @@ interface PamResourceDeleteEvent {
...
+    renewBeforeDays: string;
+  };
+}
+
+interface DisableCertificateRenewalConfigEvent {
+  type: EventType.DISABLE_CERTIFICATE_RENEWAL_CONFIG;
+  metadata: {
+    certificateId: string;
+  };
+}
```
sheensantoscapadngan, line 4065: I would also add the common name here so it's all clear from the audit logs itself

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -343,4 +343,142 @@ export const registerCertificatesRouter = async (server: FastifyZodProvider) =>
...
+    schema: {
+      hide: false,
+      tags: [ApiDocsTags.PkiCertificates],
+      params: z.object({
+        certificateId: z.string().uuid()
+      }),
+      body: z
+        .object({
+          renewBeforeDays: z.number().int().min(1).max(30).optional(),
+          disableAutoRenewal: z.boolean().optional()
```
sheensantoscapadngan: For ease of understanding, I think we should be avoiding negatives for the API properties? Maybe this could be "enableAutoRenewal"?

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -343,4 +343,142 @@ export const registerCertificatesRouter = async (server: FastifyZodProvider) =>
...
+        });
+
+        await server.services.auditLog.createAuditLog({
+          ...req.auditLogInfo,
+          projectId: data.projectId,
+          event: {
+            type: EventType.DISABLE_CERTIFICATE_RENEWAL_CONFIG,
+            metadata: {
+              certificateId: req.params.certificateId
+            }
```
sheensantoscapadngan, line 443: same remark here and along with the other relevant implementations

backend/src/services/certificate-authority/internal/internal-certificate-authority-service.ts
```diff
@@ -1210,19 +1211,21 @@ export const internalCertificateAuthorityServiceFactory = ({
...
 
-    ForbiddenError.from(permission).throwUnlessCan(
-      ProjectPermissionCertificateActions.Create,
-      ProjectPermissionSub.Certificates
-    );
+      ForbiddenError.from(permission).throwUnlessCan(
+        ProjectPermissionCertificateActions.Create,
+        ProjectPermissionSub.Certificates
+      );
+    }
```
sheensantoscapadngan: Just realized that we're not using the CertificateProfile Issue certificate permission check here for when `isFromProfile` is true

sheensantoscapadngan: I guess the same applies to the sign certificate method as well?

backend/src/services/certificate-common/certificate-utils.ts
```diff
@@ -196,3 +200,74 @@ export const convertExtendedKeyUsageArrayToLegacy = (
...
+    if (errorMessage.includes("not eligible for renewal")) {
+      return CERTIFICATE_RENEWAL_ERROR_MESSAGES[CertificateRenewalErrorType.NOT_ELIGIBLE];
+    }
+    if (errorMessage.includes("Requested validity period exceeds maximum allowed duration")) {
+      return CERTIFICATE_RENEWAL_ERROR_MESSAGES[CertificateRenewalErrorType.VALIDITY_EXCEEDS_MAXIMUM];
+    }
+    if (errorMessage.includes("not allowed by template policy")) {
+      return CERTIFICATE_RENEWAL_ERROR_MESSAGES[CertificateRenewalErrorType.NOT_ALLOWED_BY_TEMPLATE];
+    }
+  }
```
sheensantoscapadngan: hmm why don't we just throw an error with these message directly from the source of the error itself instead of parsing the message after the fact? Basing it off of the message content is very flaky..

backend/src/services/certificate-v3/certificate-v3-queue.ts
```diff
@@ -0,0 +1,164 @@
...
+                metadata: {}
+              },
+              event: {
+                type: EventType.AUTOMATED_RENEW_CERTIFICATE,
+                metadata: {
+                  certificateId: certificate.id,
+                  commonName: certificate.commonName || "",
+                  profileId: certificate.profileId!,
+                  renewBeforeDays: certificate.renewBeforeDays?.toString() || ""
+                }
```
sheensantoscapadngan: can we also log the profile name here please

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+        actorAuthMethod,
+        actorOrgId,
+        actionProjectType: ActionProjectType.CertificateManager
+      });
+
+      ForbiddenError.from(permission).throwUnlessCan(
+        ProjectPermissionCertificateProfileActions.IssueCert,
+        ProjectPermissionSub.CertificateProfiles
+      );
+    }
```
sheensantoscapadngan: it would be best if we did this right after we fetched the profile so that there's not much details we're exposing in terms of errors

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+      actorId,
+      actorAuthMethod,
+      actorOrgId,
+      templateId: profile.certificateTemplateId,
+      internal
+    });
+
+    if (!template) {
+      throw new NotFoundError({ message: "Certificate template not found for this profile" });
+    }
```
sheensantoscapadngan: if the user has appropriate access to the cert profile, we don't really have to check that they have access to the underlying template right?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+        ? originalCert.altNames.split(",").map((san) => {
+            const trimmed = san.trim();
+            const isIp =
+              trimmed.length <= 45 &&
+              (new RE2("^\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}$").test(trimmed) ||
+                new RE2("^([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$").test(trimmed));
+            return {
+              type: isIp ? CertSubjectAlternativeNameType.IP_ADDRESS : CertSubjectAlternativeNameType.DNS_NAME,
+              value: trimmed
+            };
```
sheensantoscapadngan: how about email and the other cases ?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+      certificateRequest
+    );
+
+    if (!validationResult.isValid) {
+      await certificateDAL.updateById(originalCert.id, {
+        renewalError: `Template validation failed: ${validationResult.errors.join(", ")}`
+      });
+
+      throw new BadRequestError({
+        message: `Certificate renewal failed because requested validity period exceeds maximum allowed duration by the profile template: ${validationResult.errors.join(", ")}`
```
sheensantoscapadngan: is this always the case? 

"because requested validity period exceeds maximum allowed duration by the profile template"

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+        extendedKeyUsages: parseExtendedKeyUsages(originalCert.extendedKeyUsages),
+        signatureAlgorithm: originalSignatureAlgorithm,
+        keyAlgorithm: originalKeyAlgorithm,
+        isFromProfile: true,
+        actor,
+        actorId,
+        actorAuthMethod,
+        actorOrgId,
+        internal
+      });
```
sheensantoscapadngan: shouldn't this always be internal given that we're already doing a check above?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+    await certificateDAL.updateById(newCert.id, {
+      profileId: originalCert.profileId,
+      renewBeforeDays: finalRenewBeforeDays,
+      renewedFromId: originalCert.id
+    });
+
+    await certificateDAL.updateById(originalCert.id, {
+      renewedById: newCert.id,
+      renewalError: null
+    });
```
sheensantoscapadngan: shouldn't we be doing this inside a transaction or something?

I think there will be data consistency issues when issuance succeeds but something fails before these update new certificate requests are made?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+
+    const certificateString = extractCertificateFromBuffer(certificate as unknown as Buffer);
+    const certificateChainString = extractCertificateFromBuffer(certificateChain as unknown as Buffer);
```
sheensantoscapadngan: isn't the certificate and certificateChain variables already strings in PEM format?
image

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
+        message: "Certificate is not eligible for auto-renewal: certificate was not issued from a profile"
+      });
+    }
+
+    const profile = await certificateProfileDAL.findByIdWithConfigs(certificate.profileId);
+    if (!profile) {
+      throw new NotFoundError({ message: "Certificate profile not found" });
+    }
+
+    if (profile.enrollmentType !== "api") {
```
sheensantoscapadngan: minor - let's use an ENUM here

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -479,9 +637,342 @@ export const certificateV3ServiceFactory = ({
...
```
sheensantoscapadngan: same here

backend/src/services/certificate/certificate-dal.ts
```diff
@@ -114,12 +114,55 @@ export const certificateDALFactory = (db: TDbClient) => {
...
+    limit: number;
+    offset: number;
+  }): Promise<TCertificates[]> => {
+    try {
+      const now = new Date();
+      const endOfDay = new Date(now.getFullYear(), now.getMonth(), now.getDate(), 23, 59, 59, 999);
+
+      const certs = (await db
+        .replicaNode()(TableName.Certificate)
+        .select(`${TableName.Certificate}.*`)
```
sheensantoscapadngan: minor - let's use selectAllTableCols?

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+      t.string("keyAlgorithm").nullable();
+      t.string("signatureAlgorithm").nullable();
+      t.foreign("renewedFromId").references("id").inTable(TableName.Certificate).onDelete("SET NULL");
+      t.foreign("renewedById").references("id").inTable(TableName.Certificate).onDelete("SET NULL");
+      t.index("renewedFromId");
+      t.index("renewedById");
+      t.index("renewBeforeDays");
+    });
+  }
+}
```
sheensantoscapadngan, line 27: IGNORE PLACEMENT OF COMMENT

I just noticed that months here is being interpreted as "minute" when I try to issue a certificate...

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export async function down(knex: Knex): Promise<void> {
+  if (await knex.schema.hasColumn(TableName.Certificate, "renewBeforeDays")) {
+    await knex.schema.alterTable(TableName.Certificate, (t) => {
+      t.dropForeign(["renewedFromId"]);
+      t.dropForeign(["renewedById"]);
+      t.dropIndex("renewedFromId");
+      t.dropIndex("renewedById");
+      t.dropIndex("renewBeforeDays");
+      t.dropColumn("renewBeforeDays");
```
sheensantoscapadngan, line 37: IGNORE PLACEMENT OF COMMENT

Why does clicking "View Details" here do nothing? 
image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
sheensantoscapadngan: Here's the default view when I open the "Issue certificate" modal

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+      t.index("renewedById");
+      t.index("renewBeforeDays");
+    });
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
+  if (await knex.schema.hasColumn(TableName.Certificate, "renewBeforeDays")) {
+    await knex.schema.alterTable(TableName.Certificate, (t) => {
+      t.dropForeign(["renewedFromId"]);
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

Certificate expires in 20d but trying to set renewal config to be 20d shows nothing in the UI... so it's not really clear what went wrong

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+      t.dropForeign(["renewedById"]);
+      t.dropIndex("renewedFromId");
+      t.dropIndex("renewedById");
+      t.dropIndex("renewBeforeDays");
+      t.dropColumn("renewBeforeDays");
+      t.dropColumn("renewedFromId");
+      t.dropColumn("renewedById");
+      t.dropColumn("renewalError");
+      t.dropColumn("keyAlgorithm");
```
sheensantoscapadngan, line 41: IGNORE PLACEMENT OF COMMENT

I was able to manually renew a single certificate multiple times... is this expected? 
image


ohhh.. it turns out to be a different certificate every time

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
carlosmonastyrski: Those are the fields to track which certificate is renewing that certificate and also if it's renewing another one. Great suggestion on the field names 👌

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
sheensantoscapadngan: I guess the main confusion here is that i'm seeing the certificate with "Renews in 16d" as the second entry even though it's supposed to be first one/latest? 

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
```
sheensantoscapadngan, line 4: IGNORE PLACEMENT OF COMMENT

I think "Renewed" should somehow have a different color vs "Renews in 16d"? At the moment it doesn't really register to me that there's a significant difference between the two... Maybe "Renewd" should be more of a neutral color? Like gray? 



image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2458,6 +2463,27 @@ interface AutomatedRenewPkiSubscriberCert {
   };
 }
 
+interface AutomatedRenewCertificate {
+  type: EventType.AUTOMATED_RENEW_CERTIFICATE;
```
sheensantoscapadngan, line 2467: IGNORE PLACEMENT OF COMMENT

For ineligible ones, I think we should show a status and maybe explain why it's ineligible? Just to keep it consistent 

image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2458,6 +2463,27 @@ interface AutomatedRenewPkiSubscriberCert {
...
+  metadata: {
+    certificateId: string;
+    commonName: string;
+    profileId: string;
+    renewBeforeDays: string;
+  };
+}
+
```
sheensantoscapadngan, line 2476: IGNORE PLACEMENT OF COMMENT

I think it might be more intuitive if the column header is "RENEWAL STATUS"? 

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+      t.dropColumn("renewBeforeDays");
+      t.dropColumn("renewedFromId");
+      t.dropColumn("renewedById");
+      t.dropColumn("renewalError");
+      t.dropColumn("keyAlgorithm");
+      t.dropColumn("signatureAlgorithm");
+    });
+  }
+
+  if (await knex.schema.hasColumn(TableName.PkiApiEnrollmentConfig, "renewBeforeDays")) {
```
sheensantoscapadngan, line 46: IGNORE PLACEMENT OF COMMENT

Not sure if this was introduced in this PR but the signature algorithm list and key algorithm list is empty when I open the certificate issuance modal from this certificate inventory page

image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2720,6 +2746,15 @@ interface OrderCertificateFromProfile {
...
 
+interface RenewCertificate {
+  type: EventType.RENEW_CERTIFICATE;
+  metadata: {
+    originalCertificateId: string;
+    newCertificateId: string;
+    profileName: string;
+  };
+}
+
```
sheensantoscapadngan, line 2760: IGNORE PLACEMENT OF COMMENT

Upon enabling auto-renewal again... it defaults to "20" for me but I can't actually save the configuration... should we be defaulting it to whatever's defined in the linked profile instead? 

image

image

backend/src/server/routes/v3/certificates-router.ts
```diff
@@ -343,4 +343,142 @@ export const registerCertificatesRouter = async (server: FastifyZodProvider) =>
...
+        200: z.object({
+          certificate: z.string().trim(),
+          issuingCaCertificate: z.string().trim(),
+          certificateChain: z.string().trim(),
+          privateKey: z.string().trim().optional(),
+          serialNumber: z.string().trim(),
+          certificateId: z.string()
+        })
+      }
+    },
```
sheensantoscapadngan, line 368: IGNORE PLACEMENT OF COMMENT

I think this might be better termed as "Auto-renew days before expiry"... I find it more intuitive? 

image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -340,6 +340,8 @@ export enum EventType {
   ISSUE_PKI_SUBSCRIBER_CERT = "issue-pki-subscriber-cert",
   SIGN_PKI_SUBSCRIBER_CERT = "sign-pki-subscriber-cert",
   AUTOMATED_RENEW_SUBSCRIBER_CERT = "automated-renew-subscriber-cert",
+  AUTOMATED_RENEW_CERTIFICATE = "automated-renew-certificate",
```
sheensantoscapadngan, line 343: IGNORE PLACEMENT OF COMMENT

Changing the allowed signature algorithm and allowed key algorithm did NOT trigger any sort of validation error upon renewal. So I went from SHA256-RSA to SHA512-RSA for the signature algorithm and RSA-2048 to RSA-4096 for the allowed key algorithm

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
carlosmonastyrski: Fixed on another PR: https://github.com/Infisical/infisical/pull/4740

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
carlosmonastyrski: It's an old button from the previous flow, showing the template information and other fields. For this versions we will move to a certificate details view but it'll be handled in a separate PR

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
+import { Knex } from "knex";
```
sheensantoscapadngan, line 1: IGNORE PLACEMENT OF COMMENT

I think this should just be "Renewal" here instead of "Auto-Renewal", primarily because we have renamed/repurposed the column

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+      t.text("renewalError").nullable();
+      t.string("keyAlgorithm").nullable();
+      t.string("signatureAlgorithm").nullable();
+      t.foreign("renewedFromCertificateId").references("id").inTable(TableName.Certificate).onDelete("SET NULL");
+      t.foreign("renewedByCertificateId").references("id").inTable(TableName.Certificate).onDelete("SET NULL");
+      t.index("renewedFromCertificateId");
+      t.index("renewedByCertificateId");
+      t.index("renewBeforeDays");
+    });
+  }
```
sheensantoscapadngan, line 26: IGNORE PLACEMENT OF COMMENT

A certificate that I issued through a certificate profile with no auto-renewal config shows up with renewal status of "Renews in 334d" for some reason? 

image

image

backend/src/db/migrations/20251021112356_add-certificate-auto-renewal.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export async function down(knex: Knex): Promise<void> {
+  if (await knex.schema.hasColumn(TableName.Certificate, "renewBeforeDays")) {
+    await knex.schema.alterTable(TableName.Certificate, (t) => {
+      t.dropForeign(["renewedFromCertificateId"]);
```
sheensantoscapadngan, line 32: IGNORE PLACEMENT OF COMMENT

hmmm If I delete the LATEST certificate, I am able to renew again the previous certificate which shouldn't be the case right? Because a certificate should only ever be renewed once? 
image

After deletion of most recently renewed certificate:
image

PR #4737, opened 2025-10-23
fix: validate secret names before import

victorvhs017:
# Description 📣

Issue: https://linear.app/infisical/issue/ENG-3966/fail-the-import-for-secret-syncs-if-a-secret-name-violates-constraints

Validate the secret names with the schema before import to prevent invalid characters in the name.

Error message:

image

## Type ✨

- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

- Try to import secrets with "/" in the name, and you should see an error in the import.

---

### DIFF
--- a/backend/src/services/secret-sync/secret-sync-queue.ts
+++ b/backend/src/services/secret-sync/secret-sync-queue.ts
@@ -421,9 +421,10 @@ export const secretSyncQueueFactory = ({
     }
 
     if (invalidNameCount > 0) {
-      throw new Error(
-        `Found ${invalidNameCount} invalid secret name${invalidNameCount === 1 ? "" : "s"}. ${errorMessage}`
-      );
+      throw new SecretSyncError({
+        message: `Found ${invalidNameCount} invalid secret name${invalidNameCount === 1 ? "" : "s"}. ${errorMessage}`,
+        shouldRetry: false
+      });
     }
 
     const importedSecretMap: TSecretMap = {};


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/8ba633b6-156e-49af-9ffe-713046f7a6c5?refs=1cff2622712d605148d48215ee33c79da7d8a038&source=prChecksComment) |

victorvhs017:
@greptileai

backend/src/services/secret-sync/secret-sync-queue.ts
```diff
@@ -408,6 +409,23 @@ export const secretSyncQueueFactory = ({
...
+    for (const [key] of Object.entries(importedSecrets)) {
+      const result = SecretNameSchema.safeParse(key);
+      if (!result.success) {
+        invalidNameCount += 1;
+        if (errorMessage === "") errorMessage = result.error.issues[0]?.message;
+      }
+    }
+
+    if (invalidNameCount > 0) {
+      throw new Error(
```
scott-ray-wilson: let's throw a `SecretSyncError` here with `shouldRetry` set to `false`; explicit imports fail on first attempt but if you select import during creation it runs inside the sync job and will retry 5 times - using this error tells it not to retry

backend/src/services/secret-sync/secret-sync-queue.ts
```diff
@@ -408,6 +409,23 @@ export const secretSyncQueueFactory = ({
...
     if (!Object.keys(importedSecrets).length) return {};
 
+    let invalidNameCount = 0;
+    let errorMessage = "";
+
+    for (const [key] of Object.entries(importedSecrets)) {
+      const result = SecretNameSchema.safeParse(key);
+      if (!result.success) {
+        invalidNameCount += 1;
+        if (errorMessage === "") errorMessage = result.error.issues[0]?.message;
```
scott-ray-wilson, line 419: thoughts on storing the failing keys and concatenating them to the message at the end? That way users know the problematic keys

victorvhs017: Yeah, I thought about this, but it would be clunky with too many keys in that small tooltip space, right? It was my first solution, then I moved to the number + what's the problem, so they could easily search for the characters.

victorvhs017: I just tried a solution in the middle (print the first 5 secrets and add "..." at the end if more), but the tooltip starts to glitch if there's too much text on it 😅

backend/src/services/secret-sync/secret-sync-queue.ts
victorvhs017: Good to know, thank you!

backend/src/services/secret-sync/secret-sync-queue.ts
scott-ray-wilson: true they can query on their end; this was encountered internally anyway so we can see if anyone requests it

scott-ray-wilson (approved):
Confirmed it catches invalid secret keys on import, nice work!

PR #4747, opened 2025-10-24
[ENG-3941] Implement ACME enrollment type for PKI

fangpenlin:
# Description 📣

Implement ACME enrollment type for PKI
ref: https://linear.app/infisical/issue/ENG-3941/infisical-pki-acme-enrollment-method

## Type ✨

- [ ] Bug fix
- [x] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

Since this feature is not production ready yet, but the scope is already very huge, we would like to get it reviewed and merge while the size is manageable. The PKI ACME feature is behind a feature flag `ACME_FEATURE_ENABLED`. To test this PR, you need to first enable it in your `.env` file:

```env
ACME_FEATURE_ENABLED=true
```

We also have some special treatment for HTTP-01 challenge validation for testing purpose, it allows you to map a hostname to a new target (like a localhost with a port) for local testing purpose. Therefore, you can add:

```env
# enable the dev mode for ACME
ACME_DEVELOPMENT_MODE=true
# for HTTP-01 challenge goes to `localhost`, let's remap it to your local IP address
ACME_DEVELOPMENT_HTTP01_CHALLENGE_HOST_OVERRIDES='{"localhost": "192.168.123.456:8087"}'
```

And to test it, I added some BDD tests to ensure all the happy paths work. The BDD tests are in `backend/bdd/`. They are written in Python because the [certbot](https://github.com/certbot/certbot) is written in Python and its [ACME client lib](https://github.com/certbot/certbot/tree/main/acme) is in Python as well. We may want to rewrite the BDD tests with TypeScript or other lang later, but in the meantime, it's the fastest way to implement ACME by using the right tool. To run BDD tests, you first need to install the deps with [uv](https://github.com/astral-sh/uv)
  

```bash
cd backend/bdd
uv install
```

Then, you need to copy the `backend/bdd/.env.example` to `backend/bdd/.env` with the needed value modified from your local Infisical instance. You will need:

- `INFISICAL_API_URL` - point to your local Infisical
- `INFISICAL_TOKEN` - JWT token with admin permission for the cert projects
- `PROJECT_ID` - the PKI project id for running the test under
- `CERT_CA_ID` - ID for an internal cert CA
- `CERT_TEMPLATE_ID` - ID for a cert template that allows `localhost` as the CN
- `PROFILE_ID` - ID for a ACME profile
- `EAB_SECRET` - EAB secret for accessing the ACME profile

With that, you can then run

```bash
uv run behave ./features/pki/acme/
```

Then you should see it simulates ACME client / server interactions and they shall pass. The above steps should have been integrated into CI. But for now, this is just a temporary solution.

An alternative way to perform the test is to run `certbot` command. First, have certbot installed. Then run the following script with the placeholder replaced with actual ACME values:

certbot-demo.sh:
```bash
#!/bin/bash

set -euo pipefail  # Exit on error, undefined var, pipe failure
set -x            # Echo all commands as they run

# Create a temporary directory for all certbot data
TEMP_DIR=$(mktemp -d)
echo "Using temporary directory: $TEMP_DIR"

mkdir -p ./web-root

# Generate 2048-bit RSA private key
openssl genrsa -out "$TEMP_DIR/localhost.key" 2048

# Generate CSR for localhost
openssl req -new -key "$TEMP_DIR/localhost.key" -out "$TEMP_DIR/localhost.csr" \
    -subj "/CN=localhost"

# Run certbot with full isolation in $TEMP_DIR
certbot certonly \
    --config-dir "$TEMP_DIR" \
    --work-dir "$TEMP_DIR/work" \
    --logs-dir "$TEMP_DIR/logs" \
    --server "<ACME_DIRECTORY_URL>" \
    --eab-kid "<ACME_EAB_KID>" \
    --eab-hmac-key "<ACME_EAB_SECRET>" \
    --no-eff-email \
    --email fangpen@infisical.com \
    --agree-tos \
    --preferred-challenges http \
    --http-01-port 8087 \
    --webroot-path ./web-root \
    --config-dir ./certbot/config \
    --work-dir ./certbot/work \
    --logs-dir ./certbot/logs \
    -d localhost \
    --csr "$TEMP_DIR/localhost.csr" \
    --cert-name localhost

# Turn off command echoing after certbot
set +x

echo
echo "Certificate issuance complete!"
echo "All files are in: $TEMP_DIR"
echo
echo "Certificate files:"
ls -1 "$TEMP_DIR"/*.pem 2>/dev/null || echo "  (No .pem files found — check logs)"
echo
echo "To clean up: rm -rf '$TEMP_DIR'"
```

You can use the "Reval EAB Secret" button for the ACME profile you've created:

image

Then you should see something like this, just copy paste the values into the script:

image

Then run the script:

```bash
chmod +x certbot-demo.sh
./certbot-demo.sh
```

It will ask you which way to validate, choose 6 to make it run a localhost server for you (port we are configure to use 8087 in the certbot arg) 

That's it, you should see it run throw the flow and get cert issued with ACME protocol

### DIFF
--- a/frontend/src/pages/cert-manager/PoliciesPage/components/CertificateProfilesTab/ProfileRow.tsx
+++ b/frontend/src/pages/cert-manager/PoliciesPage/components/CertificateProfilesTab/ProfileRow.tsx
@@ -140,7 +140,7 @@ export const ProfileRow = ({
           </DropdownMenuTrigger>
           <DropdownMenuContent align="start" className="p-1">
             <DropdownMenuItem
-              icon={<FontAwesomeIcon icon={isIdCopied ? faCheck : faCopy} />}
+              icon={<FontAwesomeIcon icon={isIdCopied ? faCheck : faCopy} className="w-3" />}
               onClick={() => handleCopyId()}
             >
               Copy Profile ID
@@ -151,7 +151,7 @@ export const ProfileRow = ({
                   e.stopPropagation();
                   onEditProfile(profile);
                 }}
-                icon={<FontAwesomeIcon icon={faEdit} />}
+                icon={<FontAwesomeIcon icon={faEdit} className="w-3" />}
               >
                 Edit Profile
               </DropdownMenuItem>
@@ -162,7 +162,7 @@ export const ProfileRow = ({
                   e.stopPropagation();
                   onRevealProfileAcmeEabSecret(profile);
                 }}
-                icon={<FontAwesomeIcon icon={faEye} />}
+                icon={<FontAwesomeIcon icon={faEye} className="w-3" />}
               >
                 Reveal ACME EAB
               </DropdownMenuItem>
@@ -173,7 +173,7 @@ export const ProfileRow = ({
                   e.stopPropagation();
                   handlePopUpToggle("issueCertificate");
                 }}
-                icon={<FontAwesomeIcon icon={faPlus} />}
+                icon={<FontAwesomeIcon icon={faPlus} className="w-3" />}
               >
                 Issue Certificate
               </DropdownMenuItem>
@@ -184,7 +184,7 @@ export const ProfileRow = ({
                   e.stopPropagation();
                   onDeleteProfile(profile);
                 }}
-                icon={<FontAwesomeIcon icon={faTrash} />}
+                icon={<FontAwesomeIcon icon={faTrash} className="w-3" />}
               >
                 Delete Profile
               </DropdownMenuItem>


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/cc35c662-d216-4082-bf22-adcfbdf57062?refs=91c85b33d51c89359dd99d33a9d782487f2f82b1&source=prChecksComment) |

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -107,6 +107,13 @@ const validateProfileAndPermissions = async (
     });
   }
 
+  // XXX: NOT SURE IF THIS IS SECURE TO BY PASS THE PERMISSION CHECK FOR ACME ACCOUNTS
+  //      may need to consider this carefully
+  // TODO: check actor/profile ownership as well
+  if (actor === ActorType.ACME_ACCOUNT && requiredEnrollmentType === EnrollmentType.ACME) {
+    return profile;
+  }
```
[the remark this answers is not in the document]
fangpenlin: before calling the sign cert, PKI acme service actually already checked. but just in case if the same func got used, I guess let's check it again tho

backend/src/ee/services/pki-acme/pki-acme-challenge-service.ts
```diff
@@ -0,0 +1,120 @@
...
+        );
+      }
+      const challengeUrl = new URL(`/.well-known/acme-challenge/${challenge.auth.token}`, `http://${host}`);
+      logger.info({ challengeUrl }, "Performing ACME HTTP-01 challenge validation");
+      try {
+        // Notice: well, we are in a transaction, ideally we should not hold transaction and perform
+        //         a long running operation for long time. But assuming we are not performing a tons of
+        //         challenge validation at the same time, it should be fine.
+        // TODO: bound it with timeout of the fetch request
+        const challengeResponse = await fetch(challengeUrl);
```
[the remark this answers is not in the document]
fangpenlin: The DNS name are all regex validated while inputting into the system. Private IP is now blocked

backend/src/ee/services/pki-acme/pki-acme-challenge-service.ts
```diff
@@ -0,0 +1,133 @@
...
+      try {
+        // TODO: read config from the profile to get the timeout instead
+        const timeoutMs = 10 * 1000; // 10 seconds
+        // Notice: well, we are in a transaction, ideally we should not hold transaction and perform
+        //         a long running operation for long time. But assuming we are not performing a tons of
+        //         challenge validation at the same time, it should be fine.
+        const challengeResponse = await fetch(challengeUrl, { signal: AbortSignal.timeout(timeoutMs) });
+        if (challengeResponse.status !== 200) {
+          throw new BadRequestError({ message: "ACME challenge response is not 200" });
+        }
```
sheensantoscapadngan, line 79: before we release this for actual use, we'd have to rethink this because running this potentially long process will hold up the connections available in the client pool causing cascading effects to the rest of the system

backend/src/ee/services/pki-acme/pki-acme-errors.ts
```diff
@@ -0,0 +1,574 @@
...
+// RFC 8555 Section 6.7 - Error Types
+export enum AcmeErrorType {
+  AccountDoesNotExist = "accountDoesNotExist",
+  AlreadyRevoked = "alreadyRevoked",
+  BadCsr = "badCSR",
+  BadNonce = "badNonce",
+  BadPublicKey = "badPublicKey",
+  BadRevocationReason = "badRevocationReason",
+  BadSignatureAlgorithm = "badSignatureAlgorithm",
+  CAA = "CAA",
```
sheensantoscapadngan: does the casing matter here?
image

backend/src/ee/services/pki-acme/pki-acme-errors.ts
```diff
@@ -0,0 +1,574 @@
...
+  Compound = "compound",
+  Connection = "connection",
+  DNS = "DNS",
```
sheensantoscapadngan: same question here

backend/src/ee/services/pki-acme/pki-acme-errors.ts
```diff
@@ -0,0 +1,574 @@
...
+  ExternalAccountRequired = "externalAccountRequired",
+  IncorrectResponse = "incorrectResponse",
+  IncorrectContact = "incorrectContact",
```
sheensantoscapadngan: does the mismatch here matter?
image

backend/src/ee/services/pki-acme/pki-acme-schemas.ts
```diff
@@ -0,0 +1,170 @@
...
+});
+
+// New Order payload schema
+export const CreateAcmeOrderBodySchema = z.object({
+  identifiers: z.array(
+    z.object({
+      type: z.enum(Object.values(AcmeIdentifierType) as [string, ...string[]]),
+      value: z
+        .string()
+        .regex(/^(?!-)[A-Za-z0-9-]{1,63}(?<!-)(\.(?!-)[A-Za-z0-9-]{1,63}(?<!-))*$/, "Invalid DNS identifier")
```
sheensantoscapadngan: use re2 here to prevent regex attacks... can you please check the other usages in this PR for that?

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -0,0 +1,837 @@
...
+  }: {
+    profileId: string;
+    accountId: string;
+    payload: TCreateAcmeOrderPayload;
+  }): Promise<TAcmeResponse<TAcmeOrderResource>> => {
+    // TODO: check and see if we have existing orders for this account that meet the criteria
+    //       if we do, return the existing order
+    // TODO: check the identifiers and see if are they even allowed for this profile.
+    //       if not, we may be able to reject it early with an unsupportedIdentifier error.
+
```
sheensantoscapadngan, line 482: Do we not want to demonstrate how the certificate profile handles validation? I think this is important because someone would be able to bypass the policies? Or do we do that check in the final steps?

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -0,0 +1,837 @@
...
+      status: order.status,
+      expires: order.expiresAt.toISOString(),
+      notBefore: order.notBefore?.toISOString(),
+      notAfter: order.notAfter?.toISOString(),
+      identifiers: order.authorizations.map((auth) => ({
+        type: auth.identifierType,
+        value: auth.identifierValue
+      })),
+      authorizations: order.authorizations.map((auth) => buildUrl(profileId, `/authorizations/${auth.id}`)),
+      finalize: buildUrl(profileId, `/orders/${order.id}/finalize`),
```
sheensantoscapadngan, line 301: IGNORE PLACEMENT OF COMMENT

It looks like there's eslint issues throughout the files? 
image

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -0,0 +1,837 @@
...
+            actorAuthMethod: null,
+            actorOrgId,
+            profileId,
+            csr,
+            notBefore: finalizingOrder.notBefore ? new Date(finalizingOrder.notBefore) : undefined,
+            notAfter: finalizingOrder.notAfter ? new Date(finalizingOrder.notAfter) : undefined,
+            validity: {
+              // TODO: read config from the profile to get the expiration time instead
+              ttl: (24 * 60 * 60 * 1000).toString()
+            },
```
sheensantoscapadngan: I think we can leave this blank when notBefore and notAfter is defined? this is TTL of the issued cert

backend/src/ee/services/pki-acme/pki-acme-challenge-dal.ts
```diff
@@ -0,0 +1,175 @@
+import { TDbClient } from "@app/db";
+import { TableName, TPkiAcmeChallenges } from "@app/db/schemas";
```
sheensantoscapadngan, line 2: IGNORE PLACEMENT OF COMMENT

It looks like I was able to request a cert via certbot that had a digital signature key usage but it's not allowed per the  profile's certificate template

image
image
image

backend/src/ee/services/pki-acme/pki-acme-challenge-dal.ts
```diff
@@ -0,0 +1,175 @@
...
+import { DatabaseError } from "@app/lib/errors";
+import { ormify, selectAllTableCols } from "@app/lib/knex";
+import { Knex } from "knex";
+import { AcmeAuthStatus, AcmeChallengeStatus, AcmeOrderStatus } from "./pki-acme-schemas";
+
```
sheensantoscapadngan, line 7: IGNORE PLACEMENT OF COMMENT

The same thing happens when I define an EKU 
image

image

backend/src/ee/services/pki-acme/pki-acme-challenge-dal.ts
```diff
@@ -0,0 +1,175 @@
...
+export type TPkiAcmeChallengeDALFactory = ReturnType<typeof pkiAcmeChallengeDALFactory>;
+
```
sheensantoscapadngan, line 9: IGNORE PLACEMENT OF COMMENT

Same issue with above - no validation for when used key algorithm is not a part of the template. I think we're missing the validation for the ACME flow? 
image
image

backend/src/ee/services/pki-acme/pki-acme-challenge-dal.ts
```diff
@@ -0,0 +1,175 @@
...
+          })
+          .update({ status: AcmeOrderStatus.Ready });
+      }
+
+      return challenge;
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Update certificate profile" });
+    }
+  };
+
```
sheensantoscapadngan, line 61: IGNORE PLACEMENT OF COMMENT

Minor but I think the "check" here in the COPY icon should disappear after some time so I can copy it again
image

backend/src/ee/services/pki-acme/pki-acme-errors.ts
fangpenlin: good caught. The RFC didn't say anything about it, but let's keep it the same to RFC to maximize change of it getting parsed correctly by any ACME client.

backend/src/ee/services/pki-acme/pki-acme-service.ts
fangpenlin: The final issuing is calling [the existing v3 cert signing function there](https://github.com/Infisical/infisical/blob/cb1dd9713d94c29ce691c572a5772635871fc067/backend/src/services/certificate-v3/certificate-v3-service.ts#L478-L488). The CSR validation is performed there in the final step when the user trying to request it.

Certainly, it's ideal to reject the order request early based on the bad identifiers. But there are still many nice-to-have smater features like this we can implement later. Otherwise the scope will be too big for now.

backend/src/ee/services/pki-acme/pki-acme-service.ts
fangpenlin: I see nothing from my end. We also have CI lint check pass. Not sure why you're seeing this. 

image

> IGNORE PLACEMENT OF COMMENT

Is that the error you're seeing?

backend/src/ee/services/pki-acme/pki-acme-service.ts
fangpenlin: I see, make sense

backend/src/ee/services/pki-acme/pki-acme-challenge-dal.ts
fangpenlin: hmm, that's indeed really odd. We need to look into the v3 code a bit more on that

backend/src/ee/services/pki-acme/pki-acme-challenge-dal.ts
fangpenlin: For the validation, it's relying on the `signCertificateFromProfile` function. But I need to double check where the validation happens. Maybe I miss another call before this method.

backend/src/ee/services/pki-acme/pki-acme-challenge-dal.ts
fangpenlin: Make sense, will do

backend/src/ee/services/pki-acme/pki-acme-challenge-service.ts
fangpenlin: yeah, this is something we need to improve later. Probably making it running in a queue instead. But to make things simple, it's just like that for now. Will think about how to improve it later.

PR #5026, opened 2025-12-11
fix(gitlab-oauth): preserve instanceUrl during OAuth callback flow

### LINKED ISSUE #4569: Error 401 and 400 when connecting infisical self-hosted to gitlab self-hosted
### Describe the bug
I cannot connect infisical to GitLab (both self hosted). When I try to do it with the token, it says, that it doesn't accept local IPs.. When I get to the OAuth method, self-hosted option, then, I got the error 400 and 401. I follow the steps in this guide: [Infisical Self-hosted](https://infisical.com/docs/integrations/app-connections/gitlab#self-hosted-instance-setup).  

### To Reproduce
Steps to reproduce the behaviour:
1. Go to the link above.
2. Copy and set the following Env. variables in the infisical self-hosted instance:
   2.1.  INF_APP_CONNECTION_GITLAB_OAUTH_CLIENT_ID
   2.2.  INF_APP_CONNECTION_GITLAB_OAUTH_CLIENT_SECRET
   2.3. SITE_URL (here, not sure what address to point at... I set it up as 127.0.0.1:port number, but the call back always return to the 127.0.01 but without port number which makes it fail)
2.4. Since both addresses needed to match, then, it looks like this: Env Variable: http://my-server:port.com/organization/app-connections/gitlab/oauth/callback

3. Got to these screens:

Image

Image

For some reason, where it says "you will be redirected to ..." ***I set it up as 127.0.0.1:port number, but the call back always return to the 127.0.01 but without port number which makes it fail***

How to make it call back to 127.0.01:port as a whole and not as though it is expecting a default address (localhost:8080)?

Also, the configuration in the YAML file is (so I cannot use the default port 8080 for infisical. Hope this helps):

    ports:
      - 9000:8080

### Expected behaviour
After this step it should return to the infisical instance, so I can carry out the second section of the procedure.
Is anyone else experiecing (or did in the past) this issue of not being able to connect both self-hosted instance?
I need a bit of help

digikwal:
### **Error 401 and 400 when connecting Infisical self-hosted to GitLab self-hosted**

Fixes #4569 

---

### **Summary**

This PR fixes an issue where the **GitLab OAuth flow always defaulted to `https://gitlab.com`**, even when the user entered a self-hosted GitLab instance URL.

During the OAuth redirect, the frontend serializes form data into `localStorage`.
However, `GitLabFormData` did not explicitly define `credentials.instanceUrl`, so the value was dropped during serialization/deserialization.
As a result, the backend received **no instanceUrl**, and the OAuth exchange attempted to authenticate against gitlab.com, causing:

```
Failed to exchange OAuth code: 401 invalid_client
```

---

### **Root Cause**

- Credentials is pulled from the GitLab connection type,
- But ONLY the fields used after the OAuth flow (code, refreshToken, accessToken, etc.).
- InstanceUrl is NOT part of TGitLabConnection.credentials until after validation.

So when the frontend parses the callback payload using this type, Zod or TS validation strips unknown fields, including:

`credentials.instanceUrl → removed`

This matches our observation:
- The OAuth redirect validly uses our self-hosted GitLab URL
- But the POST to /api/v1/app-connections/gitlab always lacks instanceUrl
- Backend falls back to gitlab.com
- Token exchange fails with: "invalid_client" (because the client ID is for our host, not gitlab.com)

This means:
* `GitLabFormData` was defined as:

  ```ts
  Pick<TGitLabConnection, "name" | "method" | "description" | "credentials">
  ```
* `TGitLabConnection["credentials"]` is a **discriminated union**, and in the OAuth case does *not* include `instanceUrl`.
* When the form data was stored in `localStorage`, the missing field caused:

  * `credentials.instanceUrl` → **undefined**
  * backend fell back to default GitLab URL → **[https://gitlab.com/](https://gitlab.com)**

This caused all OAuth token exchanges to fail for self-hosted GitLab.

---

### **Fix**

`GitLabFormData` is rewritten to explicitly include the OAuth credential structure:

```ts
export type GitLabFormData = BaseFormData &
  Pick<TGitLabConnection, "name" | "method" | "description"> & {
    credentials: {
      code: string;
      instanceUrl?: string;
    };
  };
```

This ensures:

* The instance URL entered on the form is stored.
* The callback handler receives it.
* The backend calls the correct `instanceUrl/oauth/token`.

---

#### **Before fix**

1. Enter self-hosted GitLab URL: `https://ops-gitlab.example.com`
2. Start OAuth flow.
3. Callback logs show:

   ```
   gitlabInstanceUrl: https://gitlab.com
   ```
4. OAuth token exchange fails with 401.

#### **After fix**

1. Enter same self-hosted GitLab URL.
2. Start OAuth flow.
3. Callback logs show:

   ```
   gitlabInstanceUrl: https://ops-gitlab.example.com
   ```
4. OAuth token exchange succeeds.
5. Connection is created.

---

### **Impact**

* Fixes OAuth authentication for all self-hosted GitLab deployments.
* No breaking changes.
* Affects GitLab OAuth flow and PAT-based connections.

### DIFF
--- a/frontend/src/pages/organization/AppConnections/OauthCallbackPage/OauthCallbackPage.types.ts
+++ b/frontend/src/pages/organization/AppConnections/OauthCallbackPage/OauthCallbackPage.types.ts
@@ -25,7 +25,12 @@ export type GitHubRadarFormData = BaseFormData &
   Pick<TGitHubRadarConnection, "name" | "method" | "description">;
 
 export type GitLabFormData = BaseFormData &
-  Pick<TGitLabConnection, "name" | "method" | "description" | "credentials">;
+  Pick<TGitLabConnection, "name" | "method" | "description"> & {
+    credentials: {
+      code: string;
+      instanceUrl?: string;
+    }
+  };
 
 export type AzureKeyVaultFormData = BaseFormData &
   Pick<TAzureKeyVaultConnection, "name" | "method" | "description"> &


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/481350ef-6f11-4192-a6e4-c0bbbb9578f5?refs=7a88dbef6dc4f6499e51c9e801ca0e8a9b799a5e&source=prChecksComment) |

PR #5063, opened 2025-12-17
fix(cdn): add index option to serve-ui configuration to fix index.html caching

victorvhs017:
## Context

The fastify/static plugin was intercepting the index.html here when it was called on the root path of the application

image

Then it never reached the custom rule in this case

image

The custom rule above is where we set the cache control header for the index file, because it should never be cached.

We can see the difference when we call the gamma base path and any other path:

image

The login route (and any other route) works because the wildcard path is false in fastify/static.

To fix the root path, too, we added the `index: false` property https://www.npmjs.com/package/@fastify/send#index, allowing our custom rule to handle it.

This is what I got from an image built after this change:

image

In the end, we have:
- All routes loading without cache 
- If CDN_HOST is not set, fastify/static serves all frontend files with 30d cache
- If CDN_HOST is not set, fastify/static serves all frontend files with 30d cache, except for the assets that are coming from the CDN_HOST URL.

## Steps to verify the change

- Build an image and do the same curl test above

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/plugins/serve-ui.ts
+++ b/backend/src/server/plugins/serve-ui.ts
@@ -79,7 +79,8 @@ export const registerServeUI = async (
       root: frontendPath,
       wildcard: false,
       maxAge: "30d",
-      immutable: true
+      immutable: true,
+      index: false
     });
 
     server.route({


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/bc7fc449-55f1-4329-981b-89368a980598?refs=f53b0c92f000d562a3c7421ca5985a27b28fa38e&source=prChecksComment) |

PR #5093, opened 2025-12-27
improvement(api): reduce identity access token delete time

fangpenlin:
## Context

ref: https://linear.app/infisical/issue/PLATFRM-94/reduce-time-to-delete-access-tokens

## Steps to verify the change

To test this PR, one needs to create a tons of `identity_access_token`. I vibe coded a script for generating testing data here:

`gen_test_data.py`
```python
#!/usr/bin/env python3
"""
Generate test data for identity_access_tokens table.
Generates 4 types of tokens: revoked, use-limit-exceeded, expired, and active.
"""

import uuid
import random
import argparse
import csv
import os
from datetime import datetime, timedelta, timezone


def generate_row(token_type: str, now: datetime) -> dict:
    """
    Generate a single row of test data.
    
    Args:
        token_type: Type of token - 'revoked', 'use_limit_exceeded', 'expired', or 'active'
        now: Current timestamp for expiration calculations
    """
    # Generate required fields
    id_val = str(uuid.uuid4())
    identity_id = str(uuid.uuid4())
    auth_method = random.choice(['password', 'oauth', 'api_key', 'sso', 'magic_link', 'token'])
    
    # Generate TTL (must be > 0)
    access_token_ttl = random.randint(3600, 2592000)  # 1 hour to 30 days in seconds
    access_token_max_ttl = access_token_ttl + random.randint(0, 2592000)  # Max TTL >= TTL
    
    # Initialize defaults
    is_access_token_revoked = False
    access_token_num_uses = 0
    access_token_num_uses_limit = 0
    
    # Generate timestamps and fields based on token type
    if token_type == 'revoked':
        # Revoked token: isAccessTokenRevoked = True
        # Should not be expired or use-limit-exceeded (mutually exclusive)
        is_access_token_revoked = True
        # Not expired: created_at >= now - ttl
        min_created = now - timedelta(seconds=access_token_ttl - random.randint(0, access_token_ttl // 2))
        created_at = min_created + timedelta(seconds=random.randint(0, int((now - min_created).total_seconds())))
        access_token_last_renewed_at = None
        # Not use-limit-exceeded: either no limit or uses < limit
        access_token_num_uses_limit = random.choice([0, random.randint(1, 200)])
        if access_token_num_uses_limit > 0:
            access_token_num_uses = random.randint(0, access_token_num_uses_limit - 1)
        else:
            access_token_num_uses = random.randint(0, 100)
    
    elif token_type == 'use_limit_exceeded':
        # Use limit exceeded: accessTokenNumUsesLimit > 0 AND accessTokenNumUses >= accessTokenNumUsesLimit
        # Should not be expired or revoked
        access_token_num_uses_limit = random.randint(1, 200)
        access_token_num_uses = random.randint(access_token_num_uses_limit, access_token_num_uses_limit + 100)
        # Not expired: created_at >= now - ttl
        min_created = now - timedelta(seconds=access_token_ttl - random.randint(0, access_token_ttl // 2))
        created_at = min_created + timedelta(seconds=random.randint(0, int((now - min_created).total_seconds())))
        access_token_last_renewed_at = None
    
    elif token_type == 'expired':
        # Expired token: (last_renewed_at or created_at) + ttl < now
        # Should not be revoked or use-limit-exceeded
        max_created = now - timedelta(seconds=access_token_ttl + random.randint(1, 86400 * 7))  # Expired by 1s to 7 days
        created_at = max_created - timedelta(days=random.randint(0, 30))
        access_token_last_renewed_at = None
        # Not use-limit-exceeded: either no limit or uses < limit
        access_token_num_uses_limit = random.choice([0, random.randint(1, 200)])
        if access_token_num_uses_limit > 0:
            access_token_num_uses = random.randint(0, access_token_num_uses_limit - 1)
        else:
            access_token_num_uses = random.randint(0, 100)
    
    else:  # token_type == 'active'
        # Active token: none of the above
        # Not expired: created_at >= now - ttl
        min_created = now - timedelta(seconds=access_token_ttl - random.randint(0, access_token_ttl // 2))
        created_at = min_created + timedelta(seconds=random.randint(0, int((now - min_created).total_seconds())))
        access_token_last_renewed_at = None
        # Not use-limit-exceeded: either no limit or uses < limit
        access_token_num_uses_limit = random.choice([0, random.randint(1, 200)])
        if access_token_num_uses_limit > 0:
            access_token_num_uses = random.randint(0, access_token_num_uses_limit - 1)
        else:
            access_token_num_uses = random.randint(0, 100)
    
    # Other fields
    access_token_last_used_at = None if random.random() > 0.3 else (created_at + timedelta(seconds=random.randint(0, 86400)))
    name = f"mock-token-{random.randint(1000, 9999)}"
    access_token_period = random.randint(0, 2592000)
    updated_at = created_at + timedelta(seconds=random.randint(0, int((now - created_at).total_seconds())))
    
    return {
        'id': id_val,
        'accessTokenTTL': access_token_ttl,
        'accessTokenMaxTTL': access_token_max_ttl,
        'accessTokenNumUses': access_token_num_uses,
        'accessTokenNumUsesLimit': access_token_num_uses_limit,
        'accessTokenLastUsedAt': access_token_last_used_at.isoformat() if access_token_last_used_at else None,
        'accessTokenLastRenewedAt': access_token_last_renewed_at.isoformat() if access_token_last_renewed_at else None,
        'isAccessTokenRevoked': is_access_token_revoked,
        'identityUAClientSecretId': None,
        'identityId': identity_id,
        'createdAt': created_at.isoformat(),
        'updatedAt': updated_at.isoformat(),
        'name': name,
        'authMethod': auth_method,
        'accessTokenPeriod': access_token_period,
        'subOrganizationId': None,
    }


def generate_csv(num_rows: int, revoked_percent: float, use_limit_exceeded_percent: float, 
                 expired_percent: float, csv_filename: str):
    """Generate CSV file with test data."""
    now = datetime.now(timezone.utc)
    
    # Calculate counts for each category
    num_revoked = int(num_rows * revoked_percent / 100)
    num_use_limit_exceeded = int(num_rows * use_limit_exceeded_percent / 100)
    num_expired = int(num_rows * expired_percent / 100)
    num_active = num_rows - num_revoked - num_use_limit_exceeded - num_expired
    active_percent = (num_active / num_rows * 100) if num_rows > 0 else 0
    
    print(f"Generating {num_rows:,} rows:")
    print(f"  Revoked tokens: {num_revoked:,} ({revoked_percent}%)")
    print(f"  Use-limit-exceeded tokens: {num_use_limit_exceeded:,} ({use_limit_exceeded_percent}%)")
    print(f"  Expired tokens: {num_expired:,} ({expired_percent}%)")
    print(f"  Active tokens: {num_active:,} ({active_percent:.2f}%)")
    print(f"  Writing to {csv_filename}...")
    
    # Column order matching the table schema
    fieldnames = [
        'id',
        'accessTokenTTL',
        'accessTokenMaxTTL',
        'accessTokenNumUses',
        'accessTokenNumUsesLimit',
        'accessTokenLastUsedAt',
        'accessTokenLastRenewedAt',
        'isAccessTokenRevoked',
        'identityUAClientSecretId',
        'identityId',
        'createdAt',
        'updatedAt',
        'name',
        'authMethod',
        'accessTokenPeriod',
        'subOrganizationId',
    ]
    
    with open(csv_filename, 'w', newline='', encoding='utf-8') as f:
        writer = csv.DictWriter(f, fieldnames=fieldnames, quoting=csv.QUOTE_MINIMAL)
        writer.writeheader()
        
        # Generate revoked tokens
        for _ in range(num_revoked):
            row = generate_row('revoked', now)
            writer.writerow(row)
        
        # Generate use-limit-exceeded tokens
        for _ in range(num_use_limit_exceeded):
            row = generate_row('use_limit_exceeded', now)
            writer.writerow(row)
        
        # Generate expired tokens
        for _ in range(num_expired):
            row = generate_row('expired', now)
            writer.writerow(row)
        
        # Generate active tokens
        for _ in range(num_active):
            row = generate_row('active', now)
            writer.writerow(row)
    
    print(f"Done! Generated {num_rows:,} rows in {csv_filename}")


def generate_sql(csv_filename: str, sql_filename: str):
    """Generate SQL \\copy command for efficient bulk insert."""
    # Use relative path since \copy is client-side and reads from client filesystem
    # The CSV file should be in the same directory as the SQL file when running psql
    
    sql_content = f"""-- Bulk insert from CSV using \\copy (client-side)
-- Usage: psql -d your_database -f {sql_filename}
-- Note: \\copy is a psql meta-command that reads from the client's filesystem.
-- Make sure the CSV file ({csv_filename}) is in the same directory or use an absolute path.

\\copy public.identity_access_tokens (
    id,
    "accessTokenTTL",
    "accessTokenMaxTTL",
    "accessTokenNumUses",
    "accessTokenNumUsesLimit",
    "accessTokenLastUsedAt",
    "accessTokenLastRenewedAt",
    "isAccessTokenRevoked",
    "identityUAClientSecretId",
    "identityId",
    "createdAt",
    "updatedAt",
    name,
    "authMethod",
    "accessTokenPeriod",
    "subOrganizationId"
) FROM '{csv_filename}' WITH (FORMAT csv, HEADER true, NULL '');
"""
    
    with open(sql_filename, 'w', encoding='utf-8') as f:
        f.write(sql_content)
    
    print(f"Generated SQL file: {sql_filename}")


def main():
    parser = argparse.ArgumentParser(
        description="Generate test data for identity_access_tokens table (CSV + SQL)",
        formatter_class=argparse.RawDescriptionHelpFormatter,
        epilog="""
Examples:
  # Generate 1000 rows with 10% revoked, 5% use-limit-exceeded, 20% expired, 65% active
  python gen_test_data_v2.py 1000 --revoked-percent 10 --use-limit-exceeded-percent 5 --expired-percent 20
  
  # Generate 1 million rows with 5% revoked, 3% use-limit-exceeded, 15% expired, 77% active
  python gen_test_data_v2.py 1000000 --revoked-percent 5 --use-limit-exceeded-percent 3 --expired-percent 15
        """
    )
    parser.add_argument(
        'num_rows',
        type=int,
        help='Number of rows to generate'
    )
    parser.add_argument(
        '--revoked-percent',
        type=float,
        default=10.0,
        help='Percentage of rows that should have revoked tokens (default: 10.0)'
    )
    parser.add_argument(
        '--use-limit-exceeded-percent',
        type=float,
        default=5.0,
        help='Percentage of rows that should have use-limit-exceeded tokens (default: 5.0)'
    )
    parser.add_argument(
        '--expired-percent',
        type=float,
        default=20.0,
        help='Percentage of rows that should have expired tokens (default: 20.0)'
    )
    parser.add_argument(
        '--csv',
        default='identity_access_tokens.csv',
        help='Output CSV filename (default: identity_access_tokens.csv)'
    )
    parser.add_argument(
        '--sql',
        default='identity_access_tokens_insert.sql',
        help='Output SQL filename (default: identity_access_tokens_insert.sql)'
    )
    
    args = parser.parse_args()
    
    # Validate percentages
    if args.revoked_percent < 0 or args.revoked_percent > 100:
        parser.error("--revoked-percent must be between 0 and 100")
    
    if args.use_limit_exceeded_percent < 0 or args.use_limit_exceeded_percent > 100:
        parser.error("--use-limit-exceeded-percent must be between 0 and 100")
    
    if args.expired_percent < 0 or args.expired_percent > 100:
        parser.error("--expired-percent must be between 0 and 100")
    
    total_percent = args.revoked_percent + args.use_limit_exceeded_percent + args.expired_percent
    if total_percent > 100:
        parser.error(f"Sum of percentages ({total_percent}%) exceeds 100%. Active tokens will be {100 - total_percent}%")
    
    if args.num_rows <= 0:
        parser.error("num_rows must be greater than 0")
    
    # Generate CSV
    generate_csv(args.num_rows, args.revoked_percent, args.use_limit_exceeded_percent, 
                 args.expired_percent, args.csv)
    
    # Generate SQL
    generate_sql(args.csv, args.sql)
    
    print("\nTo insert the data, run:")
    print(f"  psql -d your_database -f {args.sql}")
    print(f"\n  # Or use \\copy directly in psql:")
    print(f"  psql -d your_database -c \"\\copy public.identity_access_tokens FROM '{args.csv}' WITH (FORMAT csv, HEADER true, NULL '');\"")


if __name__ == "__main__":
    main()

```

Then you can generate the CSV file, here's the cmd for generating 7M records with little expired + other types, try to mimic one of the production db env:

```bash
python3 gen_test_data.py 7000000 --revoked-percent 0.01 --use-limit-exceeded-percent 0.01 --expired-percent 5
```

then import them into your local dev db:

```bash
psql postgresql://infisical:infisical@localhost:5432/infisical -c "\copy public.identity_access_tokens FROM 'identity_access_tokens.csv' WITH (FORMAT csv, HEADER true, NULL '');"
```

It takes a while to load them all. Before switching to this branch, you can run the analytics with:

```SQL
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON )
select "id"
from "identity_access_tokens"
where "isAccessTokenRevoked" = 't'
   or ("accessTokenNumUsesLimit" > '0' and
       "accessTokenNumUses" >= "identity_access_tokens"."accessTokenNumUsesLimit")
   or ("accessTokenTTL" > '0' and
    -- Check if the token's effective expiration time has passed.
    -- The expiration time is calculated by adding its TTL to its last renewal/creation time.
       COALESCE(
               "identity_access_tokens"."accessTokenLastRenewedAt", -- Use last renewal time if available
               "identity_access_tokens"."createdAt" -- Otherwise, use creation time
       ) AT TIME ZONE 'UTC'
           + make_interval(
               secs => LEAST(
                       "identity_access_tokens"."accessTokenTTL", -- Token's specified TTL
                       '315360000' -- Capped by MAX_TTL (parameterized value)
                       )
             )
           < NOW() AT TIME ZONE 'UTC' -- Check if the calculated time is before now
    )
```

The paste the query result to https://explain.dalibo.com. More likely than not, you're going to see a full table scan. Like [this](https://explain.dalibo.com/plan/ef36e944dfb2e8g4):

image

That's why it's very slow. Plus, we clean every 24 hours, it's gets more in a full day, so in this PR, we increase the frequency to per hour. After switching to this PR, the db migration should create the following indexes:

- `idx_identity_access_tokens_revoked`
- `idx_identity_access_tokens_num_uses_with_limit`
- `idx_identity_access_tokens_expiration`

And with our new query in the PR, the query planner should be able to use those index. Here's the new query you run with explain:

```sql
SET random_page_cost = 1.1;
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON)
SELECT "id"
FROM "identity_access_tokens"
WHERE "isAccessTokenRevoked" = true

UNION ALL

SELECT "id"
FROM "identity_access_tokens"
WHERE "accessTokenNumUsesLimit" > 0
  AND "accessTokenNumUses" >= "identity_access_tokens"."accessTokenNumUsesLimit"

UNION ALL

SELECT "id"
FROM "identity_access_tokens"
WHERE "accessTokenTTL" > 0
  AND (
    -- Check if the token's effective expiration time has passed.
    -- The expiration time is calculated by adding its TTL to its last renewal/creation time.
    (COALESCE(
             "identity_access_tokens"."accessTokenLastRenewedAt", -- Use last renewal time if available
             "identity_access_tokens"."createdAt" -- Otherwise, use creation time
     ) AT TIME ZONE 'UTC') -- Convert to UTC for immutable index
        + make_interval(
            secs => LEAST(
                    "identity_access_tokens"."accessTokenTTL", -- Token's specified TTL
                    315360000 -- Capped by MAX_TTL (10 years in seconds)
                    )
          )
        < '2026-01-07 04:27:08.555+00'::timestamptz AT TIME ZONE 'UTC' -- Compared to "now"
    )
```

Please note that the `SET random_page_cost = 1.1;` is needed. In modern SSD powered db, the original default `random_page_cost` value is set to 4.0, which is too high, causing psql engine to avoid using index. We need another PR to turn our infra to set this `random_page_cost` value as well. With the new indexes, instead of a full scan, you should be able to see indexes used [like this](https://explain.dalibo.com/plan/1dgbe1dcfce879e2):

image

### Test for init db race conditions

Based on the review feedbacks, it seems like we need that manual lock because the migration table initialization is not protected and could cause problems if we have multiple instances. To test and see if the locks work correctly, one can replace the app db init code with the following to simulate multiple instances trying to init the db:

```typescript

    const runApplicationMigrations = async () => {
      await ensureMigrationTables(applicationDb, logger);
      logger.info("Running application migrations.");
      const didPreviousInstanceRunMigration = !(await applicationDb.migrate
        .status(migrationConfig)
        .catch(migrationStatusCheckErrorHandler));
      if (didPreviousInstanceRunMigration) {
        logger.info("No application migrations pending: Applied by previous instance. Skipping migration process.");
        return;
      }

      // Use startup lock to ensure only one instance runs migrations at a time
      await withStartupLock(applicationDb, logger, async () => {
        await applicationDb.migrate.latest(migrationConfig);
      });
      logger.info("Finished application migrations.");
    };

    // Run multiple concurrent instances to simulate race conditions
    const concurrentInstances = 8;
    logger.info(`Starting ${concurrentInstances} concurrent migration attempts to test lock mechanism`);
    await Promise.all(Array.from({ length: concurrentInstances }, () => runApplicationMigrations()));
    logger.info(`All ${concurrentInstances} concurrent migration attempts completed`);
```

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

### DIFF
--- a/backend/src/services/identity-access-token/identity-access-token-dal.ts
+++ b/backend/src/services/identity-access-token/identity-access-token-dal.ts
@@ -99,6 +99,10 @@ export const identityAccessTokenDALFactory = (db: TDbClient) => {
     do {
       try {
         const deleteBatch = async (dbClient: Knex | Knex.Transaction) => {
+          // The default random_page_cost is 4.0, which is too high for this query.
+          // With SSD powered database, random access is way faster.
+          // We set it to 1.1 to make the query opt for random access and thus more likely to use the index.
+          await dbClient.raw(`SET LOCAL random_page_cost = 1.1`);
           const idsToDeleteQuery = getExpiredTokensQuery(dbClient, now).limit(BATCH_SIZE);
           return dbClient(TableName.IdentityAccessToken).whereIn("id", idsToDeleteQuery).del().returning("id");
         };
@@ -109,7 +113,7 @@ export const identityAccessTokenDALFactory = (db: TDbClient) => {
         } else {
           // eslint-disable-next-line no-await-in-loop
           deletedTokenIds = await db.transaction(async (trx) => {
-            await trx.raw(`SET statement_timeout = ${QUERY_TIMEOUT_MS}`);
+            await trx.raw(`SET LOCAL statement_timeout = ${QUERY_TIMEOUT_MS}`);
             return deleteBatch(trx);
           });
         }


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/465908d7-1dc2-4453-ba74-7d4535c7465b?refs=52547288974b30bdf361aaef5298a95c9b956621&source=prChecksComment) |

backend/src/auto-start-migrations.ts
```diff
@@ -83,8 +83,7 @@ export const runMigrations = async ({ applicationDb, auditLogDb, logger }: TArgs
       });
     }
 
-    await applicationDb.transaction(async (tx) => {
-      await tx.raw("SELECT pg_advisory_xact_lock(?)", [PgSqlLock.BootUpMigration]);
```
fangpenlin, line 87: To the reviewer: Is this really needed? The reason I am deleting it is because it's causing a deadlock with `CREATE INDEX CONCURRENTLY`:

image

The `CREATE INDEX CONCURRENTLY` statement needs to wait till all txn are done, but the advisory lock we've acquired before the migration is in a txn is never going to finish untill the migration finishes. Thus, it created a mutal waiting deadlock.

The knex migration system comes with its own lock system, it's unclear why this manual lock is needed in the first place. But I want to be extra careful so that I am calling it out, a bit more context is needed.

backend/src/services/identity-access-token/identity-access-token-dal.ts
```diff
@@ -30,56 +30,73 @@ export const identityAccessTokenDALFactory = (db: TDbClient) => {
...
+            < ?::timestamptz AT TIME ZONE 'UTC'                             -- Check if the calculated time is before now (cast to UTC timestamp for comparison)
+            `,
+          [MAX_TTL, nowTimestamp]
+        )
+        .select("id");
+
+      // Notice: we broken down the queyr into multiple queries and union them to avoid index usage issues.
+      //         each query got their own index for better performance, therefore, if you want to change
+      //         the query, you need to update the indexes accordingly to avoid performance regressions.
+      return revokedTokensQuery.unionAll(exceededUsageLimitQuery).unionAll(expiredTTLQuery);
```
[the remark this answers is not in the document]
fangpenlin: fixed

backend/src/auto-start-migrations.ts
sheensantoscapadngan: I think without this PG_ADVISORY_LOCK, it would be problematic for fresh DB initialization with multiple instances? Like first time deploy of Infisical with multiple instances

backend/src/db/migrations/20251227020624_add-identity-access-token-revoked-index.ts
```diff
@@ -0,0 +1,25 @@
...
+export async function up(knex: Knex): Promise<void> {
+  if (
+    (await knex.schema.hasTable(TableName.IdentityAccessToken)) &&
+    (await knex.schema.hasColumn(TableName.IdentityAccessToken, "isAccessTokenRevoked"))
+  ) {
+    await knex.raw(`
+        CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_identity_access_tokens_revoked
+        ON ${TableName.IdentityAccessToken} ("id")
+        WHERE "isAccessTokenRevoked" = 't'
+    `);
```
sheensantoscapadngan: hmmm do we have an estimate of how long this would take when ran in Infisical US? The concern is that if it exceeds 3 mins, our ECS health checks will fail and KILL the process altogether so it's never able to complete running. I think it would result in an invalid index if interrupted?

backend/src/db/migrations/20251227020625_add-identity-access-token-num-uses-index.ts
```diff
@@ -0,0 +1,27 @@
...
+    (await knex.schema.hasColumn(TableName.IdentityAccessToken, "accessTokenNumUses")) &&
+    (await knex.schema.hasColumn(TableName.IdentityAccessToken, "accessTokenNumUsesLimit"))
+  ) {
+    await knex.raw(`
+        CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_identity_access_tokens_num_uses_with_limit
+        ON ${TableName.IdentityAccessToken} ("id")
+        WHERE "accessTokenNumUsesLimit" > 0
+          AND "accessTokenNumUses" >= "identity_access_tokens"."accessTokenNumUsesLimit"
+    `);
+  }
```
sheensantoscapadngan: same concern here. perhaps we'd have to run this manually in cloud before deploying this fix if ever

backend/src/db/migrations/20251227020626_add-identity-access-token-expiration-index.ts
```diff
@@ -0,0 +1,31 @@
...
+    await knex.raw(`
+        CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_identity_access_tokens_expiration
+        ON ${TableName.IdentityAccessToken} (
+            (COALESCE("accessTokenLastRenewedAt", "createdAt") AT TIME ZONE 'UTC' + make_interval(secs => LEAST(
+            "identity_access_tokens"."accessTokenTTL",
+            '315360000'
+            )))
+        )
+        WHERE "accessTokenTTL" > 0
+    `);
```
sheensantoscapadngan: and here

backend/src/db/migrations/20251227020624_add-identity-access-token-revoked-index.ts
fangpenlin: I tested with local dev env on my MacBook with simulated data in the size of EU production db, it took around only maybe something like 10sec ~ 20sec or so to create the full index for three of them. In production and US, because the db is more busy and we have more data, so yeah, potentially it may take longer time. And yeah, if it's the case, we may have a invalid index as a result. But it's safe to have an invalid index. The worst case would be we need to remove the invalid index and manually run the create index script again. It shouldn't interrupt the operation of the production server based on my understanding.

Surely, the most reliable way is to run the SQL manually outside of the bootstrapping procedure. Or, we can roll out to EU first to see how it goes and decide whether there's a need to do it manually or not.

backend/src/auto-start-migrations.ts
fangpenlin: thanks for the context! It's very helpful. I wrote some code to try to reproduce the issue:

```
    logger.info("!!! Reproducing race condition issue !!!");
    const initializeDb = async () => {
      const currentVersion = await applicationDb.migrate.currentVersion(migrationConfig);
      logger.info(`Current database version: ${currentVersion}`);
    };
    // the currentVersion will call ensureTable, if many of them running in the same time,
    // it should be able to reproduce the race condition issue
    await Promise.all(Array.from({ length: 32 }, () => initializeDb()));
```

Replace the app db migration code with the above in `auto-start-migrations.ts`, reset db then try to boot up infisical for a few times. Indeed, I saw many errors, and the lock table ends up like this with multiple rows:

image

I read the knex source code, it seems like by design, [they assume only one row exists in this table](https://github.com/knex/knex/blob/833f2257554caf0e1565a37a6402bd1a0cc09895/lib/migrations/migrate/Migrator.js#L329-L335). If there are multiple rows in the lock table, some manual clean up will be needed.

Okay, now we know that the problem only happens in the initial bootstrap process, I am going to add the lock back but only for the initial initialization.

fangpenlin: I have updated the code to only acquire the lock when the migration tables is not initialized. Here's the code to test, it simulates running multiple instances in the same time competing for database bootstrapping:

```typescript

    logger.info("!!! Reproducing race condition issue !!!");
    const runMigration = async (db: Knex) => {
      await ensureMigrationTables(db, logger);
      await applicationDb.transaction(async () => {
        logger.info("Running application migrations.");

        const didPreviousInstanceRunMigration = !(await applicationDb.migrate
          .status(migrationConfig)
          .catch(migrationStatusCheckErrorHandler));
        if (didPreviousInstanceRunMigration) {
          logger.info("No application migrations pending: Applied by previous instance. Skipping migration process.");
          return;
        }

        try {
          await applicationDb.migrate.latest(migrationConfig);
        } catch (err) {
          // Handle MigrationLocked error gracefully - this occurs when another instance
          // is already running migrations, which is expected in multi-instance scenarios
          const isMigrationLockedError =
            (err && typeof err === "object" && "name" in err && err.name === "MigrationLocked") ||
            (err && typeof err === "object" && "type" in err && err.type === "LockError") ||
            (err instanceof Error && err.message?.includes("Migration table is already locked"));

          if (isMigrationLockedError) {
            logger.info("Migration table is locked by another instance. Skipping migration for this instance.");
            return;
          }
          logger.error(err, "Failed to run application migrations");
          throw err;
        }
        logger.info("Finished application migrations.");
      });
    };
    // the currentVersion will call ensureTable, if many of them running in the same time,
    // it should be able to reproduce the race condition issue
    await Promise.all(Array.from({ length: 8 }, () => runMigration(applicationDb)));
```

Just replace the app db init code, reset db, and try to boot up again. You will see while different promise trying to init the tables, but only one could work, and the migration will succeed without a problem (or table is already locked error in real environment, but which is fine actually, next bootup it will just work)

backend/src/db/migrations/20251227020625_add-identity-access-token-num-uses-index.ts
fangpenlin: same as mentioned here:

https://github.com/Infisical/infisical/pull/5093#discussion_r2669775003

backend/src/db/migrations/20251227020626_add-identity-access-token-expiration-index.ts
fangpenlin: same as mentioned here:

https://github.com/Infisical/infisical/pull/5093#discussion_r2669775003

backend/src/db/migrations/20251227020624_add-identity-access-token-revoked-index.ts
sheensantoscapadngan: makes sense to me!

backend/src/auto-start-migrations.ts
fangpenlin: okay, I just solved the startup db is lock error as well. Please have a look

fangpenlin:
@greptile-apps please review again

backend/src/auto-start-migrations.ts
fangpenlin: here's the new code for simulating multiple instances by the way:

```typescript

    const runApplicationMigrations = async () => {
      await ensureMigrationTables(applicationDb, logger);
      logger.info("Running application migrations.");
      const didPreviousInstanceRunMigration = !(await applicationDb.migrate
        .status(migrationConfig)
        .catch(migrationStatusCheckErrorHandler));
      if (didPreviousInstanceRunMigration) {
        logger.info("No application migrations pending: Applied by previous instance. Skipping migration process.");
        return;
      }

      // Use startup lock to ensure only one instance runs migrations at a time
      await withStartupLock(applicationDb, logger, async () => {
        await applicationDb.migrate.latest(migrationConfig);
      });
      logger.info("Finished application migrations.");
    };

    // Run multiple concurrent instances to simulate race conditions
    const concurrentInstances = 8;
    logger.info(`Starting ${concurrentInstances} concurrent migration attempts to test lock mechanism`);
    await Promise.all(Array.from({ length: concurrentInstances }, () => runApplicationMigrations()));
    logger.info(`All ${concurrentInstances} concurrent migration attempts completed`);
```

backend/src/auto-start-migrations.ts
```diff
@@ -30,6 +34,227 @@ const migrationStatusCheckErrorHandler = (err: Error) => {
...
+    const staleThreshold = new Date(Date.now() - STALE_HEARTBEAT_THRESHOLD_MS);
+    const result = await db(startupLockTableName)
+      // 1. is_locked = 0 (lock is free), OR
+      .where({ is_locked: 0 })
+      // 2. is_locked = 1 AND heartbeat is stale (previous instance crashed)
+      .orWhere((qb) => {
+        void qb.where({ is_locked: 1 }).andWhere((subQb) => {
+          void subQb.whereNull("heartbeat_updated_at").orWhere("heartbeat_updated_at", "<", staleThreshold);
+        });
+      })
```
sheensantoscapadngan, line 188: hmm could it be problematic that for `staleThreshold` we're using the instance time but heartbeat_updated_at is database time?

fangpenlin: good point. fixed. now we are using the clock from db instead of js

PR #5270, opened 2026-01-26
feat(pki): add certificate details page

saifsmailbox98:
## Context

Previously, certificates were only viewable in a table with basic info like common name, serial number, and expiry. You could perform actions via dropdown menus but there was no dedicated page to see all the details of a certificate.

This adds a certificate details page that you get to by clicking on a certificate row in the table. The page shows comprehensive certificate information including subject attributes (organization, country, etc.), SHA-256/SHA-1 fingerprints, basic constraints, key/signature algorithms, and the renewal chain if the cert was renewed from or has been renewed by another certificate. Certificate validity dates (Not Before/Not After) are displayed in UTC with local time shown on hover for convenience.

## Screenshots

cert-details

## Steps to verify the change

Generate a certificate and then click on a certificate (row) in the certificates table to view the page.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/certificate-router.ts
+++ b/backend/src/server/routes/v1/certificate-router.ts
@@ -1056,6 +1056,7 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
               })
               .optional(),
             caName: z.string().nullable().optional(),
+            caType: z.enum(["internal", "external"]).nullable().optional(),
             profileName: z.string().nullable().optional()
           })
         })
--- a/backend/src/services/certificate/certificate-dal.ts
+++ b/backend/src/services/certificate/certificate-dal.ts
@@ -465,6 +465,7 @@ export const certificateDALFactory = (db: TDbClient) => {
   type TCertificateWithRequestDetails = TCertificates & {
     caName?: string | null;
     profileName?: string | null;
+    caType?: "internal" | "external" | null;
   };
 
   // Flexible lookup filter for certificate queries - either id or serialNumber, not both
@@ -487,9 +488,15 @@ export const certificateDALFactory = (db: TDbClient) => {
           `${TableName.Certificate}.profileId`,
           `${TableName.PkiCertificateProfile}.id`
         )
+        .leftJoin(
+          TableName.InternalCertificateAuthority,
+          `${TableName.CertificateAuthority}.id`,
+          `${TableName.InternalCertificateAuthority}.caId`
+        )
         .select(selectAllTableCols(TableName.Certificate))
         .select(db.ref("name").withSchema(TableName.CertificateAuthority).as("caName"))
-        .select(db.ref("slug").withSchema(TableName.PkiCertificateProfile).as("profileName"));
+        .select(db.ref("slug").withSchema(TableName.PkiCertificateProfile).as("profileName"))
+        .select(db.ref("id").withSchema(TableName.InternalCertificateAuthority).as("internalCaId"));
 
       // Dynamic where clause based on filter
       if (filter.id) {
@@ -498,9 +505,22 @@ export const certificateDALFactory = (db: TDbClient) => {
         query = query.where(`${TableName.Certificate}.serialNumber`, filter.serialNumber);
       }
 
-      const result = (await query.first()) as TCertificateWithRequestDetails | undefined;
+      const result = (await query.first()) as
+        | (TCertificateWithRequestDetails & { internalCaId?: string | null })
+        | undefined;
 
-      return result;
+      if (!result) {
+        return undefined;
+      }
+
+      const { internalCaId, ...rest } = result;
+
+      let caType: "internal" | "external" | null = null;
+      if (result.caId) {
+        caType = internalCaId ? "internal" : "external";
+      }
+
+      return { ...rest, caType } as TCertificateWithRequestDetails;
     } catch (error) {
       throw new DatabaseError({ error, name: "Find certificate with full details" });
     }
--- a/frontend/src/hooks/api/certificates/types.ts
+++ b/frontend/src/hooks/api/certificates/types.ts
@@ -45,6 +45,7 @@ export type TCertificate = {
   };
   caName?: string | null;
   profileName?: string | null;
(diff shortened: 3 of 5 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/f82069e0-a1f9-4650-a12d-6c823675385d?refs=e70d773893d0175ed11ac8d23d5e896a5bcb57fc&source=prChecksComment) |

frontend/src/lib/fn/string.ts
```diff
@@ -17,6 +17,14 @@ export const camelCaseToSpaces = (input: string) => {
   return input.replace(/([a-z])([A-Z])/g, "$1 $2");
```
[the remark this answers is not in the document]
saifsmailbox98: Not related to my changes. Also, since it's frontend, shouldn't matter much.

backend/src/services/certificate/certificate-fns.ts
```diff
@@ -155,3 +162,47 @@ export const generatePkcs12FromCertificate = async ({
...
+      organization: parsedDn.organization,
+      organizationalUnit: parsedDn.ou,
+      country: parsedDn.country,
+      state: parsedDn.province,
+      locality: parsedDn.locality
+    };
+
+    // Calculate fingerprints and format with colons (e.g., "1A:2F:73:...")
+    const rawData = Buffer.from(certObj.rawData);
+    const formatFingerprint = (hash: string) => hash.toUpperCase().match(/.{2}/g)?.join(":") ?? hash.toUpperCase();
```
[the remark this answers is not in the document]
saifsmailbox98: Replaced with RE2

saifsmailbox98:
@greptile review this PR

backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -1032,7 +1032,48 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
...
+                pathLength: z.number().optional()
+              })
+              .optional(),
+            caName: z.string().nullable().optional(),
+            profileName: z.string().nullable().optional(),
+            renewedFromCertificate: z
+              .object({
+                id: z.string(),
+                serialNumber: z.string(),
+                commonName: z.string()
```
carlosmonastyrski: not only here, but also on the other files, why are we getting the common name here? for renewals everything should be the same except the expiration date, so the common name is going to be the same as the current certificate (mostly making this comments as on some flows common names can be empty and just use SAN and at the end it's going to be the same as the current certificate details)

backend/src/services/certificate/certificate-service.ts
```diff
@@ -109,14 +120,81 @@ export const certificateServiceFactory = ({
...
+      const kmsDecryptor = await kmsService.decryptWithKmsKey({
+        kmsId: certificateManagerKeyId
+      });
+
+      const decryptedCert = await kmsDecryptor({
+        cipherTextBlob: certBody.encryptedCertificate
+      });
+
+      // Use helper to parse the decrypted certificate
+      const parsed = parseCertificateBody(decryptedCert);
```
carlosmonastyrski: 🤔 this could be not the best move since we are going to decrypt the certificate for every get we get and it can be an expensive task, what about moving these to columns on the certificates table if we want to show them on the details, just like we do with the others?

saifsmailbox98: Makes sense. We can rely on saving these info in the certificates table and fetch from there.
I've added handling to save these info when a new entry is made in the certificates table. For the existing certs we can backfill them with a migration. But to keep it simple (discussed with Sheen), for now we can return this info from the table if present, otherwise, extract from the cert (also right now we don't add missing data back to the table when we do this, to keep it simple).

Also, I have added individual columns for each property as opposed to a single or group JSONB column. This will help us in better indexing as the inventory grows and users start filtering based on these fields.

backend/src/server/routes/v1/certificate-router.ts
saifsmailbox98: id was being used for generating the url to the linked cert and the common name was for the link's label text

since this api can be used by users directly as well, made sure that there is enough info in this object.

But as you said that the common name remains the same, we can let this API return only 2 IDs (renew from/by) and use the common name of the main cert as the link's label text.

I've removed these and now we only rely on the two IDs

backend/src/db/migrations/20260127124845_add-certificate-parsed-fields.ts
```diff
@@ -0,0 +1,44 @@
+import { Knex } from "knex";
```
carlosmonastyrski, line 1: IGNORE COMMENT PLACEMENT
We don't have an external CA details page, so for external CAs we should not make this clickable
Image

backend/src/db/migrations/20260127124845_add-certificate-parsed-fields.ts
```diff
@@ -0,0 +1,44 @@
...
+
```
carlosmonastyrski, line 2: IGNORE COMMENT PLACEMENT
CRL signing badge looks a bit off
Image

saifsmailbox98: `cRLSign` being converted from camelcase to titlecase caused this issue. Added an exception for `cRLSign` to always be shown as `CRL Sign` (which also matches the policy page)

backend/src/db/migrations/20260127124845_add-certificate-parsed-fields.ts
saifsmailbox98: Added `caType` to the API to figure out whether the CA is internal or external using a join on `internal_certificate_authorities`. Only name of the CA will be shown in case of external CAs, no links.

PR #5307, opened 2026-01-28
feat: dbt app connection & service token secret rotation

varonix0:
## Context

App connection and secret rotation for https://www.getdbt.com

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-fns.ts
+++ b/backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-fns.ts
@@ -116,7 +116,7 @@ export const dbtServiceTokenRotationFactory: TRotationFactory<
     await retrieveDbtAccount(connection).catch(() => {
       throw new BadRequestError({
         message:
-          "Failed to validate connection credentials. Are you sure your configured DBT app connection credentials are valid?"
+          "Failed to validate connection credentials. Check if your configured DBT app connection credentials are valid."
       });
     });
 
--- a/backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-schemas.ts
+++ b/backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-schemas.ts
@@ -57,7 +57,12 @@ export const DbtServiceTokenRotationGeneratedCredentialsSchema = z
   .max(2);
 
 const DbtServiceTokenRotationParametersSchema = z.object({
-  tokenName: z.string().trim().min(1).describe(SecretRotations.PARAMETERS.DBT_SERVICE_TOKEN.tokenName),
+  tokenName: z
+    .string()
+    .trim()
+    .min(1, "Token name is required")
+    .max(255, "Token name must be less than 255 characters")
+    .describe(SecretRotations.PARAMETERS.DBT_SERVICE_TOKEN.tokenName),
   permissionGrants: DbtTokenPermissionsSchema.array()
     .min(1)
     .describe(SecretRotations.PARAMETERS.DBT_SERVICE_TOKEN.permissionGrants)
--- a/frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ParametersFields/DbtServiceTokenRotationParametersFields.tsx
+++ b/frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ParametersFields/DbtServiceTokenRotationParametersFields.tsx
@@ -72,7 +72,7 @@ export const DbtServiceTokenRotationParametersFields = () => {
         name="parameters.tokenName"
         render={({ field, fieldState: { error } }) => (
           <FormControl
-            tooltipText="Enter a name for the service token to be created. Note that each secret will have a suffix of '-1' or '-2' to differentiate between rotation cycles."
+            tooltipText="Enter a name for the service token to be created."
             isError={Boolean(error?.message)}
             errorText={error?.message}
             label="Service Token Name"
--- a/frontend/src/components/secret-rotations-v2/SecretRotationV2Select.tsx
+++ b/frontend/src/components/secret-rotations-v2/SecretRotationV2Select.tsx
@@ -19,29 +19,31 @@ export const SecretRotationV2Select = ({ onSelect }: Props) => {
   }
 
   return (
-    <div className="grid grid-cols-3 gap-2">
-      {secretRotationOptions?.map(({ type }) => {
-        const { image, name, size } = SECRET_ROTATION_MAP[type];
+    <div className="max-h-[60vh] overflow-y-auto">
+      <div className="grid grid-cols-3 gap-2">
+        {secretRotationOptions?.map(({ type }) => {
+          const { image, name, size } = SECRET_ROTATION_MAP[type];
 
-        return (
-          <button
-            type="button"
-            key={type}
-            onClick={() => onSelect(type)}
-            className="group relative flex h-28 cursor-pointer flex-col items-center justify-center rounded-md border border-mineshaft-600 bg-mineshaft-700 p-4 duration-200 hover:bg-mineshaft-600"
-          >
-            <img
-              src={`/images/integrations/${image}`}
-              width={size}
-              className="mt-auto"
-              alt={`${name} logo`}
-            />
-            <div className="mt-auto max-w-xs text-center text-xs font-medium text-gray-300 duration-200 group-hover:text-gray-200">
-              {name}
-            </div>
-          </button>
-        );
-      })}
+          return (
+            <button
+              type="button"
+              key={type}
+              onClick={() => onSelect(type)}
+              className="group relative flex h-28 cursor-pointer flex-col items-center justify-center rounded-md border border-mineshaft-600 bg-mineshaft-700 p-4 duration-200 hover:bg-mineshaft-600"
+            >
+              <img
+                src={`/images/integrations/${image}`}
+                width={size}
+                className="mt-auto"
+                alt={`${name} logo`}
+              />
+              <div className="mt-auto max-w-xs text-center text-xs font-medium text-gray-300 duration-200 group-hover:text-gray-200">
+                {name}
+              </div>
+            </button>
+          );
+        })}
+      </div>
     </div>
   );
 };


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/f9fd9826-2b7a-4245-8df7-8caf7131ff9e?refs=bcd60d7d8c7fb19805df78725230ca98cab28345&source=prChecksComment) |

varonix0:
@greptile review this

docs/snippets/RotationsBrowser.jsx
```diff
@@ -30,6 +30,13 @@ export const RotationsBrowser = () => {
         "Learn how to automatically rotate Databricks service principal OAuth secrets.",
       category: "Cloud Providers",
     },
+    {
+      name: "DBT Service Token",
+      slug: "dbt-service-token",
+      path: "/documentation/platform/secret-rotation/dbt-service-token",
+      description: "Learn how to automatically rotate DBT service tokens.",
+      category: "Cloud Providers",
```
victorvhs017: This should also be `category: "Data Analytics"` like the app connection.

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/DbtConnectionForm.tsx
```diff
@@ -0,0 +1,180 @@
...
+              label="Instance URL"
+            >
+              <Input
+                value={value}
+                onChange={(e) => onChange(e.target.value)}
+                placeholder="https://bw255.us1.dbt.com"
+              />
+            </FormControl>
+          )}
+        />
```
victorvhs017, line 89: This "Instance URL" is called "Access URL" by DBT, and it can be a bit confusing if we don't specify, because they have many URLs:

Image

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ParametersFields/DbtServiceTokenRotationParametersFields.tsx
```diff
@@ -0,0 +1,165 @@
...
+                        ([permissionSet, { label, isEnterpriseOnly }]) => (
+                          <SelectItem key={permissionSet} value={permissionSet}>
+                            <div className="flex items-center gap-2">
+                              {label}
+                              {isEnterpriseOnly && <Badge variant="info">DBT Enterprise</Badge>}
+                            </div>
+                          </SelectItem>
+                        )
+                      )}
+                    </Select>
```
victorvhs017: Is this the expected behavior for the select? (taking the whole screen height)

https://github.com/user-attachments/assets/9beaffcd-3114-4a68-b215-db0bef028f4b

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ParametersFields/DbtServiceTokenRotationParametersFields.tsx
victorvhs017, line 1: The layout breaks when I add too many permissions:

Image

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ReviewFields/DbtServiceTokenRotationReviewFields.tsx
victorvhs017, line 1: The view when we have a lot of permissions is not great here, too:

Image

frontend/src/components/secret-rotations-v2/forms/schemas/dbt-service-token-rotation-schema.ts
victorvhs017, line 1: Can we check using their API if the token for the app connection is from an enterprise account or not?

If so, we could hide the enterprise options and avoid this error. What do you think?

If we can't do that, we should at least update the error message here, because the "Enterprise" is ambiguous and could mean Infisical's Enterprise plan.

Image

docs/documentation/platform/secret-rotation/dbt-service-token.mdx
```diff
@@ -0,0 +1,152 @@
...
+                    "hours": 0,
+                    "minutes": 0
+                },
+                "lastRotationMessage": null,
+                "type": "dbt-service-token",
+                "parameters": {
+                    "permissionGrants": [
+                        {
+                            "permissionSet": "account_admin",
+                            "projectId": "<project-id>, <leave empty for all projects>"
```
victorvhs017: Can we put this as an inline comment? Like:

```suggestion
projectId": "<project-id>" // empty for all projects
```

docs/documentation/platform/secret-rotation/dbt-service-token.mdx
```diff
@@ -0,0 +1,152 @@
...
+            "rotationInterval": 30,
+            "rotateAtUtc": {
+                "hours": 0,
+                "minutes": 0
+            },
+            "parameters": {
+                "permissionGrants": [
+                    {
+                        "permissionSet": "account_admin",
+                        "projectId": "<project-id>, <leave empty for all projects>"
```
victorvhs017: Same here.

backend/src/ee/routes/v2/secret-rotation-v2-routers/dbt-service-token-rotation-router.ts
victorvhs017, line 1: (ignore the placement)

We'll need to add pagination for this screen now.

Image

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ParametersFields/DbtServiceTokenRotationParametersFields.tsx
victorvhs017, line 1: I think the user should be able to add a name to the service token, or at least a prefix, the same way we do in the Open Router Secret Rotation https://infisical.com/docs/documentation/platform/secret-rotation/open-router-api-key

If we don't add a name, and the user has many secret rotations, it's hard to map which one is which in Infisical because we also don't show this name anywhere.

Image

Image

We also give them freedom to keep some naming pattern they have.

victorvhs017, line 1: The Create Service Token endpoint also supports the `writable_environment_categories` field. This gives more control of the access scope to the user, and I believe we should add it:

Image

The example in the API body is pretty bad with the "string" there, but it should be the environments that the token should be able to write, based on their [Terraform docs](https://registry.terraform.io/providers/dbt-labs/dbtcloud/latest/docs/resources/group#example-usage):

Image

backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-fns.ts
```diff
@@ -0,0 +1,223 @@
...
+        {
+          headers: {
+            Accept: "application/json",
+            Authorization: `Bearer ${connection.credentials.apiToken}`
+          }
+        }
+      );
+
+      if (!data.data.token_string || !data.data.id) {
+        logger.error(data || "Failed to create service token");
```
victorvhs017: If, for some reason, the response doesn't have an ID, but it has the token_string, this log would expose it.

We could log only the status field of the response:

Image

backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-fns.ts
```diff
@@ -0,0 +1,223 @@
...
+        {
+          headers: {
+            Accept: "application/json",
+            Authorization: `Bearer ${connection.credentials.apiToken}`
+          }
+        }
+      );
+
+      if (!data.status.is_success) {
+        throw new Error();
```
victorvhs017: nit: we could just return `false` directly here.

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ReviewFields/DbtServiceTokenRotationReviewFields.tsx
victorvhs017, line 1: When we use an app connection with an invalid token, it doesn't show any error notifications that it couldn't fetch the projects; it only shows an empty select:

Image

victorvhs017, line 1: If I delete the rotation with "Revoke Credentials" = true, and my app connection has an invalid token, the operation is successful, even if the service tokens are not deleted. 

Shouldn't we return an error for the whole secret rotation deletion in this case?

Image

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/DbtConnectionForm.tsx
varonix0: Main purpose of this was to stay in line with the other app connections, which also specify this as Instance URL. I think in terms of standardization here, but curious on your thoughts

frontend/src/components/secret-rotations-v2/forms/schemas/dbt-service-token-rotation-schema.ts
varonix0: I don't think so, and I think it's pretty clear when you select the permission set in the previous steps (says DBT Enterprise on each individual one)

backend/src/ee/routes/v2/secret-rotation-v2-routers/dbt-service-token-rotation-router.ts
varonix0: Not how it looks on my end, but I may be on a bigger monitor than you. I think we can wait a few rotations. If we add pagination now there's just nothing on the other pages. We can save this for a later time. For now I'll remove the "Coming soon" part as our rotations are pretty well fletched out at this point

CleanShot 2026-01-31 at 03 45 33

backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-fns.ts
varonix0: If is_success is false, then the request will throw a non-200 error - removing this as it will be handled by `createDbtError`

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ReviewFields/DbtServiceTokenRotationReviewFields.tsx
varonix0: This is the case for all app connection usages

Only way this would happen is if a user set up a DBT app connection with a valid token and then removed the token later

backend/src/ee/routes/v2/secret-rotation-v2-routers/dbt-service-token-rotation-router.ts
victorvhs017: I think it should behave like the app connection modal:

https://github.com/user-attachments/assets/28302d25-fab3-432d-bbe2-675c93a8ca2b

It may be out of scope for this now, but let's at least add a new issue to keep track of it!

backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-schemas.ts
```diff
@@ -57,6 +57,7 @@ export const DbtServiceTokenRotationGeneratedCredentialsSchema = z
   .max(2);
 
 const DbtServiceTokenRotationParametersSchema = z.object({
+  tokenName: z.string().trim().min(1).describe(SecretRotations.PARAMETERS.DBT_SERVICE_TOKEN.tokenName),
```
victorvhs017: We should put a max here, too. A name too long can cause errors in the dbt API:

Image

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ParametersFields/DbtServiceTokenRotationParametersFields.tsx
victorvhs017, line 1: We have this information in the name tooltip, but I don't see the suffix

Image

Image

backend/src/ee/services/secret-rotation-v2/dbt-service-token/dbt-service-token-rotation-fns.ts
```diff
@@ -125,6 +113,13 @@ export const dbtServiceTokenRotationFactory: TRotationFactory<
    * First checks if the service token exists before attempting revocation.
    */
   const revokeCredential = async (tokenId: number) => {
+    await retrieveDbtAccount(connection).catch(() => {
+      throw new BadRequestError({
+        message:
+          "Failed to validate connection credentials. Are you sure your configured DBT app connection credentials are valid?"
```
victorvhs017: nit: I think we should avoid the question here, maybe:

"Failed to validate connection credentials. Check if your configured DBT app connection credentials are valid."

Or something like that.

frontend/src/components/secret-rotations-v2/forms/SecretRotationV2ParametersFields/DbtServiceTokenRotationParametersFields.tsx
varonix0: Ah sorry this is old, fixed!

PR #5327, opened 2026-01-30
feat(pki): add webhooks to alerts

saifsmailbox98:
## Context

This adds webhook support for PKI alerts, allowing users to receive notifications before their certificates expire based on custom time thresholds.

## Screenshots

CleanShot 2026-01-31 at 00 49 20@2x

CleanShot 2026-01-31 at 00 46 41@2x

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/pki-alert-v2/pki-alert-v2-types.ts
+++ b/backend/src/services/pki-alert-v2/pki-alert-v2-types.ts
@@ -177,7 +177,11 @@ export const CreatePkiAlertV2Schema = z.object({
   alertBefore: z.string().refine(createSecureAlertBeforeValidator(), "Must be in format like '30d', '1w', '3m', '1y'"),
   filters: PkiFiltersSchema,
   enabled: z.boolean().default(true),
-  channels: z.array(CreateChannelSchema).min(1, "At least one notification channel is required").max(10)
+  channels: z
+    .array(CreateChannelSchema)
+    .min(1, "At least one notification channel is required")
+    .max(10)
+    .refine((channels) => channels.some((ch) => ch.enabled), "At least one notification channel must be enabled")
 });
 
 export type TCreatePkiAlertV2 = z.infer<typeof CreatePkiAlertV2Schema>;
--- a/frontend/src/hooks/api/pkiAlertsV2/types.ts
+++ b/frontend/src/hooks/api/pkiAlertsV2/types.ts
@@ -236,7 +236,13 @@ export const createPkiAlertV2Schema = z.object({
     .optional(),
   filters: z.array(pkiFilterRuleV2Schema),
   enabled: z.boolean().default(true),
-  channels: z.array(pkiAlertChannelV2Schema).min(1, "At least one notification channel is required")
+  channels: z
+    .array(pkiAlertChannelV2Schema)
+    .min(1, "At least one notification channel is required")
+    .refine(
+      (channels) => channels.some((ch) => ch.enabled),
+      "At least one notification channel must be enabled"
+    )
 });
 
 export const updatePkiAlertV2Schema = createPkiAlertV2Schema.partial().omit({ projectId: true });
--- a/frontend/src/views/PkiAlertsV2Page/components/CreatePkiAlertV2FormSteps.tsx
+++ b/frontend/src/views/PkiAlertsV2Page/components/CreatePkiAlertV2FormSteps.tsx
@@ -174,6 +174,7 @@ export const CreatePkiAlertV2FormSteps = ({
       config: type === PkiAlertChannelTypeV2.EMAIL ? { recipients: [] } : { url: "" },
       enabled: true
     });
+    trigger("channels");
   };
 
   // Auto-expand newly added channel (prepended at index 0)
@@ -263,9 +264,6 @@ export const CreatePkiAlertV2FormSteps = ({
   const deleteChannel = (fieldId: string, index: number, e: React.MouseEvent) => {
     e.stopPropagation();
 
-    // Prevent deleting the last channel
-    if (channelFields.length <= 1) return;
-
     removeChannel(index);
 
     // Clean up unified channel state
@@ -279,8 +277,6 @@ export const CreatePkiAlertV2FormSteps = ({
       setExpandedChannel(undefined);
     }
   };
-
-  const isLastChannel = channelFields.length === 1;
 
   const getFieldOperators = (field: PkiFilterFieldV2) => {
     if (field === PkiFilterFieldV2.INCLUDE_CAS) {
@@ -656,7 +652,7 @@ export const CreatePkiAlertV2FormSteps = ({
             </DropdownMenu>
           </div>
 
-          {channelFields.length > 0 ? (
+          {channelFields.length > 0 && (
             <Accordion
               type="single"
               collapsible
@@ -696,25 +692,22 @@ export const CreatePkiAlertV2FormSteps = ({
                               <Switch
                                 id={`channel-enabled-${index}`}
                                 isChecked={enabledField.value}
-                                onCheckedChange={enabledField.onChange}
+                                onCheckedChange={(checked) => {
+                                  enabledField.onChange(checked);
+                                  trigger("channels");
+                                }}
                               />
                             )}
                           />
-                          <Tooltip
-                            content="At least one channel is required"
-                            isDisabled={!isLastChannel}
+                          <IconButton
+                            size="sm"
+                            variant="plain"
+                            colorSchema="danger"
+                            ariaLabel="Delete channel"
+                            onClick={(e) => deleteChannel(field.id, index, e)}
                           >
-                            <IconButton
-                              size="sm"
-                              variant="plain"
-                              colorSchema="danger"
-                              ariaLabel="Delete channel"
-                              isDisabled={isLastChannel}
-                              onClick={(e) => deleteChannel(field.id, index, e)}
-                            >
-                              <FontAwesomeIcon icon={faTrash} />
-                            </IconButton>
-                          </Tooltip>
+                            <FontAwesomeIcon icon={faTrash} />
+                          </IconButton>
                         </div>
                       </div>
                     </AccordionTrigger>
@@ -836,7 +829,13 @@ export const CreatePkiAlertV2FormSteps = ({
                 );
               })}
             </Accordion>
-          ) : (
+          )}
+
+          {channelFields.length > 0 && errors.channels?.message && (
+            <span className="text-sm text-red-500">{errors.channels.message}</span>
+          )}
+
+          {channelFields.length === 0 && (
             <div className="flex flex-1 flex-col items-center justify-center gap-2">
               <span className="text-bunker-400">
                 At least one notification channel is required. Click &quot;Add Channel&quot; to add


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/ff620000-90b2-482c-ae61-59117015d1bb?refs=5595263a439ac3a1b96ab55e98fe546721b0ae29&source=prChecksComment) |

backend/src/services/pki-alert-v2/pki-alert-v2-service.ts
```diff
@@ -423,23 +557,73 @@ export const pkiAlertV2ServiceFactory = ({
...
+          friendlyName: cert.commonName,
+          serialNumber: cert.serialNumber,
+          expiryDate: cert.notAfter.toLocaleDateString()
+        }))
+      },
+      template: SmtpTemplates.PkiExpirationAlert
+    });
+  };
+
+  const sendWebhookNotification = async (
```
carlosmonastyrski: Remember to use here the blockLocalAndPrivateIpAddresses function

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
```diff
@@ -0,0 +1,37 @@
+import { Knex } from "knex";
```
carlosmonastyrski, line 1: IGNORE COMMENT PLACEMENT

If I create/edit an alert to not have any channels at all, shouldn't we mark it as disabled by default? and only allow it to be active again only if at least one channel is set?

backend/src/server/routes/v1/pki-alert-router.ts
```diff
@@ -125,6 +125,13 @@ export const registerPkiAlertRouter = async (server: FastifyZodProvider) => {
                   updatedAt: z.date()
                 })
               ),
+              lastRun: z
+                .object({
+                  timestamp: z.date(),
+                  status: z.enum(["success", "failed"]),
```
carlosmonastyrski: let's use enums for these

backend/src/server/routes/v2/pki-alert-router.ts
```diff
@@ -123,6 +123,13 @@ export const registerPkiAlertRouter = async (server: FastifyZodProvider) => {
                   updatedAt: z.date()
                 })
               ),
+              lastRun: z
+                .object({
+                  timestamp: z.date(),
+                  status: z.enum(["success", "failed"]),
```
carlosmonastyrski: same here

backend/src/server/routes/v2/pki-alert-router.ts
```diff
@@ -447,4 +454,46 @@ export const registerPkiAlertRouter = async (server: FastifyZodProvider) => {
       return result;
     }
   });
+
+  server.route({
```
carlosmonastyrski, line 459: Let's add an audit log for this one 🙏

backend/src/services/pki-alert-v2/pki-alert-v2-notification-fns.ts
```diff
@@ -0,0 +1,231 @@
+import crypto from "node:crypto";
+
+import { AxiosError } from "axios";
+
+import { request } from "@app/lib/config/request";
+import { logger } from "@app/lib/logger";
+
+import { TCertificatePreview } from "./pki-alert-v2-types";
+
+const PKI_WEBHOOK_TIMEOUT = 2 * 1000;
```
carlosmonastyrski: why are we doing a 2 second timeout? isn't it a bit short for possible web latency? 🤔 curious if we checked what is the standard for webhooks

backend/src/services/pki-alert-v2/pki-alert-v2-notification-fns.ts
```diff
@@ -0,0 +1,231 @@
...
+
+export enum PkiWebhookEventType {
```
carlosmonastyrski: is it possible to move these enums and types to `pki-alert-v2-types.ts`? so we can track types directly on that file

docs/documentation/platform/pki/alerting.mdx
```diff
@@ -15,11 +15,108 @@ To create an alert, head to your Certificate Management Project > Alerting and p
...
+
 Here's some guidance for each field in the alert configuration sequence:
 
-- Alert Type: The type of alert to create such as **Certificate Expiration**.
-- Alert Name: A slug-friendly name for the alert such as `tls-expiry-alert`.
-- Description: An optional description for the alert.
-- Alert Before: The time before certificate expiration to trigger the alert such as 30 days denoted by `30d`.
-- Filters: A list of filters that determine which certificates the alert applies to. Each row includes a **Field**, **Operator**, and **Value** to match against. For example, you can filter for certificates with a common name containing `example.com` by setting the field to **Common Name**, the operator to **Contains**, and the value to `example.com`.
-- Channels / Email Recipients: A list of email addresses to notify when the alert triggers.
+- **Alert Type**: The type of alert to create such as **Certificate Expiration**.
```
carlosmonastyrski, line 22: I think we should add a note or a section here explaining what happens if one of the channel fails, and how it'll be triggered until all channels are correctly managed.

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
```diff
@@ -0,0 +1,37 @@
...
+
```
carlosmonastyrski, line 2: IGNORE COMMENT PLACEMENT

If it's not too hard to add, I think it'll be a huge improvement if we could alert people on the notifications if one of the channels is broken

Image

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
```diff
@@ -0,0 +1,37 @@
...
+import { TableName } from "../schemas";
```
carlosmonastyrski, line 3: IGNORE COMMENT PLACEMENT
If the error message is long enough it can overflow the view and not show the full content of it. I think it's either making this tooltip scrollable or maybe clickable to show the full message in a modal?

Image

carlosmonastyrski: Also, should we limit the amount of channels available on an alert? as it can be quite risky if it's set to a huge number of APIs 🤔

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
```diff
@@ -0,0 +1,37 @@
...
+
```
carlosmonastyrski, line 4: IGNORE COMMENT PLACEMENT
Small nit, we are truncating when there is enough space to show the full url

Image

backend/src/server/routes/v2/pki-alert-router.ts
```diff
@@ -447,4 +454,46 @@ export const registerPkiAlertRouter = async (server: FastifyZodProvider) => {
...
+    method: "POST",
+    url: "/test-webhook",
```
carlosmonastyrski, line 461: any thoughts on having an structure like "/{channel-type}/test" instead? so it's scalable if we need to add it for checks on other types like Slack

carlosmonastyrski (changes requested):

backend/src/services/pki-alert-v2/pki-alert-v2-service.ts
saifsmailbox98: Done - replaced verifyHostInputValidity with blockLocalAndPrivateIpAddresses

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
saifsmailbox98: I think it's better to keep `enabled` user-controlled to avoid unexpected side effects. An alert with no channels is harmless. Users might create an alert first and add channels later, or temporarily remove channels and expect the alert to stay enabled when they add them back. We can also optimize the query to skip channel-less alerts later as volume grows, but the improvement would be insignificant.

backend/src/server/routes/v1/pki-alert-router.ts
saifsmailbox98: Done - added `PkiAlertRunStatus` enum and updated both v1 and v2 routers to use `z.nativeEnum(PkiAlertRunStatus)`

backend/src/server/routes/v2/pki-alert-router.ts
saifsmailbox98: Done

backend/src/services/pki-alert-v2/pki-alert-v2-notification-fns.ts
saifsmailbox98: Done - moved to `pki-alert-v2-types.ts`

backend/src/server/routes/v2/pki-alert-router.ts
saifsmailbox98: I think we can skip the audit log here since it's a test operation that doesn't modify any state. The actual webhook configuration is audited when the alert is created/updated. We also don't intend to show test operation logs to users.

docs/documentation/platform/pki/alerting.mdx
saifsmailbox98: Added a note explaining that all channels must succeed for the alert to complete, and that it will retry until all channels work.

backend/src/server/routes/v2/pki-alert-router.ts
saifsmailbox98: I'm not planning to add a test feature for Slack. Webhook testing makes sense because users need to verify connectivity, payload format, HMAC verification, and any integrations they're building on top of it. For Slack, users just provide a URL and it works - there's no HMAC verification, no custom parsing needed (messages are human-readable), and they can't build automated integrations based on the message format. I can't think of another channel type that would need this kind of testing.

backend/src/services/pki-alert-v2/pki-alert-v2-notification-fns.ts
saifsmailbox98: Industry standard is usually 10-20 seconds, but we kept it shorter based on our earlier discussion with Sheen - since we retry all channels when any one fails, a shorter timeout avoids wasting resources. That said, 2 seconds felt too aggressive and could lead to false timeout failures, so I've increased it to 7 seconds as a middle ground.

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
saifsmailbox98: Added max-height with thin-scrollbar to make long error messages scrollable.

CleanShot 2026-02-03 at 17 38 10@2x

saifsmailbox98: Added a limit of 10 channels per alert with frontend validation and docs.

CleanShot 2026-02-03 at 17 50 30@2x

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
saifsmailbox98: Fixed it.

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
saifsmailbox98: Added in-app notifications for project admins

CleanShot 2026-02-03 at 18 48 00@2x

saifsmailbox98:
@carlosmonastyrski 

Addressed the changes requested:
  1. Made having atleast one channel mandatory.
  2. Added retries.

Also refactored the code a bit.

backend/src/db/migrations/20260129181800_add-encrypted-config-to-pki-alert-channels.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function up(knex: Knex): Promise<void> {
+  if (await knex.schema.hasTable(TableName.PkiAlertChannels)) {
```
carlosmonastyrski, line 6: Looking really great! Really nice improvements on the UI too 🤟. Just found one thing where I'm still able to disable the last channel which will be inconsistent with alerts having at least one enabled channel (as otherwise the channel can be disabled entirely instead)

Image

saifsmailbox98: Ok, that was intentional though.

Now, I've added a check so that an alert will always require atleast one enabled channel.
Also, removed the tooltip from the delete, errors will be shown when they try to press the next button, and also when they toggle the last enabled channel disabled.

PR #5380, opened 2026-02-05
feat: implement durable queue for redis queues that needs persistence in PG

akhilmhdh:
## Context

This PR implements postgres backup and recovery for redis queues that needs persistence over wipe out. Some queues that need are
1. Dynamic secret leasing
2. PAM session expiration

This PR completely removes pg_boss dependency usage and migrated back to redis queue. 

There is two new queue jobs
1. Recovery job: That run on container bootup. This checks all pending/failed jobs are present in redis or not. Failed ones are checked in order to retry.
2. Reconcile job is the one used to stop a job and re-queue if a job takes long. As of now there is no queue that needs that long. 

All the bullmq cron job has a key parameter. This ensure we alter the timing window later.

## Steps to verify the change

1. Create a dynamic secret and generate a lease
2. Wipe out redis and bootup again
3. The lease removal job should work fine
4. Like wise for pam session expiration as well

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/db/schemas/queue-jobs.ts
+++ b/backend/src/db/schemas/queue-jobs.ts
@@ -10,7 +10,7 @@ import { TImmutableDBKeys } from "./models";
 export const QueueJobsSchema = z.object({
   id: z.string().uuid(),
   queueName: z.string(),
-  queueType: z.string(),
+  queueType: z.string().default("bullmq"),
   queueJobName: z.string(),
   jobId: z.string(),
   queueData: z.unknown(),


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/0a14af4b-83e5-44ca-8d3e-7a1cbaa9cb63?refs=3787876d2367c7406e05b42e025756da80c89d1f&source=prChecksComment) |

backend/src/queue/queue-service.ts
```diff
@@ -649,23 +924,28 @@ export const queueServiceFactory = (
...
+        queueName: name,
+        queueType: "bullmq",
+        queueJobName: job,
+        jobId,
+        queueData: data,
+        queueOptions: finalOptions,
+        status: PersistenceQueueStatus.Pending,
+        maxAttempts: opts.attempts || 1
+      });
+    }
```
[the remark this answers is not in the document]
akhilmhdh: Our assumption is other way around. Ensure it's in postgres before adding to redi. But yes we can wrap it in a transaction

backend/src/queue/queue-service.ts
```diff
@@ -529,111 +553,362 @@ export const queueServiceFactory = (
...
+
+      // Remove from Redis if exists
+      try {
+        const redisJob = await q.getJob(dbJob.jobId);
+        if (redisJob) await redisJob.remove();
+      } catch {
+        // Ignore removal errors
+      }
+      return;
+    }
```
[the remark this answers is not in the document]
akhilmhdh: On failure we have notification

backend/src/queue/queue-service.ts
```diff
@@ -529,111 +553,362 @@ export const queueServiceFactory = (
...
+          { jobId: job.jobId, queueName, adjustedDelay, wasFailedJob: isFailedJob },
+          "Job recovered and re-queued"
+        );
+      }
+
+      logger.info("Queue startup recovery completed");
+    } catch (error) {
+      logger.error(error, "Queue startup recovery failed");
+    }
+  };
```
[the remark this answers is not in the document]
akhilmhdh: Addressed

backend/src/db/migrations/20260203141935_durable-queue.ts
```diff
@@ -0,0 +1,177 @@
...
+            queueType: "bullmq",
+            queueData: job.data,
+            jobId,
+            queueOptions
+          };
+        })
+      );
+
+      totalMigrated += jobCount;
+      offset += BATCH_SIZE;
```
fangpenlin: it seems like there are some potential issues in the migration code.

first would be, we are using a pagination iterating over based on `created_on` column. If there's a job with the queue / job name we are looking, and its `created_on` value is set to a value not always increasing (like get a now() in the past but commit a bit later), then it seems like you may not get a consistent view on all the existing jobs (the pagination boundary will change).

also, for deploying process, we may have rolling update. while this db migration code got deployed, converting all the existing jobs into new format, but the older version may still create more legacy job entries.

migration like this could be very tricky when dealing with old code / new code. multiple phase rolling out might be needed

backend/src/ee/services/secret-rotation/secret-rotation-queue/secret-rotation-queue.ts
```diff
@@ -83,24 +83,7 @@ export const secretRotationQueueFactory = ({
...
-          // on prod it this will be in days, in development this will be second
-          every: appCfg.NODE_ENV === "development" ? secondsToMillis(interval) : daysToMillisecond(interval),
-          immediately: true
-        },
-        removeOnComplete: true,
-        removeOnFail: true
-      }
-    );
-  };
+  const addToQueue = async () => {};
```
fangpenlin, line 86: seems like now this func does nothing? then keep it around if it does nothing?

backend/src/ee/services/secret-rotation/secret-rotation-service.ts
```diff
@@ -205,7 +205,7 @@ export const secretRotationServiceFactory = ({
       }
       return { ...doc, outputs: outputSecretMapping, environment: folder.environment };
     });
-    await secretRotationQueue.addToQueue(secretRotation.id, secretRotation.interval);
+    await secretRotationQueue.addToQueue();
```
fangpenlin, line 208: after a rotation is created, now we are making this `addToQueue` method. but then this `addToQueue` actually does nothing now (noop).

as a result, we are not queuing a task for secret rotation? why is that? do we have code in other places which will queue a task for secret rotation instead?

backend/src/queue/queue-service.ts
```diff
@@ -516,11 +541,12 @@ export type TQueueServiceFactory = {
     startOffset?: number,
     endOffset?: number
   ) => Promise<{ delay: number; timestamp: number; repeatJobKey?: string; data?: unknown }[]>;
+  updateJobHeartbeat: <T extends QueueName>(queueName: T, jobId: string) => Promise<void>;
```
fangpenlin, line 544: while this is provided, but I don't see anywhere it's used. relying on developer to know what task are long running seems to be a bit tedious and error prone. I think there will be many bugs down the road in the production as job get prematurely killed. have you considered making the heartbeat automatic while the job handler function is running?

https://github.com/Infisical/infisical/blob/82903f6c16dfbb75f72ca12e0e9da9f81c3c3d11/backend/src/auto-start-migrations.ts#L282-L292

I did similar thing for our migration startup process to avoid race condition.

backend/src/ee/services/secret-rotation/secret-rotation-service.ts
akhilmhdh: This is secret rotation v1 and it's deprecated and not used anymore.

backend/src/ee/services/secret-rotation/secret-rotation-queue/secret-rotation-queue.ts
akhilmhdh: This is secret rotation v1 and it's deprecated and not used anymore. Nobody can use this service

backend/src/queue/queue-service.ts
akhilmhdh: Makes sense! I have added auto heartbeat mechanism now for persistant jobs.

PR #5427, opened 2026-02-10
feat(secret-sharing): cleaned up API and exposed publicly 

varonix0:
## Context

I've cleaned up the secret sharing API and added identity support. I haven't done much work on the requests side of things. If we expose the secret requests router, we can clean that one up more as well. The requests router is in much better shape than the secret sharing router was prior to this PR though.

#### Route Changes

Before | After | Notes
-- | -- | --
GET /api/v1/secret-sharing/shared/ | GET /api/v1/secret-sharing/ | Removed redundant /shared prefix
POST /api/v1/secret-sharing/shared/public/:id | POST /api/v1/secret-sharing/:id/access | Changed from POST for read to action-based endpoint. Removed /public segment.
(none) | GET /api/v1/secret-sharing/:id | New endpoint - Fetches secret metadata without revealing the value. Enables two-step access flow. Before you would hit the (now) `/access` endpoint, which would give you an error and an `isPasswordProtected` field. Then you knew you need to re-hit the same API endpoint with a password. Not restful.
POST /api/v1/secret-sharing/shared/ | POST /api/v1/secret-sharing/ | Removed /shared prefix
DELETE /api/v1/secret-sharing/shared/:id | DELETE /api/v1/secret-sharing/:id | Removed /shared prefix

#### Behavioral Changes

`GET /api/v1/secret-sharing/:id` (New)
* Returns full secret metadata (without secret value)
* Auth logic: No auth for accessType: "anyone", requires auth + org check for accessType: "organization"
* Enables frontend to determine if password prompt is needed before attempting access

`POST /:id/access`
* Renamed from POST `/public/:id` for clarity (it's an action, not a resource read)
* Removed `brandingConfig` from response (now fetched separately via `GET /branding?sharedSecretId=...`)
* Removed error field from response (errors now use proper HTTP status codes)

#### Response Schema Changes
Before (`POST /public/:id`):

```javascript
{
  isPasswordProtected: boolean;
  secret?: {
    secretValue?: string;
    // other fields
  };
  brandingConfig?: {}; // now removed
  error?: string;           // now removed
}
```

After (`GET /:id`):

```javascript
{
  id: string;
  name: string | null;
  accessType: "organization" | "anyone";
  expiresAt: string;
  expiresAfterViews: number | null;
  isPasswordProtected: boolean;
  // ... full sanitized secret (no secret value)
}
```

After (`POST /:id/access`) - new endpoint:
```javascript
{
  id: string;
  name: string | null;
  accessType: "organization" | "anyone";
  expiresAt: string;
  secretValue?: string;      // The actual secret
  orgName?: string;
  // ... full sanitized secret
}
```

#### Frontend Flow Change
Before:
1. `POST /public/:id`: Returns everything in one call (metadata + secret value + branding). Would be called multiple times in the frontend for password protected secrets.
After:
1. `GET /:id`: Get metadata (is it password protected? what access type?)
2. If not password protected: automatically call `POST /:id/access`
3. If password protected: show password prompt,  then call `POST /:id/access` with password
4. `GET /branding?sharedSecretId=...`, fetch branding separately

#### Breaking changes

In the past, secret sharing was end-to-end encrypted. We used to look up secrets using a combination of the secret's UUIDv4 ID and a hashed hex. This is over 14 months old, and I've now removed this logic from the API. This means that shared secrets that are over 14 months old will no longer work. This is not a problem, as secrets are set to expire at the latest after 30 days.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/secret-sharing-router.ts
+++ b/backend/src/server/routes/v1/secret-sharing-router.ts
@@ -96,6 +96,8 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
       response: {
         200: SanitizedSecretSharingSchema.extend({
           isPasswordProtected: z.boolean().describe("Whether the shared secret is protected by a password.")
+        }).omit({
+          authorizedEmails: true
         })
       }
     },
@@ -234,7 +236,7 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
         .object({
           name: z.string().max(50).optional().describe(SECRET_SHARING.CREATE.name),
           password: z.string().optional().describe(SECRET_SHARING.CREATE.password),
-          secretValue: z.string().describe(SECRET_SHARING.CREATE.secretValue),
+          secretValue: z.string().max(10_000).describe(SECRET_SHARING.CREATE.secretValue),
           expiresIn: z.string().default("30d").describe(SECRET_SHARING.CREATE.expiresIn),
           maxViews: z.number().min(1).optional().describe(SECRET_SHARING.CREATE.maxViews),
           accessType: z
--- a/backend/src/services/secret-sharing/secret-sharing-service.ts
+++ b/backend/src/services/secret-sharing/secret-sharing-service.ts
@@ -4,6 +4,7 @@ import { OrganizationActionScope, TOrganizations, TSecretSharing } from "@app/db
 import { TLicenseServiceFactory } from "@app/ee/services/license/license-service";
 import { OrgPermissionSecretShareAction, OrgPermissionSubjects } from "@app/ee/services/permission/org-permission";
 import { TPermissionServiceFactory } from "@app/ee/services/permission/permission-service-types";
+import { PgSqlLock } from "@app/keystore/keystore";
 import { getConfig } from "@app/lib/config/env";
 import { crypto } from "@app/lib/crypto/cryptography";
 import { BadRequestError, ForbiddenRequestError, NotFoundError, UnauthorizedError } from "@app/lib/errors";
@@ -12,6 +13,7 @@ import { ms } from "@app/lib/ms";
 import { OrgServiceActor, SecretSharingAccessType } from "@app/lib/types";
 
 import { ActorType } from "../auth/auth-type";
+import { TIdentityDALFactory } from "../identity/identity-dal";
 import { TKmsServiceFactory } from "../kms/kms-service";
 import { TOrgDALFactory } from "../org/org-dal";
 import { TSecretShareBrandConfig } from "../org/org-types";
@@ -41,6 +43,7 @@ type TSecretSharingServiceFactoryDep = {
   kmsService: TKmsServiceFactory;
   smtpService: TSmtpService;
   licenseService: Pick<TLicenseServiceFactory, "getPlan">;
+  identityDAL: Pick<TIdentityDALFactory, "findById">;
 };
 
 export type TSecretSharingServiceFactory = ReturnType<typeof secretSharingServiceFactory>;
@@ -55,7 +58,7 @@ const mapIdentifierToId = (sharedSecret: TSecretSharing): Omit<TSecretSharing, "
 
   if (!identifier) {
     throw new BadRequestError({
-      message: `Shared secret created at ${sharedSecret.createdAt.toISOString()} has no identifier`
+      message: `Shared secret with ID ${sharedSecret.id} has no identifier`
     });
   }
 
@@ -70,6 +73,7 @@ export const secretSharingServiceFactory = ({
   secretSharingDAL,
   orgAssetDAL,
   orgDAL,
+  identityDAL,
   kmsService,
   smtpService,
   userDAL,
@@ -202,23 +206,37 @@ export const secretSharingServiceFactory = ({
 
     // Loop through recipients and send out emails with unique access links
     if (authorizedEmails) {
-      const user = await userDAL.findById(actorId);
+      let displayUsername: string | undefined;
 
-      if (!user) {
-        throw new NotFoundError({ message: `User with ID '${actorId}' not found` });
+      if (actor === ActorType.USER) {
+        const user = await userDAL.findById(actorId);
+
+        if (!user) {
+          throw new NotFoundError({ message: `User with ID '${actorId}' not found` });
+        }
+
+        displayUsername = user.username;
+      } else if (actor === ActorType.IDENTITY) {
+        const identity = await identityDAL.findById(actorId);
+
+        if (!identity) {
+          throw new NotFoundError({ message: `Identity with ID '${actorId}' not found` });
+        }
+
+        displayUsername = `${identity.name} (Machine Identity)`;
       }
 
       for await (const email of authorizedEmails) {
         try {
           // Only show the username to emails which are part of the organization
-          const respondentUsername = orgEmails.includes(email) ? user.username : undefined;
+          const senderUsername = orgEmails.includes(email) ? displayUsername : undefined;
 
           await smtpService.sendMail({
             recipients: [email],
             subjectLine: "A secret has been shared with you",
             substitutions: {
               name,
-              respondentUsername,
+              senderUsername,
               secretRequestUrl: `${appCfg.SITE_URL}/shared/secret/${mappedSharedSecret.id}`
             },
             template: SmtpTemplates.SecretRequestCompleted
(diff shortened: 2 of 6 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/45bbf493-c3d8-4c9c-993f-2b00b5efbf89?refs=0283aca8017718c9dbe5db1a34da07c8dcd7e5cb&source=prChecksComment) |

varonix0:
@greptile why did you stop reviewing the PR? review my PR!

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -632,37 +695,53 @@ export const secretSharingServiceFactory = ({
...
+    }
+    if (actor === ActorType.IDENTITY) {
+      if (sharedSecret.identityId !== actorId) {
+        throw new ForbiddenRequestError({
+          message: "Identity does not have permission to delete shared secret"
+        });
+      }
+    } else {
+      throw new ForbiddenRequestError({ message: "User does not have permission to delete shared secret" });
+    }
```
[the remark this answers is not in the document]
varonix0: This is not true. The route only allows for identity and user auth, there can never be another case where the actor is not a user or an identity... I've improved the if statement though.

backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -194,29 +226,64 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
...
+            .default(SecretSharingAccessType.Organization)
+            .describe(SECRET_SHARING.CREATE.accessType),
+          authorizedEmails: z
+            .string()
+            .email()
+            .array()
+            .max(100)
+            .optional()
+            .transform((val) => (val ? [...new Set(val)] : undefined))
+            .describe(SECRET_SHARING.CREATE.authorizedEmails)
```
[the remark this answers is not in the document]
varonix0: Good catch

varonix0:
@greptile re-review PR with my comments to your comments in mind and update your summary

backend/src/db/schemas/secret-sharing.ts
```diff
@@ -28,7 +28,8 @@ export const SecretSharingSchema = z.object({
   encryptedSecret: zodBuffer.nullable().optional(),
   identifier: z.string().nullable().optional(),
   type: z.string().default("share"),
-  authorizedEmails: z.unknown().nullable().optional()
+  authorizedEmails: z.unknown().nullable().optional(),
```
maidul98, line 31: why is this unknown? does zod not have a email type?

maidul98 (commented):
`secret-sharing` sounds odd as an API. i think it would make more sense for it to be:

- /api/v1/shared-secrets/
- GET /api/v1/shared-secrets/:id
- POST /api/v1/shared-secrets/:id/access
- DELETE /api/v1/shared-secrets/:id

also what happens to all of the old share links right before this change? wil we handle redirects?

backend/src/db/schemas/secret-sharing.ts
varonix0: Arrays are infered as unknown, a quirk of our type generation

varonix0: Ideally, this should've been a table of its own, to store which users/identities are allowed to access the secret, rather than just storing an arbitrary email

backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -71,90 +78,86 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
     }
   });
 
+  server.route({
```
maidul98, line 81: q: right now if the share is for anyone, it will still spit out the authorized emails field. if someone gets a hold of these, can find a bunch of peoples email. any thoughts on this?

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -170,24 +190,25 @@ export const secretSharingServiceFactory = ({
...
+      authorizedEmails: authorizedEmails && authorizedEmails.length > 0 ? JSON.stringify(authorizedEmails) : undefined
     });
 
-    const idToReturn = `${Buffer.from(newSharedSecret.identifier!, "hex").toString("base64url")}`;
+    const mappedSharedSecret = mapIdentifierToId(newSharedSecret);
 
     // Loop through recipients and send out emails with unique access links
-    if (emails) {
+    if (authorizedEmails) {
       const user = await userDAL.findById(actorId);
```
maidul98: When actor type is machine identity, this should result in an error right?

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -170,24 +190,25 @@ export const secretSharingServiceFactory = ({
...
 
       if (!user) {
         throw new NotFoundError({ message: `User with ID '${actorId}' not found` });
       }
 
-      for await (const email of emails) {
+      for await (const email of authorizedEmails) {
         try {
           // Only show the username to emails which are part of the organization
           const respondentUsername = orgEmails.includes(email) ? user.username : undefined;
```
maidul98: nit: this is a weird var name. should be called sender name or something?

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -552,33 +625,21 @@ export const secretSharingServiceFactory = ({
...
       await secretSharingDAL.softDeleteById(sharedSecret.id);
-      return {
-        isPasswordProtected: false,
-        secretOrgId: sharedSecret.orgId,
-        error: "Unable to view secret: secret has expired"
-      };
+      throw new NotFoundError({ message: "The shared secret has expired" });
     }
 
     if (expiresAfterViews !== null && expiresAfterViews === 0) {
```
maidul98: for a secret with maxViews: 1, concurrent requests could result in multiple successful accesses. this seems similar to our project limit issue where people could create more than allowed

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -44,6 +45,26 @@ type TSecretSharingServiceFactoryDep = {
...
+  // secret requests don't use identifiers and are fully ID dependant. so for secret requests we just return it as-is
+  if (sharedSecret.type === SecretSharingType.Request) {
+    return sharedSecret;
+  }
+
+  const { identifier, ...rest } = sharedSecret;
+
+  if (!identifier) {
+    throw new BadRequestError({
+      message: `Shared secret created at ${sharedSecret.createdAt.toISOString()} has no identifier`
```
maidul98: nit: why do we need to add the time stamp? maybe without is good

backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -194,29 +226,64 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
...
-          .array()
-          .max(100)
-          .optional()
-          .transform((val) => (val ? [...new Set(val)] : undefined))
-      }),
+      body: z
+        .object({
+          name: z.string().max(50).optional().describe(SECRET_SHARING.CREATE.name),
+          password: z.string().optional().describe(SECRET_SHARING.CREATE.password),
+          secretValue: z.string().describe(SECRET_SHARING.CREATE.secretValue),
```
maidul98: we should also enforce max value here as first line for defense

backend/src/server/routes/v1/secret-sharing-router.ts
```diff
@@ -71,90 +78,86 @@ export const registerSecretSharingRouter = async (server: FastifyZodProvider) =>
     }
   });
 
+  server.route({
+    method: "GET",
+    url: "/:id",
+    config: {
+      rateLimit: publicEndpointLimit
```
maidul98, line 85: nit: someone can keep trying as many passwords and know if they got it or not by changing IPs. Maybe we need a exponential back off on the secret

backend/src/server/routes/v1/secret-sharing-router.ts
varonix0: Good catch

backend/src/server/routes/v1/secret-sharing-router.ts
varonix0: Hmm I think this can wait. Someone would need the actual ID, so the main risk is guessing the password for a link they already have, and we already have per-ip limiting. If we see abuse we can add a per-shared-secret ID backoff later; for now I think we can leave this as-is

backend/src/services/secret-sharing/secret-sharing-service.ts
varonix0: I think we should just make it display the identity name instead - pushed

backend/src/server/routes/v1/secret-sharing-router.ts
maidul98: sounds good

maidul98 (approved):

PR #5428, opened 2026-02-11
feature: add PKI Network Discovery

carlosmonastyrski:
## Context
This PR adds the PKI Network Discovery feature, enabling users to scan their network infrastructure (IP ranges, CIDR blocks, and domains) to automatically discover TLS certificates deployed across endpoints. Discovery jobs can be configured with target IPs/domains and ports, triggered manually or scheduled for recurring scans, and results are stored as certificate installations with imported certificates from them.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/frontend/src/hooks/api/pkiDiscovery/types.ts
+++ b/frontend/src/hooks/api/pkiDiscovery/types.ts
@@ -46,6 +46,7 @@ export type TPkiDiscovery = {
   isAutoScanEnabled: boolean;
   scanIntervalDays: number | null;
   gatewayId: string | null;
+  gatewayName: string | null;
   isActive: boolean;
   lastScanStatus: PkiDiscoveryScanStatus | null;
   lastScanJobId: string | null;


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/d15c4c3b-8068-4e40-884d-8e32682034a9?refs=b23e74f2b7c720ab4e36ead7664813493576ebcb&source=prChecksComment) |

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,630 @@
...
+  return new Promise((resolve) => {
+    const certificates: TScanCertificateResult[] = [];
+
+    let servername: string | undefined;
+    if (sniHostname) {
+      servername = sniHostname;
+    } else if (!isIpAddress(host)) {
+      servername = host;
+    }
+    const socket = tls.connect(
```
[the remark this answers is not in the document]
carlosmonastyrski: @greptile isn't servername in Node.js tls.connect only used for the SNI extension in the TLS handshake, it doesn't trigger DNS resolution or affect the TCP connection target (that's determined solely by host). Can you confirm this is a false positive?

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,701 @@
...
+  } else {
+    try {
+      const resolved = await resolveDomain(result.host);
+      if (resolved.length > 0) {
+        [ipAddress] = resolved;
+      }
+    } catch {
+      // do nothing
+    }
+  }
```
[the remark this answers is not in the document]
carlosmonastyrski: @greptile this re-resolution happens after the scan is already complete and is only used to populate metadata for the location fingerprint (display name). No network connection is made with the result. Can you confirm?

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,630 @@
...
+  TScanEndpointResult,
+  TScanTarget
+} from "./pki-discovery-types";
+
+const PUBLIC_DNS_RESOLVERS = [
+  { name: "Google", servers: ["8.8.8.8", "8.8.4.4"] },
+  { name: "Cloudflare", servers: ["1.1.1.1", "1.0.0.1"] },
+  { name: "Quad9", servers: ["9.9.9.9", "149.112.112.112"] },
+  { name: "OpenDNS", servers: ["208.67.222.222", "208.67.220.220"] }
+];
```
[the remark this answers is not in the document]
carlosmonastyrski: @greptile this is intentional. Private/internal domains are resolved by the gateway itself, which runs on the private network and uses its local DNS.

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,630 @@
...
+            isResolved: true,
+            originalTarget: range
+          });
+        }
+      }
+    }
+  }
+
+  return targets;
+};
```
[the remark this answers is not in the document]
carlosmonastyrski: @greptile Is the DNS rebinding scenario described possible here? scanEndpoint receives the already-resolved IP, not the domain, so rebinding has no effect. Can you confirm this is a false positive? That said, we did fix a related filter bypass where domains resolving only to private IPs would fall through and pass the raw domain to scanEndpoint, bypassing the private IP check.

carlosmonastyrski:
@greptile, re-review this PR with the context of my previous responses to your comments, and update the Overview and Summar

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
```diff
@@ -0,0 +1,139 @@
...
+import { TableName } from "../schemas";
+import { createOnUpdateTrigger, dropOnUpdateTrigger } from "../utils";
+
+export async function up(knex: Knex): Promise<void> {
+  if (await knex.schema.hasTable(TableName.Certificate)) {
+    const hasSourceColumn = await knex.schema.hasColumn(TableName.Certificate, "source");
+    if (!hasSourceColumn) {
+      await knex.schema.alterTable(TableName.Certificate, (t) => {
+        t.string("source").defaultTo("issued").notNullable();
+      });
```
sheensantoscapadngan, line 38: is it safe to make this assumption when some of the certificates could have been imported?

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
```diff
@@ -0,0 +1,139 @@
...
+      t.integer("scanIntervalDays").nullable();
+      t.uuid("gatewayId").nullable();
+      t.foreign("gatewayId").references("id").inTable(TableName.GatewayV2).onDelete("SET NULL");
+      t.boolean("isActive").defaultTo(true).notNullable();
+      t.string("lastScanStatus").nullable();
+      t.string("lastScanJobId").nullable();
+      t.string("lastScanMessage").nullable();
+      t.datetime("lastScannedAt").nullable();
+      t.integer("certificatesFound").defaultTo(0).notNullable();
+      t.integer("installationsFound").defaultTo(0).notNullable();
```
sheensantoscapadngan: is this still relevant? doesn't look like we're setting or using this anywhere

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
```diff
@@ -0,0 +1,139 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+import { createOnUpdateTrigger, dropOnUpdateTrigger } from "../utils";
+
+export async function up(knex: Knex): Promise<void> {
+  if (await knex.schema.hasTable(TableName.Certificate)) {
+    const hasSourceColumn = await knex.schema.hasColumn(TableName.Certificate, "source");
+    if (!hasSourceColumn) {
```
sheensantoscapadngan, line 11: looks like we don't have an index yet for fingerprintSha256. we should add that

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
```diff
@@ -0,0 +1,139 @@
...
+      await knex.schema.alterTable(TableName.Certificate, (t) => {
+        t.string("source").defaultTo("issued").notNullable();
```
sheensantoscapadngan: I thought we were adding a column for storing details about where a certificate was discovered? like some sort of metadata that users can always refer to even when the linked installation has already been deleted?

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  if (!(await knex.schema.hasTable(TableName.PkiCertificateInstallationCert))) {
+    await knex.schema.createTable(TableName.PkiCertificateInstallationCert, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.uuid("installationId").notNullable();
+      t.foreign("installationId").references("id").inTable(TableName.PkiCertificateInstallation).onDelete("CASCADE");
+      t.uuid("certificateId").notNullable();
+      t.foreign("certificateId").references("id").inTable(TableName.Certificate).onDelete("CASCADE");
+      t.boolean("isCurrentlyPresent").defaultTo(true).notNullable();
```
sheensantoscapadngan: why do we need this?

backend/src/ee/routes/v1/pki-discovery-router.ts
```diff
@@ -0,0 +1,487 @@
...
+      operationId: "createPkiDiscovery",
+      description: "Create a new PKI discovery configuration",
+      body: z.object({
+        projectId: z.string().describe("The ID of the project"),
+        name: z
+          .string()
+          .min(1)
+          .max(100)
+          .regex(new RE2("^[a-z0-9-]+$"), "Name must contain only lowercase letters, numbers, and hyphens")
+          .describe("Name of the discovery configuration"),
```
sheensantoscapadngan: use the slugSchema helper that we already have

backend/src/ee/routes/v1/pki-discovery-router.ts
```diff
@@ -0,0 +1,487 @@
...
+
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        projectId: req.body.projectId,
+        event: {
+          type: EventType.CREATE_PKI_DISCOVERY,
+          metadata: {
+            discoveryId: discovery.id,
+            name: discovery.name,
+            projectId: req.body.projectId
```
sheensantoscapadngan: we already store `projectId` in the audit log so there's no need to have it here as metadata as well

backend/src/ee/routes/v1/pki-discovery-router.ts
```diff
@@ -0,0 +1,487 @@
...
+        actorOrgId: req.permission.orgId
+      });
+
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        projectId: req.query.projectId,
+        event: {
+          type: EventType.GET_PKI_DISCOVERIES,
+          metadata: {
+            projectId: req.query.projectId,
```
sheensantoscapadngan: same here and for all the other places we emit events

backend/src/ee/routes/v1/pki-discovery-router.ts
```diff
@@ -0,0 +1,487 @@
...
+        discoveryId: z.string().uuid().describe("The ID of the discovery configuration")
+      }),
+      body: z.object({
+        name: z
+          .string()
+          .min(1)
+          .max(100)
+          .regex(new RE2("^[a-z0-9-]+$"), "Name must contain only lowercase letters, numbers, and hyphens")
+          .optional()
+          .describe("Name of the discovery configuration"),
```
sheensantoscapadngan: same remark here and all the other relevant places

backend/src/ee/routes/v1/pki-discovery-router.ts
```diff
@@ -0,0 +1,487 @@
...
+    },
+    handler: async (req) => {
+      if (req.body.targetConfig) {
+        const hasGateway = req.body.gatewayId !== null && req.body.gatewayId !== undefined;
+        const validation = validateTargetConfig(
+          req.body.targetConfig.ipRanges,
+          req.body.targetConfig.ports,
+          req.body.targetConfig.domains,
+          hasGateway
+        );
```
sheensantoscapadngan, line 293: Given that this is a PATCH endpoint, wouldn't this be problematic because what if targetConfig was being updated but gatewayID isn't being passed in by the client?

backend/src/ee/routes/v1/pki-installation-router.ts
```diff
@@ -0,0 +1,273 @@
...
+          metadata: {
+            projectId,
+            count: installations.length
+          }
+        }
+      });
+
+      return { installations };
+    }
+  });
```
sheensantoscapadngan: hmmm this feels off from a RESTful design perspective... can we instead just have a `certificateId` filter for the `GET /` endpoint?

backend/src/ee/routes/v1/pki-installation-router.ts
```diff
@@ -0,0 +1,273 @@
...
+      params: z.object({
+        installationId: z.string().uuid().describe("The ID of the installation")
+      }),
+      response: {
+        200: PkiCertificateInstallationsSchema.extend({
+          certificates: z
+            .array(
+              z.object({
+                id: z.string().uuid(),
+                installationId: z.string().uuid(),
```
sheensantoscapadngan: property might not be necessary anymore given that there is `installationId` in the parent object

backend/src/ee/routes/v1/pki-installation-router.ts
```diff
@@ -0,0 +1,273 @@
...
+      description: "Get a PKI certificate installation by ID with linked certificates",
+      params: z.object({
+        installationId: z.string().uuid().describe("The ID of the installation")
+      }),
+      response: {
+        200: PkiCertificateInstallationsSchema.extend({
+          certificates: z
+            .array(
+              z.object({
+                id: z.string().uuid(),
```
sheensantoscapadngan: what is the difference between this and certificateId below?

backend/src/ee/services/pki-discovery/pki-certificate-installation-cert-dal.ts
```diff
@@ -0,0 +1,88 @@
...
+        .where({ installationId, certificateId })
+        .first();
+
+      if (existing) {
+        const updateData: Partial<TPkiCertificateInstallationCerts> = {
+          lastSeenAt: data.lastSeenAt
+        };
+        if (data.isCurrentlyPresent !== undefined) {
+          updateData.isCurrentlyPresent = data.isCurrentlyPresent;
+        }
```
sheensantoscapadngan: related to my Q above but why do we care about tracking `isCurrentlyPresent` explicitly? I think when we discussed this via DM, resolution was that `lastScanned` would work just as well

backend/src/ee/services/pki-discovery/pki-certificate-installation-cert-dal.ts
```diff
@@ -0,0 +1,88 @@
...
+        .where({ installationId })
+        .orderBy("lastSeenAt", "desc");
+
+      return docs as TPkiCertificateInstallationCerts[];
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Find PKI certificate installation certs by installation ID" });
+    }
+  };
+
+  const upsertCertLink = async (
```
sheensantoscapadngan, line 25: can we do it like the following instead?

Image

If no, I do think this read + mutate DB operation should be done in the same transaction which means that if `tx` isn't passed in then we should create a tx using `db`

backend/src/ee/services/pki-discovery/pki-certificate-installation-dal.ts
```diff
@@ -0,0 +1,214 @@
...
+
+      if (discoveryId) {
+        void query
+          .join(
+            TableName.PkiDiscoveryInstallation,
+            `${TableName.PkiCertificateInstallation}.id`,
+            `${TableName.PkiDiscoveryInstallation}.installationId`
+          )
+          .where(`${TableName.PkiDiscoveryInstallation}.discoveryId`, discoveryId);
+      }
```
sheensantoscapadngan, line 90: we're joining it but we're not selecting any columns from the joined table... is this intentionally for filtering purposes?

backend/src/ee/services/pki-discovery/pki-certificate-installation-dal.ts
```diff
@@ -0,0 +1,214 @@
...
+  };
+
+  const countByProjectId = async (
+    projectId: string,
+    { discoveryId, search, tx }: { discoveryId?: string; search?: string; tx?: Knex } = {}
+  ): Promise<number> => {
+    try {
+      interface CountResult {
+        count: string;
+      }
```
sheensantoscapadngan: might be better if you just use this directly inline below instead of declaring it separately

backend/src/ee/services/pki-discovery/pki-certificate-installation-dal.ts
```diff
@@ -0,0 +1,214 @@
...
+        .join(
+          TableName.Certificate,
+          `${TableName.PkiCertificateInstallationCert}.certificateId`,
+          `${TableName.Certificate}.id`
+        )
+        .where(`${TableName.PkiCertificateInstallationCert}.installationId`, id);
+
+      if (!includeAllCerts) {
+        query = query.where(`${TableName.PkiCertificateInstallationCert}.isCurrentlyPresent`, true);
+      }
```
sheensantoscapadngan: shouldn't we always be showing all the certs of an installation regardless if it's currently present or not?

backend/src/ee/services/pki-discovery/pki-discovery-config-dal.ts
```diff
@@ -0,0 +1,134 @@
...
+            .orWhereRaw(`"lastScannedAt" + ("scanIntervalDays" * interval '1 day') < NOW()`);
+        });
+
+      return docs as TPkiDiscoveryConfigs[];
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Find PKI discovery configs due for scan" });
+    }
+  };
+
+  const findByIdWithCounts = async (id: string, tx?: Knex) => {
```
sheensantoscapadngan: withInstallationCounts

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+    return { valid: true, ipCount: 0, portCount: parsedPorts.length };
+  }
+
+  if (hasIpRanges) {
+    const cidrRanges = ipRanges.filter((r) => r.includes("/"));
+    const singleIps = ipRanges.filter((r) => !r.includes("/"));
+
+    if (cidrRanges.length > 0 && singleIps.length > 0) {
+      return { valid: false, error: "Cannot mix CIDR ranges with individual IPs. Use one or the other." };
+    }
```
sheensantoscapadngan: why don't we allow this? use-case seems reasonable

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+
+    if (cidrRanges.length > 1) {
+      return { valid: false, error: "Only one CIDR range allowed per discovery job" };
+    }
```
sheensantoscapadngan: why do we enforce this? I think the important thing is that we're enforcing max number of targets, not necessarily on how they choose to define those targets

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+
+    if (shouldBlockPrivateIps(!!hasGateway)) {
```
sheensantoscapadngan, line 105: what happens when a single domain value passed in resolves to a private/internal IP. Are we able to block that?

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+};
+
+const isIpAddress = (host: string): boolean => {
+  return net.isIP(host) !== 0;
+};
+
+export const scanEndpoint = async (
+  host: string,
+  port: number,
+  timeout = TLS_SCAN_TIMEOUT,
```
sheensantoscapadngan, line 188: I think we can make `TLS_SCAN_TIMEOUT` to be as low as 3 seconds

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+
+const parsePeerCertificate = (cert: tls.DetailedPeerCertificate): TScanCertificateResult | null => {
+  try {
+    const derBuffer = cert.raw;
+    const b64 = derBuffer.toString("base64");
+    const lines: string[] = [];
+    for (let i = 0; i < b64.length; i += 64) {
+      lines.push(b64.substring(i, i + 64));
+    }
+    const pem = `-----BEGIN CERTIFICATE-----\n${lines.join("\n")}\n-----END CERTIFICATE-----`;
```
sheensantoscapadngan: instead of manually doing this can't we just do `x509Cert.toString("pem")`?

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+
+    let extendedKeyUsages: CertExtendedKeyUsage[] = [];
+    const extKeyUsageExt = x509Cert.getExtension("2.5.29.37") as x509.ExtendedKeyUsageExtension;
+    if (extKeyUsageExt) {
+      extendedKeyUsages = extKeyUsageExt.usages
+        .map((ekuOid) => CertExtendedKeyUsageOIDToName[ekuOid as string])
+        .filter(Boolean);
+    }
+
+    const certAny = cert as { ca?: boolean; pathlen?: number };
```
sheensantoscapadngan: why is this called `certAny`? doesn't seem descriptive

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+const extractKeyAlgorithm = (cert: tls.DetailedPeerCertificate): string | undefined => {
+  const certAny = cert as { asn1Curve?: string };
+  if (cert.bits && !certAny.asn1Curve) {
+    return `RSA-${cert.bits}`;
+  }
+  if (certAny.asn1Curve) {
+    return `EC-${certAny.asn1Curve}`;
+  }
+  return undefined;
+};
```
sheensantoscapadngan: seems like there's a more reliable way extract the key algorithm using the public key like so. can you confirm? apparently the current approach could miss algorithms like DSA or Ed25519/Ed448

Image

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
```diff
@@ -0,0 +1,647 @@
...
+      if (ips.length < maxHosts) {
+        ips.push(ip);
+      }
+    });
+
+    return ips;
+  } catch {
+    return [cidr];
+  }
+};
```
sheensantoscapadngan: seems like with this implementation, we'd be silently failing when block has more than 256. Though we already have a check to ensure it always stays 256, for future reference it might be better to remove the maxHosts check here and instead just let this do the cidr expansion with no limits

backend/src/server/routes/index.ts
```diff
@@ -99,6 +99,14 @@ import { pkiAcmeOrderAuthDALFactory } from "@app/ee/services/pki-acme/pki-acme-o
 import { pkiAcmeOrderDALFactory } from "@app/ee/services/pki-acme/pki-acme-order-dal";
 import { pkiAcmeQueueServiceFactory } from "@app/ee/services/pki-acme/pki-acme-queue";
 import { pkiAcmeServiceFactory } from "@app/ee/services/pki-acme/pki-acme-service";
+import { pkiCertificateInstallationCertDALFactory } from "@app/ee/services/pki-discovery/pki-certificate-installation-cert-dal";
```
sheensantoscapadngan, line 103: IGNORE PLACEMENT OF COMMENT

Doesn't seem like I can attach a gateway during creation? (I don't have any gateways configured in this org)
Image

frontend/src/pages/cert-manager/DiscoveryDetailsByIDPage/DiscoveryDetailsByIDPage.tsx
```diff
@@ -0,0 +1,205 @@
...
+  ] as const);
+
+  if (isLoading) {
+    return <UnstablePageLoader />;
+  }
+
+  if (!discovery) {
+    return null;
+  }
+
```
sheensantoscapadngan, line 64: IGNORE PLACEMENT OF COMMENT

I recall that we've had a request about displaying a name for an installation other than just IP and port. Would it make sense if we had another column here for "Certificate name"? So in this example, the 216.150.1.1 endpoint would have certificate name of "infisical.com". Not sure what the exact column name is supposed to be but this would be helpful 

Image

frontend/src/pages/cert-manager/DiscoveryDetailsByIDPage/DiscoveryDetailsByIDPage.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const handleTriggerScan = async () => {
+    try {
+      await triggerScan.mutateAsync({ discoveryId, projectId });
+    } catch {
+      // Error handled by mutation
+    }
+  };
+
```
sheensantoscapadngan, line 72: IGNORE PLACEMENT OF COMMENT

It might be confusing that we're naming the column "Endpoint" but it's actually an Installation. Can this column just be "name" instead? 

Image

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
carlosmonastyrski: Initially the assumption was to just go with issued since it should be the general case for any certs managed by this feature so far. But makes sense to also add a bit more complexity on the migration for imports, updating it to work with that 👍

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
carlosmonastyrski: We were not going to add that since it was going to be more troublesome as the same certificate is part of multiple installation, but as you pointed out maybe at some point it'll be useful on installation remove scenarios. Adding that as well

backend/src/ee/services/pki-discovery/pki-certificate-installation-dal.ts
carlosmonastyrski: Yes, this is intentional for filtering

backend/src/ee/services/pki-discovery/pki-discovery-fns.ts
carlosmonastyrski: We handle that during the discovery process (actually the same can happen any time, where from one execution to another the value can change), so during that scan, if there is a gateway in place we allow it, otherwise it'll not proceed with the discovery attempt

backend/src/db/migrations/20260205012343_add-pki-discovery-tables.ts
```diff
@@ -0,0 +1,180 @@
...
+      await knex.schema.alterTable(TableName.Certificate, (t) => {
+        t.string("source").nullable();
+      });
+
+      // Backfill: certs with a profileId were issued, others were imported
+      await knex(TableName.Certificate).whereNotNull("profileId").update({ source: CertificateSource.Issued });
+      await knex(TableName.Certificate).whereNull("profileId").update({ source: CertificateSource.Imported });
+    }
+
+    const hasFingerprintIndex = await knex.schema.hasColumn(TableName.Certificate, "fingerprintSha256");
```
sheensantoscapadngan: minor - should be `hasFingerprintCol` instead right

backend/src/ee/routes/v1/pki-discovery-router.ts
```diff
@@ -0,0 +1,489 @@
...
+        discoveryType: req.body.discoveryType,
+        targetConfig: req.body.targetConfig as TPkiDiscoveryTargetConfig,
+        isAutoScanEnabled: req.body.isAutoScanEnabled,
+        scanIntervalDays: req.body.scanIntervalDays,
+        gatewayId: req.body.gatewayId,
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId
+      });
```
sheensantoscapadngan, line 119: looks like we're missing the permission check on whether user can use the gateway being passed in?

backend/src/ee/routes/v1/pki-discovery-router.ts
```diff
@@ -0,0 +1,489 @@
...
+    schema: {
+      hide: false,
+      tags: [ApiDocsTags.PkiDiscovery],
+      operationId: "updatePkiDiscovery",
+      description: "Update a PKI discovery configuration",
+      params: z.object({
+        discoveryId: z.string().uuid().describe("The ID of the discovery configuration")
+      }),
+      body: z.object({
+        name: slugSchema({ field: "Name", max: 100 }).optional().describe("Name of the discovery configuration"),
```
sheensantoscapadngan, line 257: same here re: gateway usage

backend/src/ee/services/pki-discovery/pki-discovery-installation-dal.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      if (existing) {
+        await (tx || db)(TableName.PkiDiscoveryInstallation).where({ id: existing.id }).update({ lastScannedAt });
+        return { ...existing, lastScannedAt };
+      }
+
+      const [created] = await (tx || db)(TableName.PkiDiscoveryInstallation)
+        .insert({ discoveryId, installationId, lastScannedAt })
+        .returning("*");
+
```
sheensantoscapadngan: I think we can use the same upsert query here to simplify?

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,690 @@
...
+                const sniHostname = target.sniHostname || target.host;
+                return scanEndpoint("localhost", proxyPort, DEFAULT_SCAN_TIMEOUT, sniHostname);
+              },
+              {
+                relayHost: targetGatewayDetails.relayHost,
+                gateway: targetGatewayDetails.gateway,
+                relay: targetGatewayDetails.relay,
+                protocol: GatewayProxyProtocol.Tcp
+              }
+            );
```
sheensantoscapadngan, line 190: we can probably do some sort of optimization here at some point where we open one persistent relay connection and then reuse that across all the sequential target scans..

just a note for the future though but this might be problematic once released. Let's see

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,690 @@
...
+
+          if (!targetGatewayDetails) {
+            consecutiveGwFailures += 1;
+            results.push({
+              host: target.host,
+              port: target.port,
+              success: false,
+              failureReason: ScanEndpointFailureReason.ConnectionFailed,
+              sniHostname: target.sniHostname
+            });
```
sheensantoscapadngan, line 177: we'd be able to show gateway-specific error messages for this one right?

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,690 @@
...
+            locationType: PkiInstallationLocationType.Network,
+            locationDetails: domainLocationDetails,
+            locationFingerprint: domainFingerprint,
+            name: domain,
+            type: PkiInstallationType.Unknown,
+            lastSeenAt: now
+          });
+        } else {
+          await deps.pkiCertificateInstallationDAL.updateById(domainInstallation.id, { lastSeenAt: now });
+        }
```
sheensantoscapadngan: let's put in the same txn

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,690 @@
...
+
+        const domainLink = await deps.pkiDiscoveryInstallationDAL.upsertLink(discoveryId, domainInstallation.id, now);
+        if (!domainLink) continue; // eslint-disable-line no-continue
+
+        for (const certId of certIds) {
+          await deps.pkiCertificateInstallationCertDAL.upsertCertLink(domainInstallation.id, certId, {
+            lastSeenAt: now
+          });
+        }
```
sheensantoscapadngan: same for these ones

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,690 @@
...
+      );
+      if (certId) {
+        certificateIds.push(certId);
+
+        await pkiCertificateInstallationCertDAL.upsertCertLink(installation.id, certId, {
+          lastSeenAt: scanTime
+        });
+      }
+    }
```
sheensantoscapadngan: should be using txn here as well

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,690 @@
...
+      });
+
+      const encryptedCertificateChain = chainPem
+        ? (await kmsEncryptor({ plainText: Buffer.from(chainPem) })).cipherTextBlob
+        : undefined;
+
+      await certificateBodyDAL.create({
+        certId: newCert.id,
+        encryptedCertificate,
+        encryptedCertificateChain
```
sheensantoscapadngan: should be using txn

backend/src/ee/services/pki-discovery/pki-discovery-service.ts
```diff
@@ -0,0 +1,410 @@
...
+      projectId,
+      name,
+      description,
+      discoveryType,
+      targetConfig,
+      isAutoScanEnabled: isAutoScanEnabled ?? false,
+      scanIntervalDays: isAutoScanEnabled ? scanIntervalDays : null,
+      gatewayId,
+      isActive: true
+    });
```
sheensantoscapadngan: we could also handle uniqueness by checking the error thrown by the DAL right? I think we've standardized to use that approach

backend/src/ee/services/pki-discovery/pki-discovery-service.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    const updatedDiscovery = await pkiDiscoveryConfigDAL.updateById(discoveryId, {
+      name,
+      description,
+      targetConfig,
+      isAutoScanEnabled,
+      scanIntervalDays: isAutoScanEnabled ? scanIntervalDays : null,
+      gatewayId,
+      isActive
+    });
```
sheensantoscapadngan: same remark here

docs/api-reference/endpoints/pki-installations/list-by-certificate.mdx
```diff
@@ -0,0 +1,4 @@
+---
```
sheensantoscapadngan: we don't need this anymore right?

docs/documentation/platform/pki/discovery/network.mdx
```diff
@@ -0,0 +1,82 @@
...
+    ```
+  </Tab>
+</Tabs>
+
+## Limits
+
+<Note>
+  The following limits apply to Network discovery jobs:
+  - **Max individual IPs**: 256
+  - **Max domains**: 5
```
sheensantoscapadngan: I think we can increase the number of domains to be 10 or 20? Seems to be very limiting with just 5

frontend/src/hooks/api/pkiDiscovery/queries.tsx
```diff
@@ -0,0 +1,168 @@
...
+    queryFn: async () => {
+      const params = new URLSearchParams({ projectId, certificateId });
+      const { data } = await apiRequest.get<TListPkiInstallationsResponse>(
+        `/api/v1/cert-manager/installations?${params.toString()}`
+      );
+      return data.installations;
+    },
+    enabled: Boolean(projectId) && Boolean(certificateId)
+  });
+};
```
sheensantoscapadngan: this can just be reduced to one `useQuery`... maybe you can pass in `certificateId` optionally in `useGetPkiInstallation`?

backend/src/ee/services/pki-discovery/pki-discovery-installation-dal.ts
```diff
@@ -0,0 +1,60 @@
+import { Knex } from "knex";
```
sheensantoscapadngan, line 1: IGNORE PLACEMENT OF COMMENT

I shouldn't be able to revoke "discovered" certificates right? 
Image

backend/src/ee/services/pki-discovery/pki-discovery-installation-dal.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import { TDbClient } from "@app/db";
```
sheensantoscapadngan, line 3: IGNORE PLACEMENT OF COMMENT

for discovered certificates, it looks like we're not showing any fingerprints? 

Image

backend/src/ee/services/pki-discovery/pki-discovery-installation-dal.ts
```diff
@@ -0,0 +1,60 @@
...
+      return created;
+    } catch (error) {
+      // PG foreign key violation — the installation was deleted between the check and the insert
+      if (error && typeof error === "object" && "code" in error && (error as { code: string }).code === "23503") {
+        return null;
+      }
+      throw new DatabaseError({ error, name: "Upsert PKI discovery installation link" });
+    }
+  };
+
```
sheensantoscapadngan, line 47: IGNORE PLACEMENT OF COMMENT

for discovered certificates, it would be helpful to see details about the issuing CAs here.. Currently we only show details for those issued through CAs in Infisical

Image

backend/src/ee/services/pki-discovery/pki-discovery-installation-dal.ts
```diff
@@ -0,0 +1,60 @@
+import { Knex } from "knex";
+
+import { TDbClient } from "@app/db";
+import { TableName, TPkiDiscoveryInstallations } from "@app/db/schemas";
+import { DatabaseError } from "@app/lib/errors";
+import { ormify } from "@app/lib/knex";
+
+export type TPkiDiscoveryInstallationDALFactory = ReturnType<typeof pkiDiscoveryInstallationDALFactory>;
+
```
sheensantoscapadngan, line 9: IGNORE PLACEMENT OF COMMENT

would be helpful to see which discovery job last saw an installation from here 
Image

backend/src/ee/services/pki-discovery/pki-discovery-installation-dal.ts
```diff
@@ -0,0 +1,60 @@
...
+      const docs = await (tx || db.replicaNode())(TableName.PkiDiscoveryInstallation)
+        .where({ discoveryId })
+        .orderBy("lastScannedAt", "desc");
+
+      return docs as TPkiDiscoveryInstallations[];
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Find PKI discovery installations by discovery ID" });
+    }
+  };
+
```
sheensantoscapadngan, line 24: IGNORE PLACEMENT OF COMMENT

is "In use" badge here still relevant? 
Image

backend/src/ee/services/pki-discovery/pki-discovery-queue.ts
```diff
@@ -0,0 +1,134 @@
+import { TGatewayV2DALFactory } from "@app/ee/services/gateway-v2/gateway-v2-dal";
+import { TGatewayV2ServiceFactory } from "@app/ee/services/gateway-v2/gateway-v2-service";
+import { logger } from "@app/lib/logger";
```
sheensantoscapadngan, line 3: IGNORE PLACEMENT OF COMMENT

It would also be very helpful to see when was the last time a specific certificate was found in this installation
Image

backend/src/ee/services/pki-discovery/pki-discovery-queue.ts
```diff
@@ -0,0 +1,134 @@
...
+  certificateDAL,
+  certificateBodyDAL,
+  projectDAL,
+  kmsService,
+  queueService,
+  gatewayV2Service,
+  gatewayV2DAL
+}: TPkiDiscoveryQueueFactoryDep) => {
+  const startPkiDiscoveryScanQueue = () => {
+    queueService.start(QueueName.PkiDiscoveryScan, async (job) => {
```
sheensantoscapadngan, line 50: IGNORE PLACEMENT OF COMMENT

Why don't we show the name here?

Image

Image

backend/src/ee/services/pki-discovery/pki-discovery-installation-dal.ts
sheensantoscapadngan, line 1: IGNORE PLACEMENT OF COMMENT

Can we add a title and description for the Jobs and Installations views? like how we have it for certificate profiles? This would make UX easier because they'll have more context in-platform

Image

E.G:

Image

backend/src/ee/services/pki-discovery/pki-discovery-scan-fns.ts
```diff
@@ -0,0 +1,690 @@
...
+          discoveryConfigId: discoveryId,
+          startedAt,
+          status: PkiDiscoveryScanStatus.Running,
+          targetsScannedCount: 0,
+          certificatesFoundCount: 0,
+          installationsFoundCount: 0
+        },
+        tx
+      );
+
```
sheensantoscapadngan, line 107: IGNORE PLACEMENT OF COMMENT

should we allow discovered certificates to be pushed out for PKI syncs?

Image

PR #5461, opened 2026-02-12
improvement(secret-rotation): validate conflicts before rotating external credentials

victorvhs017:
## Context

When creating a new secret rotation, if validation checks fail (e.g., rotation with the same name already exists, or secrets with mapped names already exist), the error is returned, but the resource credentials have already been changed.

This happens because the validation for conflicting secrets was only performed inside the transaction callback, which executes after the external credential rotation has already occurred. This leaves users in a confusing state where their password was changed on the external system, but no rotation record was created in Infisical.

To fix this we added a check before the rotation to fail fast.

## Steps to verify the change

1. Create a secret with key `DB_PASSWORD`
2. Try to create a secret rotation that maps to `DB_PASSWORD` at the same path
3. **Before:** The external password would be rotated, then the creation would fail with "secret already exists"
4. **After:** The creation fails immediately with "secret already exists" without rotating the external password

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
+++ b/backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
@@ -216,7 +216,7 @@ export const secretRotationV2ServiceFactory = ({
   }: {
     secretKeys: string[];
     folderId: string;
-    tx: Knex;
+    tx?: Knex;
     secretPath: string;
   }) => {
     if (new Set(secretKeys).size !== secretKeys.length) {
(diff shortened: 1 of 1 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/c2805a80-7f35-4459-8dbb-b8829de01e11?refs=064a2f16382a52c7bdfd252face44e5157c95f8c&source=prChecksComment) |

PR #5472, opened 2026-02-13
feat(api): new membership API for org and suborgs groups

IgorHorta:
## Context

Sub-organization groups and project groups previously used legacy group endpoints (e.g. `POST/DELETE /groups/:id/link` for org link/unlink). We're moving to the new membership API shape `{scope}/memberships/{actor-type}/{actor-id}` so org and project group membership is consistent and easier to extend.

**Before:** Org/project groups used a mix of legacy group routes; link/unlink used `POST/DELETE /groups/:id/link`.  
**After:** Org groups use `GET/POST/PATCH/DELETE /api/v1/organizations/memberships/groups` (and `.../groups/:groupId` for PATCH/DELETE). Project groups use `.../projects/:projectId/memberships/groups`. Link/unlink use the new create/delete org group membership endpoints. Legacy `/:id/link` routes are removed. Group service fetches membership + group in parallel where both are needed.

Related: PLATFRM-198

## Steps to verify the change

1. **Org groups list** – In an org (or sub-org), open Groups. Confirm list loads and shows owned + inherited groups.
2. **Org group create/edit/delete** – Create a group, rename it, delete it. Confirm all work and list refreshes.
3. **Org group link/unlink** – In a sub-org, link an existing group and set role; then unlink. Confirm list and detail update and cache invalidates for the current org.
4. **Project groups** – In a project, open Groups. Add a group, change role, remove group. Confirm list and behaviour match expectations.
5. **Sub-org inherited group** – In a sub-org, open an inherited group. Confirm name/slug are read-only; only role in this org can be updated.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/permission/org-permission.ts
+++ b/backend/src/ee/services/permission/org-permission.ts
@@ -18,7 +18,7 @@ export enum OrgPermissionActions {
 export enum OrgPermissionSubOrgActions {
   Create = "create",
   DirectAccess = "direct-access",
-  LinkRootGroup = "link-root-group"
+  LinkGroup = "link-group"
 }
 
 export enum OrgPermissionAppConnectionActions {
@@ -201,7 +201,7 @@ export const OrgPermissionSchema = z.discriminatedUnion("subject", [
     action: CASL_ACTION_SCHEMA_ENUM([
       OrgPermissionSubOrgActions.Create,
       OrgPermissionSubOrgActions.DirectAccess,
-      OrgPermissionSubOrgActions.LinkRootGroup
+      OrgPermissionSubOrgActions.LinkGroup
     ]).describe("Describe what action an entity can take.")
   }),
   z.object({
@@ -330,7 +330,7 @@ const buildAdminPermission = () => {
 
   can(OrgPermissionSubOrgActions.Create, OrgPermissionSubjects.SubOrganization);
   can(OrgPermissionSubOrgActions.DirectAccess, OrgPermissionSubjects.SubOrganization);
-  can(OrgPermissionSubOrgActions.LinkRootGroup, OrgPermissionSubjects.SubOrganization);
+  can(OrgPermissionSubOrgActions.LinkGroup, OrgPermissionSubjects.SubOrganization);
 
   // role permission
   can(OrgPermissionActions.Read, OrgPermissionSubjects.Role);
--- a/backend/src/services/membership-group/org/org-membership-group-factory.ts
+++ b/backend/src/services/membership-group/org/org-membership-group-factory.ts
@@ -61,7 +61,7 @@ export const newOrgMembershipGroupFactory = ({
       scope: OrganizationActionScope.Any
     });
     ForbiddenError.from(rootOrgPermission).throwUnlessCan(
-      OrgPermissionSubOrgActions.LinkRootGroup,
+      OrgPermissionSubOrgActions.LinkGroup,
       OrgPermissionSubjects.SubOrganization
     );
 
@@ -167,7 +167,7 @@ export const newOrgMembershipGroupFactory = ({
         scope: OrganizationActionScope.Any
       });
       ForbiddenError.from(rootOrgPermission).throwUnlessCan(
-        OrgPermissionSubOrgActions.LinkRootGroup,
+        OrgPermissionSubOrgActions.LinkGroup,
         OrgPermissionSubjects.SubOrganization
       );
     }
--- a/frontend/src/context/OrgPermissionContext/types.ts
+++ b/frontend/src/context/OrgPermissionContext/types.ts
@@ -116,7 +116,7 @@ export enum OrgPermissionGroupActions {
 export enum OrgPermissionSubOrgActions {
   Create = "create",
   DirectAccess = "direct-access",
-  LinkRootGroup = "link-root-group"
+  LinkGroup = "link-group"
 }
 
 export type AppConnectionSubjectFields = {
--- a/frontend/src/pages/organization/RoleByIDPage/components/OrgRoleModifySection.utils.ts
+++ b/frontend/src/pages/organization/RoleByIDPage/components/OrgRoleModifySection.utils.ts
@@ -128,7 +128,7 @@ const subOrganizationPermissionSchema = z
   .object({
     [OrgPermissionSubOrgActions.Create]: z.boolean().optional(),
     [OrgPermissionSubOrgActions.DirectAccess]: z.boolean().optional(),
-    [OrgPermissionSubOrgActions.LinkRootGroup]: z.boolean().optional()
+    [OrgPermissionSubOrgActions.LinkGroup]: z.boolean().optional()
   })
   .optional();
 
--- a/frontend/src/pages/organization/RoleByIDPage/components/RolePermissionsSection/OrgPermissionSubOrgRow.tsx
+++ b/frontend/src/pages/organization/RoleByIDPage/components/RolePermissionsSection/OrgPermissionSubOrgRow.tsx
@@ -13,7 +13,7 @@ import { TFormSchema } from "../OrgRoleModifySection.utils";
 const PERMISSION_ACTIONS = [
   { action: OrgPermissionSubOrgActions.Create, label: "Create" },
   { action: OrgPermissionSubOrgActions.DirectAccess, label: "Direct Access" },
-  { action: OrgPermissionSubOrgActions.LinkRootGroup, label: "Link Root Group" }
+  { action: OrgPermissionSubOrgActions.LinkGroup, label: "Link Group" }
 ] as const;
 
 type Props = {
@@ -68,7 +68,7 @@ export const OrgPermissionSubOrgRow = ({ isEditable, control, setValue }: Props)
           {
             [OrgPermissionSubOrgActions.Create]: false,
             [OrgPermissionSubOrgActions.DirectAccess]: false,
-            [OrgPermissionSubOrgActions.LinkRootGroup]: false
+            [OrgPermissionSubOrgActions.LinkGroup]: false
           },
           { shouldDirty: true }
         );
@@ -79,7 +79,7 @@ export const OrgPermissionSubOrgRow = ({ isEditable, control, setValue }: Props)
           {
             [OrgPermissionSubOrgActions.Create]: true,
             [OrgPermissionSubOrgActions.DirectAccess]: true,
-            [OrgPermissionSubOrgActions.LinkRootGroup]: true
+            [OrgPermissionSubOrgActions.LinkGroup]: true
           },
           { shouldDirty: true }
         );
@@ -90,7 +90,7 @@ export const OrgPermissionSubOrgRow = ({ isEditable, control, setValue }: Props)
           {
             [OrgPermissionSubOrgActions.Create]: true,
             [OrgPermissionSubOrgActions.DirectAccess]: true,
-            [OrgPermissionSubOrgActions.LinkRootGroup]: true
+            [OrgPermissionSubOrgActions.LinkGroup]: true
           },
           { shouldDirty: true }
         );


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/e91aba77-b6c4-443d-900d-351f863c6ea6?refs=ac08d01fd128fc43edbff75f2719c15b83af11ce&source=prChecksComment) |

backend/src/ee/services/group/group-service.ts
```diff
@@ -211,14 +213,27 @@ export const groupServiceFactory = ({
...
+        scopeData: { scope: AccessScope.Organization, orgId: actorOrgId },
+        groupId: id
+      }),
+      groupDAL.findById(id)
+    ]);
+    if (!groupMembership || !group) {
       throw new NotFoundError({ message: `Failed to find group with ID ${id}` });
     }
 
+    const isInherited = group.orgId !== actorOrgId;
```
akhilmhdh: Small nit: better word i feel like `isLinkedGroup`

backend/src/ee/services/group/group-service.ts
```diff
@@ -243,41 +258,33 @@ export const groupServiceFactory = ({
       if (isCustomRole) customRole = rolePermissionDetails?.role;
     }
 
-    const updatedGroup = await groupDAL.transaction(async (tx) => {
-      if (name) {
+    const updatedGroup = await groupDAL.transaction(async (tx): Promise<TGroups> => {
+      if (!isInherited && (name || slug) && name) {
```
akhilmhdh: This change existed in previous method itself. But let's move this check outside transaction. The reasoning is - this bad request won't be shown in UI, as it's wrapped inside transaction. Errors happening inside transaction are shown only as internal server error

backend/src/ee/services/group/group-service.ts
```diff
@@ -315,12 +322,30 @@ export const groupServiceFactory = ({
...
+        scopeData: { scope: AccessScope.Organization, orgId: actorOrgId },
+        groupId: id
+      }),
+      groupDAL.findById(id)
+    ]);
+    if (!groupMembership || !group) {
+      throw new NotFoundError({ message: `Cannot find group with ID ${id}` });
+    }
+
+    if (group.orgId !== actorOrgId) {
```
akhilmhdh: Missing transaction

backend/src/ee/services/group/group-service.ts
```diff
@@ -336,14 +361,42 @@ export const groupServiceFactory = ({
     });
     ForbiddenError.from(permission).throwUnlessCan(OrgPermissionGroupActions.Read, OrgPermissionSubjects.Groups);
 
-    const group = await groupDAL.findById(id);
-    if (!group || group.orgId !== actorOrgId) {
-      throw new NotFoundError({
-        message: `Cannot find group with ID ${id}`
-      });
+    const [groupMembership, group] = await Promise.all([
```
akhilmhdh: Hmm! I see this getting done a lot of times - should we move this into a one single query with join? I feel like that would be better

backend/src/server/routes/v1/organization-memberships-router.ts
```diff
@@ -0,0 +1,314 @@
...
+    config: {
+      rateLimit: readLimit
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    schema: {
+      hide: false,
+      operationId: "listOrganizationGroupMemberships",
+      tags: [ApiDocsTags.Groups],
+      description:
+        "List organization group memberships (org from token). Pattern: {scope}/memberships/{actor-type}. This route was restructured to follow the new membership design. Retrocompatibility is preserved (legacy routes remain available).",
```
akhilmhdh:
```
This route was restructured to follow the new membership design. Retrocompatibility is preserved (legacy routes remain available).
```

I don't think we need to mention about this part in all of the description

backend/src/server/routes/v1/organization-memberships-router.ts
```diff
@@ -0,0 +1,314 @@
...
+    config: {
+      rateLimit: writeLimit
+    },
+    onRequest: verifyAuth([AuthMode.JWT]),
+    schema: {
+      hide: false,
+      operationId: "deleteOrganizationGroupMembership",
+      tags: [ApiDocsTags.Groups],
+      description:
+        "Delete organization group membership (unlink group from current org). This route was restructured to follow the new membership design. Retrocompatibility is preserved (legacy routes remain available).",
```
akhilmhdh: I think we can make it more clear that this cannot be used where the group's origin is and remove rest of the info along `restructure`

backend/src/server/routes/v1/project-group-memberships-router.ts
```diff
@@ -0,0 +1,333 @@
...
+    config: {
+      rateLimit: writeLimit
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    schema: {
+      hide: false,
+      operationId: "createProjectGroupMembership",
+      tags: [ApiDocsTags.ProjectGroups],
+      description:
+        "Add group to project (create project group membership). This route was restructured to follow the new membership design. Retrocompatibility is preserved (legacy routes remain available).",
```
akhilmhdh: Same here. Let's remove the `This route was restructured to follow the new membership design. Retrocompatibility is preserved (legacy routes remain available).` in all description. Customers doesn't need to know about this. 

Add group to project (create project group membership).  - we should make this more clear as a single statement. Examples

```
<Add/Link - pick a word> a group from organization to the project by creating a project group membership.
```

backend/src/services/membership-group/org/org-membership-group-factory.ts
```diff
@@ -38,10 +41,60 @@ export const newOrgMembershipGroupFactory = ({
...
+      actorId: dto.permission.id,
+      orgId: dto.permission.orgId,
+      actorAuthMethod: dto.permission.authMethod,
+      actorOrgId: dto.permission.orgId,
+      scope: OrganizationActionScope.ChildOrganization
     });
+    ForbiddenError.from(permission).throwUnlessCan(OrgPermissionGroupActions.Create, OrgPermissionSubjects.Groups);
+
+    const group = await groupDAL.findById(dto.data.groupId);
+    if (!group) {
```
akhilmhdh: Let's not throw this error - because users can now find any valid group id existing without even logging into there organization

backend/src/services/membership-group/org/org-membership-group-factory.ts
```diff
@@ -83,10 +136,28 @@ export const newOrgMembershipGroupFactory = ({
     }
   };
 
-  const onDeleteMembershipGroupGuard: TMembershipGroupScopeFactory["onDeleteMembershipGroupGuard"] = async () => {
-    throw new BadRequestError({
-      message: "Organization membership cannot be deleted for organization scoped group"
+  const onDeleteMembershipGroupGuard: TMembershipGroupScopeFactory["onDeleteMembershipGroupGuard"] = async (dto) => {
+    const group = await groupDAL.findById(dto.selector.groupId);
+    if (!group) {
```
akhilmhdh, line 154: Same here - below error for both group not found and the other one as well

backend/src/services/membership-group/org/org-membership-group-factory.ts
```diff
@@ -83,10 +136,28 @@ export const newOrgMembershipGroupFactory = ({
...
+      throw new NotFoundError({ message: "Group not found" });
+    }
+    const isLinkDelete = group.orgId !== dto.permission.orgId;
+    if (!isLinkDelete) {
+      throw new BadRequestError({
+        message:
+          "Organization membership cannot be deleted for organization scoped group. Only linked groups can be unlinked."
```
akhilmhdh: I think this first statement is a bit confusing. 

What do you think? Should we just say more specifically like `Delete membership is only allowed for linked group` and something like `this group was created in this organization`

frontend/src/pages/organization/AccessManagementPage/components/OrgGroupsTab/components/OrgGroupsSection/OrgGroupModal.tsx
```diff
@@ -106,76 +122,158 @@ export const OrgGroupModal = ({ popUp, handlePopUpClose, handlePopUpToggle }: Pr
     });
   };
 
+  const isCreateMode = !popUp?.group?.data;
+  const showToggle = isSubOrganization && isCreateMode;
+
+  const modalTitle = (() => {
```
akhilmhdh: You can put this in a simple if based mode like this
```
let modalTitle = "Create Group";
if() { // model title change ]
```

No need for a wrapper function

backend/src/ee/services/group/group-service.ts
IgorHorta: agreed!

IgorHorta:
@greptile please review this again

backend/src/ee/services/group/group-service.ts
```diff
@@ -322,24 +321,32 @@ export const groupServiceFactory = ({
...
-      await membershipGroupDAL.deleteById(groupMembership.id);
+      // Linked group: unlink by deleting the membership in this org only
+      await groupDAL.transaction(async (tx) => {
+        await membershipRoleDAL.delete({ membershipId: groupMembership.id }, tx);
+        await membershipGroupDAL.deleteById(groupMembership.id, tx);
+      });
       return group;
     }
 
+    const isReferencedAsSource = await groupDAL.hasGroupsReferencingGroup(id);
```
akhilmhdh: It would be great if we can specifically say which groups this has been linked with, so that users can do it correctly

backend/src/ee/services/oidc/oidc-config-service.ts
```diff
@@ -198,14 +204,13 @@ export const oidcConfigServiceFactory = ({
...
         const foundUser = await userDAL.findById(userAlias.userId, tx);
-        const [orgMembership] = await orgDAL.findMembership(
-          {
-            [`${TableName.Membership}.actorUserId` as "actorUserId"]: userAlias.userId,
-            [`${TableName.Membership}.scopeOrgId` as "scopeOrgId"]: orgId,
-            [`${TableName.Membership}.scope` as "scope"]: AccessScope.Organization
-          },
-          { tx }
-        );
+        const orgMembership = await orgDAL.findEffectiveOrgMembership({
```
akhilmhdh: This will not be needed, because oidc login will be disabled for sub-org and they won't be able to create it.  Thus they will have org membership

backend/src/ee/services/oidc/oidc-config-service.ts
```diff
@@ -306,14 +315,13 @@ export const oidcConfigServiceFactory = ({
...
 
-        const [orgMembership] = await orgDAL.findMembership(
-          {
-            [`${TableName.Membership}.actorUserId` as "actorUserId"]: userAlias.userId,
-            [`${TableName.Membership}.scopeOrgId` as "scopeOrgId"]: orgId,
-            [`${TableName.Membership}.scope` as "scope"]: AccessScope.Organization
-          },
-          { tx }
-        );
+        const orgMembership = await orgDAL.findEffectiveOrgMembership({
```
akhilmhdh: Same here - not needed as oidc login can only be done on main org

backend/src/ee/services/permission/org-permission.ts
```diff
@@ -17,7 +17,8 @@ export enum OrgPermissionActions {
 
 export enum OrgPermissionSubOrgActions {
   Create = "create",
-  DirectAccess = "direct-access"
+  DirectAccess = "direct-access",
+  LinkRootGroup = "link-root-group"
```
akhilmhdh: Need to coordinate the right permission word for this. Let's raise a discussion internally

backend/src/ee/services/permission/permission-dal.ts
```diff
@@ -172,21 +172,34 @@ export interface TPermissionDALFactory {
 export const permissionDALFactory = (db: TDbClient): TPermissionDALFactory => {
   const getPermission: TPermissionDALFactory["getPermission"] = async ({ scopeData, tx, actorId, actorType }) => {
     try {
+      // Groups visible in this org: native (group.orgId = orgId) or linked from root (Membership in this org with actorGroupId)
       // akhilmhdh: when group has another group like sub group we would need recursively go down
-      const userGroupSubquery = (tx || db)(TableName.Groups)
-        .leftJoin(TableName.UserGroupMembership, `${TableName.UserGroupMembership}.groupId`, `${TableName.Groups}.id`)
-        .where(`${TableName.Groups}.orgId`, scopeData.orgId)
+      const conn = tx || db;
+      const groupVisibleInOrg = (groupsQb: Knex.QueryBuilder) => {
```
akhilmhdh: Given that this is three sub query - groupVisibleInOrg -> userGroupSubquery -> main query

I think we can just remove this query `groupVisibleInOrg` and remove the org filter as well from `userGroupSubquery`  and `identityGroupSubquery`

This is because we already do a filter by scopeOrgId in line 234
```
 .where(`${TableName.Membership}.scopeOrgId`, scopeData.orgId)
```

backend/src/ee/services/saml-config/saml-config-service.ts
```diff
@@ -535,14 +540,13 @@ export const samlConfigServiceFactory = ({
...
         const foundUser = await userDAL.findById(userAlias.userId, tx);
-        const [orgMembership] = await orgDAL.findMembership(
-          {
-            [`${TableName.Membership}.actorUserId` as "actorUserId"]: userAlias.userId,
-            [`${TableName.Membership}.scopeOrgId` as "scopeOrgId"]: orgId,
-            [`${TableName.Membership}.scope` as "scope"]: AccessScope.Organization
-          },
-          { tx }
-        );
+        const orgMembership = await orgDAL.findEffectiveOrgMembership({
```
akhilmhdh: I think we can remove this and go back to old check. This is because SAML is not possible in sub-org

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -408,16 +409,15 @@ export const scimServiceFactory = ({
       let orgMembership: TMemberships;
       if (userAlias) {
         user = await userDAL.findById(userAlias.userId, tx);
-        orgMembership = await membershipUserDAL.findOne(
-          {
-            actorUserId: user.id,
-            scope: AccessScope.Organization,
-            scopeOrgId: orgId
-          },
+        const effectiveMembership = await orgDAL.findEffectiveOrgMembership({
```
akhilmhdh, line 412: Same applies here - scim is not support in suborg and users can't create a config

backend/src/services/auth/auth-login-service.ts
```diff
@@ -578,11 +571,11 @@ export const authLoginServiceFactory = ({
         });
       }
 
-      // Check user membership in the root organization
-      const rootOrgMembership = await membershipUserDAL.findOne({
-        actorUserId: user.id,
-        scopeOrgId: selectedOrg.rootOrgId,
-        scope: AccessScope.Organization,
+      // Check user membership in the root organization (direct or via group)
+      const rootOrgMembership = await orgDAL.findEffectiveOrgMembership({
```
akhilmhdh: Root org should always have a direct membership

backend/src/services/org/org-dal.ts
```diff
@@ -628,6 +659,81 @@ export const orgDALFactory = (db: TDbClient) => {
...
+        void viaGroup
+          .join(
+            TableName.UserGroupMembership,
+            `${TableName.UserGroupMembership}.groupId`,
+            `${TableName.Membership}.actorGroupId`
+          )
+          .where(`${TableName.UserGroupMembership}.userId`, dto.actorId);
+      }
+
+      const first = await viaGroup.first();
```
akhilmhdh: Some small changes in this query - i think we can optimize this to be a single query. 

We first fetch all memberships available for the organization and filter out ones with direct and one with group. Like how we do it in permission service. Just that we don't do the permission filtering

Another thing is we can return all the memberships available for the user as return. This could be useful in some operations. The user could have multiple memberships like array of memberships because
1. Direct and a group membership
2. Multiple group membership in a sub-org

backend/src/services/org/org-dal.ts
```diff
@@ -165,23 +165,54 @@ export const orgDALFactory = (db: TDbClient) => {
...
-              void qb.andWhere(`${TableName.Membership}.actorIdentityId`, dto.actorId);
-            } else {
-              void qb.andWhere(`${TableName.Membership}.actorUserId`, dto.actorId);
-            }
-          });
+        // Correlate subquery: membership.scopeOrgId = outer organizations.id (quoted for PostgreSQL camelCase)
+        const scopeOrgIdEqOrgId = `"${TableName.Membership}"."scopeOrgId" = "${TableName.Organization}"."id"`;
+        void query.where((qb) => {
+          // Direct: actor has membership in the sub-org (user or identity)
+          void qb.whereExists(
```
akhilmhdh: The problem with `whereExists` query is it does a full table scan

akhilmhdh: I think we can do something like the permission dal here as well.

First keep this a subquery
```
- Find all sub-orgs
- Find all memberships of the sub-org for the user via direct or via group - refer permission dal change
- Return all sub-org ids
```

Now pass this subquery as a where filter of ids in top query. This would filter and keep the pagination correct as well. What do you think?

akhilmhdh:
Application testing feedbacks

- Would be great like identity in sub-org group section we have `Managed By` section in table that says this group is managed by which scope. Like wise in detail view of a group as well.
image

- Projects are not visible. I have created group with engineering and assign to suborg and then connected to project in suborg. But that project is not listed when i login to that suborg via the user of the group
Screenshot 2026-02-16 at 2 50 14 PM

- I am not visible in `All Projects` section as well. This part it's not saying i am part of the project
Screenshot 2026-02-16 at 2 51 51 PM

Please test the sub-org project flow for both user and identity as well. I am able to do the operations but the list ones are broken. I was able to do operations on the project - so it seems only the list is broken. Do it test for identity as well. Do a complete workflow - of creating this, adding groups, consuming operations in a project for both user and identity.

backend/src/ee/services/group/group-service.ts
IgorHorta: agreed!
will add it to the message!

backend/src/ee/services/group/group-dal.ts
```diff
@@ -595,18 +604,17 @@ export const groupDALFactory = (db: TDbClient) => {
...
-   * Used to block delete when this group is the source of linked/copied groups.
+   * Returns the groups that reference the given group via sourceGroupId (self-referential FK).
+   * Used to block delete when this group is the source of linked/copied groups; error message can list them.
    */
-  const hasGroupsReferencingGroup = async (groupId: string, tx?: Knex): Promise<boolean> => {
+  const getGroupsReferencingGroup = async (groupId: string, tx?: Knex): Promise<{ id: string; name: string }[]> => {
     try {
-      const row = await (tx || db.replicaNode())(TableName.Groups)
+      const docs = await (tx || db.replicaNode())(TableName.Groups)
         .whereRaw('"sourceGroupId" = ?', [groupId])
```
akhilmhdh: Small question: why is this raw?

backend/src/ee/services/group/group-dal.ts
```diff
@@ -472,9 +472,18 @@ export const groupDALFactory = (db: TDbClient) => {
     orderDirection?: OrderByDirection;
   }) => {
     try {
+      // Include projects in the group's org and in any child sub-orgs (so root-org groups show inherited sub-org projects)
```
akhilmhdh: This is a mistake - only render project that part of groups in that org/suborg. These information should not go outside.

akhilmhdh (changes requested):

akhilmhdh:
## Application Testing Feedback

### 1. Sub-org project visibility leak from root
Users in a sub-org should not see projects belonging to the root org. Currently, if a group has a project in root, members can see it when viewing the sub-org.

<details>
<summary>Screenshot</summary>

Screenshot

</details>

### 2. Machine identity – missing project listing fix
The project listing fix was not applied to machine identities. They still show the old behavior.

<details>
<summary>Screenshot</summary>

Screenshot

</details>

### 3. Empty membership list for group-created projects in sub-org
When a project is created in a sub-org where access comes solely through group membership, the project's member listing appears empty — no users are shown.

<details>
<summary>Screenshot</summary>

Screenshot

</details>

### 4. Group deletion failing
Deleting a group throws an error. The use of `raw` in the query is unclear, and there appears to be a reference to a non-existent column.

<details>
<summary>Screenshot</summary>

Screenshot

</details>

### 5. Group re-linking restores previously removed project access
**Steps to reproduce:** Link a group from root → add a project → remove the group → re-add the group.
**Result:** All previously removed projects reappear.
**Expected:** Project membership cleanup should persist through group removal/re-addition.

<details>
<summary>Screenshot</summary>

Screenshot

</details>

### 6. User retains direct project membership after losing suborg access
When a user's only link to a sub-org (via a group) is removed, their direct project memberships in that sub-org are not cleaned up. When the group is re-added, those memberships resurface. Project membership should be revoked when the user's last access path to the org/sub-org is removed.

Also ensure this cleanup happens for identities as well as it's not happening for the same reason

<details>
<summary>Screenshot</summary>

Screenshot

</details>

backend/src/ee/services/group/group-dal.ts
IgorHorta: fixed

backend/src/ee/services/group/group-service.ts
```diff
@@ -331,8 +331,29 @@ export const groupServiceFactory = ({
     const { group } = groupMembership;
 
     if (group.orgId !== actorOrgId) {
-      // Linked group: unlink by deleting the membership in this org only
+      // Linked group: unlink by deleting the org membership and all project memberships for this group in this sub-org
       await groupDAL.transaction(async (tx) => {
+        // Find all project-level memberships for this group in the sub-org
```
IgorHorta: lets make sure we check if this is suborg

backend/src/ee/services/group/group-service.ts
```diff
@@ -331,8 +331,29 @@ export const groupServiceFactory = ({
...
+          {
+            actorGroupId: id,
+            scopeOrgId: actorOrgId,
+            scope: AccessScope.Project
+          },
+          { tx }
+        );
+
+        // Delete roles and memberships for each project membership
+        if (projectMemberships.length > 0) {
```
IgorHorta: use in operation, we dont need n trasactions here

akhilmhdh:
@greptile this pr has made a lot of changes from your last review. Do a complete review once more

backend/src/ee/services/group/group-service.ts
```diff
@@ -289,12 +299,254 @@ export const groupServiceFactory = ({
...
+        const effectiveRows = await membershipGroupDAL.findEffectiveInOrg({
+          scopeOrgId: subOrgId,
+          excludeMembershipIds: [],
+          userIds,
+          identityIds,
+          groupIds: remainingGroupIds,
+          tx
+        });
+
+        const userIdsWithAccess = new Set<string>();
```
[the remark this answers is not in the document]
IgorHorta: group exclusion is at suborg level

akhilmhdh:
Good job @IgorHorta

PR #5562, opened 2026-03-02
feat(app-connections): credential rotations

varonix0:
## Context

This PR introduces a new concept called credential rotations to app connections. It's based on the secret rotation v2 work by @scott-ray-wilson, and allows for wider expansion later on (currently it is only supported on the Azure Key Vault app connection using the client secrets auth method).

Essentially, it is now possible to give Infisical full control over your client secret lifecycle, meaning Infisical will manage and periodically rotate the client secret.
## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/e2e-test/mocks/keystore.ts
+++ b/backend/e2e-test/mocks/keystore.ts
@@ -21,6 +21,11 @@ export const mockKeyStore = (): TKeyStoreFactory => {
       store[key] = value;
       return "OK";
     },
+    setItemWithExpiryNX: async (key, _expiryInSeconds, value) => {
+      if (store[key] !== undefined) return null;
+      store[key] = value;
+      return "OK";
+    },
     deleteItem: async (key) => {
       delete store[key];
       return 1;
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -99,7 +99,9 @@ export const KeyStorePrefixes = {
   ProjectSSEConnection: (projectId: string, connectionId: string) =>
     `project-sse-conn:${projectId}:${connectionId}` as const,
 
-  ProjectDeleteLock: (projectId: string) => `project-delete-lock-${projectId}` as const
+  ProjectDeleteLock: (projectId: string) => `project-delete-lock-${projectId}` as const,
+
+  TelemetryIdentifyIdentity: (dedupKey: string) => `telemetry-identify-identity:${dedupKey}` as const
 };
 
 export const KeyStoreTtls = {
@@ -110,7 +112,8 @@ export const KeyStoreTtls = {
   ProjectPermissionDalVersionTtl: "15m", // Project permission DAL version TTL
   MfaSessionInSeconds: 300, // 5 minutes
   WebAuthnChallengeInSeconds: 300, // 5 minutes
-  ProjectSSEConnectionTtlSeconds: 180 // Must be > heartbeat interval (60s) * 2
+  ProjectSSEConnectionTtlSeconds: 180, // Must be > heartbeat interval (60s) * 2
+  TelemetryIdentifyIdentityInSeconds: 600 // 10 minutes
 };
 
 type TDeleteItems = {
@@ -140,6 +143,12 @@ export type TKeyStoreFactory = {
     value: string | number | Buffer,
     prefix?: string
   ) => Promise<"OK">;
+  setItemWithExpiryNX: (
+    key: string,
+    expiryInSeconds: number | string,
+    value: string | number | Buffer,
+    prefix?: string
+  ) => Promise<"OK" | null>;
   deleteItem: (key: string) => Promise<number>;
   deleteItemsByKeyIn: (keys: string[]) => Promise<number>;
   deleteItems: (arg: TDeleteItems) => Promise<number>;
@@ -220,6 +229,13 @@ export const keyStoreFactory = (
     value: string | number | Buffer,
     prefix?: string
   ) => primaryRedis.set(prefix ? `${prefix}:${key}` : key, value, "EX", expiryInSeconds);
+
+  const setItemWithExpiryNX = async (
+    key: string,
+    expiryInSeconds: number | string,
+    value: string | number | Buffer,
+    prefix?: string
+  ) => primaryRedis.set(prefix ? `${prefix}:${key}` : key, value, "EX", expiryInSeconds, "NX");
 
   const deleteItem = async (key: string) => primaryRedis.del(key);
 
@@ -376,6 +392,7 @@ export const keyStoreFactory = (
     getItem,
     setExpiry,
     setItemWithExpiry,
+    setItemWithExpiryNX,
     deleteItem,
     deleteItems,
     incrementBy,
--- a/backend/src/keystore/memory.ts
+++ b/backend/src/keystore/memory.ts
@@ -22,6 +22,11 @@ export const inMemoryKeyStore = (): TKeyStoreFactory => {
       store[key] = value;
       return "OK";
     },
+    setItemWithExpiryNX: async (key, _expiryInSeconds, value) => {
+      if (store[key] !== undefined) return null;
+      store[key] = value;
+      return "OK";
+    },
     deleteItem: async (key) => {
       delete store[key];
       return 1;
--- a/backend/src/server/plugins/auth/inject-identity.ts
+++ b/backend/src/server/plugins/auth/inject-identity.ts
@@ -241,6 +241,17 @@ export const injectIdentity = fp(
           }
 
           requestContext.set("identityAuthInfo", identityAuthInfo);
+
+          // Fire-and-forget: enrich PostHog person record for this machine identity
+          void server.services.telemetry
+            .identifyIdentity(identity.identityId, {
+              name: identity.identityName,
+              authMethod: identity.authMethod
+            })
+            .catch((error) => {
+              req.log.error(error, `Failed to enrich PostHog identity [identityId=${identity.identityId}]`);
+            });
+
           break;
         }
         case AuthMode.SERVICE_TOKEN: {
--- a/backend/src/server/routes/v1/dashboard-router.ts
+++ b/backend/src/server/routes/v1/dashboard-router.ts
@@ -1524,7 +1524,8 @@ export const registerDashboardRouter = async (server: FastifyZodProvider) => {
             environment,
             secretPath,
             channel: getUserAgentType(req.headers["user-agent"]),
-            ...req.auditLogInfo
+            ...req.auditLogInfo,
+            actorType: req.permission.type
           }
         });
       }
--- a/backend/src/server/routes/v1/identity-alicloud-auth-router.ts
+++ b/backend/src/server/routes/v1/identity-alicloud-auth-router.ts
@@ -1,16 +1,19 @@
 import RE2 from "re2";
 import { z } from "zod";
 
-import { IdentityAlicloudAuthsSchema } from "@app/db/schemas";
+import { IdentityAlicloudAuthsSchema, IdentityAuthMethod } from "@app/db/schemas";
 import { EventType } from "@app/ee/services/audit-log/audit-log-types";
 import { ALICLOUD_AUTH, ApiDocsTags } from "@app/lib/api-docs";
+import { logger } from "@app/lib/logger";
 import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
 import { slugSchema } from "@app/server/lib/schemas";
+import { getTelemetryDistinctId } from "@app/server/lib/telemetry";
 import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
 import { AuthMode } from "@app/services/auth/auth-type";
 import { TIdentityTrustedIp } from "@app/services/identity/identity-types";
 import { validateArns } from "@app/services/identity-alicloud-auth/identity-alicloud-auth-validators";
 import { isSuperAdmin } from "@app/services/super-admin/super-admin-fns";
+import { PostHogEventTypes } from "@app/services/telemetry/telemetry-types";
 
 export const registerIdentityAliCloudAuthRouter = async (server: FastifyZodProvider) => {
   server.route({
@@ -91,6 +94,21 @@ export const registerIdentityAliCloudAuthRouter = async (server: FastifyZodProvi
           }
         }
       });
+
+      void server.services.telemetry
+        .sendPostHogEvents({
+          event: PostHogEventTypes.MachineIdentityLogin,
+          distinctId: `identity-${identityAliCloudAuth.identityId}`,
+          organizationId: identity.orgId,
+          properties: {
+            identityId: identityAliCloudAuth.identityId,
+            orgId: identity.orgId,
+            authMethod: IdentityAuthMethod.ALICLOUD_AUTH
+          }
+        })
+        .catch((error) => {
+          logger.error(error, `Failed to send telemetry event [identityId=${identityAliCloudAuth.identityId}]`);
+        });
 
       return {
         accessToken,
@@ -190,6 +208,21 @@ export const registerIdentityAliCloudAuthRouter = async (server: FastifyZodProvi
         }
       });
 
+      void server.services.telemetry
+        .sendPostHogEvents({
+          event: PostHogEventTypes.MachineIdentityAuthMethodAttached,
+          distinctId: getTelemetryDistinctId(req),
+          organizationId: identityAliCloudAuth.orgId,
+          properties: {
+            identityId: identityAliCloudAuth.identityId,
+            orgId: identityAliCloudAuth.orgId,
+            authMethod: IdentityAuthMethod.ALICLOUD_AUTH
+          }
+        })
+        .catch((error) => {
+          logger.error(error, `Failed to send telemetry event [identityId=${identityAliCloudAuth.identityId}]`);
+        });
+
       return { identityAliCloudAuth };
     }
   });
@@ -282,6 +315,21 @@ export const registerIdentityAliCloudAuthRouter = async (server: FastifyZodProvi
           }
         }
       });
+
+      void server.services.telemetry
+        .sendPostHogEvents({
+          event: PostHogEventTypes.MachineIdentityAuthMethodUpdated,
+          distinctId: getTelemetryDistinctId(req),
+          organizationId: identityAliCloudAuth.orgId,
+          properties: {
+            identityId: identityAliCloudAuth.identityId,
+            orgId: identityAliCloudAuth.orgId,
+            authMethod: IdentityAuthMethod.ALICLOUD_AUTH
+          }
+        })
+        .catch((error) => {
+          logger.error(error, `Failed to send telemetry event [identityId=${identityAliCloudAuth.identityId}]`);
+        });
 
       return { identityAliCloudAuth };
     }
@@ -381,6 +429,21 @@ export const registerIdentityAliCloudAuthRouter = async (server: FastifyZodProvi
           }
         }
       });
+
+      void server.services.telemetry
+        .sendPostHogEvents({
+          event: PostHogEventTypes.MachineIdentityAuthMethodRevoked,
+          distinctId: getTelemetryDistinctId(req),
+          organizationId: identityAliCloudAuth.orgId,
+          properties: {
+            identityId: identityAliCloudAuth.identityId,
+            orgId: identityAliCloudAuth.orgId,
+            authMethod: IdentityAuthMethod.ALICLOUD_AUTH
+          }
+        })
+        .catch((error) => {
+          logger.error(error, `Failed to send telemetry event [identityId=${identityAliCloudAuth.identityId}]`);
+        });
 
       return { identityAliCloudAuth };
     }
--- a/backend/src/server/routes/v1/identity-aws-iam-auth-router.ts
+++ b/backend/src/server/routes/v1/identity-aws-iam-auth-router.ts
@@ -1,10 +1,12 @@
 import { z } from "zod";
 
-import { IdentityAwsAuthsSchema } from "@app/db/schemas";
+import { IdentityAuthMethod, IdentityAwsAuthsSchema } from "@app/db/schemas";
 import { EventType } from "@app/ee/services/audit-log/audit-log-types";
 import { ApiDocsTags, AWS_AUTH } from "@app/lib/api-docs";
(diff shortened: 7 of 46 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/e9e9990d-61f7-4003-af6a-b87503dc9736?refs=902ddbd6d504e5318ba04b56b97995fe723af8ed&source=prChecksComment) |

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/shared/CredentialRotationForm.tsx
```diff
@@ -0,0 +1,109 @@
+import { ReactNode } from "react";
+import { Controller, useFormContext } from "react-hook-form";
+import { format, setHours, setMinutes } from "date-fns";
+import { twMerge } from "tailwind-merge";
+
+import { FormControl, Input, Switch } from "@app/components/v2";
+import { getRotateAtLocal } from "@app/helpers/secretRotationsV2";
+
+type Props = {
+  renderExtraFields?: ReactNode;
```
scott-ray-wilson: nit: I kind of feel like this should just be `children`?

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/shared/CredentialRotationForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+              errorText={error?.message}
+            >
+              <Switch
+                className="bg-mineshaft-400/50 shadow-inner data-[state=checked]:bg-green/80"
+                id="platform-managed"
+                thumbClassName="bg-mineshaft-800"
+                isChecked={value}
+                onCheckedChange={onChange}
+              >
+                <p className="w-[13.6rem]">Automatic Credential Rotation</p>
```
scott-ray-wilson: I think this is a bit wider than needed; the gap between the label and switch looks a little too large

Image

backend/src/db/migrations/20260220000000_app-connection-credential-rotation.ts
```diff
@@ -0,0 +1,43 @@
...
+      t.jsonb("rotateAtUtc").notNullable();
+      t.string("rotationStatus").notNullable();
+      t.timestamp("lastRotationAttemptedAt").nullable();
+      t.timestamp("lastRotatedAt").nullable();
+      t.binary("encryptedLastRotationMessage").nullable();
+      t.string("lastRotationJobId").nullable();
+      t.timestamp("nextRotationAt").nullable();
+      t.integer("activeIndex").notNullable().defaultTo(0);
+      t.binary("encryptedGeneratedCredentials").notNullable();
+      t.timestamps(true, true, true);
```
scott-ray-wilson, line 32: I noticed we're including timestamps but not the updated at trigger

backend/src/server/routes/v1/app-connection-routers/app-connection-endpoints.ts
```diff
@@ -25,15 +26,19 @@ export const registerAppConnectionEndpoints = <T extends TAppConnection, I exten
...
+    rotation?: TCreateAppConnectionCredentialRotationSchema | null;
   }>;
   updateSchema: z.ZodType<{
     name?: string;
     credentials?: I["credentials"];
     description?: string | null;
     isPlatformManagedCredentials?: boolean;
     gatewayId?: string | null;
+    isAutoRotationEnabled?: boolean | null;
+    rotation?: Partial<TCreateAppConnectionCredentialRotationSchema> | null;
```
scott-ray-wilson, line 41: more just curious why these addtions allow null when they're not nullable

backend/src/services/app-connection/azure-client-secrets/azure-client-secrets-connection-schemas.ts
```diff
@@ -46,7 +46,8 @@ export const AzureClientSecretsConnectionClientSecretInputCredentialsSchema = z.
     .uuid()
     .trim()
     .min(1, "Tenant ID required")
-    .describe(AppConnections.CREDENTIALS.AZURE_CLIENT_SECRETS.tenantId)
+    .describe(AppConnections.CREDENTIALS.AZURE_CLIENT_SECRETS.tenantId),
+  clientSecretKeyId: z.string().uuid().trim().optional()
```
scott-ray-wilson: missing describe for documentation

backend/src/services/app-connection/credential-rotation/providers/azure-client-secret/azure-client-secret-credential-rotation-fns.ts
```diff
@@ -0,0 +1,459 @@
...
+
+    const updatedCredentials = {
+      ...azureCredentials,
+      clientSecret: newCredential.clientSecret,
+      applicationObjectId
+    };
+
+    // Verify the new credential works by obtaining a fresh Graph API token with it.
+    // getGraphApiToken has built-in retry logic for propagation delays (AADSTS7000215).
+    const freshAccessToken = await getGraphApiToken(updatedCredentials);
```
scott-ray-wilson, line 395: should we move this to the create so that both initial and all further generated credentials are validated?

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-fns.ts
```diff
@@ -0,0 +1,288 @@
...
+
+export const encryptCredentialRotationGeneratedCredentials = async ({
+  orgId,
+  generatedCredentials,
+  kmsService
+}: {
+  orgId: string;
+  generatedCredentials: TAppConnectionCredentialRotationGeneratedCredentials;
+  kmsService: Pick<TKmsServiceFactory, "createCipherPairWithDataKey">;
+}) => {
```
scott-ray-wilson, line 113: shouldn't this follow the same pattern as app connection credentials and use the KMS at the level owned?

applies to all encryption methods

```
 const { encryptor } = await kmsService.createCipherPairWithDataKey(
    projectId
      ? {
          type: KmsDataKey.SecretManager,
          projectId
        }
      : {
          type: KmsDataKey.Organization,
          orgId
        }
  );
```

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-queue.ts
```diff
@@ -0,0 +1,160 @@
...
+      } catch (error) {
+        logger.error(error, "credentialRotationQueue: Queue Rotations Error:");
+        throw error;
+      }
+    } else if (job.name === QueueJobs.AppConnectionCredentialRotationSendNotification) {
+      const payload = job.data as TAppConnectionCredentialRotationSendNotificationJobPayload;
+      try {
+        logger.info(`credentialRotationQueue: Sending Failure Notification [connectionId=${payload.connectionId}]`);
+
+        // Notification sending can be expanded later with email + in-app notifications
```
scott-ray-wilson: I feel like email should be included - we can easily copy the secret rotation email template

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-schemas.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const UpdateAppConnectionCredentialRotationSchema = z.object({
+  rotationInterval: z.number().int().min(1).max(365).optional(),
+  rotateAtUtc: z
+    .object({
+      hours: z.number().int().min(0).max(23),
+      minutes: z.number().int().min(0).max(59)
+    })
+    .optional(),
+  isAutoRotationEnabled: z.boolean().optional()
```
scott-ray-wilson: why is this included here? also the only reference to this schema omits it 🤔

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-schemas.ts
```diff
@@ -0,0 +1,22 @@
...
+  rotateAtUtc: AppConnectionCredentialRotationRotateAtUtcSchema
+});
+
+export const UpdateAppConnectionCredentialRotationSchema = z.object({
+  rotationInterval: z.number().int().min(1).max(365).optional(),
+  rotateAtUtc: z
+    .object({
+      hours: z.number().int().min(0).max(23),
+      minutes: z.number().int().min(0).max(59)
+    })
```
scott-ray-wilson: nit: use the schema above with optional

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+  azureClientSecretRotationProviderFactory
+} from "./providers/azure-client-secret";
+
+const MAX_GENERATED_CREDENTIALS_LENGTH = 2;
+
+const STRATEGY_MAP: Record<
+  AppConnectionCredentialRotationStrategy,
+  {
+    app: TAppConnection["app"]; // azure-key-vault
+    method: TAppConnection["method"]; // client-secret
```
scott-ray-wilson: nit: comments needed? might cause confusion down the road

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+
+    const credentials = await getConnectionClientCredentials(connection);
+
+    const { strategyConfig, generatedCredentials, updatedCredentials } = await provider.issueInitialCredentials(
+      credentials,
+      rotationInterval
+    );
+
+    // Update the connection to use the new credential
+    await updateConnectionCredentials(connectionId, updatedCredentials as TAppConnection["credentials"], tx);
```
scott-ray-wilson: so right now, the azure client secret revokes the credentials on issuance of the new credentials - but this can be problematic if `updateConnectionCredentials` - we need to wait to revoke until the transaction completes on our side to ensure we don't end up with a dead app connection with revoked old credentials

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+          lastRotationAttemptedAt: now,
+          nextRotationAt,
+          activeIndex: 0,
+          encryptedGeneratedCredentials: encryptedGeneratedCreds
+        },
+        trx
+      );
+    };
+
+    const rotation = tx ? await create(tx) : await appConnectionDAL.transaction(create);
```
scott-ray-wilson, line 220: this only runs in a transaction with the update connection credentials above if a tx is passed - we should ensure they are always transactional coupled, whether it's passed or not

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+    if (rotateAtUtc !== undefined) {
+      updateData.rotateAtUtc = JSON.stringify(rotateAtUtc);
+    }
+
+    // Recalculate nextRotationAt if schedule changed
+    if (rotationInterval !== undefined || rotateAtUtc !== undefined) {
+      const resolvedRotateAtUtc = rotateAtUtc || (existingRotation.rotateAtUtc as { hours: number; minutes: number });
+      const resolvedInterval = rotationInterval ?? existingRotation.rotationInterval;
+
+      if (existingRotation.lastRotatedAt) {
```
scott-ray-wilson, line 294: shouldn't this always be defined? otherwise this would never get picked up for rotation right?

backend/src/db/migrations/20260220000000_app-connection-credential-rotation.ts
```diff
@@ -0,0 +1,43 @@
...
+      t.string("strategy").notNullable();
+      t.binary("encryptedStrategyConfig").notNullable();
+      t.integer("rotationInterval").notNullable();
+      t.jsonb("rotateAtUtc").notNullable();
+      t.string("rotationStatus").notNullable();
+      t.timestamp("lastRotationAttemptedAt").nullable();
+      t.timestamp("lastRotatedAt").nullable();
+      t.binary("encryptedLastRotationMessage").nullable();
+      t.string("lastRotationJobId").nullable();
+      t.timestamp("nextRotationAt").nullable();
```
scott-ray-wilson, line 29: this shouldn't be nullable right? because there isn't a rotation config if auto rotation isn't enabled

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+        connectionId,
+        queuedAt: new Date(),
+        isManualRotation: true
+      },
+      {
+        jobId: `app-connection-credential-rotation-manual-${rotation.id}-${Date.now()}`,
+        removeOnFail: true,
+        removeOnComplete: true
+      }
+    );
```
scott-ray-wilson, line 361: why not just rotate directly instead of job for manual? we do this for rotations and it gives realtime feedback - just a thought

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+      return null;
+    };
+
+    // If disabling auto rotation, delete the rotation config
+    if (isAutoRotationEnabled === false) {
+      if (tx) {
+        await update(tx);
+      } else {
+        await appConnectionDAL.transaction(update);
+      }
```
scott-ray-wilson: I feel like this is not ideal UX - if I just want to disable rotation temporarily, when I want to re-enable I now have to go get parameters again

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+          newCredential
+        ) as TAppConnection["credentials"];
+        await updateConnectionCredentials(connectionId, updatedCredentials);
+
+        // Step 3: Revoke old credential at inactive index (best-effort, non-fatal)
+        // If this fails, we have 3 secrets temporarily — they'll be cleaned up on the next cycle.
+        if (inactiveCredential) {
+          try {
+            logger.info(
+              `credentialRotation: Revoking old credential keyId=${inactiveCredential.keyId} at index=${inactiveIndex} [rotationId=${rotationId}]`
```
scott-ray-wilson: nit: explicit reference to a provider property in generic fn - will need to be removed

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
```diff
@@ -0,0 +1,541 @@
...
+        const nextRotationAt = calculateNextRotationAt({
+          rotateAtUtc,
+          isAutoRotationEnabled,
+          rotationInterval,
+          rotationStatus: AppConnectionCredentialRotationStatus.Success,
+          isManualRotation: options.isManualRotation ?? false,
+          lastRotatedAt: now
+        });
+
+        await appConnectionCredentialRotationDAL.updateById(rotationId, {
```
scott-ray-wilson: this should be transactional with the above connection credential update

backend/src/services/app-connection/app-connection-schemas.ts
scott-ray-wilson, line 1: missing describes on new properties

backend/src/services/app-connection/app-connection-service.ts
```diff
@@ -647,6 +688,57 @@ export const appConnectionServiceFactory = ({
         updatedConnection = await updateConnection(updatedCredentials);
       }
 
+      // Handle rotation updates
+      if (isAutoRotationEnabled !== undefined || rotation) {
```
scott-ray-wilson: this whole block should be transactional with the base connection update

backend/src/services/app-connection/app-connection-service.ts
```diff
@@ -872,6 +964,50 @@ export const appConnectionServiceFactory = ({
...
+        actor: actor.type,
+        actorId: actor.id,
+        projectId: appConnection.projectId,
+        actorAuthMethod: actor.authMethod,
+        actorOrgId: actor.orgId,
+        actionProjectType: ActionProjectType.Any
+      });
+
+      ForbiddenError.from(permission).throwUnlessCan(
+        ProjectPermissionAppConnectionActions.Edit,
```
scott-ray-wilson: think this should be a new permission inline with rotations

backend/src/services/app-connection/app-connection-service.ts
```diff
@@ -872,6 +964,50 @@ export const appConnectionServiceFactory = ({
...
+        actorId: actor.id,
+        actor: actor.type,
+        orgId: appConnection.orgId,
+        actorOrgId: actor.orgId,
+        actorAuthMethod: actor.authMethod,
+        scope: OrganizationActionScope.Any
+      });
+
+      ForbiddenError.from(permission).throwUnlessCan(
+        OrgPermissionAppConnectionActions.Edit,
```
scott-ray-wilson: ditto

backend/src/server/routes/v1/app-connection-routers/app-connection-endpoints.ts
varonix0: @scott-ray-wilson was trying to make TypeScript happy. The type translations are pretty convoluted here, so I ended up going with the same approach as for gateway's

backend/src/services/app-connection/credential-rotation/providers/azure-client-secret/azure-client-secret-credential-rotation-fns.ts
varonix0: Thought about that too, but wanted to avoid doing this on create as there's quite a few things already happening on create. We risk the request potentially timing out by doing this on create due to the exponential back-off retry mechanism

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-fns.ts
varonix0: Great catch

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
varonix0: Good catch

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
varonix0: It's nullable on DB level, hence the check. When the rotation is created it's null until the first rotation cycle is completed

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
varonix0: @scott-ray-wilson I think it's better to re-use the existing queue. the rotation can also take time (due to exponential back-off strategy as an example), and if we want retries to work I would say this is ideal.

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
varonix0: What do you mean by that exactly? In the section you commented on I don't see anything alarming

backend/src/services/app-connection/app-connection-service.ts
varonix0: Hmm I don't think so. I feel like this falls within the scope of the app connection itself. This is not a new concept per say, just an extension of the existing app connection functionality - we can sync about this tomorrow morning PST

backend/src/services/app-connection/app-connection-service.ts
varonix0: Same as above

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
scott-ray-wilson: this is in the generic app connection service, but you're explicitly referencing a property from the azure credential schema `inactiveCredential.keyId`

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-schemas.ts
```diff
@@ -1,22 +1,22 @@
 import { z } from "zod";
 
 export const AppConnectionCredentialRotationRotateAtUtcSchema = z.object({
-  hours: z.number().int().min(0).max(23),
-  minutes: z.number().int().min(0).max(59)
+  hours: z.number().int().min(0).max(23).describe("The hour (0-23) at which to rotate."),
+  minutes: z.number().int().min(0).max(59).describe("The minute (0-59) at which to rotate.")
```
scott-ray-wilson: we may want to mention UTC

backend/src/services/app-connection/credential-rotation/app-connection-credential-rotation-service.ts
scott-ray-wilson: I think we should make `tx` required - right now it is always passed, which is the only way this whole function acts as a single transaction (which it should) - so if anyone used this without a transaction it wouldn't be atomic - rather than refactoring requiring the tx would address this

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionsTable.tsx
```diff
@@ -233,6 +235,31 @@ export const AppConnectionsTable = ({ projectId, projectType }: Props) => {
...
+          createNotification({
+            text: `Auto-rotation ${enabling ? "enabled" : "disabled"}`,
+            type: "success"
+          });
+        },
+        onError: (error) => {
+          createNotification({
+            text: `Failed to ${enabling ? "enable" : "disable"} auto-rotation: ${error.message}`,
+            type: "error"
+          });
```
scott-ray-wilson: this will be a double error notification

PR #5574, opened 2026-03-03
feat(secrets-17): add clickhouse dynamic secret provider

PrestigePvP:
## Context

- Add Clickhouse as a dynamic secret provider

## Steps to verify the change

- Go to Dynamic Secrets
- Add a new Clickhouse Dynamic Secret 

## Type

- [ ] Fix
- [X] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [X] Tested locally

### DIFF
--- a/backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
+++ b/backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
@@ -109,18 +109,10 @@ export const ClickhouseProvider = ({
   const validateProviderInputs = async (inputs: unknown) => {
     const providerInputs = await DynamicSecretClickhouseSchema.parseAsync(inputs);
 
-    let { host } = providerInputs;
-    let hostProtocol: string | undefined;
-    if (host.startsWith("https://")) {
-      host = host.slice("https://".length);
-      hostProtocol = "https";
-    } else if (host.startsWith("http://")) {
-      host = host.slice("http://".length);
-      hostProtocol = "http";
-    }
+    const sanitizedHost = providerInputs.host.replace(/^https?:\/\//, "");
 
     const [hostIp] = await verifyHostInputValidity({
-      host,
+      host: sanitizedHost,
       isGateway: Boolean(providerInputs.gatewayId),
       isDynamicSecret: true
     });
@@ -139,14 +131,15 @@ export const ClickhouseProvider = ({
       allowedExpressions: (val) => ["username", "database"].includes(val)
     });
 
-    return { ...providerInputs, host, hostIp, hostProtocol };
+    return { ...providerInputs, hostIp };
   };
 
-  const $getClient = async (
-    providerInputs: z.infer<typeof DynamicSecretClickhouseSchema> & { hostIp: string; hostProtocol?: string }
-  ) => {
-    const protocol = providerInputs.hostProtocol ?? (providerInputs.ca ? "https" : "http");
-    const url = `${protocol}://${providerInputs.host}:${providerInputs.port}`;
+  const $getClient = async (providerInputs: z.infer<typeof DynamicSecretClickhouseSchema> & { hostIp: string }) => {
+    let url = `${providerInputs.host}:${providerInputs.port}`;
+
+    if (!url.includes("://")) {
+      url = `http${providerInputs.ca ? "s" : ""}://${url}`;
+    }
 
     const client = createClient({
       url,
--- a/docs/documentation/platform/dynamic-secrets/clickhouse.mdx
+++ b/docs/documentation/platform/dynamic-secrets/clickhouse.mdx
@@ -11,6 +11,15 @@ The Infisical ClickHouse dynamic secret allows you to generate ClickHouse databa
 
 Create a user with the required permissions in your ClickHouse instance. This user will be used to create new accounts on-demand.
 
+```sql
+CREATE USER 'infisical' IDENTIFIED WITH sha256_password BY '<strong-password>';
+GRANT CREATE USER, DROP USER, GRANT OPTION ON *.* TO 'infisical';
+```
+
+<Note>
+  The `GRANT OPTION` privilege is required so that the service user can grant permissions to the dynamically created users.
+</Note>
+
 ## Set up Dynamic Secrets with ClickHouse
 
 <Steps>


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/161ce650-7831-4bd8-a0ea-89ba6f76ba78?refs=d8969785db000e76e5ad4b4628f34483a1f00b5f&source=prChecksComment) |

docs/documentation/platform/dynamic-secrets/clickhouse.mdx
```diff
@@ -0,0 +1,159 @@
+---
```
varonix0, line 1: Ignore placement of comment:

Clickhouse is missing from the explore dynamic secrets page:

Image

backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
```diff
@@ -0,0 +1,399 @@
...
+    validateHandlebarTemplate("ClickHouse revoke", providerInputs.revocationStatement, {
+      allowedExpressions: (val) => ["username", "database"].includes(val)
+    });
+
+    return { ...providerInputs, hostIp };
+  };
+
+  const $getClient = async (providerInputs: z.infer<typeof DynamicSecretClickhouseSchema> & { hostIp: string }) => {
+    const protocol = (providerInputs.sslEnabled ?? Boolean(providerInputs.ca)) ? "https" : "http";
+    const url = `${protocol}://${providerInputs.host}:${providerInputs.port}`;
```
varonix0: This doesn't work with ClickHouse Cloud - I tried setting up a ClickHouse Cloud instance, which gives me the following connection details:

```
  https://<instance-slug>.us-east-1.aws.clickhouse.cloud:8443
```

Few things:
1. It comes with the protocol directly in the connection input. Users might be confused if it doesn't work when they input the host using the full URL (including the protocol). I think it should work both with and without the protocol. If the user specifies the protocol, it should use that protocol, otherwise it should do it based on if the CA certificate is provided or not.
2. In my case I don't need an CA certificate, but I can't use HTTPs without inputting a CA certificate

docs/documentation/platform/dynamic-secrets/clickhouse.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+The Infisical ClickHouse dynamic secret allows you to generate ClickHouse database credentials on demand based on a configured role.
+
+## Prerequisite
+
+Create a user with the required permissions in your ClickHouse instance. This user will be used to create new accounts on-demand.
+
+```sql
+CREATE USER 'infisical' IDENTIFIED WITH sha256_password BY '<strong-password>';
+GRANT CREATE USER, DROP USER, GRANT OPTION ON *.* TO 'infisical';
```
varonix0, line 16: This didn't work for me and it threw an error when I tried to grant, instead I had to do:

```sql
GRANT CREATE USER, DROP USER ON *.* TO 'infisical' WITH GRANT OPTION;
```

How come we documented that specific GRANT statement? Is it a configuration issue on my end?

varonix0, line 16: When I try to create a lease with the above statement I shared, I get this error:

```
infisical-dev-api              |       "message": "Failed to create lease from provider: infisical: Not enough privileges. To execute this query, it's necessary to have the grant SELECT ON [REDACTED].* WITH GRANT OPTION. ",
```

I was able to solve this using:
```
GRANT SELECT ON <database-name>.* TO 'infisical' WITH GRANT OPTION;
```

Did we miss this during documentation?

frontend/src/pages/secret-manager/SecretDashboardPage/components/DynamicSecretListView/EditDynamicSecretForm/EditDynamicSecretClickHouseForm.tsx
```diff
@@ -0,0 +1,649 @@
+import { Controller, useForm } from "react-hook-form";
```
varonix0, line 1: Ignore placement of comment: After provisioning a lease, I see an empty modal which doesn't contain any lease details. This tells me what we're missing a render statement for Clickhouse leases 

Image

backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
```diff
@@ -0,0 +1,399 @@
...
+import { ActorIdentityAttributes } from "../../dynamic-secret-lease/dynamic-secret-lease-types";
+import { TGatewayServiceFactory } from "../../gateway/gateway-service";
+import { TGatewayV2ServiceFactory } from "../../gateway-v2/gateway-v2-service";
+import { verifyHostInputValidity } from "../dynamic-secret-fns";
+import { DynamicSecretClickhouseSchema, PasswordRequirements, TDynamicProviderFns } from "./models";
+import { generateUsername } from "./templateUtils";
+
+const EXTERNAL_REQUEST_TIMEOUT = 10 * 1000;
+
+const DEFAULT_PASSWORD_REQUIREMENTS: PasswordRequirements = {
```
varonix0, line 22: After I tried creating a lease at first, I saw the error "password must contain at least 1 special char", and in my cloud instance of clickhouse that is a default password requirement. We should make the default password requirements include 1 symbol as well. This should also match what's displayed on the frontend for the default password requirements

backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
```diff
@@ -0,0 +1,399 @@
...
+      password: providerInputs.password,
+      database: providerInputs.database,
+      request_timeout: EXTERNAL_REQUEST_TIMEOUT,
+      ...(providerInputs.ca && {
+        tls: { ca_cert: Buffer.from(providerInputs.ca) }
+      })
+    });
+
+    return client;
+  };
```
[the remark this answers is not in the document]
varonix0: Greptile is right, the sanitized host IP is unused even though we parse it

backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
```diff
@@ -0,0 +1,399 @@
...
+
+  const executeStatements = async (statement: string, client: ReturnType<typeof createClient>) => {
+    const queries = statement
+      .split(";")
+      .map((query) => query.trim())
+      .filter(Boolean);
+
+    for (const query of queries) {
```
varonix0, line 164: can use `for await (const query of queries)` instead of `// eslint-disable-next-line no-await-in-loop`

PrestigePvP: I don't think we'd want to do that here? `queries` is a list of strings, not a readablestream or async iterable or anything like that. 

Also, it looks like every other provider uses this exact same pattern.

docs/documentation/platform/dynamic-secrets/clickhouse.mdx
PrestigePvP: No, these commands were just not tested, I assumed they were correct but that assumption was wrong :) 

I am fixing them, I did like the inclusion of them though

backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
PrestigePvP: Ah okay, good catch on this. I was just following the example from the SQL provider but it makes sense that this would need to be different for Clickhouse.

backend/src/ee/services/dynamic-secret/providers/clickhouse.ts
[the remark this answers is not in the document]
PrestigePvP: The logic I have is similar to all the other SQL clients, the only client that I can see that actually uses it is Postgres where there is no gatewayId. 

I'm honestly not sure which is the absolute correct approach here but I figured I would just match the established pattern

docs/documentation/platform/dynamic-secrets/clickhouse.mdx
```diff
@@ -11,15 +11,6 @@ The Infisical ClickHouse dynamic secret allows you to generate ClickHouse databa
 
 Create a user with the required permissions in your ClickHouse instance. This user will be used to create new accounts on-demand.
 
-```sql
-CREATE USER 'infisical' IDENTIFIED WITH sha256_password BY '<strong-password>';
```
varonix0: I actually liked this, I think we should add it back but with the correct statements

PR #5610, opened 2026-03-06
feat(pki-118): add spiffe machine auth

PrestigePvP:
## Context

- Adds SPIFFE machine auth mechanism 
 - Ability for remote bundle usage
 - Ability to use static JWKs for verifcation

## Screenshots

TODO

## Type

- [ ] Fix
- [X] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/ac19cc2d-52c4-42dd-a9cc-a435fe096842?refs=f2f10f91671fafce562356807f40122ca2ad4d20&source=prChecksComment) |

backend/src/lib/api-docs/constants.ts
```diff
@@ -745,6 +746,61 @@ export const JWT_AUTH = {
   }
 } as const;
 
+export const SPIFFE_AUTH = {
+  LOGIN: {
+    identityId: "The ID of the machine identity to login.",
+    jwt: "The JWT-SVID token to authenticate with.",
```
akhilmhdh: This is not always required for the login right? I think we need to mention that

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+  SpiffeConfigurationType
+} from "@app/services/identity-spiffe-auth/identity-spiffe-auth-types";
+import {
+  validateSpiffeAllowedAudiencesField,
+  validateSpiffeAllowedIdsField,
+  validateTrustDomain
+} from "@app/services/identity-spiffe-auth/identity-spiffe-auth-validators";
+import { isSuperAdmin } from "@app/services/super-admin/super-admin-fns";
+
+const IdentitySpiffeAuthResponseSchema = IdentitySpiffeAuthsSchema.omit({
```
akhilmhdh: It's recommended to use pick instead of omit because incase someone adds another sensitive field - that would end up in output

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+    .number()
+    .int()
+    .min(0)
+    .max(315360000)
+    .default(2592000)
+    .describe(SPIFFE_AUTH.ATTACH.accessTokenMaxTTL),
+  accessTokenNumUsesLimit: z.number().int().min(0).default(0).describe(SPIFFE_AUTH.ATTACH.accessTokenNumUsesLimit)
+});
+
+const CommonUpdateFields = z
```
akhilmhdh: Isn't this same as create with partial?

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+    accessTokenTTL: z.number().int().min(0).max(315360000).describe(SPIFFE_AUTH.UPDATE.accessTokenTTL),
+    accessTokenMaxTTL: z.number().int().min(0).max(315360000).describe(SPIFFE_AUTH.UPDATE.accessTokenMaxTTL),
+    accessTokenNumUsesLimit: z.number().int().min(0).describe(SPIFFE_AUTH.UPDATE.accessTokenNumUsesLimit)
+  })
+  .partial();
+
+const StaticConfigurationSchema = z.object({
+  configurationType: z.literal(SpiffeConfigurationType.STATIC).describe(SPIFFE_AUTH.ATTACH.configurationType),
+  caBundleJwks: z.string().min(1).describe(SPIFFE_AUTH.ATTACH.caBundleJwks),
+  bundleEndpointUrl: z.string().optional().default(""),
```
akhilmhdh: Can you share the reasoning behind static method having `bundleEndpointUrl`, I didn't understand it's purpose in static mode?

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+  bundleEndpointProfile: z.nativeEnum(SpiffeBundleEndpointProfile).optional(),
+  bundleEndpointCaCert: z.string().optional().default(""),
+  bundleRefreshHintSeconds: z.number().int().min(0).optional().default(3600)
+});
+
+const RemoteConfigurationSchema = z.object({
+  configurationType: z.literal(SpiffeConfigurationType.REMOTE).describe(SPIFFE_AUTH.ATTACH.configurationType),
+  caBundleJwks: z.string().optional().default(""),
```
akhilmhdh: Likewise remote having caBundle?

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+  bundleEndpointUrl: z
+    .string()
+    .trim()
+    .url()
+    .refine((url) => url.startsWith("https://"), "Bundle endpoint URL must use HTTPS")
+    .describe(SPIFFE_AUTH.ATTACH.bundleEndpointUrl),
+  bundleEndpointProfile: z
```
akhilmhdh: bundleEndpointProfile seems quite confusing? I think we should use a more sensible word that means the various remote accessing methods.
Should we just do a bigger discrimination instead of another one in remote - because here bundleEndpointCaCert doesn't make sense in one of them and in SPIFFE based it's more like a SPIFFE bundle rather than a ca cert.
The present API schema seems quite complex in merge and discriminated. I think this can be simplified like this how we do it for other things like app connection with one connection having multiple ways to do auth and has general fields:

```js
{
"<auth-info-key>": {
   // contains discriminated union of auth types
},
// ... all general fields
}
```

This way we can accommodate it more easier rather than fighting the ts or doing a merge with everything

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+    .nativeEnum(SpiffeBundleEndpointProfile)
+    .default(SpiffeBundleEndpointProfile.HTTPS_WEB)
+    .describe(SPIFFE_AUTH.ATTACH.bundleEndpointProfile),
+  bundleEndpointCaCert: z.string().optional().default("").describe(SPIFFE_AUTH.ATTACH.bundleEndpointCaCert),
```
akhilmhdh: If it's optional - no need for default empty string. This would cause more handling in tf side, as there will be drift.

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+        {
+          bearerAuth: []
+        }
+      ],
+      params: z.object({
+        identityId: z.string().describe(SPIFFE_AUTH.REVOKE.identityId)
+      }),
+      response: {
+        200: z.object({
+          identitySpiffeAuth: IdentitySpiffeAuthResponseSchema.omit({
```
akhilmhdh: It's recommended to use pick, i have stated the reason above

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-fns.ts
```diff
@@ -0,0 +1,96 @@
+import https from "https";
+import picomatch from "picomatch";
+import RE2 from "re2";
+
+const SPIFFE_ID_REGEX = new RE2("^spiffe:\\/\\/([^/]+)(\\/.*)?");
+
+export const isValidSpiffeId = (value: string): boolean => {
+  return SPIFFE_ID_REGEX.test(value);
```
akhilmhdh, line 10: It's better to use `URL` instead of regex.

You can do something like this
```
const a = new URL(spiffe-url)
a.protocol -> check
```

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-fns.ts
```diff
@@ -0,0 +1,96 @@
...
+};
+
+export const extractTrustDomainFromSpiffeId = (spiffeId: string): string => {
+  const match = SPIFFE_ID_REGEX.exec(spiffeId);
```
akhilmhdh, line 14: Same here

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-fns.ts
```diff
@@ -0,0 +1,96 @@
...
+const MAX_BUNDLE_SIZE_BYTES = 1_048_576; // 1 MB
+
+export const fetchRemoteBundleJwks = (url: string, caCert?: string): Promise<string> => {
+  const parsedUrl = new URL(url);
+
+  return new Promise<string>((resolve, reject) => {
+    let totalBytes = 0;
+    const chunks: Buffer[] = [];
+
+    const req = https.request(
```
akhilmhdh: Any reason for using raw http? There is a `request` module built on top of using axios and retry in place to do external request.

docs/documentation/platform/identities/spiffe-auth.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+    ![identities page remove default auth](/images/platform/identities/identities-page-remove-default-auth.png)
+
+    Now create a new SPIFFE Auth Method.
+
+    <Warning>Restrict access by properly configuring the trust domain, allowed SPIFFE IDs, and allowed audiences.</Warning>
+
+    Here's some information about each field:
+
+    **Configuration Type:**
```
akhilmhdh: It's better to move this to tabs

docs/documentation/platform/identities/spiffe-auth.mdx
```diff
@@ -0,0 +1,230 @@
...
+    - **Static:** You provide the SPIRE JWKS JSON directly. Best for environments where the trust bundle is managed externally or does not change frequently.
```
akhilmhdh: I think static is more for like gapped environment and that's what we should mention. If it's managed externally - they should use remote ones

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
```diff
@@ -0,0 +1,905 @@
...
+
+    // Fetch fresh bundle
+    if (!config.bundleEndpointUrl) {
+      throw new BadRequestError({ message: "Remote SPIFFE auth has no bundle endpoint URL configured" });
+    }
+
+    await blockLocalAndPrivateIpAddresses(config.bundleEndpointUrl);
+
+    let caCert: string | undefined;
+    if (config.bundleEndpointProfile === "https_spiffe" && config.encryptedBundleEndpointCaCert) {
```
akhilmhdh: Missing usage of enum?

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-types.ts
```diff
@@ -0,0 +1,60 @@
+import { TProjectPermission } from "@app/lib/types";
+
+export enum SpiffeConfigurationType {
+  STATIC = "static",
+  REMOTE = "remote"
+}
+
+export enum SpiffeBundleEndpointProfile {
+  HTTPS_WEB = "https_web",
```
akhilmhdh: When we do support pem mode as well - this would be more confusing? I think we need to more precise - like it's bundle or pem

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
```diff
@@ -0,0 +1,430 @@
...
+  bundleEndpointUrl: z.string().optional().default(""),
+  bundleEndpointProfile: z.nativeEnum(SpiffeBundleEndpointProfile).optional(),
+  bundleEndpointCaCert: z.string().optional().default(""),
+  bundleRefreshHintSeconds: z.number().int().min(0).optional().default(3600)
+});
+
+const RemoteConfigurationSchema = z.object({
+  configurationType: z.literal(SpiffeConfigurationType.REMOTE).describe(SPIFFE_AUTH.ATTACH.configurationType),
+  caBundleJwks: z.string().optional().default(""),
+  bundleEndpointUrl: z
```
akhilmhdh: I think calling the whole remote as bundle for a non spiffe remote access feels misleading as well. Because the endpoint is used to fetch the pem certificate as well. It would be better to call it as generic ones like `endpointUrl`

docs/documentation/platform/identities/spiffe-auth.mdx
```diff
@@ -0,0 +1,230 @@
...
+  Note over Infisical: Step 4: Token Generation
+  Infisical->>Client: Return short-lived access token
+
+  Note over Client,Infisical: Step 5: Access Infisical API with Token
+  Client->>Infisical: Make authenticated requests using the short-lived access token
+```
+
+## Concept
+
+At a high level, Infisical authenticates a workload by verifying its JWT-SVID and checking that it meets specific requirements (e.g. its SPIFFE ID matches an allowed pattern, it belongs to the expected trust domain) at the `/api/v1/auth/spiffe-auth/login` endpoint. If successful, Infisical returns a short-lived access token that can be used to make authenticated requests to the Infisical API.
```
akhilmhdh, line 37: We need to also mention that we don't support the cert based auth as of now both in api docs and here. This will help people raise it as issue when migrating from other services rather than searching for it in multiple places.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
```diff
@@ -0,0 +1,905 @@
...
+  identityDAL,
+  identitySpiffeAuthDAL,
+  membershipIdentityDAL,
+  permissionService,
+  licenseService,
+  identityAccessTokenDAL,
+  kmsService,
+  orgDAL
+}: TIdentitySpiffeAuthServiceFactoryDep) => {
+  const resolveJwks = async ({
```
akhilmhdh: Internal functions we just put a `$` at start

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
```diff
@@ -0,0 +1,889 @@
...
+  try {
+    const tokenData = crypto.jwt().verify(jwtValue, keyLike as jwt.Secret) as Record<string, unknown>;
+    return { tokenData, kid };
+  } catch (error) {
+    if (error instanceof jwt.JsonWebTokenError) {
+      throw new UnauthorizedError({ message: `Access denied: ${error.message}` });
+    }
+    throw error;
+  }
+};
```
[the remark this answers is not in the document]
akhilmhdh: Yes this is a valid concern especially since we have to maintain fips as well. I think it's better to validate algos with fips only

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
```diff
@@ -0,0 +1,905 @@
...
+  identitySpiffeAuthDAL,
+  membershipIdentityDAL,
+  permissionService,
+  licenseService,
+  identityAccessTokenDAL,
+  kmsService,
+  orgDAL
+}: TIdentitySpiffeAuthServiceFactoryDep) => {
+  const resolveJwks = async ({
+    config,
```
akhilmhdh, line 200: Additionally you don't need to do this by yourself. There is remote jwks parsing support for jose. Please refer [that](https://github.com/panva/jose/blob/main/docs/jwks/remote/functions/createRemoteJWKSet.md).

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
```diff
@@ -0,0 +1,905 @@
...
+            orgId: identity.orgId,
+            forceRefresh: true
+          }));
+          ({ tokenData } = await verifyJwtSvid(jwtValue, jwksJson));
+        } else {
+          throw verifyError;
+        }
+      }
+
+      if (!validateSpiffeClaims(tokenData, identitySpiffeAuth)) {
```
akhilmhdh, line 326: The verify already can do all the validation for the auth checks as well. We can use that instead of manually doing it. 

https://github.com/panva/jose/blob/main/docs/jwt/verify/interfaces/JWTVerifyOptions.md#audience

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
```diff
@@ -0,0 +1,905 @@
...
+        throw new BadRequestError({
+          message: "The IP is not a valid IPv4, IPv6, or CIDR block"
+        });
+      return extractIPDetails(accessTokenTrustedIp.ipAddress);
+    });
+
+    if (bundleEndpointUrl) {
+      await blockLocalAndPrivateIpAddresses(bundleEndpointUrl);
+    }
+
```
akhilmhdh, line 530: Let's do an additional check that the provided setup is working or not

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
```diff
@@ -0,0 +1,905 @@
...
+    const identitySpiffeAuth = await identitySpiffeAuthDAL.findOne({ identityId });
+
+    if (identitySpiffeAuth.configurationType !== SpiffeConfigurationType.REMOTE) {
+      throw new BadRequestError({
+        message: "Bundle refresh is only applicable to identities with remote SPIFFE Auth configuration"
+      });
+    }
+
+    await resolveJwks({ config: identitySpiffeAuth, orgId: identityMembershipOrg.scopeOrgId, forceRefresh: true });
```
akhilmhdh: Why not just return the new one instead of again fetching and decrypting it?

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
PrestigePvP: Yeah it looks like it - happy to just reuse it.

backend/src/server/routes/v1/identity-spiffe-auth-router.ts
PrestigePvP: Yeah, totally agree on this, this was sloppy.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-fns.ts
PrestigePvP: I don't know that I agrees here. The Regex is easily readable and uses RE2, which should account for the attack vectors in place. 

Also, generally doing a URL like that is going to be slower than a simple regex check. Unless you're strongly convinced, I think we should leave it.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-fns.ts
PrestigePvP: A couple reasons:
- The biggest being that I didn't know that we had Axios in place to do this. When I was looking into how to do this request this pattern came up first. 
- I also think that it might be better if we do it this way, so that we can have a max bundle size in place and end early if we're past that bundle size. 

I think I'm going to stay with this approach because the Axios request uses the content length header, which the server can lie about and force our server to download a huge bundle. Then we check the type and determine that it's way too big of a bundle.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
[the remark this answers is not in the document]
PrestigePvP: I don't think this is a real concern here because I'm pretty sure that import JWK already solves for this? When looking into this a bit further, the only thing that I could see is that they could downgrade the algorithm in the same family, but even then this protection here would not prevent that. 

I also took inspiration from the pki-acme-service.ts file and it doesn't check against it either.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
PrestigePvP: Thats kinda strange, alrighty.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
PrestigePvP: Yeah, I agree. We can let Jose take care of a lot of the audience validation and stuff, and we will handle the SPIFFE validation itself.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-types.ts
PrestigePvP: The name comes from the SPIFFE spec. A "bundle endpoint" serves trust bundles, and each      
  endpoint has a "profile" that describes how the client authenticates the endpoint.

* https_web -> uses standard web PKI (public CA)
* https_spiffe -> endpoint uses SPIFFE-issued cert (custom CA bundle)

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
PrestigePvP: Yeah, I think validating it during creation is a good idea. I'm adding that now.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
PrestigePvP: Yeah, you're totally right.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-service.ts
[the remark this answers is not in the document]
akhilmhdh: I am not exactly talking about that. We want to ensure it's only fips validated algorithms and not the non fips. This is because if they are using fips image and not the fips algo - it would throw an internal error rather than showing to ui. Like if they are using `EdDSA` it would fail

Oh! Then that's a bug. You can create a ticket there to address it.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-fns.ts
akhilmhdh: I have checked the axios module and it seems to doing what you have just did. Here is the code [snippet](https://github.com/axios/axios/blob/f7a4ee21d5d62e472b33cbadf351467b19776a3c/lib/adapters/http.js#L798)

I would recommend to switching to the battle tested module like `axios` instead of doing this by ourself. Additionally it's much more easier to read as well.

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-fns.ts
akhilmhdh: The main thing is - users will not get it first read. 

Like right now, but later everyone would need to either LLM it or some readable platform. We generally not recommend using regex only if it's a complex pattern. Otherwise - use functions. 

Not super pressing though

backend/src/services/identity-spiffe-auth/identity-spiffe-auth-types.ts
akhilmhdh: As shared in our discussions
```
[
  {
    "trustBundleDistribution": {
      "profile": "static",
      "bundle": "<--cert-->"
    },
    "trustDomain": "example.org",
    "allowedAudiences": [
      "example-audience"
    ],
    "accessTokenTTL": "3600s",
    "accessTokenMaxTTL": "7200s"
  },
  
  
  {
    "trustBundleDistribution": {
      "profile": "remote_http | remote_spiffe | remote_pem",
      "bundleEndpointCaCert": "<--cert-->",
      "bundleRefreshHintSeconds": 30
    },
    "trustDomain": "example.org",
    "allowedAudiences": [
      "example-audience"
    ],
    "accessTokenTTL": "3600s",
    "accessTokenMaxTTL": "7200s"
  }
]
```

akhilmhdh (approved):

PR #5638, opened 2026-03-09
fix(platform-230): use SSO platform identifiers

PrestigePvP:
## Context

- Fix is a bug where users who use social platforms for SSO login and then change their email will have a new user created.
- Use the social platform identifier provided as a user alias. 

## Steps to verify the change

- Log in using a social provider and then change your email. 
- Then re-login using that same social provider. If you have a new account, then this change did not work. If you were using the same existing account, this change worked. 

## Type

- [X] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [X] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: `type(scope): short description` (scope is optional, e.g., `fix: prevent crash on sync` or `fix(api): handle null response`). 
- [X] Tested locally

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -2198,7 +2198,6 @@
       "integrity": "sha512-vMqyb7XCDMPvJFFOaT9kxtiRh42GwlZEg1/uIgtZshS5a/8OaduUfCi7kynKgc3Tw/6Uo2D+db9qBttghhmxwQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.2",
@@ -4246,7 +4245,6 @@
           "url": "https://opencollective.com/csstools"
         }
       ],
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -4268,7 +4266,6 @@
           "url": "https://opencollective.com/csstools"
         }
       ],
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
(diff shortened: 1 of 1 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/576bb557-454f-49de-aed6-73f164764b6e?refs=0a7c4080b027b6ff649a6da7d95c0b28624144dc&source=prChecksComment) |

backend/src/services/auth/auth-login-service.ts
```diff
@@ -967,10 +971,47 @@ export const authLoginServiceFactory = ({
...
+
+    // Step 1: Look up user by provider alias (stable, immutable ID)
+    let user: TUsers | undefined;
+    let isNewAlias = false;
+    const existingAlias = await userAliasDAL.findOne({ externalId: providerUserId, aliasType });
+    if (existingAlias) {
+      user = await userDAL.findById(existingAlias.userId);
+    }
+
+    // Step 2: Fall back to email lookup for existing users without an alias yet
```
akhilmhdh, line 1030: One case missing here is what happens when we do the provider, but there email got changed. 

Example flow:
```
1. User google email got changed
2. Successful login with oauth into there old account.
3. There was already another user with the new email
```

How can we address this?

PrestigePvP: I think this would only be something that we would need to worry about for this one-time migration, right? Is there any way that we can check the database to ensure this case doesn't exist? 

The reason I'm saying that is because I think the actual fix for this edge case would be potentially a lot more complex for a case that can't be achieved after this change is live.

PR #5685, opened 2026-03-13
fix(secret-sync): replace non-atomic concurrency counter with Redis INCRBY to prevent deadlock

### LINKED ISSUE #5563: Critical race condition in AppConnectionConcurrentJobs causing sync deadlock
### Describe the bug

There is a race condition in the connection concurrency control used by Secret Sync and PKI Sync.

The counter (`AppConnectionConcurrentJobs`) is managed using separate read and write operations in Redis/keyStore, which are not atomic. Under concurrent load, this can cause the counter to become incorrect and remain permanently elevated.

As a result:
- Sync jobs fail with "concurrency limit reached" even when no jobs are running
- The connection appears permanently busy
- Manual counter reset is required to recover

### To Reproduce

1. Trigger multiple Secret Sync or PKI Sync jobs for the same connection at the same time
2. Observe some jobs failing with "concurrency limit reached"
3. Check Redis – the counter remains > 0 even though no jobs are active
4. Further sync attempts continue to fail

### Expected behavior

The concurrency counter should always reflect the actual number of running jobs. Jobs should only fail when the real limit is reached.

### Root Cause

Current implementation uses a non-atomic pattern:

```ts
const concurrencyCount = await keyStore.getItem(...);
await keyStore.setItemWithExpiry(...);
````

This creates a TOCTOU race condition.

### Suggested Fix

Use the atomic `incrementBy()` method instead:

```ts
await keyStore.incrementBy(key, 1);
await keyStore.incrementBy(key, -1);
```

This should prevent lost updates and stuck counters.

### Impact

* Sync operations can get permanently blocked
* More likely under high load
* Requires manual intervention to fix

Fizza-Mukhtar:
## Context

Fixes #5563

The concurrency counter used by Secret Sync and PKI Sync (`AppConnectionConcurrentJobs`) is implemented as a non-atomic read-modify-write pattern:

```ts
// BEFORE — race condition
const count = await keyStore.getItem(key);       // READ
await keyStore.setItemWithExpiry(key, count + 1); // WRITE
```

Under concurrent load, two jobs can read the same stale value before either writes back, causing lost updates. On decrement, the same race causes the counter to remain permanently elevated above zero even when no jobs are active. All subsequent sync attempts then fail with a concurrency limit error and require manual Redis intervention to recover.

**Fix:** Replace the read-modify-write pattern with `keyStore.incrementBy()`, which maps directly to Redis `INCRBY` — a single atomic command with no read-write gap:

```ts
// AFTER — atomic
await keyStore.incrementBy(key, 1);
await keyStore.setExpiry(key, TTL_SECONDS);
```

For decrement, the new value returned by `incrementBy` is checked: if still positive, the TTL is refreshed; if zero or negative, the key is left to expire naturally via the existing 15-minute safety TTL. This preserves the crash-safety behaviour of the original `setItemWithExpiry` while eliminating the race

The same buggy pattern exists identically in both `secret-sync-queue.ts` and `pki-sync-queue.ts` — both are fixed. The `keyStore` Pick types in both factory deps are updated to include `"incrementBy"` and `"setExpiry"`.

## Screenshots

N/A

## Steps to verify the change

1. Trigger 3+ concurrent Secret Sync jobs for the same app connection
2. Wait for all jobs to complete
3. Check Redis: `GET app-connection-concurrency-<connectionId>` should be `0` or the key should be absent
4. Confirm subsequent sync jobs are not incorrectly rejected with a concurrency limit error
5. `npx tsc --noEmit` in `backend/` should pass with no type errors

## Type
- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist
- [x] Tested locally

### DIFF
--- a/backend/src/services/secret-sync/secret-sync-queue.ts
+++ b/backend/src/services/secret-sync/secret-sync-queue.ts
@@ -230,17 +230,11 @@ export const secretSyncQueueFactory = ({
   };
 
   const $decrementConnectionConcurrencyCount = async (connectionId: string) => {
-    const concurrencyCount = await keyStore.getItem(KeyStorePrefixes.AppConnectionConcurrentJobs(connectionId));
-
-    const currentCount = Number.parseInt(concurrencyCount || "0", 10);
-
-    const decrementedCount = Math.max(0, Number.isNaN(currentCount) ? 0 : currentCount - 1);
-
-    await keyStore.setItemWithExpiry(
-      KeyStorePrefixes.AppConnectionConcurrentJobs(connectionId),
-      (REQUEUE_MS * REQUEUE_LIMIT) / 1000, // in seconds
-      decrementedCount
-    );
+    const key = KeyStorePrefixes.AppConnectionConcurrentJobs(connectionId);
+    const newCount = await keyStore.incrementBy(key, -1);
+    if (newCount > 0) {
+      await keyStore.setExpiry(key, (REQUEUE_MS * REQUEUE_LIMIT) / 1000);
+    }
   };
 
   const $getInfisicalSecrets = async (


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/66ee5d02-6e54-4b2b-9935-560a448f7fa7?refs=a6220052b0d272378f062ad684d5807d16c6178a&source=prChecksComment) |

Fizza-Mukhtar:
> ### Greptile Summary
> This PR replaces the non-atomic read-modify-write concurrency counter pattern with the atomic Redis `INCRBY` command (`keyStore.incrementBy`) in both `secret-sync-queue.ts` and `pki-sync-queue.ts`. The fix is correct and complete in `pki-sync-queue.ts`, but **incomplete in `secret-sync-queue.ts`**: only `$incrementConnectionConcurrencyCount` was updated; `$decrementConnectionConcurrencyCount` still uses the old `getItem` + `setItemWithExpiry` race condition. This means under concurrent load, two jobs finishing at the same time can both read the same stale counter and write back an incorrectly elevated value — keeping the counter permanently above zero and blocking future sync jobs, which is exactly the failure mode this PR intends to resolve.
> 
> Key observations:
> 
> * `pki-sync-queue.ts`: both increment and decrement correctly migrated to `incrementBy` — no issues
> * `secret-sync-queue.ts`: increment fixed, but **decrement (lines 232–244) still uses the non-atomic read-modify-write pattern**
> * The `keyStore` Pick type in `secret-sync-queue.ts` was updated to include `"incrementBy"` and `"setExpiry"`, confirming the intent was to fix both functions, but the decrement body was not changed
> 
> ### Confidence Score: 2/5
> * Not safe to merge — the race condition being fixed is only half-resolved in secret-sync-queue.ts, leaving the decrement path still vulnerable to the same deadlock bug.
> * The fix is correct and complete in pki-sync-queue.ts, but secret-sync-queue.ts's $decrementConnectionConcurrencyCount was not updated. The increment is now atomic while the decrement is still non-atomic, meaning the counter can still get stuck permanently elevated under concurrent load, reintroducing the exact failure mode the PR describes fixing.
> * backend/src/services/secret-sync/secret-sync-queue.ts — the $decrementConnectionConcurrencyCount function (lines 232–244) must be updated to match the atomic pattern already applied in pki-sync-queue.ts
> 
> ### Important Files Changed
> Filename	Overview
> backend/src/services/secret-sync/secret-sync-queue.ts	Only $incrementConnectionConcurrencyCount was updated to use the atomic incrementBy — $decrementConnectionConcurrencyCount was NOT changed and still contains the non-atomic read-modify-write race that this PR aims to fix.
> backend/src/services/pki-sync/pki-sync-queue.ts	Both $incrementConnectionConcurrencyCount and $decrementConnectionConcurrencyCount were correctly migrated to atomic incrementBy, matching the described fix. No issues found.
> Last reviewed commit: [2085abd](https://github.com/Infisical/infisical/commit/2085abd44e7864cb3741ab757c79a522a8ae8dc1)

Updated the fix — $decrementConnectionConcurrencyCount in secret-sync-queue.ts is now also migrated to atomic incrementBy. Both files now match.

PR #5697, opened 2026-03-14
improvement(pki): add RBAC policy conditions for certificate metadata

saifsmailbox98:
## Context

Adds metadata as an RBAC condition for certificates

## Screenshots

CleanShot 2026-03-15 at 00 56 55@2x

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/permission/project-permission.ts
+++ b/backend/src/ee/services/permission/project-permission.ts
@@ -764,12 +764,14 @@ const DynamicSecretConditionV2Schema = z
           key: z
             .object({
               [PermissionConditionOperators.$EQ]: PermissionConditionSchema[PermissionConditionOperators.$EQ],
+              [PermissionConditionOperators.$NEQ]: PermissionConditionSchema[PermissionConditionOperators.$NEQ],
               [PermissionConditionOperators.$IN]: PermissionConditionSchema[PermissionConditionOperators.$IN]
             })
             .partial(),
           value: z
             .object({
               [PermissionConditionOperators.$EQ]: PermissionConditionSchema[PermissionConditionOperators.$EQ],
+              [PermissionConditionOperators.$NEQ]: PermissionConditionSchema[PermissionConditionOperators.$NEQ],
               [PermissionConditionOperators.$IN]: PermissionConditionSchema[PermissionConditionOperators.$IN]
             })
             .partial()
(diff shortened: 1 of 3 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/52a19fdb-34d3-44df-9ee1-85f3a08613d4?refs=8e1972f4438e530b525dde41387381cc2d41d21a&source=prChecksComment) |

saifsmailbox98:
@greptile re-review this PR

saifsmailbox98:
@greptile
1. We had to flip the order because to check the permissions, we need the metadata from the certificate object. It is not easy to enumerate over UUIDs. 
2. I think the solution we have handles it pretty nicely and does not need any change
3. This is how condition fields work. 
4. We do similar things for other entities as well.

backend/src/ee/services/permission/project-permission.ts
```diff
@@ -470,10 +470,11 @@ export type CertificateAuthoritySubjectFields = {
```
carlosmonastyrski, line 479: Ignore comment placement
I added a not equal condition, but it's failing when it's trying to read a certificate with no metadata set, which should be fine as the metadata value is not equal 🤔 
Image

carlosmonastyrski: Also if I add two metadata values, one valid the other not valid, it's allowing me to see the value when it should block it if any of the metadata values matches that.

Image

saifsmailbox98: > I added a not equal condition, but it's failing when it's trying to read a certificate with no metadata set, which should be fine as the metadata value is not equal 🤔

elemMatch tries to find atleast one object in the array that matches the conditions

So, if there are no objects in the array, it will automatically result in false (cert won't be accessible)

> Also if I add two metadata values, one valid the other not valid, it's allowing me to see the value when it should block it if any of the metadata values matches that.

NOT EQUALS ($neq) also works the same way, if any object in the array matches the condition it will result in true -> cert will be accessible


Here,
object = one metadata key + pair
array = array of metadata on the cert

So, if one metadata says not equals, and a different object satistifies it it will allow access to that cert, and the object that should deny access will be ignored.

Removed NEQ completely from metadata conditions. Users can rely on Forbid for denying access.
Metadata key and value should always be set together in conditions, but we can probably do that later. Users can be blamed for any misconfiguration if they do not use both together.

backend/src/ee/services/permission/project-permission.ts
carlosmonastyrski, line 747: Why are we also removing the `$NEQ` condition from Dynamic Secrets? wouldn't this break any existing permissions any user may have with this? 🤔

backend/src/ee/services/permission/project-permission.ts
```diff
@@ -1227,14 +1226,12 @@ const PamAccountConditionSchema = z
           key: z
             .object({
               [PermissionConditionOperators.$EQ]: PermissionConditionSchema[PermissionConditionOperators.$EQ],
-              [PermissionConditionOperators.$NEQ]: PermissionConditionSchema[PermissionConditionOperators.$NEQ],
```
carlosmonastyrski, line 1230: same for App Connections

carlosmonastyrski: and the others as well

backend/src/ee/services/permission/project-permission.ts
saifsmailbox98: restored it for others

backend/src/ee/services/permission/project-permission.ts
saifsmailbox98: restored it for others

PR #5732, opened 2026-03-17
feat: add auto tidy feature for expired certificates

carlosmonastyrski:
## Context

Adds a project-level Certificate Cleanup setting under Certificate Manager > Settings that automatically removes expired certificates on a daily schedule. Admins can configure how many days past expiration before deletion (1–30), whether to include revoked certificates, and whether to skip certificates with active syncs. A BullMQ cron job processes eligible certificates in batches with offset-based pagination, records run status and deleted count, and emits an audit log with the serial number and CN/SAN of every removed certificate.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/2b7645a3-358e-4e61-ad26-aef8a3c855cc?refs=3356c4f017d1ff0fd335c121f39ae53a6ae94ebc&source=prChecksComment) |

backend/src/db/migrations/20260316124623_certificate-cleanup.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasTable(TableName.CertificateCleanupConfig))) {
+    await knex.schema.createTable(TableName.CertificateCleanupConfig, (tb) => {
+      tb.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      tb.string("projectId").notNullable().unique();
+      tb.foreign("projectId").references("id").inTable(TableName.Project).onDelete("CASCADE");
+      tb.boolean("isEnabled").notNullable().defaultTo(false);
+      tb.integer("daysBeforeDeletion").notNullable().defaultTo(3);
+      tb.boolean("includeRevokedCertificates").notNullable().defaultTo(false);
```
sheensantoscapadngan: why do we treat revoked certificates specially? curious why we need this flag in the first place given that we're cleaning up based only on whether they expired or not.. isn't that how other platforms do it? 

Image

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4763,6 +4756,29 @@ interface UpdateCertificateMetadataEvent {
   };
 }
 
+interface UpdateCertificateCleanupConfigEvent {
+  type: EventType.UPDATE_CERTIFICATE_CLEANUP_CONFIG;
+  metadata: {
+    projectId: string;
+    isEnabled: boolean;
+    daysBeforeDeletion: number;
```
sheensantoscapadngan: this doesn't have any indication that it's referring to days AFTER expiry. maybe this could be `postExpiryRetentionDays` instead?

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4763,6 +4756,29 @@ interface UpdateCertificateMetadataEvent {
...
+
+interface CertificateCleanupCompletedEvent {
+  type: EventType.CERTIFICATE_CLEANUP_COMPLETED;
+  metadata: {
+    deletedCount: number;
+    certificates: {
+      serialNumber: string;
+      commonName: string;
+      altNames: string | null;
+    }[];
```
sheensantoscapadngan: storing just the serial number should be enough here.. let's try to keep this as small as possible

backend/src/server/routes/v1/certificate-cleanup-router.ts
```diff
@@ -0,0 +1,103 @@
...
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    schema: {
+      hide: false,
+      operationId: "updateCertificateCleanupConfig",
+      description: "Create or update certificate cleanup configuration for a project",
+      tags: [ApiDocsTags.PkiCertificates],
+      body: z.object({
+        projectId: z.string().trim().describe("Project ID"),
+        isEnabled: z.boolean().optional().describe("Enable auto-cleanup"),
```
sheensantoscapadngan: nit - let's just refer to it as "Enable cleanup"

backend/src/server/routes/v1/certificate-cleanup-router.ts
```diff
@@ -0,0 +1,103 @@
...
+        daysBeforeDeletion: z.number().int().min(1).max(30).optional().describe("Days past expiration before deletion"),
```
sheensantoscapadngan: same remark here about property name. regardless of what name we choose, let's be consistent in the UI here 

Image

backend/src/services/certificate-cleanup/certificate-cleanup-queue.ts
```diff
@@ -0,0 +1,226 @@
...
+        .where("projectId", config.projectId)
+        .orderBy("id", "asc")
+        .select("id");
+
+      if (config.includeRevokedCertificates) {
+        query = query.where(function expiryOrRevoked() {
+          void this.where("notAfter", "<", cutoffDate).orWhere("status", CertStatus.REVOKED);
+        });
+      } else {
+        query = query.where("notAfter", "<", cutoffDate);
```
sheensantoscapadngan: each batch should be processed in one transaction for consistency right?

1st batch is tx 1, 2nd batch is tx 2, etc. 

shouldn't be too problematic in terms of lock contention

sheensantoscapadngan: when `includeRevokedCertificates` is not selected, then it seems like we're not using the status field in the query... will the index that you added still work? looks like that was a composite index so you'd need the status field as well otherwise it likely would not be used

backend/src/services/certificate-cleanup/certificate-cleanup-queue.ts
```diff
@@ -0,0 +1,226 @@
...
+          deletedCount += deleted.length;
+
+          for (const cert of deleted) {
+            pendingAuditCerts.push({
+              serialNumber: cert.serialNumber,
+              commonName: cert.commonName,
+              altNames: cert.altNames ?? null
+            });
+
+            if (pendingAuditCerts.length >= CLEANUP_AUDIT_LOG_CAP) {
```
sheensantoscapadngan: perhaps we can just keep audit_log_cap to be the same as the batch size for simplicity... unsure if 1000 entries would be problematic

backend/src/services/certificate-cleanup/certificate-cleanup-queue.ts
```diff
@@ -0,0 +1,226 @@
...
+            );
+          }
+        } catch (err) {
+          totalErrors += 1;
+          logger.error(err, `${QueueJobs.CertificateCleanup}: failed for project ${config.projectId}`);
+
+          await certificateCleanupConfigDAL.updateById(config.id, {
+            lastRunStatus: CleanupRunStatus.Error,
+            lastRunAt: new Date(),
+            lastRunCertsDeleted: 0,
```
sheensantoscapadngan: if it fails mid-way through a batch, this would not be accurate anymore right?

backend/src/services/certificate-cleanup/certificate-cleanup-service.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    const created = await certificateCleanupConfigDAL.create({
+      projectId,
+      isEnabled: isEnabled ?? false,
+      daysBeforeDeletion: daysBeforeDeletion ?? 3,
+      includeRevokedCertificates: includeRevokedCertificates ?? false,
+      skipCertsWithActiveSyncs: skipCertsWithActiveSyncs ?? true
+    });
+    return created;
```
sheensantoscapadngan: put in one txn

docs/documentation/platform/pki/settings/certificate-cleanup.mdx
```diff
@@ -0,0 +1,40 @@
+---
+title: "Certificate Cleanup"
+sidebarTitle: "Certificate Cleanup"
+description: "Learn how to configure automatic cleanup of expired certificates in Infisical"
+---
+
+## Concept
+
+Certificate cleanup is a project-level setting that automatically removes expired certificates on a daily schedule.
+Over time, expired certificates accumulate and add clutter to your certificate inventory. Enabling cleanup keeps your project organized by periodically deleting certificates that have been expired for a configurable number of days.
```
sheensantoscapadngan, line 10: how about deletion of the certificate request rows? we'd have to handle that as well right because that grows linearly with the number of certificates

docs/documentation/platform/pki/settings/certificate-cleanup.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+The cleanup job runs once daily. Each run processes eligible certificates in batches, rebuilds CRLs for any affected CAs, and records the result so you can verify it from the Settings page.
```
sheensantoscapadngan: why would it rebuild the CRL? if we're only deleting expired certificates, we don't care much about adding them to the CRL right

PR #5764, opened 2026-03-19
feat: add actor_id filter to audit logs

PrestigePvP:
## Context

- Add actor_id as a filterable item in the Audit log front end
- `Actor` was already supported on the backend, just needed to wire up the front end and add suggestions
- Also includes ID / type mismatch detection (see screenshot 2) 

## Screenshots

image

image

## Steps to verify the change

Go into audit log locally

## Type

- [ ] Fix
- [X] Feature
- [X] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/pages/organization/AuditLogsPage/components/AuditSearchFilter.tsx
+++ b/frontend/src/pages/organization/AuditLogsPage/components/AuditSearchFilter.tsx
@@ -491,7 +491,7 @@ export const AuditSearchFilter = ({
                           className={dropdownRowClass(highlightedIndex === index)}
                           onClick={() => selectSuggestion(suggestion)}
                           onMouseMove={handleMouseMove}
-                onMouseEnter={() => handleMouseEnter(index)}
+                          onMouseEnter={() => handleMouseEnter(index)}
                         >
                           <FontAwesomeIcon
                             icon={faPlus}
@@ -524,7 +524,7 @@ export const AuditSearchFilter = ({
                           className={dropdownRowClass(highlightedIndex === index)}
                           onClick={() => selectSuggestion(suggestion)}
                           onMouseMove={handleMouseMove}
-                onMouseEnter={() => handleMouseEnter(index)}
+                          onMouseEnter={() => handleMouseEnter(index)}
                         >
                           <FontAwesomeIcon
                             icon={faPlus}
@@ -555,7 +555,7 @@ export const AuditSearchFilter = ({
                     className={dropdownRowClass(highlightedIndex === index)}
                     onClick={() => selectSuggestion(suggestion)}
                     onMouseMove={handleMouseMove}
-                onMouseEnter={() => handleMouseEnter(index)}
+                    onMouseEnter={() => handleMouseEnter(index)}
                   >
                     <FontAwesomeIcon
                       icon={faPlus}


frontend/src/hooks/api/auditLogs/useAuditLogActorSuggestions.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { data: orgUsers = [], isLoading: isLoadingUsers } = useQuery({
+    queryKey: userKeys.getOrgUsers(orgId),
+    queryFn: () => fetchOrgUsers(orgId),
+    enabled: Boolean(orgId) && (!actorType || actorType === ActorType.USER)
+  });
+
+  const { data: orgIdentities, isLoading: isLoadingIdentities } = useQuery({
+    ...orgIdentityQuery.list({ limit: 100 }),
+    enabled: actorType === ActorType.IDENTITY
+  });
```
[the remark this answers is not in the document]
PrestigePvP: @greptileai 100 identities is more than enough to populate a list with. Ids that are manually pasted are not included in the suggestion limit so the 101'st ID that was manully typed would not get the warning. Whether we do 100 or 500 the problem of "what if they had more" stands. I think 100 is a valid number for suggestions in a filter list.

backend/src/ee/services/audit-log/audit-log-dal.ts
```diff
@@ -93,7 +93,12 @@ export const auditLogDALFactory = (db: TDbClient) => {
 
       // Special case: Filter by actor ID
       if (actorId) {
-        void sqlQuery.whereRaw(`"actorMetadata" @> jsonb_build_object('userId', ?::text)`, [actorId]);
+        const metadataKey = actorType
+          ? ACTOR_TYPE_TO_METADATA_ID_KEY[actorType]
+          : ACTOR_TYPE_TO_METADATA_ID_KEY[ActorType.USER];
+        if (metadataKey) {
+          void sqlQuery.whereRaw(`"actorMetadata" @> jsonb_build_object(?::text, ?::text)`, [metadataKey, actorId]);
+        }
```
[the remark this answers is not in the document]
PrestigePvP: @greptileai Thats correct behavior and intentional. Tell me how a filter for an id of an unknown user is supposed to work? 

Would an error be preferable to ignoring it?

frontend/src/pages/organization/AuditLogsPage/components/AuditSearchFilter.tsx
scott-ray-wilson, line 1: not related to your PR but something I think we should address - the display of filters doesn't show the underscore so if I try to match the filter with a space I don't see it: we should just include the underscore or handle this

Image

Image

maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/5be2ab1d-ed57-4f94-96b2-54f886e91af5?refs=97a02a9d541b205c0a9aa58b4a973fe2c9ad41bb&source=prChecksComment) |

frontend/src/pages/organization/AuditLogsPage/components/AuditSearchFilter.tsx
scott-ray-wilson, line 1: it doesn't feel intuitive I have to add the actor:identity filter to search - nothing in the UI indicates I need to do this; wonder if we should always show both or find another way to indicate this (Comment below is related)

Image

scott-ray-wilson, line 1: I don't think we should do the implicit actor type, it feels like it should search all if I don't specify it? nothing tells me I need to add the actor type to see my result

Image

Image

frontend/src/hooks/api/auditLogs/useAuditLogActorSuggestions.ts
```diff
@@ -0,0 +1,55 @@
...
+  const { data: orgUsers = [], isLoading: isLoadingUsers } = useQuery({
+    queryKey: userKeys.getOrgUsers(orgId),
+    queryFn: () => fetchOrgUsers(orgId),
+    enabled: enabled && Boolean(orgId) && (!actorType || actorType === ActorType.USER)
+  });
+
+  const { data: orgIdentities, isLoading: isLoadingIdentities } = useQuery({
+    ...orgIdentityQuery.list({ limit: 100 }),
+    enabled: enabled && actorType === ActorType.IDENTITY
+  });
```
scott-ray-wilson: for both these I think we need to do the filtering on the server side, some of are orgs have much greater numbers for both users and identities so they'll be confused why they don't see them listed

scott-ray-wilson (changes requested):
sweet feature, love the auto suggest - couple of comments

frontend/src/hooks/api/auditLogs/useAuditLogActorSuggestions.ts
scott-ray-wilson, line 1: this lists users/identities that are not part of the project when viewing project audit logs

Image

scott-ray-wilson (changes requested):
one more comment

frontend/src/pages/organization/AuditLogsPage/components/AuditSearchFilter.tsx
PrestigePvP: You only need to do identity first if you're looking for an identity. Ordinarily it will default to user. 

I agree the UI doesn't make this super clear but the alternative is mixing the two in the suggestions. If we do that tho we now have a state where if the user wants to filter by an ID that isn't found in the suggestions they would need to include the `actor` filter because there is no way we know what they're actually trying to filter by. 

I don't know a good middle ground here honestly, open to suggestions.

PrestigePvP: I don't know how else we would know what to search by?

frontend/src/hooks/api/auditLogs/useAuditLogActorSuggestions.ts
PrestigePvP: The filtering does occur on the server side, these are just to populate suggestions. The user is welcome to plop in any ID for a user/identity they have.

PrestigePvP: Oh, unless you're saying we should do server side filtering and debounce it as the user types?

frontend/src/pages/organization/AuditLogsPage/components/AuditSearchFilter.tsx
scott-ray-wilson: if there's no actor typed specified we could search on all the actor field options

If we don't do this I think if you include actor Id we should probably require actor type and not do implicit

scott-ray-wilson: we could either show both (selecting one adds the respective actor type) or require actor type for both if you only add actor id we display a warning like you did for mismatch stating that they need to specify the actor type

frontend/src/hooks/api/auditLogs/useAuditLogActorSuggestions.ts
scott-ray-wilson: yep! otherwise we're only searching a subset of the identities/users (limit: 100)

frontend/src/pages/organization/AuditLogsPage/components/AuditSearchFilter.tsx
scott-ray-wilson, line 1: When the list can scroll, arrow key navigation doesn't adjust the scroll position

https://github.com/user-attachments/assets/75b8d942-e308-4846-91bd-27a6fba7969d

scott-ray-wilson (approved):
LTGM!

PR #5987, opened 2026-04-09
feature(cache): add withCache helper and refactor Redis cache-aside usage

victorvhs017:
## Context

Introduces a small **`withCache`** cache-aside helper (`backend/src/lib/cache/with-cache.ts`) that reads from Redis, falls back to a `fetcher` on miss or Redis errors, and writes back with TTL. Centralizes the repeated get → parse → fetch → `JSON.stringify` → `setItemWithExpiry` pattern and avoids failing requests when Redis read/write fails.

Refactors existing call sites to use it (e.g. **upgrade path** config / calculate flows in `upgrade-path-service.ts`, and **super admin** server config loading where applicable) so behavior and cache keys stay the same while error handling and logging are consistent.

Adds unit tests in `with-cache.test.ts`.

## Steps to verify the change

1. Run backend unit tests: `cd backend && npm run test:unit` (or the subset for `with-cache.test.ts`).
2. **Upgrade path:** `POST /api/v1/upgrade-path/calculate` with valid `fromVersion` / `toVersion`; repeat and confirm responses match prior behavior (Redis hit on second call for cached keys).
3. **Super admin:** smoke any route that reads server config after cold start (if `getServerCfg` was refactored).

Check if the keys are being created in redis correctly

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/lib/cache/with-cache.ts
+++ b/backend/src/lib/cache/with-cache.ts
@@ -1,0 +1,35 @@
+import { TKeyStoreFactory } from "@app/keystore/keystore";
+import { logger } from "@app/lib/logger";
+
+type TWithCacheOpts<T> = {
+  keyStore: Pick<TKeyStoreFactory, "getItem" | "setItemWithExpiry">;
+  key: string;
+  ttlSeconds: number;
+  fetcher: () => Promise<T>;
+};
+
+/**
+ * Cache-aside helper: attempts to read from Redis, falls back to the fetcher on miss or Redis failure,
+ * and writes the result back to Redis. Redis errors are caught and logged — the fetcher is always the
+ * source of truth.
+ */
+export const withCache = async <T>({ keyStore, key, ttlSeconds, fetcher }: TWithCacheOpts<T>): Promise<T> => {
+  try {
+    const cached = await keyStore.getItem(key);
+    if (cached !== null) {
+      return JSON.parse(cached) as T;
+    }
+  } catch (err) {
+    logger.warn({ key, err }, `withCache: cache read failed, falling back to fetcher [key=${key}]`);
+  }
+
+  const result = await fetcher();
+
+  try {
+    await keyStore.setItemWithExpiry(key, ttlSeconds, JSON.stringify(result));
+  } catch (err) {
+    logger.warn({ key, err }, `withCache: cache write failed [key=${key}]`);
+  }
+
+  return result;
+};
--- a/backend/src/lib/cache/with-cache.ts
+++ b/backend/src/lib/cache/with-cache.ts
@@ -9,18 +9,24 @@ type TWithCacheOpts<T> = {
 };
 
 /**
- * Cache-aside helper: attempts to read from Redis, falls back to the fetcher on miss or Redis failure,
- * and writes the result back to Redis. Redis errors are caught and logged — the fetcher is always the
- * source of truth.
+ * Cache-aside helper: attempts to read from Redis, falls back to the fetcher on miss, Redis I/O failure,
+ * or invalid cached JSON, and writes the result back to Redis. Redis write errors are caught and logged
+ * — the fetcher is always the source of truth.
  */
 export const withCache = async <T>({ keyStore, key, ttlSeconds, fetcher }: TWithCacheOpts<T>): Promise<T> => {
+  let cached: string | null = null;
   try {
-    const cached = await keyStore.getItem(key);
-    if (cached !== null) {
-      return JSON.parse(cached) as T;
-    }
+    cached = await keyStore.getItem(key);
   } catch (err) {
     logger.warn({ key, err }, `withCache: cache read failed, falling back to fetcher [key=${key}]`);
+  }
+
+  if (cached !== null) {
+    try {
+      return JSON.parse(cached) as T;
+    } catch (err) {
+      logger.warn({ key, err }, `withCache: cache parse failed, falling back to fetcher [key=${key}]`);
+    }
   }
 
   const result = await fetcher();
--- a/backend/src/services/super-admin/super-admin-service.ts
+++ b/backend/src/services/super-admin/super-admin-service.ts
@@ -11,6 +11,7 @@ import {
 } from "@app/db/schemas";
 import { TLicenseServiceFactory } from "@app/ee/services/license/license-service";
 import { PgSqlLock, TKeyStoreFactory } from "@app/keystore/keystore";
+import { withCache } from "@app/lib/cache/with-cache";
 import {
   getConfig,
   getOriginalConfig,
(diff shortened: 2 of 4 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/0c3db83d-9ad2-4dae-9ebd-556b6894302d?refs=c3d5fe5489dd895d7ec96b8dc6d59ab3b11b0290&source=prChecksComment) |

backend/src/lib/cache/with-cache.ts
```diff
@@ -0,0 +1,41 @@
...
+  ttlSeconds: number;
+  fetcher: () => Promise<T>;
+};
+
+/**
+ * Cache-aside helper: attempts to read from Redis, falls back to the fetcher on miss, Redis I/O failure,
+ * or invalid cached JSON, and writes the result back to Redis. Redis write errors are caught and logged
+ * — the fetcher is always the source of truth.
+ */
+export const withCache = async <T>({ keyStore, key, ttlSeconds, fetcher }: TWithCacheOpts<T>): Promise<T> => {
```
PrestigePvP, line 16: Since this is a helper function, the approach of try-catching and swallowing an error is kinda concerning. 

If we do that because the specific libraries will throw an error and not a clean return we should scope the catch to that specific case and then fall through, rather than being a catch-all. 

For errors that we don't expect, we should catch-and-rethrow them. 

To give an example for this, if the redis errors because of connectivity it has nothing to do with our `withCache` function and it will continue to fail. We should surface this error.

backend/src/lib/cache/with-cache.ts
```diff
@@ -0,0 +1,41 @@
...
+  let cached: string | null = null;
+  try {
+    cached = await keyStore.getItem(key);
+  } catch (err) {
+    logger.warn({ key, err }, `withCache: cache read failed, falling back to fetcher [key=${key}]`);
+  }
+
+  if (cached !== null) {
+    try {
+      return JSON.parse(cached) as T;
```
PrestigePvP, line 26: Do we know that it will always be a JSON in redis? 

It seems over the top, but what would you think about encapsulating it in a parent object so that we know the type, and can make type-safe assertions?

backend/src/lib/cache/with-cache.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export const withCache = async <T>({ keyStore, key, ttlSeconds, fetcher }: TWithCacheOpts<T>): Promise<T> => {
+  let cached: string | null = null;
+  try {
+    cached = await keyStore.getItem(key);
+  } catch (err) {
+    logger.warn({ key, err }, `withCache: cache read failed, falling back to fetcher [key=${key}]`);
+  }
+
+  if (cached !== null) {
```
PrestigePvP, line 24: Can cached be undefined or an empty string? 

What if we made this `if (!!cached)` or `if (item)` so that we know it's truthy?

backend/src/lib/cache/with-cache.ts
```diff
@@ -0,0 +1,41 @@
...
+    try {
+      return JSON.parse(cached) as T;
+    } catch (err) {
+      logger.warn({ key, err }, `withCache: cache parse failed, falling back to fetcher [key=${key}]`);
+    }
+  }
+
+  const result = await fetcher();
```
PrestigePvP, line 32: Possibly unneeded, but what do we think about a try-catch then log and re-throw here? 
Makes debugging a bit easier?

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -150,26 +151,25 @@ export const superAdminServiceFactory = ({
...
     getServerCfg = async () => {
-      const config = await keyStore.getItem(ADMIN_CONFIG_KEY);
-
-      // missing in keystore means fetch from db
-      if (!config) {
-        const serverCfg = await serverCfgDAL.findById(ADMIN_CONFIG_DB_UUID);
-
-        if (!serverCfg) {
-          throw new NotFoundError({ message: "Admin config not found" });
+      const serverCfg = await withCache({
```
PrestigePvP, line 154: Love this implementation pattern. Very clean.

backend/src/services/upgrade-path/upgrade-path-service.ts
```diff
@@ -52,42 +53,35 @@ export const upgradePathServiceFactory = ({ keyStore }: TUpgradePathServiceFacto
...
-    try {
-      const yamlPath = path.join(__dirname, "..", "..", "..", "upgrade-path.yaml");
-      const resolvedPath = path.resolve(yamlPath);
-      const expectedBaseDir = path.resolve(__dirname, "..", "..", "..");
-      if (!resolvedPath.startsWith(expectedBaseDir)) {
-        throw new Error("Invalid configuration file path");
-      }
+    return withCache({
+      keyStore,
+      key: "upgrade-path:config",
```
PrestigePvP: Can we make this and the TTL const?

backend/src/services/upgrade-path/upgrade-path-service.ts
```diff
@@ -52,42 +53,35 @@ export const upgradePathServiceFactory = ({ keyStore }: TUpgradePathServiceFacto
...
+      ttlSeconds: 24 * 60 * 60,
+      fetcher: async () => {
```
PrestigePvP, line 64: I know this isn't net-new to your code but why are we caching this in redis at all? 

This should be a built constant

PrestigePvP (changes requested):
Overall I really like where we're headed with this pattern. 

I LOVE the unit-test. Super nice. I did leave a few comments tho.

backend/src/lib/cache/with-cache.ts
victorvhs017: I was on the fence on this one, and I think it depends on how we want to view Redis cache. For me, an error in the cache shouldn't prevent the request from going through.

In this case, the cache shouldn't stop the service by throwing an error, because we can still fetch the db. We are also not logging errors here because it shouldn't even pollute the logs.

I think that we should have a way to observe the errors, maybe some telemetry calls, but I don't think we really use that, right now?

backend/src/lib/cache/with-cache.ts
victorvhs017: We know because when we save, we stringify

```
await keyStore.setItemWithExpiry(key, ttlSeconds, JSON.stringify(result));
```

And we do that because we can't save the objects directly in Redis, it's all strings.

Regarding the type, we do use generics in the `withCache<T>`, so the type returned from the fetcher is forwarded through the withCache.

for example, when you inspect the editor:

image

backend/src/lib/cache/with-cache.ts
victorvhs017: Oh, this null check here is because of `await keyStore.getItem(key);`

If it returns `null`, it means that it's a cache miss.

For the value saved, we can and should save anything returned by the fetcher, even null is a valid value, and it's saved as "null" in Redis.

backend/src/lib/cache/with-cache.ts
victorvhs017: I think it falls in the same conversation we have here https://github.com/Infisical/infisical/pull/5987#discussion_r3065262782

I think this should be as transparent as possible regarding the fetcher function; we shouldn't handle anything or log anything that is not related specifically to the caching process.

Now, when the fetcher throws, it will just go through the `withCache`, and bubble up to whoever called it

backend/src/services/upgrade-path/upgrade-path-service.ts
victorvhs017: Not super acquainted to this code, too, it was just the perfect spot to switch to the new `withcache` hahaha

But from my understanding, we cache to avoid reading the yaml file, again? Agree that it doesn't seems like a huge improv

backend/src/lib/cache/with-cache.ts
PrestigePvP: WHat's the current behavior in places we would implement this?

backend/src/lib/cache/with-cache.ts
PrestigePvP: Right, so do we expect to be able to just store strings in redis anywhere?

backend/src/lib/cache/with-cache.ts
PrestigePvP: would that "null" not cause an exception to be thrown when `JSON.parse`

backend/src/services/upgrade-path/upgrade-path-service.ts
PrestigePvP: I think we should make a Linear issue to just remove this entire service. I talked to Maidul and he agrees. 

As for this change, I am OK to just resolve this and move forward for now.

backend/src/lib/cache/with-cache.ts
victorvhs017: Nope, in this case, it would be parsed to `null`

backend/src/lib/cache/with-cache.ts
victorvhs017: Not only strings, but we can also do arrays, objects, dates, anything that can be JSON serializable, because in the end it will be a string in redis.

backend/src/services/upgrade-path/upgrade-path-service.ts
victorvhs017: Alright! I created an issue for this in Linear.

backend/src/lib/cache/with-cache.ts
victorvhs017: It's a bit inconsistent on how it's done today because we didn't have a central function like this before, but looking at different places, I can see that we don't throw and fallback to the DB.

In the `getSecrets` function inside `secret-v2-bridge-service`, for example, we fallback to db.

So we already have this "degrade on cache failure" behavior.

backend/src/lib/cache/with-cache.ts
PrestigePvP: Yup - you right

PrestigePvP (approved):

PR #5990, opened 2026-04-09
feature(request-context): implement request-scoped memoization for optimized DB reads

victorvhs017:
## Context

Adds **request-scoped memoization** so repeated read-only work in a single HTTP request (permissions, project/user/identity lookups) hits the database once instead of many times—especially on batch secret paths.

**Before:** Each call path could repeat `getProjectPermission` / `getOrgPermission`, `projectDAL.findById`, and actor lookups independently within the same request.

**After:**

- Introduces `RequestMemoizer` (`getOrSet` with in-flight deduplication) stored on `@fastify/request-context` as `memoizer` (wired in `app.ts` per existing pattern in the branch).
- Adds `requestMemoize(key, fetcher)` which uses the memoizer when present and falls back to a direct fetch when there is no request context (e.g. workers).
- Centralizes memo key strings in `src/lib/request-context/memo-keys.ts` and request-context string keys in `request-context-keys.ts` (kept in sync with `RequestContextData` in `fastify.d.ts`).
- **Permission service:** `getOrgPermission` wrapped with `requestMemoize`; `getProjectPermission` uses `memoizer.getOrSet` for full-function memoization (key includes `actorOrgId`) and defers `requestContext.set` for `projectDetails` / `identityPermissionMetadata` until after the memoized load completes. DAL reads use `requestMemoize` for `projectFindById`, `userFindById`, and `identityFindById`.
- Replaces raw `requestContext.get("…")` / `set` string literals with `requestContextKeys` in touched files (logger, audit log, SAML/OIDC telemetry, assume-privilege router, permission service).
- Documents the pattern in `backend/CLAUDE.md` (Request-Scoped Memoization section).
- Unit tests in `request-memoizer.test.ts` for caching, in-flight coalescing, error behavior, and cross-request isolation.

## Steps to verify the change

1. **Knex query logging (temporary):** Enable SQL logging (e.g. Knex `debug` against the dev DB) while exercising a hot path that previously repeated lookups (e.g. batch secret read/list for one project).

`backend/src/db/instance.ts`

image

2. **API call:** Issue the same request shape you use in production or staging (same `projectId`, same auth) and compare **number of `SELECT` statements** (or log lines) before vs after—project load, permission resolution, and actor lookups should collapse to a single round-trip per distinct key within one request.

For example:

```
# Create a machine identity + access token if you don't have one
# Then fetch a single secret:
curl -H "Authorization: Bearer <identity-access-token>" \
  http://localhost:8080/api/v3/secrets/raw/<secretName>?workspaceId=<projectId>&environment=dev
```
  
```
# List all secrets (triggers getProjectPermission once, reused for all):
curl -H "Authorization: Bearer <identity-access-token>" \
  "http://localhost:8080/api/v3/secrets/raw?workspaceId=<projectId>&environment=dev"
```

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

  _Suggested title:_ `perf(backend): request-scoped memoization for permissions and project lookups`
- [x] Tested locally
- [x] Updated docs (if needed) — `backend/CLAUDE.md` updated for memoization

### DIFF
--- a/backend/src/ee/services/permission/permission-service.ts
+++ b/backend/src/ee/services/permission/permission-service.ts
@@ -25,6 +25,9 @@ import { KeyStorePrefixes, KeyStoreTtls, TKeyStoreFactory } from "@app/keystore/
 import { conditionsMatcher } from "@app/lib/casl";
 import { BadRequestError, ForbiddenRequestError, NotFoundError } from "@app/lib/errors";
 import { objectify } from "@app/lib/fn";
+import { requestMemoKeys } from "@app/lib/request-context/memo-keys";
+import { requestContextKeys } from "@app/lib/request-context/request-context-keys";
+import { requestMemoize } from "@app/lib/request-context/request-memoizer";
 import { ActorType } from "@app/services/auth/auth-type";
 import { TIdentityDALFactory } from "@app/services/identity/identity-dal";
 import { TProjectDALFactory } from "@app/services/project/project-dal";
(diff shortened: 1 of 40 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/36e9404c-fd8a-4559-b7b1-21d3325acd03?refs=32bf822329faec36f038543da5a0bcaeb7b99d5b&source=prChecksComment) |

backend/src/lib/request-context/request-context-keys.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Runtime string keys for `@fastify/request-context`.
+ * Keep in sync with `RequestContextData` in `src/@types/fastify.d.ts`.
+ */
+export const requestContextKeys = {
```
akhilmhdh: I think better we use enums

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -385,128 +398,176 @@ export const permissionServiceFactory = ({
...
+        throw new BadRequestError({
+          message: `The project is of type ${projectDetails.type}. Operations of type ${actionProjectType} are not allowed.`
         });
-        if (orgPermissionData?.length) {
-          const orgPermissionFromRoles = flattenActiveRolesFromMemberships(orgPermissionData, OrgMembershipRole.Custom);
-          const orgPermission = createMongoAbility<OrgPermissionSet>(buildOrgPermissionRules(orgPermissionFromRoles), {
-            conditionsMatcher
+      }
+
+      const permissionData = await permissionDAL.getPermission({
```
akhilmhdh, line 447: Is permission a big hitter? Generally i feel like this only called once per endpoint

akhilmhdh: My reasoning is - permission is a big object, so storing it in context may not be desirable

akhilmhdh (commented):
Overall looking good! Some questions

Pending application testing

akhilmhdh:
Application testing looking good. Just need to check on the above comments

backend/src/ee/services/permission/permission-service.ts
victorvhs017: I did some search and we have this pattern mostly on secret v3 endpoints and secrets overview, which would be enough reason to use it. But, I also find it good to have as a safeguard, because it's used a lot and we may miss a duplicated call here and there.

We can track the memory usage after the change, if the footprint rises too much, we remove it from here and fix the places we know directly

akhilmhdh (approved):

PR #6119, opened 2026-04-21
feat(secret-sync): add support for gitpod (Ona)

adilsitos:
## Context

Add a new secret sync for gitpod (now called Ona). 

## Steps to verify the change

There is a new documentation about this feature, I recommend use it so this is also reviewed. 

I also recommend test the two scopes the sync has: user and project level.  
some comments about  how this was designed: 
- Ona does not support fetching values if you are not calling the API though an environment (VM)
- The only type of secret we are taking into account now is env variable. 
- the sync works for project and user level secrets. They are expanding for org secrets now, but this is only for enterprise accounts (and don't seem to exist now)

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/0e0bd21b-995c-4497-a4c7-d89ad63765e5?refs=3c73cc03b778190d8d9f71b03b1b73b0aa5388a2&source=prChecksComment) |

docs/integrations/app-connections/ona.mdx
```diff
@@ -0,0 +1,112 @@
...
+  <Step title="Open your user settings on Ona">
+
+  Navigate to your Ona user settings.
+
+
+  </Step>
+  <Step title="Select the Personal Access Tokens tab">
+    In the bottom left, click your name, then click the gear icon to access your user settings.
+    
+      ![Open Account setting](/images/app-connections/ona/ona-account-settings.png)
```
varonix0: This screenshot should be apart of the previous step, it shows where the user settings button is

backend/src/services/app-connection/ona/ona-connection-fns.ts
```diff
@@ -0,0 +1,107 @@
...
+    methods: Object.values(OnaConnectionMethod) as [OnaConnectionMethod.PersonalAccessToken]
+  };
+};
+
+export const validateOnaConnectionCredentials = async (config: TOnaConnectionConfig) => {
+  const { credentials: inputCredentials } = config;
+
+  try {
+    await request.post(
+      `${IntegrationUrls.ONA_API_URL}${GET_AUTHENTICATED_IDENTITY_PATH}`,
```
varonix0: `IntegrationUrls` is not a pattern we use anymore for new secret syncs, it was mainly intended for native integrations. We can just define the ONA_API_URL in the file

backend/src/services/app-connection/ona/ona-connection-fns.ts
```diff
@@ -0,0 +1,107 @@
...
+      );
+
+      if (data?.projects?.length) {
+        allProjects.push(
+          ...data.projects
+            .map((project) => ({
+              id: project.id,
+              name: project.metadata?.name
+            }))
+            .filter((project): project is TOnaProject => Boolean(project.name))
```
varonix0: We don't need this assertion if we do `name: project.metadata?.name || ""` in the map function

backend/src/services/app-connection/ona/ona-connection-fns.ts
```diff
@@ -0,0 +1,107 @@
...
+      throw new BadRequestError({
+        message: `Failed to fetch Ona projects: ${
+          error.response?.data ? JSON.stringify(error.response?.data) : error.message || "Unknown error"
+        }`
+      });
+    }
+    throw error;
+  }
+
+  logger.info({ allProjects }, "Ona projects fetched successfully");
```
varonix0, line 98: Do we need this log?

backend/src/services/integration-auth/integration-list.ts
```diff
@@ -432,6 +433,15 @@ export const getIntegrationOptions = async () => {
       type: "sat",
       clientId: "",
       docsLink: ""
+    },
+    {
+      name: "Ona GitPod",
```
varonix0: Why did we add this? This creates a new entry for Ona in he native integrations (which we don't work on anymore)

Image

backend/src/services/secret-sync/ona/ona-sync-fns.ts
```diff
@@ -0,0 +1,236 @@
...
+const sleep = (ms: number) =>
+  new Promise<void>((resolve) => {
+    setTimeout(resolve, ms);
+  });
+
+const isRetryableOnaError = (error: unknown): boolean => {
+  if (!(error instanceof AxiosError)) return false;
+  if (!error.response) return true;
+  const { status } = error.response;
+  return status === 408 || status === 429;
```
varonix0: question: why are we checking for response code 408? does their API actually return this sometimes?

backend/src/services/secret-sync/ona/ona-sync-fns.ts
```diff
@@ -0,0 +1,236 @@
...
+  scope: TOnaScopeFilter
+): Promise<TOnaSecret[]> => {
+  const all: TOnaSecret[] = [];
+  let token: string | undefined;
+  let hasMore = true;
+
+  while (hasMore) {
+    // eslint-disable-next-line no-await-in-loop, @typescript-eslint/no-loop-func
+    const { data } = await withOnaRetry(() =>
+      request.post<TOnaListSecretsResponse>(
```
varonix0: nit: i'd avoid writing direct arrow function and instead do as below. makes it more clear for someone glancing over the code where the scope ends (applies to all relevent uses of `withOnaRetry`)

```typescript
  const { data } = await withOnaRetry(async () => {
    const res = await request.post<TOnaAuthenticatedIdentityResponse>(
      `${IntegrationUrls.ONA_API_URL}${ONA_GET_AUTHENTICATED_IDENTITY_PATH}`,
      {},
      { headers: getAuthHeaders(secretSync) }
    );

    return res;
  });
```

backend/src/services/secret-sync/ona/ona-sync-fns.ts
```diff
@@ -0,0 +1,236 @@
...
+    throw new SecretSyncError({
+      error,
+      secretKey
+    });
+  }
+  throw new SecretSyncError({
+    error,
+    secretKey
+  });
+};
```
varonix0: Why are we doing the if statement here? looks to be returning the same

backend/src/services/secret-sync/ona/ona-sync-fns.ts
```diff
@@ -0,0 +1,236 @@
...
+      try {
+        if (!prior) {
+          // eslint-disable-next-line no-await-in-loop
+          await createEnvVarSecret(secretSync, scope, key, secretMap[key].value);
+        } else {
+          // eslint-disable-next-line no-await-in-loop
+          await updateSecretValue(secretSync, prior.id, secretMap[key].value);
+        }
+      } catch (error) {
+        wrapApiError(error, key);
```
varonix0: I think we should just write the logic here. calling a function like this with a side-effect like throwing an error is unclear when first reading it. 

i'd move the logic into the actual catch statements instead of off-loading it to a helper function

backend/src/services/secret-sync/ona/ona-sync-schemas.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const OnaSyncDestinationConfigSchema = z.discriminatedUnion("scope", [
+  z.object({
+    scope: z.literal(OnaSyncScope.Project).describe(SecretSyncs.DESTINATION_CONFIG.ONA.scope),
+    projectId: z
+      .string()
+      .trim()
+      .min(1, "Ona project ID is required")
+      .describe(SecretSyncs.DESTINATION_CONFIG.ONA.projectId),
+    projectName: z
```
varonix0: why is the project name required?

docs/integrations/secret-syncs/ona.mdx
```diff
@@ -0,0 +1,185 @@
...
+description: "Learn how to configure an Ona Sync for Infisical."
+---
+
+**Prerequisites:**
+
+- Set up and add secrets to [Infisical Cloud](https://app.infisical.com)
+- Create an [Ona Connection](/integrations/app-connections/ona)
+
+<Note>
+    Ona Sync is **one-way** (Infisical → Ona). Importing existing Ona secrets into Infisical is not supported because Ona only exposes secret values from within active workspaces. Manage the source of truth in Infisical.
```
varonix0: I don't think we need this - seems anti-pattern compared to other syncs that don't support imports

varonix0 (changes requested):
One general concern though; I don't think we should allow syncing to the PAT user's personal secrets. Consider the scenario where someone sets up an Ona app connection, then a different team member creates a sync using that connection, they'd end up writing secrets into the connection creator's personal Ona account, which probably isn't what either party expects.

backend/src/services/app-connection/ona/ona-connection-fns.ts
adilsitos: if think this was a debug I forgot to remove xD

backend/src/services/secret-sync/ona/ona-sync-fns.ts
adilsitos: Nope, I just thought it would make sense that if the server takes too long to responde that we could try again in another request.

backend/src/services/secret-sync/ona/ona-sync-fns.ts
adilsitos: Didn't notice this. Deleted.

backend/src/services/integration-auth/integration-list.ts
adilsitos: didn't notice this. removed.

docs/integrations/secret-syncs/ona.mdx
adilsitos: removed.

backend/src/services/secret-sync/ona/ona-sync-schemas.ts
adilsitos: this is used to show the name of the ona project in the sync integration. If you think this is not necessary, I can remove it.

backend/src/services/app-connection/ona/ona-connection-fns.ts
adilsitos: removed

adilsitos:
The last commit fix this @varonix0 I saw it earlier, but forgot to fix 
image

backend/src/services/secret-sync/ona/ona-sync-schemas.ts
varonix0: I think we should make it optional, that way it's easier to create through the API

backend/src/services/secret-sync/ona/ona-sync-fns.ts
```diff
@@ -115,61 +86,50 @@ const createEnvVarSecret = async (
...
-    secretKey
+    );
+    return res;
   });
 };
 
 export const OnaSyncFns = {
   syncSecrets: async (secretSync: TOnaSyncWithCredentials, secretMap: TSecretMap) => {
-    const scope = await buildScopeFilter(secretSync);
+    const scope: TOnaScopeFilter = { projectId: secretSync.destinationConfig.projectId };
```
varonix0: Do we need this? Can't we just pass the `secretSync` directly into the functions that need it? Then we could remove the `TOnaScopeFilter` and the `scope` parameter in the `listEnvVarSecret`, etc. functions?

I see we're already passing the secretSync variable into those functions

backend/src/services/secret-sync/ona/ona-sync-fns.ts
```diff
@@ -217,7 +179,7 @@ export const OnaSyncFns = {
   },
 
   removeSecrets: async (secretSync: TOnaSyncWithCredentials, secretMap: TSecretMap) => {
-    const scope = await buildScopeFilter(secretSync);
+    const scope: TOnaScopeFilter = { projectId: secretSync.destinationConfig.projectId };
```
varonix0: Same as above

backend/src/services/secret-sync/ona/ona-sync-types.ts
```diff
@@ -16,7 +16,6 @@ export type TOnaSyncWithCredentials = TOnaSync & {
 
 export type TOnaScopeFilter = {
```
varonix0: And we can remove this type entirely

varonix0 (changes requested):

backend/src/services/secret-sync/ona/ona-sync-fns.ts
adilsitos: We don't need it. This was a artifact of the previous implementation that had the user. Fixed.

varonix0 (dismissed):

varonix0 (approved):

PR #6151, opened 2026-04-24
fix(casl): scope cert list/cert-request list/cert dashboards by metadata in SQL; guardrail translator on unknown operators

saifsmailbox98:
# fix(casl): teach CASL→SQL translator about `$elemMatch`, add unknown-operator guardrail, and scope cert list / cert request list / cert dashboards by metadata in SQL

> This template is acting as a review artifact for an uncommitted branch. It is intentionally verbose.
>
> Companion branch: `saif/pki-146-fix-shortterm-cert-metadata` — same bug, per-row in-memory filter with dashboard-denial fallback. Compare before deciding which to ship.

## Context

A security finding on PKI-146 reported that when a role scopes `Read Certificates` by `metadata` (via a CASL `$elemMatch` condition), the cert list, cert-request list, and certificate dashboard endpoints return data for every certificate in the project rather than the scoped subset. Private keys remain protected, but certificate metadata (CN, altNames, serial, friendly name, metadata tags) and aggregate stats (counts, activity trend buckets, PQC readiness trends) leak.

### Root cause

`backend/src/lib/casl/permission-filter-utils.ts::processCondition` (lines ~102-156) translates CASL MongoDB-style conditions into SQL WHERE clauses. It recognises `$regex`/`$eq`/`$in`/`$glob`/`$ne` and **silently drops anything else**, including `$elemMatch`. A rule like `{ metadata: { $elemMatch: { key: { $eq: "env" }, value: { $eq: "prod" } } } }` compiles to `allowRules: [{}]`, which `applyProcessedPermissionRulesToQuery` converts to zero SQL constraints. There is no per-row fallback on cert list paths. Cert requests use the same `Certificates` CASL subject and inherit the hole. Dashboards do only a subject-class-level `throwUnlessCan(Read, Certificates)` and run project-wide aggregate SQL with no metadata scoping.

### Why this branch takes the translator-level approach

The companion short-term branch matches the existing per-row in-memory convention used by dynamic secrets / PAM accounts / PAM resources. It's simpler and lower-risk, but it has two structural weaknesses:

1. **Dashboards can't be scoped with a per-row filter.** Aggregates don't have rows to iterate — the short-term branch resorts to denying dashboard endpoints entirely for metadata-scoped roles.
2. **The silent-drop failure mode persists.** The next CASL operator someone adds to a permission schema will ship with the same bug if it isn't also added to the translator.

This branch fixes both: the translator learns about `$elemMatch`, metadata flows through SQL so dashboards can be scoped natively, and the translator throws on any operator it doesn't recognize so the next mismatch can't ship silently.

### 1. Translator — `backend/src/lib/casl/permission-filter-utils.ts`

- `processCondition` gets an explicit `$elemMatch` branch. Instead of emitting a SQL filter, it extracts the inner `{ key: { $eq | $in }, value: { $eq | $in } }` structure into `{ key: string; value?: string }` entries and pushes them into a new per-rule `metadataFilter` bucket. `$in` on either key or value expands into multiple entries (OR-ed at the metadata level).
- **Guardrail:** if `value` is an object whose keys are `$`-prefixed CASL operators but none of them are in the recognized set (`$regex`, `$eq`, `$in`, `$glob`, `$ne`, `$elemMatch`), `processCondition` throws `BadRequestError("Unknown CASL operator … — translator is missing a handler.")`. This forces any future operator addition to the CASL permission schemas to be accompanied by a matching translator branch.
- `ProcessedPermissionRules` grows a third field, `metadataFilter: Array<{ key: string; value? }>`, populated by merging the buckets from every matching allow rule. `forbidRules` remain shape-unchanged.
- **Inverted rules containing `$elemMatch` throw.** `applyMetadataFilter` is a positive `whereExists`; a semantically correct `NOT EXISTS (x AND y)` isn't what the primitive expresses, and no current cert role uses inversion + metadata. If we ever need this, we add it with its own semantics.

### 2. Cert list — `project-service.ts::listProjectCertificates` + `certificate-dal.ts`

- Service pulls `metadataFilter` off `processedRules`, merges (AND) with the user-supplied `regularFilters.metadataFilter`, passes the combined array to the DAL for both the list and the count query.
- DAL methods `findWithPrivateKeyInfo`, `findActiveCertificatesForSync`, `countCertificatesInProject` accept the merged filter and feed it through the existing `applyMetadataFilter(query, metadataFilter, "certificateId", TableName.Certificate)` primitive (already used today for the user-supplied filter).
- The existing SQL filter via `applyProcessedPermissionRulesToQuery` is left in place — it still handles `commonName`/`status`/`friendlyName` correctly at the DB layer.
- `findActiveCertificatesForSync` previously looped over `Object.entries(filter)` and called `andWhere(key, value)` for every key. The array-typed `metadataFilter` would have SQL-broken this loop. Extracted metadata out of that loop and routed through `applyMetadataFilter` instead. Minor cleanup forced by the new shape.
- `totalCount` is now the correct post-filter SQL count — **no quirk**.

### 3. Cert request list — `certificate-request-service.ts` + `certificate-request-dal.ts`

Same pattern. CASL `metadataFilter` merges with user-supplied filter; both threaded through existing `applyMetadataFilter` calls on `certificateRequestId`. Works for `findByProjectIdWithCertificate` + `countByProjectId`.

### 4. Dashboards — `project-service.ts::{getDashboardStats, getActivityTrend, getPqcTrend}` + `certificate-dal.ts`

- Services extract `metadataFilter` from `getProcessedPermissionRules` and pass it to the DAL alongside the `projectId`.
- DAL dashboard methods (`getDashboardStats` @ `:636`, `getActivityTrend`, `getPqcTrend`) accept the optional `metadataFilter`. For **every** aggregate subquery — totals, `byAlgorithm`, `byCA`, `byEnrollmentMethod`, trend buckets, PQC buckets — they call `applyMetadataFilter(query, metadataFilter, "certificateId", TableName.Certificate)` so the `whereExists` subquery scopes the COUNT to only certs the user is authorised to see.
- `byStatus` is derived from `totals` and inherits scoping automatically.
- **Cache key correctness.** Dashboard results are cached in the keystore. Without a per-scope key, one metadata-scoped user's filtered aggregates would be served to another user via the cache. A `metadataFilterCacheHash(filter)` helper (factory-scoped, reused across all three dashboard methods) produces a stable 16-char sha256 hash from the sorted filter entries; `""` for unscoped users so the global cache is shared. `KeyStorePrefixes.CertDashboardStats`, `.CertActivityTrend`, `.CertPqcTrend` were updated to accept the hash as part of the key.

### 5. Unit tests — `permission-filter-utils.test.ts` (new)

Six tests, all passing:

1. `$glob` on a regular field — returns an `allowRules` filter, empty `metadataFilter`.
2. `$elemMatch` single entry — empty `allowRules` filter on metadata, `metadataFilter` populated.
3. `$elemMatch` with `$in` expansion — multiple entries in `metadataFilter`.
4. `$elemMatch` key-only (no value constraint) — `metadataFilter` entry with `value: undefined`.
5. Combined condition (`commonName` + `metadata`) — both structures populated, compose correctly.
6. Unknown operator → throws `BadRequestError` with the guardrail message.

Two existing test files (`certificate-policy-service.test.ts`, `certificate-profile-service.test.ts`) were updated to include the new `metadataFilter: []` field in their `ProcessedPermissionRules` fixtures. These assertions were pinning the old shape; they now match the new one.

Full unit test suite: 426 pass / 22 skipped / 0 failed.

## Translator audit (callers of `getProcessedPermissionRules` / `applyProcessedPermissionRulesToQuery`)

Per the guardrail, any subject whose rules already contain an operator outside the recognized set would start throwing. Audited:

| Caller | Subject | Conditions schema operators used |
|---|---|---|
| `project-service.ts::listProjectCertificates` | `Certificates` | `$eq`, `$ne`, `$in`, `$glob`, `$elemMatch` — all handled |
| `certificate-request-service.ts::listCertificateRequests` | `Certificates` | Shares `Certificates` subject — same set |
| `certificate-authority-service.ts` | `CertificateAuthorities` | `$eq`, `$ne`, `$in`, `$glob` — all handled |
| `certificate-profile-service.ts` | `CertificateProfiles` | `$eq`, `$ne`, `$in`, `$glob` — all handled |
| `certificate-policy-service.ts` | `CertificatePolicies` | `$eq`, `$ne`, `$in`, `$glob` — all handled |
| `pki-sync-service.ts` | `PkiSyncs` | `$eq`, `$ne`, `$in`, `$glob` — all handled |

No schema references an operator the translator doesn't handle. Guardrail will not fire on any role configured through the API.

## Files changed

| File | Change |
|---|---|
| `backend/src/lib/casl/permission-filter-utils.ts` | `$elemMatch` branch, unknown-operator guardrail, `ProcessedPermissionRules.metadataFilter` |
| `backend/src/lib/casl/permission-filter-utils.test.ts` | New — 6 unit tests |
| `backend/src/keystore/keystore.ts` | Dashboard cache keys accept metadata-filter hash |
| `backend/src/services/certificate/certificate-dal.ts` | `metadataFilter` plumbed into list, sync-list, count, and all three dashboard methods with `applyMetadataFilter` on every aggregate subquery |
| `backend/src/services/project/project-service.ts` | CASL metadata merged with user metadata in list; dashboards extract metadata + use hashed cache key |
| `backend/src/services/certificate-request/certificate-request-service.ts` | CASL metadata merged with user metadata in cert-request list |
| `backend/src/services/certificate-policy/certificate-policy-service.test.ts` | Fixture updated for new `ProcessedPermissionRules` shape |
| `backend/src/services/certificate-profile/certificate-profile-service.test.ts` | Same |
| `backend/src/lib/knex/scim.ts` | Incidental `lint:fix` import-order autofix. Unrelated to this fix. |

## Downsides (be aware before merging)

- **Shared utility touched.** `permission-filter-utils.ts` is used by every subject that uses `getProcessedPermissionRules`. Audit is included above, but anyone adding a new CASL operator to any project/org permission schema must now also teach the translator — that's the explicit intent, but it's a mechanical step future PRs will need to remember. The guardrail throws will make this loud.
- **Multi-rule metadata gets AND semantics, not OR.** CASL evaluates multiple allow rules as OR; `applyMetadataFilter` ANDs entries. When more than one matching allow rule carries `$elemMatch` conditions, the translator logs a console warning and ANDs the entries (stricter than CASL — safer direction, no leaks, potentially smaller result set than strict CASL semantics would yield). Real OR support is a future extension.
- **Forbid rules containing `$elemMatch` throw.** Inverted metadata scoping isn't currently expressible; if product ever exposes it, we'll need to add dedicated semantics (tricky — it's NOT-EXISTS with inner AND).
- **`whereExists` query plan sensitivity.** `resource_metadata` is joined per aggregate subquery. On large tenants without an index on `(certificateId, key, value)`, the subquery can be slow. Verify with `EXPLAIN` on a representative dataset before rolling to high-cardinality customers. Same concern applies to `(certificateRequestId, key, value)`.
- **Three new patterns for the team to absorb:** (1) CASL conditions feeding `applyMetadataFilter` via a bucket on `ProcessedPermissionRules`; (2) `$elemMatch` recognition in the translator; (3) per-scope cache-key hashing for dashboards. Higher review + onboarding cost than the companion branch.
- **Dynamic secrets / PAM continue to use the in-memory per-row pattern.** Only the `Certificates` subject is migrated to the SQL path here. Consistency across subjects is a follow-up if it ever matters.

## Risk assessment

- Full unit-test suite passes; two pre-existing test files had fixtures pinning the old `ProcessedPermissionRules` shape and were updated (no behavior assertions changed).
- Translator audit (table above) confirms no existing role hits the guardrail.
- Dashboard cache key includes scope — verified by reading the three dashboard methods. Two users with different metadata scopes on the same project produce different cache keys; a user with no metadata scope produces `""` (shared with the old global cache).
- No schema / migration changes. `resource_metadata.certificateId` and `.certificateRequestId` columns already exist on main.

## Steps to verify the change

End-to-end:

1. Seed two certificates — one with `metadata { env: prod }`, one with `metadata { env: staging }`.
2. Role: grant `Read Certificates` scoped by `{ metadata: { $elemMatch: { key: { $eq: "env" }, value: { $eq: "prod" } } } }`.
3. `GET /api/v2/workspace/:projectId/certificates` — expect only the prod cert; `totalCount === 1`. Repeat for cert requests with prod-tagged + staging-tagged requests — expect only the prod-tagged one.
4. `GET /api/v2/workspace/:projectId/certificates/dashboard-stats` — expect `totals.total === 1`, `byAlgorithm`/`byCA` reflect only the prod cert. Same for activity trend and PQC trend — buckets count only prod certs.
5. Combined-rule test: `{ commonName: { $glob: "prod-*" }, metadata: { $elemMatch: … } }` — list returns certs where CN matches AND metadata matches.
6. Cache scoping test: create two roles on the same project with different metadata scopes, hit dashboard as each — each sees their own numbers; hit as an unscoped user — sees global numbers. Confirm three distinct cache keys in Redis / keystore.
7. **Guardrail sanity:** temporarily add an unknown operator (e.g. `{ commonName: { $notAThing: "x" } }`) to a role via the service-layer helper — expect `BadRequestError("Unknown CASL operator …")` at list time. Remove after testing.

Local:

- `cd backend && make reviewable-api` — PASS on this branch.
- `cd backend && npm run test:unit -- permission-filter-utils` — new 6-test file passes.
- `cd backend && npm run test:unit` — full suite: 426 pass, 22 skipped, 0 failed.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally
- [ ] Updated docs (if needed) — document the new translator-guardrail expectation for anyone extending CASL permission schemas
- [ ] Updated CLAUDE.md files (if needed) — worth noting in `backend/CLAUDE.md` that any new CASL operator addition must also extend `permission-filter-utils.ts`

### DIFF
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -100,9 +100,15 @@ export const KeyStorePrefixes = {
   TelemetryGroupIdentify: (orgId: string) => `telemetry-group-identify:${orgId}` as const,
   SecretEtag: (projectId: string) => `secret-etag:${projectId}` as const,
 
-  CertDashboardStats: (projectId: string) => `cert-dashboard-stats:${projectId}` as const,
-  CertActivityTrend: (projectId: string, range: string) => `cert-activity-trend:${projectId}:${range}` as const,
-  CertPqcTrend: (projectId: string, range: string) => `cert-pqc-trend:${projectId}:${range}` as const,
+  // `metadataFilterHash` scopes dashboards per metadata-filter so users with
+  // different RBAC metadata scopes don't share cached aggregates. Pass "" for
+  // unscoped callers (the common case).
+  CertDashboardStats: (projectId: string, metadataFilterHash: string = "") =>
+    `cert-dashboard-stats:${projectId}:${metadataFilterHash}` as const,
+  CertActivityTrend: (projectId: string, range: string, metadataFilterHash: string = "") =>
+    `cert-activity-trend:${projectId}:${range}:${metadataFilterHash}` as const,
+  CertPqcTrend: (projectId: string, range: string, metadataFilterHash: string = "") =>
+    `cert-pqc-trend:${projectId}:${range}:${metadataFilterHash}` as const,
   RefreshTokenGrace: (sessionId: string) => `refresh-token-grace:${sessionId}` as const,
   InsightsCache: (projectId: string, endpoint: string) => `insights-cache:${projectId}:${endpoint}` as const
 };
--- a/backend/src/lib/casl/permission-filter-utils.test.ts
+++ b/backend/src/lib/casl/permission-filter-utils.test.ts
@@ -1,0 +1,137 @@
+import { createMongoAbility } from "@casl/ability";
+
+import { BadRequestError } from "@app/lib/errors";
+
+import { PermissionConditionOperators } from ".";
+import { getProcessedPermissionRules } from "./permission-filter-utils";
+
+describe("getProcessedPermissionRules", () => {
+  const action = "read";
+  const subject = "certificates";
+
+  test("translates $glob condition on a regular field into allowRules with empty metadataFilter", () => {
+    const ability = createMongoAbility([
+      {
+        action,
+        subject,
+        conditions: {
+          commonName: { [PermissionConditionOperators.$GLOB]: "prod-*" }
+        }
+      }
+    ]);
+
+    const result = getProcessedPermissionRules(ability, action, subject);
+
+    expect(result.forbidRules).toHaveLength(0);
+    expect(result.metadataFilter).toEqual([]);
+    expect(result.allowRules).toHaveLength(1);
+    expect(result.allowRules[0].commonName).toHaveLength(1);
+    expect(result.allowRules[0].commonName[0]).toMatchObject({ operator: "LIKE", value: "prod-*", isPattern: true });
+  });
+
+  test("extracts metadata $elemMatch into metadataFilter (single entry)", () => {
+    const ability = createMongoAbility([
+      {
+        action,
+        subject,
+        conditions: {
+          metadata: {
+            [PermissionConditionOperators.$ELEMENTMATCH]: {
+              key: { [PermissionConditionOperators.$EQ]: "env" },
+              value: { [PermissionConditionOperators.$EQ]: "prod" }
+            }
+          }
+        }
+      }
+    ]);
+
+    const result = getProcessedPermissionRules(ability, action, subject);
+
+    expect(result.metadataFilter).toEqual([{ key: "env", value: "prod" }]);
+    // The metadata field itself must not leak into SQL-translated allowRules.
+    expect(result.allowRules.every((rule) => !("metadata" in rule))).toBe(true);
+  });
+
+  test("expands $in under $elemMatch.value into multiple entries", () => {
+    const ability = createMongoAbility([
+      {
+        action,
+        subject,
+        conditions: {
+          metadata: {
+            [PermissionConditionOperators.$ELEMENTMATCH]: {
+              key: { [PermissionConditionOperators.$EQ]: "env" },
+              value: { [PermissionConditionOperators.$IN]: ["prod", "staging"] }
+            }
+          }
+        }
+      }
+    ]);
+
+    const result = getProcessedPermissionRules(ability, action, subject);
+
+    expect(result.metadataFilter).toEqual([
+      { key: "env", value: "prod" },
+      { key: "env", value: "staging" }
+    ]);
+  });
+
+  test("handles $elemMatch with only a key constraint (no value)", () => {
+    const ability = createMongoAbility([
+      {
+        action,
+        subject,
+        conditions: {
+          metadata: {
+            [PermissionConditionOperators.$ELEMENTMATCH]: {
+              key: { [PermissionConditionOperators.$EQ]: "env" }
+            }
+          }
+        }
+      }
+    ]);
+
+    const result = getProcessedPermissionRules(ability, action, subject);
+
+    expect(result.metadataFilter).toEqual([{ key: "env" }]);
+  });
+
+  test("combined conditions populate both allowRules and metadataFilter", () => {
+    const ability = createMongoAbility([
+      {
+        action,
+        subject,
+        conditions: {
+          commonName: { [PermissionConditionOperators.$GLOB]: "prod-*" },
+          metadata: {
+            [PermissionConditionOperators.$ELEMENTMATCH]: {
+              key: { [PermissionConditionOperators.$EQ]: "env" },
+              value: { [PermissionConditionOperators.$EQ]: "prod" }
+            }
+          }
+        }
+      }
+    ]);
+
+    const result = getProcessedPermissionRules(ability, action, subject);
+
+    expect(result.allowRules).toHaveLength(1);
+    expect(result.allowRules[0].commonName).toHaveLength(1);
+    expect(result.metadataFilter).toEqual([{ key: "env", value: "prod" }]);
+  });
+
+  test("throws BadRequestError on unknown operator in a condition", () => {
+    const ability = createMongoAbility([
+      {
+        action,
+        subject,
+        conditions: {
+          commonName: { $notAThing: "x" } as unknown as Record<string, unknown>
+        }
+      }
+    ]);
+
+    expect(() => getProcessedPermissionRules(ability, action, subject)).toThrow(BadRequestError);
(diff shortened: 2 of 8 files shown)


maidul98:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/maidul98/pr-checks/63d77ffc-eca8-45ac-976b-1bb77c5867c2?refs=2de9c471790d1babe63e7e2f817b176918773b7e&source=prChecksComment) |

PR #6214, opened 2026-04-27
feat(e2e-tests): fips mode

varonix0:
## Context

Updated E2E test flow to use FIPS mode in Github CI. The reason we can't run it directly on the bare runner (like before) is that FIPS mode requires OpenSSL compiled with enable-fips. `crypto.setFips(true)` will throw if the FIPS module isn't present in the Node.js runtime.

The backend image build step is cached, so when possible it will use the cached image, greatly reducing build time (such as if only frontend changes are involved)

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/.github/workflows/run-backend-tests.yml
+++ b/.github/workflows/run-backend-tests.yml
@@ -14,7 +14,7 @@ jobs:
   check-be-pr:
     name: Run integration test
     runs-on: ubuntu-latest
-    timeout-minutes: 30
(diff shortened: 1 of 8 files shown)


scott-ray-wilson (approved):
LGTM, passing in CI and changes make sense

PR #6220, opened 2026-04-28
improvement(secrets-rotation): allow move between environments and folders

adilsitos:
## Context
Add the possibility to move secret rotations between environments

## Steps to verify the change
- Create a new secret rotation 
- Move between folders and environments 
- check that the rotation still works after the move
## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/routes/v2/secret-rotation-v2-routers/secret-rotation-v2-endpoints.ts
+++ b/backend/src/ee/routes/v2/secret-rotation-v2-routers/secret-rotation-v2-endpoints.ts
@@ -454,7 +454,7 @@ export const registerSecretRotationEndpoints = <
 
   server.route({
     method: "POST",
-    url: "/move/:rotationId",
+    url: "/:rotationId/move",
     config: {
       rateLimit: writeLimit
     },
@@ -468,7 +468,8 @@ export const registerSecretRotationEndpoints = <
       }),
       body: z.object({
         destinationEnvironment: z.string().trim().min(1, "Destination environment required"),
-        destinationSecretPath: z.string().trim().min(1, "Destination secret path required")
+        destinationSecretPath: z.string().trim().min(1, "Destination secret path required"),
+        overwriteDestination: z.boolean().default(false)
       }),
       response: {
         200: z.object({ secretRotation: responseSchema })
@@ -477,11 +478,11 @@ export const registerSecretRotationEndpoints = <
     onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
     handler: async (req) => {
       const { rotationId } = req.params;
-      const { destinationEnvironment, destinationSecretPath } = req.body;
+      const { destinationEnvironment, destinationSecretPath, overwriteDestination } = req.body;
 
       const { secretRotation, sourceEnvironment, sourceSecretPath } =
         await server.services.secretRotationV2.moveSecretRotation(
-          { rotationId, type, destinationEnvironment, destinationSecretPath },
+          { rotationId, type, destinationEnvironment, destinationSecretPath, overwriteDestination },
           req.permission
         );
 
--- a/backend/src/ee/routes/v2/secret-rotation-v2-routers/secret-rotation-v2-endpoints.ts
+++ b/backend/src/ee/routes/v2/secret-rotation-v2-routers/secret-rotation-v2-endpoints.ts
@@ -454,6 +454,61 @@ export const registerSecretRotationEndpoints = <
 
   server.route({
     method: "POST",
+    url: "/move/:rotationId",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      hide: false,
+      operationId: `move${rotationTypeId}Rotation`,
+      tags: [ApiDocsTags.SecretRotations],
+      description: `Move the specified ${rotationType} Rotation to a different secret path or environment.`,
+      params: z.object({
+        rotationId: z.string().uuid().describe(SecretRotations.UPDATE(type).rotationId)
+      }),
+      body: z.object({
+        destinationEnvironment: z.string().trim().min(1, "Destination environment required"),
+        destinationSecretPath: z.string().trim().min(1, "Destination secret path required")
+      }),
+      response: {
+        200: z.object({ secretRotation: responseSchema })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const { rotationId } = req.params;
+      const { destinationEnvironment, destinationSecretPath } = req.body;
+
+      const { secretRotation, sourceEnvironment, sourceSecretPath } =
+        await server.services.secretRotationV2.moveSecretRotation(
+          { rotationId, type, destinationEnvironment, destinationSecretPath },
+          req.permission
+        );
+
+      const movedRotation = secretRotation as T;
+
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        projectId: movedRotation.projectId,
+        event: {
+          type: EventType.MOVE_SECRET_ROTATION,
+          metadata: {
+            type,
+            rotationId,
+            sourceEnvironment,
+            sourceSecretPath,
+            destinationEnvironment,
+            destinationSecretPath
+          }
+        }
+      });
+
+      return { secretRotation: movedRotation };
+    }
+  });
+
+  server.route({
+    method: "POST",
     url: "/:rotationId/rotate-secrets",
     config: {
       rateLimit: writeLimit
--- a/backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
+++ b/backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
@@ -81,6 +81,7 @@ import {
   fnSecretBulkDelete,
   fnSecretBulkInsert,
   fnSecretBulkUpdate,
+  fnUpdateMovedSecretReferences,
   reshapeBridgeSecret
 } from "@app/services/secret-v2-bridge/secret-v2-bridge-fns";
 import { TSecretVersionV2DALFactory } from "@app/services/secret-v2-bridge/secret-version-dal";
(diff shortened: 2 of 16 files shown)


backend/src/services/secret-sync/aws-parameter-store/aws-parameter-store-sync-fns.ts
```diff
@@ -470,7 +470,7 @@ export const AwsParameterStoreSyncFns = {
 
       if (!(key in secretMap) || !secretMap[key].value) {
         parametersToDelete.push(parameter);
-        deletedSecretKeys.push(parameter.Name!);
+        deletedSecretKeys.push(parameter.Name);
```
adilsitos: the linter removed this by themself

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
scott-ray-wilson, line 1: I think instead of disabling this we should just no-op on them; IE just ignore them for the sake of moving otherwise we could check if the rotation they belong to is selected and base the disabling on that - because in this scenario I don't feel like I should be blocked, I have the rotation and the secrets selected so they would be moved anyway

whatever we decide we should make delete button consistent behavior

Image

scott-ray-wilson, line 1: not related to your PR but could you add a query invalidation on move so that I don't have to refresh to see moved secrets

https://github.com/user-attachments/assets/41d021d8-e0d9-4c5d-9af5-b80cc788cae2

scott-ray-wilson, line 1: I can't move a rotation when a secret exists in the destination with that name despite having override (not I think this should only be handled for secrets; if the rotation name conflicts it should still throw

https://github.com/user-attachments/assets/f7d645b0-7f80-450b-b697-b271be35ef13

)

scott-ray-wilson, line 1: unrelated to your PR but we should patch here; I can move a secret with overwrite existing secrets to replace a rotated secret value (this should always throw even if overwrite because it's owned by the rotation)

https://github.com/user-attachments/assets/1838548e-48a3-4636-b502-7ada3d14af60

scott-ray-wilson, line 1: nit: 

redundant messages

Image

backend/src/ee/routes/v2/secret-rotation-v2-routers/secret-rotation-v2-endpoints.ts
```diff
@@ -452,6 +452,61 @@ export const registerSecretRotationEndpoints = <
     }
   });
 
+  server.route({
+    method: "POST",
+    url: "/move/:rotationId",
```
scott-ray-wilson: nit: let's invert it to match the other ones:

url: "/:rotationId/generated-credentials",

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
```diff
@@ -928,6 +942,159 @@ export const secretRotationV2ServiceFactory = ({
...
+        message: `A Secret Rotation with the name "${secretRotation.name}" already exists at the secret path "${destinationSecretPath}"`
+      });
+
+    const mappedKeys = Object.values(secretsMapping as TSecretRotationV2["secretsMapping"]);
+
+    await $throwOnConflictingSecrets({
+      secretKeys: mappedKeys,
+      folderId: destinationFolder.id,
+      secretPath: destinationSecretPath
+    });
```
scott-ray-wilson: yeah so I'm thinking here we add a parameter: overwriteSecrets or something where as long as the conflicting secrets are rotated secrets then we allow the overwrite? if false we keep this? or maybe this is over complicating it and we keep as is...

backend/src/services/secret-sync/aws-parameter-store/aws-parameter-store-sync-fns.ts
scott-ray-wilson, line 1: is this intended?

frontend/src/pages/secret-manager/OverviewPage/components/SelectionPanel/components/MoveSecretsDialog/MoveSecretsDialog.tsx
```diff
@@ -237,44 +244,100 @@ const SingleEnvContent = ({
...
       });
-    } else if (isSourceUpdated) {
-      createNotification({
-        type: "info",
-        text: "A secret approval request has been generated in the destination"
-      });
-    } else {
+    }
+
+    if (rotationFailures.length > 0) {
```
scott-ray-wilson: not sure we need this one as we get it form mutation too (see other comment)

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
```diff
@@ -928,6 +942,159 @@ export const secretRotationV2ServiceFactory = ({
     return expandSecretRotation(secretRotation, kmsService);
   };
 
+  const moveSecretRotation = async (
```
scott-ray-wilson, line 956: I think we will need to trigger a snapshot and secret sync like the normal move endpoints

- `snapshotService.performSnapshot(destinationFolder.id)` and `performSnapshot(sourceFolder.id)`
- `secretQueueService.syncSecrets({ ..., destinationFolder })` and `syncSecrets({ ..., sourceFolder })` (with `SecretImportMutation` events)

see the move endpoint/service for exact usage

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
```diff
@@ -928,6 +942,159 @@ export const secretRotationV2ServiceFactory = ({
...
+        environment: destinationEnvironment,
+        secretPath: destinationSecretPath
+      })
+    );
+
+    const conflictingRotation = await secretRotationV2DAL.findOne({
+      name: secretRotation.name,
+      folderId: destinationFolder.id
+    });
+
```
scott-ray-wilson: we should move this to the transaction block below

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
```diff
@@ -928,6 +942,159 @@ export const secretRotationV2ServiceFactory = ({
...
+    if (conflictingRotation)
+      throw new BadRequestError({
+        message: `A Secret Rotation with the name "${secretRotation.name}" already exists at the secret path "${destinationSecretPath}"`
+      });
+
+    const mappedKeys = Object.values(secretsMapping as TSecretRotationV2["secretsMapping"]);
+
+    await $throwOnConflictingSecrets({
```
scott-ray-wilson: this should also move to the transaction block below

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
```diff
@@ -928,6 +942,159 @@ export const secretRotationV2ServiceFactory = ({
...
+    });
+
+    const { encryptor: secretManagerEncryptor, decryptor: secretManagerDecryptor } =
+      await kmsService.createCipherPairWithDataKey({
+        type: KmsDataKey.SecretManager,
+        projectId
+      });
+
+    const updatedRotation = await secretRotationV2DAL.transaction(async (tx) => {
+      const sourceSecrets = await secretV2BridgeDAL.find(
```
scott-ray-wilson, line 1067: lets also check if a rotation is in progress and throw, incase there are going to be any writes

scott-ray-wilson (changes requested):
looks great overall! few concerns and some nits/thoughts

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
adilsitos: What is the rationale to have this inside the transaction? Since this is a validation, I didn't see a reason to have this there. Same case for the throwOnConflictingSecrets

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
adilsitos: I think this is not that hard, will experiment with this.

backend/src/services/secret-sync/aws-parameter-store/aws-parameter-store-sync-fns.ts
adilsitos: it was. The previous change was breaking my lint (I think I got into this file and a change happened there and I changed to this) but I'll revert this outside my code editor, so we don't have any problem.

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
adilsitos: Added an implementation that fix this 🙏

frontend/src/pages/secret-manager/OverviewPage/components/SelectionPanel/components/MoveSecretsDialog/MoveSecretsDialog.tsx
adilsitos: removed.

backend/src/ee/services/secret-rotation-v2/secret-rotation-v2-service.ts
adilsitos: Added!

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
adilsitos: This is the only one I didn't change (as we talked in private). My two cents: I tried to replicate the behavior we have when a folder is selected and we try to move it, but I agree that it is not the best ux.

scott-ray-wilson (approved):
LGTM!

PR #6229, opened 2026-04-28
feat(smtp): add SMTP_HELO_HOST to set the EHLO/HELO hostname

### LINKED ISSUE #6228: Add SMTP_HELO_HOST (or similar) env var to set the EHLO/HELO hostname for the SMTP client
### Describe the bug
Infisical does not expose any way to configure the hostname that the SMTP client announces in the EHLO/HELO command. As a result, nodemailer falls back to Node's os.hostname(). Inside a container - e.g. on Google CloudRun, Kubernetes, ECS - os.hostname() returns a random/internal identifier (a container ID or a name like localhost), which is rejected by SMTP relays that validate the sender hostname.
                                                                                                                                                                                                                               
Concretely, Gmail's SMTP relay service (smtp-relay.gmail.com) can be configured with "Only accept mail from the specified IP addresses" plus sender-hostname validation. With those settings, Infisical fails to send mail when running in a container because the EHLO hostname is not a valid FQDN owned by the sender.
                                                                                                                                                                                                                               
There is no workaround today short of patching the source.

### To Reproduce
1. Deploy Infisical in a container (e.g. Cloud Run).                                                                                                                                                                       
2. Configure SMTP to use Gmail SMTP relay (smtp-relay.gmail.com:587) with sender-hostname validation enabled in the Google Workspace admin console.
3. Trigger any email-sending action (org invite, password reset, etc.).                                                                                                                                                      
4. Mail send fails; SMTP relay rejects the session because the EHLO hostname is not an authorized FQDN.

### Expected behavior
Provide an environment variable - e.g. SMTP_HELO_HOST (or SMTP_LOCAL_HOSTNAME / SMTP_NAME) - that is forwarded to nodemailer's name option in formatSmtpConfig(). When set, the SMTP client should announce that hostname in EHLO/HELO instead of os.hostname().

quarckster:
Nodemailer falls back to os.hostname() when no `name` option is provided. Inside containers (Cloud Run, Kubernetes, ECS) that hostname is typically a random container ID, which is rejected by SMTP relays that validate the sender hostname, e.g. Gmail SMTP relay with sender-hostname checks enabled.

Expose SMTP_HELO_HOST so operators can announce a valid FQDN in the EHLO/HELO greeting. The variable is optional; existing deployments keep the previous os.hostname() behavior.

Fixes https://github.com/Infisical/infisical/issues/6228

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/lib/config/env.ts
+++ b/backend/src/lib/config/env.ts
@@ -202,6 +202,14 @@ const envSchema = z
     SMTP_PASSWORD: zpStr(z.string().optional()),
     SMTP_FROM_ADDRESS: zpStr(z.string().optional()),
     SMTP_FROM_NAME: zpStr(z.string().optional().default("Infisical")),
+    SMTP_HELO_HOST: zpStr(
+      z
+        .string()
+        .optional()
+        .describe(
+          "Hostname announced in the SMTP EHLO/HELO greeting. Defaults to the OS hostname, which may not be a valid FQDN inside containers."
+        )
+    ),
     SMTP_CUSTOM_CA_CERT: zpStr(
       z.string().optional().describe("Base64 encoded custom CA certificate PEM(s) for the SMTP server")
     ),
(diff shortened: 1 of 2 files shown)


akhilmhdh (approved):

akhilmhdh:
Thank you for the PR @quarckster

PR #6234, opened 2026-04-29
improvement(eng-4942): redis-first identity access token redesign

PrestigePvP:
## Context

- Identity access tokens were previously stored as full PG rows at issuance for every auth method. Over time this caused unbounded table growth — every login wrote a row, and cleanup only ran daily via a soft-delete scan.
- The new design makes Redis the source of truth for token validity on the auth hot path. PG is only written to on revocation (not issuance), and only for the methods that need it.
- A new `identity_access_token_revocations` table tracks revocations. It stores a row per revoked token or a sentinel row (`id = identityId`) for identity-wide revokes. Expired rows are pruned by the existing daily cleanup job.
- JWT payloads now carry all claims needed to renew a token without a PG read (`authMethod`, `accessTokenTTL`, `accessTokenMaxTTL`, `accessTokenPeriod`, `creationEpoch`, `orgName`). Legacy tokens without these claims fall back to PG on renew.
- Token Auth (the only method with an admin-managed token list in the UI) still inserts a PG row at issuance so the list UI continues to work. All other auth methods are lazy — they only insert on revoke.
- `creationEpoch` is a new JWT claim that anchors the `tokenMaxTTL` budget across renewals. `iat` gets re-stamped on every renewal so it can't serve this purpose.
- `MAX_MACHINE_IDENTITY_TOKEN_AGE` (new env var) sets a server-side ceiling on any single JWT's expiry, independent of the per-token TTL config.
- Boot hydration (`hydrateRedisFromPg`) replays the active revocations table into Redis on startup. A NX-lease prevents duplicate hydration across replicas.
- Periodic tokens (period > 0) previously used a 30-year `accessTokenMaxTTL` workaround. That's replaced by `resolveTtlInputs` which maps period mode to `maxTTL = 0` (unlimited renewals) cleanly.

## Verify

- [ ] UA login returns a token that can be used, renewed, and revoked
- [ ] Revoking a token via the API immediately rejects subsequent uses (Redis check)
- [ ] Revoking all tokens for an identity (e.g. on identity delete or auth method detach) blocks existing JWTs
- [ ] Token Auth admin revoke (by token ID, no JWT) still works
- [ ] Periodic tokens renew indefinitely without hitting a budget cap
- [ ] Usage-limited tokens reject after the limit is reached and survive a Redis flush (re-seeds from JWT claim)
- [ ] Legacy tokens (missing new claims, e.g. no `jti`) fall back to PG on renew and receive an upgraded JWT


PrestigePvP:
@claude can you re-review?

backend/src/services/identity-access-token/identity-access-token-fns.ts
```diff
@@ -0,0 +1,177 @@
...
+// run without a PG lookup. False = legacy token, requires a DAL fallback.
+export const hasFullRenewClaims = (decoded: TMinimalRenewClaims): decoded is TRenewableClaims => {
+  return (
+    !!decoded.authMethod &&
+    typeof decoded.accessTokenTTL === "number" &&
+    typeof decoded.accessTokenMaxTTL === "number" &&
+    typeof decoded.accessTokenPeriod === "number" &&
+    typeof decoded.creationEpoch === "number"
+  );
+};
```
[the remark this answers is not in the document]
PrestigePvP: Good catch fixed.

backend/src/services/identity-access-token/identity-access-token-service.ts
```diff
@@ -1,243 +1,554 @@
...
+        );
+      } else {
+        const remaining = await keyStore.hashIncrementBy(
+          stateKey,
+          IdentityTokenStateFields.UsesRemaining(token.jti),
+          -1
+        );
+        if (remaining < 0) {
+          throw new UnauthorizedError({ message: "Failed to authorize: token usage limit reached" });
+        }
```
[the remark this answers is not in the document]
PrestigePvP: Good catch, fixed.

backend/src/services/identity-access-token/identity-access-token-revocation-dal.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const identityAccessTokenRevocationDALFactory = (db: TDbClient) => {
+  const insertRevocation = async (row: TInsertRevocationInput) => {
+    try {
+      await db(TableName.IdentityAccessTokenRevocation).insert(row).onConflict(["id"]).merge({ updatedAt: db.fn.now() });
+    } catch (error) {
+      throw new DatabaseError({ error, name: "IdentityAccessTokenRevocationInsert" });
+    }
+  };
```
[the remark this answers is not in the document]
PrestigePvP: Good catch, fixed. 

Shouldn't ever get in this path but fixed anyway.

backend/src/services/identity-access-token/identity-access-token-service.ts
```diff
@@ -1,243 +1,554 @@
...
+    return {
+      identityId: token.identityId,
+      identityName: token.identityName ?? "",
+      name: token.identityName ?? "",
+      orgId: token.orgId,
+      rootOrgId: token.rootOrgId,
+      parentOrgId: token.parentOrgId,
+      orgName: undefined as string | undefined,
+      authMethod: (token.authMethod ?? "") as string
+    };
```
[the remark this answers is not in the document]
PrestigePvP: We removed the orgName from OTEL metrics for the sake of simplicity and to avoid giving out any more information than needed in the token itself.

backend/src/db/migrations/20260428145237_identity-access-token-revocations.ts
```diff
@@ -0,0 +1,30 @@
...
+    // null for per-token revocations. Stored so hydration can replay the
+    // exact revocation timestamp rather than falling back to createdAt
+    // (which reflects BullMQ job execution time, not the revocation call).
+    t.timestamp("revokedAt", { useTz: true }).nullable();
+    t.timestamps(true, true, true);
+    t.index("expiresAt");
+  });
+
+  await createOnUpdateTrigger(knex, TableName.IdentityAccessTokenRevocation);
+}
```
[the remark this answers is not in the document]
PrestigePvP: PR comment was updated. I planned to do the partition but the code was complex and not worth it. 

The optimal solution ended in tokens sitting longer than if we just do the typical delete loop so I swapped the design but forgot to update the PR description.

backend/src/@types/winrm-client.d.ts
```diff
@@ -0,0 +1,24 @@
+declare module "winrm-client" {
```
akhilmhdh: Is this related to this PR?

backend/src/db/migrations/20260428145237_identity-access-token-revocations.ts
```diff
@@ -0,0 +1,30 @@
...
+    // Populated for identity-wide revoke-all markers (id == identityId);
+    // null for per-token revocations. Stored so hydration can replay the
+    // exact revocation timestamp rather than falling back to createdAt
+    // (which reflects BullMQ job execution time, not the revocation call).
+    t.timestamp("revokedAt", { useTz: true }).nullable();
+    t.timestamps(true, true, true);
+    t.index("expiresAt");
+  });
+
+  await createOnUpdateTrigger(knex, TableName.IdentityAccessTokenRevocation);
```
akhilmhdh: Do we need an update trigger for this table. It's kinda like once used black list table right?

backend/src/lib/config/env.ts
```diff
@@ -238,6 +230,9 @@ const envSchema = z
     JWT_INVITE_LIFETIME: zpStr(z.string().default("1d")),
     JWT_MFA_LIFETIME: zpStr(z.string().default("5m")),
     JWT_PROVIDER_AUTH_LIFETIME: zpStr(z.string().default("15m")),
+    // Server-enforced ceiling on identity access token lifetime (in seconds). Caps `exp`
+    // at issuance and renewal so machine-identity JWTs cannot be immortal.
+    MAX_MACHINE_IDENTITY_TOKEN_AGE: zpStr(z.coerce.number().int().positive().default(7_776_000)),
```
akhilmhdh: Can this like our token life time above that accepts `ms` string?

backend/src/services/identity-access-token/identity-access-token-queue.ts
```diff
@@ -0,0 +1,99 @@
...
+        }
+      );
+    }
+  };
+
+  queueService.start(QueueName.IdentityAccessTokenPgMirror, async (job) => {
+    const data = job.data as TPgMirrorPayload;
+
+    switch (data.kind) {
+      case "revoke-token": {
```
akhilmhdh: It seems to be a single DAL insert? Any reason to make this queue process?

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityKubernetesAuthForm.tsx
```diff
@@ -147,6 +148,9 @@ export const IdentityKubernetesAuthForm = ({
   const { hasOrgRole } = useOrgPermission();
   const isOrgAdmin = hasOrgRole(OrgMembershipRole.Admin);
 
+  const schema = useMemo(() => buildSchema(maxAccessTokenTTL), [maxAccessTokenTTL]);
```
akhilmhdh: This can be one memo by constructing the schema inside with returning resolver

akhilmhdh: Like wise all memo in auth ui

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityJwtAuthForm.tsx
```diff
@@ -122,6 +124,9 @@ export const IdentityJwtAuthForm = ({
     enabled: isUpdate
   });
 
+  const schema = useMemo(() => buildSchema(maxAccessTokenTTL), [maxAccessTokenTTL]);
```
akhilmhdh: Likewise here - one memo with returning resolver

akhilmhdh (changes requested):
Overall two changes

1. Memo simplification
2. I feel like the queue is an overkill for a very simple revocation lookup to postgres. 

Pending application testing

backend/src/@types/winrm-client.d.ts
PrestigePvP: Removed

akhilmhdh (dismissed):

akhilmhdh (approved):

PR #6239, opened 2026-04-29
feat: allow cancellation of pending certificate requests and improve pending issuance status visibility

carlosmonastyrski:
## Context

Adds visibility and control over pending certificate requests in the Cert Manager. The cert request row now shows what stage an in-flight issuance is actually at, queued, submitting to the CA, performing the DNS-01 challenge, waiting for DNS propagation, awaiting CA validation, finalizing, surfaced as a tooltip on the Pending badge, with stage-aware messages for ACME, DigiCert, AWS PCA, AWS ACM Public CA, Azure AD CS, and Venafi TPP. Adds a Cancel Request action that immediately frees the worker (using BullMQ's AbortSignal to short-circuit the running job) and marks the request as `failed` with `errorMessage = "Cancelled by "actor"` so external clients keep seeing a terminal state.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -5407,6 +5408,14 @@ interface TriggerCertificateRequestValidationEvent {
   };
 }
 
+interface CancelCertificateRequestEvent {
+  type: EventType.CANCEL_CERTIFICATE_REQUEST;
+  metadata: {
+    certificateRequestId: string;
+    cancelled: boolean;
```
saifsmailbox98, line 6192: When can this be false?

saifsmailbox98, line 6192: Shall we also log the status and the pending message so that somebody could check how far a request had gone if somebody else cancels the request?

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -450,6 +450,7 @@ export enum EventType {
   GET_CERTIFICATE_FROM_REQUEST = "get-certificate-from-request",
   LIST_CERTIFICATE_REQUESTS = "list-certificate-requests",
   TRIGGER_CERTIFICATE_REQUEST_VALIDATION = "trigger-certificate-request-validation",
+  CANCEL_CERTIFICATE_REQUEST = "cancel-certificate-request",
```
saifsmailbox98, line 497: Update enums.tsx and constants.tsx as well.

backend/src/services/certificate-request/certificate-request-dal.ts
```diff
@@ -89,28 +89,67 @@ export const certificateRequestDALFactory = (db: TDbClient) => {
...
     try {
-      const updateData: Partial<TCertificateRequests> = { status };
+      const updateData: Partial<TCertificateRequests> = { status, pendingMessage: null };
       if (errorMessage !== undefined) {
         updateData.errorMessage = errorMessage;
       }
-      return await certificateRequestOrm.updateById(id, updateData, tx);
+      const [updated] = await (tx || db)(TableName.CertificateRequests)
+        .where({ id })
+        .whereIn("status", [CertificateRequestStatus.PENDING, CertificateRequestStatus.PENDING_VALIDATION])
```
saifsmailbox98, line 111: is updateStatus already being used elsewhere? where this change may make it skip the update?


also since we have this condition added, we shouldn't keep it named as `updateStatus`

backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -739,13 +743,67 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
...
+        });
+
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        projectId,
+        event: {
+          type: EventType.CANCEL_CERTIFICATE_REQUEST,
+          metadata: {
+            certificateRequestId: req.params.requestId,
+            cancelled
```
saifsmailbox98: Do we need to create an audit log if the request was not cancelled? Is it a log that the user tried this?

saifsmailbox98 (changes requested):

backend/src/server/routes/v1/certificate-router.ts
carlosmonastyrski: Exactly, it's a log that will show the intent of cancelling, in case it's not possible to cancel it, I think it's worth to have a log so after this they can track why in the first place the request could not be cancelled

backend/src/ee/services/audit-log/audit-log-types.ts
carlosmonastyrski: Yes, it can happen that I trigger the cancel after the certificate request is finished, so on those cases it's not possible to cancel it, and the result would be false there

carlosmonastyrski:
@claude review once

backend/src/server/routes/v1/certificate-router.ts
saifsmailbox98: I'm not really sure if this is the right way of doing it. If an action is not possible due to the current state, then we should just give them an immediate feedback rather than modifying something within the system - in this case, the addition of the audit log 

If it couldn't be cancelled, then we should just tell them that it couldn't be cancelled. The cancellation is not an asynchronous request. 

Check with Sheen once, as this appears to be a business decision regarding whether we should display the cancellation attempts

backend/src/db/migrations/20260429021227_add-pending-message-to-certificate-requests.ts
```diff
@@ -0,0 +1,25 @@
+import { Knex } from "knex";
+
```
saifsmailbox98, line 2: IGNORE COMMENT PLACEMENT

Are we allowing operators to cancel requests? I hope we are making this decision deliberately, since it's a destructive action.

carlosmonastyrski: Yes, it's intended

backend/src/server/routes/v1/certificate-router.ts
carlosmonastyrski: Yup, it's intended as it's similar to auth audit logs, consider it an audit of intention, so we track all events of this type, even when something happened so users can truly know what happened on these events and have a clear trace of events

saifsmailbox98 (approved):

PR #6243, opened 2026-04-29
feat: honey tokens

mathnogueira:
## Context

This adds the honey token feature to Infisical, allowing users to create traps that once activated, notify them so they can rotate their secrets fast in case of credential leaks.

## Steps to verify the change

1. Use ngrok to get a public endpoint for your localhost
2. When applying the CF command, use that endpoint instead of `http://localhost:3000`
3. Use AWS CLI to trigger the AWS account usage (I'll append my testing script)

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


mathnogueira:
DB schema failure is not related to our changes.

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,409 @@
...
+      const generatedKey = `htk_live_${crypto.randomBytes(24).toString("hex")}`;
+      return {
+        id: null,
+        orgId: orgPermission.orgId,
+        type,
+        connectionId: null,
+        createdAt: null,
+        updatedAt: null,
+        decryptedConfig: { webhookSigningKey: generatedKey, stackName: "infisical-honey-tokens" }
+      };
```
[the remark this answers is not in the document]
mathnogueira: We updated the flow. This doesn't happen anymore. We only show the command to deploy the CF stack after this value is saved.

backend/src/db/migrations/20260427174746_honey-tokens.ts
```diff
@@ -0,0 +1,84 @@
...
+      t.uuid("honeyTokenId").notNullable();
+      t.foreign("honeyTokenId").references("id").inTable(TableName.HoneyToken).onDelete("CASCADE");
+      t.string("eventType").notNullable();
+      t.jsonb("metadata");
+      t.timestamps(true, true, true);
+    });
+
+    await createOnUpdateTrigger(knex, TableName.HoneyTokenEvent);
+    await knex.schema.alterTable(TableName.HoneyTokenEvent, (t) => {
+      t.index(["honeyTokenId", "createdAt"]);
```
varonix0: q: we don't want to index just the honeyTokenId alone? this is a composite index that will only be used when querying for both the createdAt and honeyTokenId at the same time

backend/src/db/migrations/20260427174746_honey-tokens.ts
```diff
@@ -0,0 +1,84 @@
...
+      t.foreign("createdByUserId").references("id").inTable(TableName.Users).onDelete("SET NULL");
+      t.uuid("resetByUserId").nullable();
+      t.foreign("resetByUserId").references("id").inTable(TableName.Users).onDelete("SET NULL");
+      t.uuid("revokedByUserId").nullable();
+      t.foreign("revokedByUserId").references("id").inTable(TableName.Users).onDelete("SET NULL");
+      t.timestamps(true, true, true);
+    });
+
+    await createOnUpdateTrigger(knex, TableName.HoneyToken);
+    await knex.schema.alterTable(TableName.HoneyToken, (t) => {
```
varonix0: question: why do we alter? why don't we do it inside the create table statement? applies to most parts of this migration file

backend/src/@types/fastify.d.ts
```diff
@@ -369,6 +371,8 @@ declare module "fastify" {
       gatewayV2: TGatewayV2ServiceFactory;
       gatewayPool: TGatewayPoolServiceFactory;
       githubOrgSync: TGithubOrgSyncServiceFactory;
+      honeyToken: THoneyTokenConfigServiceFactory;
+      honeyTokenCrud: THoneyTokenServiceFactory;
```
varonix0: I'd rename `honeyToken` to `honeyTokenConfig` and `honeyTokenCrud` to just `honeyToken

backend/src/db/schemas/pam-resources.ts
```diff
@@ -21,8 +21,8 @@ export const PamResourcesSchema = z.object({
   encryptedRotationAccountCredentials: zodBuffer.nullable().optional(),
   encryptedResourceMetadata: zodBuffer.nullable().optional(),
   discoveryFingerprint: z.string().nullable().optional(),
-  domainId: z.string().uuid().nullable().optional(),
```
varonix0, line 24: Why was this changed?

backend/src/ee/routes/v1/honey-token-router.ts
```diff
@@ -0,0 +1,445 @@
...
+        type: z.nativeEnum(HoneyTokenType),
+        name: slugSchema({ field: "Name" }),
+        description: z.string().trim().max(256).nullish(),
+        secretsMapping: z.record(z.string(), z.string().min(1)),
+        environment: z.string().trim(),
+        secretPath: z.string().trim().min(1)
+      }),
+      response: {
+        200: z.object({
+          honeyToken: z.object({
```
varonix0: Can we just use the honey token zod schem and pick the fields we want to return? or if we're returning everything, we can just use the zod schema directly

backend/src/ee/routes/v1/honey-token-router.ts
```diff
@@ -0,0 +1,445 @@
...
+      }
+    },
+    handler: async (req) => {
+      const { honeyToken, stackDeployment } = await server.services.honeyTokenCrud.create(req.body, req.permission);
+      return { honeyToken, stackDeployment };
+    }
+  });
+
+  server.route({
+    url: "/:honeyTokenId",
```
varonix0: nit: instead of `honeyTokenId` as the parameter, I'd just do `id`, since it already lives on the honey token router itself

backend/src/ee/routes/v1/honey-token-router.ts
```diff
@@ -0,0 +1,445 @@
...
+          projectId: req.query.projectId
+        },
+        req.permission
+      );
+      return { honeyToken };
+    }
+  });
+
+  server.route({
+    url: "/:honeyTokenId",
```
varonix0: Same applies to the other routers

backend/src/ee/routes/v1/honey-token-router.ts
```diff
@@ -0,0 +1,445 @@
...
+      rateLimit: writeLimit
+    },
+    onRequest: verifyAuth([AuthMode.JWT]),
+    schema: {
+      params: z.object({
+        honeyTokenId: z.string().uuid()
+      }),
+      body: z.object({
+        projectId: z.string().trim(),
+        name: slugSchema({ field: "Name" }).optional(),
```
varonix0: `Name` should be lowercase. same applies to all the oher places in the router where we use this slug schema

backend/src/ee/routes/v1/honey-token-router.ts
```diff
@@ -0,0 +1,445 @@
...
+        description: z.string().trim().max(256).nullish(),
+        secretsMapping: z.record(z.string(), z.string().min(1)).optional()
+      }),
+      response: {
+        200: z.object({
+          honeyToken: z.object({
+            id: z.string().uuid(),
+            name: z.string(),
```
varonix0: And same as mentioned earlier, lets use the zod schemas directly - this comment applies to all routes

backend/src/ee/routes/v1/honey-token-router.ts
```diff
@@ -0,0 +1,445 @@
...
+    schema: {
+      params: z.object({
+        honeyTokenId: z.string().uuid()
+      }),
+      querystring: z.object({
+        projectId: z.string().trim()
+      }),
+      response: {
+        200: z.object({
+          credentials: z.record(z.string(), z.string())
```
varonix0: To make this more API friendly in the future, we should make this a discriminated union so users can always expect a certain structure of response

backend/src/ee/routes/v1/honey-token-router.ts
```diff
@@ -0,0 +1,445 @@
+import { z } from "zod";
```
varonix0, line 1: General question on the API / service design; why did we choose this pattern over something like secret rotations where everything is designated on a per-provider basis?

Example: https://github.com/Infisical/infisical/blob/main/backend/src/ee/routes/v2/secret-rotation-v2-routers/aws-iam-user-secret-rotation-router.ts

I think this might be the correct pattern to take since we will be expanding this to other providers in the future. Curious on your thoughts too as this would be quite different from what we have right now

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+    stackName: string;
+  }): Promise<{ deployed: boolean; status: string | null }> => {
+    try {
+      const appConnection = await appConnectionDAL.findById(connId);
+      if (!appConnection) {
+        return { deployed: false, status: null };
+      }
+
+      const decryptedConnection = await decryptAppConnection(appConnection, kmsService);
+      const awsConfig = decryptedConnection as unknown as TAwsConnectionConfig;
```
varonix0: I would use zod safeParse here and thorw an error if it fails to parse, that way we avoid type casting

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+      const { credentials: awsCredentials } = await getAwsConnectionConfig(awsConfig);
+
+      const cfn = new CloudFormationClient({ credentials: awsCredentials, region: "us-east-1" });
```
varonix0: So we are hardcoding the region to us-east-1? What if the user wants to provision their infra in eu-west-1 as an example?

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+  honeyTokenEventDAL,
+  permissionService,
+  kmsService,
+  licenseService,
+  appConnectionDAL,
+  orgDAL,
+  projectDAL,
+  smtpService
+}: THoneyTokenConfigServiceFactoryDep) => {
+  const verifyStackDeployment = async ({
```
varonix0: This function is missing a permissions check. Anyone can verify the stack deployment of any stack in any org

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+    });
+
+    if (existing) {
+      return honeyTokenConfigDAL.updateById(existing.id, {
+        connectionId,
+        encryptedConfig
+      });
+    }
+
+    return honeyTokenConfigDAL.create({
```
varonix0: nit: if this throws an error we lose parts of the stack trace because we aren't awaiting the function directly because the error doesn't surface correctly. applies to everywhere we return without await

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
+import crypto from "node:crypto";
```
varonix0: Never use `node:crypto` please, all cryptography needs to come from `import { crypto } from "@app/lib/crypto/cryptography"` due to FIPS regulations

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+      decryptedConfig = AwsHoneyTokenConfigSchema.parse(JSON.parse(decrypted.toString()) as unknown);
+    }
+
+    return {
+      ...config,
+      decryptedConfig
+    };
+  };
+
+  const sendTriggerNotification = async ({
```
varonix0: question: shouldn't this be in the honey token service rather than the config service?

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+      logger.error(
+        { err, orgId, honeyTokenId: honeyToken.id },
+        `Failed to send honey token trigger notification [orgId=${orgId}] [honeyTokenId=${honeyToken.id}]`
+      );
+    }
+  };
+
+  const SIGNATURE_TOLERANCE_MS = 5 * 60 * 1000;
+
+  const handleTrigger = async ({
```
varonix0: same as above

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+    const receivedBuf = Buffer.from(signatureHash, "hex");
+
+    if (expectedBuf.byteLength !== receivedBuf.byteLength || !crypto.timingSafeEqual(expectedBuf, receivedBuf)) {
+      throw new UnauthorizedError({ message: "Invalid webhook signature" });
+    }
+
+    const rawEvents = Array.isArray(payload) ? (payload as unknown[]) : [payload];
+
+    /* eslint-disable no-await-in-loop, no-continue */
+    for (const rawEvent of rawEvents) {
```
varonix0: You can do `for await (...)` and remove the eslint disable for `no-await-in-loop`

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+        );
+
+        void sendTriggerNotification({
+          orgId,
+          honeyToken,
+          eventMetadata: parsed.data
+        });
+      }
+    }
+    /* eslint-enable no-await-in-loop, no-continue */
```
varonix0: Can remove `no-await-in-loop` here as well

backend/src/ee/services/honey-token/honey-token-dal.ts
```diff
@@ -0,0 +1,111 @@
...
+export const honeyTokenDALFactory = (db: TDbClient) => {
+  const orm = ormify(db, TableName.HoneyToken);
+
+  const findByFolderIds = async (folderIds: string[], tx?: Knex) => {
+    const rows = await (tx || db.replicaNode())(TableName.HoneyToken)
+      .whereIn(`${TableName.HoneyToken}.folderId`, folderIds)
+      .join(TableName.SecretFolder, `${TableName.SecretFolder}.id`, `${TableName.HoneyToken}.folderId`)
+      .join(TableName.Environment, `${TableName.Environment}.id`, `${TableName.SecretFolder}.envId`)
+      .select(
+        db.ref("id").withSchema(TableName.HoneyToken),
```
varonix0, line 18: nit: I would group them by select statements for readability

```ts
 .select(
        db.ref("id").withSchema(TableName.HoneyToken),
        db.ref("description").withSchema(TableName.HoneyToken),
        db.ref("name").withSchema(TableName.HoneyToken),
        db.ref("type").withSchema(TableName.HoneyToken),
        db.ref("status").withSchema(TableName.HoneyToken),
        db.ref("projectId").withSchema(TableName.HoneyToken),
        db.ref("folderId").withSchema(TableName.HoneyToken),
        db.ref("connectionId").withSchema(TableName.HoneyToken),
        db.ref("secretsMapping").withSchema(TableName.HoneyToken),
        db.ref("createdAt").withSchema(TableName.HoneyToken),
        db.ref("updatedAt").withSchema(TableName.HoneyToken)
      )
      .select(
        db.ref("id").withSchema(TableName.Environment).as("envId"),
        db.ref("name").withSchema(TableName.Environment).as("envName"),
        db.ref("slug").withSchema(TableName.Environment).as("envSlug"),
        db.ref("name").withSchema(TableName.SecretFolder).as("folderName")
      );
```

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+  secretDAL,
+  secretVersionDAL,
+  secretVersionTagDAL,
+  secretTagDAL,
+  folderCommitService,
+  resourceMetadataDAL,
+  snapshotService,
+  secretQueueService
+}: THoneyTokenServiceFactoryDep) => {
+  const verifyStackDeployment = async ({
```
varonix0: Same here, no permision check

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+    const { permission } = await permissionService.getProjectPermission({
+      actor: actor.type,
+      actorId: actor.id,
+      actorAuthMethod: actor.authMethod,
+      actorOrgId: actor.orgId,
+      actionProjectType: ActionProjectType.SecretManager,
+      projectId
+    });
+
+    ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionSecretActions.Create, ProjectPermissionSub.Secrets);
```
varonix0: I think honey tokens need their own permissions like we have permissions for dynamic secrets and secret rotations. The secrets subject is specifically for static secrets

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+        orgId: actor.orgId,
+        secretPath: folderInfo.path,
+        projectId,
+        environmentSlug: folderInfo.environmentSlug,
+        excludeReplication: true
+      });
+    }
+
+    return {
+      honeyToken: {
```
varonix0: Can't we just return the whole honey token and let fastify filter it out on the router level via the zod response schema? that's more in line with what we do other places. applies to most functions in this file

backend/src/ee/services/honey-token/honey-token-config-service.ts
```diff
@@ -0,0 +1,449 @@
...
+    });
+
+    return {
+      isConnected: stackDeployment.deployed,
+      status: stackDeployment.status,
+      stackName: storedConfig.stackName
+    };
+  };
+
+  const getConfig = async ({ orgPermission, type }: { orgPermission: OrgServiceActor; type: HoneyTokenType }) => {
```
varonix0: If user has read access on the org settings they can read the webhook signing secret. seems pretty sensitive, not sure if this is the right permission here

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+    const { encryptor: credentialEncryptor } = await kmsService.createCipherPairWithDataKey({
+      type: KmsDataKey.Organization,
+      orgId: actor.orgId
+    });
+
+    const encryptedCredentials = credentialEncryptor({
+      plainText: Buffer.from(JSON.stringify(honeyTokenCredentials))
+    }).cipherTextBlob;
+
+    const honeyToken = await honeyTokenDAL.create({
```
varonix0: Shouldn't everything be inside a transaction here? what if the honey token gets created but secret creation fails as an example

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+    await snapshotService.performSnapshot(folder.id);
+    await secretQueueService.syncSecrets({
+      orgId: actor.orgId,
+      secretPath,
+      projectId,
+      environmentSlug: environment,
+      excludeReplication: true
+    });
+
+    const stackDeployment = await verifyStackDeployment({
```
varonix0: Question why do we do this at the end? Shouldn't this be one of the first things we do and then throw an error if the stack isn't successfully deployed?

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+      actorOrgId: actor.orgId,
+      actionProjectType: ActionProjectType.SecretManager,
+      projectId
+    });
+
+    ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionSecretActions.Edit, ProjectPermissionSub.Secrets);
+
+    const honeyToken = await honeyTokenDAL.findById(honeyTokenId);
+
+    if (!honeyToken || honeyToken.projectId !== projectId) {
```
varonix0: Why does this function accept a project ID? Can't we just fetch the honey token and then in the permission check do:


```ts
 const { permission } = await permissionService.getProjectPermission({
      actor: actor.type,
      actorId: actor.id,
      actorAuthMethod: actor.authMethod,
      actorOrgId: actor.orgId,
      actionProjectType: ActionProjectType.SecretManager,
      projectId: honeyToken.projectId # this here
    });
```

Then we can also get rid of the `honeyToken.projectId !== projectId` part

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+        if (conflictingSecrets.length) {
+          throw new BadRequestError({
+            message: `The following secrets already exist: ${conflictingSecrets.map(({ key }) => key).join(", ")}`
+          });
+        }
+      }
+
+      const oldSecretKeys = Object.values(oldMapping);
+
+      await fnSecretBulkDelete({
```
varonix0: Everything should be in a transaction. If someone fails we end up with corrupted data

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+        projectId: updated.projectId,
+        secretsMapping: updated.secretsMapping,
+        createdAt: updated.createdAt,
+        updatedAt: updated.updatedAt
+      }
+    };
+  };
+
+  const revokeHoneyToken = async (
+    { honeyTokenId, projectId }: { honeyTokenId: string; projectId: string },
```
varonix0: General note to all of these routes for mutating existing data: We do not need to accept the project ID. We can infer the project ID from the honey token itself and use that for things like permission checks

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+
+      try {
+        await iam.send(
+          new DeleteAccessKeyCommand({
+            UserName: decryptedCredentials.iamUserName,
+            AccessKeyId: decryptedCredentials.accessKeyId
+          })
+        );
+      } catch {
+        // Access key may already be deleted
```
varonix0: I think we should log it in that case, logger.info

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+      }
+
+      try {
+        await iam.send(new DeleteUserCommand({ UserName: decryptedCredentials.iamUserName }));
+      } catch {
+        // IAM user may already be deleted
```
varonix0: Same here

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+      }
+    }
+
+    const secretKeys = Object.values(honeyToken.secretsMapping as Record<string, string>);
+
+    await fnSecretBulkDelete({
```
varonix0: Transaction

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+        environmentSlug: folderInfo.environmentSlug,
+        excludeReplication: true
+      });
+    }
+
+    return { honeyTokenId };
+  };
+
+  const resetHoneyToken = async (
+    { honeyTokenId, projectId }: { honeyTokenId: string; projectId: string },
```
varonix0: No need for project ID, infer from the honey token itself

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+      search
+    }: {
+      projectId: string;
+      environments: string[];
+      secretPath: string;
+      search?: string;
+    },
+    actor: OrgServiceActor
+  ) => {
+    await permissionService.getProjectPermission({
```
varonix0: General note as well: We are lacking permission checks for a lot of these functions. We should have specific permission sets for honey tokens

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+    const credentials: Record<string, string> = {};
+    for (const [credentialKey, secretName] of Object.entries(mapping)) {
+      credentials[secretName] = decryptedCredentials[credentialKey] ?? "";
+    }
+
+    return { credentials };
+  };
+
+  const getHoneyTokenById = async (
+    { honeyTokenId, projectId }: { honeyTokenId: string; projectId: string },
```
varonix0: Just reiterating; no need for project ID

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,900 @@
...
+    };
+  };
+
+  const getHoneyTokenEvents = async (
+    {
+      honeyTokenId,
+      projectId,
+      offset,
+      limit
+    }: { honeyTokenId: string; projectId: string; offset?: number; limit?: number },
```
varonix0: This should be a DTO in honey-token-types.ts. Applies to most functions here and in the config service

backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -107,12 +107,37 @@ export const registerDashboardRouter = async (server: FastifyZodProvider) => {
         includeFolders: booleanSchema.describe(DASHBOARD.SECRET_OVERVIEW_LIST.includeFolders),
         includeImports: booleanSchema.describe(DASHBOARD.SECRET_OVERVIEW_LIST.includeImports),
         includeSecretRotations: booleanSchema.describe(DASHBOARD.SECRET_OVERVIEW_LIST.includeSecretRotations),
-        includeDynamicSecrets: booleanSchema.describe(DASHBOARD.SECRET_OVERVIEW_LIST.includeDynamicSecrets)
+        includeDynamicSecrets: booleanSchema.describe(DASHBOARD.SECRET_OVERVIEW_LIST.includeDynamicSecrets),
+        includeHoneyTokens: booleanSchema.optional()
```
varonix0: Missing .describe(), and why'd we make it optional?

backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -107,12 +107,37 @@ export const registerDashboardRouter = async (server: FastifyZodProvider) => {
...
       }),
       response: {
         200: z.object({
           folders: SecretFoldersSchema.extend({ environment: z.string() }).array().optional(),
           dynamicSecrets: SanitizedDynamicSecretSchema.extend({ environment: z.string() }).array().optional(),
+          honeyTokens: z
+            .object({
+              id: z.string().uuid(),
```
varonix0: Should use the honey token schema or a sanitized schema as we do elsewhere

backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -658,7 +725,8 @@ export const registerDashboardRouter = async (server: FastifyZodProvider) => {
         includeFolders: booleanSchema.describe(DASHBOARD.SECRET_DETAILS_LIST.includeFolders),
         includeDynamicSecrets: booleanSchema.describe(DASHBOARD.SECRET_DETAILS_LIST.includeDynamicSecrets),
         includeImports: booleanSchema.describe(DASHBOARD.SECRET_DETAILS_LIST.includeImports),
-        includeSecretRotations: booleanSchema.describe(DASHBOARD.SECRET_DETAILS_LIST.includeSecretRotations)
+        includeSecretRotations: booleanSchema.describe(DASHBOARD.SECRET_DETAILS_LIST.includeSecretRotations),
+        includeHoneyTokens: booleanSchema.optional()
```
varonix0: Same as above

backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -670,6 +738,30 @@ export const registerDashboardRouter = async (server: FastifyZodProvider) => {
...
           dynamicSecrets: SanitizedDynamicSecretSchema.array().optional(),
+          honeyTokens: z
+            .object({
+              id: z.string().uuid(),
+              name: z.string(),
+              description: z.string().nullable().optional(),
+              type: z.string(),
+              status: z.string(),
+              projectId: z.string(),
+              folderId: z.string().uuid(),
```
varonix0: Same as above, and generally for any of the new routes included in this pr

backend/src/services/smtp/emails/HoneyTokenTriggeredTemplate.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <Text className="text-[14px] mt-[4px]">{eventName}</Text>
+        <strong className="text-[14px]">Time</strong>
+        <Text className="text-[14px] mt-[4px]">{eventTime}</Text>
+        <strong className="text-[14px]">Source IP</strong>
+        <Text className="text-[14px] mt-[4px]">{sourceIp}</Text>
+        <strong className="text-[14px]">AWS Region</strong>
+        <Text className="text-[14px] mt-[4px]">{awsRegion}</Text>
+      </Section>
+      <Section className="text-center mt-[24px]">
+        <BaseButton href={projectUrl}>Investigate Now</BaseButton>
```
varonix0: I feel like when they press this, they shouldn't just be sent to the project. What if they have nested folders and multiple honey tokens? gonna be tough finding the right one. we should send them directly to the details page i think

docs/documentation/platform/honey-tokens/setup.mdx
```diff
@@ -0,0 +1,88 @@
...
+  naming convention Infisical uses for honey token IAM users.
+</Note>
+
+## Configure the Organization
+
+<Steps>
+  <Step title="Navigate to Organization Settings">
+    Go to **Organization Settings** → **Product Settings** and scroll to the **Honey Tokens** section.
+
+    {/* <!-- Screenshot: Organization Settings page showing the Honey Tokens section with empty form fields --> */}
```
varonix0: Are we missing a screenshot?

docs/documentation/platform/honey-tokens/usage.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+    - **Name** — a slug-friendly identifier for this honey token (must be unique within the folder).
+    - **Description** (optional) — a note about this honey token's purpose.
+    - **Environment** — the target environment where the decoy secrets will be planted.
+    - **Secret Path** — the folder path where the decoy secrets will appear.
+    - **Secrets Mapping** — the secret key names that will hold the decoy credentials:
+      - **Access Key ID** — the name of the secret that will store the AWS access key ID (e.g., `AWS_ACCESS_KEY_ID`).
+      - **Secret Access Key** — the name of the secret that will store the AWS secret access key (e.g., `AWS_SECRET_ACCESS_KEY`).
+
+    {/* <!-- Screenshot: The honey token creation form showing all fields filled out --> */}
```
varonix0: I think we are missing a lot of screenshots

docs/images/platform/honey-tokens/1.png
varonix0, line 1: A lot of the screenshots are outdated after the UI revamp you and Scott went over, we should update these as well

varonix0 (changes requested):

backend/src/db/migrations/20260427174746_honey-tokens.ts
mathnogueira: That's actually intentional:

Most of our queries we either use `createdAt` in the query (`createdAt > ?` or `ORDER BY createdAt`).

```
WHERE honeyTokenId = ?
ORDER BY createdAt DESC
```

And because how composite indexes work in Postgres (left-prefix index) as long as we use the parameters on the left, the index is used.

So, queries that only filter by `honeyTokenId` will also use the index normally.

backend/src/db/schemas/pam-resources.ts
mathnogueira: Oh, asked claude to solve the lint issues, that was an existing one from `main`. I'll revert this one.

backend/src/ee/services/honey-token/honey-token-config-service.ts
mathnogueira: Good point. It was hard-coded due to the EventBridge (as IAM is hosted in us-east-1), but now we can allow users to deploy this into any region. I'll add a new field to the settings to allow them to override this.

backend/src/ee/routes/v1/honey-token-router.ts
mathnogueira: Reworked on the API to include providers based on the type, so we can extend it with new types.

backend/src/ee/routes/v1/honey-token-routers/aws-honey-token-router.ts
```diff
@@ -0,0 +1,21 @@
...
+    type: HoneyTokenType.AWS,
+    configSchema: AwsHoneyTokenConfigSchema,
+    decryptedConfigSchema: AwsHoneyTokenConfigSchema,
+    testConnectionResponseSchema: z
+      .object({
+        isConnected: z.boolean(),
+        status: z.string().nullable(),
+        stackName: z.string()
+      })
+      .passthrough()
```
varonix0: We are sure we want to use passthrough? that would allow everything to go through even if not defined on the schema

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+
+    const folders = await folderDAL.findBySecretPathMultiEnv(projectId, environments, secretPath);
+    if (!folders.length) return 0;
+
+    const folderIds = folders.map((f) => f.id);
+    return honeyTokenDAL.countByFolderIds(folderIds, search);
+  };
+
+  const getOrgHoneyTokenLimit = async ({ projectId }: { projectId: string }, actor: OrgServiceActor) => {
+    await assertProjectPermission({ projectId, actor, action: ProjectPermissionHoneyTokenActions.Read });
```
varonix0: Couldn't someone just pass in any project ID? Looks like we're trusting the user input here

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+    return {
+      type,
+      credentials: await providerHooks.getCredentialsForDisplay({
+        encryptedCredentials: honeyToken.encryptedCredentials,
+        orgId: actor.orgId
+      })
+    };
+  };
+
+  const sendTriggerNotification = async ({
```
varonix0: Should be prefixed with `$` to indicate its an internal method

backend/src/ee/routes/v1/honey-token-routers/honey-token-endpoints.ts
```diff
@@ -0,0 +1,161 @@
...
+    method: "GET",
+    config: {
+      rateLimit: readLimit
+    },
+    onRequest: verifyAuth([AuthMode.JWT]),
+    schema: {
+      response: {
+        200: z.object({
+          config: SanitizedHoneyTokenConfigSchema.extend({
+            id: z.string().nullable(),
```
varonix0: `id`, `connectionId`, `createdAt`, and `updatedAt` doesn't need to be extended. It's already on the sanitized schema

backend/src/ee/services/honey-token/config/honey-token-config-service.ts
```diff
@@ -0,0 +1,116 @@
...
+    orgPermission: OrgServiceActor;
+    connectionId: string;
+    config: THoneyTokenConfigByType[T];
+  }) => Promise<THoneyTokenConfigRecord>;
+  testConnection: (input: { orgPermission: OrgServiceActor }) => Promise<THoneyTokenTestConnectionResponseByType[T]>;
+  getConfig: (input: { orgPermission: OrgServiceActor }) => Promise<THoneyTokenConfigWithDecrypted>;
+};
+
+const assertSupportedHoneyTokenType = (type: string): HoneyTokenType => {
+  if (Object.prototype.hasOwnProperty.call(HONEY_TOKEN_PROVIDER_MAP, type)) {
```
varonix0: question: can't we just use `Object.hasOwn()`?

backend/src/ee/services/honey-token/aws/honey-token-aws-config-provider.ts
```diff
@@ -0,0 +1,264 @@
+import { CloudFormationClient, DescribeStacksCommand } from "@aws-sdk/client-cloudformation";
```
varonix0: All of these files related specifically to AWS honey tokens, shouldn't they live in their own directory? like we have for secret rotations: `/backend/src/ee/services/secret-rotation-v2/aws-iam-user-secret/<file>`

backend/src/ee/services/honey-token/honey-token-provider-definitions.ts
```diff
@@ -0,0 +1,45 @@
...
+  credentialsResponseSchema: z.ZodType<{
+    type: T;
+    credentials: THoneyTokenDisplayCredentialsByType[T];
+  }>;
+  serviceHooksFactory: (deps: THoneyTokenServiceFactoryDep) => THoneyTokenProviderHooks;
+  configProviderFactory: (deps: THoneyTokenConfigServiceFactoryDep) => THoneyTokenConfigProvider<T>;
+  registerRouter?: (server: FastifyZodProvider) => Promise<void>;
+};
+
+export const HONEY_TOKEN_PROVIDER_DEFINITIONS: THoneyTokenProviderDefinition[] = [
```
varonix0: I also think the current value in this should be inside an AWS-specific directory. Then we can import it here like 

```typescript
export const HONEY_TOKEN_PROVIDER_DEFINITIONS: THoneyTokenProviderDefinition[] = [
    ...AWS_PROVIDER_DEFINITION,
    ...POSTGRES_PROVIDER_DEFINITION
    // etc ...
];
```

I think generally everything in this directory that is strictly specific to the provider should be in a designated folder for the provider like `src/ee/service/honey-token/aws`

backend/src/ee/services/honey-token/aws/honey-token-aws-config-provider.ts
```diff
@@ -0,0 +1,264 @@
...
+
+export const honeyTokenAwsConfigProviderFactory = ({
+  honeyTokenConfigDAL,
+  permissionService,
+  kmsService,
+  licenseService,
+  appConnectionDAL
+}: TProviderDep) => {
+  const verifyStackDeployment = async ({
+    connectionId: connId,
```
varonix0: nit: why did we inline rename this to connId? doesn't seem to be any conflicts in the file

backend/src/ee/services/honey-token/aws/honey-token-aws-config-provider.ts
```diff
@@ -0,0 +1,264 @@
...
+    stackName,
+    awsRegion
+  }: {
+    connectionId: string;
```
varonix0: Noticing a lot of inline types in a lot of files. We should have a `aws-honey-token-types.ts` file which contains all related types to AWS honey tokens

backend/src/ee/services/honey-token/config/honey-token-config-service.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (!provider) {
+      throw new BadRequestError({ message: "Unsupported honey token type" });
+    }
+    const appConnection = await deps.appConnectionDAL.findById(connectionId);
+    if (!appConnection) {
+      throw new NotFoundError({ message: `Could not find App Connection with ID ${connectionId}` });
+    }
+
+    assertHoneyTokenConnectionType(providerType, appConnection.app);
+    return provider.upsertConfig({ orgPermission, connectionId, config });
```
varonix0: very dangerous: anyone can pass in a connection ID from other orgs here, we never validate that the connection belongs to the org actor. we need to make sure that we are validating the app connection on every path possible for doing any type of mutation

backend/src/ee/services/honey-token/aws/honey-token-aws-config-provider.ts
```diff
@@ -0,0 +1,264 @@
...
+};
+
+export const honeyTokenAwsConfigProviderFactory = ({
+  honeyTokenConfigDAL,
+  permissionService,
+  kmsService,
+  licenseService,
+  appConnectionDAL
+}: TProviderDep) => {
+  const verifyStackDeployment = async ({
```
varonix0: Should be prefixed with `$` to indicate that it's an internal function

backend/src/ee/services/honey-token/honey-token-aws-config-provider.ts
```diff
@@ -0,0 +1,264 @@
...
+        type: HoneyTokenType.AWS,
+        connectionId: null,
+        createdAt: null,
+        updatedAt: null,
+        decryptedConfig: {
+          webhookSigningKey: `htk_live_${crypto.randomBytes(24).toString("hex")}`,
+          stackName: "infisical-honey-tokens",
+          awsRegion: "us-east-1"
+        }
+      };
```
varonix0: Why do we return a "fake" honey token config if nothing is found? Doesn't feel very restful. If this is for UI purposes, I think we should just either throw a 404 or return null and handle it on the UI side of things

backend/src/ee/services/honey-token/config/honey-token-config-service.ts
```diff
@@ -0,0 +1,116 @@
...
+  }: {
+    orgPermission: OrgServiceActor;
+    type: T;
+  }) => {
+    const providerType = assertSupportedHoneyTokenType(type);
+    const provider = honeyTokenConfigProviderByType[providerType];
+    if (!provider) {
+      throw new BadRequestError({ message: "Unsupported honey token type" });
+    }
+    return provider.getConfig({ orgPermission });
```
varonix0: Shouldn't we be doing permission checks before we get into the specific provider functions? Applies to all of these functions that use a specific provider

backend/src/ee/services/honey-token/honey-token-aws-service-helpers.ts
```diff
@@ -0,0 +1,142 @@
...
+}): Promise<{ deployed: boolean; status: string | null }> => {
+  try {
+    const appConnection = await appConnectionDAL.findById(connectionId);
+    if (!appConnection) return { deployed: false, status: null };
+
+    const decryptedConnection = await decryptAppConnection(appConnection, kmsService);
+    const awsConfig = parseAwsConnectionConfig(decryptedConnection);
+    const { credentials: awsCredentials } = await getAwsConnectionConfig(awsConfig);
+
+    const cfn = new CloudFormationClient({ credentials: awsCredentials, region: "us-east-1" });
```
varonix0: Still hardcoded to us-east-1

backend/src/ee/services/honey-token/aws/honey-token-aws-fns.ts
```diff
@@ -0,0 +1,142 @@
+import { CloudFormationClient, DescribeStacksCommand } from "@aws-sdk/client-cloudformation";
```
varonix0, line 1: This file should be called something like `honey-token-aws-fns.ts` to be inline with other patterns

backend/src/ee/services/honey-token/honey-token-aws-types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type TAwsHoneyTokenCredentials = z.infer<typeof AwsHoneyTokenCredentialsSchema>;
+
+export const AwsHoneyTokenDecryptedCredentialsSchema = AwsHoneyTokenCredentialsSchema.extend({
+  iamUserName: z.string()
+});
+
+export type TAwsHoneyTokenDecryptedCredentials = z.infer<typeof AwsHoneyTokenDecryptedCredentialsSchema>;
+
+export const parseAwsHoneyTokenDecryptedCredentials = (value: unknown): TAwsHoneyTokenDecryptedCredentials => {
```
varonix0: Shouldn't this function be inside the helpers file?

backend/src/ee/services/honey-token-config/honey-token-config-dal.ts
```diff
@@ -0,0 +1,10 @@
+import { TDbClient } from "@app/db";
```
varonix0, line 1: Note that the honey token config files should be in their own directory. Right now everything lives inside `honey-token`, but the config ones should live inside a `honey-token-config` folder

backend/src/ee/services/honey-token/honey-token-service-types.ts
```diff
@@ -0,0 +1,53 @@
...
+import { TSecretVersionV2DALFactory } from "@app/services/secret-v2-bridge/secret-version-dal";
+import { TSecretVersionV2TagDALFactory } from "@app/services/secret-v2-bridge/secret-version-tag-dal";
+import { TSmtpService } from "@app/services/smtp/smtp-service";
+
+import { TSecretSnapshotServiceFactory } from "../secret-snapshot/secret-snapshot-service";
+import { THoneyTokenConfigDALFactory } from "./honey-token-config-dal";
+import { THoneyTokenDALFactory } from "./honey-token-dal";
+import { THoneyTokenEventDALFactory } from "./honey-token-event-dal";
+
+export type THoneyTokenServiceFactoryDep = {
```
varonix0: To follow existing patterns, this one is fine to define inside the service itself above the factory like we do in `secret-rotation-v2-service.ts` and elsewhere

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+  secretDAL,
+  secretVersionDAL,
+  secretVersionTagDAL,
+  secretTagDAL,
+  folderCommitService,
+  resourceMetadataDAL,
+  snapshotService,
+  secretQueueService
+}: THoneyTokenServiceFactoryDep) => {
+  const honeyTokenProviderHooksByType: Record<HoneyTokenType, THoneyTokenProviderHooks> =
```
varonix0: nit: we don't need to specify the type `Record<HoneyTokenType, THoneyTokenProviderHooks>` here, as it can already be inferred from the functions return type

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+    if (!plan.honeyTokens) {
+      throw new BadRequestError({
+        message: `Failed to ${action} due to plan restriction. Upgrade plan to use honey tokens.`
+      });
+    }
+    return plan;
+  };
+
+  const getProjectPermission = async (projectId: string, actor: OrgServiceActor) =>
+    permissionService.getProjectPermission({
```
varonix0: This is too much abstraction I think, we can just call the permission service whenever we need to

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+    actor: OrgServiceActor;
+    action: ProjectPermissionHoneyTokenActions;
+  }) => {
+    const { permission } = await getProjectPermission(projectId, actor);
+    if (permission.can(action, ProjectPermissionSub.HoneyTokens)) {
+      return permission;
+    }
+
+    ForbiddenError.from(permission).throwUnlessCan(action, ProjectPermissionSub.HoneyTokens);
+    return permission;
```
varonix0: This doesn't make sense to me - first we check if the actor can do the action, if yes, then we return the permission. If the first check fails, we check the same permission and then we throw an error if not possible. Seems like The whole permission.can() statement can be discarded. Also same thing here, too much abstraction I think; better to call ForbiddenError.from() directly

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+  }) => {
+    const honeyToken = await honeyTokenDAL.findById(honeyTokenId);
+    if (!honeyToken) {
+      throw new NotFoundError({ message: `Honey token with ID "${honeyTokenId}" not found` });
+    }
+
+    try {
+      await assertProjectPermission({ projectId: honeyToken.projectId, actor, action });
+    } catch (error) {
+      if (error instanceof ForbiddenError) {
```
varonix0: Seems like a new pattern. We should just throw the forbidden error, our error handler will already handle this gracefully, and some API users might depend on this pattern for their workflows in the future:

```typescript
 } else if (error instanceof ForbiddenError) {
      void res.status(HttpStatusCodes.Forbidden).send({
        reqId: req.id,
        statusCode: HttpStatusCodes.Forbidden,
        error: "PermissionDenied",
        message: `You are not allowed to ${error.action} on ${error.subjectType}`,
        details: (error.ability as PureAbility).rulesFor(error.action as string, error.subjectType).map((el) => ({
          action: el.action,
          inverted: el.inverted,
          subject: el.subject,
          conditions: el.conditions
        }))
      });
```

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+  ) => {
+    await ensurePlanSupportsHoneyTokens(actor.orgId, "update honey token");
+    const honeyToken = await getHoneyTokenWithProjectAccess({
+      honeyTokenId,
+      actor,
+      action: ProjectPermissionHoneyTokenActions.Create
+    });
+    const { projectId } = honeyToken;
+
+    if (name && name !== honeyToken.name) {
```
varonix0: nit: we don't need to check if name is set. honeyToken.name can never be undefined so if name was undefined, the `name !== honeyToken.name` check would still fail

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+
+  const updateHoneyToken = async (
+    { honeyTokenId, name, description, secretsMapping }: THoneyTokenUpdateInput,
+    actor: OrgServiceActor
+  ) => {
+    await ensurePlanSupportsHoneyTokens(actor.orgId, "update honey token");
+    const honeyToken = await getHoneyTokenWithProjectAccess({
+      honeyTokenId,
+      actor,
+      action: ProjectPermissionHoneyTokenActions.Create
```
varonix0: Shouldn't this be an `Edit` permission? seems like we're missing the Edit permission on the honey token subject

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+        projectDAL.findById(honeyToken.projectId),
+        orgDAL.findOrgMembersByRole(orgId, OrgMembershipRole.Admin)
+      ]);
+      const adminEmails = orgAdmins.map((admin) => admin.user.email).filter(Boolean) as string[];
+      if (adminEmails.length === 0 || !project) return;
+
+      const [folderInfo] = await folderDAL.findSecretPathByFolderIds(project.id, [honeyToken.folderId]);
+
+      const cfg = getAppConfig();
+      const siteUrl = removeTrailingSlash(cfg.SITE_URL || "https://app.infisical.com");
```
varonix0: nit: we don't need to call removeTrailingSlash as we already do that in the envSchema on transform

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+      });
+    } catch (err) {
+      logger.error(
+        { err, orgId, honeyTokenId: honeyToken.id },
+        `Failed to send honey token trigger notification [orgId=${orgId}] [honeyTokenId=${honeyToken.id}]`
+      );
+    }
+  };
+
+  const handleTrigger = async ({
```
varonix0: Feels strange that the org ID isn't apart of the payload. Shouldn't that also be signed / be apart of the payload + signature? Otherwise most of this flow is allowed to run if someone passes a valid signature + payload, but then passes a fake org ID

backend/src/ee/services/honey-token/honey-token-service.ts
varonix0: Should be prefixed with $ to indicate its an internal method

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+      used,
+      limit: plan.honeyTokenLimit
+    };
+  };
+
+  const getDashboardHoneyTokens = async (
+    { projectId, environments, secretPath, search, orderBy, orderDirection, limit, offset }: THoneyTokenListInput,
+    actor: OrgServiceActor
+  ) => {
+    await getProjectPermission(projectId, actor);
```
varonix0: We aren't checking if the user has access to read honey tokens

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+      orgId: actor.orgId,
+      secretPath,
+      projectId,
+      environmentSlug: environment,
+      excludeReplication: true
+    });
+
+    // This is not a blocker, so if it fails, we just warn the user that they need
+    // to deploy the stack.
+    const stackDeployment = providerHooks.verifyDeployment
```
varonix0: I would give an error early on in the create if we aren't able to verify the deployment. Since we currently don't have a way to display/persist an error on the honey token that the deployment is misconfigured, I think it's better to throw an error early

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+      ...(description !== undefined && { description }),
+      ...(secretsMapping !== undefined && { secretsMapping })
+    });
+
+    await secretDAL.invalidateSecretCacheByProjectId(projectId);
+    await snapshotService.performSnapshot(honeyToken.folderId);
+
+    const [folderInfo] = await folderDAL.findSecretPathByFolderIds(projectId, [honeyToken.folderId]);
+    if (folderInfo) {
+      await secretQueueService.syncSecrets({
```
varonix0, line 577: Shouldn't we only do this if the secret mappings have changed? Seems like the only thing in this update function that could effect the secrets state

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+    });
+    const { projectId } = honeyToken;
+
+    if (honeyToken.status === HoneyTokenStatus.Revoked) {
+      throw new BadRequestError({ message: "Honey token is already revoked" });
+    }
+    const type = assertSupportedHoneyTokenType(honeyToken.type);
+
+    const { decryptor } = await kmsService.createCipherPairWithDataKey({
+      type: KmsDataKey.Organization,
```
varonix0: why are we using the organization data key instead of the project data key for encrypting the honey token credentials? I think this should be using the project data key. we only encrypt org level data with the org data key, project data is always encrypted with the project's data key. fyi that we need to update this in quite a lot of places

backend/src/db/schemas/honey-tokens.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const HoneyTokensSchema = z.object({
+  id: z.string().uuid(),
+  name: z.string(),
+  description: z.string().nullable().optional(),
+  type: z.string(),
+  status: z.string().default("active"),
+  projectId: z.string(),
+  folderId: z.string().uuid(),
+  connectionId: z.string().uuid(),
```
varonix0: Why do we store the connection ID on the honey token schema? shouldn't it be enough to just store it inside the org honey token config, since all honey tokens use the same org-level config for creating the actual honey tokens?

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+          actor: {
+            type: actor.type as ActorType,
+            actorId: actor.id
+          },
+          tx
+        });
+      });
+    }
+
+    const updated = await honeyTokenDAL.updateById(honeyTokenId, {
```
varonix0: This should also be apart of the transaction, as it mutates data

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+    const { decryptor } = await kmsService.createCipherPairWithDataKey({
+      type: KmsDataKey.Organization,
+      orgId: actor.orgId
+    });
+
+    const decryptedCredentials = JSON.parse(
+      decryptor({ cipherTextBlob: honeyToken.encryptedCredentials }).toString()
+    ) as Record<string, string>;
+
+    const appConnection = await appConnectionDAL.findById(honeyToken.connectionId);
```
varonix0: relates to previous comment; but we shouldn't be storing the connection ID on the honey token itself I think?

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+
+    if (appConnection) {
```
varonix0: question: can this actually be false?

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -0,0 +1,892 @@
...
+    }
+
+    return honeyTokens;
+  };
+
+  const getCredentials = async ({ honeyTokenId }: THoneyTokenByIdInput, actor: OrgServiceActor) => {
+    const honeyToken = await getHoneyTokenWithProjectAccess({
+      honeyTokenId,
+      actor,
+      action: ProjectPermissionHoneyTokenActions.Read
```
varonix0: Didn't we agree that for reading credentials, there should be a separate permission for reading the generated credentials like we have for secret rotations? `ReadGeneratedCredentials = "read-generated-credentials"`

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -553,6 +565,11 @@ const buildMemberPermissionRules = () => {
     ProjectPermissionSub.ServiceTokens
   );
 
+  can(
+    [ProjectPermissionHoneyTokenActions.Read, ProjectPermissionHoneyTokenActions.Create],
```
varonix0: curious what you think:

I think a member should only be allowed to `Read` on honey tokens (assuming we make a dedicated permission for reading the generated credentials like mentioned earlier)

this is also in line with what we do for secret rotations. feels strange that the user is able to create honey tokens, but they can't update or delete it again - should likely just be read!

backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -31,6 +31,31 @@ import {
 import { PostHogEventTypes } from "@app/services/telemetry/telemetry-types";
 
 const MAX_DEEP_SEARCH_LIMIT = 500; // arbitrary limit to prevent excessive results
+const INCLUDE_HONEY_TOKENS_OVERVIEW_DESC = "Whether to include honey tokens in the overview response.";
+const INCLUDE_HONEY_TOKENS_DETAILS_DESC = "Whether to include honey tokens in the details response.";
+
+const SanitizedHoneyTokenSchema = HoneyTokensSchema.pick({
```
varonix0: Should be defined in `backend/src/server/routes/sanitizedSchemas.ts`

backend/src/server/routes/v1/dashboard-router.ts
```diff
@@ -31,6 +31,31 @@ import {
 import { PostHogEventTypes } from "@app/services/telemetry/telemetry-types";
 
 const MAX_DEEP_SEARCH_LIMIT = 500; // arbitrary limit to prevent excessive results
+const INCLUDE_HONEY_TOKENS_OVERVIEW_DESC = "Whether to include honey tokens in the overview response.";
+const INCLUDE_HONEY_TOKENS_DETAILS_DESC = "Whether to include honey tokens in the details response.";
```
varonix0: Should be defined in `/backend/src/lib/api-docs/constants.ts` like we do for all other describe documentation, refer to other examples in this file

backend/src/services/smtp/emails/HoneyTokenTriggeredTemplate.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <Section className="px-[16px] py-[12px] text-center">
+        <Text className="text-red-600 text-[14px] font-bold m-0">ACTION REQUIRED — POTENTIAL SECRET COMPROMISE</Text>
+      </Section>
+      <Heading className="text-black text-[18px] leading-[28px] text-center font-normal p-0 mx-0 mt-[24px]">
+        Honey token <strong>{honeyTokenName}</strong> was accessed in project{" "}
+        <strong>{projectName}</strong>
+      </Heading>
+      <Text className="text-center">
+        An unauthorized party may have access to your secrets. Investigate immediately and rotate any secrets that
+        may have been exposed.
```
varonix0: There's lint errors in this file: `Delete `{"·"}⏎·······`eslintprettier/prettier`. not sure why this wasn't caught by our lint check

varonix0 (changes requested):

backend/src/ee/services/honey-token/aws/honey-token-aws-config-provider.ts
mathnogueira: also make config in another folder

backend/src/ee/services/honey-token/honey-token-aws-config-provider.ts
mathnogueira: should be done in the frontend instead of making the BE returning the default value

backend/src/ee/services/honey-token/honey-token-service.ts
mathnogueira: remove Org id and infer it from the accessKeyId

backend/src/ee/services/honey-token/honey-token-service.ts
mathnogueira: No, you cannot delete the app connection without deleting the honey token / config first.

backend/src/db/migrations/20260505120000_remove-honey-token-connection-id.ts
```diff
@@ -0,0 +1,32 @@
+import { Knex } from "knex";
```
varonix0: Combine into single migration

varonix0 (approved):

PR #6245, opened 2026-04-29
feature(platfor-313): add upgrade impact tooling

PrestigePvP:
## Context

- Adds release impact data tooling.
- Collects self-hosted upgrade evidence.
- Generates AI-authored release impact files.
- Validates committed upgrade impact data.
- Opens generated data through PRs.

## Steps to verify the change

- `npm run type:check --prefix upgrade-impact`
- `npm test --prefix upgrade-impact`
- `npm run validate --prefix upgrade-impact`

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


.github/workflows/generate-upgrade-impact.yml
```diff
@@ -0,0 +1,98 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Resolve tag
```
maidul98, line 22: test comment for review police

upgrade-impact/scripts/generate.ts
```diff
@@ -0,0 +1,152 @@
...
+  noAi?: boolean;
+};
+
+export const generateReleaseImpact = async ({
+  tag,
+  noAi = false
+}: GenerateReleaseImpactOptions) => {
+  const bundle = await collectEvidence(tag);
+  const sourceModel = noAi ? "deterministic" : MODEL;
+  const draft = noAi ? deterministicDraft(bundle) : await generateWithOpenAiAgentic(bundle);
```
maidul98, line 120: What's the purpose of deterministic draft? I don't see it being used in tests or anywhere

maidul98 (approved):
Generally looks good. Left one commnet for what looks like it might be unused code. 

Can do later but, i think it might be worth adding tests for compareVersions since if that regresses at any point the whole thing will fall apart

upgrade-impact/scripts/generate.ts
PrestigePvP: The goal was that if we wanted to disable the AI piece we could still have the feature work. 

This used to be what was fed into the AI before we used an agentic approach so I just decided to keep it in case our key expires or something and we just want to quickly swap it out. 

It was also useful to isolate during development to see what our input what and what the output from the AI was so I just decided to keep it.

maidul98: ack

PR #6250, opened 2026-04-30
improvement(security): migrate outbound Axios to safeRequest

victorvhs017:
## Context

This change strengthens **SSRF defenses** for outbound HTTP across the backend by routing user- and admin-configured URLs through **`safeRequest`** (`@app/lib/validator`) instead of raw Axios. **`safeRequest`** validates the URL, **pins DNS** to the resolved addresses (mitigating DNS rebinding / TOCTOU between check and connect), and **disables redirects** by default (`maxRedirects: 0`), aligning with the patterns documented in `backend/CLAUDE.md`.

**Before:** Many flows used `request` after optional `blockLocalAndPrivateIpAddresses`, which does **not** close the DNS rebinding window for HTTP.

**After:** User-derived HTTP calls use **`safeRequest.{get,post,put,patch,delete,request}`** where applicable. **`blockLocalAndPrivateIpAddresses`** remains for **save-time validation** and **non-HTTP clients** (LDAP, SSH, SMB, etc.). Third-party libraries that build their own HTTP stack continue to use **`buildSsrfSafeAgent`** where already integrated.

**Notable areas in this branch:**

- **Audit log streams** (Azure, Cribl, Custom, Datadog, Splunk): migrated to `safeRequest`; redundant pre-`blockLocalAndPrivateIpAddresses` before HTTP removed where `safeRequest` owns request-time protection.
- **Dynamic secret providers** (Azure Entra ID, GitHub, Couchbase, **Kubernetes**): Axios paths use `safeRequest`; Couchbase drops duplicate pre-block for the request itself.
- **Chef app connection:** synchronous `getChefServerUrl`; all HTTP via `safeRequest`.
- **Kubernetes dynamic secrets:** **`k8sHttpClient`** chooses **`safeRequest`** on the **direct** path (flat `ca` / `rejectUnauthorized`, `allowPrivateIps: true`, **no** caller `httpsAgent`) vs **raw `request`** on the **gateway** path with an explicit **`isGatewayProxied`** boolean (**caller-asserted**, not URL-inferred) to avoid DNS rebinding bypasses. New unit tests in `kubernetes.test.ts` assert this dispatch.
- **App connection gateway wrappers** (NetScaler, GitHub, HC Vault, Venafi TPP): **`requestWith*Gateway`** keeps raw `request` + relay `httpsAgent` on gateway tunnel; direct path uses **`safeRequest`** with tightened config types (`AxiosRequestConfig & { url: string }`) to avoid unsafe `url as string`.
- **Identity Kubernetes auth** (from branch context): TokenReview direct path uses `safeRequest` with TLS options and **`$resolveEffectiveVerifyTlsCertificate`** where merged from main.
- **`validate-url` / `safeRequest`:** Extended support for TLS-related agent options (e.g. **`servername`**) passed into the pinned agent; documentation updated.

**Docs:** Root `CLAUDE.md` and **`backend/CLAUDE.md`** gain an **Outbound HTTP & SSRF Protection** section (when to use `safeRequest`, `ssrfSafeGet`, `buildSsrfSafeAgent`, gateway patterns, `allowPrivateIps`, and the requirement that gateway vs direct be **caller-asserted**).

## Steps to verify the change

1. From the repo root: `make reviewable-api` (lint + typecheck for the backend).
2. Run **`backend/src/lib/validator/validate-url.test.ts`** (SSRF helper coverage).
3. Run **`backend/src/ee/services/dynamic-secret/providers/kubernetes.test.ts`** (K8s provider dispatch).
4. Smoke: **Custom audit log stream** to `http://host.docker.internal:...` from a Dockerized API; **HC Vault** dev server with `http://host.docker.internal:8200`; **Kubernetes dynamic secret** against kind with correct CA (`kubectl config view --minify`) and apiserver SANs including `host.docker.internal` if using that hostname.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/dynamic-secret/providers/rabbit-mq.ts
```diff
@@ -93,9 +102,9 @@ export const RabbitMqProvider = (): TDynamicProviderFns => {
         "Content-Type": "application/json"
       },
       maxRedirects: 0,
-      ...(providerInputs.ca && {
-        httpsAgent: new https.Agent({ ca: providerInputs.ca, rejectUnauthorized: providerInputs.sslRejectUnauthorized })
-      })
+      ...(isHttps
+        ? { httpsAgent: pinnedAgent as https.Agent | undefined }
+        : { httpAgent: pinnedAgent as import("http").Agent | undefined })
```
PrestigePvP: I don't love the dynamic import here if we can avoid it

backend/src/ee/services/oidc/oidc-config-service.ts
```diff
@@ -687,8 +688,19 @@ export const oidcConfigServiceFactory = ({
           message: "OIDC not configured correctly"
         });
       }
-      await blockLocalAndPrivateIpAddresses(oidcCfg.discoveryURL);
-      issuer = await Issuer.discover(oidcCfg.discoveryURL);
+      // Fetch discovery doc via safeRequest so the discovery leg is itself
```
PrestigePvP, line 691: Can we remove these huge chunks of comments? I know Claude likes to throw them out there but it's often times not super helpful.

backend/src/services/app-connection/azure-adcs/azure-adcs-connection-fns.ts
```diff
@@ -179,8 +186,10 @@ const testAdcsConnection = async (
 
       const shouldRejectUnauthorized = sslRejectUnauthorized;
 
-      const httpsAgent = createHttpsAgent(shouldRejectUnauthorized, sslCertificate);
+      // eslint-disable-next-line no-await-in-loop
```
PrestigePvP, line 189: Is there any reason these are sequential and not in a `Promise.`?

backend/src/services/external-migration/external-migration-fns/vault.ts
```diff
@@ -309,8 +321,12 @@ const vaultFactory = (
         getData
       );
     } else {
-      await blockLocalAndPrivateIpAddresses(baseUrl);
-      data = await getData(baseUrl);
+      const pinnedAgent = await buildSsrfSafeAgent(baseUrl, { rejectUnauthorized: true });
+      const isHttps = new URL(baseUrl).protocol === "https:";
```
PrestigePvP, line 325: Does the protocol have `:`?

backend/src/ee/services/oidc/oidc-config-service.ts
victorvhs017: I tried to leave only the ones that actually add context, like in this example, it's to explain why we are not using `Issuer.discover(oidcCfg.discoveryURL)` here, as it's not as straightforward as the `request` to `safeRequest` changes we have in other places.

backend/src/services/app-connection/azure-adcs/azure-adcs-connection-fns.ts
victorvhs017: Do you mean `Promise.all`? We can't because we use the httpsAgent in the second request

backend/src/services/external-migration/external-migration-fns/vault.ts
victorvhs017: Yes, that's the node api https://nodejs.org/api/url.html#urlprotocol

backend/src/services/app-connection/azure-adcs/azure-adcs-connection-fns.ts
PrestigePvP: Yeah, but I was thinking we'd made this a funciton and then `Promise.all(list.map(item => function(item))` which removes the eslint disable.

victorvhs017: Yep, I think this is out of scope of the request/agent creation here. But it seems intentional because we have some early returns inside the loop.

PrestigePvP (approved):

PR #6256, opened 2026-04-30
fix(kubernetes-auth): configurable TLS verification for identity Kubernetes auth

victorvhs017:
## Context

Restores **opt-out TLS certificate verification** for Identity **Kubernetes Auth** in **API** token-review mode, after [PR #6242](https://github.com/Infisical/infisical/pull/6242) enforced verification whenever no custom CA was configured—breaking clusters whose API server certs are not in the platform trust store (e.g. EKS) unless a CA is supplied.

Adds **`verifyTlsCertificate`** on the machine-identity Kubernetes auth config (DB column `verifyTlsCertificate`, API, UI). When **`true`**, Infisical verifies the Kubernetes API server’s cert using the configured **CA certificate** (required in that case). When **`false`**, HTTPS is still used but **`rejectUnauthorized`** is off (legacy / insecure, explicit customer choice). Migration **backfills `true`** for rows that already have `encryptedKubernetesCaCertificate` set; **`false`** otherwise. Attach uses `verifyTlsCertificate ?? Boolean(caCert)`; PATCH auto-promotes to `true` when a non-empty `caCert` is sent without the flag.

Renames the field from **`enableSsl`** to **`verifyTlsCertificate`** (TLS is always on; the flag only controls **verification**).

## Screenshots

image

## Steps to verify the change

- **Attach** Kubernetes Auth (API mode): with **Verify TLS Certificate** on and a valid cluster **CA**, save and **Kubernetes login** succeeds.
- Same identity with verification **off** and **no CA** (only if you accept the risk): save and login against a cluster that would fail public-CA verification.
- **PATCH** update: sending a non-empty `caCert` without `verifyTlsCertificate` promotes verification to **on**; cannot end up with verification **on** and no CA (router + service guard).
- Run migration on a dev DB: column exists; rows with stored CA get `verifyTlsCertificate: true`.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/identity-kubernetes-auth-router.ts
+++ b/backend/src/server/routes/v1/identity-kubernetes-auth-router.ts
@@ -32,7 +32,7 @@ const IdentityKubernetesAuthResponseSchema = IdentityKubernetesAuthsSchema.pick(
   allowedAudience: true,
   gatewayId: true,
   gatewayPoolId: true,
-  enableSsl: true
+  verifyTlsCertificate: true
 }).extend({
   caCert: z.string(),
   tokenReviewerJwt: z.string().optional().nullable()
@@ -189,7 +189,7 @@ export const registerIdentityKubernetesRouter = async (server: FastifyZodProvide
               }
             ),
           caCert: z.string().trim().default("").describe(KUBERNETES_AUTH.ATTACH.caCert),
-          enableSsl: z.boolean().optional().describe(KUBERNETES_AUTH.ATTACH.enableSsl),
+          verifyTlsCertificate: z.boolean().optional().describe(KUBERNETES_AUTH.ATTACH.verifyTlsCertificate),
           tokenReviewerJwt: z.string().trim().optional().describe(KUBERNETES_AUTH.ATTACH.tokenReviewerJwt),
           tokenReviewMode: z
             .nativeEnum(IdentityKubernetesAuthTokenReviewMode)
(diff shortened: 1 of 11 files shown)


backend/src/server/routes/v1/identity-kubernetes-auth-router.ts
```diff
@@ -261,6 +263,30 @@ export const registerIdentityKubernetesRouter = async (server: FastifyZodProvide
               message: "Access Token TTL cannot be greater than Access Token Max TTL."
             });
           }
+          if (
```
PrestigePvP, line 266: Kinda NIT: Structurally I would've liked to see this outlined a bit different for readability. Something like:

```
if tokenReviewMode === API {
 if verifyTLSCert && !data.caCert.length {
  addIssue...
 } else {
  addIssue...
}
}```

File: backend/src/services/identity-kubernetes-auth/identity-kubernetes-auth-service.ts [hunk 2]
```diff
@@ -476,6 +478,8 @@ export const identityKubernetesAuthServiceFactory = ({
               ? undefined
               : ((identityKubernetesAuth.gatewayV2Id ?? identityKubernetesAuth.gatewayId) as string),
             gatewayPoolId: identityKubernetesAuth.gatewayPoolId ?? undefined,
+            caCert: caCert || undefined,
+            verifyTlsCertificate: identityKubernetesAuth.verifyTlsCertificate,
```
PrestigePvP, line 482: Can these still be undefined? Do we want to add the `false` fallback?

PrestigePvP (approved):

backend/src/services/identity-kubernetes-auth/identity-kubernetes-auth-service.ts
victorvhs017: It's default to false, all good! We want this to allow backwards compatibility for folks without CAs

PR #6257, opened 2026-04-30
feature: Go sidecar first version for development

akhilmhdh:
## Context

This PR is start of Infisical backend porting Nodejs to Golang. Initially golang would be just a sidecar and nodejs would proxy various ported logic to handle for Golang. 

At some point NodeJS development would be paused and Golang would become the bigger picture and NodeJS would become sidecar to deprecated. 

This PR will not go for production until the pending TODOs are done as well. This is more towards development setup and reduced scope of our secret endpoint working. 

This PR doesn't implement following things
1. Identity access token status update queue.
2. Secret caching for both value and etag
3. External KMS support
4. HSM service is written not plugged in yet.
5. Audit log streaming
6. Assumed privilege injection

The above will be implemented and only after that prod image will be having the Go sidecar build.

## Steps to verify the change

1. Add an environment variable `GOLANG_SIDECAR_URL=http://backend-go:4040`
2. docker compose -f docker-compose.dev.yml --profile go up
3. Test the secret routes endpoints v4 List/GetByName getting forwarded to Golang

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend-go/cmd/infisical/main.go
```diff
@@ -0,0 +1,130 @@
...
+	var wg sync.WaitGroup
+	ctx, cancel := context.WithCancel(ctx)
+
+	// Start queue worker.
+	wg.Go(func() {
+		if err := queueSvc.Start(queue.ServerConfig{Concurrency: 10}); err != nil {
+			logger.ErrorContext(ctx, "queue worker error", slog.Any("error", err))
+			errc <- err
+		}
+	})
```
sheensantoscapadngan, line 116: this is for which jobs?

backend-go/internal/config/config.go
```diff
@@ -0,0 +1,616 @@
...
+
+type NodeEnv string
+
+const (
+	NodeEnvDevelopment NodeEnv = "development"
+	NodeEnvTest        NodeEnv = "test"
+	NodeEnvProduction  NodeEnv = "production"
+)
+
+type Config struct {
```
sheensantoscapadngan, line 31: would it make sense to have nesting here? wonder if it will make it easier to consume if we had `config.Db`, `config.Clickhouse`, etc.

backend-go/internal/config/config.go
```diff
@@ -0,0 +1,616 @@
...
+	// Development flags
+	RotationDevelopmentMode                     bool   `koanf:"ROTATION_DEVELOPMENT_MODE"`
+	DailyResourceCleanUpDevelopmentMode         bool   `koanf:"DAILY_RESOURCE_CLEAN_UP_DEVELOPMENT_MODE"`
+	BDDNockAPIEnabled                           bool   `koanf:"BDD_NOCK_API_ENABLED"`
+	ACMEDevelopmentMode                         bool   `koanf:"ACME_DEVELOPMENT_MODE"`
+	ACMESkipUpstreamValidation                  bool   `koanf:"ACME_SKIP_UPSTREAM_VALIDATION"`
+	ACMEDevelopmentHTTP01ChallengeHostOverrides string `koanf:"ACME_DEVELOPMENT_HTTP01_CHALLENGE_HOST_OVERRIDES"`
+	ACMEDNSResolverServers                      string `koanf:"ACME_DNS_RESOLVER_SERVERS"`
+	ACMEDNSResolveResolverServersHostEnabled    bool   `koanf:"ACME_DNS_RESOLVE_RESOLVER_SERVERS_HOST_ENABLED"`
+	DNSMadeEasySandboxEnabled                   bool   `koanf:"DNS_MADE_EASY_SANDBOX_ENABLED"`
```
sheensantoscapadngan: I assume these will all be inside `.dev` then

backend-go/internal/config/config.go
```diff
@@ -0,0 +1,616 @@
...
+}
+
+// ValidationError holds a list of config validation issues.
+type ValidationError struct {
+	Issues []string
+}
+
+func (e *ValidationError) Error() string {
+	return fmt.Sprintf("invalid environment variables:\n  - %s", strings.Join(e.Issues, "\n  - "))
+}
```
sheensantoscapadngan, line 826: should we instead centralize all errors inside a specific directory?

backend-go/internal/config/config.go
```diff
@@ -0,0 +1,616 @@
...
+
+// parseHostPortList parses a comma-separated "host:port" string into []RedisHostPort.
+// Matches the Node.js transform: val.split(",").map(el => { host, port }).
+// Returns parsing issues for invalid port values.
+func parseHostPortList(raw, envVar string) (result []RedisHostPort, issues []string) {
```
sheensantoscapadngan, line 829: this should be specific to redis right so perhaps we should add an indicator to the name

backend-go/internal/crypto/cipher/cipher_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"bytes"
+	"crypto/rand"
+	"testing"
+
+	"github.com/infisical/api/internal/crypto/cipher"
+)
+
+func TestEncryptDecryptRoundTrip(t *testing.T) {
```
sheensantoscapadngan: we should standardize the naming of the test functions... With the current setup, it will become really difficult to identify which test functions are testing which methods. Could we instead do something like: `Test<FunctionName>_<Scenario>`?

backend-go/internal/crypto/hash/hash.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// SHA384 returns the raw SHA-384 digest (48 bytes).
+func SHA384(data []byte) []byte {
+	h := sha512.Sum384(data)
+	return h[:]
+}
+
+// SHA512 returns the raw SHA-512 digest (64 bytes).
+func SHA512Bytes(data []byte) []byte {
```
sheensantoscapadngan: why does this have the "Bytes" prefix but not the others?

backend-go/internal/crypto/hash/hash.go
```diff
@@ -0,0 +1,115 @@
...
+// Use only where required for protocol compatibility.
+func HMACSHA1(key, data []byte) []byte {
+	return computeHMAC(sha1.New, key, data) //nolint:gosec // SHA-1 required for protocol compatibility
+}
+
+// TimingSafeEqual performs constant-time comparison of two byte slices.
+// Returns true only if a and b have the same length and identical contents.
+func TimingSafeEqual(a, b []byte) bool {
+	return subtle.ConstantTimeCompare(a, b) == 1
+}
```
sheensantoscapadngan: method feels misplaced. I wonder if we should just call `subtle.ConstantTimeCompare` inline instead of even providing a helper function?

backend-go/internal/crypto/hash/hash_test.go
```diff
@@ -0,0 +1,86 @@
+package hash_test
```
sheensantoscapadngan: same feedback here about conventions.

backend-go/internal/crypto/hash/hash_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"encoding/hex"
+	"testing"
+
+	"github.com/infisical/api/internal/crypto/hash"
+)
+
+func TestSHA256Hex(t *testing.T) {
```
sheensantoscapadngan: do we need to add tests even for units without any custom logic? feels like this might be redundant since these methods likely already have testing in the library

backend-go/internal/crypto/sign/sign_test.go
```diff
@@ -0,0 +1,187 @@
...
+		t.Fatal("expected error for unsupported algorithm")
+	}
+}
+
+func TestUnsupportedKeyAlgorithm(t *testing.T) {
+	_, err := sign.GeneratePrivateKey("DSA_2048")
+	if err == nil {
+		t.Fatal("expected error for unsupported key algorithm")
+	}
+}
```
sheensantoscapadngan: same feedback here re: conventions

backend-go/internal/database/pg/pg.go
```diff
@@ -0,0 +1,212 @@
+package pg
```
sheensantoscapadngan, line 1: would it make sense to have testing for this as well ?

backend-go/internal/database/redis/redis.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"crypto/tls"
+	"fmt"
+	"net"
+
+	"github.com/redis/go-redis/v9"
+
+	"github.com/infisical/api/internal/config"
+)
+
```
sheensantoscapadngan, line 13: same Q about testing, should we have it for this as well?

backend-go/internal/keystore/keystore.go
```diff
@@ -0,0 +1,162 @@
...
+		if ownedTx {
+			if err := tx.Rollback(); err != nil {
+				return nil, fmt.Errorf("acquiring advisory lock %d: %w", lockID, err)
+			}
+		}
+		return nil, fmt.Errorf("acquiring advisory lock %d: %w", lockID, err)
+	}
+
+	return &Lock{tx: tx, ownedTx: ownedTx}, nil
+}
```
sheensantoscapadngan: this method feels a bit lost

backend-go/internal/libs/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,63 @@
...
+	Name string
+	Ok   bool
+	Err  error
+}
+
+// DBReport holds the results of checking all database connections.
+type DBReport struct {
+	Primary  ConnectionStatus
+	Replicas []ConnectionStatus
+}
```
sheensantoscapadngan, line 22: need to have a standard for when to make things public/private. Seems like these can just be made private?

backend-go/internal/services/auditlog/queue_handler.go
```diff
@@ -0,0 +1,243 @@
+package auditlog
```
sheensantoscapadngan, line 1: is the idea here that if I have three queue handlers for PKI, I am supposed to just put everything inside this one file? in our node, we'd have different files

backend-go/internal/queue/tasks.go
```diff
@@ -0,0 +1,7 @@
+package queue
+
+// Task type names are defined here as constants.
+// Each task has a corresponding typed Task[P] definition that enforces payload type safety.
+const (
+	TaskNameAuditLog = "audit:log"
+)
```
sheensantoscapadngan, line 7: why do we have to define his in here rather than just declare it from inside services/auditlog/types.go?

backend-go/internal/services/auditlog/event.go
```diff
@@ -0,0 +1,168 @@
+package auditlog
```
sheensantoscapadngan, line 1: it would be a lot easier to read and maintain if we just split it up into multiple files like the ff instead of cramming everything into a single `types.go` file

auditlog/
  event.go          # EventType, EventMetadata, Event, GetSecretsEventMetadata, etc.
  actor.go          # ActorType, Actor, UserActorMetadata, etc.
  useragent.go      # UserAgentType (already exists)
  dto.go            # CreateAuditLogDTO
  service.go        # Service
  dal.go            # DAL

backend-go/internal/server/httpinfo.go
```diff
@@ -0,0 +1,50 @@
+package server
```
sheensantoscapadngan: this feels lost. should we have a middleware directory to organize everything there instead?

backend-go/internal/libs/logutil/context_handler.go
```diff
@@ -0,0 +1,40 @@
+package logutil
```
sheensantoscapadngan, line 1: do we still need the `util` suffix? the other libs do not have that. logcontext maybe?

backend-go/internal/libs/errutil/defer.go
```diff
@@ -0,0 +1,17 @@
+package errutil
```
sheensantoscapadngan, line 1: same feedback here about the `util` suffix. apperr maybe?

backend-go/internal/libs/crypto/cipher/cipher.go
```diff
@@ -0,0 +1,64 @@
+package cipher
```
sheensantoscapadngan, line 1: IGNORE PLACEMENT OF COMMENT

shouldn't crypto live inside the libs directory?

backend-go/cmd/infisical/main.go
```diff
@@ -0,0 +1,130 @@
+package main
+
```
sheensantoscapadngan, line 2: IGNORE PLACEMENT OF COMMENT


One pattern I'm noticing throughout: handlers depend on concrete service types (e.g., *permission.Service) rather than interfaces. For long-term testability, we should define interfaces at the handler level for the services we consume...

BEFORE:
```
// server/api/secretmanager/secrets/secrets.go

type Handler struct {
    auth.AuthHandler
    logger           *slog.Logger
    sharedSvc        *services.Services           // Concrete struct
    secretManagerSvc *secretmanager.Services      // Concrete struct
}

type Deps struct {
    Logger           *slog.Logger
    SharedSvc        *services.Services           // Concrete
    SecretManagerSvc *secretmanager.Services      // Concrete
}
```

AFTER
```
// server/api/secretmanager/secrets/secrets.go
// Define only what this handler NEEDS (narrow interfaces)
type permissionService interface {
    CheckProjectAccess(ctx context.Context, identity *auth.Identity, projectID string, action string) error
}
type kmsService interface {
    CreateCipherPair(ctx context.Context, dto kms.CreateCipherPairDTO) (*kms.CipherPair, error)
}
type secretService interface {
    ListSecrets(ctx context.Context, opts ListSecretsOpts) ([]Secret, error)
    GetSecretByName(ctx context.Context, opts GetSecretOpts) (*Secret, error)
}
type Handler struct {
    auth.AuthHandler
    logger     *slog.Logger
    permission permissionService    // Interface
    kms        kmsService           // Interface
    secrets    secretService        // Interface
}
type Deps struct {
    Logger     *slog.Logger
    Permission permissionService    // Interface
    KMS        kmsService           // Interface  
    Secrets    secretService        // Interface
}
```

backend-go/internal/server/api/secretmanager/secrets/secrets.go
```diff
@@ -0,0 +1,499 @@
...
+	proj, err := h.sharedSvc.Project.GetBySlug(ctx, identity.OrgID, *workspaceSlug)
+	if err != nil {
+		return "", errutil.DatabaseErr("Failed to resolve project").WithErrf("resolveProjectID(slug=%s): %w", *workspaceSlug, err)
+	}
+	if proj == nil {
+		return "", errutil.NotFound("Project not found").WithErrf("resolveProjectID: project with slug '%s' not found in org", *workspaceSlug)
+	}
+
+	return proj.ID, nil
+}
```
sheensantoscapadngan: Isn't this something that should live somewhere else though? Like in the libs for projects?

backend-go/internal/server/api/secretmanager/secrets/secrets.go
```diff
@@ -0,0 +1,499 @@
...
+		result = append(result, secrets.SecretInput{
+			ID:         sec.ID,
+			Key:        sec.Key,
+			Value:      secretValue,
+			Env:        envSlug,
+			Path:       loc.path,
+			IsImported: true,
+		})
+	}
+
```
sheensantoscapadngan: Why is there a lot of business logic in this file? I thought we wanted to separate this so that everything else should either be in services or inside libs. Right now it feels like this is a dump of everything.

backend-go/internal/crypto/hash/hash.go
```diff
@@ -0,0 +1,115 @@
...
+func SHA384(data []byte) []byte {
+	h := sha512.Sum384(data)
+	return h[:]
+}
+
+// SHA512 returns the raw SHA-512 digest (64 bytes).
+func SHA512Bytes(data []byte) []byte {
+	h := sha512.Sum512(data)
+	return h[:]
+}
```
PrestigePvP: How much do these per-algorithm wrappers actually buy us? `SHA256(data)` is `sha256.Sum256(data)[:]` — we're saving a `[:]`. `SHA256Hex` saves one line.

The two helpers in this file that *do* earn their keep:
- `NewHasher`/`NewHMAC` — runtime enum→`hash.Hash` dispatch (real value when the algo name comes from config / a DB row / a JWS header).
- `TimingSafeEqual` — wraps the `subtle.ConstantTimeCompare(a,b) == 1` footgun.

Worth trimming this file down to just the dispatch + constant-time helper, and letting callsites use stdlib directly for the rest?

backend-go/internal/crypto/hash/hash.go
```diff
@@ -0,0 +1,115 @@
+// Package hash provides cryptographic hash and HMAC functions used across
+// the Infisical backend: SHA-256, SHA-1, HMAC-SHA-256, HMAC-SHA-1, and
+// constant-time comparison.
+package hash
```
PrestigePvP: Should this whole `crypto/` tree (hash, cipher, sign) move to `/pkg/`?

It has zero Infisical-specific logic — no config, no DB, no errutil, no service deps. Pure primitives + dispatch. That's the real signal: it could compile in a fresh module with no other Infisical code.

If we ever want to share with the Infisical CLI, a Go gateway agent, or even FFI it for a TS port, `internal/` blocks that. `/pkg/` is contested as a convention but the *boundary* it represents ("this is platform, not product") is honest here in a way it wouldn't be for `services/auth` or `services/kms`.

backend-go/cmd/infisical/main.go
akhilmhdh: Audit log

backend-go/internal/config/config.go
akhilmhdh: This is just for config and it's only used by configuration. I don't think we need to move to a centralized because it's used for internally when deploying things.

backend-go/internal/config/config.go
akhilmhdh: Agreed

backend-go/internal/crypto/hash/hash.go
akhilmhdh: I think we shouldn't put in public pkg yet - until we are sure we need it. I do agree that it can be used in cli or other things. But that should be done when that decision has come or we are sure we need it and not everything. Because if cli is using this - it creates a contract and contracts should be thought about otherwise it would just break things.

backend-go/internal/crypto/cipher/cipher_test.go
akhilmhdh: Agreed

backend-go/internal/keystore/keystore.go
akhilmhdh: Can you define what you meant by lost exactly?

backend-go/internal/queue/tasks.go
akhilmhdh: It's more like the base library provides all core functionality and queue constants and everyone can use that.

backend-go/internal/services/auditlog/queue_handler.go
akhilmhdh: This is audit log queue handler! What do you mean PKI here? 

For PKI it would in PKI queue. The queue system is a bit different compared to nodejs. The base setup provides the workers running per container. You then registers the job against that,

backend-go/internal/services/auditlog/event.go
akhilmhdh: Makes sense

backend-go/internal/server/api/platform/projects/projects.go
```diff
@@ -0,0 +1,49 @@
...
+	"context"
+	"log/slog"
+
+	"github.com/infisical/api/internal/services/auth"
+
+	genprojects "github.com/infisical/api/internal/server/gen/projects"
+	"github.com/infisical/api/internal/services/permission"
+)
+
+type permissionGetter interface {
```
PrestigePvP, line 12: Should this be private?

backend-go/internal/database/pg/pg.go
```diff
@@ -0,0 +1,167 @@
...
+	"math/rand/v2"
+	"net/url"
+	"strings"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/infisical/api/internal/config"
+)
+
+type DB interface {
```
PrestigePvP, line 28: I think we're missing transactions

backend-go/internal/services/kms/kms.go
```diff
@@ -0,0 +1,866 @@
...
+	var result orgKmsInfo
+	err := row.Scan(&result.ID, &result.KmsDefaultKeyID, &result.KmsEncryptedDataKey)
+	if err != nil {
+		return nil, fmt.Errorf("finding org KMS info: %w", err)
+	}
+	return &result, nil
+}
+
+// findOrgKmsInfoTx returns org KMS info within a transaction.
+func (s *Service) findOrgKmsInfoTx(ctx context.Context, tx pgx.Tx, orgID uuid.UUID) (*orgKmsInfo, error) {
```
PrestigePvP: Why do we have a transaction and non transaction way of doing the same thing? 

I wonder if there's a way to make it so if there is a current transaction we use it, otherwise use the db? Or Can we force the func to take in the db interface and add the transaction to the interface

backend-go/internal/server/api/api.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// SecretManagerHandlers holds handlers for secret manager endpoints.
+type SecretManagerHandlers struct {
+	Secrets gensecrets.Service
+}
+
+// NewRegistry creates a new API registry with all services and handlers initialized.
+func NewRegistry(ctx context.Context, infra *Infra) (*Registry, error) {
+	platformSvc, err := newPlatformServices(ctx, infra)
```
PrestigePvP, line 46: We are accessing a non exported private type?

backend-go/internal/services/actor/actor.go
```diff
@@ -0,0 +1,90 @@
+// Package actor provides shared types and context helpers for actor identity
+// used by both auth and permission packages to avoid import cycles.
+package actor
```
PrestigePvP: Do you think we could avoid this or is this still needed?

backend-go/internal/services/kms/kms.go
```diff
@@ -0,0 +1,866 @@
...
+// Deps holds the dependencies for the KMS shared service.
+type Deps struct {
+	DB     pg.DB
+	HSM    HsmService // nil when HSM is not configured
+	Config *config.Config
+}
+
+// NewService creates a new KMS service.
+// The encryption key is decoded during Start() based on FIPS mode determination.
+func NewService(deps Deps) (*Service, error) {
```
PrestigePvP: This is different than secret.NewService and license.NewService and auth.NewAuthenticator 

Ideally these would all be uniform

backend-go/internal/database/pg/pg.go
akhilmhdh: Good catch

PrestigePvP (dismissed):

sheensantoscapadngan (approved):

PR #6259, opened 2026-04-30
improvement(audit-logs): add audit logs into audit logs

adilsitos:
## Context
Add audit logs in audit log page (view page), and also when checking audit logs details 

## Steps to verify the change
- Enable audit logs in license 
- Check both at project and org level

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/server/routes/v1/organization-router.ts
```diff
@@ -291,6 +303,41 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
...
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId,
+        auditLogId: req.params.auditLogId,
+        auditLogInfo: req.auditLogInfo
+      });
+
+      return { tracked: true };
+    }
+  });
+
```
victorvhs017: I don't think we need this one because when we call GET /audit-logs we return all the logs in the request body, so even if you don't expand the row, you already have access to the content:

Image

This would only add more requests and entries in the audit logs.

victorvhs017 (changes requested):

backend/src/server/routes/v1/organization-router.ts
adilsitos: The goal of this new endpoint is to record if someone opened a specific log, so we know which logs the user opened. Do you think this doesn't make sense?

victorvhs017 (approved):

PR #6273, opened 2026-05-01
chore(upgrade-impact): backfill recent release data

PrestigePvP:
## Summary
- Backfill upgrade impact data for v0.159.16 through v0.159.25 using local OpenAI-backed generation
- Update the upgrade impact index to reference the new release files
- Adjust agent tool schemas to satisfy the OpenAI API strict function schema requirements

## Validation
- npm run type:check
- npm test
- npm run validate

### DIFF
--- a/upgrade-impact/data/releases/v0.159.19.yaml
+++ b/upgrade-impact/data/releases/v0.159.19.yaml
@@ -1,0 +1,66 @@
+version: v0.159.19
+releasedAt: 2026-04-22T03:30:00+05:30
+sourceTag: v0.159.19
+previousTag: v0.159.18
+impactLevel: medium
+summary: Run database migrations. PAM Active Directory data moves to domains,
+  and the public upgrade-path endpoint is removed.
+requiresDbMigration: true
+breakingChanges:
+  - title: Upgrade-path API removed
+    description: The backend upgrade-path service and route were removed. Any
+      automation or UI that calls the built-in upgrade-path endpoint will stop
+      working after upgrade.
+    action: Remove dependencies on the in-app upgrade-path endpoint before
+      upgrading, or replace them with your own external process.
+    confidence: high
+    evidence:
+      - type: pr
+        ref: "#6085"
+        url: https://github.com/Infisical/infisical/pull/6085
+      - type: file
+        ref: backend/src/server/routes/v1/upgrade-path-router.ts
+        path: backend/src/server/routes/v1/upgrade-path-router.ts
+      - type: file
+        ref: backend/upgrade-path.yaml
+        path: backend/upgrade-path.yaml
+dbSchemaChanges:
+  - title: PAM domains migration
+    description: Adds pam_domains and new domain reference columns, then migrates
+      existing Active Directory PAM resources into domains. It also removes
+      adServerResourceId and enforces pam_accounts to reference either a
+      resource or a domain.
+    action: Back up the database and run migrations during maintenance. Validate PAM
+      Active Directory domains, accounts, and related resources after upgrade.
+    confidence: high
+    evidence:
+      - type: file
+        ref: backend/src/db/migrations/20260408062748_pam-domains.ts
+        path: backend/src/db/migrations/20260408062748_pam-domains.ts
+      - type: pr
+        ref: "#5982"
+        url: https://github.com/Infisical/infisical/pull/5982
+  - title: Certificates metadata column
+    description: Adds a nullable externalMetadata JSONB column to certificates. This
+      requires the normal database migration step but should not need manual
+      data changes.
+    action: Run database migrations before starting updated application nodes.
+    confidence: high
+    evidence:
+      - type: file
+        ref: backend/src/db/migrations/20260416231234_add-external-metadata-to-certificates.ts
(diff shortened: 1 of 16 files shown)


upgrade-impact/data/releases/v0.159.19.yaml
```diff
@@ -0,0 +1,66 @@
+version: v0.159.19
+releasedAt: 2026-04-22T03:30:00+05:30
+sourceTag: v0.159.19
+previousTag: v0.159.18
+impactLevel: medium
+summary: Run database migrations. PAM Active Directory data moves to domains,
```
maidul98: i think it would be worth adding at tag for what part of the platform these migration impact. for example, maybe i don't use PAM so i can ignore those. Would be good to add tags for secrets management, certs management, secret scanning, platform, etc

upgrade-impact/data/releases/v0.159.20.yaml
```diff
@@ -0,0 +1,61 @@
+version: v0.159.20
+releasedAt: 2026-04-24T12:14:15+05:30
+sourceTag: v0.159.20
+previousTag: v0.159.19
+impactLevel: medium
+summary: Run database migrations. If you use a reverse proxy, set
```
maidul98: lets ask someone from PAM to help validate this. we shouldn't need folks to do anything to their reverse proxy

upgrade-impact/data/releases/v0.159.23.yaml
```diff
@@ -0,0 +1,49 @@
...
+summary: Adds one optional SMTP setting and a gateway-pools database migration
+  for operators using those capabilities.
+requiresDbMigration: true
+breakingChanges: []
+dbSchemaChanges:
+  - title: Gateway pools tables added
+    description: Database schema adds gateway pool tables. Startup will need the new
+      migration applied before gateway pool features can work.
+    action: Run database migrations during upgrade. Verify migration jobs complete
+      before serving traffic.
```
maidul98: seeing `Verify migration jobs complete before serving traffic.` often. what do we want users to do in this case? we already run only single instance of migration even when there are many containers

upgrade-impact/data/releases/v0.159.24.yaml
```diff
@@ -0,0 +1,42 @@
+version: v0.159.24
+releasedAt: 2026-04-29T16:46:24-03:00
+sourceTag: v0.159.24
+previousTag: v0.159.23
+impactLevel: medium
+summary: Identity auth now enforces TLS certificate verification for OIDC, JWT,
```
maidul98: you can ask @victor to help validate this

upgrade-impact/data/releases/v0.159.25.yaml
```diff
@@ -0,0 +1,57 @@
...
+  and PKI internal CAs add CRL distribution point URLs.
+requiresDbMigration: true
+breakingChanges: []
+dbSchemaChanges:
+  - title: Kubernetes Auth TLS column
+    description: Adds verifyTlsCertificate to identity Kubernetes auth records.
+      Existing records with a stored CA certificate are backfilled to true;
+      others default to false.
+    action: Run database migrations before starting upgraded services. Review
+      Kubernetes Auth identities if you rely on custom CA or skipped TLS
```
maidul98: i don't think they actually need to do anything. tagging in @victorvhs017

upgrade-impact/data/releases/v0.159.25.yaml
```diff
@@ -0,0 +1,57 @@
...
+        path: backend/src/db/schemas/identity-kubernetes-auths.ts
+      - type: file
+        ref: docs/documentation/platform/identities/kubernetes-auth.mdx
+        path: docs/documentation/platform/identities/kubernetes-auth.mdx
+        description: Docs describe auto-enable with CA cert
+  - title: Internal CA CRL URLs
+    description: Adds crlDistributionPointUrls to internal certificate authorities.
+      This supports custom CRL distribution points for private CAs.
+    action: Run database migrations. If you manage PKI automation or direct database
+      integrations, account for the new nullable text array column.
```
maidul98: they shouldn't really need to account for anything other than seeing if the migration ran. 

the less responsibilities we can give the users the better. else they will be wondering what is required of them

maidul98 (commented):
left some comments and pushed up some fixes. the general notion is that we shouldn't say vauge things like: 
- run migration before serving traffic (users might not know what this means in the context of infisical)
- asking them to account for things. we want to be more prescriptive

upgrade-impact/data/releases/v0.159.23.yaml
PrestigePvP: This is pretty much just in case the migrations fail for whatever reason. Make sure they pass before attempting to serve traffic or it will not work

upgrade-impact/data/releases/v0.159.25.yaml
victorvhs017: They don't, we added backwards compatibility for people without CAs

maidul98 (dismissed):

maidul98 (approved):

PR #6320, opened 2026-05-05
improvement: re-vamp sso tab to v3 component and improve ui/ux

scott-ray-wilson:
## Context

This PR re-vamps the SSO settings tab to use v3 components, improves layout, ux and also makes switch sso methods easier

Also includes form validation improvements for sso forms

## Screenshots

CleanShot 2026-05-04 at 17 42 53@2x

## Steps to verify the change

 SSO Tab Revamp — Test Checklist

## Empty / first-time state
- Check that the **Add Provider** button is hidden and replaced by **Add Domain** when email domain verification is enabled but no domains exist
- Check that the **Add Provider** button shows a "Email domain verification required" tooltip and stays disabled when domains exist but none are verified
- Check that the **Add Provider** button is enabled once at least one domain is `verified`
- Check that when no providers are available (server config disables all login methods), the empty state shows "Single Sign-On (SSO) has been disabled"

## Provider chooser dialog
- Check that clicking **Add Provider** opens the chooser with SAML / OIDC / LDAP options
- Check that LDAP option only renders if the LDAP login method is enabled in server config
- Check that **Continue** is disabled until a provider is selected
- Check that selecting an upgrade-gated provider (e.g. OIDC on free plan) opens the upgrade modal instead of the config modal
- Check that closing/cancelling the chooser resets the selected provider

## Switch provider flow
- Check that **Switch Method** opens the chooser excluding the current provider
- Check that confirming the switch shows the destructive AlertDialog warning that members can't sign in until the new IDP is configured
- Check that confirming clears the prior IDP config (active=false, fields blanked) and opens the new provider's config modal
- Check that cancelling the AlertDialog leaves the existing IDP config untouched

## Email domains
- Check that adding a domain immediately opens the **Verify Email Domain** dialog with the new domain's TXT record
- Check that the copy buttons for record name and value populate the clipboard
- Check that the verify action transitions the domain row's badge from `pending` to `verified`

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


frontend/src/components/v3/generic/Tooltip/Tooltip.stories.tsx
```diff
@@ -16,6 +16,13 @@ import { Tooltip, TooltipContent, TooltipTrigger } from "./Tooltip";
  * with `asChild` to forward the trigger props onto a `Button`, `IconButton`,
  * or any focusable element — keep tooltip content to a single short sentence,
  * and never put interactive controls inside it.
+ *
+ * **Prefer `HoverCard`** when the hint needs rich content — multi-paragraph
```
PrestigePvP, line 20: I feel like this isn't bad info to be stored inside a Claude.md at the components level

frontend/src/hooks/api/ssoConfig/queries.tsx
```diff
@@ -59,6 +61,8 @@ export const useCreateSSOConfig = () => {
     },
     onSuccess(_, dto) {
       queryClient.invalidateQueries({ queryKey: ssoConfigKeys.getSSOConfig(dto.organizationId) });
+      queryClient.invalidateQueries({ queryKey: ldapConfigKeys.getLDAPConfig(dto.organizationId) });
```
PrestigePvP, line 64: Not really something critical to this review, but going forward do we want to do more invalidation or optimistic updates with rollback? 

I could see a responsiveness AND application benefit to the optimistic update route

PrestigePvP (approved):

frontend/src/components/v3/generic/Tooltip/Tooltip.stories.tsx
scott-ray-wilson: great point will make a note to toss it in there!

frontend/src/hooks/api/ssoConfig/queries.tsx
scott-ray-wilson: I think optimistic updates would be ideal - would love to push that, once spoke about it with @akhilmhdh and we both have our eye on https://tanstack.com/db/latest - though it can be done with query; just a bit more tedious to scaffold

PR #6325, opened 2026-05-05
fix(secret-references): resolve local keys whose names contain dots

jaydeep-pipaliya:
## Problem

Secret references fail silently when a secret key name contains a dot (e.g. `MY.SECRET`).

Using `${MY.SECRET}` in a secret value does not resolve to the local secret named `MY.SECRET`. Instead, the parser splits on `.` and treats `MY` as an environment slug, performing a cross-environment lookup for key `SECRET` in a non-existent environment `MY`. The reference expands to an empty string.

Reported in #5962.

## Root Cause

`recursivelyExpandSecret` in `secret-reference-fns.ts` branches purely on whether the interpolation key contains a dot:

```ts
const entities = interpolationKey.trim().split(".");
if (entities.length === 1) {
  // local lookup
} else {
  // always cross-env lookup — first segment assumed to be environment slug
}
```

`fetchSecret` masked the failure: when `folderDAL.findBySecretPath` returned `null` for a non-existent environment, it returned `{ value: "", tags: [] }` — identical to "key found with empty value" — so the caller could not tell the difference.

## Fix

**`fetchSecret`** now returns `null` when the target environment folder does not exist, cleanly distinguishing "environment not found" from "key exists with empty value".

**Cross-env branch**: when `fetchSecret` returns `null` (environment not found), the code falls back to a local key lookup using the full interpolation key as the key name. This means:

- `${MY.SECRET}` where `MY` is **not** an environment → resolves as local key `MY.SECRET` ✓
- `${prod.MY_SECRET}` where `prod` **is** an environment → resolves cross-env as before ✓
- `${prod.MY_SECRET}` where `prod` is an environment but `MY_SECRET` doesn't exist there → expands to empty string as before ✓

## Tests

Added 5 unit tests covering:
1. Simple local reference (no dots)
2. Local key with a single dot — the regression case
3. Local key with multiple dots
4. Cross-environment reference to an existing environment (unchanged behaviour)
5. Reference to a non-existent key in a non-existent environment (expands to the raw syntax unchanged)

```
✓ resolves a simple local reference
✓ resolves a local key containing a dot when no matching environment exists
✓ resolves a local key with multiple dots when no matching environment exists
✓ resolves a cross-environment reference when the environment exists
✓ leaves the interpolation syntax unchanged when neither cross-env nor local key is found
```

## Limitation

`getAllSecretReferences` (used for dependency tracking / cache invalidation) still classifies `${MY.SECRET}` as a cross-environment reference to env `MY`. This means a secret with a dotted local key may be over-eagerly invalidated when environment `MY` changes, but correctness of secret expansion is not affected. Fixing `getAllSecretReferences` requires access to the project's environment list at parse time and is deferred.

### DIFF
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -353,6 +353,8 @@ export const ALICLOUD_AUTH = {
     SignatureVersion: "The signature version. For STS GetCallerIdentity, this should be '1.0'.",
     SignatureNonce: "A unique random string to prevent replay attacks.",
     Signature: "The signature string calculated based on the request parameters and AccessKey Secret.",
+    SecurityToken:
+      "The Alibaba Cloud STS SecurityToken. Required only when authenticating with STS temporary credentials.",
     organizationSlug: IDENTITY_AUTH_SUB_ORGANIZATION_NAME
   },
   ATTACH: {
--- a/backend/src/server/routes/v1/identity-alicloud-auth-router.ts
+++ b/backend/src/server/routes/v1/identity-alicloud-auth-router.ts
@@ -73,7 +73,21 @@ export const registerIdentityAliCloudAuthRouter = async (server: FastifyZodProvi
           .refine((val) => new RE2("^[A-Za-z0-9+/=]+$").test(val), {
             message: "Signature must be base64 characters"
           })
-          .describe(ALICLOUD_AUTH.LOGIN.Signature)
+          .describe(ALICLOUD_AUTH.LOGIN.Signature),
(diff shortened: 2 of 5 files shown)


jaydeep-pipaliya:
Closing — this PR ended up bundling two unrelated fixes. They're now split into focused PRs:

- **#6290** — `fix(identity-alicloud-auth): accept STS-prefixed AccessKeyIds` (with SecurityToken forwarding)
- **#6492** — `fix(secret-references): resolve dotted local secret names before env.path.key` (cleaner test suite covering nested-vs-local tie-breaker)

Please review there instead. Sorry for the noise.

Issue #6280, opened 2026-05-02
Project Machine Identities: Name Sort Not Working


kevincam3:
### Product

Platform/Access controls/Other

### Describe the bug

When adding a new Machine Identity. The page that list all of them has an "arrow" that appears to be to sort them by name. Toggling that arrow doesn't change the sort.

### To Reproduce

1. Go to Access Contol
2. Go to Machine Identities
3. Click on the "Name" column heading 

### Expected behavior

Sort the list by name

### Screenshots

<img width="1920" height="919" alt="Image" src="https://github.com/user-attachments/assets/be3a9ea3-557d-4210-8958-825a00910d64" />

### Deployment Type

Self-hosted

whyujjwal:
Took a look at this one and found the likely root cause.

The frontend table is already toggling `orderBy` / `orderDirection` when you click the Name header, but the project memberships identities endpoint currently ignores those params. The v1 route only accepts `offset`, `limit`, `identityName`, and `roles`, and the membership-identity service/DAL path doesn't thread sort options through either.

So the sort chevron updates in the UI, but the backend keeps returning the same order.

I'm happy to pick this up and wire sorting through end-to-end with a regression test.

PR #6339, opened 2026-05-06
fix(identities): wire sort parameters for project machine identity list

jaydeep-pipaliya:
## Problem

The **Project Machine Identities** list does not honour the sort order selected in the UI. Identities always appear in an undefined (database insert) order, so sorting by name has no effect.

Reported in #6280.

## Root Cause

Three layers were each missing a piece:

| Layer | Issue |
|-------|-------|
| Route (`project-identity-router.ts`) | `orderBy` / `orderDirection` were not declared in the querystring schema — the API silently dropped them |
| Service (`identity-service.ts`) | `listIdentities` forwarded `search`, `offset`, `limit` to the DAL but not sort fields, even though `TListIdentityV2DTO.data` already defined them |
| DAL (`identity-dal.ts`) | `listIdentities` accepted no sort parameters and built no `ORDER BY` clause |

## Fix

**`identity-dal.ts`** — extend the filter type to accept `orderBy` and `orderDirection`, and apply `ORDER BY identity.name <direction>` to the data query before applying limit/offset.

**`identity-service.ts`** — pass `dto.data.orderBy` and `dto.data.orderDirection` through to the DAL call.

**`project-identity-router.ts`** — declare `orderBy` (`IdentityOrderBy` enum, default `name`) and `orderDirection` (`asc`/`desc`, default `asc`) as optional query parameters, and forward them into the service DTO.

The changes are additive and backward-compatible: existing callers that omit the new parameters continue to receive identities sorted by name ascending (a stable, predictable default).

### DIFF
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -353,6 +353,8 @@ export const ALICLOUD_AUTH = {
     SignatureVersion: "The signature version. For STS GetCallerIdentity, this should be '1.0'.",
     SignatureNonce: "A unique random string to prevent replay attacks.",
     Signature: "The signature string calculated based on the request parameters and AccessKey Secret.",
+    SecurityToken:
+      "The Alibaba Cloud STS SecurityToken. Required only when authenticating with STS temporary credentials.",
     organizationSlug: IDENTITY_AUTH_SUB_ORGANIZATION_NAME
   },
   ATTACH: {
--- a/backend/src/server/routes/v1/identity-alicloud-auth-router.ts
+++ b/backend/src/server/routes/v1/identity-alicloud-auth-router.ts
@@ -73,7 +73,21 @@ export const registerIdentityAliCloudAuthRouter = async (server: FastifyZodProvi
           .refine((val) => new RE2("^[A-Za-z0-9+/=]+$").test(val), {
             message: "Signature must be base64 characters"
           })
-          .describe(ALICLOUD_AUTH.LOGIN.Signature)
+          .describe(ALICLOUD_AUTH.LOGIN.Signature),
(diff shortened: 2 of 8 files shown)


backend/src/lib/api-docs/constants.ts
```diff
@@ -354,6 +353,8 @@ export const ALICLOUD_AUTH = {
     SignatureVersion: "The signature version. For STS GetCallerIdentity, this should be '1.0'.",
     SignatureNonce: "A unique random string to prevent replay attacks.",
     Signature: "The signature string calculated based on the request parameters and AccessKey Secret.",
+    SecurityToken:
```
akhilmhdh: Can you share why sort parameter has this other changes? Can you please remove this and only keep the changes required for the PR. 

I assume this is by an accident

jaydeep-pipaliya:
Hey @akhilmhdh, good catch — you were right, those changes weren't meant to be on this branch. They came from two other branches I had open at the time and got accidentally folded in. I've rebased onto current `main` so the PR now only contains the three files that actually wire the sort:

- `backend/src/server/routes/v1/project-identity-router.ts`
- `backend/src/services/identity-v2/identity-dal.ts`
- `backend/src/services/identity-v2/identity-service.ts`

While I was in there I also addressed the Codex P2: the shared `IdentityOrderBy` enum has both `Name` and `Role`, but the v2 identity DAL only knows how to ORDER BY `identity.name`. Exposing `role` on this endpoint would have silently fallen back to name sort, which is the misleading-contract case Codex flagged. I scoped the route's `orderBy` Zod enum to just `Name` so the API only advertises what the DAL implements — left a comment so whoever wires role sorting later knows where to widen it.

Let me know if you'd prefer I add the role JOIN in this PR instead; happy either way.

PR #6352, opened 2026-05-06
fix(honey-tokens): validate signing key, stack name and region

mathnogueira:
## Context

This prevents command injection into the field values by:

1. Validating the input
2. wrapping the command parameters in quotes

I didn't change the region input to be a `select` because it breaks due to the `accordion`. Will reproduce it in isolation and send to Scott.

## Screenshots

image

## Steps to verify the change

* Go to Settings  > Product Settings
* Try to save invalid values, it should fail

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/pages/organization/SettingsPage/components/OrgProductSettingsTab/HoneyTokenModal.tsx
+++ b/frontend/src/pages/organization/SettingsPage/components/OrgProductSettingsTab/HoneyTokenModal.tsx
@@ -55,7 +55,7 @@ const schema = z.object({
     .string()
     .min(1, "Webhook Signing Key is required")
     .regex(/^[a-fA-F0-9]+$/, "Signing key must be a hex string"),
-  stackName: slugSchema({max: 128, field: 'stackName'}),
+  stackName: slugSchema({ max: 128, field: "stackName" }),
   awsRegion: z
     .string()
     .trim()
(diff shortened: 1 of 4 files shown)


frontend/src/pages/organization/SettingsPage/components/OrgProductSettingsTab/HoneyTokenModal.tsx
```diff
@@ -43,11 +45,22 @@ const DEFAULT_STACK_NAME = "infisical-honey-tokens";
...
 const schema = z.object({
   connectionId: z.string().min(1, "AWS Connection is required"),
-  webhookSigningKey: z.string().min(1, "Webhook Signing Key is required"),
-  stackName: z.string().trim().min(1, "Stack name is required").max(128),
-  awsRegion: z.string().trim().min(1, "AWS region is required")
+  webhookSigningKey: z
+    .string()
+    .min(1, "Webhook Signing Key is required")
+    .regex(/^[a-fA-F0-9]+$/, "Signing key must be a hex string"),
+  stackName: slugSchema({ max: 128, field: "stackName" }),
```
adilsitos, line 58: isn't this more strict than necessary? From what I know, slugSchema only allow [a-z0-9-] but (from what I searched), stack formation allow uppercase letters. Can

adilsitos (approved):

frontend/src/pages/organization/SettingsPage/components/OrgProductSettingsTab/HoneyTokenModal.tsx
mathnogueira: The stack name is used as prefix for all the resources created by the stack, so if any resource doesn't allow uppercase characters, it will break the stack. One example is S3 buckets which don't allow uppercase characters in its name.

So allowing uppercase would break the stack deployment.

PR #6355, opened 2026-05-06
feature: announcements for cloud and self-hosted

scott-ray-wilson:
## Context

This PR adds a new announcement feature that pulls feature announcements form contentful (cloud) or bakes them in to self-hosted images (most recent 5)

## Screenshots

CleanShot 2026-05-06 at 14 22 44@2x

## Steps to verify the change

PR Reviewer Test Checklist

### Database migration
- `npm run migration:latest-dev` applies `20260506120000_users-last-seen-announcement.ts` cleanly; users table gets nullable `lastSeenAnnouncementId text`.
- `npm run migration:rollback-dev` reverses it cleanly (column dropped).
- Re-running migration is a no-op (it checks `hasColumn` first).
- Existing users have `lastSeenAnnouncementId = NULL` — no data backfill needed.

### Live (cloud) Contentful path
- With valid `CONTENTFUL_SPACE_ID` / `CONTENTFUL_DELIVERY_TOKEN` / `CONTENTFUL_ENVIRONMENT`, log in as a user older than 7 days → modal pops on `/organizations/$orgId/projects` with image, title, body, formatted date, and CTA link.
- CTA opens link in a new tab with `rel="noopener noreferrer"`.
- Dismiss via "Got it", X, click-outside, or Escape — all four paths fire `POST /api/v1/announcement/seen` and the unread dot disappears.
- Hard reload after dismiss → modal does NOT re-pop (lastSeenAnnouncementId persisted in DB).
- Pager shows "1 / N" with chevrons when ≥2 announcements; hidden when only 1.
- Megaphone button in navbar always re-opens the modal at index 0.
- Within a session, no refetch happens (staleTime: Infinity); confirm via Network tab that `/recent` is called once on initial load only.
- Open a fresh tab → /recent fires again (per-session cache, by design).

### Backend Redis cache
- First `/recent` request after expiry: backend calls Contentful, populates Redis key `announcements:recent`.
- Subsequent requests within 1 hour: Contentful is NOT called (verify in network logs / Contentful API stats).
- Stop Redis → `/recent` still works, just hits Contentful every time (graceful degradation, no 500s).

### New-user grace period
-  Sign up a brand-new account; `/recent` returns `{ announcements: [], lastSeenAnnouncementId: null }`.
-  No modal pops; no megaphone button shows.
-  After manually backdating the user's `createdAt` > 7 days, hard reload → announcements appear.

### Bundled (self-hosted) path
- `cd backend && CONTENTFUL_SPACE_ID=... CONTENTFUL_DELIVERY_TOKEN=... CONTENTFUL_ENVIRONMENT=dev npm run bake:announcements` writes `dist/announcement-assets/announcements.json` and `dist/announcement-assets/images/<sha>.<ext>`.
- Baked JSON has `imageUrl: "/api/v1/announcement/assets/<sha>.png"` (NOT `images.ctfassets.net/...`).
- Restart backend with bundled file present → log line `Loaded N bundled announcement(s) from .../announcements.json` appears.
- With Contentful env vars unset, `/recent` still returns the bundled announcements.
- `GET /api/v1/announcement/assets/<sha>.png` returns the image with `Cache-Control: public, max-age=31536000, immutable` and the right MIME type.
- `GET /api/v1/announcement/assets/../../../etc/passwd` returns 404 (regex guard).
- `GET /api/v1/announcement/assets/nonexistent.png` returns 404.
-  Asset route works without authentication (so `` tags load).
-  Run with `--network=<isolated-net>` (no internet egress) → modal still works end-to-end.

### Docker builds
- `docker build -f Dockerfile.standalone-infisical --build-arg CONTENTFUL_SPACE_ID=... --build-arg CONTENTFUL_DELIVERY_TOKEN=... --build-arg CONTENTFUL_ENVIRONMENT=dev -t test .` succeeds and build log shows `[bake-announcements] wrote N announcement(s) and M image(s)`.
- Same for `Dockerfile.fips.standalone-infisical`.
- Build WITHOUT Contentful build-args still succeeds; bake step logs the "skipping" warning and no JSON is written. Image runs in live-fetch mode.
- Inspect built image: `docker run --rm --entrypoint sh <image> -c "ls /backend/dist/announcement-assets/ /backend/dist/announcement-assets/images/"` shows the baked artifacts.

### Failure modes
- `ANNOUNCEMENTS_ENABLED=false` → `/recent` returns empty list, no modal, no navbar button. No outbound calls. No Redis write.
-  Bogus `CONTENTFUL_SPACE_ID` at runtime → backend logs warn ONCE, `/recent` returns empty list, no retry storm.
-  Contentful 5xx outage → same graceful behavior; subsequent requests served from Redis cache until expiry.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/scripts/bake-announcements.ts
```diff
@@ -0,0 +1,159 @@
...
+    console.warn(`[bake-announcements] failed to download ${url}: ${res.status}`);
+    return null;
+  }
+  const buf = Buffer.from(await res.arrayBuffer());
+  const hash = crypto.createHash("sha256").update(buf).digest("hex").slice(0, 16);
+  const filename = `${hash}${ext}`;
+  await fs.writeFile(path.join(IMAGE_DIR, filename), buf);
+  console.log(`[bake-announcements]   asset ${assetId} -> ${filename} (${buf.length} bytes)`);
+  return { filename, bytes: buf.length };
+}
```
victorvhs017, line 98: Question: Do we control where the assets are coming from? Like, is it always from Contentful or our own S3 bucket, for example?

If we do, we could also whitelist the domains here for extra security. What do you think?

backend/src/server/routes/v1/announcement-router.ts
```diff
@@ -0,0 +1,122 @@
...
+      // this is to appease the type derivation below
+      if (req.auth.authMode !== AuthMode.JWT) {
+        throw new UnauthorizedError({ message: "This endpoint can only be accessed by users" });
+      }
+      return server.services.announcement.markAnnouncementSeen({
+        userId: req.auth.userId,
+        announcementId: req.body.announcementId
+      });
+    }
+  });
```
victorvhs017, line 86: Trying to figure out if we need audit logs here for pattern consistency, and I think we should. 

In the GETs too, I think the idea is to cover everything now.

backend/src/services/announcement/announcement-service.ts
```diff
@@ -0,0 +1,192 @@
...
+  }): Promise<{ announcements: TAnnouncement[]; lastSeenAnnouncementId: string | null }> => {
+    const user = await userDAL.findById(userId);
+    const lastSeenAnnouncementId = user?.lastSeenAnnouncementId ?? null;
+
+    if (user?.createdAt && Date.now() - new Date(user.createdAt).getTime() < NEW_USER_GRACE_PERIOD_MS) {
+      return { announcements: [], lastSeenAnnouncementId };
+    }
+
+    const announcements = await getAnnouncements();
+    return { announcements, lastSeenAnnouncementId };
```
victorvhs017, line 179: I think we could return only the unseen announcements here, instead of both { announcements, lastSeenAnnouncementId }

backend/src/server/routes/v1/announcement-router.ts
scott-ray-wilson: this isn't associated with an org/project though so we can't audit it to anything? I think we're good here to not include

backend/src/services/announcement/announcement-service.ts
scott-ray-wilson: we allow them to view previous ones from the nav bar though (incase they want to review them or accidentally closed them - it's only the most recent 5 so not too much)

victorvhs017 (approved):

PR #6359, opened 2026-05-07
feat(permission): add project actions for creating and requesting access

victorvhs017:
## Context

Splits org **project** permissions into **`create`** and **`request-access`**, and enforces them in the API and UI.

- **Backend:** Introduces `OrgPermissionProjectActions` (`Create`, `RequestAccess`). Project creation checks `Create`; `requestProjectAccess` requires `RequestAccess`. Admin and member roles include both by default.
- **Frontend:** Uses the same actions for “create project” vs “request access,” hides the **All Projects** list/toggle unless the user can request access, and extends the org role form to toggle both permissions.

## Screenshots

image

## Steps to verify the change

1. As admin/member: confirm **My Projects** / **All Projects** toggle and **Request Access** behave as before when `Request Access` is allowed.
2. As a role **without** `request-access`: confirm **All Projects** is unavailable and requesting access fails at the API where expected.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/project/project-service.ts
```diff
@@ -2320,15 +2324,19 @@ export const projectServiceFactory = ({
...
       actor: permission.type,
       actorId: permission.id,
       orgId: permission.orgId,
       actorAuthMethod: permission.authMethod,
       actorOrgId: permission.orgId,
       scope: OrganizationActionScope.Any
     });
+    ForbiddenError.from(orgPermission).throwUnlessCan(
+      OrgPermissionProjectActions.RequestAccess,
+      OrgPermissionSubjects.Project
```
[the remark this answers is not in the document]
victorvhs017: That's ok, we want to only block the UI because it doesn't make sense to show the projects if you can't do anything.

backend/src/services/project/project-service.ts
```diff
@@ -2320,15 +2324,19 @@ export const projectServiceFactory = ({
...
+    );
 
     const projectMember = await permissionService
       .getProjectPermission({
```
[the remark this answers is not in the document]
victorvhs017: That's fine, as long as member and admin roles have it.

frontend/src/pages/organization/ProjectsPage/components/ProjectListToggle.tsx
```diff
@@ -25,18 +26,20 @@ export const ProjectListToggle = ({ value, onChange }: Props) => {
...
+            onChange(ProjectListView.AllProjects);
+          }}
+          size="xs"
+          className={`${
+            value === ProjectListView.AllProjects ? "bg-mineshaft-500" : "bg-transparent"
+          } min-w-[2.4rem] rounded border-none hover:bg-mineshaft-600`}
+        >
+          All Projects
+        </Button>
+      )}
```
scott-ray-wilson: I wonder if instead of just hiding all projects we should just hide the button group? this doesn't functionally do anything anymore

Image

scott-ray-wilson (commented):
one comment and just verifying we're willingly introducing a breaking change? No custom org roles will be able to request project access when previously they could? They won't even really know why, they just won't see all projects anymore - feels like a rug pull unless we back fill all custom roles with this permission

frontend/src/pages/organization/ProjectsPage/components/MyProjectView.tsx
```diff
@@ -363,7 +370,9 @@ export const MyProjectView = ({
   return (
     <div>
       <div className="flex w-full flex-row flex-wrap gap-2 md:flex-nowrap md:gap-0">
-        <ProjectListToggle value={projectListView} onChange={onProjectListViewChange} />
+        {showAllProjects && (
+          <ProjectListToggle value={projectListView} onChange={onProjectListViewChange} />
+        )}
         <Input
           className="h-[2.3rem] bg-mineshaft-800 text-sm placeholder-mineshaft-50/60 duration-200 focus:bg-mineshaft-700/80"
           containerClassName="w-full ml-2"
```
scott-ray-wilson: nit: we'd want to remove the ml-2 when showAllProjects is false

backend/src/db/migrations/20260508140759_backfill-org-custom-role-project-request-access.ts
```diff
@@ -0,0 +1,102 @@
...
+
+const PROJECT_SUBJECT = "project";
+const REQUEST_ACCESS_ACTION = "request-access";
+
+const CHUNK_SIZE = 1000;
+
+export async function up(knex: Knex): Promise<void> {
+  // Org-level custom roles are rows in `roles` where projectId IS NULL.
+  // Built-in slugs (admin/member/no-access) are handled in code and not stored here.
+  const customRoles = await knex(TableName.Role).whereNull("projectId").select("*");
```
scott-ray-wilson: this get's all rows in the db - do we know how many custom org roles are in US? maybe we should chunk read and write together?

backend/src/db/migrations/20260508140759_backfill-org-custom-role-project-request-access.ts
```diff
@@ -0,0 +1,102 @@
...
+    .filter(Boolean) as typeof customRoles;
+
+  for (let i = 0; i < toUpdate.length; i += CHUNK_SIZE) {
+    const chunk = toUpdate.slice(i, i + CHUNK_SIZE);
+    await knex(TableName.Role).insert(chunk).onConflict("id").merge();
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
+  const customRoles = await knex(TableName.Role).whereNull("projectId").select("*");
```
scott-ray-wilson: same concern

backend/src/db/migrations/20260508140759_backfill-org-custom-role-project-request-access.ts
```diff
@@ -0,0 +1,102 @@
...
+      }[];
+
+      // Skip if request-access on project is already present
+      const alreadyHasPermission = rules.some(
+        (rule) =>
+          !rule.inverted &&
+          rule.subject === PROJECT_SUBJECT &&
+          (rule.action === REQUEST_ACCESS_ACTION ||
+            (Array.isArray(rule.action) && rule.action.includes(REQUEST_ACCESS_ACTION)))
+      );
```
scott-ray-wilson: I had claude write a migration test for all scenarios and technically we don't roll back so this should be fine but if we did this would be problematic as it would add the role twice (unpack permissions normalizes so the check would say it doesn't have it - reproduced here)

should be a simple fix to handle

```
// `unpackRules` normalises `subject` and `action` to array form, so accept either shape.
const includes = (value: string | string[] | undefined, target: string) =>
  value === target || (Array.isArray(value) && value.includes(target));
```

[backfill-org-custom-role-request-access.spec.ts](https://github.com/user-attachments/files/27535888/backfill-org-custom-role-request-access.spec.ts)

fixtures:
```
import { packRules } from "@casl/ability/extra";
import { Knex } from "knex";

import { TableName } from "../schemas";
import { seedData1 } from "../seed-data";

export const MIGRATION_FIXTURE_SLUG_PREFIX = "mig_test_request_access_";

type Rule = { action: string | string[]; subject: string; inverted?: boolean };

type Fixture = {
  slug: string;
  name: string;
  rules: Rule[];
  projectId?: string;
};

export const MIGRATION_FIXTURES: Fixture[] = [
  { slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}empty`, name: "empty permissions", rules: [] },
  {
    slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}no_project_rule`,
    name: "rules but none for project",
    rules: [{ action: "read", subject: "member" }]
  },
  {
    slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}create_string`,
    name: "single create on project",
    rules: [{ action: "create", subject: "project" }]
  },
  {
    slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}create_array`,
    name: "array action on project",
    rules: [{ action: ["create", "delete"], subject: "project" }]
  },
  {
    slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}already_request_access_string`,
    name: "request-access already present (string)",
    rules: [{ action: "request-access", subject: "project" }]
  },
  {
    slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}already_request_access_array`,
    name: "request-access already present (array)",
    rules: [{ action: ["create", "request-access"], subject: "project" }]
  },
  {
    slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}inverted_request_access`,
    name: "inverted request-access rule",
    rules: [{ action: "request-access", subject: "project", inverted: true }]
  },
  {
    slug: `${MIGRATION_FIXTURE_SLUG_PREFIX}project_scoped`,
    name: "project-scoped role with project create — must be ignored",
    rules: [{ action: "create", subject: "project" }],
    projectId: seedData1.project.id
  }
];

export async function seed(knex: Knex): Promise<void> {
  await knex(TableName.Role).where("slug", "like", `${MIGRATION_FIXTURE_SLUG_PREFIX}%`).del();

  await knex(TableName.Role).insert(
    MIGRATION_FIXTURES.map((f) => ({
      slug: f.slug,
      name: f.name,
      orgId: seedData1.organization.id,
      projectId: f.projectId ?? null,
      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
      // @ts-ignore packRules type mismatch
      permissions: JSON.stringify(packRules(f.rules))
    }))
  );
}

```

scott-ray-wilson (changes requested):
few comments

backend/src/db/migrations/20260508140759_backfill-org-custom-role-project-request-access.ts
victorvhs017: Good call, I checked, and there are fewer than 500 custom roles that fall in this category. I'll do it in one pass

scott-ray-wilson (approved):
LGTM, verified migration

PR #6388, opened 2026-05-08
feat(secrets): allow duplication of secrets between environments

adilsitos:
## Context
This allows secrets to be deep copied between environments. 

## Screenshots

image

## Steps to verify the change
- Create some secrets
- Try to duplicate them between environments
 
## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretEditTableRow.tsx
```diff
@@ -1655,6 +1657,39 @@ export const SecretEditTableRow = ({
...
                   </TooltipContent>
                 </Tooltip>
+                <ProjectPermissionCan
+                  I={ProjectPermissionActions.Create}
+                  a={subject(ProjectPermissionSub.Secrets, {
+                    environment,
+                    secretPath,
+                    secretName,
+                    secretTags: ["*"]
+                  })}
```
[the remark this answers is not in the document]
adilsitos: I feel like if we create this logic, this can be a little confusing, what do you think? @scott-ray-wilson

backend/src/services/secret/secret-service.ts
```diff
@@ -3485,6 +3491,212 @@ export const secretServiceFactory = ({
...
+    const tagIdsToCopy = attributesToCopy.tags ? (sourceSecret.tags?.map((t) => t.id) ?? []) : undefined;
+    const skipMultilineEncodingToCopy = attributesToCopy.skipMultilineEncoding
+      ? Boolean(sourceSecret.skipMultilineEncoding)
+      : false;
+
+    const secretPayload = {
+      secretValue: valueToCopy,
+      secretComment: commentToCopy,
+      skipMultilineEncoding: skipMultilineEncodingToCopy,
+      tagIds: tagIdsToCopy,
```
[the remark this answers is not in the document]
adilsitos: Talked with this on the PR channels.

backend/src/services/secret/secret-service.ts
```diff
@@ -3485,6 +3491,212 @@ export const secretServiceFactory = ({
...
+
+    const { decryptor: secretManagerDecryptor } = await kmsService.createCipherPairWithDataKey({
+      type: KmsDataKey.SecretManager,
+      projectId
+    });
+
+    const valueToCopy =
+      attributesToCopy.value && sourceSecret.encryptedValue
+        ? secretManagerDecryptor({ cipherTextBlob: sourceSecret.encryptedValue }).toString()
+        : "";
```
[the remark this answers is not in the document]
adilsitos: Talked with this on the PR channels.

backend/src/services/secret/secret-service.ts
```diff
@@ -3485,6 +3491,212 @@ export const secretServiceFactory = ({
...
+          value: el.encryptedValue
+            ? secretManagerDecryptor({ cipherTextBlob: el.encryptedValue }).toString()
+            : (el.value ?? ""),
+          isEncrypted: Boolean(el.encryptedValue)
+        })) ?? [])
+      : undefined;
+    const tagIdsToCopy = attributesToCopy.tags ? (sourceSecret.tags?.map((t) => t.id) ?? []) : undefined;
+    const skipMultilineEncodingToCopy = attributesToCopy.skipMultilineEncoding
+      ? Boolean(sourceSecret.skipMultilineEncoding)
+      : false;
```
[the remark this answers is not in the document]
adilsitos: Talked with this on the PR channels.

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretEditTableRow.tsx
```diff
@@ -1655,6 +1657,39 @@ export const SecretEditTableRow = ({
...
+                    <Tooltip
+                      open={isCreatable || !isAllowed ? undefined : false}
+                      disableHoverableContent
+                    >
+                      <TooltipTrigger className="block w-full">
+                        <DropdownMenuItem
+                          onClick={() => handlePopUpOpen("duplicateSecret")}
+                          isDisabled={isCreatable || !secretId || !isAllowed}
+                        >
+                          <CopyIcon />
```
scott-ray-wilson: nit: to differentiate it from copy value let's use https://lucide.dev/icons/copy-plus

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
```diff
@@ -0,0 +1,408 @@
...
+          );
+        }}
+      />
+
+      <Controller
+        control={control}
+        name="include"
+        render={({ field: { value, onChange } }) => (
+          <Field className="mt-4">
+            <FieldLabel>Include values</FieldLabel>
```
scott-ray-wilson: let's say include properties (since value refers to a property it's a little confusing)

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
```diff
@@ -0,0 +1,408 @@
...
+                      <Checkbox
+                        id={id}
+                        variant="project"
+                        isChecked={value[opt.key]}
+                        onCheckedChange={() => onChange({ ...value, [opt.key]: !value[opt.key] })}
+                        className="mt-0.5"
+                      />
+                      <div className="flex flex-col gap-0.5">
+                        <span className="text-sm text-foreground">{opt.label}</span>
+                        <span className="text-muted-foreground text-xs">{opt.description}</span>
```
scott-ray-wilson: we don't have text-muted-foreground -> either text-muted or text-accent

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
```diff
@@ -0,0 +1,408 @@
...
+                                )}
+                              />
+                              <span className="truncate">{env.name}</span>
+                            </CommandItem>
+                          ))}
+                        </CommandGroup>
+                      </CommandList>
+                    </Command>
+                  </PopoverContent>
+                </Popover>
```
scott-ray-wilson: let's use a v3 `<FilterableSelect />` so they can see which envs are selected and click the tags to remove them

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
```diff
@@ -0,0 +1,408 @@
...
+          <Button variant="outline">Cancel</Button>
+        </DialogClose>
+        <Button
+          type="submit"
+          variant="project"
+          isDisabled={selectedEnvIds.length === 0 || !secretId || isSubmitting}
+          isPending={isSubmitting}
+        >
+          <CopyIcon />
+          {actionLabel}
```
scott-ray-wilson: since we'll show the values in the filterable select we can just make this say duplicate (and use the plus copy icon)

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
```diff
@@ -0,0 +1,408 @@
...
+};
+
+export const DuplicateSecretModal = ({ isOpen, onOpenChange, ...props }: Props) => (
+  <Dialog open={isOpen} onOpenChange={onOpenChange}>
+    <DialogContent className="max-w-xl">
+      <DialogHeader>
+        <div className="flex items-center gap-2">
+          <span className="flex size-7 items-center justify-center rounded-md border border-project/25 bg-project/10 text-project">
+            <CopyIcon className="size-3.5" />
+          </span>
```
scott-ray-wilson: nit: let's  remove the icon - we currently just do text titles unless they are third party integrations

backend/src/server/routes/v4/secret-router.ts
```diff
@@ -898,6 +908,102 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
     }
   });
 
+  server.route({
+    method: "POST",
+    url: "/duplicate",
```
scott-ray-wilson, line 913: for secrets router we usually include the secret name in the url as the identifier, rather than the ID in the payload - I would think we want to keep that consistent here but I'd check with @varonix0

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
```diff
@@ -0,0 +1,408 @@
...
+          <div className="flex items-center justify-between rounded-md border border-border bg-container px-3 py-2">
+            <div className="flex min-w-0 items-center gap-2 text-sm text-foreground">
+              <KeyIcon className="text-muted-foreground size-3.5 shrink-0" />
+              {secretPath !== "/" && (
+                <span className="text-muted-foreground/60 truncate">{secretPath}</span>
+              )}
+              <span className="truncate">{secretName}</span>
+            </div>
+            <Badge variant="info">{sourceEnvironment.name}</Badge>
+          </div>
```
scott-ray-wilson: we need to handle truncation 

Image

also I think we might be trying to squeeze too much into a single row (imagine long folder path and long env name) - we probably want to do a "field" style display for each secret name, env and folder (probably want to move this to a sheet since it's going to get taller)

Image

backend/src/server/routes/v4/secret-router.ts
```diff
@@ -898,6 +908,102 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
...
+        projectId: z.string().trim().describe(RAW_SECRETS.DUPLICATE_SECRET.projectId),
+        sourceEnvironment: z.string().trim().describe(RAW_SECRETS.DUPLICATE_SECRET.sourceEnvironment),
+        sourceSecretPath: z
+          .string()
+          .trim()
+          .default("/")
+          .transform(removeTrailingSlash)
+          .describe(RAW_SECRETS.DUPLICATE_SECRET.sourceSecretPath),
+        destinationEnvironment: z.string().trim().describe(RAW_SECRETS.DUPLICATE_SECRET.destinationEnvironment),
+        destinationSecretPath: z
```
scott-ray-wilson, line 937: let's also ask @varonix0 about this end point being single vs bulk; we might just want to wait to add bulk and this is fine but outside the UI I imagine this would typically be used as a bulk operation - especially if we're not hiding it now

backend/src/server/routes/v4/secret-router.ts
```diff
@@ -898,6 +908,102 @@ export const registerSecretRouter = async (server: FastifyZodProvider) => {
...
   });
 
+  server.route({
+    method: "POST",
+    url: "/duplicate",
+    config: {
+      rateLimit: secretsLimit
+    },
+    schema: {
+      hide: false,
```
scott-ray-wilson, line 918: this says hide false but we're not adding api docs - should we be hiding this or do we want commit to this shape?

backend/src/services/secret/secret-service.ts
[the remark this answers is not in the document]
scott-ray-wilson: ah yes we don't want it to be "" but undefined - that way the secret value doesn't get overridden

https://github.com/user-attachments/assets/66fe74f3-2323-4861-948b-cc7d87cbbeee

backend/src/services/secret/secret-service.ts
```diff
@@ -3485,6 +3491,212 @@ export const secretServiceFactory = ({
...
+        message: `Secret with ID '${secretId}' not found in source folder with path '${sourceSecretPath}' and environment slug '${sourceEnvironment}'`
+      });
+    }
+    if (sourceSecret.isRotatedSecret) {
+      throw new BadRequestError({ message: `Cannot duplicate rotated secret: ${sourceSecret.key}` });
+    }
+    if (sourceSecret.isHoneyTokenSecret) {
+      throw new BadRequestError({ message: `Cannot duplicate honey token secret: ${sourceSecret.key}` });
+    }
+
```
scott-ray-wilson, line 3565: I think we might want to check and throw if you try to duplicate to the same env/path the source is from? essentailly a no-op but good to throw probably

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
scott-ray-wilson, line 1: probably should disable submit if no property is selected 

Image

scott-ray-wilson (changes requested):
first round of feedback

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
adilsitos: I searched the codebase and I've found other places that use this class

backend/src/server/routes/v4/secret-router.ts
adilsitos: This will be moved to allow several secrets, so I don't think this will be required right?

backend/src/server/routes/v4/secret-router.ts
adilsitos: I think this makes sense to be available through the API, so I will add the docs for this. Thanks for pointing this out.

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
adilsitos: image
it is way better

backend/src/server/routes/v4/secret-router.ts
adilsitos: Talked with him and changed this to also support bulk operations 🫡

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
scott-ray-wilson: it should not be there 😭  accidentally copied from shadcn but we don't have that token (that's why the text is still white)

backend/src/ee/services/audit-log/audit-log-types.ts
scott-ray-wilson, line 1: ignore file placement:

1. nit: if I have multiple secrets selected the title should probably be plural
2. nit: we say metadata but think we should say properties as metadata is a specific property
3. I don't think the env needs to be in a badge, doesn't really add anything and I think we should make the container flex wrap so that if either the env or folder is long it push them to separate lines (or just have them on separate lines always) but still handle truncation if they go beyond that
4. not sure we need the key icon but also cute so we can just make it the correct color `text-secret` (so it's more muted then the secret name)
5. can you make the overflow only affect the content, so that the header and footer don't participate in the scroll and have borders (see create secret form and screenshot below) and also reverse the button order so duplicate is on the left and cancel on the right

Image

Image

backend/src/services/secret/secret-service.ts
```diff
@@ -3485,6 +3491,258 @@ export const secretServiceFactory = ({
...
+    await snapshotService.performSnapshot(destinationFolder.id);
+    await secretQueueService.syncSecrets({
+      actor,
+      actorId,
+      secretPath: destinationFolder.path,
+      projectId,
+      orgId: project.orgId,
+      environmentSlug: destinationFolder.environment.slug
+    });
```
scott-ray-wilson, line 3751: I think we need to include the create/update events here that the original post processing includes for webhook / event subscriptions

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretEditTableRow.tsx
```diff
@@ -1655,6 +1658,43 @@ export const SecretEditTableRow = ({
...
                   </TooltipContent>
                 </Tooltip>
+                <ProjectPermissionCan
+                  I={ProjectPermissionActions.Create}
+                  a={subject(ProjectPermissionSub.Secrets, {
+                    environment,
+                    secretPath,
+                    secretName,
+                    secretTags: ["*"]
+                  })}
```
scott-ray-wilson: I think we should remove this permission check, creating in the current environment isn't reflective of their ability to create in the target envs - we should handle permissions inside the modal or just let the backend block on them

frontend/src/pages/secret-manager/OverviewPage/components/SelectionPanel/SelectionPanel.tsx
```diff
@@ -314,6 +316,56 @@ export const SelectionPanel = ({
...
+      permission.can(
+        ProjectPermissionSecretActions.Create,
+        subject(ProjectPermissionSub.Secrets, {
+          environment: duplicateSourceEnvSlugForPermission,
+          secretPath,
+          secretName: "*",
+          secretTags: ["*"]
+        })
+      )
+  );
```
scott-ray-wilson: same here, don't think we should gate on create from source, we can filter target envs in the modal if we want to prevent or let backend throw

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
```diff
@@ -0,0 +1,378 @@
...
+
+        <Controller
+          control={control}
+          name="include"
+          render={({ field: { value, onChange } }) => (
+            <Field>
+              <FieldLabel>Include properties</FieldLabel>
+              <FieldContent>
+                <div className="divide-y divide-border rounded-md border border-border">
+                  {INCLUDE_OPTIONS.map((opt) => {
```
scott-ray-wilson, line 289: I think we should disable the secret value if the user doesn't have read value and uncheck it by default (with tooltip that says you don't have access to this kind of message)

Image

scott-ray-wilson (changes requested):
feels way smoother!

scott-ray-wilson (dismissed):
Looks great, nice work!

scott-ray-wilson (approved):
LGTM!

PR #6419, opened 2026-05-12
chore(api): create short lived token in key value store for email sign up

Thiago-AS:
## Context

This PR replaces the database-backed email sign-up token flow with a Redis-based HMAC OTP system. Instead of creating a user record before email verification, the new flow stores only a short-lived HMAC of the OTP in Redis, keyed by an HMAC of the email address. The user record is created only after the OTP has been successfully validated. A Redis lock mutex is used to protect the retry-decrement logic from race conditions.

This also adds a sign-up OTP cooldown when requesting a new token. The backend will not issue a new OTP for 60 seconds, and the frontend now displays a countdown timer to provide feedback to the user.

Follow-up needed: create a migration to remove the unused TokenType.TOKEN_EMAIL_CONFIRMATION records from the database once the new flow is fully rolled out.

## Screenshots

image

image

```
{
    "reqId": "req-rTJOrUeFZQPpSe",
    "statusCode": 400,
    "message": "Please wait before requesting another code",
    "error": "BadRequest",
    "details": {
        "cooldownSeconds": 20
    }
}
```

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [x] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/auth-token/auth-token-service.ts
+++ b/backend/src/services/auth-token/auth-token-service.ts
@@ -13,7 +13,13 @@ import { TMembershipUserDALFactory } from "../membership-user/membership-user-da
 import { TOrgDALFactory } from "../org/org-dal";
 import { TUserDALFactory } from "../user/user-dal";
 import { TTokenDALFactory } from "./auth-token-dal";
-import { TCreateTokenForUserDTO, TIssueAuthTokenDTO, TokenType, TValidateTokenForUserDTO } from "./auth-token-types";
+import {
+  TCreateTokenForUserDTO,
+  TEmailSignupOtpPayload,
+  TIssueAuthTokenDTO,
+  TokenType,
+  TValidateTokenForUserDTO
+} from "./auth-token-types";
 
 type TAuthTokenServiceFactoryDep = {
   tokenDAL: TTokenDALFactory;
@@ -358,6 +364,52 @@ export const tokenServiceFactory = ({ tokenDAL, userDAL, orgDAL, keyStore }: TAu
     return { user, tokenVersionId: token.tokenVersionId, orgId, orgName, rootOrgId, parentOrgId };
   };
 
+  const createEmailSignupToken = async (email: string): Promise<string> => {
+    const appCfg = getConfig();
+    const token = String(crypto.randomInt(10 ** 5, 10 ** 6 - 1));
+    const tokenHash = await crypto.hashing().createHash(token, appCfg.SALT_ROUNDS);
+    const ttlSeconds = KeyStoreTtls.EmailSignupOtpInSeconds;
+    const payload: TEmailSignupOtpPayload = {
+      tokenHash,
+      triesLeft: 3,
+      expiresAt: Date.now() + ttlSeconds * 1000
+    };
+    const emailHash = crypto.nativeCrypto.createHash("sha256").update(email).digest("hex");
+    await keyStore.setItemWithExpiry(KeyStorePrefixes.EmailSignupOtp(emailHash), ttlSeconds, JSON.stringify(payload));
+    return token;
+  };
+
+  const validateEmailSignupToken = async (email: string, code: string): Promise<void> => {
+    const emailHash = crypto.nativeCrypto.createHash("sha256").update(email).digest("hex");
+    const raw = await keyStore.getItem(KeyStorePrefixes.EmailSignupOtp(emailHash));
+    const parsed = raw ? (JSON.parse(raw) as TEmailSignupOtpPayload) : null;
+
+    // Always run bcrypt comparison for constant-time behaviour regardless of token presence.
+    const hashToCompare = parsed?.tokenHash ?? DUMMY_HASH;
+    const isValidToken = await crypto.hashing().compareHash(code, hashToCompare);
+
+    if (!parsed || parsed.expiresAt < Date.now()) {
+      throw new Error("Invalid token");
+    }
+
+    if (!isValidToken) {
+      if (parsed.triesLeft <= 1) {
+        await keyStore.deleteItem(KeyStorePrefixes.EmailSignupOtp(emailHash));
+      } else {
+        const remainingTtlSec = Math.max(1, Math.floor((parsed.expiresAt - Date.now()) / 1000));
+        const updated: TEmailSignupOtpPayload = { ...parsed, triesLeft: parsed.triesLeft - 1 };
+        await keyStore.setItemWithExpiry(
+          KeyStorePrefixes.EmailSignupOtp(emailHash),
+          remainingTtlSec,
+          JSON.stringify(updated)
+        );
+      }
+      throw new Error("Invalid token");
+    }
+
+    await keyStore.deleteItem(KeyStorePrefixes.EmailSignupOtp(emailHash));
+  };
+
   return {
     createTokenForUser,
     validateTokenForUser,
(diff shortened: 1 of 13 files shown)


backend/src/services/auth-token/auth-token-service.ts
```diff
@@ -358,6 +361,98 @@ export const tokenServiceFactory = ({ tokenDAL, userDAL, orgDAL, keyStore }: TAu
...
   };
 
+  const createEmailSignupToken = async (email: string): Promise<{ token: string; cooldownSeconds: number }> => {
+    const appCfg = getConfig();
+    const emailHash = computeHash(email, appCfg.AUTH_SECRET);
+
+    const cooldownKey = KeyStorePrefixes.EmailSignupResendCooldown(emailHash);
+    const hasCooldown = await keyStore.getItem(cooldownKey);
+    if (hasCooldown) {
+      const remaining = await keyStore.ttl(cooldownKey);
```
PrestigePvP, line 379: Instead of using TTL here could we just throw the date in object we get from redis for the cooldownKey? 

Doing two separate calls here could cause a racing condition where the top `getItem` would return but the ttl is -1 and the user experience is that our system is wrong, when in reality the response is just slightly delayed?

backend/src/services/auth-token/auth-token-service.ts
```diff
@@ -358,6 +361,98 @@ export const tokenServiceFactory = ({ tokenDAL, userDAL, orgDAL, keyStore }: TAu
...
+  const validateEmailSignupToken = async (email: string, code: string): Promise<void> => {
+    const appCfg = getConfig();
+    const emailHash = computeHash(email, appCfg.AUTH_SECRET);
+    const key = KeyStorePrefixes.EmailSignupOtpHash(emailHash);
+
+    // Acquire a short-lived lock to make the tries-decrement / delete atomic.
+    const lock = await keyStore.acquireLock([KeyStorePrefixes.EmailSignupOtpLock(emailHash)], 5000);
+    try {
+      const raw = await keyStore.getItem(key);
+      const parsed = raw ? (JSON.parse(raw) as TEmailSignupOtpPayload) : null;
```
PrestigePvP: NIT: If the goal is to check the validity of `raw` to see if the item is in the keystore I'd rather an explicit check than a ternary

backend/src/services/auth-token/auth-token-service.ts
```diff
@@ -13,88 +13,91 @@ import { TMembershipUserDALFactory } from "../membership-user/membership-user-da
...
 type TAuthTokenServiceFactoryDep = {
   tokenDAL: TTokenDALFactory;
   userDAL: Pick<TUserDALFactory, "findById" | "transaction">;
   orgDAL: Pick<TOrgDALFactory, "findOne" | "findEffectiveOrgMembership">;
   membershipUserDAL: Pick<TMembershipUserDALFactory, "findOne">;
-  keyStore: Pick<TKeyStoreFactory, "setItemWithExpiry" | "getItem" | "deleteItem">;
+  keyStore: Pick<
+    TKeyStoreFactory,
+    "setItemWithExpiry" | "getItem" | "deleteItem" | "acquireLock" | "deleteItemsByKeyIn" | "ttl"
+  >;
```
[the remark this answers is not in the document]
Thiago-AS: We could allow zero padding, but to avoid a possible breaking change where someone consumes that and excludes trailing zeros, I kept the old behavior.

backend/src/services/auth/auth-signup-service.ts
```diff
@@ -99,37 +87,35 @@ export const authSignupServiceFactory = ({
...
-      // If the user was valid but the token failed, same generic message.
-      throw new Error("Invalid or expired verification request");
-    }
 
-    // Reject *after* the constant-time token validation work.
-    if (shouldReject) {
+    await tokenService.validateEmailSignupToken(sanitizedEmail, code);
+
+    // Only create (or recover) the user after the OTP has been verified.
+    let user = await userDAL.findOne({ username: sanitizedEmail });
```
PrestigePvP, line 102: Could we use a `INSERT INTO ... on CONFLICT UPDATE` so this logic is 1 db query?

frontend/src/components/auth/CodeInputStep.tsx
```diff
@@ -50,24 +51,54 @@ const codeInputStylePhone = {
...
   const {
     mutateAsync: verifyCode,
     isPending: isVerifying,
     isError: isCodeError
   } = useVerifySignupEmailVerificationCode();
   const [code, setCode] = useState("");
+  const [cooldown, setCooldown] = useState(initialCooldown);
   const { t } = useTranslation();
 
+  useEffect(() => {
```
PrestigePvP, line 76: Could we not get a similar affect by storing the `startTime - now`?

PrestigePvP (changes requested):

backend/src/services/auth-token/auth-token-service.ts
Thiago-AS: I tried to avoid needing to always manipulate the object because of the hashing, so we could reduce the amount of bytes being manipulated. In an ideal world, I would use a Lua script to handle transactional Redis operations, but since we don’t necessarily know whether the Redis instances running on clients support that, I followed this approach. I’ll update it to manipulate the object instead so we can avoid that.

backend/src/services/auth-token/auth-token-service.ts
```diff
@@ -358,6 +361,98 @@ export const tokenServiceFactory = ({ tokenDAL, userDAL, orgDAL, keyStore }: TAu
...
+      expiresAt: Date.now() + ttlSeconds * 1000
+    };
+
+    await keyStore.setItemWithExpiry(
+      KeyStorePrefixes.EmailSignupOtpHash(emailHash),
+      ttlSeconds,
+      JSON.stringify(payload)
+    );
+    await keyStore.setItemWithExpiry(cooldownKey, KeyStoreTtls.EmailSignupResendCooldownInSeconds, "1");
+
```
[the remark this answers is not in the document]
Thiago-AS: Fixed

backend/src/services/auth-token/auth-token-service.ts
PrestigePvP: Gotcha, yeah, that's a solid thought. 

I mean honestly we COULD derive the TTL from the expireAt that's already in the object but an issuedAt time in the object might just be easier.

backend/src/services/auth/auth-signup-service.ts
```diff
@@ -71,23 +72,10 @@ export const authSignupServiceFactory = ({
...
-    if (!user) throw new Error("Failed to create user");
 
-    const token = await tokenService.createTokenForUser({
-      type: TokenType.TOKEN_EMAIL_CONFIRMATION,
-      userId: user.id
-    });
+    const { token, cooldownSeconds } = await tokenService.createEmailSignupToken(sanitizedEmail);
 
     await smtpService
       .sendMail({
```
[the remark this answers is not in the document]
Thiago-AS: Fixed

frontend/src/components/auth/CodeInputStep.tsx
Thiago-AS: Nice catch, that would avoid time drift.

backend/src/services/auth/auth-signup-service.ts
Thiago-AS: I don't believe this will happen anymore, since we are not pre creating the user, I only kept for short compatibility but I plan to remove afterwards.

PrestigePvP (approved):

PR #6439, opened 2026-05-12
fix(pam): web session lifecycle cleanup

saifsmailbox98:
## Context

**1. Failed connections stamped "active".**

`handlePostgresSession` caught reachability failures internally and returned normally. The service had no idea the connection failed and proceeded to mark the session "active" — leaving a dead session with an incorrect status. Fixed by letting the reachability error propagate (matching SSH and Redis handlers, which already throw on connection failure). The service's existing catch block handles cleanup, and the activation step never runs.

> *TL;DR: Target DB is unreachable but the session gets marked "active" anyway. Now the handler throws on failure like SSH/Redis already do.*

**2. Orphan DB connections on teardown.**

In the multi-tab Postgres handler, each `open-connection` request is async — it awaits a full pg.Client handshake before registering in the controllers map. If the WS disconnects mid-handshake, `cleanup()` snapshots the map, clears it, and disposes what's there — but in-flight opens aren't in the map yet. When they resolve, they register into a dead session and are never disposed. Fixed by reserving a null slot in the map before the `await`. The map size now reflects in-flight opens for the 20-connection cap, and after the `await`, a `.has()` recheck detects if cleanup wiped the slot — the late-born controller disposes itself instead of registering.

> *TL;DR: User disconnects while a tab is still opening — the pg.Client finishes connecting but nobody ever closes it. Now we reserve the slot upfront and the orphan cleans itself up.*

**3. Session lockout from stale rows.**

When a session's WebSocket disconnects, `cleanup()` tears down resources and fires an ALPN signal for the gateway to end the session in the DB. But the ALPN signal is best-effort — if it fails (or the process crashes before cleanup runs), the session stays "active" or "starting" indefinitely. The only safety net was a BullMQ delayed job in Redis, which is lost on Redis flush. These stale rows count toward the 5-session cap and can lock users out for weeks. Fixed by adding `expireOverdueSessions` — a lazy cleanup that runs right before the cap check and marks any of the user's sessions past their `expiresAt` as ended. Also added a `PAM_SESSION_END` audit log to the existing `endSessionById` call in `cleanup()` so session-end events are recorded even when the gateway's ALPN path doesn't fire.

> *TL;DR: User closes tab, session stays "active" in DB forever, locks them out after 5 sessions. Now we expire sessions that have outlived their `expiresAt` duration right when the user tries to connect again — only provably expired ones, not just disconnected.*

**4. AI summary ordering.**

The recent addition of `endSessionById` to `cleanup()` means the session is already marked "ended" in the DB by the time the gateway finishes uploading logs and calls `POST /:sessionId/end`. The service-layer `endSessionById` was returning early with `alreadyEnded: true` and skipping the AI summary queue. Fixed by always queuing the summary regardless of `alreadyEnded`, since the gateway calling that endpoint is the signal that logs are uploaded and ready for summarization.

> *TL;DR: Since `endSessionById` was recently added to `cleanup()`, the gateway's `POST /:sessionId/end` now finds the session already ended and the AI summary was getting skipped. Service-layer `endSessionById` now always queues it — the gateway calling end is the signal that logs are uploaded.*

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/pam-session/pam-session-dal.ts
```diff
@@ -73,6 +73,21 @@ export const pamSessionDALFactory = (db: TDbClient) => {
...
+      .where("userId", userId)
+      .where("projectId", projectId)
+      .whereIn("status", [PamSessionStatus.Active, PamSessionStatus.Starting])
+      .where("expiresAt", "<", now)
+      .update({ status: PamSessionStatus.Ended, endedAt: now });
+    return updatedCount;
+  };
+
   const findByProjectId = async (projectId: string, tx?: Knex) => {
     const sessions = await (tx || db.replicaNode())(TableName.PamSession)
```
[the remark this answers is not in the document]
saifsmailbox98: no logs to summarize for crashed sessions. summary gets triggered when gateway calls end after uploading logs, which we handle now

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
```diff
@@ -415,6 +418,17 @@ export const pamWebAccessServiceFactory = ({
...
 
+      // Mark session as ended AFTER resources are freed so the session slot
+      // doesn't reopen while pg connections / relay tunnel are still alive.
+      // endSessionById is a no-op if the session is already ended.
+      if (session) {
+        try {
+          await pamSessionDAL.endSessionById(session.id);
+        } catch (err) {
+          logger.error(err, `Failed to end session in DB [sessionId=${session.id}]`);
+        }
```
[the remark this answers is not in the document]
saifsmailbox98: fixed in latest commit. audit log added to cleanup, ai summary moved to gateway path since it has the logs

saifsmailbox98:
@claude review once

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
```diff
@@ -532,6 +549,9 @@ export const pamWebAccessServiceFactory = ({
         throw new BadRequestError({ message: "Gateway not configured for this resource" });
       }
 
+      // Expire overdue sessions before the cap check so stale rows don't lock the user out.
+      await pamSessionDAL.expireOverdueSessions(userId, projectId);
+
       // Check web session limit
       const activeCount = await pamSessionDAL.countActiveWebSessions(userId, projectId);
       if (activeCount >= MAX_WEB_SESSIONS_PER_USER) {
```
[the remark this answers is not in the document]
saifsmailbox98: fixed

saifsmailbox98:
@claude review once

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -247,14 +247,13 @@ export const pamSessionServiceFactory = ({
...
+
+    if (alreadyEnded && session.status !== PamSessionStatus.Ended && session.status !== PamSessionStatus.Terminated) {
+      throw new BadRequestError({ message: "Cannot end sessions that are not active or starting" });
     }
 
-    // Fire-and-forget AI summarization
+    // Always queue — the gateway calling end means logs are uploaded.
     void (async () => {
       try {
         await pamSessionAiSummaryService.queueAiSummary(sessionId, session.projectId);
```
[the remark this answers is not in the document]
saifsmailbox98: this is a pre-existing issue in queueAiSummary,  not introduced by this PR

backend/src/ee/services/pam-session/pam-session-dal.ts
```diff
@@ -79,6 +79,17 @@ export const pamSessionDALFactory = (db: TDbClient) => {
...
   };
 
+  const expireOverdueSessions = async (userId: string, projectId: string, tx?: Knex): Promise<number> => {
+    const now = new Date();
+    const updatedCount = await (tx || db)(TableName.PamSession)
+      .where("userId", userId)
+      .where("projectId", projectId)
+      .whereIn("status", [PamSessionStatus.Active, PamSessionStatus.Starting])
+      .where("expiresAt", "<", now)
+      .update({ status: PamSessionStatus.Ended, endedAt: now });
```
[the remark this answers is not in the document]
saifsmailbox98: fixed, added .where("accessMethod", "web")

backend/src/ee/services/pam-session/pam-session-dal.ts
```diff
@@ -79,6 +79,18 @@ export const pamSessionDALFactory = (db: TDbClient) => {
     return updatedCount;
   };
 
+  const expireOverdueSessions = async (userId: string, projectId: string, tx?: Knex): Promise<number> => {
+    const now = new Date();
+    const updatedCount = await (tx || db)(TableName.PamSession)
+      .where("userId", userId)
+      .where("projectId", projectId)
+      .where("accessMethod", "web")
```
x032205, line 87: Since we're only targeting web sessions, maybe we should include "web" in the function name? I like the name `endExpiredWebSessions` because it's self explanatory.

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
```diff
@@ -246,6 +246,8 @@ export const pamWebAccessServiceFactory = ({
       accountIdentity = `${domainConnectionDetails.domain}:${account.name}`;
     }
 
+    await pamSessionDAL.expireOverdueSessions(actor.id, projectId);
```
x032205: We should probably add a comment why this is here. It may seem confusing to clean up sessions when a new one is starting, but with the context that this ends ALL of the user's expired sessions it makes more sense

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -247,14 +247,13 @@ export const pamSessionServiceFactory = ({
...
     const updatedSession = await pamSessionDAL.endSessionById(sessionId);
-    if (!updatedSession) {
-      if (session.status !== PamSessionStatus.Ended && session.status !== PamSessionStatus.Terminated) {
-        throw new BadRequestError({ message: "Cannot end sessions that are not active or starting" });
-      }
-      return { session, projectId: session.projectId, alreadyEnded: true };
+    const alreadyEnded = !updatedSession;
+
+    if (alreadyEnded && session.status !== PamSessionStatus.Ended && session.status !== PamSessionStatus.Terminated) {
+      throw new BadRequestError({ message: "Cannot end sessions that are not active or starting" });
```
x032205: Saying "alreadyEnded" and then checking for ended or terminated is weird, maybe we dnt use that variable and instead say !updatedSession?

x032205 (changes requested):

backend/src/ee/services/pam-session/pam-session-dal.ts
saifsmailbox98: makes sense
renamed to `endExpiredWebSessions`

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
saifsmailbox98: added a comment

` // Sessions that outlived their expiresAt may still show as active — end them so that they don't count against the session limit.`

backend/src/ee/services/pam-session/pam-session-service.ts
saifsmailbox98: makes sense, dropped the variable and inlined !updatedSession

lb-vn (approved):

PR #6441, opened 2026-05-13
feat(external-migration): refactor usage to allow import using app connection

adilsitos:
## Context
Refactored the whole secret import to allow the usage of app connection without creating a migration config. The endpoints that create, list and delete migration config were removed. This will allow the import of secrets from users that has access to app connections, not only to admins. 

## Steps to verify the change
- Perform migrations inside projects and inside org (using doppler and vault)
- Create a migration of database info inside dynamic secrets (using vault)
- Create a migration of policies inside roles (using vault)
- Create a migration of authentication identities (using vault)

## Type

- [ ] Fix
- [x] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/app-connection/app-connection-fns.ts
```diff
@@ -407,7 +409,7 @@ export const decryptAppConnectionCredentials = async ({
     cipherTextBlob: encryptedCredentials
   });
 
-  return JSON.parse(decryptedPlainTextBlob.toString()) as TAppConnection["credentials"];
+  return JSON.parse(decryptedPlainTextBlob.toString()) as unknown as TAppConnection["credentials"];
```
adilsitos: I tried to remove this, but my linter kept complaining and adding this back.

docs/documentation/platform/external-migrations/envkey.mdx
adilsitos, line 1: let me know if you want me to add this again. The implementation for env key still exists, but as we talked, this seems pretty old. 

I plan to remove the env key implementation in another PR

backend/src/server/routes/v3/external-migration-router.ts
```diff
@@ -696,13 +520,24 @@ export const registerExternalMigrationRouter = async (server: FastifyZodProvider
 
   // ─── Doppler In-Platform Migration Routes ────────────────────────────────────
 
+  const assertExactlyOneDopplerSelector = (configId?: string, connectionId?: string) => {
```
varonix0: Why do we still allow the use of configId's? didn't we get rid of this entirely so it's fully based on connection ID's? This comment applies to many places in the PR

backend/src/services/app-connection/app-connection-fns.ts
varonix0: Might be related to your node version; works just fine without the `unknown` cast on my end

backend/src/services/app-connection/app-connection-fns.ts
```diff
@@ -395,7 +397,7 @@ export const decryptAppConnectionCredentials = async ({
   encryptedCredentials: Buffer;
   kmsService: TAppConnectionServiceFactoryDep["kmsService"];
   projectId: string | null | undefined;
-}) => {
+}): Promise<TAppConnection["credentials"]> => {
```
varonix0: Why did we add this? It should be able to infer it from the function return type

backend/src/services/external-migration/external-migration-service.ts
```diff
@@ -631,16 +441,40 @@ export const externalMigrationServiceFactory = ({
...
+  const getVaultPolicies = async ({
+    actor,
+    namespace,
+    connectionId
+  }: {
+    actor: OrgServiceActor;
+    namespace: string;
+    connectionId?: string;
+  }) => {
+    const connection = await resolveVaultConnection(actor, {
```
varonix0: I think everywhere we use this new `resolve` functions, we can just replace it with `appConnectionService.connectAppConnectionById()`, if we remove the support for configId's

backend/src/services/external-migration/external-migration-service.ts
```diff
@@ -870,63 +732,79 @@ export const externalMigrationServiceFactory = ({
...
-    const appConnection = await appConnectionService.connectAppConnectionById<TDopplerConnection>(
+    return appConnectionService.connectAppConnectionById<TDopplerConnection>(
       AppConnection.Doppler,
       config.connectionId,
       actor
     );
+  };
 
-    return appConnection;
+  const resolveDopplerConnection = (
```
varonix0: Can get rid of this too

varonix0 (changes requested):
general feedback; if we're uncertain about something, it's usually better to just sync up to make sure we're aligned on the implementation

backend/src/services/app-connection/app-connection-fns.ts
adilsitos: This was also because a type error was showing up to me. Maybe node version problem again.

backend/src/server/routes/v3/external-migration-router.ts
adilsitos: I thought the goal was to also keep the behavior with users that could possibly be using this by the API, but as we talked, this is not supported and if someone is using this way is by their own risks (since we don't document these endpoints)

varonix0 (approved):

PR #6457, opened 2026-05-13
feat: create audit log for honey token triggers

mathnogueira:
## Context

We added audit logs for honey tokens, but we never added one when the honey token is triggered.

## Screenshots

image

## Steps to verify the change

* Create honey token
* Trigger it
* See that audit log got created

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/honey-token/honey-token-service.ts
+++ b/backend/src/ee/services/honey-token/honey-token-service.ts
@@ -1035,41 +1035,41 @@ export const honeyTokenServiceFactory = ({
       if (updatedToken) {
         void $sendTriggerNotification({ orgId: honeyTokenWithOrg.orgId, honeyToken, eventMetadata: parsed.data });
         void telemetryService
-        .sendPostHogEvents({
-          event: PostHogEventTypes.HoneyTokenTriggered,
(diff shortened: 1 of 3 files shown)


backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -1020,11 +1024,17 @@ export const honeyTokenServiceFactory = ({
...
         TRIGGER_NOTIFICATION_COOLDOWN_MS
       );
-      void telemetryService
+
+      // This block only is executed once per token trigger. So, even if we get 100 events
+      // this will run only once.
+      if (updatedToken) {
+        void $sendTriggerNotification({ orgId: honeyTokenWithOrg.orgId, honeyToken, eventMetadata: parsed.data });
+        void telemetryService
         .sendPostHogEvents({
```
mathnogueira: we were sending multiple posthog events if attacker calls multiple APIs. It should count as one single "HoneyTokenTriggered" event instead.

adilsitos (approved):

PR #6460, opened 2026-05-13
fix: bug fix for honey token trigger

mathnogueira:
## Context

Fixed two edge case in triggering endpoint of honey token

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/honey-token/honey-token-service.ts
+++ b/backend/src/ee/services/honey-token/honey-token-service.ts
@@ -1029,7 +1029,7 @@ export const honeyTokenServiceFactory = ({
     if (updatedToken) {
       void $sendTriggerNotification({ orgId: honeyTokenWithOrg.orgId, honeyToken, eventMetadata });
 
-      void auditLogService
+      await auditLogService
         .createAuditLog({
           actor: {
             type: ActorType.UNKNOWN_USER,
@@ -1051,7 +1051,6 @@ export const honeyTokenServiceFactory = ({
             }
           }
         })
-        .catch(() => {});
     }
 
     return { acknowledged: true };
--- a/backend/src/ee/services/honey-token/honey-token-service.ts
+++ b/backend/src/ee/services/honey-token/honey-token-service.ts
@@ -961,33 +961,38 @@ export const honeyTokenServiceFactory = ({
     }
 
     const honeyTokenWithOrg = await honeyTokenDAL.findOneByTokenIdentifier(eventMetadata.accessKeyId);
-    if (!honeyTokenWithOrg) {
-      throw new UnauthorizedError({ message: "Invalid webhook request" });
+
+    let webhookSigningKey: string | null = null;
+    if (honeyTokenWithOrg) {
+      const config = await honeyTokenConfigDAL.findOne({
+        orgId: honeyTokenWithOrg.orgId,
+        type: HoneyTokenType.AWS,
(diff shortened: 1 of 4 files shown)


backend/src/ee/services/honey-token/honey-token-types.ts
```diff
@@ -21,16 +21,14 @@ export const AwsHoneyTokenEventMetadataSchema = z.object({
   requestParameters: z.unknown().nullable().optional()
 });
 
-export type TAwsHoneyTokenEventMetadata = z.infer<typeof AwsHoneyTokenEventMetadataSchema>;
-
-export const HoneyTokenEventMetadataSchema = z.discriminatedUnion("eventType", [
+export const HoneyTokenWebhookPayloadSchema = z.discriminatedUnion("type", [
   z.object({
```
akhilmhdh, line 25: Is this public schema or internal one? If public this would break the api contract

backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -926,8 +927,8 @@ export const honeyTokenServiceFactory = ({
     }
   };
 
-  const handleTrigger = async ({ type, signature, payload }: THandleTriggerInput) => {
-    logger.info({ payload, signature, type }, `Honey token trigger received [type=${type}]`);
+  const handleTrigger = async ({ type, signature, rawBody, payload }: THandleTriggerInput) => {
+    logger.info({ signature, type }, `Honey token trigger received [type=${type}]`);
 
     const providerType = assertSupportedHoneyTokenType(type);
```
akhilmhdh, line 937: To avoid the timing vector - we should do a dummy signature verification. Sharing a quick LLM example
```
const honeyTokenWithOrg = await honeyTokenDAL.findOneByTokenIdentifier(eventMetadata.accessKeyId);

// Fetch a dummy key to use if token not found (could be a config constant)
const dummySigningKey = "0".repeat(64);

let storedConfig: { webhookSigningKey: string } | null = null;
if (honeyTokenWithOrg) {
  const decryptor = await getOrgDataKeyDecryptor({ orgId: honeyTokenWithOrg.orgId });
  const decrypted = decryptor({ cipherTextBlob: config.encryptedConfig });
  storedConfig = AwsHoneyTokenConfigSchema.parse(JSON.parse(decrypted.toString()));
}

// Always compute HMAC - same CPU work regardless of path
const signingKey = storedConfig?.webhookSigningKey ?? dummySigningKey;
const expectedSignature = crypto.nativeCrypto
  .createHmac("sha256", signingKey)
  .update(`${timestamp}.${rawBody}`)
  .digest("hex");

// Now fail if token wasn't found (after doing the work)
if (!honeyTokenWithOrg || !storedConfig) {
  throw new UnauthorizedError({ message: "Invalid webhook request" });
}
```

akhilmhdh (changes requested):

backend/src/ee/services/honey-token/honey-token-types.ts
mathnogueira: This was actually not in use, it was dead code that we never cleaned up.

akhilmhdh (dismissed):

akhilmhdh (approved):

PR #6461, opened 2026-05-13
feat: updated the rollback to be aligned with our environment and path.

akhilmhdh:
## Context

This PR updates the rollback permission to more aligned with other permission in the dashboard. The rollback permission can now be scoped to an environment and secret path. The old permission rollback was broken for new custom roles as our UI didn't allow to add it and thus replaced it with the commit permission. 

The rollbackSnapshot service was removed because it's not used anymore and the router using it was throwing error as deprecated. 

## Steps to verify the change

1. Create a project permission with rollback under a environment and path
2. It should work in scoped environment

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/secret/secret-service.ts
```diff
@@ -2636,96 +2632,18 @@ export const secretServiceFactory = ({
...
-        if ((err as Error).message === "BadRequest: Failed to find secret") {
-          return null;
-        }
-
-        logger.error(err);
-      });
-
-    if (secretVersionV2) return secretVersionV2;
-
-    const secret = await secretDAL.findById(secretId);
```
mathnogueira, line 2667: You removed everything else in the fallback. Just want to make sure this is expected.

akhilmhdh: Yes! We don't support old projects anymore. Only upgraded one

mathnogueira (approved):

PR #6466, opened 2026-05-14
improvement(secret-share): add option to now show the external page and endpoint

adilsitos:
## Context
Add a parameter to not show the share secret page if the user defines it. Using this, people cannot use the share-secret public page.

## Steps to verify the change
- Set `DISABLE_SECRET_SHARING=true` on the .env
- Recreate the container of infisical using the current env
- Check that the path `share-secret` shows up 404 not found

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/lib/config/env.ts
+++ b/backend/src/lib/config/env.ts
@@ -54,6 +54,7 @@ const envSchema = z
       .enum(["true", "false"])
       .default("false")
       .transform((el) => el === "true"),
+    DISABLE_SECRET_SHARING: zodStrBool.default("false"),
     REDIS_URL: zpStr(z.string().optional()),
     REDIS_USERNAME: zpStr(z.string().optional()),
     REDIS_PASSWORD: zpStr(z.string().optional()),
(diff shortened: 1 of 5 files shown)


backend/src/lib/config/env.ts
```diff
@@ -54,6 +54,7 @@ const envSchema = z
       .enum(["true", "false"])
       .default("false")
       .transform((el) => el === "true"),
+    DISABLE_SECRET_SHARING: zodStrBool.default("false"),
```
varonix0: I think the env name is slightly misleading, `DISABLE_SECRET_SHARING` makes me think _all_ secret sharing is disabled. How about `DISABLE_PUBLIC_SECRET_SHARING`?

varonix0 (changes requested):

varonix0 (approved):

PR #6467, opened 2026-05-14
fix(secret-sync): make per-connection concurrency admission atomic and extend retry budget to 60 min

scott-ray-wilson:
## Context

This PR fixes the secret sync concurrency for app connection atomic and extends the retry budget to 60 minutes. Also bumps concurrency limit to 5.

## Screenshots

N/A

## Steps to verify the change

- create 50 syncs using the same app connection
- add 300 secrets to trigger all simultaneously
- ensure the syncs succeed

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/secret-sync/secret-sync-queue.ts
+++ b/backend/src/services/secret-sync/secret-sync-queue.ts
@@ -86,7 +86,7 @@ type TSecretSyncQueueFactoryDep = {
   cronJob: TCronJobFactory;
   kmsService: Pick<TKmsServiceFactory, "createCipherPairWithDataKey">;
   appConnectionDAL: Pick<TAppConnectionDALFactory, "findById" | "update" | "updateById">;
-  keyStore: Pick<TKeyStoreFactory, "acquireLock" | "setItemWithExpiry" | "getItem">;
+  keyStore: Pick<TKeyStoreFactory, "acquireLock" | "setItemWithExpiry" | "incrementBy" | "setExpiry">;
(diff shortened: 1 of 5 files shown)


backend/src/services/secret-sync/secret-sync-queue.ts
```diff
@@ -1148,84 +1137,89 @@ export const secretSyncQueueFactory = ({
...
+        }
+      } finally {
+        if (admittedConnectionSlot) {
+          await $releaseConnectionConcurrency(connectionId);
+        }
+
+        await lock.release();
+      }
+    },
+    { concurrency: 5 }
```
scott-ray-wilson: reminder to remove this before merging; it's too simulate multiple workers in a production environment

victorvhs017 (changes requested):
If the container dies (OOM or even a deploy rollout), the counter gets stuck, and I can't acquire a lock in Redis anymore, blocking any other sync.

And because we call `setExpiry`, which uses the [expire] (https://redis.io/docs/latest/commands/expire/#refreshing-expires) command, the TTL is always refreshed, and the record is never expired.

Unfortunately, the baked-in fix is only on Pro https://docs.bullmq.io/bullmq-pro/groups/concurrency. We need to think about how to handle this recovery.

victorvhs017 (dismissed):

victorvhs017 (approved):

PR #6490, opened 2026-05-16
feat: allow multiple git hub apps per organization

Thiago-AS:
## Context

Allows organizations to register and manage their own GitHub Apps for GitHub App-based connections, instead of relying solely on the instance-wide app configured through environment variables.

### What's New

- GitHub App registration via the manifest flow — Users can create a GitHub App directly from the connection form. The backend generates the app manifest, GitHub redirects back after app creation, and the app credentials (private key and client secret) are stored encrypted using the organization's KMS key. The flow is secured with a signed, single-use state token and a client-side CSRF check.
- Per-connection app selection — A new GitHubAppSelector component allows users to choose between the instance-default (shared) GitHub App and any organization-managed app. Users can also create a new app inline without losing any in-progress form state.
- GitHub App management — Added new /api/v1/github-apps endpoints to list, delete, and check the installation status of organization-managed apps. App deletion is blocked while active connections reference the app, enforced through a github_app_connections link table and foreign key constraints, making it safe against concurrent connection creation. The deletion flow also attempts to uninstall the app's GitHub installations on a best-effort basis.
- Credential resolution and security — resolveGitHubAppCredentials now resolves credentials from either an organization-managed GitHub App or the instance-level configuration. The connection's target host and instance type are derived from the app record and cannot be supplied by callers, preventing app JWTs and client secrets from being used against attacker-controlled hosts. All outbound GitHub requests are routed through safeRequest, which provides SSRF protection.

## Steps to verify the change

New custom flow:
- Go to App Connections
- Add a GitHub App connection
- Click Create New GitHub App and continue on GitHub
- After creating the app, you'll be redirected back to the modal
- Verify the app was created in GitHub: Settings → Developer Settings → GitHub Apps
- Create the app connection
- When prompted, add the GitHub App installation
- You'll be redirected back to select an existing installation or create a new one
- Create a new installation
- You'll be redirected back to the Infisical App Connections list
- Create a Secret Manager project
- Create a secret
- Sync the secret to GitHub using the GitHub App connection 

Old instance flow flow:
- Create GitHub App using the doc https://infisical.com/docs/integrations/app-connections/github#self-hosted-instance
- Export the env vars 
- Go to App Connections
- Select Share GitHub App
- Create the app connection
- When prompted, add the GitHub App installation
- You'll be redirected back to select an existing installation or create a new one
- Create a new installation
- You'll be redirected back to the Infisical App Connections list
- Create a Secret Manager project
- Create a secret
- Sync the secret to GitHub using the GitHub App connection

Reuse of existing GitHub App:
- Go to App Connections
- Select existing Custom App Connection
- Create the app connection
- When prompted, add the GitHub App installation
- You'll be redirected back to select an existing installation
- You'll be redirected back to the Infisical App Connections list
- Create a Secret Manager project
- Create a secret
- Sync the secret to GitHub using the GitHub App connection

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/db/migrations/20260529204743_create-github-apps.ts
```diff
@@ -0,0 +1,32 @@
...
+      t.string("clientId").notNullable();
+      t.binary("encryptedClientSecret").notNullable();
+      t.binary("encryptedPrivateKey").notNullable();
+      t.string("slug").notNullable();
+      t.string("owner").nullable();
+      t.string("host").nullable();
+      t.string("instanceType").notNullable().defaultTo("cloud");
+      t.timestamps(true, true, true);
+      t.unique(["orgId", "name"]);
+    });
```
scott-ray-wilson, line 27: I could be wrong, but I'm almost 90% sure Gap wants this supported at the project level too 😬 so there's no cross org-bleeding of connections - we should verify

frontend/src/pages/organization/AppConnections/GitHubManifestCallbackPage/GitHubManifestCallbackPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const returnToForm = () =>
+      navigate({
+        to: "/organizations/$orgId/app-connections",
+        params: { orgId },
+        search: { addConnectionApp: AppConnection.GitHub }
+      });
+
+    if (!gitHubAppId || !slug || !installState) {
+      createNotification({ type: "error", text: "Invalid GitHub manifest callback parameters." });
+      navigate({ to: "/organizations/$orgId/app-connections", params: { orgId } });
```
scott-ray-wilson: This always directs back to the org page, even if I started this flow from a project app connections page

it would be also great if this was supported in our inline creation flow for secret syncs / secret rotations; those work the the instance app so it would flow better if this could integrate with that

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/GitHubConnectionForm.tsx
```diff
@@ -229,6 +396,101 @@ export const GitHubConnectionForm = ({ appConnection, projectId, onSubmit }: Pro
     <FormProvider {...form}>
       <form onSubmit={handleSubmit(submitHandler)}>
         {!isUpdate && <GenericAppConnectionsFields />}
+        <div className="mb-4 flex items-center gap-3">
```
scott-ray-wilson: I preferred this section behind an advanced accordion, why'd we switch to switch?

Image

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/GitHubConnectionForm.tsx
```diff
@@ -261,6 +524,23 @@ export const GitHubConnectionForm = ({ appConnection, projectId, onSubmit }: Pro
...
+              isLoading={isGitHubAppsLoading}
+              value={selectedGitHubApp}
+              onChange={setSelectedGitHubApp}
+              host={instanceType === "server" ? watch("credentials.host") : undefined}
+              instanceType={instanceType}
+              onCreateApp={handleCreateApp}
+              isCreating={isRedirecting}
+            />
+          </FormControl>
+        )}
```
scott-ray-wilson, line 558: this section adds a lot of friction for one of the most commonly used app connections; 99.9% of users don't need to see this and seeing it will confuse them

I think we have a few options, just throwing out ideas:
1. make this a separate "method" (even if that's just a UI abstraction) - IE I select Custom GitHub App or something in the method dropdown and then I see the custom list
2.  when a user select github app there is just a regular select that defaults to instance and then they can select a custom, there can be a gear icon next to it to manage them in a popover
3. move to an advanced section accordion (with enterpise settings)

I think the biggest point though is this shouldn't be an inline list view, it should be behind a select or popover, I think we're blending two contexts - creating a github app connection and managing github apps, which should be separate. it takes a lot of real estate for something most users will never use 

Image

scott-ray-wilson: btw the actual flow for setting one up is very smooth, very well done

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/GitHubAppSelector.tsx
```diff
@@ -0,0 +1,391 @@
...
+              placeholder="deploy-bot"
+              className="font-mono"
+            />
+          </FormControl>
+          <FormControl label="GitHub account / org" isOptional className="mt-3 mb-0">
+            <Input
+              value={newAppOrg}
+              onChange={(e) => setNewAppOrg(e.target.value)}
+              placeholder="Leave blank for your personal account"
+            />
```
scott-ray-wilson: I don't think it's super clear to users when they'd want to use this, maybe an info tool tip explaining when you should do this and what it means (more context then just not personal, ie if you want the app to install on an organization)

Image

backend/src/server/routes/v1/github-app-router.ts
scott-ray-wilson, line 1: we should definitely add audit logs for these new methods!

backend/src/services/app-connection/github/github-connection-fns.ts
```diff
@@ -15,11 +16,132 @@ import { withGatewayV2Proxy } from "@app/lib/gateway-v2/gateway-v2";
...
+    if (!app) {
+      throw new BadRequestError({
+        message: `GitHub App with id ${gitHubAppId} not found in this organization.`
+      });
+    }
+
+    const { decryptor } = await kmsService.createCipherPairWithDataKey({
+      type: KmsDataKey.Organization,
+      orgId
+    });
```
scott-ray-wilson: just making a note here if we do support project scope we'll need to select the kms conditionally

scott-ray-wilson (changes requested):
first round of feedback

backend/src/db/migrations/20260529204743_create-github-apps.ts
Thiago-AS: We can do that. It doesn't really matter from the GitHub App's perspective, because what matters is where the app is installed. It would only restrict visibility within the Infisical app itself.

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/GitHubConnectionForm.tsx
Thiago-AS: It felt weird having the accordion in the middle of the screen, will add back so you can take a look at it.

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/GitHubConnectionForm.tsx
Thiago-AS: Yeah, I agree, but couldn't find a better place to put that, I don't really like opening another modal over that one, we already have a shit ton o modal navigation. Will think a little more and experiment more.

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/GitHubAppSelector.tsx
scott-ray-wilson, line 1: this is a much better experience!

The only thing I think we should change is either rename the shared label to instance or instance default or add that in parenthesis / muted text following the name 

Image

scott-ray-wilson, line 1: the fact I can select here feels a little off / disconnected, I almost think this shouldn't be selectable here, just for managing 🤔 

nits: the link / delete button don't have cursor-pointer (might be nice to have tooltip too but not required)
and bottom button should also have cursor-pointer

I also think we should have a close button at the bottom probably 

Image

scott-ray-wilson, line 1: not required but would be nice if this supported inline creation from sync/rotations like the instance level does

https://github.com/user-attachments/assets/2eecd982-7ea2-42e3-98e6-b09da881c3bc

frontend/src/pages/organization/AppConnections/OauthCallbackPage/OauthCallbackPage.tsx
```diff
@@ -556,54 +753,85 @@ export const OAuthCallbackPage = () => {
...
+                    <span className="text-xs text-mineshaft-400">{installation.accountType}</span>
+                  </div>
+                </Button>
+              ))}
+            </div>
+          )}
+          <Button
+            variant={hasInstallations ? "plain" : "outline_bg"}
+            isFullWidth
+            isDisabled={Boolean(pickedInstallationId)}
```
scott-ray-wilson: this is such a nice addition!!

could we make this a v3 card and v3 buttons (ie all v3) since it's in it's own context; no reason to stick to v2 here

backend/src/db/migrations/20260529204743_create-github-apps.ts
```diff
@@ -0,0 +1,43 @@
...
+      t.timestamps(true, true, true);
+      t.unique(["orgId", "projectId", "name"]);
+    });
+
+    // unique name for org-level apps
+    await knex.raw(`
+        CREATE UNIQUE INDEX ${UNIQUE_NAME_ORG_GITHUB_APP_INDEX}
+        ON ${TableName.GitHubApp} ("orgId", "name")
+        WHERE "projectId" IS NULL
+      `);
```
scott-ray-wilson, line 34: maybe we should also enforce unique names within the same project?

frontend/src/helpers/appConnections.ts
scott-ray-wilson, line 1: ignore file placement:

I think we should maybe prevent selecting org github apps if a user doesn't have the connect permission (the same way they wouldn't be able to select a org app connection) - this seems the most intuitive since we don't have a designated permission

Image

Image

scott-ray-wilson, line 1: ignore fileplacement:

it seems like if I'm not already logged into github I just get this; is there a way to prompt login to avoid this?

Image

scott-ray-wilson, line 1: probably just more a note for the future but it'd be nice to see where the connection is being used (or offer cascade delete)

Image

backend/src/db/migrations/20260529204743_create-github-apps.ts
scott-ray-wilson: oh seems like this might not even be doable; if I use the same name I just get a 400 

Image

```
infisical-dev-api              | [21:33:35.198] ERROR (1941): A GitHub App with name "infisical-test-inline" already exists in this project. {"reqId":"req-kJwbLapWolDmXn","severity":"ERROR"}
infisical-dev-api              |     err: {
infisical-dev-api              |       "type": "BadRequestError",
infisical-dev-api              |       "message": "A GitHub App with name \"infisical-test-inline\" already exists in this project.",
infisical-dev-api              |       "stack":
infisical-dev-api              |           BadRequest: A GitHub App with name "infisical-test-inline" already exists in this project.
infisical-dev-api              |               at Object.initiateManifestCreation (/app/src/services/github-app/github-app-service.ts:351:13)
infisical-dev-api              |               at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
infisical-dev-api              |               at async Object.handler (/app/src/server/routes/v1/github-app-router.ts:35:22)
infisical-dev-api              |       "name": "BadRequest"
infisical-dev-api              |     }
```

we might want to make the error message more clear

scott-ray-wilson (changes requested):
looks great over all, final comments

frontend/src/pages/organization/AppConnections/AppConnectionsPage/components/AppConnectionForm/GitHubAppSelector.tsx
scott-ray-wilson: nvm realized it does work, I just didn't finish the flow 😓

Thiago-AS: Done!

Thiago-AS: Done!

frontend/src/pages/organization/AppConnections/OauthCallbackPage/OauthCallbackPage.tsx
Thiago-AS: Done!

backend/src/db/migrations/20260529204743_create-github-apps.ts
Thiago-AS: Done!

frontend/src/helpers/appConnections.ts
Thiago-AS: Github does not provide a way to do that, I added a note on the information card when creating a new app.

scott-ray-wilson (commented):
LGTM

scott-ray-wilson (approved):
fr

PR #6501, opened 2026-05-18
feat: resolved tx missing in scim

akhilmhdh:
## Context

This PR is a small fix on missing tx in one of the scim operation.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/scim/scim-service.ts
+++ b/backend/src/ee/services/scim/scim-service.ts
@@ -1318,8 +1318,11 @@ export const scimServiceFactory = ({
       updatedGroup = await processReplacement(outerTx);
       await $syncNewMembersRoles(updatedGroup, members, outerTx);
     } else {
-      updatedGroup = await groupDAL.transaction(processReplacement);
-      await $syncNewMembersRoles(updatedGroup, members);
+      updatedGroup = await groupDAL.transaction(async (tx) => {
+        const replacedGroup = await processReplacement(tx);
+        await $syncNewMembersRoles(updatedGroup, members, tx);
+        return replacedGroup;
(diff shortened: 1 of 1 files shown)


backend/src/ee/services/scim/scim-service.ts
mathnogueira, line 1036: ~I don't think this will cause any issues as Knex uses savepoints behind the hood. But good to be aware that you might have a transaction running inside another transaction here.~

Nevermind.

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -1306,12 +1316,12 @@ export const scimServiceFactory = ({
     let updatedGroup: TGroups;
     if (outerTx) {
       updatedGroup = await processReplacement(outerTx);
+      await $syncNewMembersRoles(updatedGroup, members, outerTx);
     } else {
       updatedGroup = await groupDAL.transaction(processReplacement);
+      await $syncNewMembersRoles(updatedGroup, members);
```
mathnogueira: shouldn't you use the transaction from `groupDAL`?


```suggestion
    updatedGroup = await groupDAL.transaction(async (tx) => {
        const updatedGroups = await processReplacement(tx)
        await $syncNewMembersRoles(updatedGroup, members, tx);
        return updatedGroup
    });
```

akhilmhdh: Fixed

mathnogueira (approved):

PR #6519, opened 2026-05-19
feat: display both org and project machine identities at org level

Thiago-AS:
## Context

- Display both org and project machine identities at org level
- Only display project identities the user have access to

## Screenshots
Screenshot 2026-05-19 at 16 32 14
Screenshot 2026-05-19 at 16 32 19
Screenshot 2026-05-19 at 16 32 26

## Steps to verify the change

- Enter as an Admin, you should be able to see every identity in the organization, including sub projects.
- Enter as an Member, you should be able to see all identities managed by the organization, and the ones you are part of the project.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


Thiago-AS:
@claude review

backend/src/services/identity/identity-service.ts
```diff
@@ -482,6 +503,153 @@ export const identityServiceFactory = ({
...
+              });
+              return projectPermission.can(ProjectPermissionIdentityActions.Read, ProjectPermissionSub.Identity)
+                ? projectId
+                : null;
+            } catch {
+              return null;
+            }
+          })
+        );
+
```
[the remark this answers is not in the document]
Thiago-AS: The fan-out is already softened, but not flat. getProjectPermission is request-memoized and Redis-cached, so in steady state the N calls are mostly N Redis round-trips + N CASL MongoAbility builds per request, not N permissionDAL.getPermission queries. Real DB load only spikes after cache invalidations. Since this is not a hot path, it's a search/listing endpoint I don't think it should be a problem for now/

backend/src/lib/api-docs/constants.ts
```diff
@@ -208,6 +208,28 @@ export const IDENTITIES = {
...
+    orderDirection: "The direction to order identities in."
+  },
+  SEARCH_COUNT_V2: {
+    scope:
+      "Array of scopes to count. Returns a count for each requested scope. Accepts 'organization' and/or 'project'.",
+    search: {
+      desc: "The filters to apply when counting.",
+      name: "The name of the identity to filter by.",
+      role: "The role of the identity to filter by."
+    }
```
victorvhs017: Following the pattern of `IDENTITY_ADDITIONAL_PRIVILEGE_V2`, we should scope these inside `IDENTITIES_V2` with `SEARCH` and `SEARCH_COUNT`.

backend/src/server/routes/v2/identity-router.ts
victorvhs017, line 1: Let's add audit logs for both routes.

backend/src/services/identity/identity-org-dal.ts
victorvhs017, line 1: We are showing one row per membership, we should show one row per unique identity.

Image

We want to help users to visualize unique identities across the organization, because we charge per machine identity and not per membership.

backend/src/services/identity/identity-service.ts
```diff
@@ -482,6 +504,235 @@ export const identityServiceFactory = ({
...
+    actorId: string;
+    actorAuthMethod: TSearchIdentitiesV2DTO["actorAuthMethod"];
+    actorOrgId: string;
+    scope: SearchIdentitiesScope[];
+  }): Promise<{
+    uniqueScope: Set<SearchIdentitiesScope>;
+    accessibleProjectIds: string[];
+    projectPermissions: Map<string, TProjectPermissionAbility>;
+    conditionalProjectIds: Set<string>;
+  }> => {
```
victorvhs017: We should keep the `organization role - machine identity - read` permission check here. If the user doesn't have it, they shouldn't be able to access the tab:

```
ForbiddenError.from(permission).throwUnlessCan(OrgPermissionIdentityActions.Read, OrgPermissionSubjects.Identity);
```

Even if they have access to see some project-only machine identities, this data was scoped by organization on v1, and the view is scoped by organization in the UI.

backend/src/services/identity/identity-service.ts
```diff
@@ -482,6 +504,235 @@ export const identityServiceFactory = ({
...
+    }
+
+    if (requestedProject && uniqueScope.has(SearchIdentitiesScope.ProjectScope)) {
+      const { docs } = await identityOrgMembershipDAL.searchIdentitiesV2({
+        orgId: actorOrgId,
+        scope: [SearchIdentitiesScope.ProjectScope],
+        accessibleProjectIds,
+        searchFilter
+      });
+      counts.project = filterIdentitiesByProjectPermission(docs, projectPermissions, conditionalProjectIds).length;
```
victorvhs017: Do the full searchIdentitiesV2 query here only to get the length seems overkill, could we add a specific lightweight query in the DAL to get this instead?

frontend/src/hooks/api/orgIdentityMembership/mutation.tsx
```diff
@@ -35,7 +35,7 @@ export const useDeleteOrgIdentityMembership = () => {
       return data.identityMembership;
     },
     onSuccess: () => {
-      queryClient.invalidateQueries({ queryKey: identitiesKeys.searchIdentities({ search: {} }) });
+      queryClient.invalidateQueries({ queryKey: identitiesKeys.searchIdentitiesRoot });
```
victorvhs017, line 38: Shouldn't we invalidate the new count query here, too?

frontend/src/hooks/api/projectIdentity/mutations.tsx
```diff
@@ -68,6 +70,7 @@ export const useDeleteProjectIdentity = () => {
       queryClient.invalidateQueries({
         queryKey: identitiesKeys.getIdentityProjectMemberships(identityId)
       });
+      queryClient.invalidateQueries({ queryKey: identitiesKeys.searchIdentitiesRoot });
       queryClient.invalidateQueries({
```
victorvhs017, line 74: same question here

frontend/src/hooks/api/projectIdentityMembership/mutations.tsx
```diff
@@ -104,6 +106,7 @@ export const useDeleteProjectIdentityMembership = () => {
       queryClient.invalidateQueries({
         queryKey: identitiesKeys.getIdentityProjectMemberships(identityId)
       });
+      queryClient.invalidateQueries({ queryKey: identitiesKeys.searchIdentitiesRoot });
       queryClient.invalidateQueries({ queryKey: pkiApplicationKeys.all });
```
victorvhs017, line 110: and here

victorvhs017 (changes requested):

backend/src/server/routes/v2/identity-router.ts
Thiago-AS: Should we? We are only listing the identities, not actually using them or viewing any secret, looks a little overkill and would overblown the audit logs. I feel we should only be auditing log operations that really matters in terms of auditing, like viewing a secret, updating something and stuff like that. Viewing something sensible in general.

backend/src/services/identity/identity-service.ts
Thiago-AS: I agree in terms of UI, but in terms of backend, we show based on what the user has access or not, if the user does not have access to org read, we return empty. Otherwise we will have different behaviors for each scope. Are we ok with that? doesn't seem like a good modeling and it can be confusing.

backend/src/services/identity/identity-org-dal.ts
Thiago-AS: I don't know if this makes sense, because if we enter each project, if there is a membership of an org role, it is listed there, if we remove this from here, the listings would have different behavior and it breaks UX. We should have a product vision on this, all listing machine identities should behave the same (in each scope).

frontend/src/hooks/api/orgIdentityMembership/mutation.tsx
Thiago-AS: It invalidates the count query too. invalidateQueries does prefix matching on query keys.
```
  - searchIdentitiesRoot = ["identity", "search"]
  - searchIdentities(dto) = ["identity", "search", dto]
  - countIdentitiesRoot = ["identity", "search", "count"]
  - countIdentities(dto) = ["identity", "search", "count", dto]
```

backend/src/services/identity/identity-service.ts
victorvhs017: yeah, I'm on the fence on this one, but I agree that it is not good to have different behaviors. I think we can leave and go based on the scope. Let me test it again after the membership fix.

frontend/src/hooks/api/orgIdentityMembership/mutation.tsx
victorvhs017: Oh that's right, missed the same prefix there!

backend/src/lib/api-docs/constants.ts
Thiago-AS: Fixed.

backend/src/services/identity/identity-service.ts
Thiago-AS: Fixed as well, created a lighter version just so we can CASL and still count them.

victorvhs017 (changes requested):
Image

We also need to update the API reference docs. Search should be pointing to v2, and the new search count endpoint should be there.

The old one should be under the "Legacy" like we do for Secrets:

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityTable.tsx
victorvhs017, line 1: We need to remove the case before sorting

Image

victorvhs017, line 1: Can we keep the same order and use the custom role name in both places? A shared component would be great.

For the custom role name, be sure to handle big names; we'll probably need to add ellipses

Image

Image

victorvhs017, line 1: I think "Never" should come before anything

Image

victorvhs017, line 1: We need to update this filter. We could add two sections inside the list for projects and organization roles. Or even remove it and make the search also search for results in role names? We should pull Scott into this discussion.

Image

victorvhs017, line 1: Role order is changing when I change the rows per page.

https://github.com/user-attachments/assets/10383472-08ae-4dc7-a049-15fae9a156c6

victorvhs017, line 1: Can we update the screenshot we have here with the new table?

https://infisical.com/docs/documentation/platform/identities/machine-identities

victorvhs017 (changes requested):

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityTable.tsx
Thiago-AS: I kept the same old behavior, will update

Thiago-AS: Yeah, will create a shared component.

Thiago-AS: Uhm, I don't know, usually in tables the absence of a value always goes to the bottom. Users usually care more about recent activity than untouched records.

Thiago-AS: Let me just see if the project search endpoint returns the custom role name.

victorvhs017: Yep, we can go either way if it was intentional, your reason makes sense to me.

Thiago-AS: Uhmm yeah, I don't know about that, since this is the organization table, I think focusing on the organization roles are the most coherent semantic.

Thiago-AS: Done, removed the filter and added as a search OR key in the search box

victorvhs017 (approved):

PR #6540, opened 2026-05-20
feat(ldap): add mutual TLS (mTLS) client certificate support

victorvhs017:
## Context

Adds mutual TLS (mTLS) support to the LDAP SSO configuration, enabling integration with providers that require client certificate authentication during the TLS handshake (e.g. Google Workspace Secure LDAP).

Previously, LDAP connections only supported server-side TLS verification via an optional CA certificate. Now, administrators can also supply a PEM-encoded client certificate and private key that are presented during the TLS handshake. Both fields must be provided together; validation enforces this on the backend and frontend.

Additionally includes documentation for configuring Google Workspace Secure LDAP with Infisical, and updates the default on-prem feature set to enable enterprise features out of the box.

## Screenshots

image

## Steps to verify the change

- Follow the step-by-step in the new documentation.
- Test group mapping
- Test an LDAP machine identity auth

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/routes/v1/ldap-router.ts
```diff
@@ -230,7 +232,9 @@ export const registerLdapRouter = async (server: FastifyZodProvider) => {
           .trim()
           .default("(|(memberUid={{.Username}})(member={{.UserDN}})(uniqueMember={{.UserDN}}))")
           .describe(LdapSso.CREATE_CONFIG.groupSearchFilter),
-        caCert: z.string().trim().default("").describe(LdapSso.CREATE_CONFIG.caCert)
+        caCert: z.string().trim().default("").describe(LdapSso.CREATE_CONFIG.caCert),
+        clientCertificate: z.string().trim().default("").describe(LdapSso.CREATE_CONFIG.clientCertificate),
+        clientKeyCertificate: z.string().trim().default("").describe(LdapSso.CREATE_CONFIG.clientKeyCertificate)
```
Thiago-AS, line 237: Should we be returning the client private key? I don't think it should ever leave the server, once registered, if needed to be managed, needs to delete + create once again.

frontend/src/pages/organization/SettingsPage/components/OrgSsoTab/LDAPModal.tsx
```diff
@@ -171,30 +194,31 @@ export const LDAPModal = ({ popUp, handlePopUpClose, handlePopUpToggle, hideDele
...
-        caCert: watchCaCert ?? ""
+    if (Boolean(watchClientCertificate?.trim()) !== Boolean(watchClientKeyCertificate?.trim())) {
+      createNotification({
+        text: "Client Certificate and Client Private Key must be provided together for mTLS.",
+        type: "error"
       });
+      return;
+    }
+
+    const isConnected = await testLDAPConnection({
```
Thiago-AS, line 245: Why was the try catch was removed? In case of failure while testing the connection, should we have any notification like we did before: "Failed to test LDAP connection.", or just fail silently?

Thiago-AS (changes requested):

Thiago-AS:
image

nit: Is there a reason why those optional fields are not hidden behind some sort of selection like most of the other forms, like auth - TLS opens the first one or mTLS opens all?

backend/src/ee/routes/v1/ldap-router.ts
victorvhs017: Agreed! I'll update this!

frontend/src/pages/organization/SettingsPage/components/OrgSsoTab/LDAPModal.tsx
victorvhs017: It was creating a duplicated notification error because we already handle errors at the mutation level in `frontend/src/hooks/api/reactQuery.tsx`.

victorvhs017:
> image
> nit: Is there a reason why those optional fields are not hidden behind some sort of selection like most of the other forms, like auth - TLS opens the first one or mTLS opens all?

That would be a nice improvement! Let me add something there

backend/src/ee/routes/v1/ldap-router.ts
jongio: Applied in the replacement PR #2352 - added @tmeschter as owner for both airunway-aks-setup and azure-hosted-copilot-sdk.

victorvhs017: @jongio I think you commented on the wrong repo

jongio: Yes, sorry about that.

Thiago-AS (approved):

Issue #6555, opened 2026-05-20
GitLab Secret Sync: project/group selector is unusable for large organizations (2000+ repos)


ahmadalfy:
### Product

Platform/Access controls/Other

### Describe the bug

When setting up a GitLab Secret Sync using an App Connection, the project and group selector dropdowns become unusable for users who have access to a large number of GitLab repositories (e.g. 2000+).

**Root cause:** `listGitLabProjects` and `listGitLabGroups` in `backend/src/services/app-connection/gitlab/gitlab-connection-fns.ts` use Gitbeaker's `.all()` method, which exhausts all paginated pages before returning:

```ts
// gitlab-connection-fns.ts:311
const projects = await  .all({
  archived: false,
  membership: true,
  ...
});
```

For an organization with 2000+ repositories, this triggers dozens of sequential GitLab API calls before the backend can respond. The frontend then renders a single `FilterableSelect` dropdown that blocks on this entire response, resulting in a timeout or an extremely long wait with no feedback to the user.

## Impact

* Users with access to large GitLab organizations (100+ repos) are effectively unable to use the GitLab Secret Sync feature.
* The request may time out entirely, leaving no way to configure the sync.
* There is no partial loading, search-as-you-type, or progressive disclosure. It's all-or-nothing.

## Proposed Solution

GitLab's API supports server-side search and pagination natively. The fix would involve two parts:

**1. Backend: expose a search/paginate endpoint instead of bulk fetch**

Replace the `GET /app-connections/gitlab/:connectionId/projects` endpoint (which currently returns everything) with one that accepts `?search=` and `?page=` query params, proxying them directly to GitLab's API:

`GET /api/v4/projects?membership=true&search=&page=&per_page=20`

This keeps the backend response fast regardless of organization size.

**2. Frontend: switch from a static dropdown to a search-as-you-type selector**

Replace the `FilterableSelect` (which requires the full list upfront) with a combobox that fires a debounced query to the backend search endpoint as the user types. This is a common pattern already used elsewhere in the UI.

**Alternative (simpler) approach:** Add a navigation flow. First pick a GitLab group, then load only projects within that group. This reduces the dataset significantly and avoids needing a new search endpoint.

## Questions for maintainers

1. Is this a known issue? Are there plans to address it? I checked all issues related to Gitlab and couldn't find any related issuesl
2. Would you accept a PR implementing the search-as-you-type approach described above?
3. Is there a preferred pattern in the codebase for async combobox selectors we should follow?

[...]

PR #6560, opened 2026-05-20
feat(gitlab): replace project dropdown with Miller columns navigator

ahmadalfy:
Fixes a hang for users with a lot of repos by lazily loading one level at a time instead of fetching all projects upfront.

## Context

#6555 

## Screenshots

https://github.com/user-attachments/assets/46caf957-be44-4973-b369-253cc6d24eee

## Steps to verify the change

Instead of loading all accessible projects at once (which triggers dozens of GitLab API calls before rendering anything), the project selector now uses a **Miller columns (Mac Finder-style) navigator**.

  The user drills down through their group hierarchy one level at a time:

  1. The first column loads only top-level groups
  2. Clicking a group opens a new column to the right showing its subgroups and direct projects
  3. The user continues navigating until they find and select their project

  Each column fetches only the data for that level, so a user with 2000+ repos never triggers a bulk fetch. The UI responds immediately at every step.

  ### Before

  The project dropdown would hang indefinitely (or time out) for users with access to large numbers of repositories, because the backend was calling Gitbeaker's `.all()` method which exhausts all paginated pages before returning.

  ### After

  The navigator loads instantly. Each level is a single scoped GitLab API call (`/groups/:id/subgroups` or `/groups/:id/projects`), and already-visited columns are cached by React Query so navigating back is instant.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/app-connection-routers/gitlab-connection-router.ts
+++ b/backend/src/server/routes/v1/app-connection-routers/gitlab-connection-router.ts
@@ -7,6 +7,7 @@ import {
   CreateGitLabConnectionSchema,
   SanitizedGitLabConnectionSchema,
   TGitLabGroup,
+  TGitLabGroupTreeItem,
   TGitLabProject,
   UpdateGitLabConnectionSchema
 } from "@app/services/app-connection/gitlab";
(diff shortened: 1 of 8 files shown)


adilsitos:
Closing this because a similar implementation was done [here](https://github.com/Infisical/infisical/pull/6612)

PR #6565, opened 2026-05-21
feat(pki): support wildcard DNS identifiers in ACME

saifsmailbox98:
## Context

The ACME server was rejecting wildcard domains like `*.example.com` because the DNS identifier validator didn't allow `*` characters. Updated it to accept wildcard prefixes per RFC 8555 — the authorization stores the stripped domain with a `wildcard: true` flag, and only DNS-01 challenges are offered for wildcards since HTTP-01 can't prove control over all subdomains.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/db/migrations/20260519144500_add-wildcard-to-pki-acme-auths.ts
+++ b/backend/src/db/migrations/20260519144500_add-wildcard-to-pki-acme-auths.ts
@@ -1,0 +1,19 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasColumn(TableName.PkiAcmeAuth, "wildcard"))) {
+    await knex.schema.alterTable(TableName.PkiAcmeAuth, (t) => {
+      t.boolean("wildcard").nullable().defaultTo(false);
+    });
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
+  if (await knex.schema.hasColumn(TableName.PkiAcmeAuth, "wildcard")) {
+    await knex.schema.alterTable(TableName.PkiAcmeAuth, (t) => {
+      t.dropColumn("wildcard");
+    });
+  }
+}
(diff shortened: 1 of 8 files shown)


backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -785,12 +786,16 @@ export const pkiAcmeServiceFactory = ({
...
             {
               accountId: account.id,
               status: skipDnsOwnershipVerification ? AcmeAuthStatus.Valid : AcmeAuthStatus.Pending,
               identifierType: identifier.type,
-              identifierValue: identifier.value,
+              identifierValue,
+              wildcard: isWildcard || undefined,
               // RFC 8555 suggests a token with at least 128 bits of entropy
               // We are using 256 bits of entropy here, should be enough for now
               // ref: https://datatracker.ietf.org/doc/html/rfc8555#section-11.3
```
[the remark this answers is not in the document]
saifsmailbox98: fixed

saifsmailbox98:
@claude review once

backend/src/db/migrations/20260519144500_add-wildcard-to-pki-acme-auths.ts
```diff
@@ -0,0 +1,19 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasColumn(TableName.PkiAcmeAuth, "wildcard"))) {
+    await knex.schema.alterTable(TableName.PkiAcmeAuth, (t) => {
+      t.boolean("wildcard").nullable().defaultTo(false);
```
carlosmonastyrski: why is it nullable? shouldn't it be just either true/false?

carlosmonastyrski: will be simpler in the rest of the code to not look for null values

carlosmonastyrski (changes requested):

backend/src/db/migrations/20260519144500_add-wildcard-to-pki-acme-auths.ts
saifsmailbox98: yeah, it doesn't need to be nullable, fixed it everywhere. good catch.

carlosmonastyrski (approved):

PR #6567, opened 2026-05-21
improvement(webhook): add environment name on body

adilsitos:
## Context
add environment name on body

## Steps to verify the change
- Create a new webhook
- update a secret 
- check that the body has the environment name

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


adilsitos:
eng-5416

backend/src/services/secret-folder/secret-folder-dal.ts
```diff
@@ -236,18 +237,26 @@ export const secretFolderDALFactory = (db: TDbClient) => {
 
       const idMap = buildFolderIdMap(allEnvFolders);
 
-      // Map environmentSlug by envId from target folders
+      // Map env slug/name by envId from target folders
       const envSlugMap: Record<string, string> = {};
+      const envNameMap: Record<string, string> = {};
       for (const f of targetFolders) {
-        envSlugMap[f.envId] = (f as typeof f & { environmentSlug: string }).environmentSlug;
+        const typed = f as typeof f & { environmentSlug: string; environmentName: string };
```
mathnogueira: You don't need the `typed` variable, `f` already include both fields.

backend/src/services/secret-folder/secret-folder-dal.ts
```diff
@@ -236,18 +237,26 @@ export const secretFolderDALFactory = (db: TDbClient) => {
 
       const idMap = buildFolderIdMap(allEnvFolders);
 
-      // Map environmentSlug by envId from target folders
+      // Map env slug/name by envId from target folders
       const envSlugMap: Record<string, string> = {};
+      const envNameMap: Record<string, string> = {};
```
mathnogueira: Instead of two maps, why don't you use something like:

```ts
const envMap: Record<string, typeof targetFolders[number]> = {}
for (const f of targetFolders) {
    envMap[f.envId] = f;
}
```

So you just have one map and reference it like:

```ts
environmentSlug: envMap[folder.envId].environmentSlug,
environmentName: envMap[folder.envId].environmentName,
```

mathnogueira (changes requested):

mathnogueira (approved):

PR #6574, opened 2026-05-21
improvement(platfor-286): add e2e tests for saml

PrestigePvP:
## Context

- Add Playwright e2e suite under `e2e/` — covers SCIM-provisioned SAML login (SP-initiated and IdP-initiated), SCIM deactivation revoking SAML login, and SAML response rejection paths.
- Add `backend/scripts/seed-gamma-fixtures.ts` — idempotent bootstrap of the gamma e2e org, SCIM token, SAML config, and verified email domain. Refuses to run against anything that looks like prod.
- Suite runs against gamma between deploy and prod promotion; a red run blocks the prod-deploy jobs in the `infrastructure` repo's workflow.
- Mock IdP we control (no SaaS dependency); only two real secrets in CI (`E2E_SCIM_TOKEN`, `E2E_IDP_ADMIN_TOKEN`).

[PLATFOR-286](https://linear.app/infisical/issue/PLATFOR-286)

## Steps to verify the change

- `cd e2e && npm install && npm run type:check` passes.
- `cd e2e && npm test` against gamma with `.env` populated — all tests pass.

## Type

- [x] Improvement

## Checklist

- [x] Tested locally
- [x] Updated CLAUDE.md files


scott-ray-wilson (approved):
LGTM

one follow-up; I think we need to update our rejection error responses so we can tighten that test as it's pretty ambiguous right now

PR #6585, opened 2026-05-22
feat: added secret value blind index support

akhilmhdh:
## Context

This PR implements support for secret value blind index 

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/db/migrations/20260522180000_add-secret-value-blind-index.ts
+++ b/backend/src/db/migrations/20260522180000_add-secret-value-blind-index.ts
@@ -1,0 +1,46 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  // Add secretValueBlindIndex to secrets_v2
+  if (await knex.schema.hasTable(TableName.SecretV2)) {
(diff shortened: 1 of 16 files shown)


backend/src/db/migrations/20260522180000_add-secret-value-blind-index.ts
```diff
@@ -0,0 +1,46 @@
...
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  // Add secretValueBlindIndex to secrets_v2
+  if (await knex.schema.hasTable(TableName.SecretV2)) {
+    const hasColumn = await knex.schema.hasColumn(TableName.SecretV2, "secretValueBlindIndex");
+    if (!hasColumn) {
+      await knex.schema.alterTable(TableName.SecretV2, (t) => {
+        // 64 chars = 32 bytes hex encoded (HMAC-SHA256 output)
+        t.string("secretValueBlindIndex", 64).nullable();
```
varonix0, line 12: question: what's tha rationale of using hex instead of base64? you could argue that base64 would be better as it's smaller

varonix0 (changes requested):

varonix0 (approved):

PR #6587, opened 2026-05-22
feat: add support for DigiCert revocation status

carlosmonastyrski:
## Context

Adds a daily background job that detects DigiCert certificates revoked outside of Infisical and reflects that state in our inventory. For each DigiCert CA, the cron makes a single call to DigiCert's status-changes endpoint to get the last 24 hours of order activity, filters down to anything newly revoked, and updates the matching local certificates so they show up as revoked in the dashboard and on the CRL. Each detected revocation is recorded in the audit log so customers have a trail of out-of-band changes. The job runs daily off-peak, processes each CA in isolation so a single misconfigured CA doesn't affect the others.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/certificate-authority/digicert/digicert-revocation-sync-queue.ts
+++ b/backend/src/services/certificate-authority/digicert/digicert-revocation-sync-queue.ts
@@ -1,0 +1,160 @@
+/* eslint-disable no-await-in-loop */
+import { TableName } from "@app/db/schemas";
+import { EventType, TAuditLogServiceFactory } from "@app/ee/services/audit-log/audit-log-types";
+import { getConfig } from "@app/lib/config/env";
+import { CronJobName, TCronJobFactory } from "@app/lib/cron/cron-job";
+import { logger } from "@app/lib/logger";
+import { TAppConnectionDALFactory } from "@app/services/app-connection/app-connection-dal";
+import { ActorType } from "@app/services/auth/auth-type";
+import { TCertificateDALFactory } from "@app/services/certificate/certificate-dal";
+import { revocationReasonToCrlCode } from "@app/services/certificate/certificate-fns";
+import { CertStatus, CrlReason } from "@app/services/certificate/certificate-types";
+import { TKmsServiceFactory } from "@app/services/kms/kms-service";
+
+import { TCertificateAuthorityDALFactory } from "../certificate-authority-dal";
+import { CaStatus, CaType } from "../certificate-authority-enums";
+import { createDigiCertApiClient } from "./digicert-api-client";
+import {
+  castDbEntryToDigiCertCertificateAuthority,
+  getDigiCertClientCredentials
+} from "./digicert-certificate-authority-fns";
+
+// 03:00 UTC: off-peak and one hour after the existing daily certificate-cleanup cron (02:00 UTC).
+const REVOCATION_SYNC_CRON_SCHEDULE = "* * * * *";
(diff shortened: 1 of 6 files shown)


backend/src/services/certificate-authority/digicert/digicert-revocation-sync-queue.ts
```diff
@@ -0,0 +1,160 @@
...
+    serialNumber: string;
+    projectId: string;
+  }) => {
+    await certificateDAL.updateById(cert.id, {
+      status: CertStatus.REVOKED,
+      revokedAt: new Date(),
+      revocationReason: revocationReasonToCrlCode(CrlReason.UNSPECIFIED)
+    });
+
+    await auditLogService.createAuditLog({
```
saifsmailbox98, line 66: do we also need to trigger the revocation alerts for this flow?

backend/src/services/certificate-authority/digicert/digicert-revocation-sync-queue.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  const markCertificateRevokedLocally = async (cert: {
+    id: string;
+    commonName: string;
+    serialNumber: string;
+    projectId: string;
+  }) => {
+    await certificateDAL.updateById(cert.id, {
+      status: CertStatus.REVOKED,
+      revokedAt: new Date(),
```
saifsmailbox98, line 62: this will add the current time instead of the time at which it was actually revoked at digicert. is there no way to get the actual time? would that require an extra network-trip, what would that look like with the amount of traffic/revocation we predict in the near future?

backend/src/services/certificate-authority/digicert/digicert-revocation-sync-queue.ts
```diff
@@ -0,0 +1,160 @@
...
+import { CaStatus, CaType } from "../certificate-authority-enums";
+import { createDigiCertApiClient } from "./digicert-api-client";
+import {
+  castDbEntryToDigiCertCertificateAuthority,
+  getDigiCertClientCredentials
+} from "./digicert-certificate-authority-fns";
+
+// 03:00 UTC: off-peak and one hour after the existing daily certificate-cleanup cron (02:00 UTC).
+const REVOCATION_SYNC_CRON_SCHEDULE = "0 3 * * *";
+const REVOCATION_LOOKBACK_SECONDS = 24 * 60 * 60;
```
saifsmailbox98: since everything is idempotent, can we increase this to handle the following cases:
1. cron miss for an entire day
2. slight cron delays leaving a small few-minute window - any cert revoked in that won't be fetched from digicert

saifsmailbox98 (changes requested):

backend/src/services/certificate-authority/digicert/digicert-revocation-sync-queue.ts
carlosmonastyrski: Technically achievable, but at meaningful extra API cost (the only way to get it is a paginated walk of `GET /request?filters[type]=revoke&filters[status]=approved` per CA to match by `order.id`, then a per-match `Request Info` call to read `date_processed`)

saifsmailbox98 (approved):

PR #6592, opened 2026-05-23
chore: batch audit log stream for external providers and make it more resilient

Thiago-AS:
### What

Reworks audit log ingestion and external streaming to be more durable, scalable, and decoupled. Audit events are enriched at push time and appended to a single Redis stream. A single-leader (Redlock) consumer drains the stream every 5 seconds, batch-inserts records into Postgres and ClickHouse, and then fans them out to external providers through a transactional outbox (audit_log_stream_outbox). Delivery is handled by debounced flush jobs with provider-specific batch sizing, exponential backoff retries, and a dead-letter queue (DLQ).

### Why

The previous per-event delivery model provided no durability or retry guarantees. The new outbox architecture delivers events with at-least-once semantics using idempotent inserts and delivered-state tracking to deduplicate retries. It also respects provider-specific batch limits and automatically recovers stale processing claims through a sweep cron.

### Key Changes
- Push-time enrichment: buildStreamEntry now resolves organization, plan, TTL, and project metadata before events are written to the stream, eliminating database lookups during ingestion. Remains backward-compatible with existing flat event formats.
- Single ingestion queue: Uses a single queue with concurrency set to 5. Events are only trimmed after successful processing, ensuring failed batches are retried.
- Stream delivery modes: Supports both single and batch delivery modes. New streams default to batch mode, while existing Custom and Cribl streams remain on single mode for backward compatibility (Cribl streams can be upgraded).
- Provider updates: Datadog, Splunk, Azure, Cribl, and Custom provider implementations now support batched delivery with provider-specific byte and event-count limits.
- Frontend updates: Added stream mode configuration to the Cribl and Custom stream setup forms and updated related documentation.
- Test coverage: Added audit-log-queue.test.ts along with comprehensive outbox service tests.

single_queue

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


docs/documentation/platform/audit-log-streams/audit-log-streams.mdx
victorvhs017, line 1: We should be able to answer these questions after reading this document:

- What happens when an audit log fails to be pushed to an audit log stream?
- Can I manually retry it?
- How can I backfill these lost events in my SIEM?

This could mean one less support ticket

victorvhs017 (changes requested):

victorvhs017 (approved):

PR #6598, opened 2026-05-25
fix(pki): enforce authorization before cert-manager membership cleanup

saifsmailbox98:
## Context

Cert-manager access DELETE endpoints could be exploited by any authenticated user to permanently delete PKI application memberships and approval-policy approvers — even in projects they have no access to, including projects in other organizations.

**Root cause:** Each DELETE handler cleaned up PKI application memberships in a committed DB transaction *before* checking whether the caller had permission to delete from the project. If the permission check rejected the request (403), the cleanup was already permanent. The `inject-cert-manager-project-id` plugin also accepted explicit `projectId` from query/body without verifying org ownership, enabling cross-tenant targeting.

**Fix:**
- Reordered all 4 DELETE handlers (`/users`, `/users/:userId`, `/groups/:groupId`, `/identities/:identityId`) so the authorized deletion runs first, with cleanup after
- Both deletion and PKI cleanup now run in a single database transaction via new coordinating methods (`deleteMemberAndCleanup`, `deleteMembersAndCleanup`) in `pkiApplicationMembership`, using a callback pattern to pass the shared `tx`
- Added optional `externalTx` to the membership deletion services so they can participate in a caller-provided transaction
- Switched inline `actorKind` string literals to `ApplicationMemberKind` enum
- Added `isUuid()` + `isCertManagerProject(projectId, orgId)` validation for explicit project IDs in the injection plugin, matching the existing cookie-path validation

Resolves [SEC-29](https://linear.app/infisical/issue/SEC-29), [SEC-25](https://linear.app/infisical/issue/SEC-25), [SEC-15](https://linear.app/infisical/issue/SEC-15)

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/cert-manager-access-routers/groups-router.ts
+++ b/backend/src/server/routes/v1/cert-manager-access-routers/groups-router.ts
@@ -219,11 +219,6 @@ export const registerCertManagerAccessGroupsRouter = async (server: FastifyZodPr
     },
     handler: async (req) => {
       const projectId = req.internalCertManagerProjectId;
-      await server.services.pkiApplicationMembership.removeActorFromApplicationMemberships({
-        projectId,
-        actorKind: "group",
-        actorId: req.params.groupId
-      });
       const { membership: groupMembership } = await server.services.membershipGroup.deleteMembership({
         permission: req.permission,
         selector: { groupId: req.params.groupId },
@@ -236,6 +231,11 @@ export const registerCertManagerAccessGroupsRouter = async (server: FastifyZodPr
           type: EventType.REMOVE_CERT_MANAGER_GROUP,
           metadata: { groupId: req.params.groupId, membershipId: groupMembership.id }
         }
+      });
+      await server.services.pkiApplicationMembership.removeActorFromApplicationMemberships({
+        projectId,
+        actorKind: "group",
+        actorId: req.params.groupId
       });
       return {
         groupMembership: {
--- a/backend/src/server/routes/v1/cert-manager-access-routers/groups-router.ts
+++ b/backend/src/server/routes/v1/cert-manager-access-routers/groups-router.ts
@@ -12,6 +12,7 @@ import { ms } from "@app/lib/ms";
 import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
 import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
 import { AuthMode } from "@app/services/auth/auth-type";
+import { ApplicationMemberKind } from "@app/services/pki-application/pki-application-types";
 
 import { RolesUpdateBodySchema } from "./schemas";
 
@@ -234,7 +235,7 @@ export const registerCertManagerAccessGroupsRouter = async (server: FastifyZodPr
       });
       await server.services.pkiApplicationMembership.removeActorFromApplicationMemberships({
         projectId,
-        actorKind: "group",
+        actorKind: ApplicationMemberKind.Group,
         actorId: req.params.groupId
       });
       return {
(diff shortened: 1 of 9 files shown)


saifsmailbox98:
@claude review once

backend/src/server/routes/v1/cert-manager-access-routers/groups-router.ts
```diff
@@ -237,6 +232,11 @@ export const registerCertManagerAccessGroupsRouter = async (server: FastifyZodPr
           metadata: { groupId: req.params.groupId, membershipId: groupMembership.id }
         }
       });
+      await server.services.pkiApplicationMembership.removeActorFromApplicationMemberships({
+        projectId,
+        actorKind: "group",
```
sheensantoscapadngan: can we use an enum here?

sheensantoscapadngan: shouldn't we handle removal of application membership together with membership group deletion inside a single database transaction? if `removeActorFromApplicationMemberships` fails due to a transient issue then it would be orphaned no?

backend/src/server/routes/v1/cert-manager-access-routers/identities-router.ts
```diff
@@ -283,6 +278,11 @@ export const registerCertManagerAccessIdentitiesRouter = async (server: FastifyZ
           metadata: { identityId: req.params.identityId, membershipId: membership.id }
         }
       });
+      await server.services.pkiApplicationMembership.removeActorFromApplicationMemberships({
+        projectId,
+        actorKind: "identity",
+        actorId: req.params.identityId
```
sheensantoscapadngan: same remark here

backend/src/server/routes/v1/cert-manager-access-routers/users-router.ts
```diff
@@ -245,6 +240,11 @@ export const registerCertManagerAccessUsersRouter = async (server: FastifyZodPro
           }
         }
       });
+      await server.services.pkiApplicationMembership.removeUsersFromApplicationMemberships({
+        projectId,
+        emails: req.body.emails,
+        usernames: req.body.usernames
+      });
       return {
```
sheensantoscapadngan, line 253: same remark here

backend/src/server/routes/v1/cert-manager-access-routers/users-router.ts
```diff
@@ -282,6 +277,11 @@ export const registerCertManagerAccessUsersRouter = async (server: FastifyZodPro
           metadata: { userId: membership.actorUserId as string, membershipId: membership.id }
         }
       });
+      await server.services.pkiApplicationMembership.removeActorFromApplicationMemberships({
+        projectId,
+        actorKind: "user",
+        actorId: userId
+      });
```
sheensantoscapadngan: same remark here - also let's use enums for `actorKind` please

backend/src/server/routes/v1/cert-manager-access-routers/groups-router.ts
saifsmailbox98: done

saifsmailbox98: added optional externalTx to the deletion services and new coordinating methods in pkiApplicationMembership that wrap both in a single transaction

let me know what you think of this approach and if you have a better approach for this

the callback is new here but we needed it because pkiApplicationMembership doesn't depend on the deletion services and shouldn't

backend/src/server/routes/v1/cert-manager-access-routers/identities-router.ts
saifsmailbox98: done

backend/src/server/routes/v1/cert-manager-access-routers/users-router.ts
saifsmailbox98: done

backend/src/server/routes/v1/cert-manager-access-routers/users-router.ts
saifsmailbox98: done

sheensantoscapadngan (approved):

PR #6607, opened 2026-05-26
fix(acme): redact EAB HMAC key from API responses

saifsmailbox98:
## Context

`eabHmacKey` was returned in plaintext in all ACME CA read endpoints — any user with `CertificateAuthorities:Read` (including default Member and Viewer roles) could extract it. With the EAB credentials, someone could register a new ACME account under the organization's billing with the external CA provider (DigiCert, ZeroSSL, etc.) and issue certificates against it.

- Strip `eabHmacKey` from all CA API responses via `SanitizedAcmeCertificateAuthoritySchema`
- Preserve existing key on update using `__INFISICAL_UNCHANGED__` sentinel, support explicit clearing via empty string
- Mask frontend input with `type="password"`

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/certificate-authority/acme/acme-certificate-authority-fns.ts
+++ b/backend/src/services/certificate-authority/acme/acme-certificate-authority-fns.ts
@@ -132,7 +132,7 @@ type TAcmeCertificateAuthorityFnsDeps = {
     TCertificateAuthorityDALFactory,
     "create" | "transaction" | "findByIdWithAssociatedCa" | "updateById" | "findWithAssociatedCa" | "findById"
   >;
-  externalCertificateAuthorityDAL: Pick<TExternalCertificateAuthorityDALFactory, "create" | "update">;
+  externalCertificateAuthorityDAL: Pick<TExternalCertificateAuthorityDALFactory, "create" | "update" | "findOne">;
   certificateDAL: Pick<TCertificateDALFactory, "create" | "transaction" | "updateById">;
   certificateBodyDAL: Pick<TCertificateBodyDALFactory, "create">;
   certificateSecretDAL: Pick<TCertificateSecretDALFactory, "create">;
@@ -197,7 +197,6 @@ export const castDbEntryToAcmeCertificateAuthority = (
       directoryUrl: dbConfigurationCol.directoryUrl,
       accountEmail: dbConfigurationCol.accountEmail,
       eabKid: dbConfigurationCol.eabKid,
-      eabHmacKey: dbConfigurationCol.eabHmacKey,
       dnsResolver: dbConfigurationCol.dnsResolver
     },
     status: ca.status as CaStatus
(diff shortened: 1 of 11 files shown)


backend/src/services/certificate-authority/acme/acme-certificate-authority-fns.ts
```diff
@@ -197,7 +197,6 @@ export const castDbEntryToAcmeCertificateAuthority = (
       directoryUrl: dbConfigurationCol.directoryUrl,
       accountEmail: dbConfigurationCol.accountEmail,
       eabKid: dbConfigurationCol.eabKid,
-      eabHmacKey: dbConfigurationCol.eabHmacKey,
       dnsResolver: dbConfigurationCol.dnsResolver
```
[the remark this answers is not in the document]
saifsmailbox98: moved the redaction to the zod response schema layer instead so the internal model still has it for orderCertificate

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/ExternalCaModal.tsx
```diff
@@ -929,12 +922,17 @@ export const ExternalCaModal = ({ popUp, handlePopUpToggle }: Props) => {
...
+                    isOptional
                   >
                     <Input
+                      type="password"
+                      autoComplete="new-password"
                       {...field}
-                      placeholder="dGhpc2lzYW5leGFtcGxlaG1hY2tleWZvcmRpZ2ljZXJ0YWNtZXRlc3RpbmcxMjM0NTY3ODkw"
+                      placeholder={
+                        ca
+                          ? "Leave blank to keep existing"
```
carlosmonastyrski: 🤔 is this true? Isn't the placeholder the safeguard here? (only undefined is the other way to do it, but empty strings won't have this effect right?)

backend/src/services/certificate-authority/acme/acme-certificate-authority-fns.ts
```diff
@@ -813,6 +813,15 @@ export const AcmeCertificateAuthorityFns = ({
           actor
         );
 
+        let resolvedEabHmacKey = eabHmacKey;
+        if (eabHmacKey === "__INFISICAL_UNCHANGED__" || eabHmacKey === undefined) {
```
carlosmonastyrski: let's have a constant for this, also let's add it to the docs if we want to have it as a safeguard 🤔

carlosmonastyrski (changes requested):

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/ExternalCaModal.tsx
saifsmailbox98: hmm no, the placeholder was misleading, it only shows after the user clears the sentinel, at which point submitting would actually clear the key

removed the edit-mode placeholder entirely

backend/src/services/certificate-authority/acme/acme-certificate-authority-fns.ts
saifsmailbox98: added UNCHANGED_CREDENTIAL_SENTINEL constant in both backend and frontend


> let's add it to the docs

do you mean for the API users? 

API users can just omit the field (eabHmacKey) to keep existing value, sentinel is only needed because frontend forms always send all fields

carlosmonastyrski (approved):

PR #6611, opened 2026-05-26
feat: switched to new go server

akhilmhdh:
## Context

This PR implements the new chita server and couple of tickets closed with it.

1. Fixes some security things reported
2. Resolves offline license verification
3. Assume privilege permission injection

## Steps to verify the change

1. Check out how the new server is organized
2. No need to verify the services functionality to a great extend. We will be coming back to that later. 

Start the server with ` docker compose -f docker-compose.dev.yml --profile go up` and v4 secret endpoint goes to go sidecar

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend-go/internal/server/server.go
```diff
@@ -7,49 +7,66 @@ import (
...
+// Listen starts the HTTP server.
 func (s *Server) Listen(ctx context.Context, addr string, wg *sync.WaitGroup, errc chan error) {
-	var handler http.Handler = s.mux
-	// the order is other way around. Last one wraps the previous one, so request ID is set before logging.
+	var handler http.Handler = s.router
+
+	// Middleware stack (applied in reverse order - last wraps first)
 	handler = requestLogger(handler, s.logger)
 	handler = middlewares.HTTPInfoMiddleware(handler)
+	handler = middlewares.AssumePrivilege(s.services.Platform.AssumePrivilege)(handler)
```
Thiago-AS: Shouldn't `middlewares.AssumePrivilege` be defined after the per-route `IdentityMiddleware`? Otherwise `auth.IdentityFromContext` will always be null.

backend-go/internal/services/auth/auth.go
```diff
@@ -168,10 +174,14 @@ func WithIdentity(ctx context.Context, id *Identity) context.Context {
...
+// Returns an error if identity is not present (authentication required).
+func IdentityFromContext(ctx context.Context) (*Identity, error) {
+	id, ok := ctx.Value(ctxKey{}).(*Identity)
+	if !ok || id == nil {
+		return nil, errutil.Unauthorized("Authentication required")
+	}
+	return id, nil
 }
 
 // WithAuthInfo stores the actor auth info in the context.
```
Thiago-AS: Do we need withAuthInfo if we already have withIdentity and it contains the `IdentityAuthInfo`?

Thiago-AS: Having `Identity` and `IdentityAuthInfo` inside is really confusing and error prone, can we change the nomenclature so both concepts are clearly defined? Maybe `MachineAuthInfo` or something that does not interpolate between machine identities and a identity in the application context. Or maybe `Principal` instead of `Identity`.

backend-go/pkg/chita/field.go
```diff
@@ -0,0 +1,337 @@
...
+// For requests: defaults apply if not set.
+// For responses: field may be omitted if not set (with omitempty).
+// Tracks three states: absent, null, or value present.
+//
+// Implementation uses map[bool]T so that nil maps are properly omitted
+// by json.Marshal with omitempty (structs are never considered empty).
+// - nil map = not set (omitempty skips)
+// - map[false]T = explicit null
+// - map[true]T = value present
+type Optional[T any] map[bool]T
```
Thiago-AS: I don't think using map to represent the Optional is a good idea, it's fairly expensive memory/runtime cost. Each optional field allocates a map header and potentially backing storage.

Use the same approach as you were using for the other fields:

```
type Optional[T any] struct {
	value  T
	isSet  bool
	isNull bool
}

func (o *Optional[T]) Set(v T) {
	o.value = v
	o.isSet = true
	o.isNull = false
}

func (o *Optional[T]) SetNull() {
	var zero T

	o.value = zero
	o.isSet = true
	o.isNull = true
}

func (o Optional[T]) Get() T {
	return o.value
}

func (o Optional[T]) IsSet() bool {
	return o.isSet
}

func (o Optional[T]) IsNull() bool {
	return o.isSet && o.isNull
}

func (o Optional[T]) IsRequired() bool {
	return false
}

func (o Optional[T]) IsZero() bool {
	return !o.isSet
}

func (o Optional[T]) MarshalJSON() ([]byte, error) {
	if o.isNull {
		return []byte("null"), nil
	}

	return json.Marshal(o.value)
}

func (o *Optional[T]) UnmarshalJSON(data []byte) error {
	o.isSet = true

	if string(data) == "null" {
		o.isNull = true
		return nil
	}

	return json.Unmarshal(data, &o.value)
}
```

We can even use `omitzero` if wanted:

```
type User struct {
	Name Optional[string] `json:"name,omitzero"`
}
```

backend-go/pkg/chita/schema_composite.go
```diff
@@ -0,0 +1,1170 @@
...
+	for k := range fields {
+		order = append(order, k)
+	}
+	return &ObjectSchema{
+		properties:    fields,
+		propertyOrder: order,
+	}
+}
+
+func (s *ObjectSchema) Required() *ObjectSchema {
```
Thiago-AS: I don't think it's a good idea to have both monads + primitives and pointer fields + composite. Over time, this adds unnecessary complexity when dealing with nil values and creates inconsistent semantics across the API surface. We can easily let go of raw pointer and use monads to represent that for everything, so Optional and Required can be used with all types and we dont need type-specific access plumbing.

backend-go/pkg/chita/schema_composite.go
```diff
@@ -0,0 +1,1170 @@
...
+	example     any
+	title       string
+	description string
+	deprecated  bool
+	nullable    bool
+	readOnly    bool
+	writeOnly   bool
+	refName     string
+
+	validateFn  func() []ValidationError
```
Thiago-AS: The system is mixing:
- schema definition
- validation
- field presence tracking
- JSON encoding
- request parsing
- OpenAPI generation

all into the same abstraction layer.

I don't think this is a good idea, this is becoming increasingly stateful and indirect.

We should separate schema metadata from runtime values. Schemas should ideally be:
- immutable
- declarative
- metadata-only

not runtime-bound to request state.

What I think we should have:

Pure schemas: 
```
type Schema interface {
	OpenAPI() map[string]any
	Validate(any) []ValidationError
}
```
Schemas describe structure only.

Move presence tracking entirely into values:
```
type Optional[T any] struct {
	value  T
	isSet  bool
	isNull bool
}
```
Then request structs become the source of truth.

```
type CreateUserRequest struct {
	Name  Required[string]      `json:"name"`
	Email Optional[string]      `json:"email,omitzero"`
	Age   Optional[int]         `json:"age,omitzero"`
	Meta  Optional[RawMetadata] `json:"meta,omitzero"`
}
```

Now validation operates directly on values at runtime.

Validations should inspect structs not schemas: 
`Validate(req)` not `schema.Validate()`
So kill `.Validate` and `.IsPresent`
Because:
- presence belongs to values
- schemas don't own state
- validation becomes deterministic

If we use Optional and Required monads for every field we kill the need of type-specific access plumbing:

```
strFieldAccessor
boolFieldAccessor
intFieldAccessor
```

And then we can use schemas only for OpenAPI + structural validation. so it should generate:
- OpenAPI
- declarative validation rules
but never carry runtime behavior/request state.

backend-go/pkg/chita/schema_composite.go
```diff
@@ -0,0 +1,1170 @@
...
+	description string
+	deprecated  bool
+	nullable    bool
+
+	validateFn func() []ValidationError
+}
+
+// OneOf creates a discriminated union schema with a type discriminator.
+// For validation to work, call SetActiveVariant() with the discriminator value
+// after determining which variant is active (usually after JSON parsing).
```
Thiago-AS: Yeah, that's what I'm afraid, schemas have become active, stateful objects whose correctness depends on call ordering. I see we are trying to stay free from reflection, but I don't think it's worth it with the amount of coupling is being generated. 

struct tags + reflection still feels the most idiomatic way to do that. All those libs follow that: encoding/json, database/sql, gorilla/schema and so on all follow that pattern.

Request structs should be the source of truth. 

Maybe it's worth taking a look at how github.com/danielgtaylor/huma does it, I believe they follow the Struct tags + reflection pattern.

backend-go/secret.test
Thiago-AS, line 1: Need to remove this `go test -c` artifact. I believe you forgot to add the .`*.test` to `backend-go/.gitignore`.

Thiago-AS (changes requested):

Thiago-AS (approved):

PR #6612, opened 2026-05-26
improvement(secret-sync): revam search project and group on gitlab

adilsitos:
## Context

The previous sync was searching all the projects, so a user that has a lot of projects was not able to get all the data due to timeout. Changed to allow project and group search. The display now has a limit of 20 projects and groups, if the user wants more, he need to search.

## Screenshots

image
image

## Steps to verify the change
- Create a secret gitlab sync 
- check that you can select subgroups, groups
- check that you can also see groups that are personal 

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


frontend/src/hooks/api/appConnections/gitlab/queries.tsx
```diff
@@ -26,10 +50,10 @@ export const useGitLabConnectionListProjects = (
   >
 ) => {
   return useQuery({
-    queryKey: gitlabConnectionKeys.listProjects(connectionId),
+    queryKey: gitlabConnectionKeys.listProjects(connectionId, params),
     queryFn: async () => {
       const { data } = await apiRequest.get<TGitLabProject[]>(
-        `/api/v1/app-connections/gitlab/${connectionId}/projects`
+        `/api/v1/app-connections/gitlab/${connectionId}/projects${buildProjectsQueryParams(params)}`
       );
```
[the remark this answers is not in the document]
adilsitos: removed

backend/src/services/app-connection/gitlab/gitlab-connection-fns.ts
```diff
@@ -404,3 +421,133 @@ export const listGitLabGroups = async ({
...
+
+  try {
+    const groups = await client.Groups.all({
+      topLevelOnly: true,
+      orderBy: "name",
+      sort: "asc",
+      minAccessLevel: 20
+    });
+
+    return groups.filter((g) => !g.parentId).map((g) => ({ id: g.id.toString(), name: g.name, fullPath: g.fullPath }));
```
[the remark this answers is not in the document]
adilsitos: fixed

scott-ray-wilson (changes requested):
posting general feedback before code review:


https://github.com/user-attachments/assets/ceab53d5-8bb6-4344-9a69-6e0d0e02e23e

frontend/src/components/secret-scanning/forms/SecretScanningDataSourceConfigFields/GitLabDataSourceConfigFields.tsx
scott-ray-wilson, line 1: this would be a breaking change for gitlab secret scanning (they wouldn't be able to search/list past the first 20) - we should also update these

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitLabSyncFields.tsx
scott-ray-wilson, line 1: this wasn't due to your changes but could you make the paths not have spaces for the group select? it's weird only group selection has the gaps but project selection doesn't

Image

Image

scott-ray-wilson, line 1: would it be possible to make the path search able? IE if I search "group_name/project_name" right now it wouldn't match that because we filter only against project name - would be great if we could search against full name/path

backend/src/services/app-connection/gitlab/gitlab-connection-service.ts
scott-ray-wilson, line 1: ignore file placement: side not I feel like we should either truncate the middle or just always show the group/project name first and then the path so that if anything gets truncated its the path (right now you actually can't see names that are long enough)

https://github.com/user-attachments/assets/6a10fb6a-a12f-4b29-9127-05712088c242

scott-ray-wilson (changes requested):

frontend/src/components/secret-scanning/forms/SecretScanningDataSourceConfigFields/GitLabDataSourceConfigFields.tsx
adilsitos: I've moved the limit to be defined on frontend, and if we don't send anything, the previous behavior continues.

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitLabSyncFields.tsx
adilsitos: yes! added this.

adilsitos: done. This has been changed because of the other requirement, so please let me know if this is good now.

backend/src/services/app-connection/gitlab/gitlab-connection-fns.ts
```diff
@@ -309,11 +352,15 @@ export const listGitLabProjects = async ({
...
+      pagination: "offset",
+      ...(limit !== undefined ? { perPage: limit } : {}),
+      ...(search ? { search } : {}),
       archived: false,
       includePendingDelete: false,
       membership: true,
       includeHidden: false,
-      imported: false
+      imported: false,
+      searchNamespaces: !!search
```
scott-ray-wilson, line 364: don't we need to add `maxPages` like we did for group to prevent gitbeaker from getting all of them?

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitLabSyncFields.tsx
```diff
@@ -71,21 +98,61 @@ export const GitLabSyncFields = () => {
...
+  const projectId = useWatch({ name: "destinationConfig.projectId", control });
+  const projectName = useWatch({ name: "destinationConfig.projectName", control });
+  const groupId = useWatch({ name: "destinationConfig.groupId", control });
+  const groupName = useWatch({ name: "destinationConfig.groupName", control });
 
-  const { data: groups, isLoading: isGroupsLoading } = useGitLabConnectionListGroups(connectionId, {
-    enabled: Boolean(connectionId) && scope === GitLabSyncScope.Group
-  });
+  const groupListLimit = debouncedGroupSearch ? undefined : GITLAB_SYNC_LIST_LIMIT;
+  const projectListLimit = debouncedProjectSearch ? undefined : GITLAB_SYNC_LIST_LIMIT;
```
scott-ray-wilson: I'm confused, why do we not limit if there's a search? couldn't that still lead to time outs if enough search match? ie path search 

same applies to data source form

frontend/src/components/secret-scanning/forms/SecretScanningDataSourceConfigFields/GitLabDataSourceConfigFields.tsx
scott-ray-wilson, line 1: I think we need to update the help text on this form too to state only the first few results will be returned, the need to search (like you worded for sync)

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitLabSyncFields.tsx
```diff
@@ -71,21 +98,61 @@ export const GitLabSyncFields = () => {
...
+  }, [groups, groupId, groupName]);
+
+  const projectOptions = useMemo(() => {
+    const results = projects ?? [];
+    if (projectId && projectName && !results.some((project) => project.id === projectId)) {
+      return [{ id: projectId, name: projectName }, ...results];
+    }
+    return results;
+  }, [projects, projectId, projectName]);
+
```
scott-ray-wilson, line 152: we need to do this for data source as well otherwise if I edit it looks like it's unset 

Image

scott-ray-wilson (changes requested):

backend/src/services/app-connection/gitlab/gitlab-connection-fns.ts
adilsitos: we don't have `maxPages` for `projects.all`. We are making sure we have a limit using the per page. Since we will only have one page, we have this limit there.

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitLabSyncFields.tsx
adilsitos: I thought about not limiting in search because in my opinion the amount of results would be wayy smaller, but we can have this here as well.

backend/src/services/app-connection/gitlab/gitlab-connection-fns.ts
scott-ray-wilson: but gitbeaker.all gets _**all**_ projects, you're just telling it to fetch them in 20 page batches unless you set a max page?

adilsitos: I belive it fetch only the limit (20), even we calling this using the all. Since we only get 1 page, it should only return the max amount of items in a page (in this case, 20)

frontend/src/components/secret-scanning/forms/SecretScanningDataSourceConfigFields/GitLabDataSourceConfigFields.tsx
adilsitos: updated!

backend/src/services/app-connection/gitlab/gitlab-connection-fns.ts
adilsitos: I was wrong. I did a test in another PR that had this query and the max page was breaking the type, but I checked again and this works. Added it here as well. Thanks!

scott-ray-wilson (approved):
LGTM

PR #6623, opened 2026-05-27
improvement(migrations): make migration history mismatch error actionable

PrestigePvP:
## Context

When boot-up migration history doesn't match the running image, the thrown error was hard to act on — no image version, no cause, no remediation. Operators (and customers) were left guessing.

- Drop alarmist "Invalid migration history detected" phrasing for a neutral "Database migration history does not match this image".
- Add the running image version (`INFISICAL_PLATFORM_VERSION`) inline so it's clear which build is complaining.
- Add a plain-language likely cause (rollback / mixed-version rollout).
- Label each migration list with direction ("in DB but not in this image" vs "in this image but not yet applied") and a count.
- Derive a date range from migration filename timestamps and add a concrete remediation step: deploy an image released on or after the latest unknown-migration date, or restore the DB.
- Apply the same enrichment (image version + date-bounded hint) to the "ahead" warning log, plus a structured `unknownAppliedMigrationDateRange` field for dashboards.

## Steps to verify the change

- `getMigrationDateRange` parses the `YYYYMMDD` filename prefix; all 432 existing migrations match the convention, and it returns `null` (falling back to a generic hint) if a name doesn't.
- No tests assert on the message text, so wording change is safe; `backend/src/auto-start-migrations.test.ts` covers the boot-direction helpers, which are unchanged.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: `type(scope): short description`.
- [ ] Tested locally

### DIFF
--- a/backend/src/auto-start-migrations.ts
+++ b/backend/src/auto-start-migrations.ts
@@ -43,7 +43,7 @@ const getImageVersionTag = (): string => {
   if (!version) {
     return "development";
   }
-  return `v${version}`;
+  return version.startsWith("v") ? version : `v${version}`;
 };
 
 const getMigrationDateRange = (migrationNames: string[]): { earliest: string; latest: string } | null => {
--- a/backend/src/auto-start-migrations.ts
+++ b/backend/src/auto-start-migrations.ts
@@ -38,6 +38,33 @@ const migrationConfig = {
   tableName: "infisical_migrations"
 };
 
+const getImageVersionTag = (): string => {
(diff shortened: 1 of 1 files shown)


victorvhs017 (approved):
LGTM

PR #6634, opened 2026-05-27
improvement(secrets): refine secret-manager cache key strategy

PrestigePvP:
## Context

- Reshape the get-secrets service-layer cache key from `actorId-hash(fullDto)` to `actorId-permissionFingerprint-hash(permission.rules)-requestParamsHash`, where `requestParamsHash` covers only the inputs that change the response body.
- Drop transport-only `ifNoneMatch` from the key — it was folded into the old `fullDto` hash, so every distinct client ETag forked a separate `secret-manager:*` blob. This is the main key-count reducer.
- Keep `hash(permission.rules)` in the cache key (not just the fingerprint): a cache hit returns the blob without re-filtering, and `permission.rules` fold in request-time `identity.auth` that the membership-derived fingerprint doesn't capture — so two auth contexts for the same identity must not share a cached payload.
- ETag field keys on `permissionFingerprint` only — its value is a content hash of the payload, so a field shared across auth contexts can at worst miss a 304, never serve stale content. The ETag field and cache key share `requestParamsHash`, which includes `throwOnMissingReadValuePermission` (flips partial-permission reads between masking and throwing).
- Add deterministic `ORDER BY` to the permission-fingerprint query — Postgres row order was unspecified, so the hashed fingerprint could flip between identical calls and silently miss ETag/304s.
- Add OTel cache metrics — access-result counter (`not_modified`/`hit`/`miss`), entry-size histogram, oversize-skip counter — to size `MAX_SECRET_CACHE_BYTES` and measure hit rate from real traffic before any further reshaping.

An earlier revision in this PR swapped the rules hash for `permissionFingerprint` as a size optimization; reverted after review — the fingerprint omits request-time auth context and isn't a safe substitute for the cache-blob key.

https://linear.app/infisical/issue/PLATFOR-340

## Steps to verify the change

- `npm run type:check` — passes
- `npm run lint` on the touched files — passes, 0 warnings
- By code path: the 304 (If-None-Match hit), cache hit, and cache miss returns each emit one `infisical.secret.cache.access.count` sample; cache writes record `infisical.secret.cache.entry.bytes`, and over-cap writes increment `infisical.secret.cache.oversize_skip.count`.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1226,9 +1232,29 @@ export const secretV2BridgeServiceFactory = ({
...
+      secretImportReferencesBehavior,
+      viewSecretValue,
+      throwOnMissingReadValuePermission,
+      ...params
+    });
+    const cacheKey = SecretServiceCacheKeys.getSecretsOfServiceLayer({
+      projectId,
+      version: secretDalVersion,
+      actorId,
+      permissionFingerprint,
```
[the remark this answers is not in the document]
PrestigePvP: Confirmed and fixed in 5f960ce. `permission.rules` are interpolated with request-time `identity.auth` (permission-service.ts:583-599) that the membership-derived fingerprint does not capture, and cache hits return the blob without re-filtering — so this was a real cross-auth-context leak. The cache key now folds in `hash(permission.rules)` alongside the fingerprint. (The ETag field intentionally stays fingerprint-only: its value is a content hash of the payload, so a field shared across auth contexts can at worst miss a 304, never serve stale content.)

backend/src/lib/telemetry/metrics.ts
victorvhs017, line 1: We can remove this piece. I'll incorporate this in my other PR if that's ok! The reason is that we don't want to use the infisicalMeter anymore; the way it was done would blow up our cardinality in the cloud multi-tenant environment.

https://github.com/Infisical/infisical/pull/6666

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1504,9 +1524,13 @@ export const secretV2BridgeServiceFactory = ({
...
         plainText: Buffer.from(JSON.stringify(payload))
       }).cipherTextBlob;
-      if (encryptedUpdatedCachedSecrets.byteLength < MAX_SECRET_CACHE_BYTES) {
+      const cacheBytes = encryptedUpdatedCachedSecrets.byteLength;
+      const stored = cacheBytes < MAX_SECRET_CACHE_BYTES;
+      if (stored) {
         await keyStore.setItemWithExpiry(cacheKey, SECRET_DAL_TTL(), encryptedUpdatedCachedSecrets.toString("base64"));
       }
+      recordSecretCacheWriteMetric({ bytes: cacheBytes, stored });
+      recordSecretCacheAccessMetric(SecretCacheAccessResult.MISS);
```
victorvhs017: I also added all these record calls.

victorvhs017 (changes requested):

backend/src/lib/telemetry/metrics.ts
PrestigePvP: Done — reverted the metrics.ts additions in 8813747. Will pick the cache observability up from #6666 once the bounded `InfisicalCore` meter lands.

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
PrestigePvP: Dropped the four record-call sites in 8813747; the write block is back to the original shape.

victorvhs017 (approved):

PR #6635, opened 2026-05-27
improvement: refactor identity auth forms/modals to v3 components and add the ability to manage auth from table view

scott-ray-wilson:
## Context

This PR updates identity auth management forms, components and modals to v3 and adds the ability to manage auth from the table view

## Screenshots

CleanShot 2026-05-27 at 16 44 32@2x

CleanShot 2026-05-27 at 16 44 36@2x

CleanShot 2026-05-27 at 16 44 39@2x

CleanShot 2026-05-27 at 16 44 50@2x

## Steps to verify the change

- verify all identity auth method creation/edit forms
- verify all auth methods view sheets
- verify create client secret and create access token modals / tables 
- verify confirm remove auth method
- verify both details auth method table and dropdown auth method management on table
- verify behavior at all three scopes (project, org, sub-org)
- verify table auth management only shows at relevant scope

## Type

- [ ] Fix
- [x] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/identity-v2/identity-service.ts
```diff
@@ -390,6 +390,17 @@ export const identityV2ServiceFactory = ({
...
+      Promise.all(
+        rows.map(async (row) => ({
+          ...row,
+          identity: {
+            ...row.identity,
+            activeLockoutAuthMethods: await getIdentityActiveLockoutAuthMethods(row.identity.id, keyStore)
+          }
+        }))
+      );
+
```
victorvhs017, line 404: Is it practical to merge all these Redis requests into one so we only make one round-trip to Redis instead of n rows?

backend/src/services/identity/identity-service.ts
```diff
@@ -479,7 +479,17 @@ export const identityServiceFactory = ({
...
-    return { identityMemberships: docs, totalCount };
+    const docsWithLockouts = await Promise.all(
+      docs.map(async (doc) => ({
+        ...doc,
+        identity: {
+          ...doc.identity,
+          activeLockoutAuthMethods: await getIdentityActiveLockoutAuthMethods(doc.identity.id, keyStore)
+        }
+      }))
+    );
```
victorvhs017, line 491: same question

backend/src/services/membership-identity/membership-identity-service.ts
```diff
@@ -353,7 +357,17 @@ export const membershipIdentityServiceFactory = ({
...
+    const withLockouts = await Promise.all(
+      filtered.map(async (el) => ({
+        ...el,
+        identity: {
+          ...el.identity,
+          activeLockoutAuthMethods: await getIdentityActiveLockoutAuthMethods(el.identity.id, keyStore)
+        }
+      }))
+    );
+    return { ...memberships, data: withLockouts };
```
victorvhs017, line 391: same here

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/lockout/LockoutTab.tsx
victorvhs017, line 1: What do you think about setting the minimum value as a fixed placeholder here when the user deletes the default value?

Image

Then we can guarantee that the helper text won't have any gaps.

victorvhs017, line 1: We can also allow only numbers in the input now that we are putting the time unit on the select

Image

victorvhs017, line 1: This alignment feels a bit odd to me

Image

victorvhs017, line 1: Field validation is happening even when the toggle is disabled

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityAliCloudAuthForm.tsx
victorvhs017, line 1: IP field not highlighted on validation

Image

victorvhs017, line 1: Should we move away from this modal completely?

Dropping the question here because I don't know if we are tackling it in this PR or not, and this one was overlooked. But it does feel like we should.

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityAwsAuthForm.tsx
victorvhs017, line 1: No error notification when the trusted IP section is empty and you try to save.

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityAzureAuthForm.tsx
```diff
@@ -203,194 +220,221 @@ export const IdentityAzureAuthForm = ({
...
+                      handlePopUpOpen("upgradePlan", {
+                        featureName: "IP allowlisting"
+                      });
+                    }}
+                  >
+                    <XIcon />
+                  </IconButton>
+                </div>
+              ))}
+              <Button
```
victorvhs017: I feel like this token section and this IPs section could each be their own shared components; we are repeating them in a lot of places.

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityAzureAuthForm.tsx
victorvhs017, line 1: If we have the placeholder values here, why don't we validate the other 2 fields as we do for the first one?

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityKubernetesAuthForm.tsx
victorvhs017, line 1: Why does this select open upwards?

Image

victorvhs017, line 1: Odd space on these tooltips

Image

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityOciAuthForm.tsx
victorvhs017, line 1: Odd tooltip:

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentitySpiffeAuthForm.tsx
victorvhs017, line 1: This placeholder should be a tooltip instead.

Image

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityUniversalAuthForm.tsx
victorvhs017, line 1: This copy button feels a bit out of place to me.

Image

Maybe this one-line approach would make more sense?

Image

frontend/src/views/IdentityAuthMethods/helpers/IdentityAuthLockoutFields.tsx
victorvhs017, line 1: Maybe access token usage could also be its own section, too? We have it in all methods. The IPs could also be inside it, maybe?

Image

victorvhs017 (changes requested):

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityAliCloudAuthForm.tsx
scott-ray-wilson: modal comes from mutation query client so separate address as far as design but we should exhaust form validation to the extent we can on the client side so we don't show it (which we can improve in this scenario)

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentityKubernetesAuthForm.tsx
scott-ray-wilson: there's two positions for selects - popper which is based of top/bottom and the default which opens the menu directly above the select aligned with the current item; it only looks above because you're on the last item

generally I prefer popper, this just already wasn't and didn't notice

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/lockout/LockoutTab.tsx
scott-ray-wilson: hmmm I don't think we should set a min or default given the nature of this, I'd rather just say "Required" and enforce number - will udpate!

scott-ray-wilson: interesting, does it still look like that for you? 
CleanShot 2026-06-01 at 15 10 31@2x

scott-ray-wilson: I think this is fine; we don't clear the values when they disable it, we preserve the existing values so they can toggle it on again so it should always be validated; this only happens if someone intentionally leaves it blank - otherwise it'd save an invalid value

victorvhs017: sorry, I meant for this text!

image

victorvhs017: Looking good now!

victorvhs017: But then I can't update the form with the toggle off

victorvhs017 (approved):

PR #6660, opened 2026-05-28
feat(roles): convert built-in org and project roles to editable database rows

victorvhs017:
## Context

Converts built-in org and project roles (Member, Viewer, No Access, SSH Host Bootstrapper, Cryptographic Operator) from hardcoded in-memory definitions to editable database rows with an `isBuiltIn` flag, while keeping Admin as the only purely hardcoded role. This enables customization of built-in role permissions while preventing deletion/renaming of platform roles.

Key changes:
- **Migration** adds `isBuiltIn` column to roles table, seeds built-in roles for all existing orgs/projects, and backfills `customRoleId` on membership roles.
- **Permission service** now resolves role permissions from DB rows instead of switch-casing on slug, except for Admin which stays hardcoded.
- **Role slug validation** relaxed from blocking all built-in slugs to only blocking `admin`, allowing built-in roles to exist as DB rows.
- **Sub-org creation** now seeds Member/No Access roles (was missing), with sub-org-specific permission filtering (excludes billing, SSO, LDAP, SCIM, etc.).
- **Org/sub-org member invite modals** now use dynamic project role fetching instead of hardcoding the `ProjectMembershipRole` enum.
- **Legacy permission migration** in frontend: `DescribeAndReadValue` and `GrantPrivileges` actions are transparently migrated to their granular replacements during form processing.
- **Project templates** sorted alphabetically, default roles seeded when none provided, reserved role slug/name validation removed (now handled at DB level).
- **UI** prevents deletion of built-in roles with tooltip, shows "Admin role permissions cannot be changed" for admin edits.

### Org Roles
- [x] List org roles — Admin, Member, No Access all appear with "Platform" type badge
- [x] View Admin role — all policies shown but disabled, no edit/delete allowed
- [x] Edit Member role permissions — save succeeds, changes persist
- [x] Edit No Access role permissions — save succeeds
- [x] Attempt to delete a built-in role — tooltip shows "Platform roles cannot be deleted", button disabled
- [x] Delete a custom org role — succeeds
- [x] Duplicate an org role (including Admin) — succeeds without errors

### Project Roles
- [x] List project roles — Admin, Member, Viewer, No Access appear as built-in
- [x] For KMS project — Cryptographic Operator also appears
- [x] Edit built-in project role permissions — save succeeds
- [x] Attempt to delete a built-in project role — blocked with tooltip
- [x] Create a custom project role — succeeds
- [x] Duplicate a project role (including Admin) — no legacy permission errors

### Project Templates
- [x] List templates — sorted alphabetically by name
- [x] Create a template with no roles — defaults seed correctly (Admin, Member, Viewer, No Access)
- [x] Create a template with custom roles — custom roles only (no duplicate built-in injection)
- [x] Edit a template — update succeeds without type errors

### Default Roles by Project Type
- [x] Create a SecretManager project — Member, Viewer, No Access roles seeded
- [x] Create a KMS project — additionally seeds Cryptographic Operator

### Org/Sub-org Creation & Role Seeding
- [x] Create a new organization — Admin (hardcoded) + Member, No Access seeded as DB rows
- [x] Create a new sub-organization — Admin + Member, No Access seeded; Member role does NOT contain billing/SSO/LDAP/SCIM/SubOrganization permissions
- [x] Edit Member role in sub-org — save succeeds (no "invalid permission subjects" error)

### Invite Users & Role Assignment
- [x] Root org: Invite user — org role dropdown shows all org roles; project role dropdown dynamically loads roles for selected project
- [x] Root org: Select multiple projects — falls back to built-in roles (Admin, Member, Viewer, No Access)
- [x] Sub-org: Add member — org role dropdown correct; project role dropdown uses the same dynamic logic
- [ ] Sub-org: Project grouping labels — PAM, Secret Scanning, AI show correct names (not "Other")

### SSO/LDAP/SCIM Role Mapping
- [x] SAML login — default org membership role resolved correctly (uses role ID from DB, not slug)
- [x] LDAP login — same default role resolution works
- [x] OIDC login — same default role resolution works
- [x] SCIM user provisioning — role assignment works with built-in DB roles

### Machine Identities
- [x] Add identity to project with Member role — works
- [x] Add identity to project with custom role — works
- [x] Identity role shows correct permissions

### Groups
- [x] Add group to project with built-in role — works
- [x] Group members inherit correct permissions from built-in DB role

### Temporary Roles
- [x] Assign a temporary role to a user — works with both built-in and custom roles
- [x] Temporary role expires correctly

### Legacy Permission Migration (Frontend)
- [x] Open a role with legacy `DescribeAndReadValue` permission — migrated to `DescribeSecret` + `ReadValue`
- [x] Open a role with legacy `GrantPrivileges` on identities — migrated to `AssignRole` + `AssignAdditionalPrivileges`
- [x] Open a role with legacy `GrantPrivileges` on members — same migration
- [x] Open a role with legacy `GrantPrivileges` on groups — migrated to `AssignRole`

Repeat all the tests above with `rbac` permission `false`(the ones you can do).

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


victorvhs017:
@greptileai re-review

victorvhs017:
@veria-ai re-review

backend/src/db/migrations/20260525184551_add-is-built-in-to-roles.ts
scott-ray-wilson, line 1: ignore file placement: 

we need to seed the PKI roles for new projects; right now if you create a new org it only has admin

Image

scott-ray-wilson, line 1: ignore file placement: it seems like permission audit feature is broken for migrated platform roles; this user is a member but the audit shows no permissions

Image

Image

scott-ray-wilson, line 1: I think this is related to the permission audit feature not working but the migrated platform roles [1] and platform roles created post migration [2] drift; pre migration has the original slug while the post have "custom" - I would imagine we would want these to be consistent as they are equivalents?

Image

backend/src/db/migrations/20260525184551_add-is-built-in-to-roles.ts
```diff
@@ -0,0 +1,298 @@
...
+      "no-access",
+      "ssh-host-bootstrapper",
+      "cryptographic-operator"
+    ];
+    const alreadySeededProjectRows = await trx(TableName.Role)
+      .whereNull("orgId")
+      .whereNotNull("projectId")
+      .whereIn("slug", projectBuiltInSlugsToCheck)
+      .select("projectId", "slug");
+    const alreadySeededProjectSet = new Set(alreadySeededProjectRows.map((r) => `${String(r.projectId)}:${r.slug}`));
```
scott-ray-wilson: if a migration failed previously this could potentially pull an absurd amount of rows into memory 3-4 rows per project and 2 per org above - could lead to OOM - maybe a way to do with upsert/conflict check to avoid pulling into memory?

backend/src/db/migrations/20260525184551_add-is-built-in-to-roles.ts
```diff
@@ -0,0 +1,298 @@
...
+  // so a retry on next startup is always safe.
+  await knex.transaction(async (backfillTrx) => {
+    // org-scoped
+    await backfillTrx.raw(`
+      UPDATE "${TableName.MembershipRole}" mr
+      SET "customRoleId" = r.id
+      FROM "${TableName.Membership}" m,
+           "${TableName.Role}" r
+      WHERE mr."membershipId" = m.id
+        AND r.slug = mr.role
```
scott-ray-wilson: ah I'm guessing this is why we need migrated roles with that...

backend/src/db/migrations/20260525184551_add-is-built-in-to-roles.ts
```diff
@@ -0,0 +1,298 @@
...
+          permissions: packedSshBootstrap,
+          isBuiltIn: true,
+          createdAt: now,
+          updatedAt: now
+        });
+      }
+
+      if (type === ProjectType.KMS && !alreadySeededProjectSet.has(`${projectId}:cryptographic-operator`)) {
+        rows.push({
+          id: uuidv4(),
```
scott-ray-wilson: for all these rows, can't we just let pg handle uuid? unlikely to collide but I think would be safer?

backend/src/ee/routes/v1/project-template-router.ts
scott-ray-wilson, line 1: ignore file placement: 

if I rename a role in a project template it doesn't show as it's new name in the selector

Image

backend/src/ee/services/permission/default-roles.ts
scott-ray-wilson, line 1: I cannot for the life of me remember what is was but I remember we removed legacy actions once form roles and we had to add them back... not sure still applicable but I no we had to revert it once

backend/src/db/migrations/20260525184551_add-is-built-in-to-roles.ts
```diff
@@ -0,0 +1,298 @@
...
+      FROM "${TableName.Role}" r
+      WHERE r.slug = o."defaultMembershipRole"
+        AND r."orgId" = o.id
+        AND r."projectId" IS NULL
+        AND o."defaultMembershipRole" <> 'admin'
+    `);
+  });
+}
+
+export async function down(knex: Knex): Promise<void> {
```
scott-ray-wilson, line 257: down migration fails

```
migration file "20260525184551_add-is-built-in-to-roles.ts" failed
migration failed with error:
    UPDATE "organizations" o
    SET "defaultMembershipRole" = r.slug
    FROM "roles" r
    WHERE r.id = o."defaultMembershipRole"
      AND r."orgId" = o.id
      AND r."projectId" IS NULL
      AND r.slug IN ($1, $2)
   - operator does not exist: uuid = character varying

    UPDATE "organizations" o
    SET "defaultMembershipRole" = r.slug
    FROM "roles" r
    WHERE r.id = o."defaultMembershipRole"
      AND r."orgId" = o.id
      AND r."projectId" IS NULL
      AND r.slug IN ($1, $2)
   - operator does not exist: uuid = character varying
error:
    UPDATE "organizations" o
    SET "defaultMembershipRole" = r.slug
    FROM "roles" r
    WHERE r.id = o."defaultMembershipRole"
      AND r."orgId" = o.id
      AND r."projectId" IS NULL
      AND r.slug IN ($1, $2)
   - operator does not exist: uuid = character varying
    at Parser.parseErrorMessage (/Users/scott/Documents/projects/infisical/backend/node_modules/pg-protocol/src/parser.ts:368:69)
    at Parser.handlePacket (/Users/scott/Documents/projects/infisical/backend/node_modules/pg-protocol/src/parser.ts:187:21)
    at Parser.parse (/Users/scott/Documents/projects/infisical/backend/node_modules/pg-protocol/src/parser.ts:102:30)
    at Socket.<anonymous> (/Users/scott/Documents/projects/infisical/backend/node_modules/pg-protocol/src/index.ts:7:48)
    at Socket.emit (node:events:508:28)
    at Socket.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Socket.Readable.push (node:internal/streams/readable:390:5)
    at TCP.onStreamRead (node:internal/stream_base_commons:189:23)
```

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -208,7 +208,6 @@ const buildAdminPermissionRules = () => {
       ProjectPermissionMemberActions.Edit,
       ProjectPermissionMemberActions.Delete,
       ProjectPermissionMemberActions.Read,
-      ProjectPermissionMemberActions.GrantPrivileges,
```
scott-ray-wilson, line 212: I don't think we can remove these from the admin role for the old permissions system; otherwise they would fail to grant permissions on a user that had this legacy priv

backend/src/db/migrations/20260525184551_add-is-built-in-to-roles.ts
```diff
@@ -0,0 +1,298 @@
...
+  // ── Add isBuiltIn column ────
+  const hasIsBuiltIn = await knex.schema.hasColumn(TableName.Role, "isBuiltIn");
+  if (!hasIsBuiltIn) {
+    await knex.schema.alterTable(TableName.Role, (t) => {
+      t.boolean("isBuiltIn").notNullable().defaultTo(false);
+    });
+  }
+
+  // ── Seed built-in roles ────
+  await knex.transaction(async (trx) => {
```
scott-ray-wilson: aren't these transactions moot since the whole migration is in a transaction?

backend/src/ee/services/permission/default-roles.ts
scott-ray-wilson, line 1: ignore file placement:

I feel like it's a little weird I can make no-access not not access but can't change the slug or delete the role; what was the deciding point on not allowing these roles to be deleted/slug updated?

scott-ray-wilson, line 1: not sure related to your PR but duplicating admin role for sub-orgs fails

Image

scott-ray-wilson, line 1: mismatch between select trigger and dropdown menu label display for overridden names, I think this might only when it's not user selected but the default value 

Image

scott-ray-wilson (changes requested):

backend/src/ee/services/permission/default-roles.ts
victorvhs017: Yeah, I realized this when going through your review, the scope of the access is not the same, and we would be removing privileges by doing that. Reverted in the code.

victorvhs017: Initially, I allowed deletion and slug updates, but then I found some blockers.

A bunch of stuff resolves these by slug, not id: the org default membership role, SAML group provisioning (looks up no-access directly), project bootstrap, SCIM, etc. Renaming or deleting would break those lookups and orphan any default-role config pointing at them. 

Delete is also an issue: if you drop a platform role on Enterprise, downgrade, and there's no way to get it back.

Making the slug itself editable is doable but means reworking every slug-based lookup to be id-based, which widens the scope too much for this PR.

scott-ray-wilson (dismissed):
LGTM!

PR #6680, opened 2026-05-30
feat: code signing revamp

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/signer/signer-service.ts
```diff
@@ -111,57 +212,299 @@ export const signerServiceFactory = ({
...
+              scopeOrgId: dto.actorOrgId,
+              scopeProjectId: dto.projectId,
+              actorUserId: dto.actorId,
+              scopeResourceType: ResourceType.Signer,
+              scopeResourceId: signer.id,
+              isActive: true
+            },
+            tx
+          );
+          await membershipRoleDAL.create({ membershipId: newMembership.id, role: ApplicationMembershipRole.Admin }, tx);
```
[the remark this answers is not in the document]
carlosmonastyrski: A caller with `CodeSigners.Create` is creating a new resource and becomes its administrator so they can manage what they just created. The initial `members` block isn't a permission bypass — the creator just became signer admin, and signer admins have `ManageMembers`, so adding members during creation is the same operation as create-then-add

backend/src/db/migrations/20260528124443_signer-revamp.ts
```diff
@@ -0,0 +1,295 @@
...
+          policyId = signer.approvalPolicyId;
+        }
+      }
+
+      const stepIds = (await tx(TableName.ApprovalPolicySteps).where({ policyId }).select<{ id: string }[]>("id")).map(
+        (r) => r.id
+      );
+      if (stepIds.length > 0) {
+        await tx(TableName.ApprovalPolicyStepApprovers).whereIn("policyStepId", stepIds).delete();
+      }
```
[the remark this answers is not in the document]
carlosmonastyrski: Intended

backend/src/services/signer/signer-issuance-fns.ts
```diff
@@ -0,0 +1,178 @@
...
+      new x509.KeyUsagesExtension(
+        // eslint-disable-next-line no-bitwise
+        x509.KeyUsageFlags.digitalSignature | x509.KeyUsageFlags.keyEncipherment
+      )
+    ]
+  });
+  const csrPem = csrObj.toString("pem");
+
+  const signResult = await internalCertificateAuthorityService.signCertFromCa({
+    isInternal: true,
```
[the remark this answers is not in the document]
carlosmonastyrski: False positive

backend/src/services/signer/signer-service.ts
```diff
@@ -111,57 +212,299 @@ export const signerServiceFactory = ({
...
+            scopeProjectId: dto.projectId,
+            scopeResourceType: ResourceType.Signer,
+            scopeResourceId: signer.id,
+            isActive: true,
+            actorUserId: member.kind === "user" ? member.id : null,
+            actorIdentityId: member.kind === "identity" ? member.id : null,
+            actorGroupId: member.kind === "group" ? member.id : null
+          };
+          // eslint-disable-next-line no-await-in-loop
+          const newMembership = await membershipDAL.create(membershipFields, tx);
```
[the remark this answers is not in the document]
carlosmonastyrski: Creation is only possible if you are a cert-manager product admin

backend/src/db/migrations/20260528124443_signer-revamp.ts
```diff
@@ -0,0 +1,295 @@
...
+    await knex.schema.alterTable(TableName.PkiSigners, (t) => {
+      if (!hasCaId) {
+        t.uuid("caId").nullable();
+        t.foreign("caId").references("id").inTable(TableName.CertificateAuthority).onDelete("SET NULL");
+        t.index("caId");
+      }
+      if (!hasCommonName) t.string("commonName", 256).nullable();
+      if (!hasTtl) t.integer("certificateTtlDays").nullable();
+      if (!hasRenewBefore) t.integer("renewBeforeDays").nullable();
+      if (!hasFailureReason) t.text("failureReason").nullable();
```
sheensantoscapadngan: is this failure reason for issuance of certificate or for signing operations? we should ideally be very descriptive here with the names.. I think same remark can be said for "renewBeforeDays". Like it's not really descriptive with what's being renewed exactly

backend/src/db/migrations/20260528124443_signer-revamp.ts
```diff
@@ -0,0 +1,295 @@
...
+      if (!hasKeyAlgorithm) t.string("keyAlgorithm", 64).notNullable().defaultTo("RSA_2048");
```
sheensantoscapadngan, line 33: same here..

backend/src/db/migrations/20260528124443_signer-revamp.ts
```diff
@@ -0,0 +1,295 @@
...
+    t.uuid("certificateId").nullable().alter();
+  });
+
+  if (!(await knex.schema.hasColumn(TableName.ApprovalRequestGrants, "grantedByUserId"))) {
+    await knex.schema.alterTable(TableName.ApprovalRequestGrants, (t) => {
+      t.uuid("grantedByUserId").nullable();
+      t.foreign("grantedByUserId").references("id").inTable(TableName.Users).onDelete("SET NULL");
+      t.index("grantedByUserId");
+    });
+  }
```
sheensantoscapadngan: what do we need this for? don't we track this information in `approval_request_approvals`?

backend/src/db/migrations/20260528124443_signer-revamp.ts
```diff
@@ -0,0 +1,295 @@
+/* eslint-disable no-await-in-loop */
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+const DEFAULT_RENEW_BEFORE_DAYS = 7;
+const DEFAULT_CERT_TTL_DAYS = 365;
+const SIGNER_SCOPE_TYPE = "signer";
+const CERT_CODE_SIGNING_POLICY_TYPE = "cert-code-signing";
+const SIGNER_ISSUANCE_JOBS_TABLE = "pki_signer_issuance_jobs";
```
sheensantoscapadngan: would be best to specifically mention "certificate" here otherwise it's not really obvious

backend/src/db/schemas/models.ts
```diff
@@ -362,7 +363,8 @@ export enum ApplicationMembershipRole {
 }
 
 export enum ResourceType {
-  CertificateApplication = "certificate-application"
+  CertificateApplication = "certificate-application",
+  Signer = "signer"
 }
```
sheensantoscapadngan, line 368: is this exposed externally? if so perhaps we should prefix this with PKI?

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1097,3 +1092,95 @@ export const applicationAdminPermissions = buildApplicationAdminPermissionRules(
...
+
+  can(
+    [
+      ProjectPermissionMemberActions.Read,
+      ProjectPermissionMemberActions.Create,
+      ProjectPermissionMemberActions.Edit,
+      ProjectPermissionMemberActions.Delete
+    ],
+    ResourcePermissionSub.Member
+  );
```
sheensantoscapadngan: will this give signer admins the permission to delete members at the product level directly? this is confusing since you're using Project actions for a resource subject...

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1097,3 +1092,95 @@ export const applicationAdminPermissions = buildApplicationAdminPermissionRules(
...
+      ResourcePermissionSignerActions.Read,
+      ResourcePermissionSignerActions.Sign,
+      ResourcePermissionSignerActions.RequestSign,
+      ResourcePermissionSignerActions.ExportCertificate
+    ],
+    ResourcePermissionSub.Signer
+  );
+
+  can([ProjectPermissionMemberActions.Read], ResourcePermissionSub.Member);
+  can([ProjectPermissionAuditLogsActions.Read], ProjectPermissionSub.AuditLogs);
```
sheensantoscapadngan: should't this be scoped to the signer the user is assigned to directly rather than project-wide? or are we scoping that in another way?

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1097,3 +1092,95 @@ export const applicationAdminPermissions = buildApplicationAdminPermissionRules(
...
+const buildSignerAuditorPermissionRules = () => {
+  const { can, rules } = new AbilityBuilder<MongoAbility<ResourcePermissionSet>>(createMongoAbility);
+
+  can(
+    [ResourcePermissionSignerActions.Read, ResourcePermissionSignerActions.ExportCertificate],
+    ResourcePermissionSub.Signer
+  );
+
+  can([ProjectPermissionMemberActions.Read], ResourcePermissionSub.Member);
+  can([ProjectPermissionAuditLogsActions.Read], ProjectPermissionSub.AuditLogs);
```
sheensantoscapadngan: same remark here...

there should ideally be a clear and strong separation between project-level and resource-level permissions

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1097,3 +1092,95 @@ export const applicationAdminPermissions = buildApplicationAdminPermissionRules(
...
+
+  can([ResourcePermissionSignerActions.Read], ResourcePermissionSub.Signer);
+
+  can(
+    [
+      ProjectPermissionMemberActions.Read,
+      ProjectPermissionMemberActions.Create,
+      ProjectPermissionMemberActions.Edit,
+      ProjectPermissionMemberActions.Delete
+    ],
```
sheensantoscapadngan: same concern here.. why are we using project actions for resource subjects?

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -126,33 +131,53 @@ const buildProjectPermissionRules = (projectUserRoles: TBuildProjectPermissionDT
...
+  if (resourceType === ResourceType.Signer) {
+    switch (role) {
+      case ApplicationMembershipRole.Admin:
+        return signerAdminPermissions;
+      case ApplicationMembershipRole.Operator:
+        return signerOperatorPermissions;
+      case ApplicationMembershipRole.Auditor:
+        return signerAuditorPermissions;
+      case ApplicationMembershipRole.Custom:
+        throw new BadRequestError({ message: "Custom resource-level roles are not supported yet" });
```
sheensantoscapadngan: this is misleading... why are we using an application enum for signers

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1097,3 +1092,95 @@ export const applicationAdminPermissions = buildApplicationAdminPermissionRules(
...
+    ResourcePermissionSub.Signer
+  );
+
+  can([ProjectPermissionMemberActions.Read], ResourcePermissionSub.Member);
+  can([ProjectPermissionAuditLogsActions.Read], ProjectPermissionSub.AuditLogs);
+
+  return rules;
+};
+
+const buildSignerProjectAdminFallbackRules = () => {
```
sheensantoscapadngan: shouldn't this be named "buildProjectAdminSignerFallbackRules" instead since this is referring to the signer permissions of a project admin? if we change this, we should adjust the naming for applications as well

backend/src/ee/services/permission/resource-permission.ts
```diff
@@ -21,7 +21,23 @@ export enum ResourcePermissionSub {
...
-  Role = "role"
+  Role = "role",
+  Signer = "signer"
+}
+
+export enum ResourcePermissionSignerActions {
+  Read = "read",
+  Edit = "edit",
+  Delete = "delete",
+  EnableDisable = "enable-disable",
```
sheensantoscapadngan: hmmm feels a bit off. maybe this could be "manage-status"? we're already using that pattern below

backend/src/ee/services/permission/resource-permission.ts
```diff
@@ -79,6 +95,7 @@ export type ResourcePermissionSet =
   | [ProjectPermissionApprovalRequestGrantActions, ResourcePermissionSub.ApprovalRequestGrants]
   | [ProjectPermissionMemberActions, ResourcePermissionSub.Member]
   | [ProjectPermissionActions, ResourcePermissionSub.Role]
```
sheensantoscapadngan: just noticed this now... why do we use ProjectPermissions in places where it seems like it should be "ResourcePermission"?

backend/src/server/plugins/error-handler.ts
```diff
@@ -256,11 +256,10 @@ export const fastifyErrHandler = fastifyPlugin(async (server: FastifyZodProvider
         .status(error.status)
         .send({
           reqId: req.id,
-          error: error.name,
```
sheensantoscapadngan, line 259: wait why are we touching this?

backend/src/server/routes/v1/signer-membership-router.ts
```diff
@@ -0,0 +1,783 @@
...
+    method: "POST",
+    url: "/:signerId/identities",
+    config: { rateLimit: writeLimit },
+    schema: {
+      hide: false,
+      operationId: "addSignerIdentityMember",
+      tags: [ApiDocsTags.PkiSigners],
+      description: "Add a machine identity to a signer",
+      params: SignerIdParamsSchema,
+      body: z.object({ identityId: z.string().uuid(), role: z.string().min(1).default("operator") }),
```
sheensantoscapadngan: should be using enums here

backend/src/server/routes/v1/signer-membership-router.ts
```diff
@@ -0,0 +1,783 @@
...
+      .string()
+      .email()
+      .array()
+      .default([])
+      .refine((val) => val.every((el) => el === el.toLowerCase()), "Email must be lowercase"),
+    role: z.string().min(1).default("operator")
+  })
+  .refine((val) => val.userIds.length + val.emails.length > 0, {
+    message: "Provide at least one userId or email."
+  });
```
sheensantoscapadngan: why do we define this separately if we're only ever using it once? we're also not exporting this. seems like it would be a lot easier if we just defined this inline. same remark for the other entities above (if relevant)

backend/src/server/routes/v1/signer-membership-router.ts
```diff
@@ -0,0 +1,783 @@
...
+const AddUsersBodySchema = z
+  .object({
+    userIds: z.string().uuid().array().default([]),
+    emails: z
+      .string()
+      .email()
+      .array()
+      .default([])
+      .refine((val) => val.every((el) => el === el.toLowerCase()), "Email must be lowercase"),
+    role: z.string().min(1).default("operator")
```
sheensantoscapadngan: enum

backend/src/server/routes/v1/signer-membership-router.ts
```diff
@@ -0,0 +1,783 @@
...
+          type: EventType.REMOVE_PKI_SIGNER_MEMBER,
+          metadata: { signerId: req.params.signerId, kind: SignerMemberKind.Identity, memberId: req.params.identityId }
+        }
+      });
+
+      return { membershipId: result.membershipId, signerId: result.signerId };
+    }
+  });
+
+  server.route({
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

we should separate this into multiple files the same way we did it for applications

backend/src/server/routes/v1/signer-membership-router.ts
```diff
@@ -0,0 +1,783 @@
...
+    method: "GET",
+    url: "/:signerId/effective-users",
+    config: { rateLimit: readLimit },
+    schema: {
+      hide: false,
+      operationId: "listSignerEffectiveUserMembers",
+      tags: [ApiDocsTags.PkiSigners],
+      description:
+        "List users that can act on this signer either directly or via a group that's a signer member. Role is the highest the user holds across all paths.",
+      params: SignerIdParamsSchema,
```
sheensantoscapadngan: what "role" are you referring to here? and why is there a concept of "higher"? would this still make sense once we introduce custom roles?

backend/src/server/routes/v1/signer-membership-router.ts
```diff
@@ -0,0 +1,783 @@
...
+        projectId,
+        event: {
+          type: EventType.REMOVE_PKI_SIGNER_MEMBER,
+          metadata: { signerId: req.params.signerId, kind, memberId }
+        }
+      });
+
+      return { membershipId: result.membershipId, signerId: result.signerId };
+    }
+  });
```
sheensantoscapadngan: hold on... aren't these all redundant since we already have type-specific routes?

backend/src/server/routes/v1/signer-router.ts
```diff
@@ -26,8 +37,41 @@ export const registerSignerRouter = async (server: FastifyZodProvider) => {
...
+                approverUserIds: z.array(z.string().uuid()).default([]),
+                approverGroupIds: z.array(z.string().uuid()).default([])
+              })
+            ),
+            constraints: z
+              .object({
+                maxSignings: z.number().int().min(1).nullable().optional(),
+                maxWindowDuration: z.string().nullable().optional()
+              })
+              .optional()
```
sheensantoscapadngan: I thought they'd have to define either maxSignings or maxWindowDuration? are we missing a validation check here?

backend/src/server/routes/v1/signer-router.ts
```diff
@@ -180,6 +224,44 @@ export const registerSignerRouter = async (server: FastifyZodProvider) => {
...
+    handler: async (req) => {
+      return server.services.pkiSigner.getMyPermissions({
+        signerId: req.params.signerId,
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId
+      });
+    }
+  });
```
sheensantoscapadngan: just noticed that we started using this pattern for applications..

why do we need it? how come we never had to do this in the other places outside of PKI ?

backend/src/server/routes/v1/signer-router.ts
```diff
@@ -26,8 +37,41 @@ export const registerSignerRouter = async (server: FastifyZodProvider) => {
...
         description: z.string().trim().max(256).optional(),
-        certificateId: z.string().uuid(),
-        approvalPolicyId: z.string().uuid().optional()
+        caId: z.string().uuid().optional(),
+        commonName: z.string().trim().min(1).max(256).optional(),
+        certificateTtlDays: z.number().int().min(1).max(3650).optional(),
+        renewBeforeDays: z.number().int().min(1).max(30).nullable().optional(),
+        keyAlgorithm: SignerKeyAlgorithmEnum.optional(),
+        certificateId: z.string().uuid().optional(),
+        approvalPolicyId: z.string().uuid().optional(),
```
sheensantoscapadngan: same remark here about descriptive naming.. maybe put all inside an object if it feels too redundant to add prefixes again and again?

backend/src/server/routes/v1/signer-router.ts
```diff
@@ -404,6 +484,419 @@ export const registerSignerRouter = async (server: FastifyZodProvider) => {
...
+        event: {
+          type: EventType.DISABLE_PKI_SIGNER,
+          metadata: { signerId: signer.id, name: signer.name }
+        }
+      });
+
+      return signer;
+    }
+  });
+
```
sheensantoscapadngan: can just unify this into a single status endpoint for the signer

backend/src/server/routes/v1/signer-router.ts
```diff
@@ -404,6 +484,419 @@ export const registerSignerRouter = async (server: FastifyZodProvider) => {
...
+            requiredApprovals: z.number().int().min(1),
+            approverUserIds: z.array(z.string().uuid()).default([]),
+            approverGroupIds: z.array(z.string().uuid()).default([])
+          })
+        ),
+        constraints: z
+          .object({
+            maxSignings: z.number().int().min(1).nullable().optional(),
+            maxWindowDuration: z.string().nullable().optional()
+          })
```
sheensantoscapadngan: if this is a full rerewite of the approval policy based on the passed in values then this should be a "PUT" instead of a "PATCH"

backend/src/server/routes/v1/signer-router.ts
```diff
@@ -404,6 +484,419 @@ export const registerSignerRouter = async (server: FastifyZodProvider) => {
...
+        projectId: req.internalCertManagerProjectId,
+        event: {
+          type: EventType.PKI_SIGNER_REVOKE_REQUEST,
+          metadata: { signerId: req.params.signerId, requestId: req.params.requestId }
+        }
+      });
+
+      return result;
+    }
+  });
```
sheensantoscapadngan: it would be best if we broke this down into a different file altogether. this `signer-router.ts` file is getting pretty big

backend/src/services/approval-policy/approval-policy-enums.ts
```diff
@@ -6,7 +6,8 @@ export enum ApprovalPolicyType {
 
 export enum ApprovalPolicyScope {
   Project = "project",
-  PkiApplication = "pki-application"
+  PkiApplication = "pki-application",
+  Signer = "signer"
 }
```
sheensantoscapadngan, line 11: pki-signer maybe since this currently sounds pretty generic? (could be mistaken for a KMS entity as an example)

backend/src/services/approval-policy/approval-request-dal.ts
```diff
@@ -162,6 +163,112 @@ export const approvalRequestDALFactory = (db: TDbClient) => {
...
+      if (statuses?.length) {
+        void query.whereIn("status", statuses);
+      }
+      const result = (await query.count("* as count").first()) as { count: string | number } | undefined;
+      return Number(result?.count ?? 0);
+    } catch (error) {
+      throw new DatabaseError({ error, name: "CountSignerRequests" });
+    }
+  };
+
```
sheensantoscapadngan: Isn't this a bit of an anti-pattern, because now we have signing-specific logic inside the approval request DAL, which is supposed to be a generic approval system? Maybe we can create a dedicated DAL for signing instead?

backend/src/services/certificate-authority/acme/acme-certificate-authority-fns.ts
```diff
@@ -416,6 +474,19 @@ export const orderCertificate = async (
...
+  };
+  if (acmeCa.configuration.accountEmail) {
+    accountPayload.contact = [`mailto:${acmeCa.configuration.accountEmail}`];
+  }
+  try {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+    await (acmeClient as any).api.createAccount(accountPayload);
+  } catch (err) {
+    logger.error(err, `ACME pre-register account failed [caId=${caId}] — falling back to auto() registration`);
+  }
```
sheensantoscapadngan: what is this for? why did we add this as part of code signing revamp?

backend/src/services/certificate-authority/digicert/digicert-certificate-authority-fns.ts
```diff
@@ -659,6 +660,57 @@ export const DigiCertCertificateAuthorityFns = ({
...
+  const pollOrderForCertificate = async ({
+    caId,
+    orderId
+  }: {
+    caId: string;
+    orderId: number;
+  }): Promise<
+    | { status: "issued"; certificateId: number }
+    | { status: "pending"; orderStatus: string }
+    | { status: "failed"; orderStatus: string; reason: string }
```
sheensantoscapadngan: enums

backend/src/services/certificate-authority/digicert/digicert-certificate-authority-fns.ts
```diff
@@ -659,6 +660,57 @@ export const DigiCertCertificateAuthorityFns = ({
...
+
+    if (orderStatus === DigiCertOrderStatus.Pending) {
+      const checkResult = await client.checkValidation(orderId);
+      orderStatus = checkResult.order_status?.toLowerCase() ?? orderStatus;
+      certificateId = checkResult.certificate_id ?? certificateId;
+    }
+
+    const isFinalisable = DIGICERT_FINAL_ISSUED_STATUSES.some((status) => status === orderStatus);
+    if (isFinalisable && certificateId) {
+      return { status: "issued", certificateId };
```
sheensantoscapadngan: enum as well

backend/src/services/certificate-authority/certificate-issuance-queue.ts
```diff
@@ -410,14 +411,42 @@ export const certificateIssuanceQueueFactory = ({
...
+            );
+          }
+
+          const csrObj = await x509.Pkcs10CertificateRequestGenerator.create({
+            name: commonName ? [{ CN: [commonName] }] : [],
+            keys: leafKeys,
+            signingAlgorithm: keyAlg,
+            extensions: csrExtensions
+          });
+          certificateCsr = csrObj.toString("pem");
```
sheensantoscapadngan: hold on.. why did we have to update this? I don't think you're able to issue a code signing certificate using ACME anywhere... does it make sense to be able to do that with ACME?

backend/src/services/resource-cleanup/resource-cleanup-queue.ts
```diff
@@ -120,6 +120,11 @@ export const dailyResourceCleanUpQueueServiceFactory = ({
       handler: async () => {
         logger.info(`cron[${CronJobName.FrequentResourceCleanup}]: task started`);
         await identityAccessTokenDAL.removeExpiredTokens();
+        const expiredApprovalRequestIds = await approvalRequestDAL.markExpiredRequests();
+        if (expiredApprovalRequestIds.length > 0) {
+          await certificateRequestDAL.markExpiredApprovalRequests(expiredApprovalRequestIds);
+        }
+        await approvalRequestGrantsDAL.markExpiredGrants();
```
sheensantoscapadngan: wouldn't this be problematic memory-wise in the future where a LOT of requests could be expired per day?

backend/src/services/signer-membership/signer-membership-service.ts
```diff
@@ -0,0 +1,926 @@
...
+
+export const unknownSignerRoleMessage = (role: string) =>
+  `Unknown signer role '${role}'. Expected one of: ${BUILTIN_SIGNER_ROLES.join(", ")}.`;
+
+const ROLE_RANK: Record<string, number> = {
+  [ApplicationMembershipRole.Admin]: 3,
+  [ApplicationMembershipRole.Operator]: 2,
+  [ApplicationMembershipRole.Auditor]: 1
+};
+
```
sheensantoscapadngan: same remark here... why does it even matter? and wouldn't this break for when we have custom signer roles?

backend/src/services/signer-membership/signer-membership-service.ts
```diff
@@ -0,0 +1,926 @@
...
+    const { permission } = await $loadResourcePermission(signerId, projectId, {
+      actor,
+      actorId,
+      actorAuthMethod,
+      actorOrgId
+    });
+
+    ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionMemberActions.Create, ResourcePermissionSub.Member);
+
```
sheensantoscapadngan, line 204: same remark here about confusion with whether we're checking project permission or role permission... 

since this is signer-membership, I don't see why we're even checking project permission?

backend/src/services/signer-membership/signer-membership-service.ts
```diff
@@ -0,0 +1,926 @@
...
+        actorId,
+        projectId,
+        actorAuthMethod,
+        actorOrgId,
+        actionProjectType: ActionProjectType.CertificateManager
+      });
+      canCreateProjectMember = projectPermission.can(
+        ProjectPermissionMemberActions.Create,
+        ProjectPermissionSub.Member
+      );
```
sheensantoscapadngan: why do we do this? Checked it locally, and it seems like the entire permissioning logic is complicated because we're auto-adding people to the product directly if they're not there yet. I thought we settled on removing that altogether and require people to be already IN the project before adding them to either signers or applications

backend/src/services/signer-membership/signer-membership-service.ts
```diff
@@ -0,0 +1,926 @@
...
+  }: TListSignerMembersDTO): Promise<TSignerMember[]> => {
+    await $loadSignerOrThrow(signerId, projectId);
+
+    const { permission } = await $loadResourcePermission(signerId, projectId, {
+      actor,
+      actorId,
+      actorAuthMethod,
+      actorOrgId
+    });
+    ForbiddenError.from(permission).throwUnlessCan(ProjectPermissionMemberActions.Read, ResourcePermissionSub.Member);
```
sheensantoscapadngan: same remark/concern here... please check all the other relevant places

backend/src/services/signer-membership/signer-membership-service.ts
```diff
@@ -0,0 +1,926 @@
...
+      if (directActorId && role) {
+        idToEntry.set(directActorId, { role, isDirect: true, viaGroupIds: new Set() });
+      }
+    }
+
+    const groupMemberships = allMemberships.filter((m) => Boolean(m.actorGroupId));
+    const groupIds = groupMemberships.map((m) => m.actorGroupId as string);
+    if (groupIds.length > 0) {
+      const expansions =
+        kind === "user"
```
sheensantoscapadngan: enums please

backend/src/services/signer-membership/signer-membership-service.ts
```diff
@@ -0,0 +1,926 @@
...
+      }
+
+      for (const exp of expansions) {
+        const memberId = "userId" in exp ? exp.userId : exp.identityId;
+        const { groupId } = exp;
+        const groupRole = roleByGroupId.get(groupId);
+        if (groupRole) {
+          const existing = idToEntry.get(memberId);
+          if (existing) {
+            existing.role = pickHigherRole(existing.role, groupRole);
```
sheensantoscapadngan: same remark here... why do we need this?

backend/src/services/signer/signer-issuance-fns.ts
```diff
@@ -0,0 +1,177 @@
...
+    });
+  }
+  if (ca.status !== CaStatus.ACTIVE) {
+    throw new BadRequestError({ message: "The selected certificate authority is not active." });
+  }
+  if (!ca.internalCa?.id) {
+    throw new BadRequestError({
+      message:
+        "Issuing a signer certificate directly from an external Certificate Authority is not yet supported. " +
+        "Issue a code-signing certificate through Certificate Manager first and attach it to the signer by ID."
```
sheensantoscapadngan: what do you mean by this? is this relevant?

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -0,0 +1,572 @@
...
+    await signerDAL.updateById(job.signerId, {
+      certificateId,
+      status: SignerStatus.Active,
+      failureReason: null
+    });
+    await signerIssuanceJobDAL.updateById(job.id, {
+      status: SignerIssuanceJobStatus.Completed,
+      certificateId,
+      failureReason: null
+    });
```
sheensantoscapadngan, line 313: shouldn't we be using a transaction here?

backend/src/db/migrations/20260528124443_signer-revamp.ts
carlosmonastyrski: Not really, part of a test I forgot to remove 👍

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -0,0 +1,572 @@
...
+          await stepSyncWithCsr(job, "acme");
+          break;
+        case CaType.AZURE_AD_CS:
+          await stepSyncWithCsr(job, "azure");
+          break;
+        case CaType.AWS_PCA:
+          await stepSyncWithCsr(job, "aws-pca");
+          break;
+        case CaType.VENAFI_TPP:
+          await stepSyncWithCsr(job, "venafi");
```
sheensantoscapadngan: enums please

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -0,0 +1,572 @@
...
+      `signer issuance: signer attached [jobId=${job.id}] [signerId=${job.signerId}] [certificateId=${certificateId}]`
+    );
+  };
+
+  const stepDigiCert = async (job: TPkiSignerIssuanceJobs) => {
+    const ca = await certificateAuthorityDAL.findByIdWithAssociatedCa(job.caId);
+    if (!ca) {
+      await markJobFailed(job, "Certificate authority is no longer available.");
+      return;
+    }
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

if I don't define any limit, it get saved as "1 signature per approval" and "24 hrs signing window" for some reason 
Image

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -0,0 +1,572 @@
...
+    if (rows.length === 0) return;
+    await signerDAL.updateById(job.signerId, {
+      status: SignerStatus.Failed,
+      failureReason: trimmed
+    });
+    logger.info(
+      `signer issuance: job marked Failed [jobId=${job.id}] [signerId=${job.signerId}] [reason=${trimmed.slice(0, 200)}]`
+    );
+  };
+
```
sheensantoscapadngan, line 280: IGNORE PLACEMENT OF COMMENT

Removing a user at the product-level doesn't seem like it cleans up the resource-level permission

Image

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -0,0 +1,572 @@
...
+        case CaType.AWS_ACM_PUBLIC_CA:
+          await markJobFailed(
+            job,
+            "AWS ACM Public CA does not issue code-signing certificates. Choose a different CA."
+          );
+          return;
+        default:
+          await markJobFailed(job, `CA type '${job.caType}' is not supported for signer issuance.`);
+      }
+    } catch (err) {
```
sheensantoscapadngan, line 218: IGNORE PLACEMENT OF COMMENT

I only get to see the error about number of approvers in the second step which is a bit misleading

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -0,0 +1,572 @@
...
+        case CaType.ACME:
+          await stepSyncWithCsr(job, "acme");
+          break;
+        case CaType.AZURE_AD_CS:
+          await stepSyncWithCsr(job, "azure");
+          break;
+        case CaType.AWS_PCA:
+          await stepSyncWithCsr(job, "aws-pca");
+          break;
+        case CaType.VENAFI_TPP:
```
sheensantoscapadngan: were you able to test it out with these External CAs? if it's not something we're able to confirm, we should just strip it out completely. I recall that we were only able to do this with internally issued certificates previously

backend/src/services/signer/signer-types.ts
```diff
@@ -58,3 +75,79 @@ export type TListSigningOperationsDTO = {
...
+
+export type TSignerRequestStatusFilter = "pending" | "approved" | "expired" | "revoked";
+
+export type TListSignerRequestsDTO = {
+  signerId: string;
+  statuses?: TSignerRequestStatusFilter[];
+  offset?: number;
+  limit?: number;
+} & TActorPermission;
+
```
sheensantoscapadngan, line 129: IGNORE PLACEMENT OF COMMENT

is this a new modal component for danger? I haven't seen this anywhere else 
Image

backend/src/services/signer/signer-types.ts
```diff
@@ -58,3 +75,79 @@ export type TListSigningOperationsDTO = {
...
+export type TPreApproveSigningDTO = {
+  signerId: string;
+  granteeUserId?: string;
+  granteeIdentityId?: string;
+  justification: string;
+  requestedSignings?: number;
+  requestedWindowStart?: string;
+  requestedWindowEnd?: string;
+} & TActorPermission;
+
```
sheensantoscapadngan, line 147: IGNORE PLACEMENT OF COMMENT

I wonder if we should add any "reason" indicator? Not sure if it's obvious what that's referring to 

Image

Image

backend/src/services/signer/signer-issuance-fns.ts
```diff
@@ -0,0 +1,177 @@
...
+  });
+
+  const kmsEncryptor = await kmsService.encryptWithKmsKey({ kmsId: certificateManagerKmsId });
+
+  const skLeafObj = crypto.nativeCrypto.KeyObject.from(leafKeys.privateKey);
+  const skLeafPem = skLeafObj.export({ format: "pem", type: "pkcs8" }) as string;
+  const { cipherTextBlob: encryptedPrivateKey } = await kmsEncryptor({
+    plainText: Buffer.from(skLeafPem)
+  });
+
```
sheensantoscapadngan, line 165: IGNORE PLACEMENT OF COMMENT

I think it would be more appropriate to say "approval is not configured" rather than "off" since there's no explicit toggle here whatsoever
Image

backend/src/services/signer/signer-issuance-errors.ts
```diff
@@ -0,0 +1,154 @@
...
+};
+
+type TAxiosErrorBody = {
+  type?: string;
+  detail?: string;
+  errors?: { code?: string | number; message?: string; detail?: string }[];
+  error?: { message?: string; detail?: string } | string;
+  message?: string;
+};
+
```
sheensantoscapadngan, line 45: IGNORE PLACEMENT OF COMMENT

why is the "signature count" opening up as "3" by default even when my policy says max is 1
Image

backend/src/services/signer/signer-enums.ts
```diff
@@ -1,10 +1,19 @@
...
+  Disabled = "disabled",
+  Expired = "expired"
 }
 
 export enum SigningOperationStatus {
   Success = "success",
   Failed = "failed",
   Denied = "denied"
 }
+
```
sheensantoscapadngan, line 14: IGNORE PLACEMENT OF COMMENT

seems like we're missing a confirmation modal for membership deletion? 
Image

backend/src/services/signer/signer-types.ts
carlosmonastyrski: Yes, it's being used in other places like OverviewPage or RequestSecretTab as well

backend/src/server/routes/v1/signer-router.ts
carlosmonastyrski: Resource-scoped abilities aren't shipped to the client like the project ability, there's no client-side resource-ability provider. Computing "can this actor edit/sign/manage *this signer*" needs the user's project role, their signer membership. The client only sees the first, this endpoint rolls the rest up into a flat `can*` shape per signer. PKI Applications use the exact same pattern. Moving to a client-side resource-ability provider is its own refactor across both products.

backend/src/services/signer/signer-issuance-service.ts
carlosmonastyrski: Tested it with ACME, Digicert and AWS

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1096,4 +1073,71 @@ const buildApplicationProjectAdminFallbackRules = () => {
...
+  );
+
+  return rules;
+};
+
+const buildProjectAdminSignerFallbackRules = () => {
+  const { can, rules } = new AbilityBuilder<MongoAbility<ResourcePermissionSet>>(createMongoAbility);
+
+  can(
+    [ResourcePermissionSignerActions.Read, ResourcePermissionSignerActions.ManageMembers],
```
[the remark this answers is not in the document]
carlosmonastyrski: Intended

backend/src/services/signer/signer-service.ts
```diff
@@ -111,57 +212,354 @@ export const signerServiceFactory = ({
...
+      resolvedCaType = (ca.externalCa?.type as CaType | undefined) ?? CaType.INTERNAL;
     }
 
+    const isExternalCa = resolvedCaType !== null && resolvedCaType !== CaType.INTERNAL;
+
+    const keyAlgorithm: CertKeyAlgorithm = (dto.keyAlgorithm as CertKeyAlgorithm) ?? CertKeyAlgorithm.RSA_2048;
+
+    let issuedCertificateId: string | null = null;
+    if (usingCa && resolvedCaId && !isExternalCa) {
+      const { certificateId } = await issueSignerCertificate(
```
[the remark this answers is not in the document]
carlosmonastyrski: Only product admins can create code signers

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1096,4 +1073,71 @@ const buildApplicationProjectAdminFallbackRules = () => {
...
+    ResourcePermissionSub.Signer
+  );
+
+  return rules;
+};
```
sheensantoscapadngan, line 1135: IGNORE PLACEMENT OF COMMENT

project mentioned. please double check everywhere else 
Image

Image

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -6,10 +6,11 @@ import handlebars from "handlebars";
 import {
   AccessScope,
   ActionProjectType,
-  ApplicationMembershipRole,
   OrganizationActionScope,
   OrgMembershipRole,
   ProjectMembershipRole,
+  ResourceMembershipRole,
```
sheensantoscapadngan, line 12: IGNORE PLACEMENT OF COMMENT

need to update this as well 
Image

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -18,12 +19,16 @@ import {
...
   applicationOperatorPermissions,
-  applicationProjectAdminFallbackPermissions,
   cryptographicOperatorPermissions,
+  projectAdminApplicationFallbackPermissions,
   projectAdminPermissions,
+  projectAdminSignerFallbackPermissions,
   projectMemberPermissions,
   projectNoAccessPermissions,
   projectViewerPermission,
+  signerAdminPermissions,
```
sheensantoscapadngan, line 29: IGNORE PLACEMENT OF COMMENT

it would be very helpful if we could add context for the "Key Algorithm" field... This seems like the most technical property in this view but it's the one without a description 
Image

backend/src/ee/services/permission/permission-service.ts
sheensantoscapadngan: same here right

Image

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -126,33 +131,53 @@ const buildProjectPermissionRules = (projectUserRoles: TBuildProjectPermissionDT
...
+        return signerOperatorPermissions;
+      case ResourceMembershipRole.Auditor:
+        return signerAuditorPermissions;
+      case ResourceMembershipRole.Custom:
+        throw new BadRequestError({ message: "Custom resource-level roles are not supported yet" });
+      default:
+        throw new NotFoundError({ name: "SignerRoleInvalid", message: `Signer role '${role}' not found` });
+    }
+  }
+
```
sheensantoscapadngan, line 149: IGNORE PLACEMENT OF COMMENT

hold on.. this part feels a bit excessive
Image

I think we can even just remove the "Signer Detail Page" section altogether?

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -126,33 +131,53 @@ const buildProjectPermissionRules = (projectUserRoles: TBuildProjectPermissionDT
...
+  switch (role) {
+    case ResourceMembershipRole.Admin:
+      return applicationAdminPermissions;
+    case ResourceMembershipRole.Operator:
```
sheensantoscapadngan, line 153: IGNORE PLACEMENT OF COMMENT

feels like we can also just remove this altogether? 
Image

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -126,33 +131,53 @@ const buildProjectPermissionRules = (projectUserRoles: TBuildProjectPermissionDT
...
+    case ResourceMembershipRole.Auditor:
+      return applicationAuditorPermissions;
+    case ResourceMembershipRole.Custom:
+      throw new BadRequestError({ message: "Custom resource-level roles are not supported yet" });
+    default:
+      throw new NotFoundError({
+        name: "ApplicationRoleInvalid",
+        message: `Application role '${role}' not found`
+      });
+  }
```
sheensantoscapadngan, line 164: IGNORE PLACEMENT OF COMMENT

why do we mention Signing via the API as a dedicated section here? If it's not the most common usage pattern then it might be too distracting? Somehow I feel like we can instead just say that to use a signer, you either use the PKCS11 module or via API, and then just link the relevant docs

Image

backend/src/ee/services/permission/resource-permission.ts
```diff
@@ -21,7 +20,22 @@ export enum ResourcePermissionSub {
   ApprovalRequests = "approval-requests",
   ApprovalRequestGrants = "approval-request-grants",
   Member = "member",
-  Role = "role"
+  Signer = "pki-signer"
+}
+
```
sheensantoscapadngan, line 25: IGNORE PLACEMENT OF COMMENT

what do you mean by administrator doesn't want to wait? 
Image

Can we instead just say something like "Administrators can also pre-approve signing," and then maybe describe when a very specific scenario for when this would be desired

backend/src/ee/services/permission/resource-permission.ts
```diff
@@ -21,7 +20,22 @@ export enum ResourcePermissionSub {
...
+export enum ResourcePermissionSignerActions {
+  Read = "read",
+  Edit = "edit",
+  Delete = "delete",
+  ManageStatus = "manage-status",
+  ManageMembers = "manage-members",
+  ManagePolicy = "manage-policy",
```
sheensantoscapadngan, line 32: IGNORE PLACEMENT OF COMMENT

Image

What do you mean by "who you are here"? It feels a bit confusing because the tabs say "pre-approved signing" and "request to sign", and it doesn't really help me with identifying myself...

backend/src/ee/services/permission/resource-permission.ts
```diff
@@ -21,7 +20,22 @@ export enum ResourcePermissionSub {
...
+  Sign = "sign",
+  RequestSign = "request-sign",
+  PreApprove = "pre-approve",
+  RevokeRequest = "revoke-request",
+  ReissueCertificate = "reissue-certificate",
```
sheensantoscapadngan, line 37: IGNORE PLACEMENT OF COMMENT
Hold on, they're supposed to be already following the steps here because they're the approver or the rejecter, so I'm not sure if this is still even necessary to say?

Image

backend/src/ee/services/permission/resource-permission.ts
```diff
@@ -78,8 +92,7 @@ export type ResourcePermissionSet =
   | [ProjectPermissionApprovalRequestActions, ResourcePermissionSub.ApprovalRequests]
   | [ProjectPermissionApprovalRequestGrantActions, ResourcePermissionSub.ApprovalRequestGrants]
   | [ProjectPermissionMemberActions, ResourcePermissionSub.Member]
-  | [ProjectPermissionActions, ResourcePermissionSub.Role]
-  | [ProjectPermissionCertificateProfileActions, ProjectPermissionSub.CertificateProfiles]
-  | [ProjectPermissionAuditLogsActions, ProjectPermissionSub.AuditLogs];
+  | [ResourcePermissionSignerActions, ResourcePermissionSub.Signer]
```
sheensantoscapadngan, line 95: IGNORE PLACEMENT OF COMMENT

why is "last" always being marked as bold here? 
Image

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -126,33 +131,53 @@ const buildProjectPermissionRules = (projectUserRoles: TBuildProjectPermissionDT
...
-      }
-    })
+    .map(({ role }) => resolveResourceRoleRules(resourceType, role))
     .reduce((prev, curr) => prev.concat(curr), [] as RawRuleOf<MongoAbility<ResourcePermissionSet>>[])
     .sort((a, b) => Number(Boolean(a.inverted)) - Number(Boolean(b.inverted)));
 
   return rules;
 };
 
+const resolveResourceProjectAdminFallback = (resourceType: ResourceType) => {
```
sheensantoscapadngan, line 176: IGNORE PLACEMENT OF COMMENT

Why is it either approved or active? Isn't it supposed to be just one status? 
Image

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -696,15 +721,15 @@ export const permissionServiceFactory = ({
         actorType: actor
       });
 
+      const fallbackRules = resolveResourceProjectAdminFallback(resourceType);
```
sheensantoscapadngan, line 724: IGNORE PLACEMENT OF COMMENT

I'm noticing that there are a few sections we can just omit altogether. We don't necessarily have to include details like these in the documentation if it's not notable or if it's something that's easily discoverable in the UI. 
Image

backend/src/ee/services/permission/resource-permission.ts
```diff
@@ -4,7 +4,6 @@ import {
   ProjectPermissionActions,
   ProjectPermissionApprovalRequestActions,
   ProjectPermissionApprovalRequestGrantActions,
-  ProjectPermissionAuditLogsActions,
```
sheensantoscapadngan, line 7: IGNORE PLACEMENT OF COMMENT

I think it might make more sense to move the access lifecycle under "configure the approval policy" so that everything that's related to managing a request are grouped together 

Image

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1096,4 +1073,71 @@ const buildApplicationProjectAdminFallbackRules = () => {
...
+      ResourcePermissionSignerActions.RevokeRequest,
+      ResourcePermissionSignerActions.ReissueCertificate,
+      ResourcePermissionSignerActions.ExportCertificate
+    ],
+    ResourcePermissionSub.Signer
+  );
+
+  return rules;
+};
+
```
sheensantoscapadngan, line 1098: IGNORE PLACEMENT OF COMMENT

It looks like we have a duplicate for the list of guides in the PKCS#11 module page. The ones under "What's next" actually look a lot better than the ones under "Tool Guides", so perhaps we can just remove the "Tool Guides"? 

Image

Image

backend/src/services/project-membership/project-membership-service.ts
```diff
@@ -370,6 +377,17 @@ export const projectMembershipServiceFactory = ({
         tx
       );
 
+      await membershipUserDAL.delete(
```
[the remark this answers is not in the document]
sheensantoscapadngan: relevant?

backend/src/@types/knex.d.ts
```diff
@@ -407,6 +407,9 @@ import {
   TPkiScepTransactions,
   TPkiScepTransactionsInsert,
   TPkiScepTransactionsUpdate,
+  TPkiSignerCertificateIssuanceJobs,
```
sheensantoscapadngan, line 410: IGNORE PLACEMENT OF COMMENT

link to the certificate authority page if none exists
Image

backend/src/db/migrations/20260528124443_signer-revamp.ts
```diff
@@ -0,0 +1,132 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+const SIGNER_CERTIFICATE_ISSUANCE_JOBS_TABLE = "pki_signer_certificate_issuance_jobs";
+
```
sheensantoscapadngan, line 6: IGNORE PLACEMENT OF COMMENT

As I was following the documentation for OpenSSL, I realized that we don't even mention setting up authentication as one of the first steps (specifically how to define it). Right now, we go straight to the code signing invocation, which was confusing  because, as I was trying to literally follow everything from top to bottom, I encountered an auth failure.

Let's update this and the other guides please
Image

backend/src/db/schemas/pki-signer-certificate-issuance-jobs.ts
```diff
@@ -0,0 +1,41 @@
...
+  attempts: z.number().default(0),
+  maxAttempts: z.number().default(100),
+  nextPollAt: z.date(),
+  lastAttemptAt: z.date().nullable().optional(),
+  failureReason: z.string().nullable().optional(),
+  certificateId: z.string().uuid().nullable().optional(),
+  createdAt: z.date(),
+  updatedAt: z.date()
+});
+
```
sheensantoscapadngan, line 33: Perhaps it doesn't make sense to show the machine identity in the list then if it's wrong for me to select it directly? 

Also, there is a mention of a project in the error message here. 
Image

backend/src/db/schemas/pki-signers.ts
```diff
@@ -13,11 +13,17 @@ export const PkiSignersSchema = z.object({
   name: z.string(),
   description: z.string().nullable().optional(),
   status: z.string().default("active"),
-  certificateId: z.string().uuid(),
+  certificateId: z.string().uuid().nullable().optional(),
```
sheensantoscapadngan, line 16: IGNORE PLACEMENT OF COMMENT

It would be more ideal if we just show the error inside the input fields, kind of like how we do it in the other places, rather than showing them in this modal over here. 

Image

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -51,6 +51,7 @@ import {
   ResourcePermissionCertificateActions,
   ResourcePermissionPkiSyncActions,
   ResourcePermissionSet,
+  ResourcePermissionSignerActions,
```
sheensantoscapadngan, line 54: IGNORE PLACEMENT OF COMMENT

For some reason, I can't click into a request row. I was expecting it to bring me to the signing requests page, but it didn't...
Image

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -1055,15 +1034,12 @@ const buildApplicationAuditorPermissionRules = () => {
...
-  can([ProjectPermissionMemberActions.Read], ResourcePermissionSub.Member);
-  can([ProjectPermissionActions.Read], ResourcePermissionSub.Role);
   can([ProjectPermissionCertificateProfileActions.Read], ProjectPermissionSub.CertificateProfiles);
-  can([ProjectPermissionAuditLogsActions.Read], ProjectPermissionSub.AuditLogs);
 
   return rules;
 };
 
-const buildApplicationProjectAdminFallbackRules = () => {
+const buildProjectAdminApplicationFallbackRules = () => {
```
sheensantoscapadngan, line 1042: IGNORE PLACEMENT OF COMMENT

Why is the default start date of the pre-approved signing not the same as my current system time? Looks like it's in UTC, but there is no clear indication whatsoever...
Image

backend/src/services/project-membership/project-membership-service.ts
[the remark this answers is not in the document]
carlosmonastyrski: No 😅

sheensantoscapadngan (approved):

PR #6684, opened 2026-05-30
chore: mark root make targets as phony

codes-stories:
## Context

Adds a `.PHONY` declaration for the root `Makefile` command targets.
These targets are command aliases and do not produce files with matching names. Without `.PHONY`, `make` may skip a target if a file or directory with the same name exists, for example `build` or `push`.

This change makes the Makefile behavior more reliable and consistent with the other Makefiles in the repository.

## Screenshots
Not applicable. This is a Makefile-only change.

## Steps to verify the change
1. Run `make -n build`
2. Run `make -n reviewable`
3. Confirm the expected commands are printed and targets are treated as command targets.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: build push up-dev up-dev-ldap up-dev-metrics up-prod down reviewable-ui reviewable-api reviewable up-dev-sso go-generate validate-upgrade-impact generate-upgrade-impact generate-upgrade-impact-dry-run
+
 build:
 	docker-compose -f docker-compose.yml build
 


codes-stories:
Added `.PHONY` for the root Makefile targets because they are commands, not files.

This prevents `make` from skipping a target if a file or folder with the same name exists, and keeps it consistent with the other Makefiles in the repo.

PR #6689, opened 2026-06-01
chore: skip email verification whenever enforce sso is on

Thiago-AS:
## Context

  What

  When an organization enforces SAML/OIDC SSO (authEnforced), SSO logins now provision and sign in users directly without the extra email-verification step, and email/password signup is blocked for that org's verified domain(s).

  Why

  When SSO is enforced, the org's DNS-verified domain plus the IdP are already authoritative proof of email ownership, so the separate verification code is redundant friction. Blocking password signup for those domains also closes an account-takeover vector, a competing password account can no longer be created for a domain the org owns and federates.

  Changes

  - SAML / OIDC / LDAP login (*-config-service.ts): when org.authEnforced, create new users as isAccepted/isEmailVerified and skip sending the verification
  email. A shared ensureSsoAccountVerified helper also upgrades pre-existing accounts (provisioned before enforcement) on next SSO login.
  - Signup (auth-signup-service.ts): beginEmailSignupProcess rejects email/password signup when the email's domain is verified by an SSO-enforced org.
  - Frontend (OrgGeneralAuthSection.tsx): clarified the enforce-SSO copy and added a break-glass warning in the confirmation modal (non-Google).
  - Test: e2e spec covering both the SSO-session-without-verification path and the blocked password signup.

  Notes

  - Only applies to SAML/OIDC enforcement — Google SSO enforcement (googleSsoAuthEnforced, mutually exclusive) is untouched.
  - Password login for enforced orgs was already blocked; this extends the same guarantee to signup.
  - ⚠️ Admins should ensure a break-glass account exists before enabling, since password signup for the domain is blocked afterward (surfaced in the modal
  warning).

## Screenshots

Screenshot 2026-05-31 at 20 12 56
Screenshot 2026-05-31 at 22 52 47
Screenshot 2026-05-31 at 22 56 09
Screenshot 2026-05-31 at 22 58 11

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/oidc/oidc-config-service.ts
+++ b/backend/src/ee/services/oidc/oidc-config-service.ts
@@ -46,7 +46,7 @@ import { TTelemetryServiceFactory } from "@app/services/telemetry/telemetry-serv
 import { PostHogEventTypes } from "@app/services/telemetry/telemetry-types";
 import { TUserDALFactory } from "@app/services/user/user-dal";
 import { TUserAliasDALFactory } from "@app/services/user-alias/user-alias-dal";
-import { ensureSsoAccountVerified } from "@app/services/user-alias/user-alias-fns";
+import { ensureSsoAccountVerified, isStaleSsoAlias } from "@app/services/user-alias/user-alias-fns";
 import { UserAliasType } from "@app/services/user-alias/user-alias-types";
 
 import { TEmailDomainDALFactory } from "../email-domain/email-domain-dal";
@@ -223,6 +223,11 @@ export const oidcConfigServiceFactory = ({
     const skipEmailVerification = Boolean(organization.authEnforced);
 
     let user: TUsers;
+    // A stale, still-unverified alias may point at another user's account. Don't mutate that
+    // account's org membership / group state until the IdP proves control of it (the
+    // email-verification fallback below issues no session). Resolved against the existing alias
+    // before any mutation; freshly created aliases are never stale.
+    let isStaleAlias = false;
     if (userAlias) {
       user = await userDAL.transaction(async (tx) => {
         const foundUser = await userDAL.findById(userAlias.userId, tx);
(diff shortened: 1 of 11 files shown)


backend/src/ee/services/oidc/oidc-config-service.ts
```diff
@@ -432,6 +451,17 @@ export const oidcConfigServiceFactory = ({
 
     await oidcConfigDAL.update({ orgId }, { lastUsed: new Date() });
 
+    // When SSO is enforced, mark the user + alias as verified/accepted before issuing a session.
+    if (skipEmailVerification) {
```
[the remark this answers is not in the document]
Thiago-AS: Even if email_verified is absent, we trust the organization's IdP to manage identities correctly in this case.

backend/src/services/user-alias/user-alias-fns.ts
```diff
@@ -0,0 +1,86 @@
...
+ * metadata, or group memberships onto that account (nor issue a session for it) until ownership is
+ * proven via the separate email-verification flow. A verified alias is already trusted, so it is
+ * never considered stale.
+ */
+export const isStaleSsoAlias = ({ user, userAlias, assertedEmail }: TIsStaleSsoAliasDTO): boolean => {
+  if (userAlias.isEmailVerified) return false;
+
+  const normalizedAssertedEmail = assertedEmail?.toLowerCase().trim();
+  const accountEmails = new Set(
+    [user.username, user.email, ...(userAlias.emails ?? [])]
```
akhilmhdh, line 34: Username and email are same. We made sure that in the recent auth revamp.I am still confused about this? What are we solving this line?

akhilmhdh (commented):
Question: So does this mean an existing user using the forgot account option - can configure the account with email + password flow? This is used when there IDP access is lost

backend/src/services/user-alias/user-alias-fns.ts
Thiago-AS: Checking wether the email attempting to SSO is the same on the alias/user previous created.

Thiago-AS:
> Question: So does this mean an existing user using the forgot account option - can configure the account with email + password flow? This is used when there IDP access is lost

From what I understood the break the glass comes from admins being able to login using username + password right? I can double check regarding forgot account. But it should not affect that flow, since the account is already created.

akhilmhdh (approved):

PR #6693, opened 2026-06-01
fix(auth): reject non-JWT 4+ segment tokens in peekJWTClaims

### LINKED ISSUE #6651: peekJWTClaims misclassifies 4-part tokens as valid JWTs
### Product

### Describe the bug

peekJWTClaims uses SplitN(token, ".", 4) but checks len(parts) < 3. A 4-part token like "a.b.c.d" has 4 parts, passes the check, and gets classified as a valid JWT. The function should use SplitN(..., 3) and check != 3.

### To Reproduce

1. Pass a token with 4 dot-separated segments to peekJWTClaims
2. Observe it is classified as a valid JWT instead of being rejected

### Expected behavior

Only tokens with exactly 3 dot-separated parts should be classified as valid JWTs.

### Screenshots

_No response_

### Deployment Type

Self-hosted

### Additional context

classify.go:47-49. Found via Faultmark (faultmark.com), an AI code scanner that uses a multi-model debate to verify bugs against actual source code before reporting them.

jaydeep-pipaliya:
## Context

Closes #6651.

\`peekJWTClaims\` in \`backend-go/internal/services/auth/apiauth/classify.go\` used:

\`\`\`go
parts := strings.SplitN(token, ".", 4)
if len(parts) < 3 {
    return peekClaims{}
}
\`\`\`

\`SplitN(s, sep, 4)\` returns **at most** 4 elements; anything past the third \`.\` is folded into the last element. So a four-segment token like \`a.b.c.d\` came back as \`["a", "b", "c.d"]\` — three parts — and passed the \`< 3\` gate. The function then base64-decoded \`parts[1]\` and \`json.Unmarshal\`ed it. If a non-JWT input happens to have a JWT-looking middle segment, the function returned a populated \`peekClaims\` and \`ClassifyToken\` treated the token as a JWT.

## Changes

\`backend-go/internal/services/auth/apiauth/classify.go\`:
- Drop \`SplitN\`'s collapsing-tail behaviour: switch to plain \`strings.Split\` so all segments stay separate.
- Require **exactly three** parts (\`len(parts) != 3\` returns empty), matching the RFC 7519 JWT shape.
- Add an inline comment explaining the pitfall so this doesn't regress under a future "performance" sweep.

## Type

- [x] Fix

## Steps to verify the change

- \`cd backend-go && go test ./internal/services/auth/apiauth/ -run TestClassifyToken\` passes both the existing cases and the new regression case.
- The new \`four-segment token is not classified as JWT\` subtest constructs a header + payload + signature + extra suffix and asserts \`ClassifyToken\` returns the empty \`AuthMode\` (the rejection sentinel).
- Verified the new subtest fails on \`main\` (\`expected empty (4-segment token rejected), got "jwt"\`) and passes after the change.

## Checklist

- [x] Tested locally — Go unit tests pass, new regression case fails on main
- [ ] Updated docs (no public API change)

### DIFF
--- a/backend-go/internal/services/auth/apiauth/classify.go
+++ b/backend-go/internal/services/auth/apiauth/classify.go
@@ -44,8 +44,12 @@ func ClassifyToken(token string) auth.AuthMode {
 // peekJWTClaims decodes the JWT payload segment (no signature verification)
 // to inspect the authTokenType claim.
 func peekJWTClaims(token string) peekClaims {
-	parts := strings.SplitN(token, ".", 4)
-	if len(parts) < 3 {
+	// A valid JWT has exactly three dot-separated segments (header.payload.signature).
+	// SplitN(..., 4) used to collapse 4+ segments into a single trailing element,
+	// which let tokens like "a.b.c.d" through with a usable parts[1]. Use unlimited
+	// Split + an exact-3 check so anything else is rejected up front.
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
 		return peekClaims{}
 	}
 
--- a/backend-go/internal/services/auth/apiauth/classify_test.go
+++ b/backend-go/internal/services/auth/apiauth/classify_test.go
@@ -69,6 +69,23 @@ func TestClassifyToken_VariousInputTypes(t *testing.T) {
 		}
 	})
 
+	// Regression: peekJWTClaims used SplitN(..., 4) + len(parts) < 3, which
+	// let 4+ segment tokens through because SplitN collapses the tail into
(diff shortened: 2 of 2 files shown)


jaydeep-pipaliya:
Closing — upstream's `ClassifyToken` refactor on `main` already covers this. The `peekJWTClaims` helper has been removed entirely, the shape check is now inlined in `ClassifyToken` using `SplitN(..., 4) + != 3` which correctly rejects 4-segment tokens, and `classify_test.go` even includes a `four parts` test case. Nothing left for this PR to add.

PR #6694, opened 2026-06-01
fix(auth): guard against missing org row in validateJWT

### LINKED ISSUE #6644: nil dereference panic in validateJWT when org row is missing
### Product

### Describe the bug

findOrgByID is called at line 213 of authenticator.go but there is no nil check before org.Name is accessed at line 230. If the org row does not exist in the database, this causes a nil pointer dereference panic.

### To Reproduce

1. Issue a JWT for an org
2. Delete the org from the database
3. Attempt to validate the JWT
4. Server panics

### Expected behavior

validateJWT should handle a missing org gracefully and return an authentication error instead of panicking.

### Screenshots

_No response_

### Deployment Type

Self-hosted

### Additional context

authenticator.go:213. Found via Faultmark (faultmark.com), an AI code scanner that uses a multi-model debate to verify bugs against actual source code before reporting them.

jaydeep-pipaliya:
## Context

Closes #6644.

\`findOrgByID\` in \`backend-go/internal/services/auth/apiauth/authenticator.go\` returns \`(nil, nil)\` on \`pgx.ErrNoRows\`:

\`\`\`go
err := row.Scan(...)
if errors.Is(err, pgx.ErrNoRows) {
    return nil, nil
}
\`\`\`

The function is called from three branches in \`validateJWT\` / \`validateIdentityAccessToken\`:
- **Line 177** (sub-org branch) — checks \`if subOrg == nil\` and returns a clean error.
- **Line 329** (identity-access-token branch) — checks \`if org == nil\` and returns a clean error.
- **Line 213** (regular-org branch) — **missing the nil guard**. Goes straight to \`org.Name\` on line 230, which panics the worker.

The panic is reachable from any session that holds a JWT whose organization row was deleted between issue and validation (e.g. org-deletion flow, test-tenant cleanup, manual SQL).

## Changes

Add the \`if org == nil\` check between lines 216 and 218 (after the error check, before the membership lookup), mirroring the existing pattern in the other two branches. The error is reported via \`errutil.BadRequest\` so the caller sees the same 400 shape it would for any other JWT-with-stale-claim failure.

## Type

- [x] Fix

## Steps to verify the change

1. Issue a JWT for org X.
2. \`DELETE FROM organizations WHERE id = '<X>'\` (or run the org-deletion flow).
3. Hit any authenticated endpoint with the JWT.
4. Before: worker panics on \`org.Name\` deref.
5. After: returns \`400 Bad Request — Organization <id> not found\`, matching the sub-org branch's behavior.

## Checklist

- [x] Tested locally — \`go build ./...\` is clean, the change matches the existing nil-check pattern in two other branches of the same function
- [ ] Updated docs (no public API change)

### DIFF
--- a/backend-go/internal/services/auth/apiauth/authenticator.go
+++ b/backend-go/internal/services/auth/apiauth/authenticator.go
@@ -218,7 +218,7 @@ func (a Authenticator) validateJWT(ctx context.Context, token string) (*auth.Ide
 				// findOrgByID returns (nil, nil) on pgx.ErrNoRows. The org row can be
 				// missing if it was deleted between JWT issue and validation; without
 				// this guard `org.Name` below panics.
-				return nil, errutil.BadRequest("Organization %s not found", claims.OrganizationID).WithErrf("validateJWT(orgId=%s): organization not found", claims.OrganizationID)
+				return nil, errutil.NotFound("Organization %s not found", claims.OrganizationID).WithErrf("validateJWT(orgId=%s): organization not found", claims.OrganizationID)
 			}
 
 			orgMembership, err := a.findEffectiveOrgMembership(ctx, auth.ActorTypeUser, user.ID, claims.OrganizationID, "accepted")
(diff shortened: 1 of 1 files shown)


akhilmhdh (approved):

akhilmhdh:
Thank you for the fix

PR #6705, opened 2026-06-02
feat: completed support for external kms, hsm and production hardening before shadowing

akhilmhdh:
## Context

This PR closes PLATFOR-389, PLATFOR-383, PLATFOR-382, SECRETS-357

## Screenshots

To verify
1. Set environment variable `GOLANG_SIDECAR_URL=http://backend-go:4040`
2. Run docker compose under root
```
docker compose -f docker-compose.dev.yml --profile go up
```
3. Configure external kms or hsm as needed and everything should work fine

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


.github/workflows/run-backend-go-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+    timeout-minutes: 30
+
+    steps:
+      - name: Free up disk space
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf "/usr/local/share/boost"
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+          docker system prune -af
```
mathnogueira, line 66: Why you need to do this?

backend-go/internal/ee/services/externalkms/externalkms.go
```diff
@@ -0,0 +1,82 @@
...
+// It implements the ExternalKmsProvider interface expected by the core KMS service.
+type Service struct {
+	logger *slog.Logger
+}
+
+// Deps holds dependencies for the external KMS service.
+type Deps struct{}
+
+// NewService creates a new external KMS service.
+func NewService(_ context.Context, logger *slog.Logger, _ *Deps) (*Service, error) {
```
mathnogueira, line 30: Why are you receiving the context here? Even if you are going to reuse later, keep in mind you should never store the context, it should be passed to the function (Encrypt and Decrypt for example) when it gets called, otherwise you might get inconsistent logs/traces with OTEL or other libraries that rely on context.Values.
Passing a context to a constructor is only fine if you need to do any IO operation inside the constructor. But that is kind of a weird thing to do anyways...

backend-go/internal/ee/services/externalkms/externalkms.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Encrypt encrypts plaintext using the specified external KMS provider.
+// The config parameter contains the decrypted provider configuration JSON.
+func (s *Service) Encrypt(ctx context.Context, provider string, config, plaintext []byte) ([]byte, error) {
+	p, err := s.createProvider(ctx, provider, config)
+	if err != nil {
+		return nil, err
+	}
+	defer p.Close() //nolint:errcheck // best-effort cleanup
```
mathnogueira: Instead of using disabling the linter here, why don't you use `errutil.DeferErr`?

backend-go/internal/ee/services/externalkms/externalkms.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Decrypt decrypts ciphertext using the specified external KMS provider.
+// The config parameter contains the decrypted provider configuration JSON.
+func (s *Service) Decrypt(ctx context.Context, provider string, config, ciphertext []byte) ([]byte, error) {
+	p, err := s.createProvider(ctx, provider, config)
+	if err != nil {
+		return nil, err
+	}
+	defer p.Close() //nolint:errcheck // best-effort cleanup
```
mathnogueira: Same here

backend-go/internal/ee/services/externalkms/externalkms.go
```diff
@@ -0,0 +1,82 @@
...
+// NewService creates a new external KMS service.
+func NewService(_ context.Context, logger *slog.Logger, _ *Deps) (*Service, error) {
+	return &Service{
+		logger: logger.With(slog.String("service", "externalkms")),
+	}, nil
+}
+
+// Encrypt encrypts plaintext using the specified external KMS provider.
+// The config parameter contains the decrypted provider configuration JSON.
+func (s *Service) Encrypt(ctx context.Context, provider string, config, plaintext []byte) ([]byte, error) {
```
mathnogueira: nit: one thing I like doing in those `string` parameters that we have specific values that can be provided is to type it like:

```go
type ProviderType string

const (
  ProviderAWS:ProviderType = "aws"
  ProviderGCP ProviderType = "gcp"
)

func (s *Service) Encrypt(ctx context.Context, provider ProviderType, config, plaintext []byte) ([]byte, error) { }
```

This way you make:

1. Make sure a random string is not passed unless a type conversion is done;
2. Easier to now which values are valid

backend-go/internal/ee/services/externalkms/provider_aws.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	"github.com/aws/aws-sdk-go-v2/service/kms"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+)
+
+// AWS credential types matching Node.js constants.
+const (
+	AwsCredentialTypeAccessKey  = "access-key"
+	AwsCredentialTypeAssumeRole = "assume-role"
+)
```
mathnogueira, line 22: nit: you can use a custom type here too

backend-go/internal/ee/services/externalkms/provider_gcp.go
```diff
@@ -0,0 +1,86 @@
...
+	keyName string
+}
+
+func newGcpProvider(ctx context.Context, cfg *GcpConfig) (*gcpProvider, error) {
+	credentialJSON, err := json.Marshal(cfg.Credential)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling GCP credential: %w", err)
+	}
+
+	client, err := kms.NewKeyManagementClient(ctx, option.WithAuthCredentialsJSON(option.ServiceAccount, credentialJSON))
```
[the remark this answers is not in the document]
mathnogueira: Should take a look

backend-go/internal/ee/services/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+}
+
+// keyByAuthToken extracts the bearer token from the Authorization header for rate limiting.
+// Falls back to IP if no token present.
+func keyByAuthToken(r *http.Request) (string, error) {
+	authHeader := r.Header.Get("Authorization")
+	if len(authHeader) > 7 && authHeader[:7] == "Bearer " {
+		return authHeader[7:], nil
+	}
```
mathnogueira, line 174: I thought rate limits were imposed in the Org-level, not at the user level. Won't this make it possible for users to bypass rate limits by using multiple Machine Identities? Or calling `login` again to generate a new bearer token?

Shouldn't we use the `orgId` instead here?

backend-go/internal/server/middlewares/cors.go
```diff
@@ -0,0 +1,97 @@
...
+			"Accept",
+			"Authorization",
+			"Content-Type",
+			"X-Request-Id",
+			"X-Requested-With",
+		}
+	}
+	if cfg.MaxAge == 0 {
+		cfg.MaxAge = 86400
+	}
```
mathnogueira, line 47: `cfg.MaxAge = 0` is a valid option, which means: "disable preflight caching"

Maybe an option pattern here would be better:

```go
CORS(
  WithAllowedMethods("GET", "POST", "PUT"),
  WithMaxAge(0)
)
```

backend-go/internal/server/middlewares/timeout.go
```diff
@@ -0,0 +1,79 @@
...
+						w.Header().Set("Content-Type", "application/json")
+						w.WriteHeader(http.StatusGatewayTimeout)
+
+						resp := errorResponse{
+							ReqID:      reqID,
+							StatusCode: http.StatusGatewayTimeout,
+							ErrorData:  "GatewayTimeoutError",
+							Message:    "Request timed out",
+						}
+						_ = json.NewEncoder(w).Encode(resp)
```
mathnogueira: Shouldn't you use `tw` instead of `w` here?

backend-go/internal/server/server.go
```diff
@@ -84,3 +93,42 @@ func requestLogger(next http.Handler, logger *slog.Logger) http.Handler {
 		)
 	})
 }
+
+// buildCORSConfig creates CORS config from server configuration.
+// Matches the Node.js @fastify/cors configuration.
+func (s *Server) buildCORSConfig() *middlewares.CORSConfig {
```
mathnogueira, line 98: Shouldn't it also generate a `MaxAge`?

backend-go/internal/services/kms/kms.go
```diff
@@ -311,39 +245,137 @@ func (s *Service) generateEncryptedKeyMaterial() ([]byte, error) {
...
 
-// decryptKmsKeyMaterial decrypts the key material from a kmsKeyWithAssociatedKms record.
-// Branches on internal (local AES-GCM) vs external (cloud provider — not yet implemented).
-func (s *Service) decryptKmsKeyMaterial(kmsKeyDoc *kmsKeyWithAssociatedKms) ([]byte, error) {
-	// TODO(go): fix this before mileston
-	if kmsKeyDoc.ExternalKmsID.Valid {
-		panic("external KMS not implemented")
+	if externalKmsID.Valid {
+		return nil, fmt.Errorf("KMS: cannot get key material for external KMS")
 	}
```
mathnogueira, line 279: What's the logic behind this? Why it fails if this is set? Should it only have the `internalKmsID`? If so, maybe that error message can be a bit more clear about this.

backend-go/internal/services/kms/kms.go
```diff
@@ -311,39 +245,137 @@ func (s *Service) generateEncryptedKeyMaterial() ([]byte, error) {
...
 
-	if !kmsKeyDoc.InternalKmsID.Valid {
-		return nil, fmt.Errorf("KMS: key %s has no internal or external KMS association", kmsKeyDoc.ID.String())
+	if !internalKmsID.Valid {
+		return nil, fmt.Errorf("KMS: key %s has no internal or external KMS association", id.String())
 	}
```
mathnogueira, line 283: Based on the error message, I would expect this condition to be:

```go
if !internalKmsID.Valid && !externalKmsID.Valid {
```

backend-go/tests/infra/infisical_secrets.go
```diff
@@ -0,0 +1,255 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&loginResp); err != nil {
+		return "", fmt.Errorf("decoding login response: %w", err)
+	}
+
+	return loginResp.AccessToken, nil
+}
+
+// listSecrets fetches secrets from the specified path.
+func (c *infisicalClient) listSecrets(ctx context.Context, secretPath string) (map[string]string, error) {
+	secretsURL := fmt.Sprintf("%s/api/v3/secrets/raw?workspaceId=%s&environment=%s&secretPath=%s",
```
mathnogueira, line 134: I wonder if we should use v4 instead of v3 here

backend-go/tests/infra/infisical_secrets.go
```diff
@@ -0,0 +1,255 @@
...
+func (s *InfisicalSecrets) Get(key string) string {
+	return s.secrets[key]
+}
+
+// Require returns a secret value by key, or skips the test if not found.
+func (s *InfisicalSecrets) Require(t *testing.T, key string) string {
+	t.Helper()
+	value, ok := s.secrets[key]
+	if !ok || value == "" {
+		t.Skipf("skipping: required secret %q not found in Infisical test project", key)
```
mathnogueira, line 246: I think we should probably fail instead of Skip, this will make things more explicit that something is wrong.

backend-go/tests/secretmanager/secrets/list_secrets_integration_test.go
```diff
@@ -893,3 +893,57 @@ func TestListSecretsV4_HTTP(t *testing.T) {
...
+			Environment:     proj.EnvSlug,
+			SecretPath:      new("/"),
+			ViewSecretValue: new(true),
+		})
+
+		require.NoError(t, err)
+		require.Len(t, result.Secrets, 1)
+		assert.Equal(t, "DEV_SECRET", result.Secrets[0].SecretKey)
+	})
+}
```
mathnogueira, line 962: ignore placement:

Not 100% sure, will let you decide. What do you think about checking if the row still exists in the DB (so we ensure it's a soft delete and not a hard delete).

mathnogueira (changes requested):

backend-go/internal/ee/services/externalkms/externalkms.go
akhilmhdh: The context is never stored. Each function accepts context as first argument. All services follows the convention of three argument: context, logger and other service dependencies. 

The context is taken service to setup somethings like a timer but keep watching for it on sig int

backend-go/internal/server/middlewares/cors.go
akhilmhdh: Ya! I haven't exposed it because nodejs doesn't expose it yet. 
```
   CORS_ALLOWED_ORIGINS: zpStr(
      z
        .string()
        .optional()
        .transform((val) => {
          if (!val) return undefined;
          return JSON.parse(val) as string[];
        })
    ),
    CORS_ALLOWED_HEADERS: zpStr(
      z
        .string()
        .optional()
        .transform((val) => {
          if (!val) return undefined;
          return JSON.parse(val) as string[];
        })
    ),
    ```

backend-go/internal/server/middlewares/timeout.go
akhilmhdh: Good catch

backend-go/internal/server/server.go
akhilmhdh: Not exposed in NodeJS

backend-go/internal/services/kms/kms.go
akhilmhdh: That logic is for decrypting an interal KMS key with a kms key id and used for both project and organization. |

If the provided kms key is valid external kms configuration then it should not be used with this function.

backend-go/internal/services/kms/kms.go
akhilmhdh: This is because previous check already isolated that there was no external kms connection and then it also doesn't have valid internal kms - means there is no association with anything

backend-go/tests/infra/infisical_secrets.go
akhilmhdh: This is intentional. This is dog food our internal service inside the test pipeline to connect with external system. This should not block the existing test you are running locally. CI CD will configure this in place.

backend-go/tests/secretmanager/secrets/list_secrets_integration_test.go
akhilmhdh: Good catch

backend-go/internal/ee/services/ratelimit/ratelimit.go
akhilmhdh: So if you expand the usage - you see it's configured for the mfa endpoint. This rate limiter is specifically used for MFA and the MFA follows the same convention in nodejs where it's by the token

.github/workflows/run-backend-go-tests.yml
akhilmhdh: Just preparing before the test container workflow starts

mathnogueira (dismissed):

mathnogueira (approved):

PR #6707, opened 2026-06-02
fix(migration): guard webauthn down() trigger drop with hasTable check

### LINKED ISSUE #6654: dropOnUpdateTrigger unconditional in down() causes rollback failure
### Product

### Describe the bug

In the add-webauthn-support migration, up() only creates the trigger if the table didn't exist, but down() calls dropOnUpdateTrigger unconditionally. Rolling back on a database where the table was never created will fail trying to drop a trigger that doesn't exist.

### To Reproduce

1. Run the migration on a database where the table already existed before the migration
2. Roll back the migration
3. Observe failure on dropOnUpdateTrigger

### Expected behavior

down() should mirror the conditional logic in up() and only drop the trigger if it was created.

### Screenshots

_No response_

### Deployment Type

Self-hosted

### Additional context

backend/src/db/migrations/20251118190904_add-webauthn-support.ts:28. Found via Faultmark (faultmark.com), an AI code scanner that uses a multi-model debate to verify bugs against actual source code before reporting them.

jaydeep-pipaliya:
## Context

Closes #6654.

The webauthn-support migration's \`up()\` creates the \`WebAuthnCredential\` table and trigger only when the table doesn't exist:

\`\`\`ts
if (!(await knex.schema.hasTable(TableName.WebAuthnCredential))) {
  await knex.schema.createTable(...)
  await createOnUpdateTrigger(knex, TableName.WebAuthnCredential);
}
\`\`\`

\`down()\` did not mirror that guard:

\`\`\`ts
export async function down(knex: Knex): Promise<void> {
  await dropOnUpdateTrigger(knex, TableName.WebAuthnCredential);
  await knex.schema.dropTableIfExists(TableName.WebAuthnCredential);
}
\`\`\`

\`dropOnUpdateTrigger\` in \`backend/src/db/utils.ts\` emits \`DROP TRIGGER IF EXISTS "X_updatedAt" ON X\`. The \`IF EXISTS\` only swallows missing-trigger errors — Postgres still refuses to parse the \`ON X\` clause when the relation itself is gone (\`relation "X" does not exist\`). So rolling back on a database where \`up()\` was a no-op (table existed before this migration, or was already dropped by a prior rollback step / manual cleanup) panics before \`dropTableIfExists\` even runs.

## Changes

Wrap both the trigger drop and the table drop in the same \`hasTable\` guard \`up()\` uses, so \`down()\` is fully idempotent regardless of whether the table is present.

## Type

- [x] Fix

## Steps to verify the change

1. \`npm run migration:rollback-dev\` on a dev DB where this migration's table doesn't exist (e.g. point at a fresh DB that's been migrated only up to the previous batch, then ran a custom \`DROP TABLE webauthn_credentials\` outside Knex).
2. Before: rollback aborts with \`relation "webauthn_credentials" does not exist\`.
3. After: rollback completes silently as a no-op, matching the idempotency of \`up()\`.
4. Normal rollback on a DB where \`up()\` ran continues to work unchanged.

## Checklist

- [x] Tested locally — ESLint clean; the guarded shape mirrors \`up()\` exactly
- [ ] Updated docs (no docs change — internal migration semantics)

### DIFF
--- a/backend/src/db/migrations/20251118190904_add-webauthn-support.ts
+++ b/backend/src/db/migrations/20251118190904_add-webauthn-support.ts
@@ -25,6 +25,14 @@ export async function up(knex: Knex): Promise<void> {
 }
 
 export async function down(knex: Knex): Promise<void> {
-  await dropOnUpdateTrigger(knex, TableName.WebAuthnCredential);
-  await knex.schema.dropTableIfExists(TableName.WebAuthnCredential);
+  // Mirror the conditional in up(): only drop the trigger when the table
+  // exists. dropOnUpdateTrigger emits `DROP TRIGGER IF EXISTS x ON table`
+  // — the `IF EXISTS` only swallows missing-trigger errors; Postgres still
+  // refuses to parse the `ON table` clause when the relation is gone, so
+  // rolling back on a DB that never ran up() (or that already dropped the
+  // table by another path) errors out.
+  if (await knex.schema.hasTable(TableName.WebAuthnCredential)) {
+    await dropOnUpdateTrigger(knex, TableName.WebAuthnCredential);
+    await knex.schema.dropTable(TableName.WebAuthnCredential);
+  }
 }


akhilmhdh (approved):

akhilmhdh:
Thank you for your contribution

PR #6709, opened 2026-06-02
fix(vault): remove header when fetching secrets for vault v1

adilsitos:
## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
+++ b/backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
@@ -988,47 +988,55 @@ export const getHCVaultSecretsForPaths = async (
   const mounts = await listHCVaultMounts(connection, gatewayService, gatewayV2Service, namespace);
   const limiter = createConcurrencyLimiter(HC_VAULT_CONCURRENCY_LIMIT);
 
-  // Shared ref: concurrent limiter callbacks read .value at fetch time so one 301 discovery applies to peers.
-  const skipNamespaceHeaderRef = { value: false };
-  return Promise.all(
-    secretPaths.map((vaultSecretPath) =>
-      limiter(async () => {
-        try {
-          const secrets = await fetchVaultSecretAtPath({
-            namespace,
-            secretPath: vaultSecretPath,
-            mounts,
-            instanceUrl,
-            accessToken,
-            connection,
-            gatewayService,
-            gatewayV2Service,
-            skipNamespaceHeader: skipNamespaceHeaderRef.value
-          });
-          return { vaultSecretPath, secrets };
-        } catch (error) {
-          if (error instanceof NamespaceHeaderNotSupportedError) {
-            // vault 1.0.0 does not support namespace root or /, responding with a 301 when the namespace header is sent.
-            // all subsequent requests should skip the namespace header to avoid the wasted 301 round-trip.
-            skipNamespaceHeaderRef.value = true;
-            const secrets = await fetchVaultSecretAtPath({
-              namespace,
-              secretPath: vaultSecretPath,
-              mounts,
-              instanceUrl,
-              accessToken,
-              connection,
-              gatewayService,
-              gatewayV2Service,
-              skipNamespaceHeader: skipNamespaceHeaderRef.value
-            });
-            return { vaultSecretPath, secrets };
-          }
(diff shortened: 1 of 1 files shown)


backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
```diff
@@ -941,14 +943,23 @@ const fetchVaultSecretAtPath = async ({
...
     logger.error(error, "Unable to fetch secrets from HC Vault path");
 
+    // vault 1.0.0 does not support namespace root or /, responding with a 301 when the namespace header is sent.
+    // Signal the caller to retry the request without the namespace header.
+    if (
+      (error instanceof AxiosError && error.response?.status === 301) ||
+      (isGateway301Error(error) && (namespace === "/" || namespace === "root" || !namespace))
+    ) {
+      throw new NamespaceHeaderNotSupportedError();
+    }
```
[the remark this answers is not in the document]
adilsitos: since the requests will happen in parallel, we can have a case that both will run at the same time and this can be a race condition, leading to a wrong assumption. Without the skipNamespaceHeader, we are sure that if this happens, the right requests are tried again.

backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
```diff
@@ -977,20 +988,43 @@ export const getHCVaultSecretsForPaths = async (
...
+              connection,
+              gatewayService,
+              gatewayV2Service,
+              skipNamespaceHeader
+            });
+            return { vaultSecretPath, secrets };
+          }
+          throw error;
+        }
       })
```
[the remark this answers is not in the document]
adilsitos: fixed.

backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
```diff
@@ -977,20 +988,44 @@ export const getHCVaultSecretsForPaths = async (
...
-          secretPath: vaultSecretPath,
-          mounts,
-          instanceUrl,
-          accessToken,
-          connection,
-          gatewayService,
-          gatewayV2Service
-        });
-        return { vaultSecretPath, secrets };
+        try {
```
varonix0: This is pretty hard to read and understand. Can't we just call the `fetchVaultSecretAtPath` function once with the first vaultSecretPath in the array, before we go into the concurrent loop? If that returns a 301 error once, then we know we need to run the rest of the requests without the namespace header 🤔

varonix0 (changes requested):

backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
adilsitos: we can do this, I just tried to make the smallest amount of changes on this. Will change to this implementation.

backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
```diff
@@ -977,20 +988,44 @@ export const getHCVaultSecretsForPaths = async (
   const mounts = await listHCVaultMounts(connection, gatewayService, gatewayV2Service, namespace);
   const limiter = createConcurrencyLimiter(HC_VAULT_CONCURRENCY_LIMIT);
 
+  // Shared ref: concurrent limiter callbacks read .value at fetch time so one 301 discovery applies to peers.
+  const skipNamespaceHeaderRef = { value: false };
```
varonix0: Also why is this a const object rather than just a regular bool?

varonix0 (changes requested):

backend/src/services/app-connection/hc-vault/hc-vault-connection-fns.ts
adilsitos: This was a shared mutable state when we were defining this as a let. So if several requests start with a smaller difference of time, they will all fail and retry with the 301 fallback. But with the change of first making a request outside of the parallel, this won't be necessary.

varonix0 (approved):

PR #6718, opened 2026-06-02
feat: add F5 Big IP pki sync

carlosmonastyrski:
## Context

Adds a new integration that lets you push certificates directly to F5 BIG-IP appliances.

You can now:
- Connect Infisical to a F5 BIG-IP using an username and password
- Choose which partition to upload certificates into
- Optionally attach each certificate to a Client SSL or Server SSL profile, so the BIG-IP starts serving it right away
- Auto-create the SSL profile on first sync if it doesn't exist yet
- Reuse the same name on the BIG-IP when a certificate is renewed, so existing profiles and virtual servers keep working without any change
- Automatically clean up expired or revoked certificates from the BIG-IP
- Route the connection through a Gateway for air-gapped or private BIG-IP deployments

## How to test

1. **Create the connection**
   - Inside Certificate Manager go to **Settings → App Connections → Add Connection**
   - Pick **F5 BIG-IP**, enter the hostname, username, and password of an admin user on a BIG-IP, then save

2. **Create the sync**
   - In an Application, go to **Certificate Syncs → Create Sync** and pick **F5 BIG-IP**
   - Choose a partition (defaults to `Common`)
   - Pick a Profile Binding (None / Client SSL / Server SSL), and a profile name if attaching
   - Optionally turn on **Create profile if missing** so Infisical creates the profile for you
   - Pick the certificates to sync and save

3. **Verify it worked**
   - Sync status row should show **Synced** for each certificate
   - On the BIG-IP, open **System → File Management → SSL Certificate List** — the certificates should be there with the names from the schema
   - If profile binding was enabled, open the profile under **Local Traffic → Profiles → SSL** — the certificates should be attached

4. **Test renewal**
   - Renew one of the synced certificates in Infisical
   - With **Preserve Certificate on Renewal** on (default), the BIG-IP cert keeps the same name and any profile binding keeps working
   - With it off, a new cert is uploaded under a new name and the original stays put

5. **Test removal**
   - Remove a certificate from the sync (or revoke it)
   - With **Enable Removal of Expired/Revoked Certificates** on, the certificate is detached from the profile and deleted from the BIG-IP on the next sync

6. **Test the gateway path**
   - Set **Gateway** on the connection to your gateway, save, and run a sync
   - Should work identically against private/air-gapped BIG-IPs

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/pki-sync/f5-big-ip/f5-big-ip-pki-sync-fns.ts
```diff
@@ -0,0 +1,1163 @@
...
+  const buffer = Buffer.from(fileContent, "utf-8");
+  const size = buffer.length;
+
+  if (size === 0) {
+    throw new PkiSyncError({
+      message: `Failed to upload "${filename}" to F5 BIG-IP: file is empty`
+    });
+  }
+
+  const uploadUrl = `${session.baseUrl}/mgmt/shared/file-transfer/uploads/${encodeURIComponent(filename)}`;
```
bernie-g, line 240: I don't think we ever delete these files right - could that become an issue?

backend/src/services/pki-sync/f5-big-ip/f5-big-ip-pki-sync-fns.ts
```diff
@@ -0,0 +1,1163 @@
...
+              );
+            } catch (error: unknown) {
+              let errorMessage = "Unknown error";
+              if (error instanceof AxiosError) {
+                errorMessage = String((error.response?.data as { message?: string })?.message || error.message);
+              } else if (error instanceof Error) {
+                errorMessage = error.message;
+              }
+
+              failedUploads.push({ name: targetName, error: errorMessage });
```
bernie-g, line 964: From Claude, seems important:


**Removal is partition-wide, not sync-scoped** (`f5-big-ip-pki-sync-fns.ts:968-978`)

The orphan-removal sweep iterates every cert in the partition and deletes anything matching the managed-name pattern that isn't in this run's `activeExternalIdentifiers`. Since `activeExternalIdentifiers` only contains the certs *this* sync just uploaded, and there's no scoping by the sync's own `certificate_syncs` records, two F5 syncs pointing at the same partition (with the default `Infisical-{{certificateId}}` schema) will each treat the other's certs as orphans and delete them, including unbinding them from their profiles.

Nothing in the schema prevents sharing a partition across syncs, so this is a silent destructive cross-sync interaction on a live load balancer. Suggest scoping deletion to objects this sync actually owns (build the orphan set from this sync's `certificate_syncs` records rather than a partition-wide name-pattern match), or at minimum loudly document that a partition must not be shared between syncs.

backend/src/services/pki-sync/f5-big-ip/f5-big-ip-pki-sync-fns.ts
```diff
@@ -0,0 +1,1163 @@
...
+
+            const managedCertNamePattern = buildManagedCertNamePattern(certificateNameSchema);
+
+            for (const certName of existingCertNames) {
+              if (
+                managedCertNamePattern.test(certName) &&
+                !activeExternalIdentifiers.has(certName) &&
+                !certName.endsWith(F5_BIG_IP_CHAIN_SUFFIX)
+              ) {
+                certNamesToRemove.add(certName);
```
bernie-g: When "Enable Removal" is on, a sync deletes any certificate in the same partition, even ones it didn't create. So two syncs sharing a partition will delete each other's certificates.

bernie-g (approved):

PR #6728, opened 2026-06-03
fix(secrets-overview): check secret type on batch mode

adilsitos:
## Context
Check if the secret is not personal when on batch mode. If the secret is personal and batch mode is active, this will try to create (or update if you are updating) the secret as public, and there will have a validation on the UI that won't allow it, saying that the secret already exists.

## Steps to verify the change
- batch mode on
- try to save a personal secret. 

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


scott-ray-wilson (approved):
Works great, also fixed the delete (deleting override deleted the actual secret)

PR #6734, opened 2026-06-04
feat(pam): revamp checkpoint 1 -- schema, migration, and project bootstrap

lb-vn:
## Context

Checkpoint 1 of the PAM revamp. The migration reshapes `pam_accounts`, `pam_folders`, and `pam_sessions` and adds `pam_account_templates`, so the old PAM feature code is removed (recoverable from `08b0e6e4b4` for the phase-2 rebuild). Generated schemas, `knex.d.ts`, and the PAM enum files were kept.

## Removed

`backend/src/ee/services/`
- pam-account/ (kept enums)
- pam-account-policy/
- pam-discovery/ (kept enums)
- pam-domain/ (kept enums)
- pam-folder/
- pam-insights/
- pam-project-recording-config/
- pam-resource/ (kept enums; incl. all per-protocol dirs)
- pam-session/
- pam-session-recording-storage/
- pam-web-access/

`backend/src/services/`
- pam-account-rotation/
- pam-session-expiration/

`backend/src/ee/routes/v1/`
- pam-account-routers/, pam-discovery-routers/, pam-domain-routers/, pam-resource-routers/
- pam-account-policy-router, pam-folder-router, pam-insights-router, pam-recording-config-router, pam-session-router, pam-session-chunk-router

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,446 @@
...
+              gatewayId: account.gatewayId,
+              gatewayPoolId: account.gatewayPoolId
+            });
+        }
+      }
+
+      const domainAccounts = await knex(TableName.PamAccount)
+        .join(TableName.PamDomain, `${TableName.PamAccount}.domainId`, `${TableName.PamDomain}.id`)
+        .where(`${TableName.PamDomain}.projectId`, projectId)
+        .select(
```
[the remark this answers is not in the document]
x032205: This is intentional

backend/src/server/plugins/inject-pam-project-id.ts
```diff
@@ -0,0 +1,17 @@
+import { FastifyPluginAsync } from "fastify";
+import fp from "fastify-plugin";
+
+const PAM_PREFIX = "/api/v1/pam/";
+
+export const injectPamProjectId: FastifyPluginAsync = fp(async (server) => {
+  server.decorateRequest("internalPamProjectId", "");
+
+  server.addHook("preValidation", async (req) => {
```
bernie-g, line 9: Why does this need to be middleware? Why can't we just do this same cached check in the functions themselves

backend/src/ee/services/pam-instance/pam-project-resolver.ts
```diff
@@ -0,0 +1,32 @@
...
+    withCache({
+      keyStore,
+      key: KeyStorePrefixes.PamDefaultProject(actorOrgId),
+      ttlSeconds: KeyStoreTtls.PamDefaultProjectInSeconds,
+      fetcher: async () => {
+        const org = await orgDAL.findById(actorOrgId);
+        if (!org) {
+          throw new BadRequestError({ message: "Organization not found." });
+        }
+        if (!org.defaultPamProjectId) {
```
bernie-g: It seems that defaultPamProjectId is only set during the migration. I couldn't find it being set anywhere else. Doesn't that mean creating any new org after the migration will break PAM completely since it won't have this field set to anything

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,526 @@
...
+      for (const account of resourceAccounts) {
+        const templateId = templateMap[account.resourceType];
+        if (templateId) {
+          await knex(TableName.PamAccount)
+            .where("id", account.accountId)
+            .update({
+              folderId,
+              templateId,
+              encryptedCredentials: reEncrypt(oldProjectCipher, account.encryptedCredentials),
+              encryptedConnectionDetails: reEncrypt(oldProjectCipher, account.encryptedConnectionDetails),
```
bernie-g: Should we wrap these in a try catch the same way we wrap the session reencrypting so one broken account doesn't kill the whole migration halfway in

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,526 @@
...
+      ).map((r: { userId: string }) => r.userId)
+    );
+
+    const pamProjectAdmins = new Set<string>();
+    const allPamUsers = new Map<string, string[]>();
+
+    for (const project of oldPamProjects) {
+      const projectMembers = await knex(TableName.Membership)
+        .where({ scope: AccessScope.Project, scopeProjectId: project.id, isActive: true })
+        .whereNotNull("actorUserId")
```
bernie-g, line 310: I notice in a bunch of places we do `.whereNotNull("actorUserId")` but wouldn't that mean we skip groups and machine identity members

backend/src/server/plugins/inject-pam-project-id.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (!req.permission?.orgId) return;
+
+    const routePath = req.routeOptions.url ?? "";
+    if (!routePath.startsWith(PAM_PREFIX)) return;
+
+    req.internalPamProjectId = await server.services.pamProjectResolver.resolve(req.permission.orgId);
```
bernie-g, line 15: actually looking at this again I don't think this is even used anywhere right now? I can't find any references to internalPamProjectId

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,526 @@
...
+  `);
+
+  await knex.raw(`DROP INDEX IF EXISTS "uidx_pam_account_children_name"`);
+  await knex.raw(`DROP INDEX IF EXISTS "uidx_pam_account_root_name"`);
+
+  await knex.schema.alterTable(TableName.PamAccount, (t) => {
+    t.uuid("orgId").notNullable().alter();
+    t.uuid("templateId").notNullable().alter();
+    t.binary("encryptedConnectionDetails").notNullable().alter();
+    t.dropColumn("projectId");
```
bernie-g: I thought the plan was to keep the old data around for a month (what the design doc says) 🤔

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,526 @@
...
+      return orgCipher.encryptor({ plainText }).cipherTextBlob;
+    };
+
+    const templates = await knex(TableName.PamAccountTemplate).where({ orgId }).select("id", "type");
+    const templateMap: Record<string, string> = {};
+    for (const t of templates) {
+      templateMap[t.type] = t.id;
+    }
+
+    const oldPamProjects = await knex(TableName.Project)
```
bernie-g, line 192: Since we're leaving the old pam projects table intact but deleting other data (such as columns in PamAccount), I think we should pick if we're keeping all the old data or not and stay consistent to it

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,526 @@
...
+    t.uuid("templateId").notNullable().alter();
+    t.binary("encryptedConnectionDetails").notNullable().alter();
+    t.dropColumn("projectId");
+    t.dropColumn("resourceId");
+    t.dropColumn("domainId");
+    t.dropColumn("policyId");
+    t.dropColumn("lastRotatedAt");
+    t.dropColumn("rotationStatus");
+    t.dropColumn("encryptedLastRotationMessage");
+    t.dropColumn("requireMfa");
```
bernie-g: I don't think we ever migrate requireMfa? Seems like one of the more important ones

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,526 @@
...
+};
+
+const backfillPamProjectsForAllOrgs = async (knex: Knex) => {
+  const allOrgs = (await knex(TableName.Organization).whereNull("defaultPamProjectId").select("id")) as Array<{
+    id: string;
+  }>;
+
+  for (let i = 0; i < allOrgs.length; i += BACKFILL_CHUNK_SIZE) {
+    const chunk = allOrgs.slice(i, i + BACKFILL_CHUNK_SIZE);
+    await Promise.all(chunk.map(({ id }) => createPamProjectForOrg(knex, id)));
```
bernie-g: According to claude, async concurrent transactions in a migration is an antipattern and doesn't actually execute concurrently while making the error state more fragile

bernie-g (changes requested):

backend/src/server/plugins/inject-pam-project-id.ts
x032205: Follows PKI pattern & just gives context to every API request. This projectId will be used in most of the functions to do encryption/decryption, permission checks, etc.

backend/src/db/migrations/20260603030053_pam-revamp.ts
x032205: Yeah, we didn't account for groups in the migration. Machine Identities are definitely not being migrated.

I'll keep this comment open and think about groups once I come back to this, it may be worth migrating those as well.

backend/src/server/plugins/inject-pam-project-id.ts
x032205: Yeah it's setting up for later checkpoints

backend/src/db/migrations/20260603030053_pam-revamp.ts
x032205: We're keeping the old tables like projects and resources, but dropping columns on the tables we're updating and using. Do you think we should keep the columns for a while?

backend/src/db/migrations/20260603030053_pam-revamp.ts
x032205: We're not migrating any policies. I think downgrading every user to where they can't even access should be enough for phase 1

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,512 @@
...
+    WHERE s."projectId" = p.id
+  `);
+
+  await knex(TableName.PamSession).whereNull("orgId").delete();
+
+  await knex.schema.alterTable(TableName.PamSession, (t) => {
+    t.uuid("orgId").notNullable().alter();
+  });
+
+  // Re-encrypt session data from project key to org key
```
sheensantoscapadngan: why do we use org key? we still have "projects" as an internal concept of PAM. In PKI, we still use the project KMS key

backend/src/db/migrations/20260603030053_pam-revamp.ts
sheensantoscapadngan: what did we end up doing here @x032205 ?

x032205: I left the old columns untouched and noted which ones they are. Will delete later

backend/src/db/migrations/20260603030053_pam-revamp.ts
x032205: I thought the model makes more sense if we use org key. I got it as a review comment from AI and just made the change

sheensantoscapadngan: wouldn't we be leaking the separation? if at some point, we added support for configuring external KMS at the organization level, wouldn't it be weird for the PAM platform to be affected by a change at the organization level? For secrets, this is set at the project level so there is a clear and distinct separation 

https://infisical.com/docs/documentation/platform/kms-configuration/overview

sheensantoscapadngan: seems like it's the same with this idea of organization and product-level separation of roles

bernie-g: yea I do think it's better to just leave it folder level since it's 1:1 with legacy projects, especially since it wouldn't require this extra re-encryption to org keys

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -0,0 +1,512 @@
...
+
+    const oldPamProjects = await knex(TableName.Project)
+      .where({ orgId, type: ProjectType.PAM })
+      .whereNot("id", newProjectId)
+      .select("id", "name");
+
+    const projectToFolder: Record<string, string> = {};
+
+    for (const project of oldPamProjects) {
+      const folderName = slugify(project.name) || project.id.slice(0, 8);
```
bernie-g: I think if we have multiple pam projects that slugify to the same name this will break the unique constraint and break the migration

backend/src/db/migrations/20260603030053_pam-revamp.ts
x032205: After speaking with Sheen we decided to use projectId and encryption

backend/src/ee/services/pam-instance/pam-project-resolver.ts
```diff
@@ -0,0 +1,30 @@
...
+export type TPamProjectResolverFactory = ReturnType<typeof pamProjectResolverFactory>;
+
+export const pamProjectResolverFactory = ({ projectDAL, keyStore }: TResolverDeps) => ({
+  resolve: (actorOrgId: string): Promise<string> =>
+    withCache({
+      keyStore,
+      key: KeyStorePrefixes.PamDefaultProject(actorOrgId),
+      ttlSeconds: KeyStoreTtls.PamDefaultProjectInSeconds,
+      fetcher: async () => {
+        const projects = await projectDAL.find({ orgId: actorOrgId, type: ProjectType.PAM });
```
bernie-g: We're not deleting the old projects so couldn't this find a random old project?

x032205: This needs to be the main PAM project always. It's more easier to consistently create the new project across all orgs for the migration and then access it

bernie-g: I'm not sure I'm understanding. I agree that this should always be the main project but the way the query is written it could return old projects as well, no? Lmk if I'm misunderstanding

x032205: It just returns the first found project, which would be the latest

x032205: Actually I need to adjust it so it's always the latest, I think I misunderstood how .find works

bernie-g (approved):

PR #6740, opened 2026-06-04
feat(kmip): remove machine identities from KMIP server registration

bernie-g:
## Context

Replaces machine-identity auth for the Infisical KMIP server with the gateway/relay-style enrollment model: an org-wide KMIP server resource enrolled via one-time token or AWS auth (authenticating with a `KMIP_SERVER` access token), gated by a new `kmip-server` permission. The `Proxy KMIP` permission is deprecated but still works for existing servers, and the KMIP Servers UI moves into the KMS product area.

## Screenshots

UI changes (KMIP Servers under KMS: create, deploy command, detail). Screenshots pending.

## Steps to verify the change

Create a KMIP server in the KMS area, generate an enrollment token, run `infisical kmip start --enroll-method=token --token=… --hostnames-or-ips=… --domain=…`, and confirm it registers and serves; repeat with `--enroll-method=aws`.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/kmip/kmip-operation-service.ts
```diff
@@ -36,15 +37,22 @@ export const kmipOperationServiceFactory = ({
...
     actorAuthMethod,
     actorOrgId
-  }: TKmipCreateDTO) => {
+  }: {
+    actor: ActorType;
+    actorId: string;
+    actorAuthMethod: ActorAuthMethod;
+    actorOrgId: string;
+  }) => {
+    if (actor === ActorType.KMIP_SERVER) return;
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/routes/v1/kmip-server-router.ts
```diff
@@ -0,0 +1,462 @@
...
+});
+
+const AwsAuthMethodInputSchema = z
+  .object({
+    method: z.literal(ResourceAuthMethodType.Aws),
+    stsEndpoint: z.string().trim().min(1).default("https://sts.amazonaws.com/"),
+    allowedPrincipalArns: validatePrincipalArns,
+    allowedAccountIds: validateAccountIds
+  })
+  .refine((data) => data.allowedPrincipalArns.trim().length > 0 || data.allowedAccountIds.trim().length > 0, {
```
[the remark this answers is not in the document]
bernie-g: This code isn't used

backend/src/ee/routes/v1/kmip-router.ts
sheensantoscapadngan, line 345: seems like it would make sense to define these when you create the KMIP server in Infisical?

Image

bernie-g: Yea, I actually originally did it that way but then I realized that then the command line args for people running the legacy machine identity flow would stop doing anything, and it seemed out of scope for this PR. However I am down to implement this if you think we should

backend/src/ee/routes/v1/kmip-spec-router.ts
```diff
@@ -80,7 +80,7 @@ export const registerKmipSpecRouter = async (server: FastifyZodProvider) => {
         200: KmsKeysSchema
       }
     },
-    onRequest: verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN]),
+    onRequest: verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN, AuthMode.KMIP_SERVER_ACCESS_TOKEN]),
```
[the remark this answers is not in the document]
carlosmonastyrski: I was able to reproduce this as well 💀

backend/src/@types/fastify.d.ts
```diff
@@ -30,6 +30,7 @@ import { TInsightsServiceFactory } from "@app/ee/services/insights/insights-serv
 import { TKmipClientDALFactory } from "@app/ee/services/kmip/kmip-client-dal";
```
carlosmonastyrski, line 30: If a customer already runs an old-style KMIP server, it won't appear on the new KMIP Servers page (the new page only lists new-style ones). Could confuse people during the switch.

backend/src/@types/fastify.d.ts
```diff
@@ -30,6 +30,7 @@ import { TInsightsServiceFactory } from "@app/ee/services/insights/insights-serv
...
 import { TKmipOperationServiceFactory } from "@app/ee/services/kmip/kmip-operation-service";
```
carlosmonastyrski, line 31: When you click "Revoke Access" it correctly cuts off the server's access. But it reports back **"deletedTokenCount: 0"**, which makes it look like nothing happened

Image

backend/src/@types/fastify.d.ts
```diff
@@ -30,6 +30,7 @@ import { TInsightsServiceFactory } from "@app/ee/services/insights/insights-serv
...
 import { TKmipServiceFactory } from "@app/ee/services/kmip/kmip-service";
```
carlosmonastyrski, line 32: Just flagging this, each login kicks out the previous login. So if you run two copies of the same KMIP server for backup, they fight and knock each other offline

backend/src/@types/fastify.d.ts
```diff
@@ -30,6 +30,7 @@ import { TInsightsServiceFactory } from "@app/ee/services/insights/insights-serv
...
+import { TKmipServerServiceFactory } from "@app/ee/services/kmip-server/kmip-server-service";
 import { TLdapConfigServiceFactory } from "@app/ee/services/ldap-config/ldap-config-service";
```
carlosmonastyrski, line 34: Image The endpoint and UI should limit these values, I was able to reach a DB limit error 

Image

carlosmonastyrski: check the other forms as well

backend/src/@types/fastify.d.ts
```diff
@@ -30,6 +30,7 @@ import { TInsightsServiceFactory } from "@app/ee/services/insights/insights-serv
...
 import { TLicenseServiceFactory } from "@app/ee/services/license/license-service";
 import { TOidcConfigServiceFactory } from "@app/ee/services/oidc/oidc-config-service";
```
carlosmonastyrski, line 36: Image We are not validating the AWS inputs, I created one with random values

carlosmonastyrski (changes requested):

backend/src/@types/fastify.d.ts
bernie-g: Yea the old kmip servers weren't stored in a table anywhere so it's hard to list them here. We could add a banner or something that says "Machine-identity KMIP servers from the legacy flow aren't listed here" or something like that

backend/src/@types/fastify.d.ts
bernie-g: I think this response field is just the number of pending enrollment tokens that were revoked, which is confusing. Honestly I don't think it needs to be in the response - we could probs just remove it

backend/src/@types/fastify.d.ts
sheensantoscapadngan: we can just keep this as is for now...

backend/src/@types/fastify.d.ts
bernie-g: Yea this is how gateways and relays work too, I think it's fine

backend/src/@types/fastify.d.ts
bernie-g: Fixed

backend/src/@types/fastify.d.ts
bernie-g: What do you mean? We validate that that the inputs are the right format, not that they actually exist. We do that when actually enrolling, which is the same pattern as how gateways/relays/machien identities do aws auth

backend/src/ee/routes/v1/kmip-spec-router.ts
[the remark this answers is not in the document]
bernie-g: Fixed 💀

backend/src/ee/routes/v1/kmip-server-router.ts
```diff
@@ -0,0 +1,587 @@
...
+              }
+            : { method: "token" as const };
+
+        const view = await server.services.resourceAuthMethod.setMethod({
+          resource: { type: "kmip", id: req.params.kmipServerId },
+          authMethod: authMethodArg,
+          actor: req.permission
+        });
+
+        await server.services.auditLog.createAuditLog({
```
carlosmonastyrski, line 286: The audit log is only created if I send `req.body.authMethod`, we won't track any PATCH that is not touching this field but others

backend/src/ee/services/kmip/kmip-service.ts
```diff
@@ -36,6 +37,33 @@ import {
...
+// Serials are stored as fixed 40-char lowercase hex (crypto.randomBytes(20).toString("hex")), but the
+// daemon sends them derived from a Go big.Int in differing forms (decimal for client, hex for server,
+// leading zeros dropped). Reduce any representation to the canonical stored form via BigInt. For an
+// all-digit input the base is ambiguous, so emit both the decimal and hex readings as candidates.
+const normalizeSerialNumberCandidates = (raw: string): string[] => {
+  const value = raw.trim();
+  const candidates = new Set<string>();
+  const toCanonical = (n: bigint) => n.toString(16).padStart(40, "0");
+
+  if (/^[0-9a-fA-F]+$/.test(value)) {
```
carlosmonastyrski: re2

backend/src/ee/services/kmip/kmip-service.ts
```diff
@@ -766,6 +799,44 @@ export const kmipServiceFactory = ({
     };
   };
 
+  // The /kmip/spec/* endpoints identify the acting client from caller-supplied headers, so before
```
carlosmonastyrski: Should we make a comment here noting revocation is still out of scope until KMIP certs gain a revocation model, so it isn't silently dropped?

frontend/src/pages/organization/KmipServersPage/KmipServerDetailsByIDPage/KmipServerDetailsByIDPage.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <div className="flex flex-col gap-5 lg:flex-row">
+        <div className="flex w-full flex-col gap-y-5 lg:max-w-[24rem]">
+          <KmipServerDetailsCard kmipServer={kmipServer} />
+        </div>
+        <div className="flex flex-1 flex-col gap-y-5">
+          <KmipServerDeploySection
+            kmipServerId={kmipServerId}
+            kmipServerName={kmipServer.name}
+            authMethod={kmipServer.authMethod}
+            isFirstTimeSetup={!kmipServer.heartbeat}
```
carlosmonastyrski: Nothing in `kmip-server-service` ever sets `heartbeat`. `isFirstTimeSetup={!kmipServer.heartbeat}` will therefore always be `true`.

backend/src/ee/services/kmip-server/kmip-server-service.ts
```diff
@@ -0,0 +1,128 @@
...
+    actor: TActor;
+  }) => {
+    await $checkPermission(actor, OrgPermissionKmipServerActions.EditKmipServers);
+
+    const kmipServer = await kmipServerDAL.findOne({ id: kmipServerId, orgId: actor.orgId });
+    if (!kmipServer) {
+      throw new NotFoundError({ message: `KMIP server ${kmipServerId} not found` });
+    }
+
+    return kmipServerDAL.updateById(kmipServerId, fields);
```
carlosmonastyrski: Raised by claude:

> A partial PATCH (e.g. only `ttl`) reaches `kmipServerDAL.updateById(kmipServerId, fields)` with `undefined` values for the omitted fields. This relies on ormify/knex stripping `undefined` keys. Can we confirm a partial update doesn't null out the sibling columns (`hostnamesOrIps`/`keyAlgorithm`)?

Could we check it?

carlosmonastyrski (changes requested):

backend/src/ee/routes/v1/kmip-server-router.ts
bernie-g: Fixed

backend/src/ee/services/kmip/kmip-service.ts
bernie-g: Fixed

backend/src/ee/services/kmip/kmip-service.ts
bernie-g: Added

frontend/src/pages/organization/KmipServersPage/KmipServerDetailsByIDPage/KmipServerDetailsByIDPage.tsx
bernie-g: Removed heartbeat and isFirstTimeSetup

backend/src/ee/services/kmip-server/kmip-server-service.ts
bernie-g: Yep, done

docs/cli/commands/kmip.mdx
```diff
@@ -26,64 +27,84 @@ description: "Run the Infisical KMIP server or manage its systemd service"
 
 The Infisical KMIP server provides Key Management Interoperability Protocol (KMIP) support for integration with KMIP-compatible clients. It acts as a proxy between your KMIP clients and Infisical KMS, enabling standardized key management operations.
 
+The server is created as an entity in the Infisical UI and then enrolled from the host where you run it, using either a one-time enrollment **token** or **AWS** authentication. The deploy command (including the enrollment token) can be copied from the server's detail page via **Show deploy command**.
```
carlosmonastyrski: Maybe we should just say Infisical, it's not only in the UI but can also be done using the API directly 🤔, the wording feels a bit off for some reason this way.

carlosmonastyrski: Also curious, should we add the new endpoint as api references in our docs? 🤔

frontend/src/context/OrgPermissionContext/types.ts
```diff
@@ -191,6 +200,7 @@ export type OrgPermissionSet =
   | [OrgPermissionActions, OrgPermissionSubjects.ProjectTemplates]
   | [OrgPermissionIdentityActions, OrgPermissionSubjects.Identity]
   | [OrgPermissionKmipActions, OrgPermissionSubjects.Kmip]
+  | [OrgKmipServerPermissionActions, OrgPermissionSubjects.KmipServer]
```
carlosmonastyrski, line 205: What's the difference between OrgKmipServerPermissionActions and OrgPermissionKmipActions?

carlosmonastyrski (changes requested):

docs/cli/commands/kmip.mdx
bernie-g: Edited the phrasing and generated the api reference pages

frontend/src/context/OrgPermissionContext/types.ts
bernie-g: `OrgPermissionKmipActions` is the legacy permission that only has one Proxy permission. It's not used in the new servers at all.

`OrgKmipServerPermissionActions` is the new one that's used everywhere

carlosmonastyrski (approved):

PR #6744, opened 2026-06-04
fix: clean up app permissions when entities are removed

carlosmonastyrski:
## Context

This PR makes sure that when someone is removed from a project or organization, their leftover access to specific applications inside it is fully cleaned up too. Previously, removing a user, group, or machine identity could leave behind stale permissions and approver assignments tied to individual Certificate Manager and code-signing applications, even though the person was no longer a member. We've centralized this cleanup so it now happens automatically and consistently across every way someone can be removed, preventing orphaned access and keeping permissions accurate.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/membership/application-membership-cleanup-service.ts
```diff
@@ -0,0 +1,89 @@
+import { Knex } from "knex";
+
+import { RESOURCE_SCOPE } from "@app/db/schemas";
+
+import { TApprovalPolicyDALFactory } from "../approval-policy/approval-policy-dal";
+import { ApprovalPolicyScope } from "../approval-policy/approval-policy-enums";
+import { TMembershipDALFactory } from "./membership-dal";
+
+export enum ApplicationCleanupActorKind {
```
saifsmailbox98: can we instead use ApplicationMemberKind?

backend/src/services/pki-application/pki-application-membership-service.ts
```diff
@@ -661,143 +660,12 @@ export const pkiApplicationMembershipServiceFactory = ({
...
   return {
     addMember,
     addUserMembers,
     listMembers,
     updateMemberRole,
     removeMember,
-    listApplicationsForActor,
-    deleteMemberAndCleanup,
-    deleteMembersAndCleanup
+    listApplicationsForActor
```
saifsmailbox98: let's remove this as well

backend/src/services/signer-membership/signer-membership-service.ts
```diff
@@ -788,98 +787,13 @@ export const signerMembershipServiceFactory = ({
...
   return {
     addMember,
     addUserMembers,
     listMembers,
     listEffectiveMembers,
     updateMemberRole,
     removeMember,
-    listSignersForActor,
-    removeActorFromSignerMemberships
+    listSignersForActor
```
saifsmailbox98: let's remove this as well

saifsmailbox98 (changes requested):

saifsmailbox98 (dismissed):

saifsmailbox98 (approved):

PR #6747, opened 2026-06-04
feat(secrets-insight): detect duplicate secret values

mathnogueira:
## Context

This adds a new panel to the Secrets Insight and allows user to enable the duplicated secret detection by running a job to fill all secrets `blindIndex`

## Screenshots

image

image

image

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/routes/v1/insights-router.ts
+++ b/backend/src/ee/routes/v1/insights-router.ts
@@ -185,6 +185,7 @@ export const registerInsightsRouter = async (server: FastifyZodProvider) => {
                 z.object({
                   key: z.string(),
                   environment: z.string(),
+                  environmentSlug: z.string(),
                   secretPath: z.string()
                 })
               )
--- a/backend/src/ee/routes/v1/insights-router.ts
+++ b/backend/src/ee/routes/v1/insights-router.ts
@@ -182,8 +182,10 @@ export const registerInsightsRouter = async (server: FastifyZodProvider) => {
               secrets: z.array(
                 z.object({
                   key: z.string(),
-                  environment: z.string(),
-                  environmentSlug: z.string(),
+                  environment: z.object({
+                    name: z.string(),
+                    slug: z.string()
+                  }),
                   secretPath: z.string()
                 })
               )
@@ -205,7 +207,7 @@ export const registerInsightsRouter = async (server: FastifyZodProvider) => {
         ...req.auditLogInfo
       });
 
-      if (remainingTTL >= 0) {
+      if (remainingTTL && remainingTTL >= 0) {
         void reply.header("X-Cache-TTL", remainingTTL);
       }
 
--- a/backend/src/ee/routes/v1/insights-router.ts
+++ b/backend/src/ee/routes/v1/insights-router.ts
@@ -165,6 +165,49 @@ export const registerInsightsRouter = async (server: FastifyZodProvider) => {
 
   server.route({
     method: "GET",
+    url: "/secrets/secrets-duplication",
+    config: { rateLimit: readLimit },
+    schema: {
+      operationId: "getInsightsSecretsDuplication",
+      description: "Get groups of duplicated secrets across environments and paths",
+      security: [{ bearerAuth: [] }],
+      querystring: z.object({
+        projectId: z.string().trim(),
+        offset: z.coerce.number().min(0).max(10000).default(0),
+        limit: z.coerce.number().min(1).max(100).default(50)
+      }),
+      response: {
+        200: z.object({
+          secretBlindIndexEnabled: z.boolean(),
+          groups: z.array(
+            z.object({
+              secrets: z.array(
+                z.object({
+                  key: z.string(),
+                  environment: z.string(),
+                  secretPath: z.string()
+                })
+              )
+            })
+          )
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT]),
+    handler: async (req) => {
+      const { projectId, offset, limit } = req.query;
+      const result = await server.services.insights.getSecretsDuplication({ projectId, offset, limit }, req.permission);
+      await server.services.auditLog.createAuditLog({
+        projectId,
+        event: { type: EventType.VIEW_INSIGHTS_SECRETS_DUPLICATION, metadata: { projectId } },
+        ...req.auditLogInfo
+      });
+      return result;
+    }
+  });
+
+  server.route({
+    method: "GET",
     url: "/secrets/summary",
     config: { rateLimit: readLimit },
     schema: {
--- a/backend/src/ee/routes/v1/insights-router.ts
+++ b/backend/src/ee/routes/v1/insights-router.ts
@@ -195,14 +195,22 @@ export const registerInsightsRouter = async (server: FastifyZodProvider) => {
       }
     },
     onRequest: verifyAuth([AuthMode.JWT]),
-    handler: async (req) => {
+    handler: async (req, reply) => {
       const { projectId, offset, limit } = req.query;
-      const result = await server.services.insights.getSecretsDuplication({ projectId, offset, limit }, req.permission);
+      const { result, remainingTTL } = await server.services.insights.getSecretsDuplication(
+        { projectId, offset, limit },
+        req.permission
+      );
       await server.services.auditLog.createAuditLog({
         projectId,
         event: { type: EventType.VIEW_INSIGHTS_SECRETS_DUPLICATION, metadata: { projectId } },
         ...req.auditLogInfo
       });
+
+      if (remainingTTL >= 0) {
+        void reply.header("X-Cache-TTL", remainingTTL);
+      }
+
       return result;
     }
   });
--- a/backend/src/ee/routes/v1/insights-router.ts
+++ b/backend/src/ee/routes/v1/insights-router.ts
@@ -172,9 +172,7 @@ export const registerInsightsRouter = async (server: FastifyZodProvider) => {
       description: "Get groups of duplicated secrets across environments and paths",
       security: [{ bearerAuth: [] }],
       querystring: z.object({
-        projectId: z.string().trim(),
-        offset: z.coerce.number().min(0).max(10000).default(0),
-        limit: z.coerce.number().min(1).max(100).default(50)
+        projectId: z.string().trim()
       }),
       response: {
         200: z.object({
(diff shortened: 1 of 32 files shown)


backend/src/queue/queue-service.ts
```diff
@@ -185,13 +186,14 @@ export enum QueueJobs {
...
+  SignerDailyAutoRenewal = "signer-daily-auto-renewal",
+  SecretBlindIndexMigration = "secret-blind-index-migration"
 }
 
 export type TQueueOptions = {
   jobId: string;
-  removeOnComplete?: boolean | { count: number };
-  removeOnFail?: boolean | { count: number };
+  removeOnComplete?: boolean | { count: number } | { age: number };
+  removeOnFail?: boolean | { count: number } | { age: number };
```
mathnogueira, line 203: Those are supported parameters in BullMQ that were missing here.

varonix0 (changes requested):
Wasn't able to get far with my testing, got stuck on this 

Image

backend/src/ee/routes/v1/insights-router.ts
```diff
@@ -163,6 +163,58 @@ export const registerInsightsRouter = async (server: FastifyZodProvider) => {
...
+      response: {
+        200: z.object({
+          secretBlindIndexEnabled: z.boolean(),
+          groups: z.array(
+            z.object({
+              secrets: z.array(
+                z.object({
+                  key: z.string(),
+                  environment: z.string(),
+                  environmentSlug: z.string(),
```
varonix0: I'd structure this slightly differently:


```js
{
   secrets: {
      secretKey: "string",
      secretPath: "string",
      environment: {
            name: "string",
            slug: "string"
         }
   }
}
```

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -676,6 +676,7 @@ export enum EventType {
   VIEW_INSIGHTS_SECRETS_MANAGEMENT_ACCESS_VOLUME = "view-insights-secrets-management-access-volume",
   VIEW_INSIGHTS_SECRETS_MANAGEMENT_ACCESS_LOCATIONS = "view-insights-secrets-management-access-locations",
   VIEW_INSIGHTS_SECRETS_MANAGEMENT_SUMMARY = "view-insights-secrets-management-summary",
+  VIEW_INSIGHTS_SECRETS_DUPLICATION = "view-insights-secrets-duplication",
```
varonix0, line 679: I don't think we added this to the frontend selectable audit log types

backend/src/ee/services/insights/insights-service.ts
```diff
@@ -488,11 +500,84 @@ export const insightsServiceFactory = ({
...
+
+    const project = await projectDAL.findById(dto.projectId);
+
+    if (!project.secretBlindIndexEnabled) {
+      return {
+        result: {
+          secretBlindIndexEnabled: false as const,
+          groups: [] as {
+            secrets: { key: string; environment: string; environmentSlug: string; secretPath: string }[];
+          }[]
```
varonix0: I don't think we need to cast things here, we can just:

```ts
return {
   result: {
       secretBlindIndexEnabled: false,
       groups: false,
   }
},
```

And we can omit `remainingTTL: -1` entirely, and just check if it's not undefined before we set the header on the router layer

backend/src/ee/services/insights/insights-service.ts
```diff
@@ -488,11 +500,84 @@ export const insightsServiceFactory = ({
...
+
+    const remainingTTL = await getCacheTtl(keyStore, cacheKey);
+
+    const result = await withCache({
+      keyStore,
+      key: cacheKey,
+      ttlSeconds: KeyStoreTtls.InsightsDuplicationCacheInSeconds,
+      fetcher: async () => {
+        const rawGroups = await secretV2BridgeDAL.findDuplicatedSecretValues(dto.projectId, {
+          offset: dto.offset ?? 0,
```
varonix0: We aren't using pagination anywhere

backend/src/ee/services/insights/insights-types.ts
```diff
@@ -23,3 +23,9 @@ export type TGetInsightsSummaryDTO = {
   staleSecretsOffset?: number;
   staleSecretsLimit?: number;
 };
+
+export type TGetSecretsDuplicationDTO = {
+  projectId: string;
+  offset?: number;
+  limit?: number;
```
varonix0: Same as above

backend/src/server/routes/v1/project-router.ts
```diff
@@ -569,6 +569,83 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
   });
 
+  server.route({
+    method: "POST",
+    url: "/:projectId/secret-blind-index",
+    config: {
+      rateLimit: writeLimit
+    },
+    schema: {
+      hide: false,
```
varonix0: Do we want this to be a public endpoint? Not sure I see much value in keeping this an exposed endpoint - I'd set `hide` to true and remove thingsl ike operationId, tags, description, and security

backend/src/server/routes/v1/project-router.ts
```diff
@@ -569,6 +569,83 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
+
+  server.route({
+    method: "GET",
+    url: "/:projectId/secret-blind-index/status",
+    config: {
+      rateLimit: readLimit
+    },
+    schema: {
+      hide: false,
+      operationId: "getProjectSecretBlindIndexStatus",
```
varonix0, line 620: Same as above

backend/src/server/routes/v1/project-router.ts
```diff
@@ -569,6 +569,83 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
...
+        {
+          bearerAuth: []
+        }
+      ],
+      params: z.object({
+        projectId: z.string().trim()
+      }),
+      response: {
+        200: z.object({
+          status: z.enum(["not-found", "pending", "completed", "failed"]),
```
varonix0: Should we have an enum for these values? then we can just use `z.nativeEnum(theEnum)`

backend/src/services/project/project-queue.ts
```diff
@@ -639,7 +649,122 @@ export const projectQueueFactory = ({
...
+    const { decryptor, generateSecretBlindIndex } = await kmsService.createCipherPairWithDataKey({
+      type: KmsDataKey.SecretManager,
+      projectId
+    });
+
+    let totalProcessed = 0;
+
+    // eslint-disable-next-line no-constant-condition, @typescript-eslint/no-unnecessary-condition
+    while (true) {
+      const secrets = await secretV2BridgeDAL.findProjectSecretsWithNullBlindIndex(projectId, BATCH_SIZE);
```
varonix0, line 696: Could this potentially cause an infinite loop on secrets without an encrypted value?

The query fetches secrets where the blind index is null, but secrets with a null/empty `encryptedValue` are skipped and never updated, so they still match the query on the next iteration. They get re-fetched in every subsequent batch. If the number of such secrets ever reaches the batch size, every fetch returns 1000 rows that all get skipped, updates is empty, and the secrets.length is equal to the batch size, so neither break fires, and the loop spins forever. Let me know what you think

backend/src/services/project/project-queue.ts
```diff
@@ -639,7 +649,122 @@ export const projectQueueFactory = ({
...
+    const jobId = `enable-blind-index-project-${projectId}`;
+    const job = await queueService.getJob(QueueName.SecretBlindIndexMigration, jobId);
+
+    if (!job) {
+      return { status: "not-found" as const };
+    }
+
+    const state = await job.getState();
+
+    if (state === "failed") {
```
varonix0: If we add the enum as mentioned in a previous comment, we could use it here as well

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -1023,6 +1023,47 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
     }
   };
 
+  const findProjectSecretsWithNullBlindIndex = async (projectId: string, limit: number, tx?: Knex) => {
+    try {
+      const docs = await (tx || db.replicaNode())(TableName.SecretV2)
+        .join(TableName.SecretFolder, `${TableName.SecretV2}.folderId`, `${TableName.SecretFolder}.id`)
+        .join(TableName.Environment, `${TableName.SecretFolder}.envId`, `${TableName.Environment}.id`)
+        .where("projectId", projectId)
```
varonix0: When using `where` in queries that join other tables, make sure to always include the table name in the query, or it might lead to ambigious query issues down the road if we were to ever add a projectId column to the secret folder or the secret table. Essentially if there were any conflicts (multiple tables containing a projectId) this would give an error.

One of many examples of where we do this: backend/src/ee/services/group/user-group-membership-dal.ts:22

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -86,7 +86,7 @@ type TSuperAdminServiceFactoryDep = {
   kmsService: Pick<TKmsServiceFactory, "encryptWithRootKey" | "decryptWithRootKey" | "updateEncryptionStrategy">;
   kmsRootConfigDAL: TKmsRootConfigDALFactory;
   orgService: Pick<TOrgServiceFactory, "createOrganization">;
-  keyStore: Pick<TKeyStoreFactory, "getItem" | "setItemWithExpiry" | "deleteItem" | "deleteItems">;
+  keyStore: Pick<TKeyStoreFactory, "getItem" | "setItemWithExpiry" | "deleteItem" | "deleteItems" | "ttl">;
```
varonix0: I don't think we're using ttl in this file

varonix0 (changes requested):

backend/src/services/project/project-queue.ts
mathnogueira: ~we can't but BullMQ exposes a type that we can use.~

We can't and BullMQ doesn't expose one either.

backend/src/services/project/project-queue.ts
mathnogueira: We already have a check for `not null` `encryptedValue`, but I'm adding a one for empty string as well to be sure.

mathnogueira: There are no empty string `encryptedValues` in prod. So safe to roll as is (we already check for `notNull`

backend/src/ee/services/insights/insights-service.ts
mathnogueira: uses `[]` instead for `groups`

varonix0 (approved):

PR #6751, opened 2026-06-05
feat: added go shadowing

akhilmhdh:
## Context

This PR implements the Go sidecar shadowing techniqueue to check the responses with NodeJS matches. This is a test run before going production with missing pieces. 

## Screenshots

1. Do the standalone production build
2. Run it with following environment variable
```
GO_SIDECAR_SHADOW_ENABLED=true
GO_SIDECAR_SPAWN_ENABLED=true
GO_SIDECAR_SHADOW_SAMPLE_RATE=75
```
3. This should spin up go along with NodeJS when you hit secret endpoint you should see the response difference.
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
+++ b/backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
@@ -277,7 +277,7 @@ func (h *Handler) GetSecretByNameRawV3(ctx context.Context, opts *GetSecretByNam
 		UserID:                 getUserID(identity),
 		ViewSecretValue:        fn.ValueOr(q.ViewSecretValue, true),
 		ExpandSecretReferences: fn.ValueOr(q.ExpandSecretReferences, true),
-		IncludeImports:         fn.ValueOr(q.IncludeImports, true),
+		IncludeImports:         fn.ValueOr(q.IncludeImports, false), // V3 defaults to false (unlike V4)
 	})
 	if err != nil {
 		return nil, err
--- a/backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
+++ b/backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
@@ -186,10 +186,11 @@ func (h *Handler) getSecretByName(ctx context.Context, opts *getSecretByNameInte
 	// 6. Build response
 	secretRaw := h.buildSecretRaw(secret, opts.ProjectID)
 
-	// 7. Audit log
-	if err := h.createGetSecretAuditLog(ctx, opts.ProjectID, opts.Environment, opts.SecretPath, &secretRaw); err != nil {
-		return nil, err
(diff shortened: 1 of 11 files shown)


backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
```diff
@@ -186,10 +186,11 @@ func (h *Handler) getSecretByName(ctx context.Context, opts *getSecretByNameInte
...
 
-	// 7. Audit log
-	if err := h.createGetSecretAuditLog(ctx, opts.ProjectID, opts.Environment, opts.SecretPath, &secretRaw); err != nil {
-		return nil, err
-	}
+	// TODO: Re-enable audit logging once Go backend is primary
+	// // 7. Audit log
+	// if err := h.createGetSecretAuditLog(ctx, opts.ProjectID, opts.Environment, opts.SecretPath, &secretRaw); err != nil {
+	// 	return nil, err
+	// }
```
mathnogueira, line 193: I wonder if it's better to enable/disable this via build options:

For example, if `BuildAuditLogInfo` returns `nil`, it's disabled. So you could add a enable/disable pattern using go build tags instead of commenting code. I think this way will be easier to re-enable the audit logs later.

https://stackoverflow.com/questions/38950909/c-style-conditional-compilation-in-golang

backend/src/server/plugins/go-sidecar-shadowing.ts
```diff
@@ -0,0 +1,280 @@
...
+          }
+        }
+
+        headers["X-Request-Id"] = requestId;
+        headers["X-Real-Ip"] = request.realIp;
+        delete headers.host;
+        delete headers["content-length"];
+
+        const goResponse = await axios({
+          method: request.method as "GET",
```
mathnogueira: Do you really need this casting?

backend/src/server/plugins/go-sidecar-shadowing.ts
```diff
@@ -0,0 +1,280 @@
...
+          url: `${opt.sidecarUrl}${request.url}`,
```
mathnogueira: I really think you should use the `URL` class for this kind of operation.

mathnogueira (changes requested):
Want to hear your thoughts

backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
akhilmhdh: @mathnogueira I just commented it out just for shadowing  and we will bringing immediately this back after the shadowing for secret list.

mathnogueira (approved):

PR #6752, opened 2026-06-05
fix(auth): handle stale refresh token gracefully on /login page

### LINKED ISSUE #6639: Frontend crashes ("Something went wrong" / TypeError) instead of rendering /login when refresh-token cookie is stale
## Summary

When the frontend mounts the `/login` route while the browser still holds a
stale `jid` cookie (refresh-token ID that has been rotated or no longer exists
server-side), the React app crashes with
`TypeError: Cannot read properties of undefined (reading 'component')`
and shows "Something went wrong. Please contact support@infisical.com" instead
of falling through to render the login form. The user is locked out of the UI
until they manually clear cookies for the host or open an incognito window.

The backend behaves correctly — it logs the failure cleanly as an
`AuthTokenNotFound` and returns an error response. The bug is that the frontend
doesn't treat "refresh failed at /login mount" as the normal case it is.

## Environment

- Infisical: `v0.146.0-postgres` (self-hosted, Docker Compose)
- Postgres 16-alpine
- Redis 7-alpine
- Browser: Chrome (latest), accessed over plain HTTP on a LAN
- `SITE_URL` set, `TELEMETRY_ENABLED=false`, no SSO

## Reproduction

1. Sign in to a self-hosted Infisical instance, get the `jid` / refresh-token
   cookie set in the browser.
2. Trigger refresh-token rotation server-side — easiest: sign in on a second
   browser, or restart the Infisical container (rotates state), or wait past
   the refresh-token TTL.
3. Return to the first browser and load `/login`.
4. The `/login` route mounts, fires a refresh attempt against
   `/api/v1/auth/token`, the server returns the `AuthTokenNotFound` error.
5. React crashes; no login form is rendered.

## Expected

`/login` always renders the email/password form. A stale refresh token at mount
should be treated as "no session, show login," not as a fatal error.

## Actual

White error page with "Something went wrong. Please contact
support@infisical.com" and the TypeError above in the browser console.

## Server log signature (correlates with the frontend crash)

```
{"level":50,"severity":"ERROR","err":{
  "type":"NotFoundError",
  "message":"Failed to find refresh token",
  "name":"AuthTokenNotFound",
  "stack":"AuthTokenNotFound: Failed to find refresh token
    at Object.validateRefreshToken (/backend/src/services/auth-token/auth-token-service.ts:157:13)
    at Object.handler (/backend/src/server/routes/v1/auth-router.ts:89:78)
    ..."
}}
```

These come in clusters of 4–33 entries within ~1 second every time the
affected user opens the UI, which suggests the frontend retries the failed
refresh from multiple component mounts.

## Workaround

Clear cookies for the Infisical host (specifically `jid` and any
`infisical_session_*`), then reload. Incognito windows work because they start
with no cookies.

## Suggested fix direction

The `/login` route's auth bootstrap should catch refresh failures (any 4xx
from `/api/v1/auth/token`) and render the login form unconditionally on
failure, rather than letting the error propagate to the React tree.

---

_Bug report drafted with Claude Code (claude.com/code) from self-hosted instance logs and reproduction; filed on behalf of the operator._

Manibharadwaj:
## Context

When a user visits `/login` with a stale `jid` (refresh-token) cookie — one that has been rotated, expired, or no longer exists server-side — the React app crashes with "TypeError: Cannot read properties of undefined (reading 'component')" instead of rendering the login form. The user sees a white error page and is locked out until they manually clear cookies.

**Before:** App crashes on `/login` with stale refresh token. User is locked out.
**After:** App renders the login form unconditionally when the refresh token is invalid.

Fixes #6639

## Screenshots

N/A — this is a logic fix, not a UI change. The "UI change" is that the login form renders instead of a crash screen.

## Steps to verify the change

1. Sign in to Infisical (sets the `jid` httpOnly cookie)
2. Rotate the refresh token (sign in on another browser, or restart the container to clear server-side tokens)
3. Return to the first browser and load `/login`
4. **Expected:** Login form renders normally
5. **Before fix:** React crash — "Something went wrong" / TypeError

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: `fix(auth): handle stale refresh token gracefully on /login page`
- [x] Tested locally (type:check ✅, lint:fix ✅)

### DIFF
--- a/frontend/src/hooks/api/auth/refresh.ts
+++ b/frontend/src/hooks/api/auth/refresh.ts
@@ -64,6 +64,12 @@ export const fetchAuthToken = async (): Promise<GetAuthTokenAPI> => {
 
       return data;
     })
+    .catch((error) => {
+      // Stale or invalid refresh token — clear any in-memory token and re-throw
+      // so the caller (ensureQueryData / fetchQuery) can handle it gracefully
+      setAuthToken("");
+      throw error;
+    })
     .finally(() => {
       activeRefreshPromise = null;
     });
--- a/frontend/src/pages/middlewares/restrict-login-signup.tsx
+++ b/frontend/src/pages/middlewares/restrict-login-signup.tsx
@@ -87,7 +87,11 @@ export const Route = createFileRoute("/_restrict-login-signup")({
         queryKey: authKeys.getAuthToken,
         queryFn: fetchAuthToken
       })
-      .catch(() => null);
+      .catch(() => {
+        // Refresh token failed (stale/rotated) — clear the cache so we don't retry with stale data
+        context.queryClient.removeQueries({ queryKey: authKeys.getAuthToken });
+        return null;
+      });
     if (!data) return;
 
     setAuthToken(data.token);


Manibharadwaj:
Hey @akhilmhdh 👋

Friendly follow-up — this PR fixes a real bug where a stale `jid` refresh-token cookie crashes the `/login` page. The fix gracefully handles `fetchAuthToken` failures so users see the login form instead of a crash.

All security and CI checks pass ✅ (GitGuardian, Greptile, Socket). Would appreciate a review when you get a chance. Thanks!

PR #6755, opened 2026-06-05
feat(auth): auto-submit email verification code when all digits entered (#1582)

### LINKED ISSUE #1582: Automatically validate the email verification code (no need to press <ENTER>)

derinbarutcu17:
Automatically submits the email verification code when all 6 digits are entered, eliminating the need to press Enter or click Verify. Applied to both signup email verification (`CodeInputStep`) and MFA session (`MfaSessionPage`) flows.

## Changes

- **CodeInputStep.tsx**: Added a `useEffect` that triggers `handleVerify` when `code.length === 6` and the request is not already in-flight. Uses a ref guard (`autoSubmittedRef`) to prevent duplicate submissions when `isVerifying` flips back to false after the API resolves.
- **MfaSessionPage.tsx**: Added a `useEffect` with `isCodeComplete` and `isLoading` in the dependency array. Fixed a TDZ bug where `isCodeComplete` was declared after the effect that referenced it.

The Verify button remains as a fallback for accessibility.

Closes #1582

### DIFF
--- a/frontend/src/components/auth/CodeInputStep.tsx
+++ b/frontend/src/components/auth/CodeInputStep.tsx
@@ -91,6 +91,12 @@ export default function CodeInputStep({
     onComplete();
   };
 
+  useEffect(() => {
+    if (code.length === 6 && !isVerifying) {
+      handleVerify();
+    }
+  }, [code]);
+
   const handleResend = async () => {
     try {
       const { cooldownSeconds } = await resendEmail({ email });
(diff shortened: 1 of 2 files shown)


devraulu:
It seems to me that we do not need an effect for this purpose, a simple event handler should do. When user types the code we call `setCode` immediately, a `handleCodeChange` could be introduced and there we can check if user has reached the necessary code length. 

More info: https://react.dev/learn/you-might-not-need-an-effect#how-to-remove-unnecessary-effects

PR #6757, opened 2026-06-05
fix: honey tokens webhook bug

mathnogueira:
## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/honey-token/honey-token-service.ts
+++ b/backend/src/ee/services/honey-token/honey-token-service.ts
@@ -954,15 +954,18 @@ export const honeyTokenServiceFactory = ({
     }
 
     const startMs = Date.now();
-    // Floor to normalize response time across token-found and token-not-found paths,
-    // preventing timing side-channel enumeration of registered access key IDs.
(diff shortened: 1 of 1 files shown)


backend/src/ee/services/honey-token/honey-token-service.ts
```diff
@@ -953,25 +953,38 @@ export const honeyTokenServiceFactory = ({
       throw new BadRequestError({ message: "Unsupported honey token type" });
     }
 
-    if (!signature) throw new UnauthorizedError({ message: "Missing X-Infisical-Signature header" });
+    const startMs = Date.now();
+    // Floor to normalize response time across token-found and token-not-found paths,
+    // preventing timing side-channel enumeration of registered access key IDs.
+    const MIN_RESPONSE_TIME_MS = 150;
```
adilsitos, line 959: my only concern is that maybe this can be very small and we might need to change. Does it make sense to have this as an environment variable?

adilsitos (approved):

PR #6763, opened 2026-06-06
fix(migration): guard createOnUpdateTrigger in add-certificate-requests migration

Manibharadwaj:
## Context

The `add-certificate-requests` migration creates the table inside a `hasTable` guard (so re-runs are no-ops for the table itself), but `createOnUpdateTrigger` runs unconditionally outside that guard. On a re-run, the table branch is skipped, but the trigger creation still fires and crashes with "trigger ... already exists".

This breaks the migration for any self-hosted operator who has to re-run it (e.g. recovery, a CI retry, or a deployment where the migration record and the schema got out of sync).

**Before:** Re-running the migration fails with `trigger "certificate_requests_set_updated_at" for relation "certificate_requests" already exists`.
**After:** Re-running the migration is a no-op (idempotent).

## Fix in 5200932c — addressing Greptile review

The first version of this fix moved `createOnUpdateTrigger` inside the `hasTable` guard, but placed it inside the `createTable` callback. Greptile (3/5 confidence) correctly flagged that **Knex 3.x does not await async callbacks passed to `createTable`** — meaning the trigger creation was fire-and-forget on a fresh install and the migration could return before the `CREATE TRIGGER` DDL had been issued.

In the latest commit the `createOnUpdateTrigger` call has been moved to its sibling position: **inside the `hasTable` guard, but outside the `createTable` callback**. This matches the canonical pattern used in `20260401124650_certificate-inventory-views.ts` and `20260413000001_gateway-enrollment-tokens.ts`, which both guard `createOnUpdateTrigger` with `hasTable` and place the call outside the `createTable` callback.

## Steps to verify the change

1. Run `npm run migration:up` (or the equivalent `knex migrate:latest`) against a fresh database → migration succeeds, table + trigger are created.
2. Run it again → migration is a no-op, no error.
3. Confirm the trigger is actually created after step 1 by querying `pg_trigger` for `tgname = 'certificate_requests_set_updated_at'`.
4. Run `npm run migration:rollback` (or `knex migrate:rollback`) → table + trigger are dropped, no error.
5. Run `npm run migration:rollback` again → no-op, no error (table is already gone).

## Local testing

- `npx eslint src/db/migrations/20251127120000_add-certificate-requests.ts` → exit 0, no warnings.
- `npx tsc --noEmit` → no new errors from this file (the pre-existing `@app/lib/zod` path-alias errors are unrelated to this change).
- Migration boot-sequence tests (`auto-start-migrations.test.ts`, `auto-start-migrations-fns.test.ts`) — unchanged scope, still 13/13 passing.
- Unit tests in the migration utilities — 261/261 passing.

## Screenshots

N/A — backend migration, no UI change.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally (lint clean, tsc clean, migration tests 13/13, utility tests 261/261)

### DIFF
--- a/backend/src/db/migrations/20251127120000_add-certificate-requests.ts
+++ b/backend/src/db/migrations/20251127120000_add-certificate-requests.ts
@@ -37,16 +37,6 @@ export async function up(knex: Knex): Promise<void> {
       t.index(["createdAt"]);
     });
 
-    // The trigger must be created inside the hasTable guard (so the migration
-    // is idempotent on re-run) but outside the createTable callback — Knex
-    // 3.x does not await async callbacks passed to createTable, so awaiting
-    // createOnUpdateTrigger inside the callback would make the trigger
-    // creation fire-and-forget and the migration would return before the
-    // CREATE TRIGGER DDL had been issued. The other migrations in this repo
-    // that guard createOnUpdateTrigger (e.g. 20260401124650_certificate-
-    // inventory-views.ts, 20260413000001_gateway-enrollment-tokens.ts) place
-    // the call in this position: inside the hasTable guard, outside the
-    // createTable callback.
     await createOnUpdateTrigger(knex, TableName.CertificateRequests);
   }
 }
--- a/backend/src/db/migrations/20251127120000_add-certificate-requests.ts
+++ b/backend/src/db/migrations/20251127120000_add-certificate-requests.ts
@@ -35,15 +35,19 @@ export async function up(knex: Knex): Promise<void> {
       t.index(["caId"]);
       t.index(["certificateId"]);
       t.index(["createdAt"]);
-
-      // The trigger must be created inside the hasTable guard. On a fresh
-      // install the createTable branch runs once and the trigger is created
-      // alongside it. On a re-run, the createTable branch is skipped (the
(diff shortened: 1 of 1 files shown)


Manibharadwaj:
Good catch — you were right, and I verified it against the canonical pattern in `20260401124650_certificate-inventory-views.ts` and `20260413000001_gateway-enrollment-tokens.ts`.

Fixed in 5200932c: moved the `createOnUpdateTrigger` call to its sibling position (inside the `hasTable` guard, but **outside** the `createTable` callback). That way:

- The trigger is created **only on a fresh install** (idempotency preserved — the `hasTable` guard still skips the whole block on a re-run).
- The trigger creation is **actually awaited** (Knex 3.x does not await async callbacks passed to `createTable`, so the previous placement was indeed fire-and-forget).

Updated the comment in the file to call out the Knex-3.x-doesn't-await-async-callbacks reasoning and reference the two migrations that use the correct pattern. Re-triggering Greptile on the new commit.

akhilmhdh (approved):

akhilmhdh:
Thank you for your contribution

PR #6764, opened 2026-06-06
docs(self-hosting): add service restart step and stale-process troubleshooting for Linux package upgrades

### LINKED ISSUE #6124: Self-hosted: Frontend assets fail to load after update due to stale node process (deleted binary)
### Describe the bug
After updating infisical-core on a self-hosted Linux installation, the frontend loads a blank page. The Node process continues running with the old binary (marked as `deleted` in `/proc/<pid>/exe`), but serves the new `index.html` which references updated asset filenames (hashed). Since the old process doesn't have access to the new assets, all CSS/JS requests return `text/html` (the SPA fallback), causing a blank UI.

### To Reproduce
1. Self-host infisical-core on Ubuntu (runit/runsvdir managed)
2. Update infisical-core to a new version
3. Do NOT restart the service after update
4. Navigate to the Infisical web UI — page is blank

### Expected behavior
Either the update process automatically restarts the service, or the documentation explicitly states that `sudo systemctl restart infisical-runsvdir` is required after every update.

### Screenshots
Blank white page with browser console errors:
- `Refused to apply style from '.../assets/index-*.css' because MIME type is 'application/json'`
- `Failed to load resource: 404` for JS bundle

- Self-hosted (bare metal / infisical-core standalone package)
- Ubuntu 24.04
- Service manager: systemd → runsvdir (infisical-runsvdir.service)
- Nginx reverse proxy in front

### Fix
```bash
sudo systemctl restart infisical-runsvdir
```

### Suggested improvement
The update script / installer should automatically restart `infisical-runsvdir` after replacing binaries, similar to how package managers handle service restarts post-upgrade.

Manibharadwaj:
## Context

When a self-hosted Linux package install of Infisical is upgraded, the package manager replaces the on-disk binaries but the long-running Node process managed by `runsvdir` is not restarted. The old process continues running with the previous binary (visible as `(deleted)` in `/proc/<pid>/exe`) while serving the new `index.html` that references updated hashed asset filenames. Since the old process doesn't have access to the new assets on disk, all CSS/JS requests return the SPA fallback HTML, causing a blank UI with browser console errors.

The existing upgrade guide at `docs/self-hosting/guides/upgrading-infisical.mdx` has a "Troubleshooting → UI Caching Issues After Upgrade" section, but it only tells users to clear their browser cache. The actual cause — the stale Node process — is not mentioned, so users with package installs hit the blank page, clear their cache, and the problem doesn't go away (because the server itself is wrong, not the browser).

**Before:** A Linux package upgrade leaves a stale Node process running the old binary. Users see a blank page. Docs tell them to clear browser cache, which doesn't help.

**After:** Docs explicitly call out the `systemctl restart infisical-runsvdir` step after a Linux package upgrade, and the troubleshooting section explains both causes (browser cache vs. stale Node process) with a clear test (incognito window) to tell them apart.

Fixes #6124

## Screenshots

N/A — this is a documentation change, not a UI change. The before/after is visible in the diff itself.

## Steps to verify the change

1. Open `docs/self-hosting/guides/upgrading-infisical.mdx` and review the new "Restart the Infisical Service" step (#5) and the expanded "UI Caching Issues After Upgrade" troubleshooting section.
2. The "Restart" step is scoped to Linux package installs only, with a note that Docker / Kubernetes users do not need it.
3. The troubleshooting section now distinguishes between browser cache and stale Node process, with a test (incognito window) to tell them apart.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: `docs(self-hosting): add service restart step and stale-process troubleshooting for Linux package upgrades`
- [x] Tested locally (mDX tags balanced: 3 opening, 3 closing; reviewed rendered preview)
- [x] Updated docs (this IS the docs change)

### DIFF
--- a/docs/self-hosting/guides/upgrading-infisical.mdx
+++ b/docs/self-hosting/guides/upgrading-infisical.mdx
@@ -53,7 +53,15 @@ Now, migrations run automatically during boot-up. This improvement streamlines t
    - In environments with multiple instances, one instance will acquire a lock and perform the migration while the other instances wait.
    - Once the migration is complete, all instances will operate with the updated schema.
 
-5. **Verify the Upgrade:**  
+5. **Restart the Infisical Service (Linux package installs only):**
+   - If you installed Infisical via the Linux package (`.deb` / `.rpm`), the upgrade replaces the on-disk binaries but does not automatically restart the long-running Node process started by `runsvdir`. The old process continues running with the previous binary (visible as `(deleted)` in `/proc/<pid>/exe`) while serving the new `index.html` that references updated asset filenames. This causes a blank UI on the next page load.
+   - After the package upgrade completes, restart the service so the new binary is loaded:
+     ```bash
+     sudo systemctl restart infisical-runsvdir
+     ```
+   - This step is not required for Docker, Kubernetes, or other container-based deployments — those restart containers automatically.
+
+6. **Verify the Upgrade:**  
    - Review the logs for any migration errors or warnings.
    - Test basic functionality to ensure the upgrade was successful.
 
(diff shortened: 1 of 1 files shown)


Manibharadwaj:
Hey @akhilmhdh 👋

Friendly follow-up on this docs PR — it adds a missing service restart step and stale-process troubleshooting for Linux package-based Infisical upgrades. Small but important fix — without the restart step, users hit caching issues after upgrading.

All security checks pass ✅. Would appreciate a review when you get a chance. Thanks!

PR #6771, opened 2026-06-08
feat: add kmip client registration with CSR

sheensantoscapadngan:
## Context
This PR adds support for KMIP client certificate generation with CSRs

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/routes/v1/kmip-router.ts
```diff
@@ -287,16 +287,21 @@ export const registerKmipRouter = async (server: FastifyZodProvider) => {
...
       }),
-      body: z.object({
-        keyAlgorithm: z.nativeEnum(CertKeyAlgorithm),
-        ttl: z.string().refine((val) => ms(val) > 0, "TTL must be a positive number")
-      }),
+      body: z
+        .object({
+          keyAlgorithm: z.nativeEnum(CertKeyAlgorithm).optional(),
+          ttl: z.string().refine((val) => ms(val) > 0, "TTL must be a positive number"),
+          csr: z.string().trim().min(1, "CSR cannot be empty").optional()
```
bernie-g: can we add a max() 4096 here

docs/documentation/platform/kms/kmip-dell-poweredge.mdx
```diff
@@ -0,0 +1,64 @@
...
+    4. Enter the KMIP server address and port (default: 5696)
+    5. Click **Next**
+  </Step>
+
+  <Step title="Generate Certificate Request on iDRAC">
+    When iDRAC prompts you to generate a certificate request, you'll need to enter subject values. You can find the required values in Infisical by clicking **Generate Certificate** on your KMIP client and selecting the **CSR** request method - the modal will display the exact **Client ID** and **Project ID** to use.
+    
+    1. Click **Generate CSR** on iDRAC
+    2. Enter the certificate information:
+       - **Common Name (CN)**: Enter the **Client ID** shown in Infisical
```
bernie-g, line 35: not a blocker if we need to get this out ASAP but it would be useful to have some screenshots here, this is a lot of instruction text

docs/documentation/platform/kms/kmip-dell-poweredge.mdx
```diff
@@ -0,0 +1,64 @@
...
+  <Step title="Upload Certificate to iDRAC">
+    1. In iDRAC, upload the signed client certificate
+    2. Upload the certificate chain as the KMS CA certificate
+    3. Click **Test Network Connection** to verify connectivity
+    4. Complete the SEKM configuration
+  </Step>
+
+  <Step title="Enable Encryption on Storage Controller">
+    Once SEKM is configured, you can enable encryption on your storage controller (PERC, HBA, or NVMe) through iDRAC.
+  </Step>
```
bernie-g, line 57: ignore comment placement

Image

nit: Can we change this modal title to "KMIP Client Certificate" since we're not really generating a cert in this case? Also maybe change the right click option to "Create Certificate" instead of "Generate Certificate" but that's kind of a nit

sheensantoscapadngan: oh we technically are... 

the result of passing in a CSR is a completely new certificate

docs/documentation/platform/kms/kmip-dell-poweredge.mdx
sheensantoscapadngan: I intentionally left out the screenshots here because the exact form factor is bound to change and this will get outdated pretty quickly. Would be best for us to stabilize the product first before then

docs/documentation/platform/kms/kmip-dell-poweredge.mdx
sheensantoscapadngan: but yes, let me update that. should be pretty quick

backend/src/ee/routes/v1/kmip-router.ts
sheensantoscapadngan: will do

docs/documentation/platform/kms/kmip-dell-poweredge.mdx
bernie-g: Hm ok I see, I guess I was confused by the button saying "Sign certificate" instead of "Generate cert". Makes sense to me.

bernie-g (approved):

PR #6775, opened 2026-06-09
improvement(secret-sync): allow gateway on azure key vault  

adilsitos:
## Context
Add gateway support for azure key vault. This allow the usage of vault running on private networks to be reached by Infisical using a gateway 

## Steps to verify the change
- create a private azure vault instance
- create a virtual machine that can reach the private azure vault instace 
- add a gateway and a relay on the vm that has access
- create an app connection that uses a gateway 
- check that a sync can happen using the gateway

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


frontend/src/components/v3/platform/GatewayPicker/GatewayPicker.tsx
```diff
@@ -83,7 +83,7 @@ export const GatewayPicker = ({
       <SelectTrigger className={className ?? "w-full"} isError={isError}>
         <SelectValue placeholder={placeholder ?? "Select gateway..."} />
       </SelectTrigger>
-      <SelectContent position="popper">
+      <SelectContent position="popper" className="z-[100]">
```
adilsitos: this was a bug introduced in a PR earlier today. If this PR tooks long to get merged, I will change this in another PR and send a patch

backend/src/services/app-connection/azure-key-vault/azure-key-vault-connection-fns.ts
```diff
@@ -24,6 +33,148 @@ import {
...
+    });
+  }
+
+  const gatewayId =
+    gatewayPoolId && gatewayPoolService
+      ? await gatewayPoolService.resolveEffectiveGatewayId({ gatewayId: directGatewayId, gatewayPoolId })
+      : directGatewayId;
+
+  const url = new URL(requestConfig.url as string);
+  await blockLocalAndPrivateIpAddresses(url.toString(), Boolean(gatewayId));
```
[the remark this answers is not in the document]
adilsitos: this is not required. If we do the validation at the end, it would be the same, since they both came from the same place

backend/src/services/app-connection/azure-key-vault/azure-key-vault-connection-fns.ts
```diff
@@ -24,6 +33,148 @@ import {
...
+
+      const finalRequestConfig: AxiosRequestConfig = {
+        ...requestConfig,
+        url: url.toString(),
+        httpsAgent,
+        headers: {
+          ...requestConfig.headers,
+          Host: targetHost
+        }
+      };
```
mathnogueira, line 151: question: there are some implementation differences here. Just want to confirm is they are intended:

* We enforce `https` instead of checking if it's `https`
* We don't add the `https.Agent` anymore in the `AxiosRequestConfig`

mathnogueira (dismissed):

mathnogueira (approved):

PR #6784, opened 2026-06-09
improvement(router): change route creation to avoid memory stack exceeded error

adilsitos:
## Context
The secret sync router being created was consuming all the node stack memory because it was not releasing the context right away, it was adding all the possible paths. The solution was to return the `onReady` in an async way.  

## Steps to verify the change
- Add new secret syncs (you can just add the router; do this by just copying the other ones that already exist) 
- Add around 10 of them 
- Check that no stack memory error happens  

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


adilsitos:
@greptile check the changes again, they are different from the first commit. Also, update the summary

carlosmonastyrski (approved):

PR #6803, opened 2026-06-10
fix: use gateway in case of private GHE server

Thiago-AS:
## Context

- Allow GitHub App manifest creation for private hosted GHE server.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/github-app/github-app-service.ts
+++ b/backend/src/services/github-app/github-app-service.ts
@@ -131,7 +131,11 @@ export const gitHubAppServiceFactory = ({
   // Mirrors the gateway checks in resolveUserInstallations so the manifest exchange can route
   // through a gateway when the GitHub host (e.g. a private GitHub Enterprise Server) isn't
   // directly reachable from the Infisical backend.
-  const assertGatewayUsable = async (gatewayId: string, orgPermission: OrgServiceActor) => {
+  const assertGatewayUsable = async (
+    gatewayId: string,
+    orgPermission: OrgServiceActor,
+    prefetchedOrgPermission?: Awaited<ReturnType<typeof permissionService.getOrgPermission>>["permission"]
+  ) => {
     const plan = await licenseService.getPlan(orgPermission.orgId);
     if (!plan.gateway) {
       throw new BadRequestError({
@@ -140,14 +144,18 @@ export const gitHubAppServiceFactory = ({
       });
     }
 
-    const { permission } = await permissionService.getOrgPermission({
-      actor: orgPermission.type,
-      actorId: orgPermission.id,
-      orgId: orgPermission.orgId,
-      actorAuthMethod: orgPermission.authMethod,
-      actorOrgId: orgPermission.orgId,
-      scope: OrganizationActionScope.Any
-    });
+    const permission =
+      prefetchedOrgPermission ??
+      (
+        await permissionService.getOrgPermission({
+          actor: orgPermission.type,
+          actorId: orgPermission.id,
+          orgId: orgPermission.orgId,
+          actorAuthMethod: orgPermission.authMethod,
+          actorOrgId: orgPermission.orgId,
+          scope: OrganizationActionScope.Any
+        })
+      ).permission;
 
     ForbiddenError.from(permission).throwUnlessCan(
       OrgPermissionGatewayActions.AttachGateways,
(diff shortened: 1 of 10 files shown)


backend/src/services/github-app/github-app-service.ts
```diff
@@ -503,18 +555,41 @@ export const gitHubAppServiceFactory = ({
...
+            method: "POST",
+            data: {},
+            headers: {
+              Accept: "application/vnd.github+json",
+              "X-GitHub-Api-Version": "2022-11-28"
             }
-          );
+          };
+
+          const { data } = gatewayId
```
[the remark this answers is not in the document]
Thiago-AS: Not an issue for now; the callback is almost always instantaneous since it's automated.

backend/src/services/github-app/github-app-service.ts
```diff
@@ -126,6 +127,50 @@ export const gitHubAppServiceFactory = ({
...
+    );
+
+    const [gateway] = await gatewayDAL.find({ id: gatewayId, orgId: orgPermission.orgId });
+    const [gatewayV2] = await gatewayV2DAL.find({ id: gatewayId, orgId: orgPermission.orgId });
+    if (!gateway && !gatewayV2) {
+      throw new NotFoundError({
+        message: `Gateway with ID ${gatewayId} not found for org`
+      });
+    }
+  };
```
victorvhs017: We should add gateway pool support, as we have in the app-connection-service.ts.

backend/src/services/github-app/github-app-service.ts
```diff
@@ -126,6 +127,50 @@ export const gitHubAppServiceFactory = ({
...
   };
 
+  // Validates that the actor may attach the given gateway and that it exists for the org.
+  // Mirrors the gateway checks in resolveUserInstallations so the manifest exchange can route
+  // through a gateway when the GitHub host (e.g. a private GitHub Enterprise Server) isn't
+  // directly reachable from the Infisical backend.
+  const assertGatewayUsable = async (
+    gatewayId: string,
+    orgPermission: OrgServiceActor,
+    prefetchedOrgPermission?: Awaited<ReturnType<typeof permissionService.getOrgPermission>>["permission"]
```
victorvhs017: we can drop this because the permission fetch is memoized.

backend/src/services/github-app/github-app-service.ts
```diff
@@ -503,18 +555,41 @@ export const gitHubAppServiceFactory = ({
...
+
+          const { data } = gatewayId
+            ? await requestWithGitHubGateway<TGitHubAppManifestResponse>(
+                { gatewayId },
+                gatewayService,
+                gatewayV2Service,
+                requestConfig,
+                await getGitHubGatewayConnectionDetails(gatewayId, apiBaseUrl, gatewayV2Service)
+              )
+            : await safeRequest.request<TGitHubAppManifestResponse>(requestConfig);
```
victorvhs017, line 602: We had issues with safeRequest before and had to roll back. The code was updated, but it's only being used in webhooks for now.

We need to really test this through; we should do it against hosts using only ipv4, only ipv6, and both. (this is where it broke the last time)

backend/src/services/github-app/github-app-service.ts
Thiago-AS: We didn't support before, that's why I did not add it here, but can do.

backend/src/services/github-app/github-app-service.ts
Thiago-AS: Tested already, it was working just fine, but since they are using private hosted git hub, it wasn't reaching their instance.

victorvhs017: perfect!

victorvhs017: just keep an eye if another issue shows up, because it could be one of the factors

victorvhs017 (changes requested):
One thing that is not really clear to me is that this screen is using the gateway from the initial form on the GitHub callback.

Image

And it may fail if your gateway is not healthy. For example, I got this error when I tried to create an app with a gateway (both an invalid gateway and the default gateway).

Image

I think we should at least show in this modal a message like:

"Using X gateway"

Image

victorvhs017 (approved):

PR #6814, opened 2026-06-11
improvement(vault): add fallback to list only authorized paths on migration

adilsitos:
## Context
Some users don't want to have a wide open policy on vault, and the current approach always need the list on the root of the folder, otherwise it will cause an unauthorized error.   

## Screenshots

image

## Steps to verify the change
- Create an instance on vault cloud (ask me if you need help)
- define an app role that only has access on a certain amount of folders and secrets 
- create an app connection on infisical 
- try to perform a migration and verify that only the paths you have access are returned. 

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


adilsitos:
@greptile review again and update the summary. For context, the paths that have a + wildcard will be discarded. The `*` wildcard don't show up on the response, so we don't need to check those.

mathnogueira (dismissed):
codewise this looks good

mathnogueira (approved):

PR #6818, opened 2026-06-11
feat(webhook): add get-by-id route and allow machine identity auth

PrestigePvP:
## Context

- Add `GET /api/v1/webhooks/:webhookId`, returning the webhook with its decrypted URL.
- Accept `IDENTITY_ACCESS_TOKEN` on all webhook routes (previously JWT-only) so machine identities can manage webhooks.
- Unblocks the `infisical_webhook` Terraform provider resource, whose Read/Import call this route. The provider release must lag this deploy, since without this route every provider read 404s, not just imports.
- ENG-5124

## Steps to verify the change

- `npm run type:check` clean; `eslint` clean.
- Not yet exercised at runtime: `GET /webhooks/:id` should enforce project `Webhooks` read permission, return the decrypted `url`, and 404 on unknown id.

## Type

- [x] Feature

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/webhook-router.ts
+++ b/backend/src/server/routes/v1/webhook-router.ts
@@ -257,7 +257,7 @@ export const registerWebhookRouter = async (server: FastifyZodProvider) => {
     config: {
       rateLimit: writeLimit
     },
-    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    onRequest: verifyAuth([AuthMode.JWT]),
     schema: {
       operationId: "testWebhook",
       params: z.object({
--- a/backend/src/server/routes/v1/webhook-router.ts
+++ b/backend/src/server/routes/v1/webhook-router.ts
@@ -45,7 +45,7 @@ export const registerWebhookRouter = async (server: FastifyZodProvider) => {
     config: {
       rateLimit: writeLimit
     },
-    onRequest: verifyAuth([AuthMode.JWT]),
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
     schema: {
       operationId: "createWebhook",
       body: z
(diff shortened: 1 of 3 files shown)


backend/src/services/webhook/webhook-service.ts
```diff
@@ -251,10 +252,36 @@ export const webhookServiceFactory = ({
     });
   };
 
+  const getWebhookById = async ({ id, actor, actorId, actorAuthMethod, actorOrgId }: TGetWebhookByIdDTO) => {
+    const webhook = await webhookDAL.findById(id);
+    if (!webhook) throw new NotFoundError({ message: `Webhook with ID '${id}' not found` });
```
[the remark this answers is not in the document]
PrestigePvP: Fixed in de6eef3e0b. Unauthorized callers (non-member, cross-org, or missing Webhooks read permission) now get the same 404 as a missing webhook, so the endpoint no longer reveals existence. The decrypt/return path is only reached after the read check passes. Left the pre-existing update/delete/test routes unchanged as out of scope for this PR.

backend/src/server/routes/v1/webhook-router.ts
```diff
@@ -257,7 +257,7 @@ export const registerWebhookRouter = async (server: FastifyZodProvider) => {
     config: {
       rateLimit: writeLimit
     },
-    onRequest: verifyAuth([AuthMode.JWT]),
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
```
akhilmhdh: I don't think the webhook test needs to be public. It's something just for the UI

akhilmhdh (changes requested):

backend/src/server/routes/v1/webhook-router.ts
PrestigePvP: Good call, reverted in 9f96d09802. The test route is back to JWT-only. Machine-identity auth now stays only on the routes the Terraform provider actually uses (create/update/delete/get-by-id), plus list as a general read. Happy to lock list down too if you'd prefer, but left it open since it's a pure read rather than a UI action.

akhilmhdh (approved):

PR #6823, opened 2026-06-11
feat(folders): add move operation

adilsitos:
## Context
Allow the move operation of secrets. At the moment, only folders that contain only static secrets can be moved. 

## Screenshots

image

## Steps to verify the change

There are several scenarios in this change. 

### Scenario 1: Folder has secrets that cannot be moved
- Try to move a folder that has secrets that can't be moved 
- Check that the UI does not allow to move it 
- if you try to call the move endpoint directly, this will also fail 

### Scenario 2: Move folder between environments 
- try to move a folder 
- it should move a folder to another environment 
- all secrets under the folder were moved 

### Scenario 3: Approval policy exists on the source folder/environment
- create a policy to a sub folder
- try to move a folder (from the `root`)
- The move button is not allowed

### Scenario 4: Approval policy exists on the destination folder/environment 
- create a policy to the destination 
- try to move a folder (from the `root`)
- check that the secret was deleted on the source and an approval was created on the destination
- if the approval is approved, the folder is created on the destination

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/secret-folder/secret-folder-service.ts
```diff
@@ -1448,13 +1508,390 @@ export const secretFolderServiceFactory = ({
...
+        throw new BadRequestError({
+          message: `Folder '${folderName}' is already located at path '${destinationPath}'`
+        });
+      }
+    }
+
+    // 4. perform the whole move inside ONE transaction so it is atomic and the source-folder row lock is held for
+    // the entire operation: enumerate + lock the source subtree, recreate the tree at the destination, move every
+    // folder's secrets (sharing this tx), then delete the emptied source root. a failure anywhere rolls all of
+    // it back. side effects (snapshots/syncs/cache) are dispatched after the commit.
```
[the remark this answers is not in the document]
adilsitos: Right now when we move secrets, those are also not moved. This can be discussed, but in this case this would also change how the move secrets happen. cc: @akhilmhdh

backend/src/server/routes/v2/secret-folder-router.ts
```diff
@@ -441,4 +441,72 @@ export const registerSecretFolderRouter = async (server: FastifyZodProvider) =>
...
+          folderId: z.string(),
+          sourceEnvironment: z.string(),
+          sourcePath: z.string(),
+          destinationEnvironment: z.string(),
+          destinationPath: z.string(),
+          hasApprovalRequests: z.boolean()
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
```
akhilmhdh: Hmm! May be let's not expose to identity yet and keep it UI only. Like the other ones

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -2493,6 +2494,18 @@ interface DeleteFolderEvent {
...
 
+interface MoveFolderEvent {
+  type: EventType.MOVE_FOLDER;
+  metadata: {
+    folderId: string;
+    sourceEnvironment: string;
+    sourcePath: string;
+    destinationEnvironment: string;
+    destinationPath: string;
+    hasApprovalRequests: boolean;
```
akhilmhdh: I thought we don't allow secret. move for ones with approval request configured

akhilmhdh (commented):
Pending application testing

backend/src/ee/services/audit-log/audit-log-types.ts
adilsitos: forgot to remove this from audit logs.

akhilmhdh (dismissed):

akhilmhdh (approved):

backend/src/services/secret-folder/secret-folder-service.ts
```diff
@@ -1448,13 +1521,527 @@ export const secretFolderServiceFactory = ({
...
+      if (existingDestinationFolder) {
+        throw new BadRequestError({
+          message: `A folder named '${folderName}' already exists at path '${destinationPath}' in environment '${destinationEnvironment}'`
+        });
+      }
+
+      const destinationFolderRoot = finalDestinationPath;
+      const destinationEnvId = destinationParentFolder.envId;
+      // locking the source folder rows blocks concurrent secret inserts into them (secret -> folder FK) until commit.
+      const subtree = await folderDAL.findByEnvsDeep({ parentIds: [sourceFolder.id] }, tx);
```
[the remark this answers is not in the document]
adilsitos: nit

PR #6827, opened 2026-06-11
feat(platfor-415): usage metering pipeline for License Server v2

PrestigePvP:
## Context

License Server v2 introduces per-entity pricing, so the product needs to **observe** the entity counts that drive billing/caps and **report** them to the License Server. This adds the product-side usage-metering pipeline for the 4 committed meters.

It ships in **Phase 1, before reads cut over**, so it is **structural-but-inert**: gated behind `LICENSE_SERVER_V2_ENABLED` (default off). Deploying with the flag off is a product no-op.

- New `services/license-client/usage/` module: org-keyed snapshot reporter (`POST /v1/usage/snapshots`), per-meter count queries on the read replica (sum across the org's live projects), a BullMQ worker (`QueueName.UsageEvent`, 60s debounce via deterministic jobId, report-on-change), and a self-hosted 30-min flush cron (`CronJobName.UsageFlush`, gated to non-cloud).
- Fire-and-forget emit hooks on metered create/delete in the 4 metered services: identity, internal CA, active certificate, PAM resource. Emits never block or throw into the request path and sit outside DB transactions.
- Registers the live-count fns with the SDK (`registerCounter`) so `canUse()` can resolve current usage. No call site invokes it yet.
- Inert when off: emits no-op, the worker no-ops without a reporter, the cron is self-hosted-only.

Out of scope (tracked separately): entitlement dual-read/compare + read cutover and `canUse()` cap enforcement at call sites — PLATFOR-449 / PLATFOR-418. Secret-scanning's meter is intentionally omitted until its pricing unit is settled.

Ticket: PLATFOR-415

## Screenshots

No UI changes.

## Steps to verify the change

- `cd backend && npm run test:unit -- src/services/license-client/usage/usage-metering.test.ts` → 16 tests pass (counters registry, emit gating + debounce, reporter build, worker report-on-change, and `canUse` usage examples).
- `make reviewable-api` → lint + type-check clean.
- With `LICENSE_SERVER_V2_ENABLED=false` (default): create/delete an identity, internal CA, certificate, or PAM resource → nothing is enqueued and create/delete behavior is unchanged.
- With it enabled + a snapshot endpoint configured: create an identity → exactly one org-keyed snapshot is POSTed after the 60s debounce window; an unchanged count does not re-report.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the conventional commit format: `type(scope): short description`
- [x] Tested locally

### DIFF
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -168,6 +168,7 @@ export const KeyStoreTtls = {
   AuditLogMigrationAlertInSeconds: 604800, // 7 days
   LicenseCloudPlanInSeconds: 300, // 5 minutes
   LicenseEntitlementsInSeconds: 1800, // 30 minutes
+  UsageLastReportedInSeconds: 7776000, // 90 days (~3 billing cycles) so orphaned meter keys self-clean
   AiMcpEndpointOAuthFlowInSeconds: 300, // 5 minutes
   OauthAuthorizationCodeInSeconds: 600, // 10 minutes
   AiMcpServerOAuthSessionInSeconds: 600, // 10 minutes
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -132,7 +132,8 @@ export const KeyStorePrefixes = {
   AuditLogMigrationAlert: "audit-log-migration-alert-last-row-count",
   LicenseCloudPlan: (orgId: string) => `infisical-cloud-plan-${orgId}` as const,
   LicenseEntitlements: (orgId: string) => `license-entitlements-${orgId}` as const,
-  UsageLastReported: (orgId: string, featureKey: string) => `usage-last-reported-${orgId}-${featureKey}` as const,
+  LicenseUsageLastReported: (orgId: string, featureKey: string) =>
+    `license-usage-last-reported-${orgId}-${featureKey}` as const,
   IdentityLockoutState: (identityId: string, authMethod: string, slug: string) =>
     `lockout:identity:${identityId}:${authMethod}:${slug}` as const,
   IdentityLockoutStateByMethodPattern: (identityId: string, authMethod: string) =>
@@ -168,7 +169,7 @@ export const KeyStoreTtls = {
   AuditLogMigrationAlertInSeconds: 604800, // 7 days
   LicenseCloudPlanInSeconds: 300, // 5 minutes
   LicenseEntitlementsInSeconds: 1800, // 30 minutes
-  UsageLastReportedInSeconds: 7776000, // 90 days (~3 billing cycles) so orphaned meter keys self-clean
+  LicenseUsageLastReportedInSeconds: 7776000, // 90 days (~3 billing cycles) so orphaned meter keys self-clean
   AiMcpEndpointOAuthFlowInSeconds: 300, // 5 minutes
   OauthAuthorizationCodeInSeconds: 600, // 10 minutes
   AiMcpServerOAuthSessionInSeconds: 600, // 10 minutes
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -132,6 +132,7 @@ export const KeyStorePrefixes = {
   AuditLogMigrationAlert: "audit-log-migration-alert-last-row-count",
   LicenseCloudPlan: (orgId: string) => `infisical-cloud-plan-${orgId}` as const,
   LicenseEntitlements: (orgId: string) => `license-entitlements-${orgId}` as const,
+  UsageLastReported: (orgId: string, featureKey: string) => `usage-last-reported-${orgId}-${featureKey}` as const,
   IdentityLockoutState: (identityId: string, authMethod: string, slug: string) =>
     `lockout:identity:${identityId}:${authMethod}:${slug}` as const,
   IdentityLockoutStateByMethodPattern: (identityId: string, authMethod: string) =>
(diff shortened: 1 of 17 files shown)


backend/src/services/license-client/usage/usage-reporter.ts
```diff
@@ -0,0 +1,57 @@
...
+      return;
+    }
+
+    const url = new URL(SNAPSHOTS_PATH, serverUrl);
+    const res = await fetch(url, {
+      method: "POST",
+      headers: { Authorization: `Bearer ${serviceKey}`, "Content-Type": "application/json" },
+      body: JSON.stringify({ snapshots }),
+      redirect: "manual"
+    });
```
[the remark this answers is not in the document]
PrestigePvP: Addressed in 1f0dc44 — `buildUsageReporter` now rejects a malformed or non-`https` `LICENSE_SERVER_V2_URL` (returns `null`, so reporting stays inert) before the service-key bearer can be sent anywhere. Note the already-merged read SDK (`license-client-backends.ts`) consumes the same env var, so a stricter hostname allowlist would be shared hardening across both; since this is operator-controlled config rather than user input, I kept it to scheme + parse validation here.

backend/src/services/license-client/usage/usage-event-queue.ts
```diff
@@ -0,0 +1,109 @@
...
+        org_id: orgId,
+        feature_key: featureKey,
+        value,
+        observed_at: observedAt.toISOString(),
+        idempotency_key: `${source}:${orgId}:${featureKey}:${minuteBucketSeconds}`,
+        source
+      }
+    ]);
+
+    await keyStore.setItem(lastReportedKey, String(value));
```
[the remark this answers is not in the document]
PrestigePvP: Addressed in 1f0dc44 — last-reported keys now use `setItemWithExpiry` with a 90-day TTL (`UsageLastReportedInSeconds`, ~3 billing cycles) so orphaned `{orgId}:{featureKey}` entries self-clean.

backend/src/services/license-client/usage/usage-event-queue.ts
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  // Self-hosted true-up: online instances also report event-driven, so this is a periodic backstop
+  // for cap visibility and renewal. Cloud is event-driven only.
+  const flushAllOrgs = async () => {
+    if (!usageReporter) {
+      return;
+    }
+
+    const orgs = await orgDAL.find({});
```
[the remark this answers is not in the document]
PrestigePvP: Addressed in 1f0dc44 — `flushAllOrgs` now pages through orgs in batches of 500 (`limit`/`offset`) instead of materializing the whole table, with tests covering the pagination loop and the null-reporter no-op.

backend/src/services/license-client/usage/usage-counters.ts
```diff
@@ -0,0 +1,24 @@
...
+  usageCounterDAL: TUsageCounterDALFactory;
+};
+
+// The single source of truth pairing each metered feature with its live-count fn. Reused by the
+// worker (count + report) and by registerCounter (so canUse() resolves current usage).
+export const buildMeteredFeatures = ({ licenseDAL, usageCounterDAL }: TBuildMeteredFeaturesDep): TMeteredFeature[] => [
+  { feature: MaxIdentities, count: (orgId) => licenseDAL.countOrgUsersAndIdentities(orgId) },
+  { feature: MaxInternalCas, count: (orgId) => usageCounterDAL.countInternalCas(orgId) },
+  { feature: MaxActiveCerts, count: (orgId) => usageCounterDAL.countActiveCerts(orgId) },
+  { feature: MaxResources, count: (orgId) => usageCounterDAL.countPamResources(orgId) }
```
akhilmhdh: I think it;'s better to call it MaxPamResources

backend/src/keystore/keystore.ts
```diff
@@ -167,6 +168,7 @@ export const KeyStoreTtls = {
   AuditLogMigrationAlertInSeconds: 604800, // 7 days
   LicenseCloudPlanInSeconds: 300, // 5 minutes
   LicenseEntitlementsInSeconds: 1800, // 30 minutes
+  UsageLastReportedInSeconds: 7776000, // 90 days (~3 billing cycles) so orphaned meter keys self-clean
```
akhilmhdh: Better make it more aligned like `LicenseUsage` something like that?

backend/src/lib/cron/cron-job.ts
```diff
@@ -37,7 +37,8 @@ export const CronJobName = {
   SignerDailyAutoRenewal: "signer-daily-auto-renewal",
   SignerIssuancePolling: "signer-issuance-polling",
   AuditLogStreamOutboxStaleClaimSweeper: "audit-log-stream-outbox-stale-claim-sweeper",
-  AuditLogStreamOutboxCleanup: "audit-log-stream-outbox-cleanup"
+  AuditLogStreamOutboxCleanup: "audit-log-stream-outbox-cleanup",
+  UsageFlush: "usage-flush"
```
akhilmhdh: Here as well.

akhilmhdh (changes requested):

backend/src/services/license-client/usage/usage-counters.ts
PrestigePvP: Renamed to `MaxPamResources` in 505eb98. I kept the underlying feature key as `max_resources` (the constant's string) to match the seeded feature registry / Usage Reporting SPEC §3.1 — if you'd rather the key be `max_pam_resources` too, say the word and I'll change it here and on the server registry together.

backend/src/keystore/keystore.ts
PrestigePvP: Done in 505eb98 — renamed to `LicenseUsageLastReported` / `LicenseUsageLastReportedInSeconds` (value `license-usage-last-reported-…`) to align with `LicenseCloudPlan` / `LicenseEntitlements`.

backend/src/lib/cron/cron-job.ts
PrestigePvP: Done in 505eb98 — `CronJobName.UsageFlush` → `LicenseUsageFlush` (`license-usage-flush`), same alignment.

akhilmhdh (approved):

PR #6829, opened 2026-06-12
fix: netlify bug where different contexts would be deleted

mathnogueira:
## Context

When you have a variable with multiple values, a sync would cause the other environment values to be deleted. This fixes it and simplifies the code on how to run the sync.

## Steps to verify the change

1. Create a netlify connection
2. create two syncs for different environments (dev and prod for example)
3. Create secrets in the project, make sure to create variables with same name (ENVIRONMENT: prod/dev values for example)
4. Run the syncs and make sure all variables were added and the values are preserved

1. Create a secret in netlify with same name as a variable you have in Infisical
2. Sync it
3. See that the other environment values are deleted (this is intended because we can't get the existing value of a secret, it's resolved as `****************`

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
+++ b/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
@@ -94,23 +94,30 @@ class NetlifyPublicClient {
     }
   }
 
-  async getVariables(
-    connection: TNetlifyConnectionConfig,
-    { account_id, ...params }: NetlifyParams,
-    limit: number = 50,
-    page: number = 1
-  ) {
-    const res = await this.send<TNetlifyVariable[]>(connection, {
-      method: "GET",
-      url: `/accounts/${account_id}/env`,
-      params: {
-        ...params,
-        limit,
-        page
-      }
-    });
+  async getVariables(connection: TNetlifyConnectionConfig, { account_id, ...params }: NetlifyParams) {
+    const PAGE_SIZE = 100;
+    const allVariables: TNetlifyVariable[] = [];
+    let page = 1;
 
-    return res;
+    // eslint-disable-next-line no-constant-condition, @typescript-eslint/no-unnecessary-condition
+    while (true) {
+      const res = await this.send<TNetlifyVariable[]>(connection, {
+        method: "GET",
+        url: `/accounts/${account_id}/env`,
+        params: {
+          ...params,
+          limit: PAGE_SIZE,
+          page
+        }
+      });
+
+      allVariables.push(...res);
+
+      if (res.length < PAGE_SIZE) break;
+      page += 1;
+    }
+
+    return allVariables;
   }
 
   async createVariable(
--- a/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
+++ b/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
@@ -102,7 +102,7 @@ class NetlifyPublicClient {
   }
 
   async getVariables(connection: TNetlifyConnectionConfig, { account_id, ...params }: NetlifyParams) {
-    // This public endpoint doesn't support validation, but it also returns all variables.
+    // This public endpoint doesn't support pagination, but it also returns all variables.
     // Tested with 900 secrets.
     return this.send<TNetlifyVariable[]>(connection, {
       method: "GET",
--- a/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
+++ b/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
@@ -202,13 +202,13 @@ class NetlifyPublicClient {
 
   async deleteVariableValue(
     connection: TNetlifyConnectionConfig,
-    { account_id, value_id, ...params }: NetlifyParams & { value_id: string },
+    { account_id, ...params }: NetlifyParams,
     variable: Pick<TNetlifyVariable, "key" | "id">
   ) {
     try {
       const res = await this.send<TNetlifyVariable>(connection, {
         method: "DELETE",
-        url: `/accounts/${account_id}/${variable.key}/value/${value_id}`,
+        url: `/accounts/${account_id}/env/${variable.key}/value/${variable.id}`,
         params
       });
 
--- a/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
+++ b/backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
@@ -5,10 +5,10 @@ import { AxiosInstance, AxiosRequestConfig, AxiosResponse, HttpStatusCode, isAxi
 
 import { createRequestClient } from "@app/lib/config/request";
 import { IntegrationUrls } from "@app/services/integration-auth/integration-list";
+import { NetlifySyncContext } from "@app/services/secret-sync/netlify";
 
 import { NetlifyConnectionMethod } from "./netlify-connection-constants";
 import { TNetlifyAccount, TNetlifyConnectionConfig, TNetlifySite, TNetlifyVariable } from "./netlify-connection-types";
-import { NetlifySyncContext } from "@app/services/secret-sync/netlify";
 
 export function getNetlifyAuthHeaders(connection: TNetlifyConnectionConfig): Record<string, string> {
   switch (connection.method) {
@@ -104,7 +104,7 @@ class NetlifyPublicClient {
   async getVariables(connection: TNetlifyConnectionConfig, { account_id, ...params }: NetlifyParams) {
     // This public endpoint doesn't support validation, but it also returns all variables.
     // Tested with 900 secrets.
-    return await this.send<TNetlifyVariable[]>(connection, {
+    return this.send<TNetlifyVariable[]>(connection, {
       method: "GET",
       url: `/accounts/${account_id}/env`,
       params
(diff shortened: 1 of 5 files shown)


backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
```diff
@@ -41,78 +41,124 @@ export const NetlifySyncFns = {
...
       }
     }
 
-    if (disableSecretDeletion) return;
+    for await (const key of variablesToUpdate) {
+      try {
+        const existingVar = existingInNetlify[key];
+
+        if (existingVar.is_secret) {
+          await NetlifyPublicAPI.deleteVariable(secretSync.connection, params, { key });
```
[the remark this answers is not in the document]
mathnogueira: Well, this behavior was not changed. We might need to discuss if this is something we want to prevent. The only think that is not clear is if we want to allow users to use `secrets` in Netlify or just variables.

backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
```diff
@@ -94,23 +94,30 @@ class NetlifyPublicClient {
...
+          ...params,
+          limit: PAGE_SIZE,
+          page
+        }
+      });
 
-    return res;
+      allVariables.push(...res);
+
+      if (res.length < PAGE_SIZE) break;
```
adilsitos: maybe a good validation to have on this loop is to check if the response status is a 429, and in this case, perform a sleep. we can even have a maximum amount of retries in case several 429 happens.

backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
```diff
@@ -41,78 +41,124 @@ export const NetlifySyncFns = {
...
 
-    const existing = Object.fromEntries(variables.map((variable) => [variable.key, variable]));
+    const params = {
+      ...baseParams,
+      context_name: config.context,
+    };
 
-    for await (const key of Object.keys(secretMap)) {
+    const variables = await NetlifyPublicAPI.getVariables(secretSync.connection, baseParams);
+    const existingInNetlify = Object.fromEntries(variables.map((variable) => [variable.key, variable]));
```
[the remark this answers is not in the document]
adilsitos: if you check how `getVariables` work you will see there is a pagination implementation there.

backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
```diff
@@ -41,78 +41,126 @@ export const NetlifySyncFns = {
...
+          await NetlifyPublicAPI.updateVariable(secretSync.connection, params, {
+            key,
+            is_secret: false,
+            values: mergedValues
+          });
+        }
+      } catch (error) {
+        throw new SecretSyncError({ error, secretKey: key });
+      }
+    }
```
adilsitos: this new approach is fine for me, but the upsert operation is not being used anymore. Maybe we can remove that.

backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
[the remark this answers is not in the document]
adilsitos: gotcha. I would send a message on the secret channel to validate which approach we should follow. 

as a reference, in other syncs that we cannot change the secrets (github and travis ci), we do not let the user choose if they want to import and delete, we force them to handle everything inside infisical. In this case, I think this would also make sense, but maybe it is good to check with @varonix0

backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
```diff
@@ -41,78 +41,126 @@ export const NetlifySyncFns = {
...
+          const variable = existingSecrets[secret];
+          const remainingValues = variable.values.filter((v) => v.context !== config.context);
+
+          // We skip updating variables in case of `is_secret`, otherwise we override the
+          // correct value with a masked one.
+          if (remainingValues.length > 0 && !variable.is_secret) {
+            await NetlifyPublicAPI.updateVariable(secretSync.connection, params, {
+              key: secret,
+              values: remainingValues
+            });
```
adilsitos: why are we updating the secrets on the delete operation? I don't think I understood the idea behind this.

backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
```diff
@@ -94,23 +94,30 @@ class NetlifyPublicClient {
...
-      method: "GET",
-      url: `/accounts/${account_id}/env`,
-      params: {
-        ...params,
-        limit,
-        page
-      }
-    });
+  async getVariables(connection: TNetlifyConnectionConfig, { account_id, ...params }: NetlifyParams) {
+    const PAGE_SIZE = 100;
```
adilsitos: this pagination approach is not working properly, it is causing an infinite loop 

https://github.com/user-attachments/assets/b66fe7bd-7534-4767-8919-f982bb080ba4

adilsitos (changes requested):

backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
mathnogueira: because netlify's way to keep multiple environments (contexts is the name they use) is to keep an array of values. 

So if you are deleting prod's value, I still need to update the array in case there's remaining values (staging, dev, etc). If prod is the only one and you delete it, this falls into the `else` block where we delete the variable.

adilsitos (changes requested):
One more thing that I forgot to mention. If we let the user choose all contexts to make the syncs, we need to make this more clear on the UI (and also on the docs) that this will erase what happen with other secret syncs. Maybe something similar to what we have [here](https://infisical.com/docs/integrations/secret-syncs/vercel#vercel-sync)

```
Team-scoped sync — initial import scope matching
On the initial import, Infisical only imports Vercel shared environment variables whose scope exactly matches this sync’s configuration (same target environments, target projects, and Apply to all Custom Environments flag). Variables covering a broader set of environments — or split across separate records per environment — will not be imported. The next sync will detach those records and create a new one for the sync’s scope using the value from Infisical.
Example: Vercel has a single shared variable FOO scoped to Production and Preview, but the sync targets only Production. FOO is not imported. On the next sync, the existing FOO is rescoped to Preview (value preserved) and a new FOO record is created for Production with the value from Infisical.
```

it is not the same problem, but worth letting the user know about it.

backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
```diff
@@ -41,78 +42,110 @@ export const NetlifySyncFns = {
...
+
+        // We skip updating variables in case of `is_secret`, otherwise we override the
+        // correct value with a masked one.
+        if (remainingValues.length > 0 && !variable.is_secret) {
+          await NetlifyPublicAPI.updateVariable(secretSync.connection, params, {
+            key,
+            values: remainingValues
           });
+        } else {
+          await NetlifyPublicAPI.deleteVariable(secretSync.connection, params, { key });
```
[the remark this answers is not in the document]
mathnogueira: we don't check for is_secret anymore.

backend/src/services/secret-sync/netlify/netlify-sync-fns.ts
```diff
@@ -41,78 +42,113 @@ export const NetlifySyncFns = {
...
+    const variablesToCreate = Object.keys(secretMap).filter((key) => !existingInNetlify[key]);
+    const variablesToUpdate = Object.keys(secretMap).filter((key) => existingInNetlify[key]);
+    const variablesToDelete = disableSecretDeletion
+      ? []
+      : // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+        Object.keys(existingInNetlify).filter(
+          (key) =>
+            !secretMap[key] &&
+            matchesSchema(key, environment?.slug || "", keySchema) &&
+            existingInNetlify[key].values.some((v) => v.context === config.context)
```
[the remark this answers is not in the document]
mathnogueira: removed fallback to `all` and send `undefined` instead in update.

docs/integrations/secret-syncs/netlify.mdx
```diff
@@ -39,6 +39,9 @@ description: "Learn how to configure a Netlify Sync for Infisical."
         - **Account**: The Netlify Account to be used.
         - **Site**: The Netlify site where secrets should be synced.
         - **Context**: The Netlify deployment context where secrets should be created in.
+        <Info>
```
adilsitos, line 42: I think it is worth adding a summarized version of this on the frontend as well in case the context is not selected. If you want to add a more descriptive one, we can have something like the second image.

Image

Image

backend/src/services/app-connection/netlify/netlify-connection-public-client.ts
```diff
@@ -94,23 +101,14 @@ class NetlifyPublicClient {
...
 
-  async getVariables(
-    connection: TNetlifyConnectionConfig,
-    { account_id, ...params }: NetlifyParams,
-    limit: number = 50,
-    page: number = 1
-  ) {
-    const res = await this.send<TNetlifyVariable[]>(connection, {
+  async getVariables(connection: TNetlifyConnectionConfig, { account_id, ...params }: NetlifyParams) {
+    // This public endpoint doesn't support validation, but it also returns all variables.
```
adilsitos: I think it is `pagination`, right?

adilsitos (commented):
it is working! Just two small final comments

docs/integrations/secret-syncs/netlify.mdx
mathnogueira: Image

adilsitos (approved):

PR #6833, opened 2026-06-12
feat: add support for revoking certificates using thumbprints

carlosmonastyrski:
## Context

Adds a new `POST /api/v1/cert-manager/certificates/revoke-by-thumbprint` endpoint to revoke a certificate by its thumbprint, alongside the existing revoke-by-ID flow.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/certificate-router.ts
+++ b/backend/src/server/routes/v1/certificate-router.ts
@@ -7,6 +7,7 @@ import { EventType } from "@app/ee/services/audit-log/audit-log-types";
 import { ApiDocsTags, CERTIFICATES } from "@app/lib/api-docs";
 import { NotFoundError } from "@app/lib/errors";
 import { ms } from "@app/lib/ms";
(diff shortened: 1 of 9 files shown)


backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -1688,6 +1688,70 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
     }
   });
 
+  server.route({
+    method: "POST",
+    url: "/revoke-by-thumbprint",
```
bernie-g: Are we planning on adding any more ways to revoke such as by serial number? If so it might be cleaner to deprecate the /:id/revoke endpoint and make a new one like /revoke that takes a json body which can be `id`, `serialNumber`, or `thumbprint`.

bernie-g (approved):

PR #6848, opened 2026-06-13
feat(kms): add unexportable flag to KMS keys

bernie-g:
## Context

Add a first-class `isExportable` attribute to KMS keys so a key can be permanently marked as non-exportable at creation time, preventing raw key material from ever being retrieved via API regardless of RBAC permissions.

## Steps to verify the change

- Create a new KMS key with "Allow Export" toggled off
- Verify the key cannot be exported via single or bulk export
- Verify encrypt/decrypt/sign/verify operations still work normally
- Verify existing keys default to exportable
- Verify the setting cannot be changed after key creation

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/cmek/cmek-service.ts
```diff
@@ -314,6 +314,7 @@ export const cmekServiceFactory = ({
     if (!key) throw new NotFoundError({ message: `Key with ID "${keyId}" not found` });
     if (!key.projectId || key.isReserved) throw new BadRequestError({ message: "Key is not customer managed" });
     if (key.isDisabled) throw new BadRequestError({ message: "Key is disabled" });
+    if (!key.isExportable) throw new BadRequestError({ message: "Key is not exportable" });
```
saifsmailbox98: a user who doesn't have the permission to export shouldn't see "Key is not exportable" (info leak), rather we should respond with you are not allowed to export

frontend/src/pages/kms/OverviewPage/components/CmekBulkImportModal.tsx
```diff
@@ -49,6 +49,7 @@ type ParsedKey = {
   keyMaterial?: string;
```
saifsmailbox98, line 49: IGNORE COMMENT PLACEMENT

the import API accepted isExportable, but the modal doesn't show that in the "expected format", we should make the users aware of the following:
1. existence of isExportable
2. its default value, if not passed (and that it's optional)
3. what it means and that it is irreversible
4. confirmation in the preview table (with similar indication as the main table)

Image

frontend/src/pages/kms/OverviewPage/components/CmekTable.tsx
```diff
@@ -568,7 +574,21 @@ export const CmekTable = () => {
...
                         <TableCell className="uppercase">{encryptionAlgorithm}</TableCell>
                         <TableCell>
-                          <Badge variant={variant}>{label}</Badge>
+                          <div className="flex items-center gap-1">
+                            <Badge variant={variant}>{label}</Badge>
+                            {!isExportable && (
+                              <Tooltip>
+                                <TooltipTrigger asChild>
+                                  <span className="inline-flex">
+                                    <Badge variant="neutral">Non-Exportable</Badge>
```
saifsmailbox98: we shouldn't mix the exportability indicator with the status because status can change over time and has nothing to do with exportability. can we use an icon here, like a lock icon, instead of a badge? either at the same place or in a small icon column for alignment, similar to what we have for the actions menu (`...`)


Image

backend/src/db/migrations/20260612142200_add-is-exportable-to-kms-keys.ts
```diff
@@ -0,0 +1,23 @@
+import { Knex } from "knex";
```
saifsmailbox98, line 1: the file name can be a little more precise, right now it's just a day, though the full file name will not collide with any other, but not sure how migrations would be treated in the future, better to avoid collision at the timestamp level as well

saifsmailbox98 (changes requested):

backend/src/db/migrations/20260612142200_add-is-exportable-to-kms-keys.ts
bernie-g: Fixed

frontend/src/pages/kms/OverviewPage/components/CmekTable.tsx
bernie-g: Added the lock icon

frontend/src/pages/kms/OverviewPage/components/CmekBulkImportModal.tsx
bernie-g: Added all of those

backend/src/services/cmek/cmek-service.ts
bernie-g: Changed the msg

saifsmailbox98 (approved):

PR #6849, opened 2026-06-14
feat(dynamic-secret): make GCP IAM token OAuth scopes configurable

dangtony98:
Screenshot 2026-06-13 at 9 43 05 PM

## Context

The GCP IAM dynamic secret provider mints short-lived OAuth access tokens by having an Infisical-controlled service account impersonate a user-specified target service account (via google-auth-library's `Impersonated` client).

The OAuth scopes on the minted token were **hardcoded** to `iam` + `cloud-platform`, with no way to change them. As a result, tokens could not carry scopes like `https://www.googleapis.com/auth/drive.readonly`, so any Google API outside `cloud-platform`'s coverage (Drive, Sheets, etc.) failed. Because OAuth scopes can only be **narrowed** after issuance, never widened, this could not be worked around on the consuming side; the scope must be set at mint time, in the provider.

This PR adds a configurable `tokenScopes` input that **defaults to the current two scopes**, so existing leases are unaffected.

**Backward compatibility:** the Zod `.default([...iam, ...cloud-platform])` fills in the original scopes for any dynamic secret stored before this change, so existing leases mint identical tokens. Provider inputs are stored as an encrypted JSON blob (`encryptedInput`), not as table columns, so **no database migration is required**.

**Scope of change:** service-account impersonation path only (`targetScopes`). It does **not** add domain-wide delegation / user impersonation (`subject`), which would be a separate, larger change and is out of scope here.

### Changes
- **Backend** — added `tokenScopes` to `DynamicSecretGcpIamSchema` and threaded it through `$getToken` into the `Impersonated` client's `targetScopes`. The source JWT client's `cloud-platform` scope (which governs the impersonation API call, not the minted token) is unchanged.
- **Frontend** — added `tokenScopes` to the GCP IAM DTO and surfaced it in the create and edit forms as an add/remove list, mirroring the existing Kubernetes `audiences` field pattern.
- **Docs** — documented the new `Token Scopes` parameter.

## Steps to verify the change

1. Create a GCP IAM dynamic secret leaving Token Scopes at the defaults → lease mints a token as before.
2. Create one with `https://www.googleapis.com/auth/drive.readonly` added → minted token carries the Drive scope (verify via `https://oauth2.googleapis.com/tokeninfo?access_token=<token>`).
3. Edit an existing (pre-change) secret → scope list shows the two defaults; saving and generating a new lease works.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: `type(scope): short description`.
- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/dynamic-secret/providers/models.ts
+++ b/backend/src/ee/services/dynamic-secret/providers/models.ts
@@ -615,7 +615,12 @@ export const DynamicSecretTotpSchema = z.discriminatedUnion("configType", [
 ]);
 
 export const DynamicSecretGcpIamSchema = z.object({
-  serviceAccountEmail: z.string().email().trim().min(1, "Service account email required").max(128)
+  serviceAccountEmail: z.string().email().trim().min(1, "Service account email required").max(128),
+  tokenScopes: z
+    .array(z.string().trim().min(1))
+    .min(1, "At least one scope is required")
+    .default(["https://www.googleapis.com/auth/iam", "https://www.googleapis.com/auth/cloud-platform"])
+    .describe("OAuth scopes for the generated access token.")
 });
 
 export const DynamicSecretGithubSchema = z.object({
--- a/docs/documentation/platform/dynamic-secrets/gcp-iam.mdx
+++ b/docs/documentation/platform/dynamic-secrets/gcp-iam.mdx
@@ -108,7 +108,7 @@ The Infisical GCP IAM dynamic secret allows you to generate GCP service account 
        		The email tied to the service account created in earlier steps.
     	</ParamField>
     	<ParamField path="Token Scopes" type="string[]" default="https://www.googleapis.com/auth/iam, https://www.googleapis.com/auth/cloud-platform">
-       		The OAuth scopes granted to the generated access token. Defaults to the `iam` and `cloud-platform` scopes. Add scopes such as `https://www.googleapis.com/auth/drive.readonly` to access Google APIs outside `cloud-platform`'s coverage.
+       		The OAuth scopes granted to the generated access token. Defaults to `https://www.googleapis.com/auth/iam` and `https://www.googleapis.com/auth/cloud-platform` scopes. You can add additional scopes such as `https://www.googleapis.com/auth/drive`.
     	</ParamField>
   </Step>
   <Step title="Click `Submit`">
--- a/docs/documentation/platform/dynamic-secrets/gcp-iam.mdx
+++ b/docs/documentation/platform/dynamic-secrets/gcp-iam.mdx
@@ -6,11 +6,17 @@ description: "Learn how to dynamically generate GCP service account tokens."
 The Infisical GCP IAM dynamic secret allows you to generate GCP service account tokens on demand based on service account permissions.
 
 <Warning>
-    GCP service account access tokens cannot be revoked. As such, revoking or regenerating a token does not invalidate the old one; it remains active until it expires.
+  GCP service account access tokens cannot be revoked. As such, revoking or
+  regenerating a token does not invalidate the old one; it remains active until
+  it expires.
 </Warning>
 
 <Note>
-    You must enable the [IAM API](https://console.cloud.google.com/apis/library/iam.googleapis.com) and [IAM Credentials API](https://console.cloud.google.com/apis/library/iamcredentials.googleapis.com) in your GCP console as a prerequisite
+  You must enable the [IAM
+  API](https://console.cloud.google.com/apis/library/iam.googleapis.com) and
+  [IAM Credentials
+  API](https://console.cloud.google.com/apis/library/iamcredentials.googleapis.com)
+  in your GCP console as a prerequisite
 </Note>
 
 <Accordion title="Self-Hosted Instance">
@@ -41,6 +47,7 @@ The Infisical GCP IAM dynamic secret allows you to generate GCP service account 
             4. You can now use GCP integration with service account impersonation.
         </Step>
     </Steps>
+
 </Accordion>
 
 ## Create GCP Service Account
@@ -72,6 +79,7 @@ The Infisical GCP IAM dynamic secret allows you to generate GCP service account 
 
         ![Service Account Page](/images/app-connections/gcp/service-account-grant-access.png)
     </Step>
+
 </Steps>
 
 ## Set up Dynamic Secrets with GCP IAM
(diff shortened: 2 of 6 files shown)


frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/CreateDynamicSecretForm/GcpIamInputForm.tsx
```diff
@@ -187,6 +201,41 @@ export const GcpIamInputForm = ({
...
+                          </IconButton>
+                        </div>
+                      ))}
+                      <Button variant="outline_bg" onClick={() => append("")} type="button">
+                        Add Scope
+                      </Button>
+                    </div>
+                  </FormControl>
+                )}
+              />
```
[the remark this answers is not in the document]
dangtony98: Good catch, fixed in 71f48fbd7b. Each scope row is now wrapped in its own `Controller` with an indexed name (`provider.tokenScopes.${index}`) that surfaces `fieldState.error` inline, so a blank scope now shows an error on that row instead of silently blocking submit. Adopted the per-item pattern already used by the MongoDB Atlas form.

frontend/src/pages/secret-manager/SecretDashboardPage/components/DynamicSecretListView/EditDynamicSecretForm/EditDynamicSecretGcpIamForm.tsx
```diff
@@ -166,6 +178,41 @@ export const EditDynamicSecretGcpIamForm = ({
...
+                    <Button variant="outline_bg" onClick={() => append("")} type="button">
+                      Add Scope
+                    </Button>
+                  </div>
+                </FormControl>
+              )}
+            />
           </div>
         </div>
         <div className="mt-4 flex items-center space-x-4">
```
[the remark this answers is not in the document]
dangtony98: Fixed in 71f48fbd7b, same per-item `Controller` change applied here as in the create form.

frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/CreateDynamicSecretForm/GcpIamInputForm.tsx
```diff
@@ -41,7 +44,7 @@ const formSchema = z
     path: ["maxTTL"],
     message: "Max TTL must be greater than or equal to Default TTL"
   });
-type TForm = z.infer<typeof formSchema>;
+type TForm = z.infer<typeof formSchema> & FieldValues;
```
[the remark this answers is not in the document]
dangtony98: Verified this is required, not removable. `tokenScopes` is a primitive `string[]`, and `useFieldArray`'s `name` must extend `FieldArrayPath<TForm>`, which resolves to `never` for a primitive-array path. Without the `& FieldValues` intersection, type-check fails with `TS2322: Type 'string' is not assignable to type 'never'` on the `name` argument. This is the same workaround the existing Kubernetes dynamic-secret form uses for its primitive `audiences` string array, so it also keeps us consistent with the established pattern. Leaving it as-is.

dangtony98:
@greptileai

docs/documentation/platform/dynamic-secrets/gcp-iam.mdx
```diff
@@ -99,6 +107,9 @@ The Infisical GCP IAM dynamic secret allows you to generate GCP service account
     	<ParamField path="Service Account Email" type="string" required>
        		The email tied to the service account created in earlier steps.
     	</ParamField>
+    	<ParamField path="Token Scopes" type="string[]" default="https://www.googleapis.com/auth/iam, https://www.googleapis.com/auth/cloud-platform">
+       		The OAuth scopes granted to the generated access token. Defaults to the `iam` and `cloud-platform` scopes. Add scopes such as `https://www.googleapis.com/auth/drive.readonly` to access Google APIs outside `cloud-platform`'s coverage.
```
saifsmailbox98: > access Google APIs outside `cloud-platform`'s coverage.

Better to just mention the defaults and then mention that you can add more scopes to it, for example, to access drive, etc.

frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/CreateDynamicSecretForm/GcpIamInputForm.tsx
```diff
@@ -187,6 +201,54 @@ export const GcpIamInputForm = ({
...
+
+              <Controller
+                control={control}
+                name="provider.tokenScopes"
+                render={({ fieldState: { error } }) => (
+                  <FormControl
+                    label="Token Scopes"
+                    isError={Boolean(error?.message)}
+                    errorText={error?.message}
+                    helperText="OAuth scopes granted to the generated access token. Scopes can only be narrowed after issuance, never widened."
```
saifsmailbox98: What is the process of narrowing the scope if it has got nothing to do with Infisical or dynamic secrets? I think we should not mention it.

backend/src/ee/services/dynamic-secret/providers/models.ts
```diff
@@ -615,7 +615,12 @@ export const DynamicSecretTotpSchema = z.discriminatedUnion("configType", [
 ]);
 
 export const DynamicSecretGcpIamSchema = z.object({
-  serviceAccountEmail: z.string().email().trim().min(1, "Service account email required").max(128)
+  serviceAccountEmail: z.string().email().trim().min(1, "Service account email required").max(128),
+  tokenScopes: z
```
saifsmailbox98, line 619: Are we handling users passing duplicate scopes? Both here and on the front end

frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/CreateDynamicSecretForm/GcpIamInputForm.tsx
saifsmailbox98: > & FieldValues kills type safety for the whole form. Model scopes as z.object({ value: z.string() }).array() instead of z.string().array() so useFieldArray works without this

same in the edit form as well

saifsmailbox98 (changes requested):

backend/src/ee/services/dynamic-secret/providers/models.ts
dangtony98: Good call. Handled in d8c8f61c36, but deduping silently rather than rejecting, to stay in line with how dynamic secrets already treat list inputs.

GCP IAM was actually the only provider doing schema-level uniqueness validation. SSH is the one provider that dedups, and it does so silently: the backend collapses principals via `new Set()` at consume time and never errors. So GCP IAM now matches that, the backend dedups with `targetScopes: [...new Set(tokenScopes)]` before the impersonation call, and both forms dedup on submit. No validation error is surfaced.

frontend/src/pages/secret-manager/SecretDashboardPage/components/ActionBar/CreateDynamicSecretForm/GcpIamInputForm.tsx
dangtony98: Done in d8c8f61c36. Modeled `tokenScopes` as `z.object({ value: z.string() }).array()` in both the create and edit forms, dropped the `& FieldValues` intersection, and `TForm` is now a plain `z.infer<typeof formSchema>`. `useFieldArray` is fully typed without the workaround, and values are mapped to/from `string[]` at the API boundary.

This follows the existing MongoDB Atlas form, which already uses the object-array pattern with a plain `z.infer` type.

dangtony98:
@greptileai

saifsmailbox98 (approved):

PR #6856, opened 2026-06-14
fix(secrets): respect personal overrides on imported secrets

### LINKED ISSUE #4842: Override of imported secrets not respected
### Describe the bug
When personal override of some secret is set in one folder `/A`. Secrets from the folder `/A` are imported to another folder `/B`, and than infisical cli is used to export secrets from folder `/B`. In that case the overridden secrets in `/A` are not imported with overridden values, but with the original values. What should happen is that exporting secrets from B, which has imported secrets from A, should in any case contain overridden values of folders A and B.

### To Reproduce
- install infisical CLI
- infisical login with your personal credentials
- create project for secrets called `Proj`
- in `Proj` create folder called `/A`, which contains some secrets
- in `Proj` create folder called `/B`, which has different secrets.
- both folders are on the same level in the `Proj`
- import secrets from `/A` to `/B`
- override single secret with personal override in `/A`
- use Infisical CLI with `export` command to export and save secrets from folder /B to some .env file.
- `infisical export --format=dotenv --projectId=some-guid --env=dev --path=/B --output-file=.env`
- observe that exported secret does not contain overriden values in A
- adding `--secret-overriding=True` flag does not have any effect on the faulty behaviour. 

### Expected behavior
Overridden secrets in A, should be exported with overridden values, not the default ones, when B secrets are exported.

OS: Fedora Linux 42 (Workstation Edition) x86_64
Kernel: Linux 6.17.7-200.fc42.x86_64

When overriding values in `/B` and exporting them, it works OK.

It might be partially related to this issue: #4627

sanidhyasin:
## Context

Closes #4842.

When a folder imports secrets from another folder and a source secret has a **personal override**, reading the importing folder returned the original (shared) value instead of the override. This is most visible with `infisical export` of a folder that imports secrets — the exported values ignored personal overrides on the imported secrets (and `--secret-overriding` had no effect on them). Direct (non-imported) overrides already worked.

Two issues in the secret v2 bridge caused this:

1. **`getSecrets` gated the actor on `expandPersonalOverrides`.** It passed `userId: expandPersonalOverrides ? actorId : undefined` into `fnSecretsV2FromImports`. The v3 raw endpoint (used by the CLI export) requests `personalOverridesBehavior: IncludeAll` but does not set `expandPersonalOverrides`, so personal secrets were never fetched for imports. Direct secrets always pass the actor and let `personalOverridesBehavior` decide inclusion, so imports now do the same.

2. **`fnSecretsV2FromImports` de-duplicated imported secrets by key.** A shared secret and its personal override share the same key but are distinct secrets (exactly as they are for direct secrets under `IncludeAll`), so the override was collapsed away. De-duplication now keys on `key + type` for `IncludeAll`, while keeping key-only de-duplication for the behaviors that resolve to a single effective value per key (`Priority`, default).

## Steps to verify the change

1. In a project, create a secret in `/staging` (e.g. `OVERRIDE_KEY=shared-value`).
2. Add a personal override for it (`OVERRIDE_KEY=personal-value`).
3. In `/dev`, add a secret import from `/staging`.
4. Read `/dev` with imports included (e.g. `infisical export --env=dev --path=/`, or `GET /api/v3/secrets/raw?...&include_imports=true`).
5. The imported `OVERRIDE_KEY` now exposes the personal override; before this change it only returned `shared-value`.

A new e2e regression test (`backend/e2e-test/routes/v1/secret-import.spec.ts`) covers this. Existing secret/import/reference e2e suites continue to pass.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/e2e-test/routes/v1/secret-import.spec.ts
+++ b/backend/e2e-test/routes/v1/secret-import.spec.ts
@@ -812,6 +812,8 @@ describe.each([{ path: "/" }, { path: "/deep" }])(
 // a personal override on a secret in the source folder must be respected when the
 // secret is read through an import (e.g. CLI export of a folder that imports it).
 describe("Secret import personal override testing", () => {
+  // Setup and teardown live in beforeAll so cleanup runs unconditionally, even if an
+  // assertion in the test below fails.
   beforeAll(async () => {
     const devImportFromStaging = await createSecretImport({
       authToken: jwtAuthToken,
@@ -822,18 +824,6 @@ describe("Secret import personal override testing", () => {
       importEnv: "staging"
     });
 
-    return async () => {
-      await deleteSecretImport({
-        id: devImportFromStaging.id,
-        workspaceId: seedData1.projectV3.id,
-        environmentSlug: seedData1.environment.slug,
-        secretPath: "/",
-        authToken: jwtAuthToken
-      });
-    };
-  });
-
-  test("Personal override on an imported secret is respected", async () => {
     // shared secret living in the imported (staging) folder
     await createSecretV2({
       environmentSlug: "staging",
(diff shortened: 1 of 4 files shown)


sanidhyasin:
Hi @akhilmhdh, gentle nudge on this when you have time. It makes imported secrets respect personal overrides, with e2e tests added. Happy to rebase or change anything if needed.

sanidhyasin:
@varonix0 friendly ping — this fixes personal overrides being ignored on imported secrets (#4842). Checks are passing. Could you take a look when you get a chance? Thanks!

Issue #6552, opened 2026-05-20
SCIM v2 Group endpoints fail: POST 500 'Find one', PUT/PATCH 422 missing schemas/members[].display, ServiceProviderConfig 404


derrix060:
# Infisical SCIM v2 — multiple defects observed via Authentik 2026.2 client

## Environment
- Infisical SaaS (`https://app.infisical.com`)
- Org: nethermind (SCIM endpoint `https://app.infisical.com/api/v1/scim`)
- Identity Provider: Authentik 2026.2 (`compatibility_mode = default`); same behaviour observed across `aws`, `slack`, `sfdc`
- Observed: 2026-05-20 ~15:46–17:35 UTC

## Defect 1 — `ServiceProviderConfig` endpoint returns 404 (RFC 7644 §4 violation)

```
$ curl -H "Authorization: Bearer <scim_token>" \
    https://app.infisical.com/api/v1/scim/v2/ServiceProviderConfig
{"message":"Route GET:/api/v1/scim/v2/ServiceProviderConfig not found","error":"Not Found","statusCode":404}
```

RFC 7644 §4 mandates this endpoint so SCIM clients can discover capabilities (patch support, bulk support, filter, etag, etc.). Authentik can't negotiate down to what Infisical accepts because the discovery surface is missing.

## Defect 2 — `POST /Groups` returns HTTP 500 "Find one"

Every group create attempt returns 500. Reproduced on both pre-existing and freshly deleted groups (so it is not a stale-state issue).

```
{
  "reqId": "req-HXLZlgagbMm0xp",
  "statusCode": 500,
  "message": "Something went wrong",
  "error": "Find one"
}
```

Other example reqIds: `req-a2kQWCwfWKvm07` (role_operations_devops).

The error string "Find one" looks like an internal ORM `findOne` call returning null on the create path. Worth inspecting whether the create handler tries to look up a related record (org? membership? schema?) and 500s when it can't find it.

## Defect 3 — `PUT /Groups/<id>` rejects body without explicit `schemas` array (422 ValidationFailure)

```
{
  "reqId": "req-y6hcivQtXTLFJg",
  "statusCode": 422,
  "error": "ValidationFailure",
  "message": [
    {
      "code": "invalid_type",
      "expected": "array",
      "received": "undefined",
      "path": ["schemas"],
      "message": "Required"
    }
  ]
}
```

[...]

PR #6860, opened 2026-06-15
fix(scim): make PUT /Groups schemas optional per RFC 7644 §3.4.2

jaydeep-pipaliya:
## Context

Fixes part of #6552 — the PUT /Groups endpoint rejects the request when the body has no `schemas` field.

Today the route requires `schemas` as an array. But RFC 7644 §3.4.2 says the server can figure out the schema from the URL, so clients like Authentik don't send it. The PATCH route on the same file already treats `schemas` as optional. This PR does the same for PUT.

Only changed one Zod field. Left the other three defects from #6552 alone — they each need their own PR.

## Type

- [x] Fix

## Steps to verify the change

Send a PUT to `/api/v1/scim/Groups/<groupId>` without `schemas`:

- Before: 422, `schemas Required`
- After: 200, group updated

Requests that still include `schemas` keep working as before.

## Checklist

- [x] Tested locally with curl

### DIFF
--- a/backend/src/ee/routes/v1/scim-router.ts
+++ b/backend/src/ee/routes/v1/scim-router.ts
@@ -507,7 +507,11 @@ export const registerScimRouter = async (server: FastifyZodProvider) => {
         groupId: z.string().trim()
       }),
       body: z.object({
-        schemas: z.array(z.string()),
+        // Optional per RFC 7644 §3.4.2 — the server can infer the resource schema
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Closing - Defect 3 (PUT /Groups schemas optional) and ServiceProviderConfig were addressed by #6885 which has since been merged. This PR is superseded.

PR #6861, opened 2026-06-15
fix(scim): add ServiceProviderConfig endpoint per RFC 7644 §5

jaydeep-pipaliya:
## Context

Fixes another defect from #6552 — `GET /api/v1/scim/ServiceProviderConfig` returns 404.

RFC 7644 §5 makes this endpoint mandatory. SCIM clients hit it to find out what the server supports (patch, bulk, filter, etag, etc.). Without it, Authentik and other clients can't figure out what operations to use and fall back to assumptions that don't match Infisical's actual surface.

Added a small static handler that returns the config. Values match what Infisical actually supports today: patch yes, filter yes, bulk/changePassword/sort/etag no, bearer token auth.

## Type

- [x] Fix

## Steps to verify the change

```
curl -H "Authorization: Bearer <scim_token>" \
  https://<host>/api/v1/scim/ServiceProviderConfig
```

Before: 404. After: 200 with the config JSON.

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/routes/v1/scim-router.ts
+++ b/backend/src/ee/routes/v1/scim-router.ts
@@ -45,7 +45,49 @@ const ScimGroupSchema = z.object({
   })
 });
 
+// SCIM /ServiceProviderConfig response per RFC 7644 §5. Describes which SCIM
+// capabilities Infisical's implementation supports so clients (Authentik etc.)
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Closing - ServiceProviderConfig endpoint was added in #6885 which has since been merged. This PR is superseded.

PR #6862, opened 2026-06-15
fix(signup): read user row from primary to avoid stale replica after deletion

### LINKED ISSUE #6034: Bug: Account data persists after deletion and re-registration with same email
### Describe the bug
After deleting an account and signing up again with the same email, the previous account information is still retained.

Because of this, the user is unable to recover the password, and deleting/re-registering does not solve the issue.

### To Reproduce
1. Sign up with an email account
2. Delete the account
3. Sign up again with the same email
4. Try to log in or use the password recovery feature
5. See that the previous account data is still retained and the issue is not resolved

### Expected behavior
If an account is deleted, signing up again with the same email should create a completely new account.

The user should also be able to recover the password or access the account normally after re-registering.

### Screenshots
N/A

- Web

### Additional context
There is currently no workaround for the user.

Even after deleting the account and registering again with the same email, the issue still remains.

It would help if:
- account deletion fully removed all related account data, or
- an admin password reset feature was provided

jaydeep-pipaliya:
## Context

Fixes #6034 — after a user deletes their account and immediately signs up again with the same email, they get the "account already exists" email instead of the OTP, and password recovery doesn't work either.

The cause is replica lag. `beginEmailSignupProcess` calls `userDAL.findOne` without a tx, so it goes to the read replica. The delete happened on the primary, but the replica might still have the old row with `isAccepted=true`. The flow sees it and bails into the "existing account" branch.

Fix is small: wrap the lookup in `userDAL.transaction` so it hits the primary. Single-statement Postgres transactions don't cost more than a normal query, so the signup hot path is unchanged.

## Type

- [x] Fix

## Steps to verify the change

1. Sign up with email A and complete the flow.
2. Delete the account from Personal Settings.
3. Immediately sign up again with the same email A.
4. Before: "account already exists" email arrives, no OTP, password recovery fails.
5. After: fresh OTP arrives as expected.

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/auth/auth-signup-service.ts
+++ b/backend/src/services/auth/auth-signup-service.ts
@@ -125,23 +125,33 @@ export const authSignupServiceFactory = ({
     await tokenService.validateEmailSignupToken(sanitizedEmail, code);
 
     // Only create (or recover) the user after the OTP has been verified.
-    let user = await userDAL.findOne({ username: sanitizedEmail });
-    if (!user) {
-      user = await userDAL.create({
-        authMethods: [AuthMethod.EMAIL],
-        username: sanitizedEmail,
-        email: sanitizedEmail,
-        isGhost: false,
-        isEmailVerified: true
-      });
-    } else if (!user.isAccepted) {
-      // Migration path: pre-user created by the old flow before this refactor.
-      await userDAL.updateById(user.id, { isEmailVerified: true });
-      user = { ...user, isEmailVerified: true };
-    } else {
+    // Run the lookup + create/update inside a transaction so the read hits
+    // the primary. If replication still hasn't caught up with the delete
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Good catch by Greptile — the same unguarded `findOne` exists in `verifyEmailSignup` at line 128 and has the exact same replica-lag exposure. Even though the OTP step happens seconds after the begin step (so the replica has more time to catch up), under load that lag can spike to multiple seconds and the user would hit "Invalid or expired verification request" right after a valid OTP.

Pushed a follow-up that wraps the lookup + create/update in `verifyEmailSignup` inside the same `userDAL.transaction` pattern. Bonus: doing the create/update inside the same tx makes the lookup-then-write atomic, which is the right shape for this branch anyway.

PR #6872, opened 2026-06-15
fix: replace 500 error with bad request due to bitbucket returning 429

mathnogueira:
## Context

We had a customer complaining about bitbucket failing with 500 error. However, after investigation, it was failing because BitBucket was returning 429 errors. 

We, unfortunately, cannot give more details because BitBucket doesn't return more information: for example, how long to wait. Bitbucket rate limit windows are 1h long.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/app-connection/bitbucket/bitbucket-connection-fns.ts
+++ b/backend/src/services/app-connection/bitbucket/bitbucket-connection-fns.ts
@@ -146,32 +146,15 @@ export const listBitbucketRepositories = async (
   const encodedSlug = encodeURIComponent(workspaceSlug);
 
   try {
+    const baseUrl = new URL(`${IntegrationUrls.BITBUCKET_API_URL}/2.0/repositories/${encodedSlug}`);
+    baseUrl.searchParams.set("pagelen", String(BITBUCKET_PAGE_SIZE));
+    baseUrl.searchParams.set("sort", "slug");
     if (search) {
-      const baseUrl = new URL(`${IntegrationUrls.BITBUCKET_API_URL}/2.0/repositories/${encodedSlug}`);
-      baseUrl.searchParams.set("pagelen", String(BITBUCKET_PAGE_SIZE));
-      baseUrl.searchParams.set("sort", "slug");
       baseUrl.searchParams.set("q", `name ~ "${search.replace(/"/g, "")}"`);
-
-      const { data } = await request.get<BitbucketPaginatedResponse<TBitbucketRepo>>(baseUrl.toString(), { headers });
-      return data.values;
     }
 
-    // Fetch repos per-project to avoid Bitbucket's 1,000-result pagination cap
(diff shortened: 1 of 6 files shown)


backend/src/services/app-connection/bitbucket/bitbucket-connection-fns.ts
```diff
@@ -124,22 +141,26 @@ export const listBitbucketRepositories = async (appConnection: TBitbucketConnect
...
       )
-    )
-  );
+    );
 
-  return reposByProject.flat();
+    return reposByProject.flat();
+  } catch (error) {
+    ensureBitbucketRateLimitNotExceeded(error);
+  }
```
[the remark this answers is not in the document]
mathnogueira: false positive.

mathnogueira:
@greptile review this again

backend/src/services/app-connection/bitbucket/bitbucket-connection-fns.ts
```diff
@@ -154,30 +189,8 @@ export const listBitbucketEnvironments = async (
...
-    if (data.next) {
-      environmentsUrl = data.next;
-    } else {
-      hasNextPage = false;
-    }
-    iterationCount += 1;
-  }
-
-  return environments;
+  return paginateBitbucketRequest<TBitbucketEnvironment>(
```
mathnogueira, line 177: environments don't support searching by a term, so it's the only endpoint we still need to run this logic.

backend/src/services/app-connection/bitbucket/bitbucket-connection-fns.ts
```diff
@@ -94,27 +110,32 @@ interface BitbucketPaginatedResponse<T> {
...
 const paginateBitbucketRequest = async <T>(url: string, headers: Record<string, string>): Promise<T[]> => {
   let allItems: T[] = [];
   let nextUrl: string | undefined = url;
   let iterationCount = 0;
 
-  while (nextUrl && iterationCount < BITBUCKET_MAX_PAGES) {
-    // eslint-disable-next-line no-await-in-loop
-    const { data }: { data: BitbucketPaginatedResponse<T> } = await request.get(nextUrl, { headers });
+  try {
+    while (nextUrl && iterationCount < BITBUCKET_MAX_PAGES) {
```
varonix0, line 119: Why are we still doing pagination? Wouldn't that mean that the 429 error would still be present because we're still looping through everything if there's no search?

I think we should just get a single page based on the search, and if no search is supplied just return a single page, otherwise we'll have the same issue

varonix0 (changes requested):

varonix0 (approved):

PR #6874, opened 2026-06-15
feat: add FQDN support to PKI Syncs schemas

carlosmonastyrski:
## Context

This PR makes certificate sync naming far more flexible. Until now, when you set up a certificate sync to a destination like NetScaler, Azure, AWS, Chef, Cloudflare, or F5, you could only name the synced certificates using the certificate's internal ID. This update lets you build those names from meaningful certificate details instead, you can now use the certificate's common name, its profile, and the application it belongs to, in any combination with your own text. The names are automatically cleaned up to fit each destination's rules (for example, dots in a domain name are adjusted where a destination doesn't allow them), and any name that wouldn't be accepted by a destination is caught upfront when you save the sync rather than failing silently later.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/pki-sync/aws-secrets-manager/aws-secrets-manager-pki-sync-fns.ts
+++ b/backend/src/services/pki-sync/aws-secrets-manager/aws-secrets-manager-pki-sync-fns.ts
@@ -20,6 +20,11 @@ import { TCertificateDALFactory } from "@app/services/certificate/certificate-da
 import { TCertificateSyncDALFactory } from "@app/services/certificate-sync/certificate-sync-dal";
 import { CertificateSyncStatus } from "@app/services/certificate-sync/certificate-sync-enums";
 import { createConnectionQueue, RateLimitConfig } from "@app/services/connection-queue";
+import {
+  certificateNameSchemaHasFreeTextPlaceholder,
+  sanitizeCertificateNameValue
+} from "@app/services/pki-sync/pki-sync-certificate-name-fns";
+import { PkiSync } from "@app/services/pki-sync/pki-sync-enums";
 import { matchesCertificateNameSchema } from "@app/services/pki-sync/pki-sync-fns";
 import { TCertificateMap, TPkiSyncWithCredentials } from "@app/services/pki-sync/pki-sync-types";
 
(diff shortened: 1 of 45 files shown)


frontend/src/components/pki-syncs/forms/PkiSyncOptionsFields/PkiSyncOptionsFields.tsx
saifsmailbox98: this field is optional but you can't clear it as it snaps back to the original value

frontend/src/components/pki-syncs/forms/PkiSyncOptionsFields/PkiSyncOptionsFields.tsx
```diff
@@ -434,7 +434,23 @@ export const PkiSyncOptionsFields = ({ destination }: Props) => {
...
+                    </li>
+                    <li>
+                      <code>{"{{applicationId}}"}</code> - The ID of the application the sync
+                      belongs to
+                    </li>
                   </ul>
+                  <span className="mt-1 text-xs text-bunker-300">
+                    The schema must include <code>{"{{certificateId}}"}</code> to guarantee a unique
+                    name per certificate. Characters not allowed by the destination (e.g. dots in a
+                    common name for Azure Key Vault) are automatically replaced with hyphens.
```
saifsmailbox98: mentioning azure key vault could confuse users setting up some other sync, maybe we can make it either dynamic or remove it

frontend/src/components/pki-syncs/forms/PkiSyncOptionsFields/PkiSyncOptionsFields.tsx
saifsmailbox98, line 463: most users won't need to be told the forbidden characters, instead of showing it upfront, can we just rely on validation errors?

frontend/src/components/pki-syncs/forms/PkiSyncOptionsFields/PkiSyncOptionsFields.tsx
```diff
@@ -434,7 +434,23 @@ export const PkiSyncOptionsFields = ({ destination }: Props) => {
...
+                      <code>{"{{profileId}}"}</code> - The certificate profile ID (falls back to the
+                      certificate ID when none is set)
+                    </li>
+                    <li>
+                      <code>{"{{applicationId}}"}</code> - The ID of the application the sync
+                      belongs to
+                    </li>
                   </ul>
+                  <span className="mt-1 text-xs text-bunker-300">
+                    The schema must include <code>{"{{certificateId}}"}</code> to guarantee a unique
```
saifsmailbox98: this is a bit confusing - it says characters are "not allowed" but also "automatically replaced", so it's unclear what actually happens. maybe something like "your template can only use letters, numbers, dashes, underscores, and slashes. when placeholders resolve, unsupported characters are replaced with hyphens" could be clearer

saifsmailbox98 (changes requested):

frontend/src/components/pki-syncs/forms/PkiSyncOptionsFields/PkiSyncOptionsFields.tsx
carlosmonastyrski: Changing it to a more generic version 👍

backend/src/services/pki-sync/aws-secrets-manager/aws-secrets-manager-pki-sync-fns.ts
```diff
@@ -20,6 +20,11 @@ import { TCertificateDALFactory } from "@app/services/certificate/certificate-da
 import { TCertificateSyncDALFactory } from "@app/services/certificate-sync/certificate-sync-dal";
```
saifsmailbox98, line 19: Ignore comment placement


tested aws secrets manager with schema `{{commonName}}-{{certificateId}}` and CN `hello2.com` - the secret name came out as `hello2-com-1dea1094-b726-45ea-98c3-30b0e75ab2c2`, the dashes aren't being stripped



Image

saifsmailbox98 (changes requested):

saifsmailbox98 (approved):

PR #6885, opened 2026-06-16
fix: added serviceconfig scim endpoint, resovled duplicate user error

akhilmhdh:
## Context

This resolves ENG-5114 in which users reported following things
1. Requirement for the ServiceProviderConfig
2. Duplicate user listing with duplicate saml alias or oidc alias. This. is done by filtering user org configured alias method and also by selected latest  alias of the user.
3. Resolved optional field as schemas

## Steps to verify the change

1. Added test to replicate the scenerio. It should pass
2. Configure saml/oidc and then scim. Then just duplicate the user aliases and this should avoid. This can be duplicate due to external id changing etc. 
3. The scim should continue working. 

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/e2e-test/routes/v1/scim.spec.ts
+++ b/backend/e2e-test/routes/v1/scim.spec.ts
@@ -1,0 +1,425 @@
+/* eslint-disable no-await-in-loop */
+import crypto from "node:crypto";
+
+import jwt from "jsonwebtoken";
+import { Knex } from "knex";
+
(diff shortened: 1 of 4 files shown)


scott-ray-wilson (approved):
Looks good but I'm curious, this only mitigates the issue on read; shouldn't we also prevent the duplication from occurring in the first place?

PR #6888, opened 2026-06-16
fix(secret-rotation): planet scale authentication

adilsitos:
## Context
Update the username parsing on alter table so we can also take into account planet scale users. Also updated the docs 

## Screenshots
image

## Steps to verify the change
- Create a new postgresql app connection using planetscale (ask me for credentials)
- Create a new secret rotation for postgres
- follow the docs, and add the new users with the planetscale branch id at the end (eg. infisical_user.branch_id)
- check that the rotation is created 

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


adilsitos:
@greptile on the comment regarding "Safety check doesn't account for PlanetScale branch suffixes" this doesn't happen because the connection username is defined with the dot, which will make the check true and won't let the creation of this to happen. re review with this in mind and also update the summary

varonix0 (approved):

PR #6898, opened 2026-06-17
feat(billing): license server v2 customer billing page

PrestigePvP:
## Context

- Adds a new org Billing page (customer portal) behind the `LICENSE_SERVER_V2_ENABLED` env flag. The existing billing page is served unchanged when the flag is off.
- New backend surface under `/api/v1/organizations/:orgId/billing/v2/*` (overview, catalog, checkout-session, portal-session, payment-method), invisible (404) when the flag is off via a plugin-scoped `onRequest` gate.
- Reads (overview, catalog) come from the License Server through the `licenseClient` SDK. The Infisical backend acts as an authorizing proxy: it runs the CASL billing checks, mints a short-lived HS256 service JWT, and attaches `org_id`.
- Checkout and portal/payment actions create real Stripe sessions on the License Server and redirect. The catalog is fully data-driven from the License Server (products, plans, per-plan/cadence prices, comparison); no hardcoded catalog.
- CTA precedence: self-serve products show Upgrade/Activate, sales-led-only products show Contact sales, and a product offering both keeps a Contact sales option in the product sheet.
- Surfaces the flag to the frontend via `licenseServerV2Enabled` on the admin server config.

Linear: PLATFOR-416

### Known limitations (follow-ups)

- Machine-identity usage count and plan limits are not yet wired (the members count is real).
- Customer modes (enterprise-contract read-only, enterprise-stripe payment-only) are not yet differentiated; `mode` is `self-serve`.
- Completing a real checkout requires Stripe price IDs seeded on the License Server.

## Screenshots

UI billing page; screenshots can be attached on review.

## Steps to verify the change

- Flag off (default): the existing billing page renders and behaves as before, and `/billing/v2/*` returns 404.
- Flag on: the page renders catalog + subscription state; Upgrade/Activate redirects to Stripe checkout and Manage redirects to the Stripe billing portal.
- `make reviewable-api` and `make reviewable-ui`: backend and frontend type-check and lint pass.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/routes/v1/license-v2-router.ts
```diff
@@ -0,0 +1,228 @@
...
+
+export const registerLicenseV2Router = async (server: FastifyZodProvider) => {
+  // Every route is gated by LICENSE_SERVER_V2_ENABLED; the surface is invisible when the flag is off.
+  server.addHook("onRequest", async () => {
+    if (!server.services.licenseV2.isEnabled()) {
+      throw new NotFoundError({ message: "License Server v2 is not enabled" });
+    }
+  });
+
+  const buildActor = (permission: FastifyRequest["permission"]) => ({
```
akhilmhdh, line 101: I am a bit confused with this - like isn't this same as req.permission?

backend/src/ee/routes/v1/license-v2-router.ts
```diff
@@ -0,0 +1,228 @@
...
+  noun: z.string(),
+  monthly: z.number(),
+  annual: z.number(),
+  included: z.number()
+});
+
+const BillingV2CompareRowSchema = z.object({
+  label: z.string(),
+  pro: z.union([z.string(), z.boolean()]),
+  ent: z.union([z.string(), z.boolean()])
```
akhilmhdh, line 33: I think calling it `enterprise` would be better than `ent`. Took me a couple of seconds to understand the ent part 😅

backend/src/ee/routes/v1/license-v2-router.ts
```diff
@@ -0,0 +1,228 @@
...
+  const buildActor = (permission: FastifyRequest["permission"]) => ({
+    type: permission.type,
+    id: permission.id,
+    authMethod: permission.authMethod,
+    orgId: permission.orgId,
+    rootOrgId: permission.rootOrgId,
+    parentOrgId: permission.parentOrgId
+  });
+
+  server.route({
```
akhilmhdh, line 110: organization id can be removed from the URL. This was an old thing and now we can always infer it from the token. You can use the rootOrgId

backend/src/ee/services/license-v2/license-v2-service.ts
```diff
@@ -0,0 +1,435 @@
...
+};
+
+const centsToDollars = (cents: number | null | undefined): number => {
+  if (!cents) {
+    return 0;
+  }
+  return cents / 100;
+};
+
+const formatDate = (unixSeconds: number | null | undefined): string | null => {
```
akhilmhdh, line 65: You can use just the date-fns as well

backend/src/ee/services/license-v2/license-v2-service.ts
```diff
@@ -0,0 +1,435 @@
...
+    if (!catalog) {
+      return { products: [] };
+    }
+
+    const products = catalog.products.map(toCatalogProduct);
+    return { products };
+  };
+
+  // Stripe billing portal: manages the existing subscription, payment methods, and billing details.
+  const openPortal = async (orgId: string, returnPath?: string) => {
```
akhilmhdh, line 391: I think it would be better to keep internal functions as appended as `$`. This helps in review as well

frontend/src/pages/organization/BillingV2Page/components/primitives.tsx
```diff
@@ -0,0 +1,217 @@
+import {
+  ButtonHTMLAttributes,
+  CSSProperties,
+  InputHTMLAttributes,
+  MouseEvent,
+  ReactNode
+} from "react";
+
+import { IconExternal, IconX } from "../icons";
```
akhilmhdh: 🤔 May be you would need to catch up with @scott-ray-wilson on this - because are we ok with rolling out an entire suite of components for this? Can't we use the new v3 components

frontend/src/pages/organization/BillingV2Page/billing-v2.css
```diff
@@ -0,0 +1,463 @@
+/* Infisical — License v2 billing surface. Scoped under .billing-v2 to avoid global collisions. */
+
+.billing-v2 * { box-sizing: border-box; }
```
akhilmhdh: Same here! Not sure about this change as this would go wildly different from our existing setups.

akhilmhdh (changes requested):

akhilmhdh (approved):

PR #6900, opened 2026-06-17
fix(pki): mark subjectAltName critical for empty-subject certificates

### LINKED ISSUE #6899: Internal CA issues RFC 5280-invalid leaves over ACME (empty subject + non-critical SAN)
**Product:** Certificate Manager (PKI)

**Deployment Type:** Self-hosted

### Describe the bug

The internal CA's CSR-based issuance (ACME, EST, SCEP) produces X.509 leaves that violate **RFC 5280 §4.1.2.6**: when the CSR has no Common Name, the leaf gets an **empty subject DN** *and* a **non-critical** `subjectAltName`. The RFC requires the SAN to be **critical** whenever the subject is empty.

Standards-compliant ACME clients (Caddy/certmagic, certbot, lego, acme.sh) omit the Subject CN by default, so this hits the **default** ACME path, not an edge case. Apple's Security framework (macOS/iOS) rejects these certs; `openssl`/`curl`/Go validate via the SAN and accept them, which hides the bug from CLI testing.

Root cause (`backend/src/services/certificate-authority/internal/internal-certificate-authority-service.ts`): the SAN extension is built with the critical flag hardcoded to `false` on every path — `new x509.SubjectAlternativeNameExtension(altNamesArray, false)` — with no empty-subject check. In `signCertFromCa`, `subjectOverride` resolves to `""` for a CN-less CSR, so the leaf ends up empty-subject + non-critical SAN.

Same class of bug previously fixed in step-ca: smallstep/certificates#302 → smallstep/crypto commit `a28955b` (`Critical: subjectIsEmpty`). Go's stdlib does not set this automatically (golang/go#22249), so the CA must.

### To Reproduce

1. Internal CA (RSA) + policy allowing a hostname SAN + ACME enabled (EAB).
2. Enroll with a default client, e.g. Caddy:
   ```
   svc.example {
     tls { ca https://<infisical>/api/v1/pki/acme/<...>/directory
           eab <kid> <hmac> }
   }
   ```
3. Inspect the leaf:
   ```
   echo | openssl s_client -connect <host>:443 -servername svc.example 2>/dev/null \
     | openssl x509 -noout -subject -ext subjectAltName
   # subject=
   # X509v3 Subject Alternative Name:        <-- not marked critical
   #     DNS:svc.example
   ```
4. Open `https://svc.example` on macOS/Safari → rejected (`CSSMERR_TP_INVALID_CERTIFICATE` / "Invalid empty SubjectName"). A leaf with a CN from the same CA verifies fine — only the empty subject differs.

### Expected behavior

Given an empty-subject CSR, the CA issues an RFC 5280-conformant leaf: `subjectAltName` marked **critical** when the subject DN is empty.

### Additional context

- Confirmed on `infisical/infisical:v0.161.0` and current `main`.
- Introduced by #4920 (empty-CN ACME CSRs) without the RFC-required critical-SAN handling.
- RFC 5280 §4.1.2.6; §4.2.1.6.

molchalih:
## Context

Closes #6899.

The internal CA issued RFC 5280-invalid leaves on its CSR paths (ACME/EST/SCEP). When a CSR has no Common Name — the default for Caddy/certmagic, certbot, lego, acme.sh — the leaf got an **empty subject DN** with a **non-critical** `subjectAltName`. RFC 5280 §4.1.2.6 requires the SAN to be **critical** when the subject is empty; Apple's verifier (macOS/iOS) rejects the cert otherwise. `openssl`/curl/Go accept it (they read the SAN), which masked the bug.

Cause: `signCertFromCa` / `issueCertFromCa` built the SAN with the critical flag hardcoded `false`, with no empty-subject check. Same fix smallstep made in step-ca (smallstep/certificates#302, smallstep/crypto `a28955b`); Go's stdlib won't set it for you (golang/go#22249).

Fix: derive criticality from the resolved subject DN via a single helper, and apply it at both issuance sites. Also removes a duplicate `subjectAltName` push in the explicit-`altNames` branch of `signCertFromCa` (introduced in `e7a6f46`) that emitted the extension twice. Behavior is unchanged when the subject is non-empty.

```ts
export const createSubjectAltNameExtension = (altNames: TAltNameMapping[], subjectDn: string) =>
  new x509.SubjectAlternativeNameExtension(altNames, !subjectDn);
```

## Steps to verify the change

- `cd backend && npm run test:unit -- certificate-authority` — covers both the helper (empty subject ⇒ critical SAN; present ⇒ non-critical) and `signCertFromCa` end to end (real signing; asserts the issued leaf has an empty subject with a single, critical SAN, and a present subject with a non-critical SAN). Reverting the fix turns these red.
- Or issue via ACME with a CN-less CSR and inspect: `openssl x509 -noout -subject -ext subjectAltName` shows an empty subject with the SAN marked `critical`.

## Type

- [x] Fix

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/certificate-authority/certificate-authority-fns.test.ts
+++ b/backend/src/services/certificate-authority/certificate-authority-fns.test.ts
@@ -1,8 +1,15 @@
-import { describe, expect, it } from "vitest";
+import { webcrypto } from "node:crypto";
 
-import { CertKeyAlgorithm } from "@app/services/certificate/certificate-types";
+import * as x509 from "@peculiar/x509";
+import { beforeAll, describe, expect, it } from "vitest";
 
-import { buildCrlDistributionPointUrls, signatureAlgorithmToAlgCfg } from "./certificate-authority-fns";
+import { CertKeyAlgorithm, TAltNameType } from "@app/services/certificate/certificate-types";
+
+import {
+  buildCrlDistributionPointUrls,
+  createSubjectAltNameExtension,
+  signatureAlgorithmToAlgCfg
+} from "./certificate-authority-fns";
 
 // Helper to access properties on the union return type
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


molchalih:
Good catch. `issueCertificate`/`issueCertificateWithTemplate` always build the CSR with `CN=${commonName}`, so the subject is never empty on those paths and they can't hit the §4.1.2.6 violation today. Kept this PR scoped to the ACME/EST/SCEP regression — happy to fold the helper into those two sites as a follow-up so the invariant stays uniform.

JustinBack:
Is there any news on this PR? The ACME feature is still broken... Whats missing for the merge?

molchalih:
> Is there any news on this PR? The ACME feature is still broken... Whats missing for the merge?

that's a good question honestly. should we give a notice to maintainers via email? that's a tiny fix left unnoticed for almost a month already.

JustinBack:
> > Is there any news on this PR? The ACME feature is still broken... Whats missing for the merge?
> 
> 
> 
> that's a good question honestly. should we give a notice to maintainers via email? that's a tiny fix left unnoticed for almost a month already. 

It seems that PR's are ignored even after contacting the maintainers via Slack (being reassured it's taken a look at) and pinging them on the Repo (i.e. a PR of mine is open for 3 months now without any feedback on that matter). 

Guess we'll have to play the waiting game...

Issue #6854, opened 2026-06-14
Spurious "Commit Changes" modal after bulk tagging secrets on Overview page


flare19:
### Product

Secret Manager

### Describe the bug

After bulk-assigning tags to multiple secrets on the Overview page, hovering over an affected secret and clicking "View Tags" shows an empty tag list despite the bulk tag operation succeeding. This immediately triggers a spurious "Commit Changes" modal showing the tag as removed (before: tag1, after: -). If the user commits, the tag is permanently deleted server-side. If the user declines, the tag reappears correctly in the View Tags popover.

### To Reproduce

1. Create two or more secrets on the Overview page
2. Create a tag (e.g. tag1)
3. Enable batch mode
4. Select all secrets and click the add tags button in the selection panel
5. Add tag1 and apply — confirm the success notification appears
6. Immediately hover over one of the affected secrets and click View Tags
7. Observe that the tag list is empty despite the bulk tag succeeding
8. Observe the spurious "Commit Changes" modal appears showing tag1 as removed
9. Click Commit Changes — the tag is permanently deleted server-side

### Expected behavior

When bulk-assigning tags to secrets, the tags should appear immediately in the View Tags popover after the operation succeeds. No pending changes modal should be triggered, as the user has not made any manual edits to the affected secrets.

### Deployment Type

Self-hosted

jaydeep-pipaliya:
Hi, I've traced the root cause and would like to fix this.

Root cause: in `SecretTagForm.tsx`, the `useEffect` that calls `onTagsChange` fires on the initial mount (after the 500ms debounce). When a secret's tags are applied via bulk-tagging, the parent row's form is still initialized with `[]` from before the tags existed. React Hook Form freezes `defaultValues` at mount time and doesn't update them when props change. So the form's `watchedTags` stays at `[]` while `originalTagsRef` is updated to reflect the new server tags. When the user opens the tag popover, `SecretTagForm` mounts and—even without any user interaction—fires `onTagsChange([])` after 500ms. That sets the parent form's tags to `[]`, which the batch auto-apply timer detects as a change vs the server state and creates a pending change.

Fix: skip the first `onTagsChange` call in the effect (guard with `isFirstRun = useRef(true)`). On mount, the component should not propagate its initial form value upstream—only subsequent user interactions should. Assigning myself to implement this.

PR #6903, opened 2026-06-17
fix(secret-manager): skip initial onTagsChange on mount in batch mode tag form

jaydeep-pipaliya:
## What

Fixes a data-loss bug (#6854) where opening the "View Tags" popover on a secret in batch mode could immediately trigger a spurious "Commit Changes" modal showing the tag as removed -- even though the user made no changes. Confirming that modal deleted the tag server-side.

## Root cause

`SecretTagForm` runs a debounced `useEffect` that calls `onTagsChange` whenever `watchedTags` changes. The effect also fires once on initial mount with whatever value `watch("tags")` returns at that moment.

In the bug scenario:
1. The user bulk-assigns a tag to several secrets.
2. React Query invalidates and refetches the secret rows. `originalTagsRef` in the parent row is updated to reflect the new tag.
3. But React Hook Form freezes `defaultValues` at mount time, so the row's `watchedTags` form field stays at `[]` (its value from before the bulk operation).
4. When the user opens the tag popover, `SecretTagForm` mounts and -- after 500 ms -- fires `onTagsChange([])`.
5. The parent's `handleTagsChange` calls `setValue("tags", [])`, triggering the batch auto-apply timer.
6. The timer compares `watchedTags = []` against `originalTagsRef = [{id, slug}]`, detects a diff, and creates a pending change showing the tag as removed.

## Fix

Add an `isFirstRun` ref in `SecretTagForm` and skip the first execution of the effect. Only actual user interactions in the tag picker should propagate to the parent -- not the initial form state on mount.

## Test plan

- [ ] Enable batch mode on the Secret Manager overview page
- [ ] Create a tag and bulk-assign it to one or more secrets
- [ ] Immediately open the "View Tags" popover on an affected secret -- no "Commit Changes" modal should appear
- [ ] Manually add or remove a tag inside the popover -- the pending change should appear as expected
- [ ] Confirm bulk-assigning tags with no pending changes still works correctly

### DIFF
--- a/frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretTagForm.tsx
+++ b/frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretTagForm.tsx
@@ -91,9 +91,17 @@ export const SecretTagForm = ({
   // In batch mode, debounce tag changes to parent form
   const watchedTags = watch("tags");
   const debounceTimer = useRef<ReturnType<typeof setTimeout>>();
+  const isFirstRun = useRef(true);
 
   useEffect(() => {
     if (!isBatchMode) return () => {};
+
+    // Skip the initial mount so we don't propagate the form's defaultValues back
+    // to the parent. Only actual user interactions should trigger onTagsChange.
+    if (isFirstRun.current) {
+      isFirstRun.current = false;
+      return () => {};
+    }
 
     if (debounceTimer.current) {
       clearTimeout(debounceTimer.current);


jaydeep-pipaliya:
Good catch from Greptile — confirmed the parent is already covered. In `SecretEditTableRow.tsx`, `handleTagsChange` is wrapped in `useCallback` with `[setValue]` as the dependency:

```tsx
const handleTagsChange = useCallback(
  (newTags: { id: string; slug: string }[]) =>
    setValue('tags', newTags.map((t) => ({ id: t.id, slug: t.slug }))),
  [setValue]
);
```

React Hook Form's `setValue` is stable across renders, so `handleTagsChange` (and therefore `onTagsChange` in `SecretTagForm`) maintains a stable reference and won't re-trigger the effect after the first mount.

jaydeep-pipaliya:
Closing this. I missed that @flare19 already opened #6855 against the same issue. Sorry for the noise.

Issue #6643, opened 2026-05-28
intToString rune conversion breaks IncrementBy in keystore


R0hit-Sriram:
### Product

Platform/Access controls/Other

### Describe the bug

string(rune(65)) returns "A", not "65". In IncrementBy, the result of incrementing an integer is converted to a string using rune conversion, producing a letter instead of a digit. The downstream stringToInt parser returns 0 for any non-digit character, so IncrementBy is fully broken for any numeric value.

### To Reproduce

1. Call IncrementBy on any keystore key with a numeric value
2. Read the value back
3. The value is 0 instead of the incremented number

### Expected behavior

IncrementBy should increment the stored integer and persist the correct string representation of the new value.

### Deployment Type

Self-hosted

### Additional context

backend-go/internal/keystore/memory.go:131. Found via Faultmark (faultmark.com), an AI code scanner that uses a multi-model debate to verify bugs against actual source code before reporting them.

ssshreyans26:
I've submitted a fix for this in PR #6682. The one-line change replaces `string(rune(n))` with `strconv.FormatInt(n, 10)` in `intToString`, and adds unit tests covering `IncrementBy` (new key, existing key, multiple increments, negative values) and a round-trip test for `intToString`/`stringToInt`. Lint and unit tests pass locally.

jaydeep-pipaliya:
Hi, I can reproduce this - `string(rune(65))` in Go returns `"A"` (the Unicode character at codepoint 65), not `"65"`. So any call to `IncrementBy` stores the wrong string and subsequent reads parse it back as 0.

The fix is one line: replace `string(rune(n))` with `strconv.FormatInt(n, 10)` in the `intToString` helper. I'll take this.

PR #6904, opened 2026-06-17
fix(keystore): use strconv.FormatInt instead of rune cast in intToString

jaydeep-pipaliya:
## What

Fixes a bug (#6643) where `IncrementBy` in the in-memory keystore silently persisted the wrong value for any input.

## Root cause

`intToString` used `string(rune(n))` to convert an integer to a string. In Go, `string(rune(n))` converts the integer to its Unicode code-point character, not its decimal representation. For example `string(rune(65))` returns `"A"`, not `"65"`. So `IncrementBy` would compute the correct new value (e.g. 5), return it from the function, but store `"E"` (U+0005... actually the control character at codepoint 5) in the map. The next call reads that back through `stringToInt`, which skips non-digit bytes and returns 0, so all subsequent increments start from 0 again.

## Fix

Replace `string(rune(n))` with `strconv.FormatInt(n, 10)`, which produces the correct decimal string.

## Test plan

- [ ] Verify the memory keystore `IncrementBy` returns incrementally correct values across multiple calls
- [ ] Existing integration tests in `tests/platform/` should continue to pass

### DIFF
--- a/backend-go/internal/keystore/memory.go
+++ b/backend-go/internal/keystore/memory.go
@@ -2,6 +2,7 @@ package keystore
 
 import (
 	"context"
+	"strconv"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Closing this. I missed that @ssshreyans26 already opened #6682 against the same issue with the same fix. Sorry for the noise.

Issue #6864, opened 2026-06-15
decryptInternalKmsKey reads from replica inside an open primary transaction


flare19:
### Product

KMS

### Describe the bug

In backend-go/internal/services/kms/, both ensureProjectDataKey and ensureOrgDataKey open a primary transaction, acquire an advisory lock, and re-check whether a KMS key exists via findProjectKmsInfo(ctx, tx, ...) / findOrgKmsInfo(ctx, tx, ...) — correctly using the transaction handle.
However, in the else branch (key already exists), both functions call decryptInternalKmsKey which hardcodes s.db.Replica().QueryRow() internally and accepts no tx/querier parameter. This means the read escapes the transaction and hits the replica.
Under replication lag, a concurrent request that just created the KMS key on the primary will cause decryptInternalKmsKey to return pgx.ErrNoRows on the replica, producing a Failed to find KMS key error and a 500 for the user.
The inconsistency is visible by comparison — findProjectKmsInfo and findOrgKmsInfo both accept pg.Querier specifically to support both transactional and non-transactional callers. decryptInternalKmsKey was never given the same treatment.

### To Reproduce

Two concurrent requests hitting getProjectDataKey or getOrgDataKey for the same project/org for the first time, under any replication lag.

### Expected behavior

decryptInternalKmsKey should use the caller's transaction handle so it reads from the same primary connection that created the KMS key.

### Deployment Type

Self-hosted

jaydeep-pipaliya:
Hi, I've verified the bug in the code. In both `ensureProjectDataKey` and `ensureOrgDataKey`, a primary transaction (`tx`) is opened and advisory locks are acquired before re-checking whether the KMS key exists via `findProjectKmsInfo(ctx, tx, ...)` / `findOrgKmsInfo(ctx, tx, ...)` -- correctly using the transaction. But in the else branch (key exists), the call to `decryptInternalKmsKey` hardcodes `s.db.Replica().QueryRow()` and never accepts a `pg.Querier` parameter, so the read escapes the transaction.

The fix is to add a `pg.Querier` parameter to `decryptInternalKmsKey` (matching the existing pattern from `findProjectKmsInfo` / `findOrgKmsInfo`) and have the two callers pass their `tx`. Taking this.

PR #6905, opened 2026-06-17
fix(kms): pass primary transaction to decryptInternalKmsKey to avoid replica lag

jaydeep-pipaliya:
## What

Fixes a race condition (#6864) where `decryptInternalKmsKey` reads from the replica inside an open primary transaction, causing intermittent 500 errors under replication lag.

## Root cause

Both `ensureProjectDataKey` and `ensureOrgDataKey` open a primary transaction and acquire advisory locks before re-checking whether a KMS key exists:

```go
project, err := s.findProjectKmsInfo(ctx, tx, projectID)  // correctly uses tx
```

`findProjectKmsInfo` and `findOrgKmsInfo` both accept a `pg.Querier` parameter and correctly receive the transaction. In the else branch (key already exists), however, the call to `decryptInternalKmsKey` hardcodes `s.db.Replica().QueryRow()` and was never given the same treatment:

```go
// before fix -- escapes the transaction
kmsKeyMaterial, err = s.decryptInternalKmsKey(ctx, project.KmsSecretManagerKeyID.V)
```

Under any replication lag, a concurrent request that just created the KMS key on the primary passes the `findProjectKmsInfo` check (read from the locked primary tx), then hits the replica in `decryptInternalKmsKey` and gets `pgx.ErrNoRows` -- producing a `Failed to find KMS key` 500.

## Fix

Add a `pg.Querier` parameter to `decryptInternalKmsKey`, matching the existing pattern from `findProjectKmsInfo` / `findOrgKmsInfo`. Both callers now pass their open primary transaction so all reads within the ensure function stay on the same connection.

```go
// after fix
kmsKeyMaterial, err = s.decryptInternalKmsKey(ctx, tx, project.KmsSecretManagerKeyID.V)
```

## Test plan

- [ ] Two concurrent requests for the same project/org data key on first use should not produce a 500 under replica lag
- [ ] Existing KMS integration tests in `tests/platform/kms/` should continue to pass

### DIFF
--- a/backend-go/internal/services/kms/kms.go
+++ b/backend-go/internal/services/kms/kms.go
@@ -248,7 +248,8 @@ func (s *Service) generateEncryptedKeyMaterial() ([]byte, error) {
 
 // decryptInternalKmsKey loads a KMS key record and decrypts its key material using the root key.
 // For internal KMS keys only. Returns error for external KMS keys.
-func (s *Service) decryptInternalKmsKey(ctx context.Context, kmsKeyID uuid.UUID) ([]byte, error) {
+// q must be the same connection/transaction used by the caller to avoid replica lag.
+func (s *Service) decryptInternalKmsKey(ctx context.Context, q pg.Querier, kmsKeyID uuid.UUID) ([]byte, error) {
 	query := `
 		SELECT
 			kmsKey.id,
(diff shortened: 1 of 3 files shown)


jaydeep-pipaliya:
Closing this. I missed that @flare19 already opened #6865 against the same issue. Sorry for the noise.

Issue #6907, opened 2026-06-17
feat(tls-cert-auth): validate SAN alongside CN (enables SPIFFE X.509-SVID auth without JWT-SVID)


sinnwise:
**Product:** Machine Identities (Auth)

**Deployment Type:** Self-hosted

### Feature

Extend **TLS Certificate Auth** (`/api/v1/auth/tls-cert-auth/login`) to optionally validate the client certificate's **Subject Alternative Name (SAN)** alongside the existing Common Name (CN) check.

Today the auth method only matches the cert's Subject CN against `allowedCommonNames` (see `identity-tls-cert-auth-service.ts`, where `subjectDetails.CN` is compared to the allow-list). There is no way to authorize based on a SAN entry.

### Motivation: SPIFFE X.509 without JWT-SVID

In SPIFFE/SPIRE X.509 deployments where JWT-SVID is **not** enabled across the estate, workloads only have X.509-SVIDs. A SPIFFE X.509-SVID carries its identity in the **SAN URI** field (`spiffe://trust-domain/path`) and leaves the Subject CN empty — so the current CN-only check cannot authorize these workloads.

With SAN validation, an X.509 SPIFFE setup could:
1. Provide the SPIRE trust bundle / roots to Infisical (as the trusted CA for TLS Cert Auth).
2. Use TLS Cert Auth where the workload presents its X.509-SVID fetched from the SPIRE agent socket.
3. Authorize on the **SAN URI** (the SPIFFE ID), completing service identity validation without needing JWT-SVID.

### Proposed behaviour

- Add an optional `allowedSubjectAlternativeNames` (or SAN-type-aware) allow-list to the TLS Cert Auth identity config.
- During login, match against the cert's SAN entries (URI / DNS / IP) in addition to / instead of CN.
- Keep CN validation as-is for backwards compatibility; SAN validation is additive and opt-in.

### Additional context

- Existing CN check: `backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-service.ts`.
- Complements the existing SPIFFE JWT-SVID auth (#5610) by covering X.509-only SPIRE estates.

PR #6908, opened 2026-06-17
feat(tls-cert-auth): validate SAN alongside CN (enables SPIFFE X.509-SVID auth without JWT-SVID)

sinnwise:
## Context

Extends **TLS Certificate Auth** to optionally validate the client certificate's **Subject Alternative Name (SAN)** alongside the existing Common Name (CN) check.

Today the auth method only matches the certificate's Subject CN against `allowedCommonNames`. This adds an optional `allowedSubjectAltNames` allow-list. When both are configured, the certificate must satisfy both checks; either may be left empty.

### SAN matching behaviour

Matching is **type-aware** — an allow-list entry only matches a certificate SAN of the same type, so a configured value can never satisfy a SAN of a different type:

- **Type-prefixed entries** match the corresponding SAN type, e.g. `URI:spiffe://example.org/svc`, `IP:10.0.0.1`, `DNS:svc.example.com`.
- **Bare entries** (no prefix) are treated as **DNS** names, e.g. `svc.example.com`.
- **DNS name matching is case-insensitive** (per [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#:~:text=When%20comparing%20DNS,the%20indicated%20subtree.)); URI/IP values are matched exactly.

### Motivating use case

A common driver is workload identity where the certificate carries its identity in a **URI SAN** (e.g. a SPIFFE ID `spiffe://trust-domain/path`) with an empty Subject CN — CN-only matching cannot authorize these workloads. With SAN validation you can supply the issuing trust bundle as the CA certificate and authorize on `URI:spiffe://...`.

Tracks issue #6907.

## Changes

- **DB:** migration + schema for a new nullable `allowedSubjectAltNames` `text` column on `identity_tls_cert_auths`.
- **Service / fns:** `parseCertificateSubjectAltNames` normalizes the certificate's SANs into canonical `type:value` tokens; `isSubjectAltNameAllowed` performs type-aware, case-correct matching against the allow-list during login. Denials use a dedicated `subject_alt_name_not_allowed` reason code. Persisted on attach/update.
- **Router / api-docs / audit log:** new field wired through attach + update bodies and audit metadata.
- **Frontend:** types, mutations, auth form input, and read-only view.
- **Docs:** `tls-cert-auth.mdx` documents the field and the type-aware matching rules.
- **Tests:** unit tests for SAN parsing and matching (type collision, case-insensitivity, bare-vs-prefixed entries).

## Type

- [ ] Fix
- [X] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Steps to verify the change

1. Configure TLS Cert Auth on an identity with the issuing CA as the CA certificate and `allowedSubjectAltNames` set to e.g. `URI:spiffe://example.org/svc`.
2. Authenticate with a client certificate whose URI SAN matches → succeeds (`200`, Bearer token).
3. Authenticate with a certificate whose SAN does not match (wrong value, or a same-string SAN of a different type) → `401` with reason code `subject_alt_name_not_allowed`.

## Checklist

- [X] Title follows the conventional commit format.
- [X] Tested locally
- [X] Updated docs (if needed)
- [X] Read the contributing guide


sinnwise:
Completed local testing against a local dev stack (type-aware SAN matching).

**Setup:** Root → Intermediate → Leaf chain via `openssl`. Leaf has no CN — identity carried in the URI SAN (`spiffe://example.org/test-svc`). Intermediate supplied as the CA certificate; no CN allow-list.

| # | Allowed SAN entry | Certificate SAN | Expected | Result |
|---|---|---|---|---|
| 1 | `URI:spiffe://example.org/test-svc` | `URI:spiffe://example.org/test-svc` | allow | **200** |
| 2 | `spiffe://example.org/test-svc` (bare → DNS) | `URI:spiffe://example.org/test-svc` | deny (type mismatch) | **401** |
| 3 | `URI:spiffe://example.org/other` | `URI:spiffe://example.org/test-svc` | deny | **401** |
| 4 | `SVC.EXAMPLE.COM` (DNS) | URI SAN only | deny | **401** |
| 5 | `svc.example.com` (DNS) | `DNS:SVC.EXAMPLE.COM` | allow (case-insensitive) | **200** |

Confirms:
- **Type-aware matching** — a bare/DNS-defaulted value cannot satisfy a URI SAN (no cross-type collision).
- **Case-insensitive DNS** — `svc.example.com` matches `DNS:SVC.EXAMPLE.COM`.
- Denials return `401` with `reasonCode: subject_alt_name_not_allowed`.

Also verified:
- Migration applies cleanly — `allowedSubjectAltNames` is a `text` column (no length cap) in `identity_tls_cert_auths`.
- Unit tests: 27/27 passing.

sinnwise:
Resolving as complete in another PR. 😊

sinnwise:
https://github.com/Infisical/infisical/pull/6957

PR #6929, opened 2026-06-18
fix(approval-policy): allow user added through groups

adilsitos:
## Context
A user that is added into a project by a group cannot be added as a single approver on secrets and access policies. This was happening because the check only validated users, and not users by group memberships. 

There was also another bug that was fixed on this PR. group membership were not being checked when creating an approval policy, which allowed the addition of groups as approvers that did not belong to a project. 

### Bug 1: User added into a project through group 
- create an approval policy and select the approver as an user that belong to a project by a group 
- check that this can be created (it was causing an error earlier) 

### Bug 2: add a group that does not belong to a project as approver 
- (the ui does not show groups that don't belong to a project, but this can be done using API)
- try to create an approval policy using an external org. 
- check that this fails  

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally
- [  ] Updated docs (if needed)


backend/src/ee/services/access-approval-policy/access-approval-policy-service.ts
```diff
@@ -83,20 +90,10 @@ export const accessApprovalPolicyServiceFactory = ({
...
-
-    if (projectMemberships.length !== userIds.length) {
-      const projectMemberUserIds = new Set(projectMemberships.map((member) => member.userId));
-      const userIdsNotInProject = userIds.filter((id) => !projectMemberUserIds.has(id));
-      throw new BadRequestError({
-        message: `Some users are not members of the project: ${userIdsNotInProject.join(", ")}`
-      });
-    }
-  };
+  const { verifyProjectUserMembership, verifyProjectGroupMembership } = approvalPolicyMembershipVerifierFactory({
```
akhilmhdh: Overall looking good but one question i have is why not something like `findEffectiveOrgMembership` in `org-dal`. Like that would precisely compute a given user or users memberships available in a suborganization as example which could be through group or direct membership

adilsitos: I didn't know about this function. Will change to use it.

akhilmhdh (approved):

PR #6934, opened 2026-06-18
improvement(project-role): add endpoint to get project role by id

adilsitos:
## Context
Add a new endpoint to get project role by ID. This was added to support the project_role import using terraform  

## Steps to verify the change
- Create a new project role inside infisical 
- Use the following curl to get a project role: 
```
OCcurl -X GET "http://localhost:4000/api/v1/projects/roles/<ROLE_ID>" \
  -H "Authorization: Bearer <YOUR_TOKEN>" \
  -H "Content-Type: application/json"
```

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/routes/v1/project-role-router.ts
+++ b/backend/src/ee/routes/v1/project-role-router.ts
@@ -354,7 +354,7 @@ export const registerProjectRoleRouter = async (server: FastifyZodProvider) => {
         }
       ],
       params: z.object({
-        roleId: z.string().trim().uuid().describe("The ID of the role")
+        roleId: z.string().trim().uuid().describe(PROJECT_ROLE.GET_ROLE_BY_ID.roleId)
       }),
       response: {
         200: z.object({
--- a/backend/src/ee/routes/v1/project-role-router.ts
+++ b/backend/src/ee/routes/v1/project-role-router.ts
@@ -347,6 +347,12 @@ export const registerProjectRoleRouter = async (server: FastifyZodProvider) => {
     schema: {
       hide: false,
       tags: [ApiDocsTags.ProjectRoles],
+      description: "Get a project role by ID",
+      security: [
+        {
+          bearerAuth: []
+        }
+      ],
       params: z.object({
         roleId: z.string().trim().describe(PROJECT_ROLE.GET_ROLE_BY_ID.roleId)
       }),
--- a/backend/src/ee/routes/v1/project-role-router.ts
+++ b/backend/src/ee/routes/v1/project-role-router.ts
@@ -340,6 +340,37 @@ export const registerProjectRoleRouter = async (server: FastifyZodProvider) => {
 
   server.route({
     method: "GET",
+    url: "/roles/:roleId",
+    config: {
+      rateLimit: readLimit
+    },
+    schema: {
+      hide: false,
+      tags: [ApiDocsTags.ProjectRoles],
+      params: z.object({
+        roleId: z.string().trim().describe(PROJECT_ROLE.GET_ROLE_BY_ID.roleId)
+      }),
+      response: {
+        200: z.object({
+          role: SanitizedRoleSchema
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const role = await server.services.role.getProjectRoleById({
+        permission: req.permission,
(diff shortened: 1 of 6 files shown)


backend/src/ee/routes/v1/project-role-router.ts
```diff
@@ -338,6 +338,43 @@ export const registerProjectRoleRouter = async (server: FastifyZodProvider) => {
...
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const role = await server.services.role.getProjectRoleById({
+        permission: req.permission,
+        selector: {
+          id: req.params.roleId
+        }
+      });
+
+      return { role: { ...role, projectId: role.projectId } };
```
varonix0: Why is this needed? Why not just `return role`

backend/src/ee/routes/v1/project-role-router.ts
```diff
@@ -338,6 +338,43 @@ export const registerProjectRoleRouter = async (server: FastifyZodProvider) => {
...
+      hide: false,
+      tags: [ApiDocsTags.ProjectRoles],
+      description: "Get a project role by ID",
+      security: [
+        {
+          bearerAuth: []
+        }
+      ],
+      params: z.object({
+        roleId: z.string().trim().describe(PROJECT_ROLE.GET_ROLE_BY_ID.roleId)
```
varonix0: Role ID's can only be uuid's, lets enforce UUID on the roleId input

backend/src/services/role/role-service.ts
```diff
@@ -243,6 +244,35 @@ export const roleServiceFactory = ({
...
+      throw new NotFoundError({ message: `Role with id ${selector.id} not found` });
+    }
+
+    const scopeData = {
+      scope: AccessScope.Project,
+      orgId: permission.orgId,
+      projectId: role.projectId
+    } as const;
+
+    await scopeFactory[AccessScope.Project].onGetRoleByIdGuard({ permission, scopeData, selector });
```
varonix0: Why aren't we just calling the `getRoleById` function directly? Seems like there's a lot of logic in there that we aren't replicating here, I think we could just call the `getRoleById` and use the new `getProjectRoleById` function as a wrapper to get the details we need before calling `getRoleById`

varonix0 (changes requested):

backend/src/services/role/role-service.ts
```diff
@@ -243,6 +244,35 @@ export const roleServiceFactory = ({
...
+
+    return { ...role, projectId: role.projectId, permissions: unpackPermissions(role.permissions) };
```
varonix0: And same here, I think we could just do `return {...role, permissions: unpackPermissions(..)`}

varonix0 (changes requested):

backend/src/ee/routes/v1/project-role-router.ts
adilsitos: This is the same output we have for other endpoints that fetch project role. We need to have the `projectId: role.projectId` because this is optional on the type that is used. Fixed this, so we return the right type from the service.

backend/src/services/role/role-service.ts
adilsitos: I just changed the return to be the output of the `getRoleById` with projectId, so the type is correct.

backend/src/services/role/role-service.ts
adilsitos: fixed it.

backend/src/ee/routes/v1/project-role-router.ts
```diff
@@ -354,7 +354,7 @@ export const registerProjectRoleRouter = async (server: FastifyZodProvider) => {
         }
       ],
       params: z.object({
-        roleId: z.string().trim().describe(PROJECT_ROLE.GET_ROLE_BY_ID.roleId)
+        roleId: z.string().trim().uuid().describe("The ID of the role")
```
varonix0: Why did you remove the `PROJECT_ROLE.GET_ROLE_BY_ID.roleId`?

varonix0 (changes requested):

varonix0:
@veria

varonix0 (approved):

PR #6938, opened 2026-06-19
feat: add HSM connector to PKI code signing

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/db/migrations/20260618143217_pki-hsm-connector.ts
```diff
@@ -0,0 +1,205 @@
+import { Knex } from "knex";
+
+import { TableName } from "@app/db/schemas";
+import { createOnUpdateTrigger, dropOnUpdateTrigger } from "@app/db/utils";
+
+const HSM_CONNECTORS_GATEWAY_XOR_CHECK = "hsm_connectors_gateway_xor_chk";
+const CERTS_KEY_SOURCE_CHECK = "certificates_key_source_chk";
```
sheensantoscapadngan: nit - we can really just say "chk" here

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -68,6 +69,12 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
     },
     schema: {
       operationId: "gatewayHeartbeat",
+      body: z
+        .object({
+          capabilities: z.record(z.unknown()).optional()
+        })
```
sheensantoscapadngan, line 79: let's define the schema here rather than using unknown with passthrough

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -974,12 +974,24 @@ export const gatewayV2ServiceFactory = ({
...
+    orgPermission,
+    capabilities
+  }: {
+    orgPermission: OrgServiceActor;
+    capabilities?: Record<string, unknown>;
+  }) => {
+    const persistCapabilities = async (gatewayId: string) => {
+      if (capabilities === undefined) return;
+      await gatewayV2DAL.updateById(gatewayId, { capabilities });
+    };
```
sheensantoscapadngan: why do we define this method inside the heartbeat?

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -974,12 +974,24 @@ export const gatewayV2ServiceFactory = ({
...
-  const heartbeat = async ({ orgPermission }: { orgPermission: OrgServiceActor }) => {
+  const heartbeat = async ({
+    orgPermission,
+    capabilities
+  }: {
+    orgPermission: OrgServiceActor;
+    capabilities?: Record<string, unknown>;
+  }) => {
+    const persistCapabilities = async (gatewayId: string) => {
+      if (capabilities === undefined) return;
```
sheensantoscapadngan: if it's undefined, shouldn't we update capabilities to be empty?

backend/src/server/routes/v1/cert-manager-hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+  description: z.string().nullable(),
+  projectId: z.string(),
+  gatewayId: z.string().uuid().nullable(),
+  gatewayPoolId: z.string().uuid().nullable(),
+  slotLabel: z.string(),
+  keyNamePrefix: z.string().nullable(),
+  createdAt: z.date(),
+  updatedAt: z.date()
+});
+
```
sheensantoscapadngan: this isn't necessarily for cert manager right? asking because we've placed it in a cert-manager specific file

backend/src/server/routes/v1/cert-manager-hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+  .refine(
+    (data) => {
+      if (data.gatewayId === undefined && data.gatewayPoolId === undefined) return true;
+      return gatewayPickRefiner({
+        gatewayId: data.gatewayId ?? null,
+        gatewayPoolId: data.gatewayPoolId ?? null
+      });
+    },
+    { message: GATEWAY_PICK_MESSAGE, path: ["gatewayId"] }
+  );
```
sheensantoscapadngan: same remark here.. also shouldn't we make the names more specific to prevent ambiguity?

backend/src/server/routes/v1/hsm-connector-router.ts
```diff
@@ -0,0 +1,239 @@
+import { z } from "zod";
```
sheensantoscapadngan, line 2: same remark here about the filename. this isn't specific to certificate manager

backend/src/server/routes/v1/cert-manager-hsm-connector-router.ts
```diff
@@ -0,0 +1,239 @@
...
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const connector = await server.services.hsmConnector.getHsmConnectorById(
+        { connectorId: req.params.connectorId },
+        req.permission
+      );
+      const result = await server.services.hsmConnector.deleteHsmConnector(
+        { connectorId: req.params.connectorId },
+        req.permission
+      );
```
sheensantoscapadngan: shouldn't we just be doing getConnectorById inside the deleteHsmConnector method and return the full connector from there?

backend/src/server/routes/v1/cert-manager-hsm-connector-router.ts
```diff
@@ -0,0 +1,239 @@
...
+          type: EventType.TEST_HSM_CONNECTOR,
+          metadata: {
+            connectorId: req.params.connectorId,
+            name: connector.name,
+            ok: result.ok,
+            memberCount: result.members.length
+          }
+        }
+      });
+      return result;
```
sheensantoscapadngan: same remark here... I think the pattern everywhere else is 1 service method invocation per router

backend/src/server/routes/v1/cert-manager-hsm-connector-router.ts
```diff
@@ -0,0 +1,239 @@
...
+    }
+  });
+
+  server.route({
+    method: "GET",
+    url: "/:connectorId/linked-certificates",
```
sheensantoscapadngan: perhaps we make this generic ahead of time and just say "linked-resources" since this could potentially be used in other areas outside of certificate manager?

backend/src/server/routes/v1/index.ts
```diff
@@ -240,6 +241,7 @@ export const registerV1Routes = async (server: FastifyZodProvider) => {
       await pkiRouter.register(registerCertificateCleanupRouter, { prefix: "/certificate-cleanup" });
       await pkiRouter.register(registerCertificateInventoryViewRouter, { prefix: "/certificate-inventory-views" });
       await pkiRouter.register(registerCertManagerAccessRouter, { prefix: "/access" });
+      await pkiRouter.register(registerCertManagerHsmConnectorRouter, { prefix: "/hsm-connectors" });
```
sheensantoscapadngan: hmm.. shouldn't this live outside of PKI ?

backend/src/services/certificate-authority/azure-ad-cs/azure-ad-cs-certificate-authority-fns.ts
```diff
@@ -1112,7 +1112,7 @@ export const AzureAdCsCertificateAuthorityFns = ({
     isCancelled
   }: {
     caId: string;
-    profileId: string;
+    profileId?: string;
```
sheensantoscapadngan, line 1115: why does this become optional? seems a bit off for profileId not to be passed in when method is specifically for issuing using a profile

backend/src/services/hsm-connector/hsm-connector-dal.ts
```diff
@@ -0,0 +1,71 @@
...
+          .clone()
+          .select("id", "commonName", "status", "notAfter", "hsmKeyLabel", "createdAt")
+          .orderBy("createdAt", "desc")
+          .limit(limit)
+          .offset(offset),
+        baseQuery.clone().count<{ count: string }[]>("id as count").first()
+      ]);
+
+      return {
+        rows: rows.map((r) => ({
```
sheensantoscapadngan: maybe this should be "certificates" instead?

backend/src/services/hsm-connector/hsm-connector-fns.ts
```diff
@@ -0,0 +1,81 @@
...
+}): Promise<THsmConnectorCredentials> => {
+  const { decryptor } = await kmsService.createCipherPairWithDataKey({
+    type: KmsDataKey.SecretManager,
+    projectId
+  });
+  const plainText = decryptor({ cipherTextBlob: encryptedCredentials });
+  return HsmConnectorCredentialsSchema.parse(JSON.parse(plainText.toString()));
+};
+
+export const sanitizeHsmConnector = (args: {
```
sheensantoscapadngan: hmm do we still need this? we usually define a sanitized zod schema instead rater than doing it like this... or maybe i'm misunderstanding this

backend/src/services/hsm-connector/hsm-connector-routing.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+    if (!connector.gatewayPoolId) {
+      throw new BadRequestError({ message: "Connector has neither gatewayId nor gatewayPoolId set." });
+    }
+    const healthy = await gatewayPoolService.listHealthyGateways(connector.gatewayPoolId);
+    const capable = healthy.filter((g) => isPkcs11Capable(g.capabilities) && !triedGatewayIds.has(g.id));
+    if (capable.length === 0) {
+      throw new BadRequestError({
+        message:
+          "No healthy PKCS#11-enabled gateway in the pool. Start at least one gateway with --pkcs11-module pointing at the HSM."
```
sheensantoscapadngan: at the HSM client files right

backend/src/services/hsm-connector/hsm-connector-service.ts
```diff
@@ -0,0 +1,457 @@
...
+  };
+
+  const assertLicense = async (orgId: string) => {
+    const plan = await licenseService.getPlan(orgId);
+    if (!plan.pkiHsm) {
+      throw new BadRequestError({
+        message: "Your plan does not include HSM-backed code signing. Please upgrade to use HSM Connectors."
+      });
+    }
+  };
```
sheensantoscapadngan, line 90: would it make sense for us to just use the existing hsm license flag feature since this isn't really specific to PKI?

backend/src/services/hsm-connector/hsm-connector-service.ts
```diff
@@ -0,0 +1,457 @@
...
+    action: ProjectPermissionHsmConnectorActions
+  ) => {
+    await assertLicense(actor.orgId);
+    const { permission } = await permissionService.getProjectPermission({
+      actor: actor.type,
+      actorId: actor.id,
+      projectId,
+      actorAuthMethod: actor.authMethod,
+      actorOrgId: actor.orgId,
+      actionProjectType: ActionProjectType.CertificateManager
```
sheensantoscapadngan: this isn't specific to certificate manager

backend/src/services/hsm-connector/hsm-connector-service.ts
```diff
@@ -0,0 +1,457 @@
...
+        gatewayPoolId: dto.gatewayPoolId ?? null,
+        encryptedCredentials
+      });
+    } catch (err) {
+      // 23505 = unique constraint violation: UNIQUE(projectId, name) on hsm_connectors.
+      const cause = (err as { error?: { code?: string } })?.error;
+      if (cause?.code === "23505") {
+        throw new BadRequestError({
+          message: `An HSM Connector named "${dto.name}" already exists in this project.`
+        });
```
sheensantoscapadngan, line 165: no need to specifically mention "project" here since it would be invalid for PKI right ?

backend/src/services/hsm-connector/hsm-connector-service.ts
```diff
@@ -0,0 +1,457 @@
...
+    await assertProjectPermission(actor, row.projectId, ProjectPermissionHsmConnectorActions.Delete);
+
+    const refCount = await hsmConnectorDAL.countReferencingCertificates(row.id);
+    if (refCount > 0) {
+      throw new BadRequestError({
+        message: `HSM Connector is in use by ${refCount} certificate(s). Open the connector to see the linked certificates panel and re-issue or retire them first.`
+      });
+    }
+
+    await hsmConnectorDAL.deleteById(row.id);
```
sheensantoscapadngan: this should be ran inside a txn

backend/src/services/hsm-connector/hsm-connector-service.ts
```diff
@@ -0,0 +1,457 @@
...
+      offset: dto.offset,
+      limit: dto.limit
+    });
+    return { linkedCertificates: rows, totalCount };
+  };
+
+  const testHsmConnector = async (
+    dto: TTestHsmConnectorDTO,
+    actor: THsmConnectorServiceActor
+  ): Promise<THsmConnectorTestResult> => {
```
sheensantoscapadngan: any reason why we have to explicitly mention the return schema?

backend/src/services/signer/signer-issuance-fns.ts
```diff
@@ -173,3 +177,232 @@ export const issueSignerCertificate = async (
...
+};
+
+export const renewHsmBackedSignerCertificate = async (
+  deps: THsmBackedRenewDeps,
+  input: THsmBackedRenewInput
+): Promise<{ certificateId: string; hsmKeyLabel: string }> => {
+  const ca = await deps.certificateAuthorityDAL.findByIdWithAssociatedCa(input.caId);
+  if (!ca || ca.projectId !== input.projectId) {
+    throw new BadRequestError({ message: `Certificate authority '${input.caId}' is not in this project.` });
+  }
```
sheensantoscapadngan, line 336: project mentioned

backend/src/services/signer/signer-issuance-fns.ts
```diff
@@ -173,3 +177,232 @@ export const issueSignerCertificate = async (
...
+    connectorId: input.hsmConnectorId,
+    projectId: input.projectId,
+    keyLabel: input.hsmKeyLabel
+  });
+  if (input.expectedPublicKeySpkiDer) {
+    const stored = input.expectedPublicKeySpkiDer;
+    if (stored.length !== publicKeySpkiDer.length || !stored.equals(publicKeySpkiDer)) {
+      throw new BadRequestError({
+        message:
+          "HSM returned a different public key than the one recorded at issuance. Renewal aborted; investigate the HSM and the Gateway."
```
sheensantoscapadngan: maybe we could mention when this scenario could happen so that we're able to point them to the right direction

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -122,6 +137,74 @@ export const signerIssuanceServiceFactory = ({
 
     await signerIssuanceJobDAL.cancelOpenForSigner(input.signerId, "Superseded by a newer issuance request.");
 
+    if (input.hsm) {
+      if (input.hsm.actor) {
+        await hsmConnectorService.assertAttachPermission(input.hsm.actor, input.hsm.hsmConnectorId, input.projectId);
+      }
+      const reuseExistingKey = Boolean(input.hsm.hsmKeyLabel && input.hsm.hsmPublicKeySpki);
```
sheensantoscapadngan, line 142: hmm I wonder if we should allow people to do this.. wouldn't this result in very sensitive keys being used? curious to know what other platforms are doing here

backend/src/services/signer/signer-issuance-service.ts
```diff
@@ -122,6 +137,74 @@ export const signerIssuanceServiceFactory = ({
...
+        hsmKeyLabel: keyLabel,
+        hsmPublicKeySpki: publicKeySpkiDer,
+        nextPollAt: new Date()
+      });
+
+      void (async () => {
+        try {
+          await processJob(hsmJob.id);
+        } catch (err) {
+          logger.warn(
```
sheensantoscapadngan: wouldn't we want this to be inside a txn?

backend/src/services/signer/signer-issuance-service.ts
sheensantoscapadngan: same here... shouldn't these be in the same txn?

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -63,8 +63,11 @@ In **Certificate Manager → Code Signing → Signers**, click **Create Signer**
     | Field | Description |
     |-------|-------------|
     | **Certificate Authority** | The CA that issues the certificate. Internal CAs issue immediately. External CAs (AWS Private CA, Azure AD CS) issue asynchronously and the Signer enters **Pending** until the cert lands. |
+    | **Key source** | Where the leaf key for this Signer is generated. **Infisical-managed** (default) lets Infisical generate and manage the key for you. **HSM-backed** generates the key inside your own HSM through an [HSM Connector](/documentation/platform/pki/settings/hsm-connectors). |
     | **Common Name** | The legal name shown on the certificate, for example `Acme Mobile, Inc.`. Fixed once the certificate is issued. To change it you must reissue. |
-    | **Key algorithm** | The key for the signing certificate: `RSA-2048`, `RSA-3072`, `RSA-4096`, `ECDSA P-256`, `ECDSA P-384`, or `ECDSA P-521`. |
+    | **Key algorithm** | The key for the signing certificate: `RSA-2048`, `RSA-3072`, `RSA-4096`, `ECDSA P-256`, `ECDSA P-384`, or `ECDSA P-521`. Shown only when **Key source = Infisical-managed**. |
```
sheensantoscapadngan: why is this only for when key source is Infisical-managed ?

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -63,8 +63,11 @@ In **Certificate Manager → Code Signing → Signers**, click **Create Signer**
...
+    | **HSM Connector** | The Connector that decides which Gateway routes operations to your HSM. Shown only when **Key source = HSM-backed**. See [HSM Connectors](/documentation/platform/pki/settings/hsm-connectors). |
```
sheensantoscapadngan: hmm it's not ideal that Gateway is even mentioned here since that's really supposed to be "internal"

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -63,8 +63,11 @@ In **Certificate Manager → Code Signing → Signers**, click **Create Signer**
...
+    | **HSM key algorithm** | Algorithm Infisical asks the HSM to generate: `RSA_2048`, `RSA_4096`, `ECC_P256`, or `ECC_P384`. Shown only when **Key source = HSM-backed**. |
```
sheensantoscapadngan: why do we use a different field? shouldn't we just make it uniform and use one field for both HSM and Infisical-backed certs?

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -93,6 +96,19 @@ In **Certificate Manager → Code Signing → Signers**, click **Create Signer**
...
+## HSM-backed Signers
+
+When you pick **Key source = HSM-backed** on the Certificate step, two extra fields appear:
+
+| Field | Description |
+|-------|-------------|
+| **HSM Connector** | The [HSM Connector](/documentation/platform/pki/settings/hsm-connectors) Infisical uses to reach your HSM. Set the Connector up first under **Cert Manager Settings > HSM Connectors**. |
+| **HSM key algorithm** | Algorithm the HSM generates for this Signer: `RSA_2048`, `RSA_4096`, `ECC_P256`, or `ECC_P384`. |
+
+The private key never leaves the HSM. Every sign operation routes through the linked Connector. Infisical only ever holds the public key and the issued certificate. Key source is fixed at create time. To switch a Signer between Infisical-managed and HSM-backed, create a new Signer with the desired key source.
```
sheensantoscapadngan: curious, why aren't you able to just switch the key source? what's the blocker?

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -93,6 +96,19 @@ In **Certificate Manager → Code Signing → Signers**, click **Create Signer**
...
+
+See [HSM Connectors](/documentation/platform/pki/settings/hsm-connectors) for the Gateway + Connector setup an HSM-backed Signer depends on.
```
sheensantoscapadngan: same here, no need to mention Gateway. We should be focusing on what an HSM connector is conceptually

docs/documentation/platform/pki/settings/hsm-connectors.mdx
```diff
@@ -0,0 +1,157 @@
+---
+title: "HSM Connectors"
+sidebarTitle: "HSM Connectors"
+description: "Register a Hardware Security Module that Infisical reaches through one of your Gateways."
```
sheensantoscapadngan, line 4: I don't think we should be mentioning gateways as the center piece. It should be more about how Gateway will serve as the bridge, not that HSM connector IS a gateway, because technically it isn't...

docs/documentation/platform/pki/settings/hsm-connectors.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+Three pieces have to be in place for an HSM Connector to be usable:
+
+1. **A Gateway** running on a host that can talk to your HSM, started with `--pkcs11-module=<path to vendor .so/.dll>`.
+2. **An HSM Connector** that points at that Gateway (or a Pool that contains it) and carries the slot credentials.
+3. **A consumer**, a feature that supports HSM-backed keys. See the feature's own docs for how it references an HSM Connector.
+
+## Run a Gateway with PKCS#11 enabled
+
+The same Gateway binary you already deploy can serve HSM operations: just pass the PKCS#11 driver path at start time. A Gateway started without that flag stays a regular Gateway. HSM Connectors will skip it automatically.
```
sheensantoscapadngan: regardless of whether the driver path is passed in or not, it's always going to be a "regular" gateway...

docs/documentation/platform/pki/settings/hsm-connectors.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+<Steps>
+  <Step title="Install your HSM's PKCS#11 driver on the Gateway host">
+    Install the PKCS#11 library shipped by your HSM vendor on the same machine where the Gateway will run. Note the absolute path to the `.so` or `.dll`. You will pass it to the Gateway in the next step.
+
+    For example, a Fortanix DSM deployment typically installs its driver at `/opt/fortanix/pkcs11/fortanix_pkcs11.so`. Any vendor with a PKCS#11 driver works the same way.
```
sheensantoscapadngan: wait what do you mean by this? Fortanix installs it at that path? I thought you're supposed to just download the files and put it wherever?

docs/documentation/platform/pki/settings/hsm-connectors.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+    Follow the vendor's onboarding to provision a slot, configure operator credentials, and confirm `pkcs11-tool --module <driver> --list-slots` returns the expected slot label.
+  </Step>
+
+  <Step title="Create the Gateway in the Infisical UI">
+    Go to **Organization Settings > Networking > Gateways**, click **Create Gateway**, name it, and pick an auth method. Then click **Show deploy command** and copy the install command for your platform. This is the same flow as a regular Gateway. See [Gateway Deployment](/documentation/platform/gateways/gateway-deployment) for full setup detail (relay, firewall, AWS or token auth).
```
sheensantoscapadngan: same concern here about referring to it as "regular'... it's always a regular gateway since PKCS11 enabled doesn't make it special

docs/documentation/platform/pki/settings/hsm-connectors.mdx
```diff
@@ -0,0 +1,157 @@
...
+  </Step>
+
+  <Step title="(Optional) Add the Gateway to a Pool">
+    For high availability, deploy two or more PKCS#11-enabled Gateways and add them to the same [Gateway Pool](/documentation/platform/gateways/gateway-pools). HSM Connectors that target the pool route each request to a randomly chosen, healthy, PKCS#11-capable member. Non-capable members in the same pool are silently skipped.
+  </Step>
+</Steps>
+
+<Note>
+PKCS#11 driver libraries are vendor-specific binaries. Place them on the Gateway host with the same file permissions the Gateway process expects to read, and verify the driver matches your HSM firmware version before connecting it to production traffic.
+</Note>
```
sheensantoscapadngan, line 105: what file permission? we should point them directly rather than let them guess

docs/documentation/platform/pki/settings/hsm-connectors.mdx
```diff
@@ -0,0 +1,157 @@
...
+    Nothing. Infisical removes its reference to the key label. The actual key object remains on your HSM. Delete it through your HSM's own tooling if you want to free the slot.
+  </Accordion>
+  <Accordion title="Can I route HSM traffic through the same Gateway that serves my databases or app connections?">
+    Yes. Append `--pkcs11-module` to the start command of an existing Gateway and that Gateway will serve HSM operations alongside its other traffic. Restart is required for the driver to load.
+  </Accordion>
+  <Accordion title="Does a Gateway Pool need every member to be PKCS#11-capable?">
+    No. Connectors routed through a pool ignore non-capable members. You can mix one or two PKCS#11-enabled Gateways with the rest of your pool freely.
+  </Accordion>
+  <Accordion title="Which mechanisms are supported?">
+    The PKCS#11 bridge supports `CKM_RSA_PKCS`, `CKM_SHA256_RSA_PKCS`, `CKM_SHA384_RSA_PKCS`, `CKM_SHA512_RSA_PKCS`, and `CKM_ECDSA_SHA256` / `SHA384` / `SHA512`. Key algorithms supported for generation are `RSA_2048`, `RSA_4096`, `ECC_P256`, and `ECC_P384`.
```
sheensantoscapadngan: what are CKM_RSA_PKCS, CKM_SHA256_RSA_PKCS, etc.? seems cryptic, we should give them a label that's understandable

docs/documentation/platform/pki/settings/hsm-connectors.mdx
```diff
@@ -0,0 +1,157 @@
+---
```
sheensantoscapadngan, line 1: we should also have a document specifically for Fortanix DSM...

backend/src/lib/gateway-v2/pkcs11-rpc.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export type Pkcs11RpcSuccess<T> = { ok: true; status: number; result: T };
+export type Pkcs11RpcFailure = {
+  ok: false;
+  status: number;
+  errorCode: string;
+  errorMessage: string;
+};
+export type Pkcs11RpcResponse<T> = Pkcs11RpcSuccess<T> | Pkcs11RpcFailure;
+
```
sheensantoscapadngan, line 21: IGNORE PLACEMENT OF COMMENT

add a link to documentation here as well 
Image

backend/src/lib/gateway-v2/pkcs11-rpc.ts
```diff
@@ -0,0 +1,122 @@
...
+export const PKCS11_RPC_TIMEOUT_MS = 30_000;
+
```
sheensantoscapadngan, line 23: IGNORE PLACEMENT OF COMMENT

text overflow
Image

backend/src/lib/gateway-v2/pkcs11-rpc.ts
```diff
@@ -0,0 +1,122 @@
...
+export const callPkcs11Endpoint = async <T>(args: {
+  port: number;
+  endpoint: Pkcs11RpcEndpoint;
+  body: Pkcs11RpcRequestBody;
+  timeoutMs?: number;
+}): Promise<Pkcs11RpcResponse<T>> => {
+  const payload = JSON.stringify(args.body);
+  return new Promise<Pkcs11RpcResponse<T>>((resolve, reject) => {
+    const req = http.request(
+      {
```
sheensantoscapadngan, line 33: IGNORE PLACEMENT

this page seems very technical... for a new user, they wouldn't know what a Gateway is and what PKCS11 is so it could feel overwhelming... 
Image

here's some suggestions:
Image

backend/src/lib/gateway-v2/pkcs11-rpc.ts
```diff
@@ -0,0 +1,122 @@
...
+        host: "127.0.0.1",
+        port: args.port,
+        path: args.endpoint,
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          "Content-Length": Buffer.byteLength(payload).toString(),
+          Connection: "close"
+        },
+        timeout: args.timeoutMs ?? PKCS11_RPC_TIMEOUT_MS
```
sheensantoscapadngan, line 43: when referring to a new entity altogether in the docs, we should be adding links for the very first mention... in this case, we should add one for Gateway. this makes it a lot easier for people to connect entities

Image.

backend/src/lib/gateway-v2/pkcs11-rpc.ts
```diff
@@ -0,0 +1,122 @@
...
+          const text = Buffer.concat(chunks).toString("utf8");
+          const status = res.statusCode ?? 0;
+          if (!text) {
+            resolve({
+              ok: false,
+              status,
+              errorCode: "internal",
+              errorMessage: `Empty response body from Gateway (status ${status})`
+            });
+            return;
```
sheensantoscapadngan, line 58: IGNORE PLACEMENT OF COMMENT

what do you mean by vendor's path? isn't this something that you get to decide for yourself? 
Image

sheensantoscapadngan: also, I wonder if we should just point them directly to the docs rather than show them here... since it might be too overwhelming

backend/src/lib/gateway-v2/pkcs11-rpc.ts
sheensantoscapadngan: we also use hosts here which could be confusing...
Image

backend/src/server/routes/v1/hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
+import { z } from "zod";
+
```
sheensantoscapadngan, line 2: IGNORE PLACEMENT

I don't think we should specifically call out code signing here...
Image

Image

sheensantoscapadngan: also, notice "hosts" here

backend/src/server/routes/v1/hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+const isProvided = (v: unknown) => v !== undefined && v !== null && v !== "";
+
+export const GATEWAY_PICK_MESSAGE = "Exactly one of gatewayId or gatewayPoolId must be set";
+export const gatewayPickRefiner = (data: { gatewayId?: string | null; gatewayPoolId?: string | null }) => {
+  const hasGatewayId = isProvided(data.gatewayId);
+  const hasPool = isProvided(data.gatewayPoolId);
+  return hasGatewayId !== hasPool;
+};
+
```
sheensantoscapadngan, line 17: IGNORE PLACEMENT

I don't think we should specifically mention "heartbeat" since that's just an internal detail right? I think we do this in the docs as well 
Image

backend/src/server/routes/v1/hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+      z.discriminatedUnion("ok", [
+        z.object({
+          gatewayId: z.string().uuid(),
+          ok: z.literal(true),
+          slotInfo: z.object({
+            manufacturer: z.string(),
+            model: z.string(),
+            firmware: z.string()
+          })
+        }),
```
sheensantoscapadngan, line 31: IGNORE PLACEMENT

we should also link docs to the gateway if we're mentioning it here? Important note here is that people setting up HSM connector could potentially be a different group of people setting up gateways...

Image

backend/src/server/routes/v1/hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+          gatewayId: z.string().uuid(),
+          ok: z.literal(false),
+          errorCode: z.string(),
+          errorMessage: z.string()
+        })
+      ])
+    )
+    .min(1)
+});
+
```
sheensantoscapadngan, line 42: IGNORE PLACEMENT

since we're already calling it "Key Source", does it still make sense to say "-backed" in the labels? Seems like having it as just "Infisical" and "HSM" would be ok ?

Image

backend/src/server/routes/v1/hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+export const CreateBodySchema = z
+  .object({
+    name: slugSchema({ min: 1, max: HSM_CONNECTOR_NAME_MAX, field: "name" }),
+    description: z.string().max(HSM_CONNECTOR_DESCRIPTION_MAX).optional(),
+    gatewayId: z.string().uuid().optional(),
+    gatewayPoolId: z.string().uuid().optional(),
+    credentials: HsmConnectorCredentialsSchema
+  })
+  .refine(gatewayPickRefiner, { message: GATEWAY_PICK_MESSAGE, path: ["gatewayId"] });
+
```
sheensantoscapadngan, line 52: IGNORE PLACEMENT

I don't think we should be mentioning the two extra fields as a description. Doesn't seem notable to explicitly mention it? 
Image

backend/src/server/routes/v1/hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+export const ConnectorIdParamSchema = z.object({
+  connectorId: z.string().uuid()
+});
+
+export const LinkedCertificatesQuerySchema = z.object({
+  offset: z.coerce.number().min(0).default(0),
+  limit: z.coerce.number().min(1).max(100).default(20)
+});
+
```
sheensantoscapadngan, line 80: IGNORE PLACEMENT

seems like we can just remove the "Caller" altogether from the diagram here? 
Image

backend/src/server/routes/v1/cert-manager-hsm-connector-router-schemas.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const ConnectorIdParamSchema = z.object({
```
sheensantoscapadngan: IGNORE PLACEMENT

people connecting this will likely not know about Gateway fully... so this would be misleading/noise. We need to make that assumption 

Image

sheensantoscapadngan: also, note on the "regular gateway" usage...

backend/src/server/routes/v1/hsm-connector-router.ts
```diff
@@ -0,0 +1,239 @@
...
+
+import { EventType } from "@app/ee/services/audit-log/audit-log-types";
+import { ApiDocsTags } from "@app/lib/api-docs";
+import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
+import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
+import { AuthMode } from "@app/services/auth/auth-type";
+
```
sheensantoscapadngan, line 10: IGNORE PLACMENT

seems excessive to mention this... let's keep the doc as packed as possible in terms of content.. if it doesn't add a lot of value, we can just trim it out 
Image

backend/src/server/routes/v1/hsm-connector-router.ts
```diff
@@ -0,0 +1,239 @@
...
+import {
+  ConnectorIdParamSchema,
+  CreateBodySchema,
+  HsmConnectorSanitizedSchema,
+  HsmConnectorTestResultSchema,
+  LinkedCertificatesQuerySchema,
+  LinkedCertificatesResponseSchema,
+  UpdateBodySchema
+} from "./cert-manager-hsm-connector-router-schemas";
+
```
sheensantoscapadngan, line 19: IGNORE PLACEMENT

we should add pkcs11-module to the CLI docs as well..

backend/src/db/schemas/hsm-connectors.ts
```diff
@@ -0,0 +1,26 @@
+// Code generated by automation script, DO NOT EDIT.
+// Automated by pulling database and generating zod schema
+// To update. Just run npm run generate:schema
+// Written by akhilmhdh.
+
+import { z } from "zod";
```
sheensantoscapadngan, line 6: IGNORE PLACEMENT OF COMMENT

here's a few more claude-generated feedback for the docs...see which one would make sense for us here 
Image

backend/src/lib/gateway-v2/pkcs11-rpc.ts
```diff
@@ -0,0 +1,122 @@
...
+    });
+    req.on("error", (err) => {
+      logger.warn({ err }, `PKCS#11 RPC request error [port=${args.port}]`);
+      reject(err);
+    });
+    req.write(payload);
+    req.end();
+  });
+};
+
```
sheensantoscapadngan, line 111: IGNORE PLACEMENT OF COMMENT

I might've already mentioned this but we don't want to refer to a Gateway as a first-class citizen outside the context of HSM connector since it would just make it a lot more confusing
Image

backend/src/services/hsm-connector/hsm-connector-service.ts
carlosmonastyrski: Used a different one initially as it'll enable this for anyone with the current KMS HSM feature, but agree that we can also reuse it now and maybe separate them in the future if needed 👍

backend/src/services/signer/signer-issuance-service.ts
carlosmonastyrski: Checked the major HSM-backed PKI platforms, AWS Private CA, GCP Certificate Authority Service, DigiCert Signing Manager, and HashiCorp Vault PKI all reuse the HSM key across cert renewals by default. Key rotation is always a separate, explicit action. The HSM key is the long-lived signing identity, and the HSM enforces its protection regardless of how many signatures it produces. The cert is just the wrapper that rotates around it, minting a new key on every renewal would defeat the point of putting it in an HSM in the first place. Customers who actually want a fresh key create a new Signer instead of renewing.

docs/documentation/platform/pki/code-signing/signers.mdx
carlosmonastyrski: There wasn’t a real blocker, just a scoping decision. Added it.

backend/src/server/routes/v1/hsm-connector-router-schemas.ts
```diff
@@ -15,18 +18,7 @@ export const gatewayPickRefiner = (data: { gatewayId?: string | null; gatewayPoo
...
-  description: z.string().nullable(),
-  projectId: z.string(),
-  gatewayId: z.string().uuid().nullable(),
-  gatewayPoolId: z.string().uuid().nullable(),
-  slotLabel: z.string(),
-  keyNamePrefix: z.string().nullable(),
-  createdAt: z.date(),
-  updatedAt: z.date()
-});
+export const HsmConnectorSanitizedSchema = SharedHsmConnectorSanitizedSchema;
```
sheensantoscapadngan: wait why do we import it as "SharedHsmConnectorSanitizedSchema" but then still use the exact same name?

Image

backend/src/server/routes/v1/hsm-connector-router.ts
```diff
@@ -7,16 +7,17 @@ import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
...
-  LinkedCertificatesQuerySchema,
-  LinkedCertificatesResponseSchema,
-  UpdateBodySchema
-} from "./cert-manager-hsm-connector-router-schemas";
+  ListHsmConnectorsQuerySchema,
+  UpdateHsmConnectorBodySchema
+} from "./hsm-connector-router-schemas";
 
-export const registerCertManagerHsmConnectorRouter = async (server: FastifyZodProvider) => {
+export const registerHsmConnectorRouter = async (server: FastifyZodProvider) => {
```
sheensantoscapadngan: huddle talk

backend/src/services/certificate-authority/aws-pca/aws-pca-certificate-authority-fns.ts
```diff
@@ -848,11 +848,34 @@ export const AwsPcaCertificateAuthorityFns = ({
...
+    isRenewal?: boolean;
+    originalCertificateId?: string;
+    csr?: string;
+    organization?: string;
+    organizationalUnit?: string;
+    country?: string;
+    state?: string;
+    locality?: string;
+    isCancelled?: () => Promise<boolean>;
+  }) => orderCertificateFromProfile(args);
```
sheensantoscapadngan: wait why did we need to add this?

backend/src/services/certificate-authority/azure-ad-cs/azure-ad-cs-certificate-authority-fns.ts
```diff
@@ -1612,12 +1612,30 @@ export const AzureAdCsCertificateAuthorityFns = ({
...
+    validity: { ttl: string };
+    notBefore?: Date;
+    notAfter?: Date;
+    signatureAlgorithm?: string;
+    keyAlgorithm?: CertKeyAlgorithm;
+    isRenewal?: boolean;
+    originalCertificateId?: string;
+    isCancelled?: () => Promise<boolean>;
+  }) => orderCertificateFromProfile(args);
+
```
sheensantoscapadngan: same question here... might not be needed at all

backend/src/services/certificate-authority/azure-ad-cs/azure-ad-cs-certificate-authority-fns.ts
sheensantoscapadngan: rename

backend/src/services/hsm-connector/hsm-connector-fns.ts
sheensantoscapadngan: remove and check

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -63,11 +63,10 @@ In **Certificate Manager → Code Signing → Signers**, click **Create Signer**
...
     | **Common Name** | The legal name shown on the certificate, for example `Acme Mobile, Inc.`. Fixed once the certificate is issued. To change it you must reissue. |
-    | **Key algorithm** | The key for the signing certificate: `RSA-2048`, `RSA-3072`, `RSA-4096`, `ECDSA P-256`, `ECDSA P-384`, or `ECDSA P-521`. Shown only when **Key source = Infisical-managed**. |
-    | **HSM Connector** | The Connector that decides which Gateway routes operations to your HSM. Shown only when **Key source = HSM-backed**. See [HSM Connectors](/documentation/platform/pki/settings/hsm-connectors). |
-    | **HSM key algorithm** | Algorithm Infisical asks the HSM to generate: `RSA_2048`, `RSA_4096`, `ECC_P256`, or `ECC_P384`. Shown only when **Key source = HSM-backed**. |
+    | **Key algorithm** | The signing key algorithm. With **Key source = Infisical**: `RSA-2048`, `RSA-3072`, `RSA-4096`, `ECDSA P-256`, `ECDSA P-384`, or `ECDSA P-521`. With **Key source = HSM**: `RSA_2048`, `RSA_4096`, `ECC_P256`, or `ECC_P384`. |
```
sheensantoscapadngan: hmm why is it that Key Algorithm is still different between HSM and Infisical? Is ECC_P256 here the same as ECDSA P-256 and can we just unify it so that regardless of the key source, it will always accept ECDSA P-256 and do transformation in our backend if needed? Same thing applies between RSA-2048 and RSA_2048 and the rest. Seems confusing for users that they have to think about different formats

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -98,16 +97,11 @@ In **Certificate Manager → Code Signing → Signers**, click **Create Signer**
...
-| **HSM key algorithm** | Algorithm the HSM generates for this Signer: `RSA_2048`, `RSA_4096`, `ECC_P256`, or `ECC_P384`. |
+You can later switch an existing Signer between Infisical and HSM by reissuing the certificate with a different key source. Hit the [reissue endpoint](/api-reference/endpoints/code-signing/signers/reissue) with a `certificate.keySource` body, plus `hsmConnectorId` and `hsmKeyAlgorithm` when switching to HSM. The signer keeps its name, members, policy, and CA. Only the underlying key and the cert wrapping it change.
 
-The private key never leaves the HSM. Every sign operation routes through the linked Connector. Infisical only ever holds the public key and the issued certificate. Key source is fixed at create time. To switch a Signer between Infisical-managed and HSM-backed, create a new Signer with the desired key source.
-
-See [HSM Connectors](/documentation/platform/pki/settings/hsm-connectors) for the Gateway + Connector setup an HSM-backed Signer depends on.
+See [HSM Connectors](/documentation/platform/pki/settings/hsm-connectors) for the setup an HSM-backed Signer depends on.
```
sheensantoscapadngan: can just be "for the setup."

docs/documentation/platform/pki/code-signing/signers.mdx
```diff
@@ -183,8 +177,10 @@ To sign with a Signer, pick the integration that matches your tooling:
   <Accordion title="What happens to signed artifacts when I delete a Signer?">
     Artifacts already signed remain valid as long as the certificate they were signed with is still valid and the verifier trusts the issuing CA's chain. Deleting the Signer in Infisical only removes Infisical's ability to sign new artifacts with that certificate. It does not invalidate past signatures.
   </Accordion>
-  <Accordion title="Can I migrate an existing Infisical-managed Signer to an HSM?">
-    Not in place. Key source is fixed at create time so the Signer's certificate and its key never go out of sync. Create a new HSM-backed Signer alongside the old one, route new sign traffic to it, and disable or delete the old Signer once you're satisfied.
+  <Accordion title="Can I migrate an existing Signer between Infisical and an HSM?">
+    Yes. Reissue the Signer's certificate with a different `certificate.keySource` via the [reissue endpoint](/api-reference/endpoints/code-signing/signers/reissue). When switching to HSM you also pass `hsmConnectorId` and `hsmKeyAlgorithm`. The signer keeps its name, members, approval policy, and CA. Only the key and the cert wrapping it change.
```
sheensantoscapadngan: are you also able to do this via the UI? if so, perhaps we can highlight that instead?

docs/documentation/platform/pki/settings/hsm-connectors-fortanix-dsm.mdx
```diff
@@ -0,0 +1,93 @@
...
+| **Key label prefix** | Optional. Useful if your DSM hosts keys for multiple environments. |
+
+On submit Infisical runs a Verify round-trip through the Gateway. A wrong PIN, missing slot label, or unreachable Gateway is caught here and the Connector is not persisted.
+
+## Troubleshooting
+
+- **"slot not found" on Verify**: confirm the slot label matches what `pkcs11-tool --list-slots` reports. Fortanix slot labels are case-sensitive.
+- **"login failed" on Verify**: re-check the PIN. The PIN is the Fortanix application's PKCS#11 PIN, not the DSM account password.
+- **"gateway unreachable"**: the Gateway must be running with HSM support enabled. Check the Gateway log for the PKCS#11 module load line.
+- **Slow operations**: if the Gateway host can't reach DSM over HTTPS with low latency, every sign call pays the round-trip. Run the Gateway in the same network as DSM whenever possible.
```
sheensantoscapadngan: nit - let's use an accordion here like the following? We should ideally use mintlify components where applicable so it doesn't look too plain

Image

docs/documentation/platform/pki/settings/hsm-connectors-fortanix-dsm.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+Download the Fortanix PKCS#11 client from the Fortanix DSM dashboard (Settings -> Client configurations) for the OS that will run the Gateway. Install it and confirm the driver file path is readable by the Gateway user:
+
+```bash
+ls -l /opt/fortanix/pkcs11/fortanix_pkcs11.so
+```
+
+The default install path on Linux is `/opt/fortanix/pkcs11/fortanix_pkcs11.so`. If you put it somewhere else, use that path everywhere in the steps below.
+
+Create or copy the Fortanix client configuration file at `/etc/fortanix/pkcs11/pkcs11.conf`. The minimum keys are the DSM endpoint and the API key for the application that will be used as the PKCS#11 user. Fortanix provides the file content directly from the dashboard.
```
sheensantoscapadngan: im confused what this pkcs11.conf file is for...

is this supposed to be where users extract the values to use for configuring the HSM connector? if so, we should make it a lot clearer. When reading through this, I was under the impression that we'd also have to reference this somewhere but it doesn't seem like it? Also, could they instead just fetch the values for the connector from Fortanix UI or does it have to be from this file?

docs/documentation/platform/pki/settings/hsm-connectors-fortanix-dsm.mdx
```diff
@@ -0,0 +1,93 @@
...
+description: "Connect Infisical to Fortanix Data Security Manager via an HSM Connector."
+---
+
+This page covers the end-to-end setup for connecting Fortanix Data Security Manager (DSM) to Infisical through an HSM Connector. Read [HSM Connectors](/documentation/platform/pki/settings/hsm-connectors) first for the concept.
+
+## Prerequisites
+
+- A Fortanix DSM account with at least one group and an application configured for PKCS#11 access.
+- A host inside your network that can reach your DSM endpoint over HTTPS, where you can install the Fortanix PKCS#11 client and run an [Infisical Gateway](/documentation/platform/gateways/overview).
+- Project-level permission to create HSM Connectors in Infisical (`HSM Connectors > Create`).
```
sheensantoscapadngan: project mentioned

backend/src/db/migrations/20260618143217_pki-hsm-connector.ts
```diff
@@ -0,0 +1,205 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasTable(TableName.HsmConnector))) {
+    await knex.schema.createTable(TableName.HsmConnector, (t) => {
+      t.uuid("id", { primaryKey: true }).defaultTo(knex.fn.uuid());
+      t.string("name", 32).notNullable();
+      t.string("description", 256);
+      t.string("projectId").notNullable();
+      t.foreign("projectId").references("id").inTable(TableName.Project).onDelete("CASCADE");
+
```
sheensantoscapadngan, line 36: IGNORE PLACEMENT

I don't think we have to explicitly mention "signing" here because it might be understood as just relevan tot "code signing".. After this initiative, we'll follow with an HSM integration for Internal CAs

Image

sheensantoscapadngan: same remark here... perhaps we call it "private" instead? 
Image

sheensantoscapadngan:
let's also use mintlify's "Steps" component here please
image

sheensantoscapadngan (approved):

PR #6941, opened 2026-06-19
fix(scim): split filter string on ' eq ' to avoid false split on values containing 'eq'

jaydeep-pipaliya:
## What

`parseScimFilter` was splitting the SCIM filter string on the bare
substring `"eq"` instead of the operator token `" eq "`. Any attribute
value that contains `"eq"` — a group named `"Equity Team"`, a username
like `"frequency@example.com"`, or any similar value — gets split at the
first occurrence of those two letters inside the value, truncating the
search term.

Concrete example with `displayName eq "Equity Team"`:

```
"displayName eq \"Equity Team\"".split("eq")
//  → ["displayName ", " \"E", "uity Team\""]
//  parsedName  = "displayName"
//  parsedValue = " \"E"  →  "E"  (after trim + strip-quotes)
```

The lookup ends up searching for `name = "E"` instead of
`name = "Equity Team"`, so the group is never found and the next sync
creates a duplicate instead of updating the existing one.

## Fix

Split on `" eq "` (with mandatory surrounding spaces). RFC 7644
§3.4.2.2 defines the filter grammar as `attrPath SP compareOp SP
compValue`, so the operator is always surrounded by whitespace. With
the space-delimited split the entire value is captured intact:

```
"displayName eq \"Equity Team\"".split(" eq ")
//  → ["displayName", "\"Equity Team\""]
//  parsedValue = "Equity Team"  (after trim + strip-quotes)  ✓
```

I considered adding a full SCIM filter parser to handle `co`, `sw`,
`ne`, `gt`, etc. but kept the minimal fix because this function is
only called for equality filters sent by IdPs (displayName/userName)
and a broader parser is a separate, larger change.

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/scim/scim-fns.ts
+++ b/backend/src/ee/services/scim/scim-fns.ts
@@ -20,7 +20,7 @@ export const buildScimUserList = ({
 
 export const parseScimFilter = (filterToParse: string | undefined) => {
   if (!filterToParse) return {};
-  const [parsedName, parsedValue] = filterToParse.split("eq").map((s) => s.trim());
+  const [parsedName, parsedValue] = filterToParse.split(" eq ").map((s) => s.trim());
 
   let attributeName = parsedName;
   if (parsedName === "userName") {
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Good catch. Added a guard for the undefined case in the follow-up commit - returns an empty object (same pattern as the existing early-return for a missing filter), so callers like `findGroups` get an empty filter object rather than a thrown TypeError.

PR #6942, opened 2026-06-19
fix(scim): batch-fetch group members and fix totalResults in listScimGroups

jaydeep-pipaliya:
## Problem

`listScimGroups` loaded group members with a separate DB query per group (N+1 pattern). For an org with 50 groups, listing groups triggered 51 queries. Under load this adds up and slows SCIM provisioning syncs noticeably.

A second bug: `buildScimGroupList` always set `totalResults` to the length of the current page instead of the total count across all pages. RFC 7644 §3.4.2 requires `totalResults` to be the total number of resources matching the query, not just the current page. Clients that rely on this to drive pagination or display counts get wrong numbers.

## Fix

- Added `findGroupMembershipsByGroupIdsInOrg` to `user-group-membership-dal.ts` - same join logic as the existing single-ID variant but accepts a `string[]` and uses `whereIn`. Fetches members for all groups in one query.
- Added `countGroups` to `group-dal.ts` - counts rows with the same filter used for `findGroups`, so `totalResults` reflects the true total rather than page size.
- `listScimGroups` now fires `findGroups` and `countGroups` in parallel (`Promise.all`), then builds a `Map<groupId, members[]>` in memory to populate each group without extra round-trips.
- `buildScimGroupList` now accepts an explicit `totalResults` parameter instead of computing it from `scimGroups.length`.

## Testing

Covered by existing SCIM integration test suite. The batch query path handles the empty-group list correctly (skips the query when `groupIds` is empty to avoid a `whereIn([])` no-op).

### DIFF
--- a/backend/src/ee/services/group/group-dal.ts
+++ b/backend/src/ee/services/group/group-dal.ts
@@ -40,6 +40,19 @@ export const groupDALFactory = (db: TDbClient) => {
     }
   };
 
+  const countGroups = async (filter: TFindFilter<TGroups>, tx?: Knex): Promise<number> => {
+    try {
+      const result = await (tx || db.replicaNode())(TableName.Groups)
+        // eslint-disable-next-line
+        .where(buildFindFilter(filter))
+        .count<{ count: string }>({ count: "*" })
+        .first();
+      return Number(result?.count ?? 0);
+    } catch (err) {
+      throw new DatabaseError({ error: err, name: "Count groups" });
+    }
+  };
+
   const findByOrgId = async (orgId: string, tx?: Knex) => {
     try {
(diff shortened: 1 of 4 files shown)


jaydeep-pipaliya:
Fair flag, but the double `whereIn` on `groupId` mirrors the existing single-id variant `findGroupMembershipsByGroupIdInOrg` at lines 187-218 of the same file. Both belt-and-braces against the case where a group row exists but its Membership row is gone (or pending). The batch function intentionally copies that shape for parity.

If we want to drop the redundancy, it should be a separate PR that touches both functions together so they stay consistent. Doing it in this PR would introduce drift between the two helpers.

Happy to open that follow-up if you'd like, but it felt out of scope for the N+1 fix.

PR #6952, opened 2026-06-19
fix(secret-sync): fix personal tokens and sealed values on railway sync

adilsitos:
## Context
This PR fixes 3 bugs from railway secret sync. The first one is the creation of app connections using personal tokens, the second one is about checking the deployment status to avoid errors and the last one is about handling null values returned by railway. 

This also fixes: secrets-386 and secrets-306

### Bug 1: Personal tokens 
- Try to create an app connection using a personal token
- With this new app connection, try to create a secret sync
- check that you can list projects and teams (this was also not working before) 

### Bug 2: null argument error (you can run this with validation 1)
- In railway, create a secret that is sealed (do not forget to redeploy the service)
- create an app connection 
- create a secret sync and use the service that has this sealed variable 
- trigger the sync 
- check that no error happen 

### Bug 3: Distinguish regular deployments and cron services in Railway syncs
- Create a cron job in railway 
- try to make this in a invalid status (check `RAILWAY_NON_REDEPLOYABLE_STATUSES`)
- check that no error happen in this case.

### Validation 1: sealed values are not imported
For context: We should not interfer in sealed values, even if the user wants to delete it.
- create a railway service with a sealed value 
- create a sync 
- check that the sealed value is not imported 

### validation 2: value is not sealed at the import but it becomes sealed
- create a railway service without sealed values
- create a sync 
- check that the value is imported 
- On railway, change the variable to be a sealed one 
- delete this value (or modify it) inside infisical 
- check that this is not deleted inside railway

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/e2e-test/routes/v1/secret-folder.spec.ts
+++ b/backend/e2e-test/routes/v1/secret-folder.spec.ts
@@ -36,6 +36,8 @@ const deleteFolder = async (dto: { path: string; id: string }) => {
 };
 
 describe("Secret Folder Router", async () => {
+  vi.setConfig({ testTimeout: 10_000 });
+
   test.each([
     { name: "folder1", path: "/" }, // one in root
     { name: "folder1", path: "/level1/level2" }, // then create a deep one creating intermediate ones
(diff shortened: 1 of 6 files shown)


adilsitos:
@greptile check again and update the summary

mathnogueira (dismissed):

varonix0 (approved):

PR #6956, opened 2026-06-19
feat(pam): add MongoDB and MS SQL Server account types

bernie-g:
## Summary
- Adds MongoDB and Microsoft SQL Server as supported PAM account types for CLI access.
- Both types use the existing schema-driven form system, so no frontend changes are needed.

## Type of change
- [x] New feature (non-breaking change which adds functionality)

### DIFF
--- a/backend/src/ee/services/pam-account/pam-account-schemas.ts
+++ b/backend/src/ee/services/pam-account/pam-account-schemas.ts
@@ -1,9 +1,8 @@
 /* eslint-disable no-underscore-dangle */
 import dns from "dns";
-import { promisify } from "util";
-
 import ConnectionString from "mongodb-connection-string-url";
 import RE2 from "re2";
+import { promisify } from "util";
 import { z } from "zod";
 
 import { BadRequestError } from "@app/lib/errors";
--- a/backend/src/ee/services/pam-account/pam-account-schemas.ts
+++ b/backend/src/ee/services/pam-account/pam-account-schemas.ts
@@ -236,7 +236,7 @@ export const ACCOUNT_TYPE_CONFIGS = {
             ctx.addIssue({
               code: z.ZodIssueCode.custom,
               message:
-                "Credentials should not be included in the connection string. They are managed separately per account"
+                "Credentials should not be included in the connection string. Use the Username and Password fields instead"
             });
             return z.NEVER;
           }
--- a/backend/src/ee/services/pam-account/pam-account-schemas.ts
+++ b/backend/src/ee/services/pam-account/pam-account-schemas.ts
@@ -175,6 +175,7 @@ export const ACCOUNT_TYPE_CONFIGS = {
     }),
     ui: {
       port: { defaultValue: 1433 },
+      database: { defaultValue: "master" },
       authMethod: {
         label: "Auth Method",
         defaultValue: "sql-login",
--- a/backend/src/ee/services/pam-account/pam-account-schemas.ts
+++ b/backend/src/ee/services/pam-account/pam-account-schemas.ts
@@ -142,15 +142,8 @@ export const ACCOUNT_TYPE_CONFIGS = {
     name: "MongoDB",
     icon: "MongoDB.png",
     connectionDetails: z.object({
-      host: z.string().trim().min(1).max(255),
-      port: z.coerce.number(),
+      connectionString: z.string().trim().min(1).max(2048),
       database: z.string().trim().min(1).max(255),
-      authSource: z
-        .string()
-        .trim()
-        .max(255)
-        .transform((v) => v || undefined)
-        .optional(),
       sslEnabled: z.boolean(),
       sslRejectUnauthorized: z.boolean(),
       sslCertificate: z
@@ -170,8 +163,12 @@ export const ACCOUNT_TYPE_CONFIGS = {
     }),
     sanitizedCredentials: z.object({ username: z.string().optional() }),
     ui: {
-      port: { defaultValue: 27017 },
-      authSource: { label: "Auth Source" },
+      connectionString: {
+        label: "Connection String",
+        widget: PamFieldWidget.Textarea,
+        tooltip: "Supports mongodb:// and mongodb+srv:// URIs. Do not include credentials or database in the URI."
+      },
+      database: { defaultValue: "admin" },
       sslEnabled: { label: "SSL Enabled" },
       sslRejectUnauthorized: {
         label: "Reject Unauthorized",
@@ -295,12 +292,23 @@ export const extractGatewayTarget = (
     case PamAccountType.SSH:
     case PamAccountType.Postgres:
     case PamAccountType.MySQL:
-    case PamAccountType.MongoDB:
     case PamAccountType.MsSQL:
       return {
         host: (validated as { host: string; port: number }).host,
         port: (validated as { host: string; port: number }).port
       };
+    case PamAccountType.MongoDB: {
+      const { connectionString } = validated as { connectionString: string };
+      try {
+        const url = new URL(connectionString);
+        return {
+          host: url.hostname,
+          port: url.port ? Number(url.port) : undefined
+        };
+      } catch {
+        return { host: connectionString };
+      }
+    }
     default:
       throw new Error(`No gateway target extraction defined for account type '${accountType}'`);
   }
@@ -361,7 +369,10 @@ export const PamFieldDescriptorSchema = z.object({
   defaultValue: z.union([z.string(), z.number(), z.boolean()]).optional(),
 
   // Only render when the referenced field equals this value
-  showWhen: z.object({ field: z.string(), equals: z.union([z.string(), z.boolean()]) }).optional()
+  showWhen: z.object({ field: z.string(), equals: z.union([z.string(), z.boolean()]) }).optional(),
+
+  // Info tooltip shown next to the label
+  tooltip: z.string().optional()
 });
 
 type PamFieldDescriptor = z.infer<typeof PamFieldDescriptorSchema>;
@@ -383,6 +394,7 @@ type TFieldUiHint = {
   secret?: boolean;
   defaultValue?: string | number | boolean;
   showWhen?: PamFieldDescriptor["showWhen"];
+  tooltip?: string;
 };
 
 const humanizeKey = (key: string) => {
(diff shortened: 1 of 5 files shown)


backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -205,8 +205,18 @@ export const pamSessionServiceFactory = ({
...
 
+    const mergedCredentials: Record<string, unknown> = { ...connectionDetails, ...credentials };
+
+    if (account.accountType === PamAccountType.MongoDB) {
+      const host = connectionDetails.host as string;
+      const port = connectionDetails.port as number;
+      const authSource = (connectionDetails.authSource as string) || undefined;
+      const qs = authSource ? `?authSource=${encodeURIComponent(authSource)}` : "";
+      mergedCredentials.connectionString = `mongodb://${host}:${port}${qs}`;
+    }
```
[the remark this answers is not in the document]
bernie-g: Wrong

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +97,95 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      sslCertificate: {
+        label: "SSL Certificate",
+        widget: PamFieldWidget.Textarea,
+        showWhen: { field: "sslEnabled", equals: true }
+      },
+      password: { widget: PamFieldWidget.Password, secret: true }
+    }
+  },
+
+  [PamAccountType.MongoDB]: {
```
saifsmailbox98, line 213: The MongoDB resource on main (`pam-resource/mongodb/mongodb-resource-schemas.ts`) uses a connection string model -- a full `mongodb://` or `mongodb+srv://` URI validated with the `mongodb-connection-string-url` library, with credentials and database stripped out and managed separately. This config uses individual host/port fields instead, which means no SRV support and the connection string gets manually assembled in the session service as `mongodb://host:port?authSource=...` (missing the `/database` path segment). Should this match the main resource schema's approach?

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +97,95 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+    name: "MongoDB",
+    icon: "MongoDB.png",
+    connectionDetails: z.object({
+      host: z.string().trim().min(1).max(255),
+      port: z.coerce.number(),
+      database: z.string().trim().min(1).max(255),
```
saifsmailbox98: Main's MongoDB resource schema caps database at 64 chars and validates with `regex(/^[a-zA-Z0-9_-]+$/)`. This allows 255 chars and no regex.

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +97,95 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      sslEnabled: z.boolean(),
+      sslRejectUnauthorized: z.boolean(),
+      sslCertificate: z
+        .string()
+        .trim()
+        .transform((v) => v || undefined)
+        .optional()
+    }),
+    credentials: z.object({
+      username: z.string().trim().min(1).max(255),
```
saifsmailbox98, line 270: Main's MongoDB account schema uses max 256 for username, not 255.

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +97,95 @@ export const ACCOUNT_TYPE_CONFIGS = {
     }
   },
 
+  [PamAccountType.MsSQL]: {
```
saifsmailbox98, line 107: Main's MsSQL resource supports three auth methods: SqlLogin, NTLM, and Kerberos. NTLM adds a `domain` field, Kerberos adds `realm`, `kdcAddress`, and `spn`. This only has username/password. Is that intentional for CLI access, or should we support the other auth methods?

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +97,95 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+    name: "Microsoft SQL Server",
+    icon: "MsSql.png",
+    connectionDetails: z.object({
+      host: z.string().trim().min(1).max(255),
+      port: z.coerce.number(),
+      database: z.string().trim().min(1).max(128),
```
saifsmailbox98: Main's MsSQL resource uses the shared `BaseSqlResourceConnectionDetailsSchema` which has `database` max 255. This uses 128.

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +97,95 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      sslEnabled: z.boolean(),
+      sslRejectUnauthorized: z.boolean(),
+      sslCertificate: z
+        .string()
+        .trim()
+        .transform((v) => v || undefined)
+        .optional()
+    }),
+    credentials: z.object({
+      username: z.string().trim().min(1).max(128),
```
saifsmailbox98: Main's MsSQL account schema uses max 63 for username (from `BaseSqlAccountCredentialsSchema`). This uses 128.

saifsmailbox98 (commented):
Compared these schemas against the existing PAM resource/account schemas on main. A few mismatches worth discussing.

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-schemas.ts
saifsmailbox98, line 213: Some connection string that won't work on this branch, but works on `main`:
- `mongodb+srv://cluster0.abc.mongodb.net` -- SRV/Atlas-style, no port; PR requires a port
- `mongodb://mongo1:27017,mongo2:27017,mongo3:27017/?replicaSet=myRS` -- replica set with multiple hosts; PR has a single host field
- `mongodb://mongo.example.com:27017/?retryWrites=true&w=majority` -- query params beyond authSource are lost
- `mongodb://primary:27017,secondary:27018` -- different ports per node; single host/port can't represent this

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +97,157 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      },
+      password: { widget: PamFieldWidget.Password, secret: true }
+    }
+  },
+
+  [PamAccountType.MongoDB]: {
+    name: "MongoDB",
+    icon: "MongoDB.png",
+    connectionDetails: z.object({
+      connectionString: z.string().trim().min(1).max(2048),
```
saifsmailbox98: Main validates the connection string with `mongodb-connection-string-url` and rejects:
- Embedded credentials (e.g. `mongodb+srv://user:pass@host`)
- Database in the URI path (e.g. `mongodb://host/mydb`)
- Invalid protocol (must be `mongodb://` or `mongodb+srv://`)

This is just `z.string().min(1).max(2048)` with no validation. Two reasons to add it:
- Credentials and database are managed as separate fields, so we should reject them in the connection string to avoid confusion and duplication
- `connectionDetails` is returned unsanitized in API responses, so someone could save credentials in the connection string and the password would be exposed

bernie-g: Added validation

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -206,10 +402,23 @@ export const extractGatewayTarget = (
     case PamAccountType.SSH:
     case PamAccountType.Postgres:
     case PamAccountType.MySQL:
+    case PamAccountType.MsSQL:
       return {
         host: (validated as { host: string; port: number }).host,
         port: (validated as { host: string; port: number }).port
```
saifsmailbox98, line 414: Two issues with `extractGatewayTarget` for MongoDB:

1. Uses `new URL()` which doesn't handle multi-host strings correctly (`mongodb://host1:27017,host2:27017` -- URL parser treats everything after the first host as a path). `ConnectionString` is already imported and handles this.

2. For SRV URIs, main resolves the SRV record server-side (`dns.resolveSrv`) so the relay gets a concrete host:port to tunnel to, and randomizes across replica set members. Both call sites are already async so making this async is just adding `await` in two places.

Main's approach is in `mongodb-resource-factory.ts` (`parseMongoConnectionString` + `resolveMongoSrvHost`).

saifsmailbox98 (commented):
ignore

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -97,6 +98,199 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+              message: "Invalid MongoDB connection string. Must start with mongodb:// or mongodb+srv://"
+            });
+            return z.NEVER;
+          }
+
+          if (cs.username || cs.password) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message:
+                "Credentials should not be included in the connection string. They are managed separately per account"
```
saifsmailbox98: Says "managed separately per **account**" but Database message says "Use the Database **field** instead." For consistency, change the credentials message to something like "Use the Username and Password fields instead."

Because this message was used when they were configuring resources and there was a separate configuration needed on accounts. But now this itself is the account, so we can refer to the "fields" directly and also be consistent.

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Fixed

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Fixed

saifsmailbox98 (approved):

PR #6959, opened 2026-06-20
perf(signer): bulk-revoke active grants instead of looping updateById

jaydeep-pipaliya:
## What

`revokeRequest` in `signer-policy-service.ts` fetched every active grant for a signer approval request and then issued one `updateById` per grant inside the transaction. For a request with N grants that is N+1 round trips (1 find + N updates) when one bulk update covers it.

Replaced the find + loop with a single `approvalRequestGrantsDAL.update({ requestId, status: Active }, { ... })`. The grant DAL is `ormify`-based so this is one `WHERE requestId = ? AND status = 'active'` UPDATE.

## Behavior note

On existing rows the result is identical. If another transaction inserts a fresh Active grant for the same request between the original find and the per-row updates, the old code would miss revoking it. The bulk version revokes everything still Active at the moment of the UPDATE, which is what you want when cancelling a request.

Also bumped the `Pick<TApprovalRequestGrantsDALFactory, ...>` in the factory deps to expose `update`.

## Checklist

- [ ] I have added tests (signer service has no unit tests for revokeRequest; this is a perf refactor with no behavior change on existing rows)

### DIFF
--- a/backend/src/services/signer/signer-policy-service.ts
+++ b/backend/src/services/signer/signer-policy-service.ts
@@ -55,7 +55,7 @@ type TSignerPolicyServiceFactoryDep = {
   signerRequestDAL: TSignerRequestDALFactory;
   approvalRequestStepsDAL: Pick<TApprovalRequestStepsDALFactory, "create">;
   approvalRequestStepEligibleApproversDAL: Pick<TApprovalRequestStepEligibleApproversDALFactory, "create">;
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Dropped `find` and `updateById` from the Pick — both are unused after the bulk-update refactor (only `create` and `update` are called now). e3fbe15.

PR #6967, opened 2026-06-22
perf(access-approval-policy): batch the per-env existence check

jaydeep-pipaliya:
## What

`createAccessApprovalPolicy` and `updateAccessApprovalPolicy` looped over the env list and called the DAL once per env to detect a conflicting policy for the same `secretPath`:

```ts
for (const env of envs) {
  // eslint-disable-next-line no-await-in-loop
  if (await $policyExists({ envId: env.id, secretPath })) {
    throw new BadRequestError({ ... });
  }
}
```

The DAL helper `findPolicyByEnvIdAndSecretPath` already accepts `envIds: string[]` and returns the conflicting policy with its `environments` array. The loop was N round trips when one query covers it.

This PR replaces the loop with a single batched call.

## Preserves behavior

When a conflict exists, the new code intersects `policy.environments` with the requested `envs` to identify the offending env, so the thrown error still names the same env slug as before. The update site keeps the existing self-exclusion via `excludePolicyId` (a policy is not allowed to conflict with itself).

Removes `$policyExists` since both call sites move to the new helper.

## Checklist

- [ ] I have added tests (no unit tests exist for this service path; the access-approval-policy DAL function being called was already used in batch form elsewhere)

### DIFF
--- a/backend/src/ee/services/access-approval-policy/access-approval-policy-dal.ts
+++ b/backend/src/ee/services/access-approval-policy/access-approval-policy-dal.ts
@@ -149,7 +149,7 @@ export interface TAccessApprovalPolicyDALFactory
     | undefined
   >;
   findPolicyByEnvIdAndSecretPath: (
-    { envIds, secretPath }: { envIds: string[]; secretPath: string },
+    { envIds, secretPath, excludePolicyId }: { envIds: string[]; secretPath: string; excludePolicyId?: string },
     tx?: Knex
   ) => Promise<{
     name: string;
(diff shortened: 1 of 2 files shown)


jaydeep-pipaliya:
Codex P1 was right — the old per-env loop checked each env independently, so a self-policy first-match could not hide a third-party conflict on a different env. Pushed b4ae0fd that adds `excludePolicyId` to the DAL and uses it from the service, so the query never returns self and the `policy != null` check is sufficient.

Greptile P2 (the `?? envs[0]` fallback) — the DAL filter `envId IN (envIds)` guarantees the returned policy shares at least one env with the caller's list, so the fallback is now structurally unreachable. Kept as a defensive last resort.

PR #6970, opened 2026-06-22
fix(webhook): skip soft-deleted projects so webhooks don't fire during the delete grace window

jaydeep-pipaliya:
## The bug

`webhookFindQuery` and `findAllWebhooks` in `backend/src/services/webhook/webhook-dal.ts` join `project_environments` and filter `Environment.deleteAfter IS NULL`, but they never join the `projects` table or filter `Project.deleteAfter IS NULL`.

The two soft-deletes are independent — soft-deleting a project does **not** soft-delete its environments. So both queries return webhooks for projects that are sitting in their delete grace window.

## Why this is dangerous

`fnTriggerWebhook` runs from the `QueueName.SecretWebhook` BullMQ worker. Two ways the deleted project's webhook can fire:

1. **Queue lag at the moment of deletion.** A trigger event is enqueued for project P, then P is soft-deleted before the worker dequeues it. The worker pulls the job, calls `findAllWebhooks(P.id, env)` and fires the webhook to the user's external URL.
2. **Activity during the grace window.** Cron-driven secret rotation, replication catch-up, queued retries — anything that writes to the project's secrets during the grace window enqueues a fresh trigger job for a project the user has marked for deletion. The webhook still fires.

Webhooks are the canonical external side effect — they ship secret payloads to a customer-controlled URL. Once a user has clicked delete, no external traffic for that project should leave the system.

The existing API path is already protected upstream by `projectDAL.findById` (it returns nothing for a soft-deleted project), but the queue worker calls into `findAllWebhooks` directly with the queued projectId, never going through `projectDAL.findById` for a soft-delete check.

## Fix

Adds a `Project` join and `whereNull(Project.deleteAfter)` to both `webhookFindQuery` and `findAllWebhooks`. The two queries now match the pattern already in `org-product-stats-dal.ts`, which correctly filters both `Project.deleteAfter` and `Environment.deleteAfter` on every cross-project enumeration.

Defense in depth — the API path was already safe via the upstream `projectDAL.findById`, but the queue path was bypassing it.

## Checklist

- [ ] I have added tests (no DAL-level unit tests in this area; the change is the documented pattern from `org-product-stats-dal.ts`)


jaydeep-pipaliya:
Codex P1 was right — the new `projects` join brings in a second `slug` column and Postgres would reject the query with "column reference 'slug' is ambiguous" whenever `findAllWebhooks(projectId, environment)` is called (which the SecretWebhook worker always does).

Pushed 2676cfb that qualifies both predicates to their owning tables (`project_environments.slug`, `webhooks.secretPath`). The `secretPath` one wasn't ambiguous today but qualifying it makes the query robust if `secretPath` ever gets added to another joined table.

PR #6973, opened 2026-06-22
fix(billing): correct license server v2 customer portal behavior

PrestigePvP:
## Context

Fixes a batch of issues found during License Server v2 gamma testing of the custom "Usage & Billing" Customer Portal. All changes are in this repo (the portal UI and its billing-v2 service); catalog data (pricing, per-plan feature lists, product ordering) is owned by the License Server and handled separately.

- **Checkout vs Stripe portal routing** is now decided per product by entitlement. An owned product opens the Stripe portal ("Manage in Stripe"); a not-yet-owned self-serve product goes through checkout ("Continue to Checkout"), even when the org already has a subscription. Previously any existing subscriber was sent to the Stripe portal for every product, so adding a new product never reached checkout.
- **Two-outcome checkout response**: the checkout endpoint now returns either `checkout_created` (redirect to Stripe) or `subscription_updated` (the product was added straight to the existing subscription). The UI redirects only for `checkout_created`; for `subscription_updated` it refreshes the overview and notifies in place instead of navigating to an empty URL. The client schema tolerates the legacy `{ url }` shape so the cloud and License Server can deploy in any order.
- **Top-right plan badge** is derived from the real plan slug instead of the `groups` feature flag, so a Pro org that holds an SSO/groups add-on is no longer mislabeled "Enterprise". Logic is shared via `getSubscriptionPlanLabel` so NavBar and SidebarHeader cannot drift.
- **Plan slug under v2 mode**: the entitlement projection only carries feature flags, so `getPlan` now sets the slug from the subscription tier (enterprise > pro), non-fatally. This also repairs other `subscription.slug` consumers (member-limit bypass, Pro custom-roles banner, etc.) that were silently treating every v2 org as having no tier.
- **Usage card** shows the real machine-identity count via `countAllOrgIdentities` (it was hardcoded to `0`). "Unlimited" remains only when a cap is genuinely absent.
- **Enterprise option** in the product sheet now exposes a working "Contact sales" button on the Enterprise card for products that also have a self-serve Pro tier (routes to talk-to-us).
- **Empty state**: removed the redundant "Get started" card whose button only scrolled to the product list already on screen.

## Screenshots

Behavioral/correctness changes, visible on the org Usage & Billing page (badge, usage counts, product sheet CTAs, checkout routing).

## Steps to verify the change

- As an org with an existing subscription, open a not-yet-owned product and confirm the CTA is "Continue to Checkout". When the License Server returns `subscription_updated`, the sheet closes, a success toast shows, and the overview refreshes with no Stripe redirect; when it returns `checkout_created`, it redirects to Stripe. An owned product still shows "Manage in Stripe".
- Open a product that has both Pro and Enterprise tiers and confirm the Enterprise card shows a working "Contact sales" button.
- Confirm the top-right badge reflects the actual plan (a Pro org with the groups feature no longer reads "Enterprise").
- Confirm the Usage card shows the real machine-identity count, and the empty "Get started" card is gone for orgs with no subscription.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -240,10 +240,12 @@ export const licenseServiceFactory = ({
           // it non-fatal so a subscription read failure doesn't drop the org to the free fallback.
           try {
             const subscription = await licenseClient.getSubscription(rootOrgId);
-            const tiers = (subscription?.items ?? []).map((item) => item.plan.toLowerCase());
-            if (tiers.some((tier) => tier.includes("enterprise"))) {
+            const paidTiers = (subscription?.items ?? [])
+              .map((item) => item.plan.toLowerCase())
+              .filter((tier) => tier !== "free");
+            if (paidTiers.some((tier) => tier.includes("enterprise"))) {
               currentPlan.slug = "enterprise";
-            } else if (tiers.length > 0) {
+            } else if (paidTiers.length > 0) {
               currentPlan.slug = "pro";
             }
           } catch (error) {
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -67,7 +67,7 @@ type TLicenseServiceFactoryDep = {
   licenseDAL: TLicenseDALFactory;
   keyStore: Pick<TKeyStoreFactory, "setItemWithExpiry" | "getItem" | "deleteItem">;
   projectDAL: TProjectDALFactory;
-  licenseClient?: Pick<TLicenseClientFactory, "getEntitlements">;
+  licenseClient?: Pick<TLicenseClientFactory, "getEntitlements" | "getSubscription">;
   licenseDualRead?: Pick<TDualReadServiceFactory, "compareInBackground">;
 };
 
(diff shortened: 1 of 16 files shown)


backend/src/ee/services/license/license-service.ts
```diff
@@ -234,6 +234,21 @@ export const licenseServiceFactory = ({
...
+          // subscription tier; otherwise the org-level plan label can't reflect the real tier. Keep
+          // it non-fatal so a subscription read failure doesn't drop the org to the free fallback.
+          try {
+            const subscription = await licenseClient.getSubscription(rootOrgId);
+            const tiers = (subscription?.items ?? []).map((item) => item.plan.toLowerCase());
+            if (tiers.some((tier) => tier.includes("enterprise"))) {
+              currentPlan.slug = "enterprise";
+            } else if (tiers.length > 0) {
+              currentPlan.slug = "pro";
+            }
```
scott-ray-wilson: can't this incorrectly label a free tier as Pro?

seems like else where we reference the free tier 

Image

backend/src/services/license-client/license-client-types.ts
```diff
@@ -109,6 +109,18 @@ export const sessionResponseSchema = z
...
+// directly to an existing subscription (subscription_updated). `url` is the legacy shape, tolerated so
+// the cloud and license server can deploy in any order.
+export const checkoutResultSchema = z
+  .object({
+    outcome: z.enum(["checkout_created", "subscription_updated"]).optional(),
+    checkoutUrl: z.string().optional(),
+    subscriptionId: z.string().optional(),
+    url: z.string().optional()
+  })
+  .passthrough();
```
scott-ray-wilson, line 120: question: a little confused how this would be tolerated, we're updating the backend and frontend here to not account for url other than here?

scott-ray-wilson (changes requested):
intitial code review comments, pending application testing

backend/src/ee/services/license/license-service.ts
PrestigePvP: Good catch, fixed in 0e1e702. A `free`-plan subscription item was counting toward the `else if (items.length > 0)` branch and getting stamped `pro`. The derivation now filters out the `free` tier first, so a free org keeps the projection default (null slug -> "Free").

backend/src/services/license-client/license-client-types.ts
PrestigePvP: Agreed, removed in 0e1e702. The `url` field was only there as a transitional fallback for deploy ordering, but since neither system is deployed yet we can treat the two-outcome response as the canonical contract. `outcome` is now required and `url` is gone, so the schema matches the new shape everywhere.

scott-ray-wilson (dismissed):
LGTM

scott-ray-wilson (approved):

PR #6975, opened 2026-06-22
fix: check all orgs before failing request

mathnogueira:
## Steps to verify the change

* Create a secret in a suborg
* Share secret with Org
* Open it in another account that has access to said root org and suborg and has root org selected at the time
* Secret opens normally

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/secret-sharing/secret-sharing-service.ts
+++ b/backend/src/services/secret-sharing/secret-sharing-service.ts
@@ -679,7 +679,13 @@ export const secretSharingServiceFactory = ({
   };
 
   /** Gets password-less secret. validates all secret's requested (must be fresh). */
-  const accessSharedSecret = async ({ sharedSecretId, orgId, actorId, actor, password }: TGetActiveSharedSecretByIdDTO) => {
+  const accessSharedSecret = async ({
+    sharedSecretId,
+    orgId,
+    actorId,
+    actor,
+    password
+  }: TGetActiveSharedSecretByIdDTO) => {
     const result = await secretSharingDAL.transaction(async (tx) => {
       await tx.raw("SELECT pg_advisory_xact_lock(?)", [PgSqlLock.AccessSharedSecret(sharedSecretId)]);
 
--- a/backend/src/services/secret-sharing/secret-sharing-service.ts
+++ b/backend/src/services/secret-sharing/secret-sharing-service.ts
@@ -1,7 +1,7 @@
 import { ForbiddenError } from "@casl/ability";
 import { Knex } from "knex";
 
-import { OrganizationActionScope, TOrganizations, TSecretSharing } from "@app/db/schemas";
+import { OrganizationActionScope, OrgMembershipStatus, TOrganizations, TSecretSharing } from "@app/db/schemas";
 import { TLicenseServiceFactory } from "@app/ee/services/license/license-service";
 import { OrgPermissionSecretShareAction, OrgPermissionSubjects } from "@app/ee/services/permission/org-permission";
 import { TPermissionServiceFactory } from "@app/ee/services/permission/permission-service-types";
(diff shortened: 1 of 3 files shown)


backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -670,12 +685,13 @@ export const secretSharingServiceFactory = ({
...
-      if (accessType === SecretSharingAccessType.Organization && orgId !== sharedSecret.orgId) {
-        throw new ForbiddenRequestError();
+      if (accessType === SecretSharingAccessType.Organization) {
+        if (orgId === undefined) {
+          throw new UnauthorizedError();
+        }
+        if (sharedSecret.orgId) {
+          await $assertOrgAccess(sharedSecret.orgId, orgId, actorId);
+        }
       }
```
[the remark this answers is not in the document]
mathnogueira: False positive: if `sharedSecret.orgId` is empty, it means a secret that can shared to external users.

mathnogueira:
@greptile review this once more

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -107,6 +107,18 @@ export const secretSharingServiceFactory = ({
   const $hasExternalEmailAccess = (sharedSecret: TSecretSharing): boolean =>
     Boolean(sharedSecret.allowExternalEmails && sharedSecret.password);
 
+  // Verifies the actor has access to secretOrgId by checking they share the same root org.
+  const $assertOrgAccess = async (secretOrgId: string, actorOrgId: string): Promise<void> => {
+    if (secretOrgId === actorOrgId) return;
+    const [actorRootOrg, secretRootOrg] = await Promise.all([
+      orgDAL.findRootOrgDetails(actorOrgId),
+      orgDAL.findRootOrgDetails(secretOrgId)
+    ]);
```
mathnogueira, line 121: This covers 3 important cases, considering we have a root org and a couple of sub orgs and user has access to all. The secret is shared from a sub org.

1. User has access to both root org and sub-org, but is currently logged in the root org: it can still see the secret as it has access to the sub-org
2. User is currently in a different sub org from the same root: user has access to the secret
3. User has logged in a different org: cannot see the secret. This is done because org might require MFA and it would be unsafe to allow user to see the secret without requiring it to be logged in that org.

mathnogueira:
@veria-ai check this again, please

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -107,6 +107,18 @@ export const secretSharingServiceFactory = ({
   const $hasExternalEmailAccess = (sharedSecret: TSecretSharing): boolean =>
     Boolean(sharedSecret.allowExternalEmails && sharedSecret.password);
 
+  // Verifies the actor has access to secretOrgId by checking they share the same root org.
+  const $assertOrgAccess = async (secretOrgId: string, actorOrgId: string): Promise<void> => {
+    if (secretOrgId === actorOrgId) return;
```
mathnogueira, line 117: The reason I left this here (original check) was because this is a cheap check. No DB check or anything. So it's a good default for most users. If that fails, we check the DB for the fallback.

backend/src/services/secret-sharing/secret-sharing-service.ts
```diff
@@ -107,6 +107,18 @@ export const secretSharingServiceFactory = ({
...
+    const [actorRootOrg, secretRootOrg] = await Promise.all([
+      orgDAL.findRootOrgDetails(actorOrgId),
+      orgDAL.findRootOrgDetails(secretOrgId)
+    ]);
+    if (!actorRootOrg || !secretRootOrg || actorRootOrg.id !== secretRootOrg.id) {
+      throw new ForbiddenRequestError({ message: "You do not have access to this secret" });
+    }
+  };
```
adilsitos, line 137: I believe there is a better way of validating this. We can check this by using the membership table, making sure that the actor requesting it has access into this organization. I feel like this if statement at the end might be harder to understand in the future.

adilsitos (changes requested):

adilsitos (approved):

PR #6981, opened 2026-06-23
feat(pam): registry-driven template policies

lb-vn:
## Context

Registry-driven template policies

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/db/migrations/20260603030053_pam-revamp.ts
+++ b/backend/src/db/migrations/20260603030053_pam-revamp.ts
@@ -32,7 +32,6 @@ const seedDefaultTemplates = async (knex: Knex, projectId: string) => {
         projectId,
         name: template.name,
         type: template.type,
-        accessPolicy: template.accessPolicy,
         settings: template.settings
       })
       .onConflict(["projectId", "name"])
@@ -85,7 +84,7 @@ export async function up(knex: Knex): Promise<void> {
     t.string("type").notNullable();
     t.index("type");
 
-    t.jsonb("accessPolicy").nullable();
+    t.jsonb("policies").nullable();
     t.jsonb("settings").nullable();
 
     t.uuid("gatewayId").nullable();
--- a/backend/src/ee/services/pam/pam-policies.test.ts
+++ b/backend/src/ee/services/pam/pam-policies.test.ts
@@ -1,0 +1,41 @@
+import { describe, expect, test } from "vitest";
+
+import { PamAccountType } from "./pam-enums";
+import { getApplicablePolicies, PamPolicyType, validatePolicyValues } from "./pam-policies";
+
+describe("getApplicablePolicies", () => {
+  test("returns the universal policies for any account type", () => {
+    const keys = getApplicablePolicies(PamAccountType.Postgres).map((p) => p.key);
+    expect(keys).toContain(PamPolicyType.RequireMfa);
+    expect(keys).toContain(PamPolicyType.RequireReason);
+    expect(keys).toContain(PamPolicyType.MaxSessionDuration);
+  });
+});
+
+describe("validatePolicyValues", () => {
+  test("returns empty data for null/undefined", () => {
+    expect(validatePolicyValues(PamAccountType.SSH, null)).toEqual({ ok: true, data: {} });
+    expect(validatePolicyValues(PamAccountType.SSH, undefined)).toEqual({ ok: true, data: {} });
+  });
+
+  test("accepts valid policy values", () => {
+    expect(
+      validatePolicyValues(PamAccountType.SSH, {
+        [PamPolicyType.RequireMfa]: true,
+        [PamPolicyType.MaxSessionDuration]: 3600
+      })
+    ).toEqual({
+      ok: true,
+      data: { [PamPolicyType.RequireMfa]: true, [PamPolicyType.MaxSessionDuration]: 3600 }
+    });
+  });
+
+  test("rejects an unknown policy", () => {
+    expect(validatePolicyValues(PamAccountType.SSH, { "not-a-policy": true }).ok).toBe(false);
+  });
+
+  test("rejects an invalid value shape", () => {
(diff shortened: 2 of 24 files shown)


backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -85,7 +84,7 @@ export async function up(knex: Knex): Promise<void> {
     t.string("type").notNullable();
     t.index("type");
 
-    t.jsonb("accessPolicy").nullable();
+    t.jsonb("policies").nullable();
```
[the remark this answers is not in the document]
x032205: This is not yet in main so I think it's okay

backend/src/db/migrations/20260603030053_pam-revamp.ts
```diff
@@ -85,7 +84,7 @@ export async function up(knex: Knex): Promise<void> {
     t.string("type").notNullable();
     t.index("type");
```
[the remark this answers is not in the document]
x032205: Not yet in main so I think it's okay

frontend/src/pages/pam/PamAccountAccessPage/ReasonGate.tsx
```diff
@@ -9,23 +9,22 @@ import {
...
-
 export const resolveAccessPolicy = (account: TPamAccount): TAccessPolicy => {
-  const templatePolicy = account.templateAccessPolicy as Partial<TAccessPolicy> | null;
-  return { ...DEFAULT_ACCESS_POLICY, ...templatePolicy };
+  const policies = (account.templatePolicies ?? {}) as Record<string, unknown>;
+  const duration = policies[PamPolicyType.MaxSessionDuration];
+  return {
+    requireReason: policies[PamPolicyType.RequireReason] === true,
+    requireMfa: policies[PamPolicyType.RequireMfa] === true,
+    maxSessionDurationSeconds: typeof duration === "number" ? duration : 3600
```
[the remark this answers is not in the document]
x032205: Not an issue

backend/src/ee/services/pam/pam-policies.ts
```diff
@@ -0,0 +1,103 @@
...
+export const validatePolicyValues = (
+  accountType: PamAccountType,
+  policies: Record<string, unknown> | null | undefined
+): TValidatePolicyValuesResult => {
+  if (!policies) return { ok: true, data: {} };
+
+  const data: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(policies)) {
+    if (!(key in PAM_POLICY_DEFINITIONS)) {
+      return { ok: false, message: `Unknown PAM policy '${key}'` };
```
saifsmailbox98: can we also return the valid ones here in the message? for example, `"Unknown PAM policy 'require-mfaa'. Valid policies for ssh: require-mfa, require-reason, max-session-duration" `

backend/src/ee/services/pam/pam-policies.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    const policy = key as PamPolicyType;
+    if (!policyAppliesTo(policy, accountType)) {
+      return { ok: false, message: `Policy '${key}' does not apply to account type '${accountType}'` };
+    }
+
+    const parsed = PAM_POLICY_DEFINITIONS[policy].schema.safeParse(value);
+    if (!parsed.success) {
+      return { ok: false, message: `Invalid configuration for policy '${key}'` };
```
saifsmailbox98: This suppresses the zod error details. The user gets no hint about what's wrong. Can we include atleast one issue in this message?

backend/src/ee/services/pam/pam-policies.ts
```diff
@@ -0,0 +1,103 @@
...
+};
+
+export type TPamAccessControls = {
+  requireReason: boolean;
+  requireMfa: boolean;
+  maxSessionDurationSeconds: number | null;
+};
+
+export const resolveAccessControls = (policyMap: unknown): TPamAccessControls => {
+  const duration = resolvePolicy(policyMap, PamPolicyType.MaxSessionDuration);
```
saifsmailbox98, line 108: when we add type-scoped policies, would we need to add accountType to this function signature?

backend/src/ee/services/pam/pam-policies.test.ts
```diff
@@ -0,0 +1,41 @@
+import { describe, expect, test } from "vitest";
```
saifsmailbox98, line 1: can we add test coverage for `resolveAccessControls`? I think it has some interesting trivial behaviors that could cause data corruption if changed by mistake during development by someone.

frontend/src/pages/pam/components/policyEditors/DurationPolicyEditor.tsx
```diff
@@ -0,0 +1,26 @@
...
+  description,
+  value,
+  onChange
+}: PolicyEditorProps) => (
+  <Field>
+    <FieldLabel>{label}</FieldLabel>
+    <FieldContent>
+      <Input
+        type="number"
+        placeholder="e.g. 3600 (leave empty for no limit)"
```
saifsmailbox98: Placeholder says "leave empty for no limit" but empty actually falls back to 1 hour.


Also the description says the max is 86400. So when an admin leaves this empty, does "no limit" mean indefinitely, 86400 (the max), or 1 hour? right now it's 1 hour but the UI doesn't say that.


 Also, pinching in or out on the input changes the number value

saifsmailbox98 (changes requested):

backend/src/ee/services/pam/pam-policies.ts
x032205: No. Write-time validation (`validatePolicyValues`) already blocks storing a policy that doesn't apply to the type, so the stored map only ever has valid policies. This function just reads by key, and anything missing falls back to a default. Adding a type-scoped policy doesn't change that, so no `accountType` needed here

saifsmailbox98 (approved):

PR #6982, opened 2026-06-23
fix(pki): honor CA-issuance policy denial on certificate renewal

### LINKED ISSUE #6971: Certificate renewal bypasses CA-issuance policy denial
### Product

Certificate Manager (PKI)

### Describe the bug

When renewing a certificate via **certificate-v3-service.ts**, the renewal path validates the request against the certificate policy via `certificatePolicyService.validateCertificateRequest`, but the certificateRequest object it builds **(around certificate-v3-service.ts:2417-2434)** never populates basicConstraints from originalCert.isCA. 

The policy's CA-related checks in validateRequestAgainstPolicy **(certificate-policy-service.ts:~307)** are gated on **request.basicConstraints?.isCA === true**, which is always falsy for a renewal request — so a policy that denies CA issuance is silently bypassed on renewal, even though it's correctly enforced on direct issuance (certificate-v3-service.ts:1290, :1608) and on approval (certificate-approval-fns.ts:393).

### To Reproduce

- Create a certificate policy with CA issuance allowed (basicConstraints.isCA: "allowed").
- Issue a CA certificate under that policy.
- Update the policy to deny CA issuance (basicConstraints.isCA: "denied").
- Renew the previously-issued CA certificate.

### Expected behavior

Renewal should be rejected with the same "CA certificate issuance is not allowed by this policy" error that direct issuance and approval paths throw.

### Actual behavior
Renewal succeeds, issuing a new CA-flagged certificate despite the policy now denying it.

### Screenshots

_No response_

### Deployment Type

Infisical Cloud also : noting it reproduces on self-hosted too (since the code path is identical) so triage doesn't dismiss it as cloud-only.

### Additional context

The three existing CA-policy checks `(certificate-v3-service.ts:1290, :1608, certificate-approval-fns.ts:393)` are near-identical inline copies rather than calls to one shared function — worth extracting into a single validateCaIssuancePolicy(ca, policy, shouldIssueAsCA) helper in certificate-issuance-utils.ts so the renewal path (and any future issuance path) can't miss it again.

jaydeep-pipaliya:
Fixes #6971.

## What

The renewal path in `certificateV3Service` runs the same `certificatePolicyService.validateCertificateRequest` that direct issuance runs, but the `certificateRequest` it builds (`certificate-v3-service.ts:2457-2475`) never populated `basicConstraints` from the original certificate. `validateRequestAgainstPolicy` gates its CA-related checks on `request.basicConstraints?.isCA === true`, so the check was always falsy on renewal. A policy that denied CA issuance was silently bypassed when renewing an existing CA certificate.

Repro from the issue:

1. Create a policy with CA issuance allowed.
2. Issue a CA cert under that policy.
3. Tighten the policy to deny CA issuance.
4. Renew the existing CA cert → succeeds. The renewal still issues a CA-flagged certificate even though the active policy denies it.

The same denial fires correctly on the two other issuance paths (`certificate-v3-service.ts:1290` for direct issuance, `certificate-approval-fns.ts:393` for the approval flow), so renewal was the lone gap.

## Fix

Carry `originalCert.isCA` and `originalCert.pathLength` into the renewal `certificateRequest`. The policy's deny / allow / require + pathLength checks then run on renewal exactly the way they run on issuance.

## Not in this PR

The reporter also suggested extracting the three near-identical inline CA checks into a single shared helper so the next issuance path can't miss it. That is a sensible follow-up, but is out of scope here — keeping the fix minimal so it can land as a targeted patch.

## Checklist

- [ ] I have added tests (no existing unit tests on this service path; the policy check itself is already covered by the issuance path tests)

### DIFF
--- a/backend/src/services/certificate-v3/certificate-v3-service.ts
+++ b/backend/src/services/certificate-v3/certificate-v3-service.ts
@@ -2553,6 +2553,8 @@ export const certificateV3ServiceFactory = ({
             signatureAlgorithm: originalSignatureAlgorithm,
             keyAlgorithm: originalKeyAlgorithm,
             isFromProfile: true,
+            basicConstraints: originalCert.isCA ? { isCA: true, pathLength: originalCert.pathLength } : undefined,
+            pathLength: originalCert.pathLength ?? undefined,
             organization: originalCert.subjectOrganization || undefined,
             ou: originalCert.subjectOrganizationalUnit || undefined,
             country: originalCert.subjectCountry || undefined,
--- a/backend/src/services/certificate-v3/certificate-v3-service.ts
+++ b/backend/src/services/certificate-v3/certificate-v3-service.ts
@@ -2471,7 +2471,10 @@ export const certificateV3ServiceFactory = ({
           ttl
         },
         signatureAlgorithm: originalCert.signatureAlgorithm || undefined,
-        keyAlgorithm: originalCert.keyAlgorithm || undefined
+        keyAlgorithm: originalCert.keyAlgorithm || undefined,
+        basicConstraints: originalCert.isCA
+          ? { isCA: true, pathLength: originalCert.pathLength ?? undefined }
+          : undefined
       };
 
       let validationResult: { isValid: boolean; errors: string[] } = { isValid: true, errors: [] };


jaydeep-pipaliya:
Codex P1 was right — validation alone wasn't enough. Without forwarding `basicConstraints` + `pathLength` into `internalCaService.issueCertFromCa`, the renewal would have validated as CA but issued as a leaf, marking the original as renewed. Worse than the original bug.

Pushed af43762 that mirrors what the direct-issuance path already does at line 1315-1316: passes the original cert's isCA + pathLength through to the issuance call.

PR #6985, opened 2026-06-23
feat: implemented caching and etag support for Go

akhilmhdh:
## Context

This PR implements the following things in Go Sidecar
1. Etag implementation
2. Caching implementation
3. The new handler.go pattern in which the spec only api functions exist in handler
4. Test improvements for integration in which we group test by domain.

There cache keys will have a mismatch between computed for request as the serialization is different for both. This is good as it won't cause conflict.  

Closes PLATFOR-386 and PLATFOR-473

## Steps to verify the change

1. Add the `.env` 
```
GOLANG_SIDECAR_URL=http://host.docker.internal:4040
GO_SIDECAR_SHADOW_ENABLED=true
GO_SIDECAR_SHADOW_SAMPLE_RATE=75
```

2. Just do secrets request and you will see the shadow is success
3. Do make lint and make test and all should pass. Go etag and caching should work fine - but it should not be 

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend-go/internal/server/middlewares/etag.go
```diff
@@ -0,0 +1,118 @@
...
+	return `"` + base64.StdEncoding.EncodeToString(hash[:]) + `"`
+}
+
+func etagMatches(etag, ifNoneMatch string) bool {
+	etag = strings.TrimSpace(etag)
+	ifNoneMatch = strings.TrimSpace(ifNoneMatch)
+
+	if etag == ifNoneMatch {
+		return true
+	}
```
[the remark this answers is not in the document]
akhilmhdh: This is expected with nodejs behavoir

backend-go/internal/services/services.go
```diff
@@ -0,0 +1,131 @@
...
+		return nil, nil, err
+	}
+
+	// Initialize groups
+	s.initPlatform(ctx)
+	s.initSecrets(ctx)
+
+	cleanup := func() {
+		s.KMS.Close()
+	}
```
adilsitos, line 71: I think this can become too big if we start adding new cleanup. Do you see any problem if this is a method for the Service struct where we can call a cleanup operation? This way, on the `main.go` we can call this from the service instance.

backend-go/internal/keystore/keystore.go
```diff
@@ -19,62 +23,123 @@ type KeyStore interface {
...
+	// PgGetIntItem returns the integer value for a key from the PostgreSQL key_value_store table.
+	// Returns 0 if key doesn't exist or is expired.
+	PgGetIntItem(ctx context.Context, key string) (int64, error)
 }
 
-type redisKeyStore struct {
-	client redis.UniversalClient
+type keyStore struct {
+	redis redis.UniversalClient
+	db    pg.DB
```
adilsitos, line 50: nit: Here we define db, but this is actually Postgres right? Maybe we could have it as `postgres` instead of `db`, since the interface supports any db, so if in the future we need another db I believe it is more straightforward

backend-go/internal/keystore/keystore.go
```diff
@@ -19,62 +23,123 @@ type KeyStore interface {
...
 	}).Result()
 }
+
+func (k *keyStore) PgGetIntItem(ctx context.Context, key string) (int64, error) {
+	var integerValue *int64
+	err := k.db.Replica().QueryRow(ctx, `
+		SELECT "integerValue"
+		FROM key_value_store
+		WHERE key = @key
+		AND ("expiresAt" IS NULL OR "expiresAt" > NOW())
```
adilsitos, line 133: question: one pattern that I see a lot on our knex queries is that if we receive a transaction we hit the primary database instead of the replica.

I saw the usage of this and it seems taht we are always hitting the replica (`pgGetIntItem` on `keystore.ts`) but the increment (`pgIncrementBy.ts` can receive a transaction, which will bump the wrong one). In my opinion, this should always query the main database, so we know that a cache was invalidated.

backend-go/internal/keystore/memory.go
adilsitos, line 1: I believe that we are going to use this to save items in memory right? Not sure if it is necessary, but since we are using go we could have a go routine to check the expiry items, and invalidate them. This would be lightweight, and we could have a low frequency of when this can be done.

backend-go/internal/keystore/memory.go
```diff
@@ -49,13 +58,21 @@ func (m *MemoryKeyStore) GetItem(_ context.Context, key string) (string, error)
 func (m *MemoryKeyStore) SetExpiry(_ context.Context, key string, expiry time.Duration) (bool, error) {
```
adilsitos, line 58: it feels confusing to me that the setexpiry is the only method that applies changes on both hashes and items. Maybe we could have it with a different name? Or have an expiry for each field, one for hashes the other for items.

backend-go/internal/keystore/memory.go
```diff
@@ -49,13 +58,21 @@ func (m *MemoryKeyStore) GetItem(_ context.Context, key string) (string, error)
...
-	return true, nil
+
+	// Check hashes
+	if hash, ok := m.hashes[key]; ok {
+		hash.expiresAt = time.Now().Add(expiry)
+		return true, nil
+	}
+
+	return false, nil
 }
```
adilsitos, line 77: to avoid the micro seconds difference, we can compute the expiry before the map lookup and define them with the same value. 

```suggestion

    expiryAt := time.Now().Add(expiry)

	// Check regular items first
	if item, ok := m.items[key]; ok {
		item.expiresAt = expiryAt
		m.items[key] = item
		return true, nil
}

	// Check hashes
	if hash, ok := m.hashes[key]; ok {
		hash.expiresAt =  expiryAt
		return true, nil
	}

	return false, nil
}

```

backend-go/internal/keystore/memory.go
```diff
@@ -122,34 +139,77 @@ func (m *MemoryKeyStore) IncrementBy(_ context.Context, key string, value int64)
...
+}
+
 func (m *MemoryKeyStore) StreamAdd(_ context.Context, stream, id string, values map[string]string) (string, error) {
 	// Simplified implementation - just return a fake ID
 	return "0-0", nil
 }
 
+func (m *MemoryKeyStore) PgGetIntItem(_ context.Context, key string) (int64, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
```
adilsitos, line 194: is there a reason for us to lock on the get operations as well? would be to make sure that we are getting the most up to date value? makes sense for me, I would just like to know the thought process behind.

backend-go/internal/libs/cache/hash.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// HashBytes creates a URL-safe base64 SHA256 hash of raw bytes.
+func HashBytes(data []byte) string {
+	hash := sha256.Sum256(data)
+	encoded := base64.StdEncoding.EncodeToString(hash[:])
+	encoded = strings.ReplaceAll(encoded, "+", "-")
+	encoded = strings.ReplaceAll(encoded, "/", "_")
+	encoded = strings.TrimRight(encoded, "=")
+	return encoded
```
adilsitos, line 32: Is there a reason for not using `return base64.RawURLEncoding.EncodeToString(hash[:])`? It seems to do the same replace operations we do right here (it also remove the last `=`)

backend-go/internal/server/api/secrets/secret/gen.go
```diff
@@ -1136,6 +1156,14 @@ func (o *ListSecretsV4ServiceRequestOptions) Validate() error {
...
 		}
 	}
+
+	if o.Header != nil {
+		if v, ok := any(o.Header).(runtime.Validator); ok {
+			if err := v.Validate(); err != nil {
+				errors = errors.Append("Header", err)
+			}
+		}
+	}
```
adilsitos, line 1166: these "common" validations here seem to be the same on the endpoints defined here (and I believe that all secrets endpoint that have header and queries will need this), maybe we could factor this out in a separate function that we can just call here? This way we avoid duplicating the same common validation in every endpoint.

backend-go/internal/server/api/secrets/secret/list_secrets.go
```diff
@@ -40,23 +43,43 @@ type listSecretsInternalOpts struct {
...
+type listSecretsResponseWithETag struct {
+	Response    *listSecretsResponse
+	ETag        string
+	NotModified bool
 }
 
 // listSecrets is the unified internal method for listing secrets.
 // Both V3 and V4 handlers call this with different options.
-func (h *Handler) listSecrets(ctx context.Context, opts *listSecretsInternalOpts) (*listSecretsResponse, error) {
+func (h *Handler) listSecrets(ctx context.Context, opts *listSecretsInternalOpts) (*listSecretsResponseWithETag, error) {
```
adilsitos, line 64: I got a little confused here. I saw that we call this on the handler.go (that lives on this same folder.  But all the operation that happen here seem to be `service` operations, that should live on the service layer, is there a reason for us to have it here? I check how the node server is defined, and this lives in the service.

backend-go/internal/server/api/secrets/secret/list_secrets.go
adilsitos, line 326: There is a type error here for me. 

Image

Can't we use something like this?  

Image

backend-go/internal/services/permission/permission.go
```diff
@@ -789,6 +802,167 @@ func (s *Service) findIdentityName(ctx context.Context, identityID uuid.UUID) (s
...
+	ProjectID string
+	OrgID     uuid.UUID
+	ActorID   uuid.UUID
+	ActorType auth.ActorType
+}
+
+// GetPermissionFingerprint returns a lightweight hash of membership DB rows for ETag validation.
+// This is faster than computing the full permission rules hash since it only reads IDs + timestamps.
+// Port of permission-dal.ts:947-1050.
+func (s *Service) GetPermissionFingerprint(ctx context.Context, args *GetPermissionFingerprintArgs) (string, error) {
```
adilsitos, line 816: I noticed that at the moment we are not adding the `tx` parameter to hit the main database, is this going to be implemented in another PR in the future?

backend-go/internal/server/api/secrets/secret/handler.go
```diff
@@ -0,0 +1,326 @@
...
+		IfNoneMatch:               ifNoneMatch,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if result.NotModified {
+		resp := NewListSecretsV4ResponseData(&ListSecretsV4Response{
+			Secrets: []SecretRaw{},
+		})
```
adilsitos, line 148: if the result is a 304, this should not return the body right? Only the status code with the right header

backend-go/internal/server/api/secrets/secret/handler.go
```diff
@@ -0,0 +1,326 @@
...
+		secrets:     deps.Secrets,
+		kms:         deps.KMS,
+		secretCache: deps.SecretCache,
+	}
+}
+
+// --- OpenAPI Spec Methods ---
+
+// ListSecretsV4 is the handler for listing secrets (V4).
+func (h *Handler) ListSecretsV4(ctx context.Context, opts *ListSecretsV4ServiceRequestOptions) (*ListSecretsV4ResponseData, error) {
```
adilsitos, line 103: I did a check here, and when I try to list secrets using the sidecar, the returned status code is different from the nodejs implementation. Not sure if this is intended at the moment, but the etag header being received is the same on both servers.

go logs (I add my name so it is easier to find, do not judge me) The value in front of the header is the cache key.

Image

node server: 
```
[17:56:19.567] ERROR (1131): [Shadow:error] Status code mismatch [requestId=req-4ZwOsSR08Aj23L] [route=/api/v4/secrets] [nodeStatus=304] [goStatus=200] {"reqId":"req-4ZwOsSR08Aj23L","orgId":"4d7e15a6-16ad-4998-b336-91475c820c99","severity":"ERROR","requestId":"req-4ZwOsSR08Aj23L","route":"/api/v4/secrets","nodeStatus":304,"goStatus":200}
```

backend-go/internal/server/api/secrets/secret/handler.go
```diff
@@ -0,0 +1,326 @@
...
+	})
+	if result.ETag != "" {
+		resp.Headers = make(http.Header)
+		resp.Headers.Set("ETag", result.ETag)
+	}
+	return resp, nil
+}
+
+// ListSecretsRawV3 is the handler for listing raw secrets (V3, deprecated).
+func (h *Handler) ListSecretsRawV3(ctx context.Context, opts *ListSecretsRawV3ServiceRequestOptions) (*ListSecretsRawV3ResponseData, error) {
```
adilsitos, line 166: I had the same response difference on this case. And the etag being sent is the same. It might be because the way that the cache is calculated is different between both cases, but maybe it is worth checking this, because we might end up with several shadow errors on logs.

backend-go/internal/server/api/secrets/secret/handler.go
```diff
@@ -0,0 +1,326 @@
...
+		IfNoneMatch:               ifNoneMatch,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if result.NotModified {
+		resp := NewListSecretsRawV3ResponseData(&ListSecretsRawV3Response{
+			Secrets: []SecretRaw{},
+		})
```
adilsitos, line 216: This is also returning the empty body, shouldn't we return only the headers with the 304 status code?

backend-go/internal/server/api/secrets/secret/handler.go
adilsitos, line 1: Regarding the cache problems. The cache being created by node is different thatn the one created by go, that is why I faced those problems. Not sure if this is intended, but maybe they should be the same? Just something we can keep in mind

backend-go/internal/services/secrets/secretcache/secretcache.go
adilsitos, line 1: from claude: reason for the difference between the cachekey on nodejs and go: 

```
  The two never agree, so a client moving between the Go and Node backends
  (or Go reading an ETag Node wrote) will never get a 304, and neither
  will hit the other's cached payload.

  1. requestParamsHash diverges — this breaks the ETag field and the cache
  key

  This is the most likely thing you're actually seeing. The ETag is stored
  in Redis under field actorId:permissionFingerprint:requestParamsHash.
  That hash is computed completely differently:

  Go (list_secrets.go:102) passes a map[string]any to GenerateHash →
  json.Marshal. Go's json.Marshal sorts map keys alphabetically. Node's
  JSON.stringify uses insertion order. So even with identical keys/values
  the JSON strings differ → different SHA → different field.

  But the keys aren't even identical:

  Go map: environment, path, recursive, includeImports,
    expandSecretReferences, viewSecretValue, personalOverridesBehavior,
    tagSlugs, metadataFilter
  Node object: environment, path, recursive, includeImports,
    expandSecretReferences, expandPersonalOverrides,
    personalOverridesBehavior, secretImportReferencesBehavior,
    viewSecretValue, throwOnMissingReadValuePermission, ...params

```

backend-go/internal/libs/cache/hash.go
```diff
@@ -0,0 +1,33 @@
...
+
+// GenerateHash creates a URL-safe base64 SHA256 hash of the JSON-serialized data.
+// Port of generateCacheKeyFromData from backend/src/lib/crypto/cache.ts.
+func GenerateHash(data any) string {
+	jsonBytes, err := json.Marshal(data)
+	if err != nil {
+		return ""
+	}
+	return HashBytes(jsonBytes)
+}
```
adilsitos, line 18: Maybe we should return the error here? Or maybe even panic this.

backend-go/internal/services/permission/permission.go
```diff
@@ -789,6 +802,167 @@ func (s *Service) findIdentityName(ctx context.Context, identityID uuid.UUID) (s
...
+	if args.ActorType == auth.ActorTypeUser {
+		actorCondition = `(membership."actorUserId" = @actorID OR membership."actorGroupId" IN (` + groupSubquery + `))`
+	} else {
+		actorCondition = `(membership."actorIdentityId" = @actorID OR membership."actorGroupId" IN (` + groupSubquery + `))`
+	}
+
+	// Build additional privilege join condition
+	var apActorCondition string
+	if args.ActorType == auth.ActorTypeIdentity {
+		apActorCondition = `membership."actorIdentityId" = addlPriv."actorIdentityId"`
```
adilsitos, line 844: got this with the help of claude, but it seems to make sense. We had a similar problem with secret access, where we were not checking users inside projects though a group. 

```
     "summary": "GetPermissionFingerprint joins additional_privileges on
  a column-to-column predicate (membership.actorUserId =
  addlPriv.actorUserId) instead of the literal @actorID used by the fixed
  getPermission query — so it misses additional-privilege changes for
  actors whose only project access is via a group, leaving the ETag
  fingerprint unchanged.",


      "failure_scenario": "A user whose project access is solely through a
  group is granted (or has revoked) an additional privilege. No secret
  write occurs, so invalidateSecretCacheByProjectId is never called and
  the ETag hash is not deleted. Group-derived membership rows have
  actorUserId = NULL, so this join drops the privilege from the
  fingerprint and it stays identical. On the next ListSecrets with
  If-None-Match, the 304 fast-path (secretcache.go:86-96) — which keys
  purely on actorID:fingerprint:paramsHash — matches the stored ETag and
  returns 304 Not Modified. The client keeps a stale secret list
  reflecting the old privilege for up to the 15m ETag TTL / UTC-day
  boundary. The sibling getPermission query (line 598-602) was
  deliberately changed to the literal form for exactly this case, and the
  TS reference (permission-dal.ts:976) also uses the literal actorId — the
  fingerprint port did not get the fix."
```

adilsitos (commented):
some comments. I think the most imported one would be how the cache is being calculated

backend-go/internal/server/api/secrets/secret/gen.go
akhilmhdh: This is a generated file @adilsitos  from the library based of the openapi

adilsitos: 😢 gotcha.

backend-go/internal/keystore/memory.go
akhilmhdh: I just implemented it for testing only. But we don't use it as of now. So we can kinda ingore it. I can delete it as well. But it's just a stub implementation

backend-go/internal/keystore/keystore.go
akhilmhdh: Not sure about another database. I used `pg.DB` because of two reasons. 
1. This is something used everywhere in the app. So most of them will get it as pg itself.
2. I feel it would be strange to call is `pg.Postgres`

backend-go/internal/services/services.go
akhilmhdh: Great suggestion!!

backend-go/internal/libs/jitter/jitter.go
```diff
@@ -0,0 +1,17 @@
+// Package jitter applies random jitter to time durations, used to spread out
```
lobocv, line 1: nit: one thing I found helpful for the understanding of the codebase is to minimize the number of global general-purpose utility functions and packages like this one. A little bit of duplication, in my opinion, is worth the added cohesion of packages. 

For example, this jitter, is going to be used for cache TTLs. The other place that I could see it being used is on retries. I think since the library is so small (just a single Apply func), it can probably just be duplicated into each of those libraries themselves as a private function.

backend-go/internal/libs/cache/hash.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+)
+
+// GenerateHash creates a URL-safe base64 SHA256 hash of the JSON-serialized data.
+// Port of generateCacheKeyFromData from backend/src/lib/crypto/cache.ts.
+func GenerateHash(data any) string {
```
lobocv, line 12: Same comment here. Having this as a global package also forces the hash function to be the same everywhere. Since hashing is basically in the standard library, I don't think we need to wrap it.

backend-go/tests/infra/nodejs/access_tokens.go
```diff
@@ -0,0 +1,113 @@
+//go:build integration
+
+package nodejs
```
lobocv, line 3: I'd suggest renaming the package to `e2e`. NodeJS is an implementation detail. When the backend gets written in Go then this would no longer make sense.

backend-go/internal/services/secrets/secretcache/secretcache.go
```diff
@@ -0,0 +1,231 @@
...
+func NewService(_ context.Context, logger *slog.Logger, deps *Deps) *Service {
+	return &Service{
+		logger:   logger.With(slog.String("service", "secretcache")),
+		keyStore: deps.KeyStore,
+	}
+}
+
+// CheckListSecrets checks for a cached list secrets response or ETag match.
+// Returns nil if no cache hit.
+func (s *Service) CheckListSecrets(
```
lobocv, line 77: Question: Have we considered using the postgres read replica as the "cache"? Are we optimizing for latency or performance here?

backend-go/internal/services/secrets.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// SecretsGroup holds secrets-related services.
+type SecretsGroup struct {
+	Secret       *secret.Service
+	SecretFolder *secretfolder.Service
+	SecretImport *secretimport.Service
+	SecretCache  *secretcache.Service
+	Environment  *environment.Service
+}
```
lobocv, line 20: Good idea to group these "services" together. This looks to be everything in Secrets Management. We talked about renaming this grouping to "Service" and having the components be called something else (they're really just internal libraries)

Just a suggestion, we could call the main service that we're mounting SecretsGroup to the "Server" as it's the one hosting the transport layer. It makes sense to me because we deploy one server and servers usually have one or more services

backend-go/internal/services/services.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/infisical/api/internal/keystore"
+	"github.com/infisical/api/internal/queue"
+	"github.com/infisical/api/internal/services/assumeprivilege"
+	"github.com/infisical/api/internal/services/auditlog"
+	"github.com/infisical/api/internal/services/kms"
+	"github.com/infisical/api/internal/services/permission"
+)
+
+// Infra holds the external infrastructure dependencies.
+type Infra struct {
```
lobocv, line 24: As I mentioned before, there are a lot of benefits to a layered architecture. If we go that route, I would love to see this abstracted a bit more, so that the infrastructure is exposed through interfaces and the implementation details are hidden from the application layer. Right now, we sort of do this with key store and HSM service, but some of the infrastructure implementations are leaking through a little bit with Postgres and the redis.UniversalClient

backend-go/internal/services/services.go
```diff
@@ -0,0 +1,131 @@
...
+	Redis    redis.UniversalClient
+	HSM      kms.HsmService
+	License  *license.Service
+	KeyStore keystore.KeyStore
+	Queue    *queue.Service
+}
+
+// Services holds all initialized services.
+type Services struct {
+	infra *Infra
```
lobocv, line 37: I would consider having the infrastructure dependency injected into the service's rather than staying on the Server. Otherwise service's will have a dependency on the Server as well as the other way around. It's a good idea to always have dependencies flow one way, Server ---> Service

backend-go/internal/services/services.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// Top-level services (frequently used across handlers)
+	KMS             *kms.Service
+	Permission      *permission.Service
+	AuditLog        *auditlog.Service
+	AssumePrivilege *assumeprivilege.Service
+	RateLimit       *ratelimit.Service
+	License         *license.Service
+
```
lobocv, line 46: These seems like infrastructure concerns to me.

backend-go/llm/TESTING_GUIDELINE.md
```diff
@@ -248,56 +273,38 @@ func TestMain(m *testing.M) {
 }
 ```
 
-### Writing an integration test
+### Per-test isolation + seed facade
 
-Each test gets a clean transaction that rolls back at the end, so tests never pollute each other:
+The stack is shared, so isolation comes from each test seeding its **own** project (and
+identities/secrets under it) rather than transaction rollback. Seed through the
```
lobocv, line 279: We discussed a fixture generation library used solely by tests. A goal is to reduce boilerplate that distracts from the intent of the test when each test seeds it's own fixtures. We will need to spend time to get this right because it's a key balance between optimizing for readability and coupling. As long as it's opt in and bare bones then we should be alright. It's something to be mindful of when designing the fixture generator

Not sure what you mean by "rather than transaction rollback". Can you clarify that?

backend-go/llm/TESTING_GUIDELINE.md
```diff
@@ -248,56 +273,38 @@ func TestMain(m *testing.M) {
...
+		Build()
+	// ... exercise the Go handler against the seeded project ...
 }
 ```
 
-Always call `t.Helper()` so failure messages point to the calling test, not the seed function.
+Because every test owns a freshly-named project, suites run with `t.Parallel()` without
+cross-contamination. Add new seed endpoints as builders in the matching `tests/infra/nodejs`
+domain file (e.g. `secrets.go`, `identities.go`) so new params never break existing call
+sites.
```
lobocv, line 307: Not clear what this last sentence means.

backend-go/tests/infra/nodejs/access_tokens.go
```diff
@@ -0,0 +1,113 @@
...
+		ClientID string `json:"clientId"`
+	} `json:"identityUniversalAuth"`
+}
+
+// CreateClientSecretRequest is the request body for POST /api/v1/auth/universal-auth/identities/{id}/client-secrets.
+type CreateClientSecretRequest struct {
+	Description  string `json:"description"`
+	TTL          int    `json:"ttl"`
+	NumUsesLimit int    `json:"numUsesLimit"`
+}
```
lobocv, line 35: I like the *Request/*Response naming, it's very clear. I also like that it references the API endpoint and is transport agnostic (no HTTP leaking through) 👍🏽 
While not the highest priority right now, before we roll this out to teams to implement, we should figure out how we're going to be creating this client SDK and what guard rails we can put in to make sure it's uniform when each team is contributing to it

backend-go/tests/infra/http.go
```diff
@@ -0,0 +1,324 @@
...
+	"github.com/infisical/api/internal/services/auth"
+)
+
+// ClientBuilder constructs an HTTPClient with extensible configuration.
+type ClientBuilder struct {
+	t              *testing.T
+	router         http.Handler
+	identity       *TestIdentity
+	defaultHeaders http.Header
+}
```
lobocv, line 30: We may want to separate out the testing from the client SDK. If we want to write scripts or separately deployed jobs it would be very beneficial to have a SDK. Also, eventually i'm assuming that the SDK is going become a client facing product so that customers do not have to write their own. At that point we should more strongly consider generating clients from the OpenAPI spec so that we can easily support multiple languages

backend-go/tests/infra/http.go
```diff
@@ -0,0 +1,324 @@
...
+
+// NewClientBuilder creates a new client builder for the given router.
+func NewClientBuilder(t *testing.T, router http.Handler) *ClientBuilder {
```
lobocv, line 33: I don't quite understand this package and why we have a client builder and are hosting the httptest server inside the client? I thought we were testing against a real server?

backend-go/tests/secrets/secrets/list_secrets_expansion_test.go
```diff
@@ -0,0 +1,206 @@
...
+		require.NoError(t, err)
+		return resp.Secrets
+	}
+
+	t.Run("nested expansion", func(t *testing.T) {
+		secrets := expanded(t)
+
+		endpoint := findSecret(secrets, "ENDPOINT")
+		require.NotNil(t, endpoint)
+		assert.Equal(t, "myhost.com:5432", endpoint.SecretValue)
```
lobocv, line 70: `testify/assert` marks the failure but then continues on with the rest of the test, while `testify/require` immediately stops the test. I would highly recommend that we only use testify require, because continuing on with the test when one step fails doesn't really make sense. It adds a lot of noise and confusion to the output because you're going to see a cascading amount of failures, and you have to figure out which one is the first one that caused it. 

We can easily make a linting rule (forbidigo) to stop this systematically

backend-go/tests/secrets/secrets/list_secrets_expansion_test.go
```diff
@@ -0,0 +1,206 @@
...
+// The expansion algorithm (circular refs, self-reference, max depth, chained
+// absolute refs, import ordering) is covered by the unit tests in
+// internal/services/secrets/secret/expansion_test.go. These integration tests
+// cover the end-to-end wiring: expansion and imports resolved through the real
+// handler, DB, and permission layer for a full list response.
+
+// TestListSecrets_Imports and the import-response structure tests live in
+// list_secrets_imports_test.go.
+
+func TestListSecrets_Expansion(t *testing.T) {
```
lobocv, line 24: I like how clearly these tests read 👍🏽

PR #6989, opened 2026-06-23
feat(license-client): send org identity (name, slug, region) to entitlements call

PrestigePvP:
## Context

The License Server populates an org directory from the entitlements call; this sends the org identity it needs. Also refactors the v2 entitlements client to take a single org-like object so future fields are additive instead of new positional params.

- `getEntitlements`/`fetchEntitlements` now take `{ id; name?; slug? }` (only `id` required)
- Send `org_name`, `org_slug` (per-org) and `region` (from `INTERNAL_REGION`, baked into the backend at construction) on `GET /v1/entitlements`
- v1/legacy path untouched; entire path only active when `LICENSE_SERVER_V2_MODE != off`

## Steps to verify the change

- Differential `tsc --noEmit` over the backend: identical error count with and without this change, none in the changed files (this change introduces no new type errors).
- With `LICENSE_SERVER_V2_MODE != off`, an org reading entitlements produces a `GET /v1/entitlements?org_id=&org_name=&org_slug=&region=` request to the License Server.

## Type

- [x] Feature

## Checklist

- [x] Tested locally (typecheck; see steps above)

### DIFF
--- a/backend/src/ee/services/license-v2/license-v2-service.ts
+++ b/backend/src/ee/services/license-v2/license-v2-service.ts
@@ -9,6 +9,7 @@ import { TLicenseClientFactory } from "@app/services/license-client";
 import {
   TCatalogProduct,
   TCheckoutLineItem,
+  TEntitlementOrg,
   TSubscriptionResponse
 } from "@app/services/license-client/license-client-types";
 import { TOrgDALFactory } from "@app/services/org/org-dal";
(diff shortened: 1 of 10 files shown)


PrestigePvP:
Addressed the entitlement-cache propagation concern (greptile P1 / codex P2):

The resolver now forwards org identity **past** the cache. When `name`/`slug` are present and not yet synced this TTL window, it fetches directly (forwarding `org_name`/`org_slug`/`region` to the license server), refreshes the shared entitlements cache, and sets a per-org identity-synced key (30-min TTL). So a high-frequency identity-less feature check can no longer seed the cache and suppress identity sync — identity reaches the server at least once per TTL window. Type Check + integration pass.

Also: the **Lint** failure (prettier) is fixed, and history is squashed to a single commit. The **BDD** check failed on a GitHub Actions Docker build-cache fetch error (`buildx bake ... failed to copy ... actions-cache`), unrelated to this change; it should clear on re-run.

akhilmhdh (approved):

PR #6994, opened 2026-06-23
improvement: revamp access requests ui

scott-ray-wilson:
## Context

This PR revamps the access request tab UI in secrets management; migrating to v3 components and various UI/UX improvements

## Screenshots

CleanShot 2026-06-23 at 14 04 41@2x
CleanShot 2026-06-23 at 14 04 46@2x
CleanShot 2026-06-23 at 14 04 51@2x

## Steps to verify the change

 List renders in v3 (search, environment + type filters, pagination)
 Loading skeleton + empty states display correctly
 Create an access request works (temp/permanent, etc.)
approval flow (many reviews, single review, by pass, reject, approve, etc.)
 Edit request works

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/components/EditAccessRequestModal.tsx
adilsitos, line 1: not related to the PR. But I feel like `policies` should be inside access control, not inside `approvals` 

Image

backend/src/ee/services/access-approval-request/access-approval-request-fns.ts
adilsitos, line 1: @scott-ray-wilson not sure what was decided from your conversation with @varonix0 but if we are going to allow only dynamic secrets and static secrets to be received by this endpoint I feel like we should enforce this on the zod validation, making sure that nothing gets in. I'm afraid of someone sending something that the `humanizeSlug` does not catch, and the approver receives a wrong message.

adilsitos, line 1: We can send a request like this: 

```
{
    "permissions": [
      [
        "read-root-credential",
        "dynamic-secrets",
                                                                                                                                                                                                                                            {
                                                                                                                                                                                                                                                               "environment": "dev",
                                                                                                                                                                                                                                         "secretPath": {
            "$glob": "/",
            "$in": ["/private/**"]
          }
        }
      ],
      [
        "edit-root-credential",
        "dynamic-secrets",
        {
          "environment": "dev",
          "secretPath": {
            "$glob": "/",
            "$in": ["/private/**"]
          }
        }
      ]
    ],
    "isTemporary": false
  }'
``` 
Where we pass a `in` clause as well. Nothing happens, the requester does not get access into envs other than `/`, but we allow this request. Maybe we should validate that only the `glob` path is received. Can be tackled in another PR.

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/components/RequestAccessForm.tsx
```diff
@@ -0,0 +1,422 @@
...
+});
+
+type TSecretPermissionForm = z.infer<typeof secretPermissionSchema>;
+
+const PERMISSIONS = [
+  { name: "read", label: "View", description: "Read secret values", Icon: EyeIcon },
+  { name: "create", label: "Create", description: "Create new secrets", Icon: PlusIcon },
+  { name: "edit", label: "Modify", description: "Update existing secrets", Icon: PencilIcon },
+  { name: "delete", label: "Delete", description: "Delete existing secrets", Icon: Trash2Icon }
+] as const;
```
adilsitos, line 76: Something that confused me is that we say only `secrets`, maybe we should say that they are requesting access to `static secrets`? Maybe we can have a label saying this.

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/components/RequestAccessForm.tsx
adilsitos, line 1: nit. 

When we click to request access, the path could be filled with the current path 

https://github.com/user-attachments/assets/a7679bc2-0378-4e3f-a39a-2e9611f2e5bc

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
```diff
@@ -346,274 +452,337 @@ export const AccessApprovalRequest = ({
...
+            </Tabs>
+            <div className="flex flex-wrap items-center gap-2 2xl:mr-auto 2xl:flex-nowrap">
+              <InputGroup className="xl:w-[26rem]">
+                <InputGroupAddon>
+                  <SearchIcon />
+                </InputGroupAddon>
+                <InputGroupInput
+                  value={search}
+                  onChange={(e) => setSearch(e.target.value)}
+                  placeholder="Search by requesting user or environment..."
```
adilsitos, line 545: The search happens using environment slug, maybe we could change the copy?

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
adilsitos, line 1: nit: if I click on edit with a requester that has a `no access` role I get this modal 


Image

and it says that the approver was removed, but I believe this happens because I cannot see other users, maybe this could have another text in case the user cannot access anything?

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
adilsitos: The same copy exists on the review flow

backend/src/ee/services/access-approval-request/access-approval-request-fns.ts
adilsitos: If we are going to allow everything, maybe we could have a message like this on the email (and maybe on the UI as well?) saying that the user requested things outside of static secrets. Since this is not allowed right now on the UI, this could help the reviewer IMO.

Image

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/components/EditAccessRequestModal.tsx
scott-ray-wilson: hmm, not sure on this one, I'd run it by @varonix0 but I'd personally push back since it pertains to approval requests explicitly

backend/src/ee/services/access-approval-request/access-approval-request-fns.ts
scott-ray-wilson: we now list all the resources that are being requested, verified with @varonix0, we don't need to make it like a warning because all the permissions are enumerated in the email and UI

scott-ray-wilson: agree we should flag this for a separate ticket!

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/components/RequestAccessForm.tsx
scott-ray-wilson: We refer to the resource just as "Secrets" everywhere else (permission, access control, etc.) so I think its best to keep this consistent. Once we update the UI to update other resources I think that will also be clearer since I can see which others I can select

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
scott-ray-wilson: lol at your note

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/components/RequestAccessForm.tsx
scott-ray-wilson: We thought about this before but deferred because this only applies if it's an exact match and you might have others that have partial match ("/**", etc.) so we don't want to influence what they pick (at least what we decided at the time) if you think it's worth adding we can explore in a separate ticket!

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
scott-ray-wilson: I added support for searching by environment name!

adilsitos: I tried searching using `devel...` and nothing was returned for me. Let me check it again

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/components/RequestAccessForm.tsx
adilsitos: Gotcha!

frontend/src/pages/secret-manager/SecretApprovalsPage/components/AccessApprovalRequest/AccessApprovalRequest.tsx
adilsitos: I tried to break it

adilsitos (approved):

PR #7011, opened 2026-06-24
fix: query invalidation

varonix0:
## Context

Fixed query invalidation preventing the frontend from correctly getting the config when the user is in an authenticated state.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


mathnogueira (approved):
Tested it and it works. Thanks @varonix0!

PR #6957, opened 2026-06-19
feat: add SAN validation for workload identity

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: `type(scope): short description` (scope is optional, e.g., `fix: prevent crash on sync` or `fix(api): handle null response`). 
- [ ] Tested locally
- [ ] Updated docs (if needed)
- [ ] Updated CLAUDE.md files (if needed)
- [ ] Read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview)

PR #7021, opened 2026-06-24
feat(identity-tls-cert-auth): optional trust-anchor chain verification for client certificates

### LINKED ISSUE #7020: Identity TLS Cert Auth: optionally validate the client certificate chain against the configured CA as a trust anchor
## What I'm trying to do

I'd like to authenticate workloads to Infisical using their SPIRE X.509-SVIDs through TLS Certificate Auth, with no static credentials and no JWT.

A SPIRE workload already holds an X.509 identity whose chain looks like this:

```
leaf (SPIFFE SAN)  ->  SPIRE signing intermediate  ->  SPIRE root CA
```

That seems like a perfect fit for TLS Cert Auth — the workload presents its leaf, Infisical validates it against a CA. The problem is *which* CA you're allowed to configure.

## Where it breaks

TLS Cert Auth today is single-hop: the leaf must be signed **directly** by the configured `caCertificate`. In SPIRE, the cert that directly signs the leaf is the **signing intermediate**, and that intermediate rotates often — in our deployment roughly daily (SPIRE stages a "prepared" intermediate ahead of time and swaps it). The root, by contrast, is stable for years.

So to keep auth working, I'd have to configure the rotating intermediate as the `caCertificate` and run a controller that watches for each rotation and `PATCH`es the identity's CA on every swap. One per trust domain. We run ~31 clusters, each its own SPIRE trust domain, so that's 31 rotation controllers — each one a moving part that, if it falls behind, breaks authentication for that cluster.

Cross-cluster makes it worse: SPIRE federation only shares **roots**, never intermediates. So a central Infisical can't even obtain a peer cluster's current intermediate to sync it. The intermediate is reachable only from inside the cluster that owns it.

## What would fix it

The thing I can reliably get my hands on — locally and through the federated trust bundle — is the **stable root**. I just can't use it today, because single-hop verification won't accept a root that didn't directly sign the leaf.

The fix is the same pattern every other SPIFFE consumer (Envoy, Vault, ...) already uses: treat the configured CA as a **trust anchor** and let the verifier walk the chain the client presents up to it. Pin the root once, done — no sync controller, no static credentials, and it works the same way across clusters from the federated root.

## Proposal

Add an opt-in, per-identity boolean (something like `verifyClientCertificateChain`, default `false`):

* `false` **(today's behavior, unchanged):** the leaf must be signed directly by the configured CA.
* `true`**:** the configured CA is a trust anchor. The client presents `leaf + intermediate(s)` (exactly what a SPIRE X.509-SVID already contains), and Infisical validates a path from the leaf up to the anchor — checking the issuer relationship and signature at each hop, that issuers are CAs, and that every certificate on the path is within its validity window.

The configured CA stays the only trusted certificate. A presented intermediate is just an untrusted input: if it doesn't chain to the anchor, it's rejected. And everything [#6913](<https://github.com/Infisical/infisical/issues/6913>) added still holds — the leaf must be an end-entity certificate (a CA cert presented as the client cert is still refused), and SAN/CN checks ([#6957](<https://github.com/Infisical/infisical/issues/6957>)) still apply to the leaf.

## References

* [#3852](<https://github.com/Infisical/infisical/issues/3852>) — the original TLS Certificate Auth (single-hop, admin supplies the CA).
* [#6913](<https://github.com/Infisical/infisical/issues/6913>) — hardened it: end-entity leaf required, CA-as-client-cert rejected, cert sourced from the trusted proxy. This proposal keeps all of that.
* [#6957](<https://github.com/Infisical/infisical/issues/6957>) — SAN validation for workload identity, which this builds on for the SPIFFE URI SAN.
* Infisical/infisical#6907 / [#6908](<https://github.com/Infisical/infisical/issues/6908>) — earlier SPIFFE X.509-SVID context.

Happy to send a PR — I have an implementation with tests ready to go.

sinnwise:
## Context

TLS Certificate Auth verifies the client certificate single-hop: the presented leaf has to be signed directly by the configured `caCertificate`, and any intermediates the client sends are dropped (`extractX509CertFromChain(...)[0]`).

That's a problem for any PKI where the certificate that signs the leaf isn't long-lived. The case I hit is SPIRE X.509-SVIDs: the signing intermediate rotates frequently (about daily for us) while the root is stable for years. Single-hop forces you to configure the rotating intermediate as the CA and re-`PATCH` it on every rotation — and across many trust domains, or cross-cluster where federation only shares roots, that quickly stops being workable.

This PR adds an opt-in mode that lets you pin the stable root instead and have Infisical validate the chain the client presents up to it. The motivation is written up in more detail in #7020.

## What this changes

A new per-identity option, `verifyClientCertificateChain`, defaulting to `false`:

- **`false` (default):** unchanged — the configured CA must be the leaf's direct issuer.
- **`true`:** the configured CA is treated as a trust anchor. The client presents its `leaf + intermediate(s)` (which a SPIRE X.509-SVID already is), and login builds a path from the leaf through those intermediates up to the anchor (RFC 5280-style): issuer/subject + signature at each hop, CA basic-constraints on issuers, and validity windows on every certificate involved.

Nothing else about the flow moves. The intermediates already arrive in the same client-certificate header — `extractX509CertFromChain` parses a multi-cert PEM today, login just ignored everything past the leaf. In chain mode it now uses the rest as candidate issuers. No proxy or header changes.

On the surface: a toggle on the **Advanced** tab of the TLS Cert Auth form (off by default), with the same value shown in the read-only auth-method view.

## Security

I wanted to be sure this doesn't loosen anything from #6913, so the new branch is deliberately narrow:

- The branch only decides *how the leaf is anchored to the CA*. The end-entity-leaf requirement (a CA cert presented as the client cert is rejected), the expiry / not-yet-valid checks, and the CN / SAN (#6957) checks all sit **outside** the branch and run in both modes.
- The configured CA is still the only trusted certificate. A forged or unrelated intermediate can't construct a path to it, so it's rejected.
- The column defaults to `false`, so existing identities behave exactly as before — same code path, same `ca_verification_failed` reason code.
- The client certificate is still sourced from the trusted-proxy header; the nginx configs are untouched.

## Testing

Unit tests cover the path builder against real `leaf -> rotating-intermediate -> stable-root` chains: a valid chain to the root anchor, a leaf issued directly by the anchor, a missing intermediate, the wrong root, a forged extra intermediate presented alongside the real one, and expired / not-yet-valid leaf and intermediate. Backend and frontend both type-check and lint clean.

I also ran it end to end against a running instance (migration applied, real HTTP logins) and confirmed both the new behavior and that the defaults are untouched:

| scenario | expected | result |
| --- | --- | --- |
| attach with the flag omitted | stored `false` | ✓ |
| default single-hop, CA = direct issuer, leaf | 200 | ✓ |
| default single-hop, CA = root (not the issuer), leaf | 401 `ca_verification_failed` | ✓ |
| default mode, CA cert presented as the client cert | 401 `ca_certificate_not_allowed` | ✓ |
| chain mode, leaf + intermediate to root anchor | 200 | ✓ |
| chain mode, leaf only (no intermediate) | 401 | ✓ |
| chain mode, CA cert presented as the client cert | 401 `ca_certificate_not_allowed` | ✓ |

## Type

- [x] Feature

## Checklist

- [x] Tested locally (unit + end-to-end against a running instance)
- [x] Updated docs

Closes: #7020


backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
```diff
@@ -1,3 +1,132 @@
...
+    }
+
+    if (depth >= maxDepth) return { ok: false, reasonCode: "ca_verification_failed" };
+
+    // No path to the anchor was found through this node. Default to the generic reason, but
+    // surface a more specific validity failure if the only viable issuer was rejected for being
+    // outside its validity window (so an expired/not-yet-valid intermediate is reported as such).
+    let bestFailure: TVerifyClientCertificateChainResult = { ok: false, reasonCode: "ca_verification_failed" };
+
+    for (let i = 0; i < presentedChain.length; i += 1) {
```
Thiago-AS: Hey @sinnwise, thanks for the contribution! The only part that worries me is this bit of code. It ends up exploring permutations over the nodes. 

Because each branch gets its own copy of used, once a branch finishes exploring and returns, the certs it consumed are effectively "released" for sibling branches. As a result, the search doesn't globally remember that it has already explored a given cert, so it ends up re-exploring the same cert for every different prefix that can reach it.

Since this uses the synchronous crypto.verify(), an attacker could potentially block the entire Node.js event loop and take down the server by sending a single request containing a sufficiently large number of self-signed certificates that all share the same key pair and have the same subject == issuer.

Would having a shared visited set work here? memoize? So we take from O(N!) to O(N^2).

Thiago-AS (changes requested):

Thiago-AS (dismissed):

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
sinnwise: Memoized traversal results by certificate DER and added shared in-progress cycle tracking, so each certificate is expanded once rather than once per preceding path. Added a regression test for shared dead-end issuer paths. Fixed in PR #7021 on `feat/tls-cert-auth-chain-verification`: https://github.com/Infisical/infisical/blob/269ea7547141967a0b00c7d3aae2de3a00939f8b/backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts#L75-L141

Thiago-AS (approved):

Thiago-AS (approved):

PR #7027, opened 2026-06-25
fix(pki): honor CA-issuance policy denial on certificate renewal

### LINKED ISSUE #6971: Certificate renewal bypasses CA-issuance policy denial

Abdul-Moiz31:
## Context

Fixes #6971.

Certificate renewal in `certificateV3Service.renewCertificate` runs the same `certificatePolicyService.validateCertificateRequest` that direct issuance runs, but the `certificateRequest` it built never populated `basicConstraints`. `validateRequestAgainstPolicy`'s CA-related checks are gated on `request.basicConstraints?.isCA === true`, which was always falsy on renewal — so a policy that denies CA issuance was silently bypassed when renewing a previously-issued CA certificate.

Separately, even once validation is fixed, the renewal's call to `internalCaService.issueCertFromCa` never passed `basicConstraints`/`pathLength` at all — so the renewed certificate would have lost its CA extension regardless of what the policy said.

Repro from the issue:
1. Create a policy with CA issuance allowed.
2. Issue a CA cert under that policy.
3. Tighten the policy to deny CA issuance.
4. Renew the existing CA cert → previously succeeded and silently re-issued a CA-flagged cert.

## Fix

- `renewCertificate` now derives `shouldIssueAsCA` from `originalCert.isCA` and populates `certificateRequest.basicConstraints` (with `originalCert.pathLength`) before the generic policy validation runs.
- Added an explicit `assertCaIssuancePolicyAllowed(policy, shouldIssueAsCA)` guard right after that validation — the same defense-in-depth check the other three issuance paths (direct issuance, CSR/profile issuance, approval finalization) already run.
- The renewal's `issueCertFromCa` call now passes `basicConstraints: { isCA: true, pathLength: policy?.basicConstraints?.maxPathLength }` and `pathLength: originalCert.pathLength`, so a renewed CA certificate keeps its CA status and gets re-validated against the *current* policy's max path length and the issuing CA's own constraints — not just blindly copied forward.

## Refactor (per the issue's suggestion)

The CA-denial check existed as three near-identical inline copies in `certificate-v3-service.ts` (direct issuance, CSR/profile issuance) and `certificate-approval-fns.ts` (approval finalization). Extracted them into a single `assertCaIssuancePolicyAllowed(policy, shouldIssueAsCA)` helper in `certificate-issuance-utils.ts`, used by all four issuance paths now (including renewal). This also normalizes the error message, which previously differed slightly between the approval path and the other two.

## Tests

Added two unit tests to `certificate-v3-service.test.ts`:
- Renewing a CA certificate under a policy that now denies CA issuance throws, and `issueCertFromCa` is never called.
- Renewing a CA certificate under a policy that allows it correctly passes `basicConstraints`/`pathLength` through to `issueCertFromCa`.

`npx vitest run -c vitest.unit.config.mts src/services/certificate` → 198/198 passing. Type-check and lint are clean for all changed files (verified against the project's pre-existing, unrelated lint/type-check debt — error counts are identical before/after this diff).

## Type
- [x] Fix

## Checklist
- [x] Tested locally (unit tests added)


Abdul-Moiz31:
Hi @adilsitos  Can you please review this Please
Thankyou

PR #7048, opened 2026-06-26
improvement(billing): render metered pricing in the catalog

PrestigePvP:
## Context

Renders usage-based (metered) pricing in the billing v2 catalog and hardens catalog parsing against new License Server price kinds.

- Honor the License Server price `kind`: a metered dimension renders as usage-based ("First N included, then $X / noun / period" plus a "Usage-based" badge) instead of a fixed per-unit price. The metered flag is tracked per cadence, so a dimension that is `per_unit` on one cadence and `metered` on the other only badges the metered cadence.
- Defensive `kind` parsing: the catalog schema accepts any string (like `model`/`tier`/`cadence`) so a new License Server kind can't break parsing for an un-upgraded consumer. Unknown kinds normalize in code via a single `KNOWN_PRICE_KINDS` allow-list and degrade to non-purchasable, never mis-sold as `per_unit`.
- `buildCheckoutItems`: only `per_unit` dimensions carry a quantity; metered and base lines are added server-side, so a metered-only plan now checks out (previously returned "not available for self-serve checkout").
- Built on main's redesigned `ProPricing` (base-fee headline plus per-dimension usage list). Base-only self-serve checkout already landed on main via `planHasPricing`, so this folds `kind`-awareness into that helper rather than re-adding the base-only path.
- Pairs with Infisical/license-server-go#31, which surfaces metered prices on `/v1/products`.

## Steps to verify the change

- `make reviewable-api` and `make reviewable-ui`: `tsc --noEmit` and ESLint clean for all changed files.
- A plan priced only on a metered dimension shows the usage-based framing and reaches checkout, instead of a fixed per-unit price or "Contact sales".

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/routes/v1/license-v2-router.ts
+++ b/backend/src/ee/routes/v1/license-v2-router.ts
@@ -24,7 +24,9 @@ const BillingV2DimSchema = z.object({
   noun: z.string(),
   monthly: z.number(),
   annual: z.number(),
-  included: z.number()
+  included: z.number(),
+  meteredMonthly: z.boolean().optional(),
+  meteredAnnual: z.boolean().optional()
 });
 
 const BillingV2CompareRowSchema = z.object({
--- a/backend/src/ee/services/license-v2/license-v2-service.ts
+++ b/backend/src/ee/services/license-v2/license-v2-service.ts
@@ -82,6 +82,12 @@ const toModel = (model: string): BillingV2Model => {
   return "usage";
 };
 
+// Price kinds we understand; an unknown kind is treated as non-purchasable, never sold as per_unit.
+const PER_UNIT_KIND = "per_unit";
+const METERED_KIND = "metered";
+const KNOWN_PRICE_KINDS: readonly string[] = [PER_UNIT_KIND, METERED_KIND];
+const isKnownPriceKind = (kind: string): boolean => KNOWN_PRICE_KINDS.includes(kind);
+
 const centsToDollars = (cents: number | null | undefined): number => {
   if (!cents) {
     return 0;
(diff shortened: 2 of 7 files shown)


scott-ray-wilson:
## Review finding: base-only self-serve plans still route to "Contact sales"

The PR description says "metered- or base-only plans now check out," but a truly base-only plan (empty `prices[]` with only `basePriceMonthlyCents`/`basePriceAnnualCents` set) still cannot. I verified this end to end against `license-server-go` `main` (PR #31 is merged there). Metered rendering is unaffected; this is specifically the base-only half of the claim.

### Where it breaks

`findPaidSelfServePlan` gates on a known-kind price row:

```ts
// backend/src/ee/services/license-v2/license-v2-service.ts:138
product.plans.find(
  (plan) =>
    plan.selfServe === true && plan.tier !== "free" && plan.prices.some((price) => isKnownPriceKind(price.kind))
);
```

For a base-only plan `prices` is empty, so this returns `undefined`. `buildCheckoutItems` then short-circuits on its first guard (`if (!paidSelfServePlan) return null`) and the new `hasBase` branch is never reached. `toCatalogProduct` likewise never sets `pro.base`/`pro.planKey`, so the UI shows no price and the "Contact sales" CTA.

### Why base-only is a real shape (verified in license-server-go)

The catalog deliberately keeps base out of `prices[]` and exposes it only as scalars:

- `loadPlanPrices` selects `WHERE ... pp.kind IN ('per_unit', 'metered')` (these become `plan.prices[]`).
- `loadBasePrices` selects `WHERE ... pp.kind = 'base'` (these become `basePrice*`), and base rows carry no dimension: `CHECK ((kind = 'base') = (dimension_id IS NULL))`.

So a plan with only a base fee serializes as `{ prices: [], basePriceMonthlyCents: 5000 }`.

It is a configurable shape, not hypothetical:
- the admin catalog API accepts product `model` `oneof=seat usage limit flat` and price `kind` in `base, per_unit, metered`.
- the frontend already has a dedicated `model === "flat" || "limit"` branch in `proPriceParts` that reads `prod.pro?.base`. That branch is dead for this shape today, because `pro.base` only populates when `findPaidSelfServePlan` succeeds.

### Caveat

Current seed data is only `seat` products with a single `per_unit` price; no metered, base, or flat rows are seeded. So the metered rendering this PR adds is itself forward-looking, and base-only is exactly as reachable as metered once catalog data is configured via the admin tooling. Not a blocker on the metered path, but worth closing since the description promises it and the frontend flat branch is built for it.

### Suggested fix

```ts
const findPaidSelfServePlan = (product: TCatalogProduct) =>
  product.plans.find(
    (plan) =>
      plan.selfServe === true &&
      plan.tier !== "free" &&
      (plan.prices.some((price) => isKnownPriceKind(price.kind)) ||
        plan.basePriceMonthlyCents != null ||
        plan.basePriceAnnualCents != null)
  );
```

This makes the `hasBase` branch reachable and lets `toCatalogProduct` populate `pro.base`/`pro.planKey`, so the frontend flat branch renders and checkout sends `quantities: {}` with the base added server-side (mirroring the metered-only path).

scott-ray-wilson (approved):
LGTM

PR #7060, opened 2026-06-27
feat(pam): add Windows AD account type with multi-host selection

lb-vn:
## Context

Add Windows AD account type with multi-host selection

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/db/migrations/20260612142201_pam-revamp.ts
```diff
@@ -182,6 +182,23 @@ export async function up(knex: Knex): Promise<void> {
...
+    ) => {
+      if (!blob) return undefined;
+      const details = JSON.parse(oldCipher.decryptor({ cipherTextBlob: blob }).toString("utf-8")) as Record<
+        string,
+        unknown
+      >;
+      const transformed = {
+        ...details,
+        hosts: details.dcAddress ? [details.dcAddress] : [],
+        rdpPort: 3389
```
bernie-g, line 197: Why is this hardcoded to 3389?

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
```diff
@@ -300,7 +300,13 @@ export const registerPamWebAccessRouter = async (server: FastifyZodProvider) =>
           .trim()
           .optional()
           .describe("Session duration (e.g. '1h', '30m'). Capped at the account's max session duration."),
-        mfaSessionId: z.string().max(64).optional().describe("MFA session ID from a completed MFA verification")
+        mfaSessionId: z.string().max(64).optional().describe("MFA session ID from a completed MFA verification"),
+        targetHost: z
+          .string()
+          .trim()
+          .max(255)
```
bernie-g, line 308: Can we add the hostPattern regex to targetHost in /access and /web-access-ticket

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -422,14 +423,22 @@ export const ACCOUNT_TYPE_CONFIGS = {
     }),
     credentials: z.object({
       username: z.string().trim().min(1).max(255),
-      password: z.string().trim().min(1).max(255)
+      password: optionalTrimmedString
     }),
     sanitizedCredentials: z.object({
       username: z.string()
```
bernie-g, line 429: ignore comment placement

Image
Let's add a tooltip to DC address since we did all the other fields

frontend/src/pages/pam/PamAccountAccessPage/useRdpSession.ts
```diff
@@ -17,6 +17,40 @@ const DEFAULT_DESKTOP_SIZE = { width: 1920, height: 1080 };
 
 const STATUS_BAR_HEIGHT = 33;
 
+// Mirrors the IronErrorKind enum, indexed by its discriminant
+const IRON_ERROR_KIND_NAMES = [
```
bernie-g: I think we can import this from the ironrdp package

bernie-g: Also this is lowkey not that useful cause I tried to signin with invalid credentials and just got a General error instead of a better one (I think it's IronRDPs fault but still)

bernie-g (changes requested):

backend/src/db/migrations/20260612142201_pam-revamp.ts
x032205: This is because in the previous model the RDP port would be stored on each individual resource that the domain account can access. Since we're not compiling every resource as part of the migration, we're hardcoding the port. Even if we did get every resource and it's RDP port, there's a chance each one has a different port which would be unfavorable.

frontend/src/pages/pam/PamAccountAccessPage/useRdpSession.ts
x032205: Yeah after testing it seems we can't even trigger cases where errors are informative. For now I completely removed it, we can expand later

bernie-g (approved):

bernie-g (approved):

PR #7063, opened 2026-06-28
fix(ui): secret actions menu

varonix0:
## Context

Refactored secret actions buttons to be a single menu rather than list of icons.
CleanShot 2026-06-28 at 23 51 38@2x

CleanShot 2026-06-28 at 23 51 42@2x

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretEditTableRow.tsx
scott-ray-wilson, line 1: When batch mode is enabled you now can't edit additional properties or modify existing changes. We need to expose all the edit valid functionality and disable the anything that shouldn't be useable when an edit is pending (ie everything previously behind the disabled dropdown)

https://github.com/user-attachments/assets/f50a62ba-a3d8-4ecf-9f7d-8f96d5d66afa

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretEditTableRow.tsx
```diff
@@ -1524,209 +1301,517 @@ export const SecretEditTableRow = ({
...
+                            variant="ghost"
+                            size="xs"
+                            className="w-6 border-0 text-muted hover:text-foreground"
+                            onClick={() => setIsTagOpen(true)}
+                          >
+                            <TagsIcon className="size-3.5" />
+                          </IconButton>
+                        </TooltipTrigger>
+                        <TooltipContent>
+                          {tags.length} tag{tags.length > 1 ? "s" : ""}
```
scott-ray-wilson: This was previously View Tags which was consistent with the other tooltips; Maybe we could do "View # Tags"? But at least make the casing consistent, ie "# Tags"

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretEditTableRow.tsx
```diff
@@ -1524,209 +1301,517 @@ export const SecretEditTableRow = ({
...
+                        </TooltipContent>
+                      </Tooltip>
+                    ) : null}
+                    {reminder && (
+                      <Tooltip>
+                        <TooltipTrigger asChild>
+                          <IconButton
+                            variant="ghost"
+                            size="xs"
+                            className="w-6 border-0 text-muted hover:text-foreground"
```
scott-ray-wilson: nit:

applies to all edit buttons; the width is now messed up and the button icons are taller than wide, they should still be square like the other actions

https://github.com/user-attachments/assets/fa60cee4-c9e7-411d-b416-1b412097b144

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretEditTableRow.tsx
```diff
@@ -1524,209 +1301,517 @@ export const SecretEditTableRow = ({
...
-                >
-                  <TooltipTrigger className="block w-full">
-                    <DropdownMenuItem
-                      onClick={() => {
-                        if (!subscription?.secretAccessInsights) {
-                          handlePopUpOpen("accessInsightsUpgrade");
-                        } else {
-                          setIsAccessInsightsOpen(true);
+                      <PencilLineIcon />
+                      Annotate
```
scott-ray-wilson: nit it feels weird the chevron points right but opens left, we support moving it to the other side but you'd need to figure out what to do with the icon

Image

scott-ray-wilson (changes requested):
I like this overall, just some nits and one issue

scott-ray-wilson (changes requested):
Image

I should be able to edit multiline in batch mode and I should NOT be able to access version history as it has a rollback mechanic and can mess with secret values

also feel like view access should be disabled but not technically an issue, just feels weird

scott-ray-wilson (approved):
looks great!

PR #7065, opened 2026-06-29
fix(dynamic-secret): skip soft-deleted projects in DAL queries

jaydeep-pipaliya:
## What

Three queries in `dynamic-secret-dal.ts` joined `project_environments` and `projects` but only filtered `Environment.deleteAfter` — never `Project.deleteAfter`:

- `findByGatewayId` (used when listing dynamic secrets tied to a gateway)
- `findByGatewayPoolId` (same, but for gateway pools)
- `countByProject` (insights / org stats)

Soft-deleting a project does not soft-delete its environments, so these queries kept returning dynamic-secret rows for projects sitting in the delete grace window. The gateway-detail and insights pages would still surface a soft-deleted project's dynamic secrets, and gateway/pool delete flows would seq-scan over them.

## Fix

Adds `whereNull(Project.deleteAfter)` to all three, matching the pattern already used in `org-product-stats-dal.ts` and the recently shipped webhook (#6970), integration (#6983), and secret-rotation-v2 (#7002) fixes.

## Checklist

- [ ] I have added tests (DAL-only change; same pattern as the existing org-product-stats-dal)

### DIFF
--- a/backend/src/ee/services/dynamic-secret/dynamic-secret-dal.ts
+++ b/backend/src/ee/services/dynamic-secret/dynamic-secret-dal.ts
@@ -223,6 +223,7 @@ export const dynamicSecretDALFactory = (db: TDbClient): TDynamicSecretDALFactory
       .join(TableName.Environment, `${TableName.SecretFolder}.envId`, `${TableName.Environment}.id`)
       .whereNull(`${TableName.Environment}.deleteAfter`)
       .join(TableName.Project, `${TableName.Environment}.projectId`, `${TableName.Project}.id`)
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Codex P2 right — without filtering the counts, the gateway list could show `connectedResourcesCount = 5` while expanding the details would show 3 (because findByGatewayId / findByGatewayPoolId hide the soft-deleted-project rows but the counts don't). Pushed cb61d99 that mirrors the same env + project soft-delete filter onto both count queries.

PR #7067, opened 2026-06-29
fix(folder-commit): skip soft-deleted projects in PIT history queries

jaydeep-pipaliya:
## What

Three queries in the folder-commit DALs join `project_environments` with `andOnNull(Environment.deleteAfter)` but never join `projects` or filter `Project.deleteAfter`:

- `folder-commit-dal.findLatestCommit`
- `folder-commit-dal.findById`
- `folder-commit-changes-dal.findByCommitId`

Soft-deleting a project does not soft-delete its environments. The point-in-time (PIT) commit history and rollback UI for a soft-deleted project would still surface its commits during the grace window.

## Fix

Adds a parallel `leftJoin(Project, ...).andOnNull(Project.deleteAfter)` to each query — same style this file already uses on environments. Mirrors the pattern from `org-product-stats-dal.ts` and the previously shipped webhook (#6970), integration (#6983), secret-rotation-v2 (#7002), dynamic-secret (#7065), and snapshot (#7066) fixes.

## Checklist

### DIFF
--- a/backend/src/services/folder-commit-changes/folder-commit-changes-dal.ts
+++ b/backend/src/services/folder-commit-changes/folder-commit-changes-dal.ts
@@ -101,6 +101,11 @@ export const folderCommitChangesDALFactory = (db: TDbClient) => {
             `${TableName.Environment}.deleteAfter`
           );
         })
+        .leftJoin(TableName.Project, function joinActiveProjectForFolderCommit() {
+          this.on(`${TableName.Environment}.projectId`, `${TableName.Project}.id`).andOnNull(
+            `${TableName.Project}.deleteAfter`
+          );
+        })
         .where((qb) => {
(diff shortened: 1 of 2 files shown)


jaydeep-pipaliya:
Greptile P1 and Codex P2 both right — `andOnNull` in a leftJoin only nulls the joined columns, it doesn't drop the row. Pushed af735d1 that:

- Converts the env and project joins from leftJoin to innerJoin
- Uses the canonical `innerJoin` + `whereNull(Project.deleteAfter)` pattern from org-product-stats-dal.ts and the other shipped fixes
- folder_commits.envId is NOT NULL with a CASCADE FK so leftJoin → innerJoin on env doesn't drop any otherwise-valid rows

Soft-deleted project's commits are now actually excluded.

PR #7069, opened 2026-06-29
fix(audit-logs): misattribution on change secret

adilsitos:
## Context

When a secret was changed, the actor being saved on audit logs was the approver instead of the commiter (the user that requested the change). This was causing confusion.

This PR also fixes the redirect to the right overview page when a click on a node at the dependency graph happens. 

## Screenshots

https://github.com/user-attachments/assets/4362dbb5-be29-4a70-af14-735bba111f1b

### audit logs saving wrong actor 
- create a secret change policy 
- perform a secret change using an account 
- With another account, accept this change
- check the audit logs and validate that on the secret change the actor that is saved is the commiter, not the approver 

### redirect on dependency graph 
- create a secret reference between a "source secret" and a "destination secret"
- check the "source secret" dependency graph by clicking on `Secret References`
- Check the `Dependency Graph` and click on a referenced secret
- Make sure that you are being redirected to the new overview page. 

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
+++ b/frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
@@ -443,24 +443,53 @@ const OverviewPageContent = () => {
     userAvailableEnvs?.[0]?.id ? [userAvailableEnvs[0].id] : []
   );
 
-  // Apply the environment filter when linked via the `environments` search param (e.g. from the
-  // secret reference tree). This runs reactively rather than on mount only, because the tree is
-  // rendered inside this page, so navigating from a node updates the param without remounting.
+  // Apply one-shot deep-link inputs to local filters, then strip them from the URL in a SINGLE
+  // navigate. These arrive either from a notification/email link (`search`, `filterBy`) or from
+  // the secret reference tree (`environments` + `search`). Handling them in one effect/navigate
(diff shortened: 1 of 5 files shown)


frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
```diff
@@ -443,17 +443,24 @@ const OverviewPageContent = () => {
...
+      .filter((env) => envSlugs.includes(env.slug))
+      .map((env) => env.id);
+    if (envIds.length > 0) {
+      setStoredEnvIds(envIds);
     }
-  }, []);
+
+    // Treat the param as a one-shot deep-link input (mirrors the `search` handling below) so it
+    // doesn't override the user's stored env filter on subsequent renders.
+    navigate({ search: (prev) => ({ ...prev, environments: [] }), replace: true });
```
adilsitos: I thought that the previous implementation would work, but it seems that we need to do a navigation, so a re render happens and the right envs are displayed. Please let me know if there is a better way of doing this.

scott-ray-wilson (approved):
LGTM!

PR #7084, opened 2026-06-30
feat(pam): add GCP IAM service account access type

bernie-g:
## Context

Adds GCP IAM as a new PAM account type, enabling brokered credential-less access to GCP service accounts through the gcloud CLI. Supports both workload identity impersonation and static key auth methods.

Companion CLI PR: https://github.com/Infisical/cli/pull/290

## Steps to verify the change

1. Create a GCP IAM PAM account with a valid service account email
2. Start a PAM session and verify gcloud commands route through the proxy with injected credentials

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -208,6 +211,65 @@ export const pamSessionServiceFactory = ({
...
+      if (credentials.authMethod === "static-key") {
+        const keyJson = JSON.parse(credentials.serviceAccountKeyJson as string) as {
+          client_email?: string;
+          private_key?: string;
+        };
+        sourceClient = new JWT({
+          email: keyJson.client_email,
+          key: keyJson.private_key,
+          scopes: ["https://www.googleapis.com/auth/cloud-platform"]
+        });
```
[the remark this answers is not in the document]
bernie-g: We'll add connection tests later

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -210,6 +213,60 @@ export const pamSessionServiceFactory = ({
       }
     }
 
+    if (account.accountType === PamAccountType.GcpIam) {
+      const serviceAccountEmail = connectionDetails.serviceAccountEmail as string;
+      const remainingSeconds = Math.max(1, Math.floor((new Date(session.expiresAt).getTime() - Date.now()) / 1000));
+      const sessionTtlSeconds = Math.min(remainingSeconds, 3600);
+
+      let sourceClient;
+      if (credentials.authMethod === "static-key") {
```
sheensantoscapadngan: let's use enum

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -387,6 +387,42 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+    connectionDetails: z.object({
+      serviceAccountEmail: z.string().trim().min(1).max(255)
+    }),
+    credentials: z.discriminatedUnion("authMethod", [
+      z.object({
+        authMethod: z.literal("impersonation")
+      }),
+      z.object({
+        authMethod: z.literal("static-key"),
+        serviceAccountKeyJson: z.string().trim().min(1).max(8192)
```
saifsmailbox98: let's valid that this is a JSON and it has the required keys: client_email and private_key

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -387,6 +387,42 @@ export const ACCOUNT_TYPE_CONFIGS = {
     }
   },
 
+  [PamAccountType.GcpIam]: {
+    name: "Google Cloud IAM",
+    icon: "Google Cloud Platform.png",
+    connectionDetails: z.object({
+      serviceAccountEmail: z.string().trim().min(1).max(255)
```
saifsmailbox98: Add email format validation, should also end with .iam.gserviceaccount.com (let me know if that's not the case)

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -387,6 +387,42 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+    sanitizedCredentials: z.object({
+      authMethod: z.string()
+    }),
+    ui: {
+      serviceAccountEmail: { label: "Service Account Email" },
+      authMethod: {
+        label: "Auth Method",
+        defaultValue: "impersonation",
+        options: [
+          { label: "Impersonation (Recommended)", value: "impersonation" },
```
saifsmailbox98: Do we need to say "Recommended"?

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -387,6 +387,42 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      }),
+      z.object({
+        authMethod: z.literal("static-key"),
+        serviceAccountKeyJson: z.string().trim().min(1).max(8192)
+      })
+    ]),
+    sanitizedCredentials: z.object({
+      authMethod: z.string()
+    }),
+    ui: {
```
saifsmailbox98, line 431: Add a tooltip on the auth method options explaining what impersonation vs static key means

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -210,6 +213,60 @@ export const pamSessionServiceFactory = ({
...
+        const keyJson = JSON.parse(credentials.serviceAccountKeyJson as string) as {
+          client_email?: string;
```
saifsmailbox98: Are these two optional? what would happen if they are not present? should they ever be not present?

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -387,6 +387,42 @@ export const ACCOUNT_TYPE_CONFIGS = {
     }
   },
 
+  [PamAccountType.GcpIam]: {
+    name: "Google Cloud IAM",
```
saifsmailbox98: I hope "Google Cloud IAM" doesn't confuse the users. In case of AWS, you're actually assuming an IAM role, but here you're impersonating a service account to access all of GCP, not just IAM - this is where the confusion comes from.

Do you think "GCP Service Account" is more accurate? Also, if we want to extend this to support Google Workspace or other Google APIs, it's a one-line scope change here: *.googleapis.com

Might be worth calling it "Google Service Account"? Let me know what you think.

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-schemas.ts
sheensantoscapadngan: agreed ^^

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Yeah I think we should try to push this as much as possible

backend/src/ee/services/pam-session/pam-session-service.ts
bernie-g: No they're both required. Fixed

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Renamed to Google Service Account

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Done

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Done

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Done

backend/src/ee/services/pam-session/pam-session-service.ts
bernie-g: Done

backend/src/ee/services/pam-account/pam-account-schemas.ts
saifsmailbox98: I think you got it renamed to "GCP Service Account" instead of "Google Service Account", also should we update the other entities/enums to not use Iam (as those are also user facing)?

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -387,6 +387,67 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      })
+    ]),
+    sanitizedCredentials: z.object({
+      authMethod: z.string()
+    }),
+    ui: {
+      serviceAccountEmail: { label: "Service Account Email" },
+      authMethod: {
+        label: "Auth Method",
+        tooltip:
```
saifsmailbox98, line 435: Does this show up on the UI for you?

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: My bad, it actually shows up in the UI now. Also I added a tooltip for the email as well

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Renamed to Google Service Account and renamed the enums and entities too

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -387,6 +387,71 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      })
+    ]),
+    sanitizedCredentials: z.object({
+      authMethod: z.string()
+    }),
+    ui: {
+      serviceAccountEmail: {
+        label: "Service Account Email",
+        tooltip:
+          "The email address of the GCP service account to broker access to (must end in .iam.gserviceaccount.com)."
```
saifsmailbox98: `must end in .iam.gserviceaccount.com` is probably not needed, there is already validation error for that.

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: Removed

saifsmailbox98 (approved):

PR #7085, opened 2026-06-30
feat: add HSM-backed internal certificate authorities

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/lib/api-docs/constants.ts
```diff
@@ -2659,6 +2659,11 @@ export const CertificateAuthorities = {
         "The maximum number of intermediate CAs that may follow this CA in the certificate / CA chain. A maxPathLength of -1 implies no path limit on the chain.",
       keyAlgorithm:
         "The type of public key algorithm and size, in bits, of the key pair for the CA; when you create an intermediate CA, you must use a key algorithm supported by the parent CA.",
+      keySource:
+        "Where the CA's signing key is generated and stored. 'infisical' keeps the key in Infisical's KMS; 'hsm' generates and stores the key in the HSM reached through the specified HSM Connector.",
+      hsmConnectorId:
+        "The ID of the HSM Connector to generate and store the CA's signing key in. Required when keySource is 'hsm'.",
+      hsmKeyLabel: "The label of the CA signing key on the HSM. Set automatically for HSM-backed CAs.",
```
sheensantoscapadngan: what do you mean by "Set automatically"? Is this not a prefix? why do we use a prefix for signers but not for this?

backend/src/services/certificate-authority/internal/internal-certificate-authority-service.ts
```diff
@@ -336,23 +342,141 @@ export const internalCertificateAuthorityServiceFactory = ({
...
+        await hsmConnectorService.assertAttachPermission(
+          { type: dto.actor, id: dto.actorId, authMethod: dto.actorAuthMethod, orgId: dto.actorOrgId },
+          dto.hsmConnectorId,
+          projectId
+        );
+      }
+
+      // Kept outside the transaction below (network round-trip); an aborted tx orphans a harmless
+      // unreferenced HSM key rather than holding a DB connection open across the call.
+      const generated = await hsmConnectorService.generateKeyPair({
```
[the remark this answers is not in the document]
sheensantoscapadngan: yeah, wouldn't it be potentially expensive to hold an orphaned HSM slot? we should add a clean-up if the transaction does end up throwing something

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CreateCaWizard/BasicsStep.tsx
```diff
@@ -0,0 +1,74 @@
...
+            </FieldLabel>
+            <FieldContent>
+              <Input {...field} placeholder="my-internal-ca" isError={Boolean(error)} />
+              <FieldDescription>A unique slug used to reference this CA.</FieldDescription>
+              <FieldError errors={[error]} />
+            </FieldContent>
+          </Field>
+        )}
+      />
+
```
sheensantoscapadngan, line 44: IGNORE PLACEMENT

I know that right now we don't auto delete the keys on the HSM side, which I think is fine. Perhaps we should make it clear in the UI that there will be a key which will get orphaned so that they at least know? Showing the actual key label would be great as well for reference...

Image

backend/src/lib/api-docs/constants.ts
```diff
@@ -2663,7 +2663,8 @@ export const CertificateAuthorities = {
         "Where the CA's signing key is generated and stored. 'infisical' keeps the key in Infisical's KMS; 'hsm' generates and stores the key in the HSM reached through the specified HSM Connector.",
       hsmConnectorId:
         "The ID of the HSM Connector to generate and store the CA's signing key in. Required when keySource is 'hsm'.",
-      hsmKeyLabel: "The label of the CA signing key on the HSM. Set automatically for HSM-backed CAs.",
+      hsmKeyLabel:
+        "The label of the CA's signing key on the HSM. Not user-supplied: it is the HSM Connector's configured key name prefix followed by a unique, server-generated per-CA suffix (ca-<uuid>).",
```
sheensantoscapadngan: hmm why don't we just use the name of the CA as the prefix instead of "ca-" ?

sheensantoscapadngan (approved):

PR #7088, opened 2026-06-30
fix(honey-token): skip soft-deleted projects in honey token DAL queries

jaydeep-pipaliya:
## What

`findByFolderIds` in `honey-token-dal.ts` joined `project_environments` and filtered `Environment.deleteAfter` but never joined `projects` or filtered `Project.deleteAfter`. Soft-deleting a project does not soft-delete its environments, so honey tokens for projects sitting in the delete grace window were still returned.

Honey tokens are tripwires — fake credentials that route the alerting pipeline when accessed. If a soft-deleted project's tokens still fire alerts, the alert names a project the user has already marked for deletion. That noise during a deletion window is exactly the case the soft-delete pattern was designed to suppress.

## Fix

Adds the `Project` join + `whereNull(Project.deleteAfter)`. Same pattern as `org-product-stats-dal.ts` and the previously shipped webhook (#6970), integration (#6983), secret-rotation-v2 (#7002), dynamic-secret (#7065), snapshot (#7066), and folder-commit (#7067) fixes.

## Checklist

### DIFF
--- a/backend/src/ee/services/honey-token/honey-token-dal.ts
+++ b/backend/src/ee/services/honey-token/honey-token-dal.ts
@@ -59,7 +59,13 @@ export const honeyTokenDALFactory = (db: TDbClient) => {
   };
 
   const countByFolderIds = async (folderIds: string[], search?: string, tx?: Knex) => {
-    const query = (tx || db.replicaNode())(TableName.HoneyToken).whereIn(`${TableName.HoneyToken}.folderId`, folderIds);
+    const query = (tx || db.replicaNode())(TableName.HoneyToken)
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Greptile P1 right — countByFolderIds is paired with findByFolderIds by the dashboard, so with only the find filtered the count would stay non-zero for a soft-deleted project while the list renders empty. Pushed 78d67d4 adding the same env + project soft-delete filter onto countByFolderIds. (countByProjectId and countByOrgId were already patched in the follow-up on this branch.)

PR #7125, opened 2026-07-02
fix(app-connection): skip soft-deleted projects when listing linked secret rotations

jaydeep-pipaliya:
## What

The secret-rotation listing helper inside \`app-connection-dal.ts\` (used when showing which secret rotations are tied to a given app connection) already joins \`TableName.Project\` but never filters \`Project.deleteAfter\`. Soft-deleting a project does not soft-delete its environments, so rotations for projects sitting in the delete grace window were still surfaced on the connection details view.

## Fix

One-line: add \`.whereNull(Project.deleteAfter)\` after the existing Project join. Same shape as the previously shipped webhook (#6970), integration (#6983), secret-rotation-v2 (#7002), dynamic-secret (#7065), snapshot (#7066), folder-commit (#7067), honey-token (#7088), secret-import (#7089), secret-approval-request (#7090, merged), access-approval-request (#7105), access-approval-policy (#7106), secret-approval-policy (#7107), and reminder (#7124) fixes.

## Checklist

### DIFF
--- a/backend/src/services/app-connection/app-connection-dal.ts
+++ b/backend/src/services/app-connection/app-connection-dal.ts
@@ -106,6 +106,7 @@ export const appConnectionDALFactory = (db: TDbClient) => {
       .join(TableName.Environment, `${TableName.SecretFolder}.envId`, `${TableName.Environment}.id`)
       .whereNull(`${TableName.Environment}.deleteAfter`)
       .join(TableName.Project, `${TableName.Environment}.projectId`, `${TableName.Project}.id`)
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Greptile P1 right — the same findAppConnectionUsageById function had three more Project joins without whereNull(Project.deleteAfter): secretSyncs, externalCas, and dataSources. Pushed 1255bef adding the same filter to all three so the connection-usage view is fully consistent.

PR #7137, opened 2026-07-02
feat(secrets): add activation modal

adilsitos:
## Context
This PR adds the possibility of prompting users on recently created organizations to add more users. The triggers happen in orgs that are at most 2 months old or have less than 5 users. This check is happening by session, to avoid triggering it several times. 

## Steps to verify the change

There are three scenarios to be validated: 
- a recently created account opens up the secrets product, and when it adds a new secret, the modal shows up. 
- After returning to the secrets product after 3 days, the modal shows up.
- After returning to the secrets product after 7 days, the modal shows up.

### to validate the scenarios, create the following:
- create a new org 
- create a new secrets project
- add a secret 
- verify that you received the activation modal
- (if possible, verify the posthog events. I have credentials if you need)
- Change the date of `firstSecretCreatedAt` to check the 3 days and 7 days modal (remember to invalidate the session, since the call for this endpoint is cached on the UI)

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/user-activation/user-activation-service.ts
```diff
@@ -0,0 +1,118 @@
...
+      const now = new Date();
+
+      // First interaction: creating the row is itself the signal that this is the user's first
+      // action, so we stamp firstSecretCreatedAt. This also covers orgs that predate the feature.
+      if (!existingActivation) {
+        const [createdActivation] = await userActivationDAL.upsert(
+          [{ userId: actorId, orgId: actorOrgId, firstSecretCreatedAt: now }],
+          ["userId", "orgId"],
+          tx
+        );
```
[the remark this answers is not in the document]
adilsitos: this should be fine in this case.

backend/src/lib/config/env.ts
```diff
@@ -238,6 +238,22 @@ const envSchema = z
...
+        .string()
+        .default("3d")
+        .transform((val) => ms(val))
+    ),
+    SECRETS_ACTIVATION_SECOND_NUDGE_DELAY: zpStr(
+      z
+        .string()
+        .default("7d")
+        .transform((val) => ms(val))
+    ),
```
scott-ray-wilson: I feel like these shouldn't have defaults? I could be wrong but feel like this shouldn't be active by default in self-hosted, only cloud?

frontend/src/pages/secret-manager/OverviewPage/components/InviteMembersModal/InviteMembersModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+                    <TooltipTrigger asChild>
+                      <button
+                        type="button"
+                        aria-label="Role information"
+                        className="text-muted transition-colors hover:text-foreground"
+                      >
+                        <InfoIcon className="size-3" />
+                      </button>
+                    </TooltipTrigger>
+                    <TooltipContent>New members join with the Member role.</TooltipContent>
```
scott-ray-wilson: this get's auto-focused when the modal opens 

Image

scott-ray-wilson: I think we should let them select the role (we can set it to the project default which is usually member anyway)

frontend/src/pages/secret-manager/OverviewPage/components/InviteMembersModal/InviteMembersModal.tsx
```diff
@@ -0,0 +1,228 @@
...
+                      !orgUsers?.some(({ user }) => user.email === input || user.username === input)
+                    }
+                    placeholder="teammate@company.com"
+                    isMulti
+                    name="members"
+                    components={{ DropdownIndicator: null, IndicatorSeparator: null }}
+                    value={field.value}
+                    onChange={field.onChange}
+                    isError={!!errors.members?.length}
+                  />
```
scott-ray-wilson: since we can't select existing users I don't think there's any payoff using a creatable select here, let's just use a text area like the org invite

Image

frontend/src/pages/secret-manager/OverviewPage/components/InviteMembersModal/InviteMembersModal.tsx
scott-ray-wilson: will also need to bring over the validation from that page as well

frontend/src/const.ts
```diff
@@ -57,6 +57,10 @@ export const plans = plansProd || plansDev;
 export const leaveConfirmDefaultMessage =
   "Your changes will be lost if you leave the page. Are you sure you want to continue?";
 
+// Name of the sample project created during onboarding (see UserInfoStep). Referenced when
+// excluding the demo project from growth nudges; use this constant instead of the bare literal.
+export const EXAMPLE_PROJECT_NAME = "Example Project";
```
scott-ray-wilson, line 62: thoughts: I'm not sure we should exclude this; I could see a team trying it out, deleting the seeded secrets and not caring/noticing the project name?

frontend/src/const.ts
scott-ray-wilson, line 1: doesn't seem like the exclude is working anyway, it auto popped as soon as I navigated to this page in an example project 

Image

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
```diff
@@ -2108,8 +2129,24 @@ const OverviewPageContent = () => {
           : "Changes saved successfully",
         type: requiresApproval ? "info" : "success"
       });
+
+      // If the commit actually created secrets (not just an approval request), check whether to
+      // surface the activation nudge.
+      const createdSecret = changes.secrets.some((s) => s.type === PendingAction.Create);
+      if (createdSecret && !requiresApproval) {
+        console.log("createdSecret", createdSecret);
```
scott-ray-wilson: nit: console log

scott-ray-wilson (changes requested):
not sure if this is intentional but we're incorrectly tracking time to first secret and telemetry if they decide to do any of this via the API and then get auto-popping modal instead of triggered by action

backend/src/lib/config/env.ts
adilsitos: makese sense. Those should only work for cloud

frontend/src/pages/secret-manager/OverviewPage/components/InviteMembersModal/InviteMembersModal.tsx
adilsitos: This is something I talked with @varonix0 . We decided to have a "leaner" modal to be easier to the user to invite new members and not think about things like roles. But open to discuss it.

frontend/src/const.ts
adilsitos: Weird, this should not be happening. One thing that I thought during the development was that we could only call it when a secret is created, which would make this more predictable. Maybe something we can discuss it.

adilsitos: I removed this validation!

scott-ray-wilson (approved):
LGTM!

PR #7139, opened 2026-07-03
chore: make recovery code account wide and revamp 2FA settings UI

Thiago-AS:
## Context

Overhauls the MFA/2FA experience with a focus on session-bound verification, step-up authentication for sensitive operations, account-level recovery codes, and brute-force lockout protection.

**Key changes:**

* **Session-bound MFA and sudo**: MFA verification and sudo (step-up) state are now bound to the specific session that established them instead of being shared across all user sessions. Sudo mode persists for 10 minutes after login, ensuring users are not locked out of step-up operations if they lose access to their MFA device during that session.
* **Step-up authentication for sensitive actions**: Enabling or disabling 2FA, removing MFA methods, and viewing or rotating recovery codes now require a fresh MFA challenge. Enabling an MFA method also requires re-verifying the method being configured.
* **Account-level recovery codes**: Recovery codes have been moved from a per-TOTP configuration model to a single account-level pool backed by the new `user_mfa_recovery_codes` table. Existing TOTP recovery codes are automatically backfilled, and the legacy `user.mfaMethods` field has been removed. A confirmation email is now sent whenever a recovery code is used.
* **Lockout protection**: Introduced a new `mfa-lockout-service` that enforces a maximum of five MFA and step-up attempts by reserving an attempt before validating a code. Lockout is enforced when creating an MFA session. Also fixed a broken login flow when an account is locked and prevented failed step-up attempts from locking users out of their primary sign-in flow.
* **Frontend refactor**: Rebuilt the MFA settings experience by splitting `MFASection` into `MfaMethodsCard`, `RecoveryOptionsCard`, per-method dialogs, and reusable hooks. Also introduced a guided enrollment flow (`VerifyStep` and `RecoveryCodesStep`), dialog-based passkey management, and a shared step-up hook.

### 1. Enable MFA (Email - Default)

1. Log in and navigate to **Organization Settings → Security**.
2. Enable 2FA.
   * Expect recovery codes to be displayed once in a dialog.
   * The **Continue** button should remain disabled until you confirm you've saved the codes.
3. Log out and sign back in.
   * Expect an email MFA challenge.

---

### 2. Set Up Authenticator (TOTP)

1. In the **MFA Methods** card, set up an authenticator by scanning the QR code (or entering the secret manually) and verifying it with a 6-digit code.
2. Select **Authenticator (TOTP)** as the preferred MFA method.
4. Log out and sign back in.
   * Expect a TOTP challenge.

---

### 3. Set Up and Manage Passkeys (WebAuthn)

1. Register a passkey (using either a virtual authenticator via browser DevTools or a physical security key/Touch ID).
2. Select **Passkey** as the preferred MFA method.
3. Log out and sign back in.
   * Expect a passkey challenge.
---

### 4. Recovery Codes (Step-Up MFA)

1. Open **Recovery Options** and choose **View** or **Regenerate** recovery codes.
2. Expect a step-up MFA challenge.
   * The challenge should use the currently selected MFA method (email, TOTP, or passkey).
3. After completing the challenge:
   * Recovery codes are displayed.
   * Regenerating creates a new set and invalidates the previous one.
   * Within the 5-minute MFA session window, viewing or regenerating again should not require another challenge.

---

### 5. Recovery Code Login (All MFA Methods)

1. Log out.
2. At the MFA challenge, choose **Use a recovery code** and enter a valid recovery code.
3. Verify:
   * Login succeeds.
   * The recovery code is consumed.
4. Attempt to log in again using the same recovery code.
   * Expect an **"Invalid recovery code"** error.
5. Repeat this flow with each MFA method selected:
   * Email
   * Authenticator (TOTP)
   * Passkey

Recovery codes should work regardless of the selected MFA method (this is the primary regression check compared to the previous TOTP-only implementation).

---

### 6. Organization-Enforced MFA

1. As an organization administrator, enable **Enforce MFA**.
2. Invite a new user (or create a new account) and sign in.
   * Expect the user to be routed into the guided **MfaEnrollment** flow.
   * Complete MFA setup.
   * Save the recovery codes.
   * Continue into the organization.
3. As an MFA-enabled member, navigate to **Security** and attempt to disable MFA.
   * Expect the error:
     > "MFA is required by your organization and cannot be disabled."

---

### 7. Disable MFA (Non-Enforced Organization)

1. In an organization without MFA enforcement, disable 2FA.
2. Verify the database:
   * The `user_mfa_recovery_codes` row is deleted.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/db/migrations/20260701120200_add-mfa-recovery-codes.ts
```diff
@@ -0,0 +1,165 @@
...
+    await createOnUpdateTrigger(knex, TableName.UserMfaRecoveryCode);
+
+    // Backfill from existing TOTP recovery codes. Both stores use
+    // kmsService.encryptWithRootKey() (root-key AES-GCM, table-independent),
+    // so the ciphertext is portable and can be copied without decrypting.
+    await knex.raw(
+      `INSERT INTO ?? ("userId", "encryptedRecoveryCodes")
+       SELECT "userId", "encryptedRecoveryCodes"
+       FROM ??
+       WHERE "isVerified" = true
```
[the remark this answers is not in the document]
Thiago-AS: That's intended.

backend/src/services/totp/totp-service.ts
```diff
@@ -238,51 +189,22 @@ export const totpServiceFactory = ({ totpConfigDAL, kmsService, userDAL, keyStor
...
-      });
-    });
+    // If the user's preferred method is the TOTP config we just removed, reset it
+    // so getRequiredMfaMethod doesn't keep challenging for a method that no longer
+    // has a verified config, which would force the user back into enrollment. Fall
+    // back to email, which is always available.
+    const user = await userDAL.findById(userId);
+    if (user?.selectedMfaMethod === MfaMethod.TOTP) {
+      await userDAL.updateById(userId, { selectedMfaMethod: MfaMethod.EMAIL });
+    }
```
scott-ray-wilson: given the comment, this should probably be in a transaction with the delete?

backend/src/services/webauthn/webauthn-service.ts
```diff
@@ -339,6 +340,18 @@ export const webAuthnServiceFactory = ({
...
+    // method, reset the preference so getRequiredMfaMethod doesn't keep
+    // challenging for webauthn with no credential left, which would lock the
+    // user out. Fall back to email, which is always available.
+    const remainingCredentials = await webAuthnCredentialDAL.find({ userId });
+    if (remainingCredentials.length === 0) {
+      const user = await userDAL.findById(userId);
+      if (user?.selectedMfaMethod === MfaMethod.WEBAUTHN) {
+        await userDAL.updateById(userId, { selectedMfaMethod: MfaMethod.EMAIL });
+      }
+    }
```
scott-ray-wilson: same here, ideally in a transaction with the above delete

backend/src/services/user/user-service.ts
```diff
@@ -95,27 +112,101 @@ export const userServiceFactory = ({
...
+      }
+    } else if (method === MfaMethod.WEBAUTHN) {
+      const credentials = await webAuthnCredentialDAL.find({ userId });
+      if (credentials.length === 0) {
+        throw new BadRequestError({
+          message: "Cannot select a passkey without a registered passkey"
+        });
+      }
+    }
+  };
```
scott-ray-wilson, line 140: not sure if here is the right place but we should also block setting up email MFA if SMTP has not been configured

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/MFASection.tsx
```diff
@@ -1,841 +1,296 @@
...
-
-      createNotification({
-        text: "Successfully renamed passkey",
-        type: "success"
-      });
-
-      handlePopUpClose("renameWebAuthnCredential");
-      setSelectedCredentialId("");
-      setCredentialName("");
+      await setMfaMethod({ selectedMfaMethod: method });
```
scott-ray-wilson: I think we ideally guard on email if smtp is not setup here so we can show a modal (though toast might be sufficient from backend blocking)

frontend/src/components/auth/Mfa.tsx
```diff
@@ -140,10 +162,13 @@ export const Mfa = ({ successCallback, closeMfa, hideLogo, email, method }: Prop
       SecurityClient.setMfaToken("");
       SecurityClient.setToken(result.token);
 
-      await successCallback();
-      if (closeMfa) {
-        closeMfa();
+      if (!showRecoveryCodeInput && !wasMfaEnabledRef.current) {
+        const { recoveryCodes } = await activateMfa({ selectedMfaMethod: method });
```
scott-ray-wilson: nit: unlikely but activate and verify are in the same try block so if verify succeeds but activate fails users get stuck thinking the verification failed code and keep submitting but can't progress; we should probably move this to it's own try/catch and just forward them to login as it will self heal

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/MFASection.tsx
```diff
@@ -1,841 +1,296 @@
...
-                    setShouldShowRecoveryCodes.off();
-                  }}
-                  disabled={isLoading}
-                  className="border-mineshaft-500 text-mineshaft-300 hover:border-mineshaft-400"
-                >
-                  Cancel
-                </Button>
-              </div>
-            )}
+        <SheetContent
```
scott-ray-wilson: not a blocker but I kind of feel like this is small enough to warrant a dialog over a sheet?

Image

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/MFASection.tsx
```diff
@@ -1,841 +1,296 @@
...
-                  setCredentialName("");
-                  handlePopUpOpen("registerPasskey");
-                }}
-                isLoading={isGeneratingOptions}
+          <SheetFooter className="flex-col items-stretch gap-3 border-t">
+            <div className="flex items-center gap-2">
+              <Checkbox
+                id="acknowledge-recovery-codes"
+                isChecked={hasAcknowledgedCodes}
+                onCheckedChange={(checked) => setHasAcknowledgedCodes(checked === true)}
```
scott-ray-wilson: total nit but cuter if org variant to match 💅

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/MFASection.tsx
scott-ray-wilson, line 1: I know you shared screenshots before but now as I'm actually using this when I first set up I was confused where the other methods were because I was thinking my context was a single card; I think it might be better if all this was in a single card but we sectioned it with separators?

Image

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/PasskeyManagerDialog.tsx
```diff
@@ -0,0 +1,113 @@
...
+                      <p className="text-xs text-muted">
+                        Added {new Date(credential.createdAt).toLocaleDateString()}
+                        {credential.lastUsedAt &&
+                          ` · Last used ${new Date(credential.lastUsedAt).toLocaleDateString()}`}
+                      </p>
+                    </div>
+                  </div>
+                  <Button variant="danger" size="sm" onClick={() => handleDelete(credential.id)}>
+                    <TrashIcon /> Remove
+                  </Button>
```
scott-ray-wilson: think this should be behind a confirmation delete modal (typed)

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/PasskeyManagerDialog.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <div className="flex-1">
+              <Input
+                value={name}
+                onChange={(e) => setName(e.target.value)}
+                placeholder="New passkey name (optional)"
+              />
+            </div>
+            <Button variant="org" isPending={isRegistering} onClick={handleAdd}>
+              <PlusIcon /> Add
+            </Button>
```
scott-ray-wilson: adding a passkey from fullscreen on chrome mac always times out (not sure if you can do anything about this)

https://github.com/user-attachments/assets/cf9a73ee-d74c-4ccc-b6f0-89eac28d3fea

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/PasskeyManagerDialog.tsx
scott-ray-wilson, line 1: total polish but I think it might be nice if when I add a passkey (or other method) I get a confirmation modal asking if I would like to set it as my default/active 2fa; I think users might add one but not realize they need to select it from the dropdown to make it apply

scott-ray-wilson, line 1: ignore file placement:

we make users verify to disable 2FA but I wonder if we should do the same to switch which method is active? Thinking if I had an active session with TOTP I can bypass that deactivation by switching method and using a passkey or email to deactivate instead

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/MethodSetupDialog.tsx
```diff
@@ -0,0 +1,37 @@
...
+  method: MfaMethod | null;
+  onOpenChange: (isOpen: boolean) => void;
+  onCompleted: () => void;
+};
+
+// Lightweight single-method setup used from the enabled view (MFA is already on
+// and recovery codes already exist, so no enable/recovery steps are needed).
+export const MethodSetupDialog = ({ method, onOpenChange, onCompleted }: Props) => {
+  return (
+    <Sheet open={Boolean(method)} onOpenChange={onOpenChange}>
```
scott-ray-wilson: same here I see this and think probably just in a dialog but not a blocker

Image

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/MethodSetupDialog.tsx
scott-ray-wilson, line 1: ignore file placement:

nit: if I try to disable 2fa with totp and input a bad code I can't retry and also don't get any feedback/window close to retry

Image

scott-ray-wilson: ah yeah maybe we should probably auto close it too cause if I just switch tabs I have to hard refresh to try again but would expect 3 attempts like usual

Image

scott-ray-wilson, line 1: ignore file placement:

not sure if easy to do but I kind of expected to be able to cancel enabling 2fa; once this modal opens I have to go through the disable process; think a confirmation of enabling might be good

Image

scott-ray-wilson, line 1: if out of scope for this PR can we add a ticket to update docs/images? 

Image

scott-ray-wilson, line 1: ignore file placement:

getting locked out of an account by failing mfa breaks the UI if you try logging back in again 

https://github.com/user-attachments/assets/e73d8d14-7d4a-4440-b984-f4d80f85f510

scott-ray-wilson: don't think this is related to your changes but weird that when I'm locked out the date is the only thing set, seems like we don't set/use the boolean? 🤔 

Image

scott-ray-wilson, line 1: when a user regenerates recovery codes I think we should do the confirmation to close like we do for creating? or neither 🤷‍♂️ 

Image

frontend/src/components/mfa/setup/VerifyStep.tsx
```diff
@@ -0,0 +1,125 @@
...
+    try {
+      await enrollMfa({ method: MfaMethod.TOTP, totp: totp.trim() });
+      createNotification({ text: "Authenticator app configured", type: "success" });
+      await onVerified();
+    } catch (error: any) {
+      createNotification({
+        text: error?.response?.data?.message || "Invalid verification code",
+        type: "error"
+      });
+    }
```
scott-ray-wilson: nit: this toast is duplicate of the one thrown from the mutation itself and blocks the more informative one

Image

frontend/src/components/mfa/setup/VerifyStep.tsx
scott-ray-wilson, line 1: ala clood:

This PR introduces an account-level recovery-code pool that acts as a universal second-factor bypass at login. In `backend/src/services/auth/auth-login-service.ts:710-714`, `verifyMfaToken` short-circuits to `verifyAndConsumeRecoveryCode` whenever `isRecoveryCode` is set, **before** the `mfaMethod !== requiredMfaMethod` check runs, so a recovery code satisfies any `requiredMfaMethod`, including WebAuthn. `verifyAndConsumeRecoveryCode` (`mfa-recovery-code-service.ts:56`) is keyed on `userId` only, so the pool is account-wide and org-agnostic.

The step-up challenge that gates viewing and rotating those codes is chosen by `getStepUpMfaMethod` (`user-service.ts:490-494`), which resolves the required method from only the single current org via `getRequiredMfaMethod(org, user)`. This contrasts with `deactivateMfa` (`user-service.ts:177-189`), which correctly enumerates all of the user's org memberships and enforces the strongest policy (`organizations.some((org) => org.enforceMfa)`). Because one account-wide pool bypasses every org's MFA, gating its exposure on only the current (possibly weaker) org's method lets an attacker extract the codes through a low-enforcement org and defeat a stricter org's enforced factor.

Confirmed net-new via git history: on `main`, `verifyMfaToken` checked the method match first and recovery codes were TOTP-only (nested inside the `MfaMethod.TOTP` branch). The account-level pool, the view/rotate endpoints, and `getStepUpMfaMethod` do not exist on `main`.

### Exploit Scenario
A victim belongs to Org A (`enforceMfa=true`, `selectedMfaMethod=WEBAUTHN`) and Org B (`enforceMfa=false`); the victim's personal `selectedMfaMethod` is `EMAIL`. An attacker holds the victim's password and email inbox access, but not the passkey.

1. Attacker logs in and calls `/select-organization` for Org B. Per `auth-fns.ts:18-24`, since Org B does not enforce MFA, `requiredMfaMethod` falls back to the user preference `EMAIL`. The attacker reads the emailed code and passes login MFA.
2. With the Org-B-scoped JWT, the attacker calls `GET /api/v1/user/me/mfa/recovery-codes`. `ensureStepUpMfa` uses `req.permission.orgId` (Org B), so `getStepUpMfaMethod` picks `EMAIL` for the step-up. The attacker reads a second emailed code, verifies the step-up session, and receives the account-level recovery codes (`POST` even rotates and returns fresh codes).
3. The attacker logs in again, selects Org A (issuing a WebAuthn challenge token without verifying the passkey), and submits a recovery code to `POST /api/v2/mfa/verify/recovery-code`. The `isRecoveryCode` short-circuit bypasses Org A's WebAuthn requirement, yielding a fully authorized Org A session. The phishing-resistant factor Org A enforced is never exercised.

### Recommendation
Make `getStepUpMfaMethod` (and therefore `ensureStepUpMfa`) resolve the strongest required MFA method across all of the user's org memberships, mirroring the all-orgs enumeration already used in `deactivateMfa`, so the gate protecting the universal recovery-code pool is never weaker than any org's enforced factor. Additionally, consider disallowing recovery-code login from satisfying a WebAuthn-enforced org, so a compromised email factor cannot substitute for a phishing-resistant one.

scott-ray-wilson (changes requested):

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/PasskeyManagerDialog.tsx
Thiago-AS: It worked here, skill issues 🫣

frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/MethodSetupDialog.tsx
Thiago-AS: Just checked, and we only set isLocked after we've definitively locked the user's account, once the exponential cooldown period has been exhausted.

backend/src/db/migrations/20260701120200_add-mfa-recovery-codes.ts
```diff
@@ -0,0 +1,165 @@
...
+  // deploy the previous backend version is still serving and reads this column
+  // unconditionally (getUserTotpConfig / registerUserTotp / verifyUserTotpConfig
+  // and the recovery-code login path); dropping it would 500 those pods for
+  // every TOTP operation until they roll. Making it nullable lets the new code
+  // insert TOTP configs without it while the old code keeps reading the existing
+  // ciphertext. A follow-up migration can drop the column once every pod runs
+  // the new code.
+  if (await knex.schema.hasColumn(TableName.TotpConfig, "encryptedRecoveryCodes")) {
+    await knex.schema.alterTable(TableName.TotpConfig, (t) => {
+      t.binary("encryptedRecoveryCodes").nullable().alter();
```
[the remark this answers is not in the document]
Thiago-AS: That's ok, not much value on using a staged migration only to mitigate a few minutes of duplicated usage of recovery codes.

scott-ray-wilson (approved):
nice

PR #7142, opened 2026-07-03
feat: pam rotation revamp

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/@types/fastify.d.ts
```diff
@@ -36,6 +36,7 @@ import { TLicenseServiceFactory } from "@app/ee/services/license/license-service
 import { TLicenseV2ServiceFactory } from "@app/ee/services/license-v2/license-v2-service";
```
x032205, line 37: ignore placement

This dropdown looks a little weird with the spacing on the left. Can we move the checkmark spot to the right of the items at the end?

Image

backend/src/@types/fastify.d.ts
```diff
@@ -36,6 +36,7 @@ import { TLicenseServiceFactory } from "@app/ee/services/license/license-service
...
 import { TOidcConfigServiceFactory } from "@app/ee/services/oidc/oidc-config-service";
```
x032205, line 38: ignore placement

Image

This UI says that you can manually trigger rotation when the toggle is off, but you can't actually do that.

backend/src/@types/fastify.d.ts
```diff
@@ -36,6 +36,7 @@ import { TLicenseServiceFactory } from "@app/ee/services/license/license-service
...
 import { TPamAccountServiceFactory } from "@app/ee/services/pam-account/pam-account-service";
+import { TPamAccountRotationServiceFactory } from "@app/ee/services/pam-account-rotation/pam-account-rotation-service";
 import { TPamAccountTemplateServiceFactory } from "@app/ee/services/pam-account-template/pam-account-template-service";
```
x032205, line 42: ignore placement

If we end up keeping this tab and UI, I think we should move the "Rotate Now" button to be at the bottom, full width, and have the button type be "pam" variant.

Image

backend/src/@types/fastify.d.ts
```diff
@@ -36,6 +36,7 @@ import { TLicenseServiceFactory } from "@app/ee/services/license/license-service
...
 import { TPamFolderServiceFactory } from "@app/ee/services/pam-folder/pam-folder-service";
 import { TPamMembershipServiceFactory } from "@app/ee/services/pam-membership/pam-membership-service";
```
x032205, line 44: ignore placement

When you put in values that are outside of the form schema, we don't surface the errors properly. In the example below I'm inputting a password length of 1:

Image

backend/src/@types/fastify.d.ts
```diff
@@ -428,6 +429,7 @@ declare module "fastify" {
       pamAccountTemplate: TPamAccountTemplateServiceFactory;
       pamFolder: TPamFolderServiceFactory;
```
x032205, line 435: ignore placement

The query for getting candidate rotation accounts in the rotation account dropdown uses "ReadAccounts" perm, but actually setting the account uses "ViewCredentials" perm. We should make both use the same permission

backend/src/@types/fastify.d.ts
```diff
@@ -428,6 +429,7 @@ declare module "fastify" {
...
       pamAccount: TPamAccountServiceFactory;
+      pamAccountRotation: TPamAccountRotationServiceFactory;
       pamMembership: TPamMembershipServiceFactory;
       pamSession: TPamSessionServiceFactory;
```
x032205, line 439: ignore placement

I think we should make the rotator accounts connect to THEIR OWN database for rotation rather than the target account's database. This is because they could possibly not be part of that database yet. Roles are server-global so connecting to the rotator account database still allows the rotation query to be ran

backend/src/ee/services/pam-account-rotation/pam-rotation-fns.test.ts
```diff
@@ -0,0 +1,101 @@
+import { describe, expect, test } from "vitest";
+
+import { PamAccountType } from "../pam/pam-enums";
```
x032205, line 3: Coverage in this test file is only the pure helpers. The recovery-probe branch in `rotateWithinLock` (pending-works → promote / current-works → clear / neither → defer) has no test and is the easiest thing to regress. We should probably make `PAM_ROTATION_FACTORY_MAP` injectable and add one test driving that decision tree with a fake handler

backend/src/@types/fastify.d.ts
```diff
@@ -36,6 +36,7 @@ import { TLicenseServiceFactory } from "@app/ee/services/license/license-service
 import { TLicenseV2ServiceFactory } from "@app/ee/services/license-v2/license-v2-service";
 import { TOidcConfigServiceFactory } from "@app/ee/services/oidc/oidc-config-service";
 import { TPamAccountServiceFactory } from "@app/ee/services/pam-account/pam-account-service";
+import { TPamAccountRotationServiceFactory } from "@app/ee/services/pam-account-rotation/pam-account-rotation-service";
```
x032205, line 41: ignore placement

Here's a comment pointed out by Claude:

MSSQL Windows-auth accounts (`ntlm`/`kerberos`) have no SQL password to rotate. The set-time check that rejects them only runs when `credentialConfigured` is true. But those accounts can have `credentialConfigured: false` (password is optional for Windows auth), so they slip past the check, get marked "ready," and then fail every rotation forever, retrying hourly.

Fix: validate the target's auth method always, not just when `credentialConfigured` is true.

lb-vn (changes requested):

backend/src/@types/fastify.d.ts
x032205, line 34: ignore placement

When enabling rotation and selecting a rotation account, you can then edit another form in another tab like editing the name and it'll clear the rotation account.

x032205, line 36: ignore placement

I think that we should turn the whole rotation tab into a form and have discard/save logic like the other tabs

backend/src/@types/fastify.d.ts
```diff
@@ -37,6 +37,7 @@ import { TLicenseServiceFactory } from "@app/ee/services/license/license-service
 import { TLicenseV2ServiceFactory } from "@app/ee/services/license-v2/license-v2-service";
```
x032205, line 37: ignore placement

Confirmed with Sheen that we should allow manual rotation when automatic rotation is turned off. I think a good way to do this is to rename the top toggle to something like "Allow Rotation" and add a new "Manual" option in the schedule dropdown.

Image

backend/src/@types/fastify.d.ts
x032205, line 45: ignore placement

Image

when rotation fails, it says "check rotation status". But where? I don't see anywhere on the UI to see the error besides the brief error toast

lb-vn (changes requested):

backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
```diff
@@ -0,0 +1,632 @@
...
+  }> => {
+    const account = await loadAccount(accountId, projectId);
+    await checkAccount(
+      accountId,
+      account.folderId,
+      projectId,
+      ResourcePermissionPamResourceActions.ManageRotation,
+      ctx
+    );
+    // Manual rotation follows the same readiness rule as the schedule, so a not-ready account can't be rotated out-of-band.
```
x032205: This is a stale comment

backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
```diff
@@ -0,0 +1,632 @@
...
+
+  const rotateNow = async ({
+    accountId,
+    projectId,
+    ...ctx
+  }: TRotatePamAccountDTO & TActorContext): Promise<{
+    accountType: string;
+    rotationAccountId: string | null;
+    rotationStatus: string;
+    message?: string;
```
x032205, line 524: ignore placement

Can we make the error popup thing appear right away when a rotation error happens? Right now you need to reload to see it.

backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
```diff
@@ -0,0 +1,632 @@
...
+  }> => {
+    const account = await loadAccount(accountId, projectId);
+    await checkAccount(
+      accountId,
```
x032205, line 528: ignore placement

Image

How would it look if we move the "Last rotation failed" popup to be within the top card right under the Rotate Now button? Putting it at the top level breaks our design system of having cards

lb-vn (changes requested):

lb-vn (dismissed):

sheensantoscapadngan (approved):

PR #7146, opened 2026-07-03
feat: move helm chart docs to self hosting section

mathnogueira:
## Context

We are centralizing all helm docs in our docs website. All helms are placed under "Self hosting" > "Helm charts"

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


docs/self-hosting/deployment-options/kubernetes-helm.mdx
```diff
@@ -176,788 +180,240 @@ For production deployments with many users or secrets, increase these values acc
...
+    ingressClassResource:
+      # -- Use a unique IngressClass name so the bundled controller only serves Infisical's Ingress,
+      # -- not other Ingresses in the cluster that use the common "nginx" class.
+      name: infisical-nginx
+      controllerValue: k8s.io/infisical-nginx
+      # -- Do not set this as the default IngressClass for the cluster
+      default: false
+    ingressClass: infisical-nginx
 ```
-
```
jakehulberg, line 320: Do we want to keep this warning? the warning about not deleting the Postgres PVC because it holds all your encrypted secrets. It's dropped in this move and I couldn't find it anywhere else in the docs.

Maybe we move it onto the new helm-charts overview or into production-hardening?

mathnogueira: Created a production hardening section that includes the additional configurations and the warning

jakehulberg:
@greptile re-review pls

docs/self-hosting/deployment-options/kubernetes-helm.mdx
```diff
@@ -454,7 +554,7 @@ redis:
     spec:
       podSelector:
         matchLabels:
-          component: infisical
+          app.kubernetes.io/name: infisical
```
jakehulberg: I believe this should be component: infisical. Our chart puts that label on pods (see infisical.matchLabels in helm-charts/infisical-standalone-postgres/templates/_helpers.tpl), and it never applies app.kubernetes.io/name, so this selector would silently match zero pods. Same for the other three selectors below (ServiceMonitor, topology spread, PDB).

docs/self-hosting/deployment-options/kubernetes-helm.mdx
```diff
@@ -566,7 +666,7 @@ redis:
     spec:
       selector:
         matchLabels:
-          component: infisical
+          app.kubernetes.io/name: infisical
```
jakehulberg: same as above

docs/self-hosting/deployment-options/kubernetes-helm.mdx
```diff
@@ -593,7 +693,7 @@ redis:
           whenUnsatisfiable: ScheduleAnyway
           labelSelector:
             matchLabels:
-              component: infisical
+              app.kubernetes.io/name: infisical
```
jakehulberg: same as above

jakehulberg:
Pushed a one-line fix (ca6bfcb) for the verify step in `docs/snippets/kubernetes-operator-install.mdx`: it checked `kubectl get pods -n infisical`, but none of the install commands on the page use that namespace (cluster-wide installs land in the current context namespace, and the scoped examples use `single-namespace`/`ns1`/`ns2`/`operator-namespace`), so every path ended with "No resources found". It now uses a `<namespace>` placeholder with a note for cluster-wide installs. @mathnogueira shout if you'd rather handle it differently. Thanks for the quick turnaround on the label selectors!

jakehulberg (approved):
LGTM. Verified the label selectors now match what the chart stamps on pods (`component: infisical`), and pushed a one-liner fixing the operator verify-step namespace (ca6bfcb, see comment above for rationale). Remaining nits (PKI issuer OCI vs classic repo install path, the `#kubernetes-via-helm-chart` anchor in automated-bootstrapping.mdx) are non-blocking.

PR #7157, opened 2026-07-04
fix(offline-usage-report): skip soft-deleted projects in per-project metric queries

jaydeep-pipaliya:
## What

The offline-usage-report DAL builds two per-project metric queries (\`getProjectMetrics\` / \`getSecretMetrics\`) that leftJoin projects for aggregation but never filter \`Project.deleteAfter\`. Soft-deleting a project does not soft-delete its environments, so the offline usage report bundled by self-hosted operators keeps counting secrets against projects sitting in the delete grace window.

## Fix

Adds \`andOnNull("p.deleteAfter")\` to each Project leftJoin so a soft-deleted project's rows have \`p\` columns nulled out; the subsequent \`whereNotNull("p.id")\` already in the query then drops them from the aggregation. Mirrors the existing env-side \`andOnNull\` pattern used in the same file.

Same behavior as the previously shipped webhook (#6970), integration (#6983), secret-rotation-v2 (#7002), dynamic-secret (#7065), snapshot (#7066), folder-commit (#7067), honey-token (#7088), secret-import (#7089), secret-approval-request (#7090, merged), access-approval-request (#7105), access-approval-policy (#7106), secret-approval-policy (#7107), reminder (#7124), app-connection (#7125), secret-approval-request-secret (#7126), secret-folder-version (#7155), and secret-version (#7156) fixes.

## Checklist

### DIFF
--- a/backend/src/services/offline-usage-report/offline-usage-report-dal.ts
+++ b/backend/src/services/offline-usage-report/offline-usage-report-dal.ts
@@ -111,7 +111,9 @@ export const offlineUsageReportDALFactory = (db: TDbClient) => {
       .leftJoin(`${TableName.Environment} as e`, function joinActiveEnvForFolder() {
         this.on("sf.envId", "e.id").andOnNull("e.deleteAfter");
       })
-      .leftJoin(`${TableName.Project} as p`, "e.projectId", "p.id")
+      .leftJoin(`${TableName.Project} as p`, function joinActiveProjectForFolder() {
+        this.on("e.projectId", "p.id").andOnNull("p.deleteAfter");
+      })
       .where("p.type", ProjectType.SecretManager)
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Codex P2 both right — projectMetrics was still counting raw `TableName.Project` (feeds "Total Projects" / "Projects <type>" in the CSV) and totalSecrets was still counting raw `TableName.SecretV2` (feeds "Total Secrets" in the CSV). Pushed 4501e23:

- Added `whereNull(Project.deleteAfter)` to the project-type count.
- Reshaped the totalSecrets query to join SecretFolder → Environment → Project with the same env + project soft-delete filter the per-project breakdown uses, so the aggregate and the per-project totals are internally consistent.

PR #7158, opened 2026-07-04
feat: add native Windows ADCS

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/server/routes/v1/certificate-authority-routers/adcs-certificate-authority-router.ts
```diff
@@ -0,0 +1,80 @@
...
+    schema: {
+      hide: false,
+      operationId: "getAdcsTemplates",
+      description: "Get available certificate templates from ADCS CA",
+      params: z.object({
+        caId: z.string().describe("ADCS CA ID")
+      }),
+      querystring: z.object({
+        projectId: z.string().optional().describe(openApiHidden())
+      }),
```
sheensantoscapadngan: project detected

backend/src/server/routes/v1/certificate-authority-routers/internal-certificate-authority-router.ts
```diff
@@ -761,6 +762,60 @@ export const registerInternalCertificateAuthorityRouter = async (server: Fastify
     }
   });
 
+  server.route({
+    method: "POST",
+    url: "/:caId/install-certificate-adcs-native",
+    config: {
```
sheensantoscapadngan, line 768: `install-certificate-native-adcs`

backend/src/server/routes/v1/certificate-authority-routers/internal-certificate-authority-router.ts
```diff
@@ -718,7 +719,7 @@ export const registerInternalCertificateAuthorityRouter = async (server: Fastify
       hide: false,
       operationId: "installCaCertificateAdcs",
       tags: [ApiDocsTags.PkiCertificateAuthorities],
-      description: "Install a CA certificate via Azure AD CS",
+      description: "Install a CA certificate via Azure AD CS (Web Enrollment)",
```
sheensantoscapadngan, line 722: IGNORE PLACEMENT

seems like it would be best for us to hide the web enrollment approach altogether? from docs and from the UI? we just need to make sure people who are already using it will not be affected but I don't think we should let people continue creating new external CAs especially given that it's being referred to internally as "AZURE ADCS" which doesn't make sense?

from what i'm seeing, this web enrollment approach is not being done by other platforms and can be considered as more of a hack

backend/src/services/certificate-authority/adcs/adcs-certificate-authority-fns.ts
```diff
@@ -0,0 +1,666 @@
...
+          alg = keyAlgorithmToAlgCfg(CertKeyAlgorithm.ECDSA_P384);
+          break;
+        case "ECDSA-SHA512":
+        case "SHA512WITHECDSA":
+          alg = keyAlgorithmToAlgCfg(CertKeyAlgorithm.ECDSA_P521);
+          break;
+        default:
+          alg = keyAlgorithmToAlgCfg(keyAlgorithm);
+          break;
+      }
```
sheensantoscapadngan, line 478: unrelated but i noticed that we do this multiple times in the code-base.. could we instead extract this a shared helper function?

backend/src/services/certificate-authority/certificate-authority-maps.ts
```diff
@@ -23,6 +24,11 @@ export const CERTIFICATE_AUTHORITIES_CAPABILITIES_MAP: Record<CaType, CaCapabili
     CaCapability.RENEW_CERTIFICATES
     // Note: REVOKE_CERTIFICATES intentionally omitted - not supported by ADCS connector
   ],
+  [CaType.ADCS]: [
+    CaCapability.ISSUE_CERTIFICATES,
+    CaCapability.RENEW_CERTIFICATES
+    // Note: REVOKE_CERTIFICATES intentionally omitted - MS-WCCE enrollment does not support revocation
```
sheensantoscapadngan: yikes.. this is actually a big expected use-case...  perhaps it's exposed via a different protocol other than MS-WCCE? let's look into this

backend/src/server/routes/v1/certificate-authority-routers/adcs-certificate-authority-router.ts
```diff
@@ -0,0 +1,80 @@
...
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const projectId = req.internalCertManagerProjectId;
+      const templates = await server.services.certificateAuthority.getADCSTemplates({
+        caId: req.params.caId,
+        projectId,
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId
```
sheensantoscapadngan, line 58: IGNORE PLACEMENT

we have a "Limitations" heading but only a call out component is under it...
Image

backend/src/server/routes/v1/certificate-authority-routers/adcs-certificate-authority-router.ts
```diff
@@ -0,0 +1,80 @@
...
+      });
+
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        projectId,
+        event: {
+          type: EventType.GET_ADCS_TEMPLATES,
+          metadata: {
+            caId: req.params.caId,
```
sheensantoscapadngan, line 67: IGNORE PLACEMENT

would be best for us to be more specific here.. like what ports? 
Image

backend/src/server/routes/v1/certificate-authority-routers/adcs-certificate-authority-router.ts
```diff
@@ -0,0 +1,80 @@
...
+            amount: templates.length
+          }
+        }
+      });
+
+      return { templates };
```
sheensantoscapadngan, line 73: IGNORE PLACEMENT

once we remove Azure ADCS, now we can just call this "Microsoft ADCS"
Image

backend/src/server/routes/v1/certificate-authority-routers/adcs-certificate-authority-router.ts
```diff
@@ -0,0 +1,80 @@
...
+      response: {
+        200: z.object({
+          templates: z.array(
+            z.object({
+              id: z.string().describe("Template identifier"),
+              name: z.string().describe("Template display name")
+            })
+          )
+        })
+      }
```
sheensantoscapadngan, line 47: IGNORE PLACEMENT

is it necessary to repeat this in both pages?
Image
Image

backend/src/server/routes/v1/certificate-authority-routers/adcs-certificate-authority-router.ts
```diff
@@ -0,0 +1,80 @@
...
+        ...req.auditLogInfo,
+        projectId,
+        event: {
+          type: EventType.GET_ADCS_TEMPLATES,
+          metadata: {
+            caId: req.params.caId,
+            amount: templates.length
+          }
+        }
+      });
```
sheensantoscapadngan, line 71: IGNORE PLACMEMENT

it doesn't seem like this would be relevant for normal users? we should make the docs as helpful as possible for people setting it up without bombarding them with content

Image

backend/src/server/routes/v1/certificate-authority-routers/adcs-certificate-authority-router.ts
```diff
@@ -0,0 +1,80 @@
...
+          )
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) => {
+      const projectId = req.internalCertManagerProjectId;
+      const templates = await server.services.certificateAuthority.getADCSTemplates({
+        caId: req.params.caId,
+        projectId,
```
sheensantoscapadngan, line 54: IGNORE PLACEMENT

it doesn't seem like this is necessary to mention here in the app connection docs? 

Image

sheensantoscapadngan (approved):

PR #7164, opened 2026-07-05
feat(pam): create PAM project lazily on first use

bernie-g:
Creates an org's consolidated PAM project lazily on first use instead of eagerly for every org during the migration. Targets `pam-revamp`.

- The migration now only creates a project for orgs that already have PAM data; the rest are skipped.
- Orgs with no PAM data get their project created the first time someone opens PAM, guarded by a per-org lock so simultaneous requests can't create duplicates.
- That first-use creation seeds the org's admins (users, identities, and groups) as PAM admins. This also fixes a pre-existing bug: the eager migration created projects for empty orgs with no admins, which nobody could then administer.

Tested on a fresh database: empty orgs get no project from the migration; a data org still migrates correctly; 12 simultaneous first-use requests produce exactly one project; and org admins (user/identity/group) are seeded while non-admins are not.

### DIFF
--- a/backend/src/ee/routes/v1/pam-routers/pam-project-router.ts
+++ b/backend/src/ee/routes/v1/pam-routers/pam-project-router.ts
@@ -10,6 +10,8 @@ export const registerPamProjectRouter = async (server: FastifyZodProvider) => {
     method: "GET",
     url: "/",
     schema: {
+      // Internal endpoint; PAM projects are never user-facing, so keep it out of the API docs
+      hide: true,
       operationId: "getPamProject",
       description: "Resolve the organization's Privileged Access Manager project, creating it on first access",
       tags: [ApiDocsTags.PamFolders],
@@ -21,8 +23,7 @@ export const registerPamProjectRouter = async (server: FastifyZodProvider) => {
     },
     config: { rateLimit: readLimit },
     onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
-    // injectPamProjectId (preValidation) resolves and lazily bootstraps the project, so by the time
-    // this handler runs internalPamProjectId is always set.
+    // injectPamProjectId (preValidation) has already resolved/bootstrapped the project by now.
     handler: async (req) => ({ projectId: req.internalPamProjectId })
   });
 };
--- a/backend/src/ee/routes/v1/pam-routers/pam-project-router.ts
+++ b/backend/src/ee/routes/v1/pam-routers/pam-project-router.ts
@@ -1,0 +1,28 @@
+import z from "zod";
+
+import { ApiDocsTags } from "@app/lib/api-docs/constants";
+import { readLimit } from "@app/server/config/rateLimiter";
+import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
+import { AuthMode } from "@app/services/auth/auth-type";
+
+export const registerPamProjectRouter = async (server: FastifyZodProvider) => {
+  server.route({
+    method: "GET",
+    url: "/",
+    schema: {
+      operationId: "getPamProject",
+      description: "Resolve the organization's Privileged Access Manager project, creating it on first access",
+      tags: [ApiDocsTags.PamFolders],
+      response: {
+        200: z.object({
+          projectId: z.string()
+        })
+      }
+    },
+    config: { rateLimit: readLimit },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    // injectPamProjectId (preValidation) resolves and lazily bootstraps the project, so by the time
+    // this handler runs internalPamProjectId is always set.
+    handler: async (req) => ({ projectId: req.internalPamProjectId })
+  });
+};
--- a/backend/src/ee/services/pam-project/pam-project-resolver.ts
+++ b/backend/src/ee/services/pam-project/pam-project-resolver.ts
@@ -1,33 +1,97 @@
-import { ProjectType } from "@app/db/schemas";
+import { Knex } from "knex";
+
+import { AccessScope, OrgMembershipRole, ProjectType, TableName } from "@app/db/schemas";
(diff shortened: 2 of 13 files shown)


backend/src/ee/services/pam-project/pam-project-resolver.ts
```diff
@@ -1,33 +1,97 @@
...
+
+  // Orgs with no PAM data get no consolidated project during the migration; the first PAM request
+  // creates it here, seeded with the org's current admins so it is immediately usable (otherwise
+  // no one could administer it — there is no org-admin fallback in the PAM permission model).
+  const ensureDefaultProject = (orgId: string): Promise<string> =>
+    db.transaction(async (tx) => {
+      // Serialize concurrent first-use bootstraps for the same org. There is no DB unique
+      // constraint to rely on (old zombie PAM projects also have type=pam, so an org can legitimately
+      // have several), and the advisory lock auto-releases when this transaction ends.
+      await tx.raw("SELECT pg_advisory_xact_lock(hashtext(?))", [`pam-bootstrap:${orgId}`]);
```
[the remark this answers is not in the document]
bernie-g: Craziest nit ever

backend/src/ee/services/pam-project/pam-project-resolver.ts
```diff
@@ -1,33 +1,106 @@
...
+
+      // Re-check inside the lock on the primary, with the same semantics as findDefaultProjectId
+      // (newest first, soft-deleted excluded). This resolves the race winner and, when several PAM
+      // projects exist (consolidated + old zombies), returns the consolidated one rather than an
+      // arbitrary row.
+      const existing = await tx(TableName.Project)
+        .where({ orgId, type: ProjectType.PAM })
+        .whereNull("deleteAfter")
+        .orderBy("createdAt", "desc")
+        .first("id");
```
x032205: I wonder if we can use findDefaultProjectId inplace of this query. We'd just need to add `.whereNull("deleteAfter")` to it.

backend/src/ee/services/pam-project/pam-project-resolver.ts
```diff
@@ -1,33 +1,106 @@
...
+      { orgId, type: ProjectType.PAM },
+      { sort: [["createdAt", "desc"]], limit: 1 }
+    );
+    return projects.length ? projects[0].id : null;
+  };
+
+  // Orgs with no PAM data get no consolidated project during the migration; the first PAM request
+  // creates it here, seeded with the org's current admins so it is immediately usable (otherwise
+  // no one could administer it, since the PAM permission model has no org-admin fallback).
+  const ensureDefaultProject = (orgId: string): Promise<string> =>
```
x032205, line 45: IGNORE PLACEMENT

`getOrgById` returns `pamProjectId` back to the frontend state, but that becomes null currently in cases where the project wasnt bootstrapped. We need to make `getOrgById` call `pamProjectResolver.resolve` I think.

backend/src/ee/services/pam-project/pam-project-resolver.ts
```diff
@@ -1,33 +1,106 @@
...
-            message: "This organization has no Privileged Access Manager project. Contact your administrator."
-          });
+export const pamProjectResolverFactory = ({
+  db,
+  projectDAL,
+  membershipDAL,
+  membershipRoleDAL,
+  keyStore
+}: TResolverDeps) => {
+  const findDefaultProjectId = async (orgId: string): Promise<string | null> => {
```
x032205: IGNORE PLACEMENT

We should mention this bootstrapping logic in PAM's CLAUDE.md

lb-vn (changes requested):

backend/src/ee/services/pam-project/pam-project-resolver.ts
bernie-g: Doing the project creation in `getOrgById` is a bad idea because then it happens anytime an org is loaded which defeats the point of this PR. Instead I created a tiny endpoint to fetch the pam project ID and we call that in the pam layout.tsx. We could also hack one of the existing endpoints to return the pam project ID if we don't want a whole new endpoint but it seems fine to me. lmk

backend/src/ee/services/pam-project/pam-project-resolver.ts
bernie-g: Done

backend/src/ee/services/pam-project/pam-project-resolver.ts
```diff
@@ -1,33 +1,103 @@
...
+  return {
+    resolve: (actorOrgId: string): Promise<string> =>
+      withCache({
+        keyStore,
+        key: KeyStorePrefixes.PamDefaultProject(actorOrgId),
+        ttlSeconds: KeyStoreTtls.PamDefaultProjectInSeconds,
+        fetcher: async () => {
+          const existingId = await findDefaultProjectId(actorOrgId);
+          if (existingId) return existingId;
+          return ensureDefaultProject(actorOrgId);
```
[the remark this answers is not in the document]
bernie-g: This isn't a privilege escalation and generally I don't think this really matters

backend/src/ee/routes/v1/pam-routers/pam-project-router.ts
```diff
@@ -0,0 +1,28 @@
...
+import { AuthMode } from "@app/services/auth/auth-type";
+
+export const registerPamProjectRouter = async (server: FastifyZodProvider) => {
+  server.route({
+    method: "GET",
+    url: "/",
+    schema: {
+      operationId: "getPamProject",
+      description: "Resolve the organization's Privileged Access Manager project, creating it on first access",
+      tags: [ApiDocsTags.PamFolders],
```
x032205, line 17: We should probably add `hide: true` to the schema in order to hide this endpoint from the OpenAPI spec (and our API docs). The concept of projects should never be user facing

lb-vn (changes requested):

backend/src/ee/routes/v1/pam-routers/pam-project-router.ts
bernie-g: Good call

lb-vn (approved):

PR #7170, opened 2026-07-06
feat: resolved vercel sync failing on conflict witht preview branch

akhilmhdh:
### Problem

Project-scoped Vercel syncs failed with `ENV_CONFLICT ... already exists for the target preview on branch undefined` when a key had more than one Vercel record (e.g. a branch-agnostic Preview record alongside a branch-specific one). The failing sync could also silently delete a branch-specific override.

### Root cause

 Vercel allows multiple project records with the same key when their `(target, gitBranch)` scopes don't overlap, so the key-only Map kept just one record and dropped the siblings. The loop then acted on the wrong record: it would delete a branch-specific record on a sensitivity change and create a branch-agnostic one, which collided with the surviving sibling the Map had hidden.

This was done correctly for Vercel Team,  but not for project one.

### Testing

Reproduced against a live Vercel project with two Preview records for one key (branch-agnostic + `staging`): the sync now updates the owned record

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -14,6 +14,7 @@ import { routeTree } from "./routeTree.gen";
 
 import "@fontsource/inter/400.css";
 import "@fontsource/inter/500.css";
+import "@fontsource/inter/600.css";
 import "@xyflow/react/dist/style.css";
 import "nprogress/nprogress.css";
 import "@fortawesome/fontawesome-svg-core/styles.css";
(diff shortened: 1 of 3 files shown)


frontend/src/main.tsx
adilsitos, line 1: is this change necessary? It does not seem related.

adilsitos (approved):

PR #7182, opened 2026-07-06
fix: correct License Server v2 dual-read comparison

PrestigePvP:
## Context

Found via EU prod dual-read (read-compare) logs showing a high error rate and a large diff flood. All fixes are on the dual-read shadow path plus the v2 key mapping it shares with the on-mode projection.

- **Crash fix**: `p.rateLimits.readLimit` in the v1 extractor threw when a v1 plan had `rateLimits` undefined, aborting the whole org's compare (this was `dual_read.error.count`). Optional-chain to `null` so a missing cap becomes a normal diff.
- **Wrong v2 keys**: the dual-read mapping used v2 keys absent from the License Server v2 registry, so those features were permanently `v2_missing` (and in on-mode the v2 -> v1 projection silently served free-tier defaults for them). Aligned with the authoritative v1 -> v2 map: `rate_limits_*` -> `{read,write,secrets}_rate_limit`, `pki_acme` -> `acme_client`.
- **Dead keys removed**: dropped `workspace_limit`, `member_limit`, `enforce_identity_limit`, `enterprise_certificate_syncs`, `pki_legacy_templates`, which have no v2 feature (`member_limit` is already covered by `identityLimit -> max_identities`). Excluded the corresponding v1 fields so the completeness test still passes.
- **Unlicensed orgs**: v2 resolves an org with no license to all-defaults (`source=default`), so comparing against the org's real v1 plan flooded diffs. Now compare v2 against the v1 free baseline (`getDefaultOnPremFeatures()`) when the response is entirely default.

Read-path safety: removals are read-neutral (the projection skips v2 keys absent from the response, keeping the base value); the renames change on-mode projection correctly (those features were served at free defaults before because the keys never matched); the unlicensed-compare change is shadow-path only.

## Steps to verify the change

- `tsc --noEmit` clean
- `eslint` clean on changed files
- dual-read comparator + projection tests pass (13/13)

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -167,6 +167,8 @@ export const licenseServiceFactory = ({
           .filter((tier) => tier !== "free");
         if (paidTiers.some((tier) => tier.includes("enterprise"))) {
           currentPlan.slug = "enterprise";
+        } else if (paidTiers.some((tier) => tier.includes("advanced"))) {
+          currentPlan.slug = "advanced";
         } else if (paidTiers.length > 0) {
           currentPlan.slug = "pro";
         }
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -317,6 +317,8 @@ export const licenseServiceFactory = ({
               .filter((tier) => tier !== "free");
             if (paidTiers.some((tier) => tier.includes("enterprise"))) {
               currentPlan.slug = "enterprise";
+            } else if (paidTiers.some((tier) => tier.includes("advanced"))) {
+              currentPlan.slug = "advanced";
             } else if (paidTiers.length > 0) {
               currentPlan.slug = "pro";
             }
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -366,7 +366,7 @@ export const licenseServiceFactory = ({
 
         // read-compare bake: serve v1 but shadow-compare against v2 in the background ahead of the cutover.
         if (envConfig.LICENSE_SERVER_V2_MODE === "read-compare") {
-          licenseDualRead?.compareInBackground(rootOrgId, currentPlan);
+          licenseDualRead?.compareInBackground(rootOrgId, currentPlan, getDefaultOnPremFeatures);
         }
 
         return currentPlan;
--- a/backend/src/ee/services/license/license-service.ts
+++ b/backend/src/ee/services/license/license-service.ts
@@ -370,7 +370,7 @@ export const licenseServiceFactory = ({
 
         // read-compare bake: serve v1 but shadow-compare against v2 in the background ahead of the cutover.
         if (envConfig.LICENSE_SERVER_V2_MODE === "read-compare") {
-          licenseDualRead?.compareInBackground(rootOrgId, currentPlan, getDefaultOnPremFeatures);
+          licenseDualRead?.compareInBackground(rootOrgId, currentPlan);
         }
 
         return currentPlan;
--- a/backend/src/services/license-client/dual-read/dual-read-service.ts
+++ b/backend/src/services/license-client/dual-read/dual-read-service.ts
@@ -19,7 +19,7 @@ export const dualReadServiceFactory = ({ licenseClient, envConfig }: TDualReadSe
   // Fire-and-forget shadow compare against the already-resolved v1 plan. Never awaited, never throws into
   // the request path; inert unless mode is read-compare. Reads v2 through the real client SDK so the
   // production read path is exercised under real traffic ahead of the cutover.
-  const compareInBackground = (orgId: string, planV1: TFeatureSet) => {
+  const compareInBackground = (orgId: string, planV1: TFeatureSet, getFreeTierPlan: () => TFeatureSet) => {
     if (!envConfig.isLicenseDualReadEnabled) {
       return;
     }
@@ -32,7 +32,16 @@ export const dualReadServiceFactory = ({ licenseClient, envConfig }: TDualReadSe
         return;
       }
 
-      const discrepancies = compareEntitlements(planV1, entitlements, FEATURE_MAPPINGS);
+      // v2 resolves an org with no license to the free tier (every feature source=default). Comparing that
+      // against the org's real v1 plan is noise, so compare v2 against the v1 free baseline instead.
+      const featureValues = Object.values(entitlements.features);
+      const isV2Unlicensed = featureValues.length > 0 && featureValues.every((f) => f.source === "default");
+      let planForCompare = planV1;
+      if (isV2Unlicensed) {
+        planForCompare = getFreeTierPlan();
+      }
+
+      const discrepancies = compareEntitlements(planForCompare, entitlements, FEATURE_MAPPINGS);
       const diffs = discrepancies.filter((d) => d.kind !== DualReadDiffKind.Match);
 
       for (const d of diffs) {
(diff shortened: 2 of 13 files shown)


backend/src/services/license-client/dual-read/dual-read-service.ts
```diff
@@ -32,7 +32,16 @@ export const dualReadServiceFactory = ({ licenseClient, envConfig }: TDualReadSe
         return;
       }
 
-      const discrepancies = compareEntitlements(planV1, entitlements, FEATURE_MAPPINGS);
+      // v2 resolves an org with no license to the free tier (every feature source=default). Comparing that
+      // against the org's real v1 plan is noise, so compare v2 against the v1 free baseline instead.
+      const featureValues = Object.values(entitlements.features);
+      const isV2Unlicensed = featureValues.length > 0 && featureValues.every((f) => f.source === "default");
```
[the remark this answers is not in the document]
PrestigePvP: Intentional: the check requires an explicit all-default response and only then swaps to the free baseline. If `source` is ever absent (version skew), `every(=== "default")` is false and we fall back to comparing the org's real v1 plan, which is the safe default (compare real data rather than fabricate a free baseline). The v2 resolver sets source on every feature today, and this is shadow-path only, so there's no serving impact either way.

scott-ray-wilson: it seems like we don't set default on all but just return an empty set? or is this populated elsewhere?
```
// freeResolution is the entitlement set for an org with no active license: no products and no
// features, so consumers fall back to their free-tier defaults.
func freeResolution(orgID string) entitlement.Resolution {
	return entitlement.Resolution{
		License:  entitlement.LicenseInfo{LicenseID: "", OrgID: orgID, BillingMethod: "", DeploymentMode: "cloud", Status: "free"},
		Account:  entitlement.AccountInfo{SfdcAccountID: "", Name: ""},
		Products: []entitlement.ProductEntitlement{},
		Features: map[string]entitlement.Feature{},
	}
}
```

backend/src/services/license-client/dual-read/feature-mapping/limits.ts
```diff
@@ -1,18 +1,6 @@
 import { TFeatureMapping, unlimitedWhenNull } from "../dual-read-types";
 
 export const limitsMappings: TFeatureMapping[] = [
-  {
-    v2Key: "workspace_limit",
-    v1Field: "workspaceLimit",
-    extractV1: (p) => p.workspaceLimit,
-    normalize: unlimitedWhenNull
-  },
```
scott-ray-wilson, line 9: doesn't this exist as `secrets_management_workspace_limit` in the seed script? or are we not using it?

backend/src/ee/services/license/license-service.ts
```diff
@@ -366,7 +368,7 @@ export const licenseServiceFactory = ({
 
         // read-compare bake: serve v1 but shadow-compare against v2 in the background ahead of the cutover.
         if (envConfig.LICENSE_SERVER_V2_MODE === "read-compare") {
-          licenseDualRead?.compareInBackground(rootOrgId, currentPlan);
+          licenseDualRead?.compareInBackground(rootOrgId, currentPlan, getDefaultOnPremFeatures);
```
scott-ray-wilson: It looks like our cloud free tier that is seeded deviates from on prem defaults

| field | on-prem default (what the PR compares) | v1 cloud starter | v2 catalog default |
|---|---|---|---|
| rateLimits.readLimit | 60 | 200 | 200 |
| rateLimits.writeLimit | 200 | 90 | 90 |
| rateLimits.secretsLimit | 40 | 120 | 120 |
| secretVersioning | true | false | false |
| pkiAcme / acme_client | true | false | false |
| auditLogStreamLimit | 3 | 0 | 0 |
| auditLogsRetentionDays | 0 | 0 | 30 |
| environmentLimit | null (unlimited) | 2 | 3 |
| identityLimit / max_identities | null (unlimited) | 5 | 5 |

backend/src/ee/services/license/license-service.ts
```diff
@@ -317,6 +317,8 @@ export const licenseServiceFactory = ({
               .filter((tier) => tier !== "free");
             if (paidTiers.some((tier) => tier.includes("enterprise"))) {
               currentPlan.slug = "enterprise";
+            } else if (paidTiers.some((tier) => tier.includes("advanced"))) {
+              currentPlan.slug = "advanced";
```
scott-ray-wilson, line 323: do we need to add this to `syncSelfHostedV2Features` too? it has the same tier branching

frontend/src/hooks/api/subscriptions/fns.ts
```diff
@@ -13,5 +13,8 @@ export const getSubscriptionPlanLabel = (subscription: SubscriptionPlan, suffix
   if (!slug || slug === SubscriptionPlanTypes.Starter) {
     return `Free${suffix}`;
   }
+  if (slug === SubscriptionPlanTypes.Advanced) {
+    return `Advanced${suffix}`;
+  }
```
scott-ray-wilson, line 18: I think we also need to update PreviewSection.tsx otherwise advanced shows the upgrade banner

```
return (
    <div>
      {subscription &&
        subscription?.slug !== SubscriptionPlanTypes.Enterprise &&
        subscription?.slug !== SubscriptionPlanTypes.Pro &&
        subscription?.slug !== SubscriptionPlanTypes.ProAnnual && (
```

backend/src/services/license-client/dual-read/dual-read-service.ts
```diff
@@ -32,7 +32,16 @@ export const dualReadServiceFactory = ({ licenseClient, envConfig }: TDualReadSe
...
+      let planForCompare = planV1;
+      if (isV2Unlicensed) {
+        planForCompare = getFreeTierPlan();
```
scott-ray-wilson: I'm not fulling tracking this comment but Claude surfaced it locally:

When this swap fires for an org whose v1 plan is paid (real plan in v1, license never
provisioned in v2), the compare now reports `diffs=0`: identical telemetry to a genuinely
free org. Those are exactly the orgs that get downgraded to free at cutover, and neither
the per-diff metric (feature+kind only) nor the summary log records that the swap happened
or what `planV1.slug` was, so the bake can't surface them.

`planV1` is already in scope and the free baseline has `slug: null`, so one guarded line
keeps the noise reduction while making the unmigrated-paid population visible and countable:

​```ts
if (isV2Unlicensed && planV1.slug && planV1.slug !== "starter") {
  logger.warn(`license-dual-read: paid v1 org unlicensed in v2 [orgId=${orgId}] [v1Slug=${planV1.slug}]`);
}
​```

scott-ray-wilson (changes requested):
few questions, some of these may be known/intentional

backend/src/ee/services/license/license-service.ts
PrestigePvP: Claude added it but I removed it because in the case where the entitlement fails for self hosted it won't have an advanced tier returned. It'll either be enterprise or fall through to pro.

backend/src/services/license-client/dual-read/dual-read-service.ts
PrestigePvP: Good catch here, the intention was that `Features` returned a list of the feature defaults. Corrected here: https://github.com/Infisical/license-server-go/pull/53

backend/src/services/license-client/dual-read/feature-mapping/access.ts
```diff
@@ -51,11 +51,6 @@ export const accessMappings: TFeatureMapping[] = [
     v1Field: "githubOrgSync",
     extractV1: (p) => p.githubOrgSync
   },
-  {
-    v2Key: "enforce_identity_limit",
```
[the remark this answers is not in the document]
PrestigePvP: `enforce_identity_limit` has no corresponding License Server v2 feature, so there is nothing for the projection to source in v2 mode. This mapping was dropped in an earlier commit on the PR (not the latest push); preserving enterprise identity-limit enforcement at cutover is a separate design decision, tracked outside this shadow-compare fix. Resolving.

scott-ray-wilson (approved):
LGTM

Issue #50, opened 2022-11-28
Add code linter to Repo


maidul98:
Having a code linter would standardize our format and catch invalid code during pull request. It would be best to set up a Github action such as https://github.com/marketplace/actions/super-linter. I have attempted to add the linter but have ran into the following issue. 

```
level=warning msg="[runner] Can't run linter goanalysis_metalinter: inspect: failed to load package : could not load export data: no export data for \"github.com/Infisical/infisical/packages/cmd\""
level=error msg="Running error: 1 error occurred:
	* can't run linter goanalysis_metalinter: inspect: failed to load package : could not load export data: no export data for \"github.com/Infisical/infisical/packages/cmd\"
```

This needs further investigation

PR #7184, opened 2026-07-06
improvement(license-client): switch to org-scoped license-server paths

PrestigePvP:
## Context

The license server is reshaping its public API so org-scoped calls carry the org in the path instead of an `?org_id=` query param (Infisical/license-server-go#50). This updates the cloud license-client to match.

- Cloud backend builds `/v1/organizations/{orgId}/...` via an `orgScoped` helper and drops the `org_id` query param (keeps `org_name`/`org_slug`/`region`)
- `usage-reporter` posts to `/v1/organizations/{orgId}/usage-snapshots`
- Self-hosted (license-key) backend unchanged: token-scoped bare paths carry no org_id

Pairs with Infisical/license-server-go#50. Hard cutover for the service surface, so deploy together; safe under `LICENSE_SERVER_V2_MODE=read-compare`.

## Steps to verify the change

- `npm run type:check` (backend) passes
- `eslint` clean on the changed files

## Type

- [x] Improvement
- [x] Breaking

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/license-client/license-client-backends.ts
+++ b/backend/src/services/license-client/license-client-backends.ts
@@ -27,18 +27,15 @@ import {
   TSubscriptionResponse
 } from "./license-client-types";
 
+// Token-scoped paths for the self-hosted (license-key) backend: the key identifies the license, so
+// no org_id is carried. The global catalog is org-independent and shared by both backends.
 const ENTITLEMENTS_PATH = "/v1/entitlements";
 const ENTITLEMENTS_REFRESH_PATH = "/v1/entitlements/refresh";
 const PRODUCTS_PATH = "/v1/products";
 const SUBSCRIPTION_PATH = "/v1/subscription";
-const CLOUD_PLAN_PATH = "/v1/cloud-plan";
-const BILLING_PROFILE_PATH = "/v1/billing/profile";
-const CHECKOUT_SESSION_PATH = "/v1/billing/checkout-session";
-const PORTAL_SESSION_PATH = "/v1/billing/portal-session";
-const SUBSCRIPTION_PREVIEW_PATH = "/v1/billing/subscription/preview";
-const SUBSCRIPTION_ITEMS_PATH = "/v1/billing/subscription/items";
-const SUBSCRIPTION_CANCEL_PATH = "/v1/billing/subscription/cancel";
-const SUBSCRIPTION_RESUME_PATH = "/v1/billing/subscription/resume";
+
+// Cloud (service-JWT) callers address the org in the path instead of an org_id query param.
+const orgScoped = (orgId: string, suffix: string): string => `/v1/organizations/${encodeURIComponent(orgId)}${suffix}`;
 
 // Issuer/audience/subject match the license server's in-code constants (it validates iss/aud against
 // them). They are public claims, not per-deployment config, so they live here rather than in env.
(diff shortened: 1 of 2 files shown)


backend/src/services/license-client/usage/usage-reporter.ts
scott-ray-wilson, line 1: ignore file placement:

we should also update the readme (and any other references) to the original urls:

```
### 3. Call the API

Service-account callers must pass `org_id`, and that org must already have a
license in the local DB (the server maps org to license):

```bash
curl 'http://localhost:4040/v1/entitlements?org_id=<ORG_ID>' \
  -H "Authorization: Bearer <JWT>"
```

scott-ray-wilson: also claude.md:

```
### Three request surfaces (see `internal/server/api/api.go`)

1. **Public client API** (`/v1/entitlements`, `/v1/usage/snapshots`, `/v1/products`, `/v1/subscription`, `/v1/cloud-plan`, `/v1/billing/*`). Guarded by `auth.RequireAuth`, which accepts either a **license key** (`Authorization: Bearer infisical_lk_...`, bound to one license) or a **cloud service JWT** (no prefix, not license-bound, so the caller must pass `?org_id=`). Service JWT is off unless `SERVICE_JWT_ENABLED=true`.
2. **Stripe webhook** (`POST /v1/webhooks/stripe`). Authenticated by Stripe signature **inside** the handler, so no auth middleware is applied at the route.
3. **Admin API + SPA** (`/v1/admin/*` and the root catch-all SPA). Guarded by `adminauth` (Google Workspace OIDC session; `/auth/login` + `/auth/callback` are outside the gated group). Fails closed with 401 unless OAuth is configured or `ADMIN_DEV_AUTH_BYPASS=true` is set locally. The server refuses to boot if bypass is combined with a real `ADMIN_OAUTH_CLIENT_ID`. The SPA is served from root via a `Handle("/*")` catch-all registered last (chi panics if you `Mount("/")` alongside other routes).

```

scott-ray-wilson (changes requested):
LGTM just nits on documentation drift

scott-ray-wilson (approved):
LGTM

scott-ray-wilson (approved):

backend/src/services/license-client/license-client-backends.ts
```diff
@@ -218,8 +208,7 @@ export const licenseServerBackend = (
     orgId: string,
     payload: TSubscriptionPreviewPayload
   ): Promise<TSubscriptionPreview> => {
-    const url = new URL(SUBSCRIPTION_PREVIEW_PATH, serverUrl);
-    url.searchParams.set("org_id", orgId);
+    const url = new URL(orgScoped(orgId, "/subscription/preview"), serverUrl);
```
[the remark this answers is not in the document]
PrestigePvP: Intentional and consistent with the paired server change (Infisical/license-server-go#50). The license server now serves subscription mutations at `/v1/organizations/{org_id}/subscription/{preview,items,cancel,resume}` (the `billing/` segment was dropped there too), and this client calls exactly those. Checkout/portal/profile stay under `/billing/` on both sides. The `/:organizationId/billing/v2/subscription/...` routes you referenced are infisical's own internal billing-v2 router, not the license-server API, so they are unaffected. Confirmed the #50 server routes match these client paths.

backend/src/services/license-client/usage/usage-reporter.ts
PrestigePvP: Thanks. Fixed in the license-server repo (where the README lives): Infisical/license-server-go#50 replaces the `?org_id=` examples with the org-scoped `/v1/organizations/{org_id}/...` paths.

PrestigePvP: Same, updated the CLAUDE.md "Three request surfaces" section in Infisical/license-server-go#50 to describe the token-scoped vs org-addressed surfaces. No code change needed on this PR.

PR #7185, opened 2026-07-06
feat: make certificate properties optional for ADCS CAs

carlosmonastyrski:
## Context

This PR streamlines the certificate request form when issuing through an external CA like ADCS. It now shows only the fields that apply and skips the ones the external CA's template already manages.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


sheensantoscapadngan (commented):
IGNORE PLACEMENT

renewing fails here....

Image

backend/src/services/certificate-policy/certificate-policy-service.ts
```diff
@@ -387,12 +387,8 @@ export const certificatePolicyServiceFactory = ({
           errors.push(`${attrType} is not allowed by template policy (not defined in template)`);
         }
       }
-    } else if (requestAttributes.size > 0) {
-      // No subject policies defined but request has subject attributes - deny all
-      for (const [attrType] of requestAttributes) {
-        errors.push(`${attrType} is not allowed by template policy (no subject policies defined)`);
-      }
     }
+    // No subject policy defined means no subject constraint (allow all)
```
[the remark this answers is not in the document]
carlosmonastyrski: Making omitted policy sections mean "allow all" is the intended product behavior here (policy properties are optional). CA issuance is the one privilege-boundary exception and is unaffected: `basicConstraints.isCA` still defaults to `DENIED`, so a policy must explicitly opt into CA issuance. Restrictive-by-default for the other sections is the previous behavior this change intentionally removes.

backend/src/services/certificate-policy/certificate-policy-service.ts
```diff
@@ -308,7 +308,7 @@ export const certificatePolicyServiceFactory = ({
   const validateRequestAgainstPolicy = (
     template: TCertificatePolicy,
     request: TCertificateRequest,
-    options?: { skipRequired?: boolean }
+    options?: { skipRequired?: boolean; skipCaManagedFieldValidation?: boolean }
   ): TPolicyValidationResult => {
     const errors: string[] = [];
```
sheensantoscapadngan, line 329: ignore placement of comment

this would actually be a breaking change for folks with empty subject attributes policies (and SAN) because before this, it would be "REJECT ALL" whereas after, it gets treated as an "ALLOW ALL" rule...

Image

sheensantoscapadngan: Maybe we should make it so that we only ALLOW ALL if the subject, SAN, key usages, or whatever field is undefined rather than an empty array? If it's an empty array, it seems like it would be more fitting to refer to that as "nothing allowed".

frontend/src/pages/cert-manager/CertificatePolicyDetailsByIDPage/CertificatePolicyDetailsByIDPage.tsx
```diff
@@ -62,6 +63,21 @@ const Page = () => {
...
+      policy?.keyUsages?.allowed?.length ||
+      policy?.keyUsages?.denied?.length ||
+      policy?.extendedKeyUsages?.required?.length ||
+      policy?.extendedKeyUsages?.allowed?.length ||
+      policy?.extendedKeyUsages?.denied?.length ||
+      policy?.algorithms?.signature?.length ||
+      policy?.algorithms?.keyAlgorithm?.length ||
+      policy?.validity?.max ||
+      policy?.basicConstraints?.isCA
+  );
```
sheensantoscapadngan, line 75: is this outdated? An empty array for subject means reject all, which should really mean there is enforcement, right? I know via UI you can't create a policy with an empty array for Subject or for Sans, but this is currently possible via API...

backend/src/services/certificate-policy/certificate-policy-service.ts
```diff
@@ -308,7 +308,7 @@ export const certificatePolicyServiceFactory = ({
   const validateRequestAgainstPolicy = (
     template: TCertificatePolicy,
     request: TCertificateRequest,
-    options?: { skipRequired?: boolean }
+    options?: { skipRequired?: boolean; skipCaManagedFieldValidation?: boolean }
```
sheensantoscapadngan: is this part here still needed? asking because I feel like this is just something that we want to do in the frontend to show the fields that people should be filling up or to hide ones where it gets auto-filled by the certificate authority. if they're using ADCS, they shouldn't have any policies defined for the key usages, extended key usages, validity, etc right?  I think our logic right now will just skip this altogether if so.. 

it might be better to be explicit rather than have the backend policy just bypass the defined rules silently if for some weird reason somebody using ADCS CA will define key usages, etc.

frontend/src/pages/cert-manager/PoliciesPage/components/CertificateProfilesTab/CreateProfileModal.tsx
```diff
@@ -576,7 +609,28 @@ export const CreateProfileModal = ({ isOpen, onClose, profile, mode = "create" }
...
+        if (!sanTypes.includes(sanPolicy.type as CertSubjectAlternativeNameType)) {
+          sanTypes.push(sanPolicy.type as CertSubjectAlternativeNameType);
+        }
+      });
+      allowedSanTypes = sanTypes;
+    } else {
+      allowedSanTypes = Object.values(
+        CertSubjectAlternativeNameType
+      ) as CertSubjectAlternativeNameType[];
     }
```
sheensantoscapadngan, line 626: is this part still correct with regards to:

undefined = allow all
[] = allow NONE

frontend/src/pages/cert-manager/PoliciesPage/components/CertificateProfilesTab/CreateProfileModal.tsx
```diff
@@ -541,33 +564,43 @@ export const CreateProfileModal = ({ isOpen, onClose, profile, mode = "create" }
...
+      ? templateData.algorithms.keyAlgorithm.map((templateAlgorithm: string) => {
+          const apiAlgorithm = mapPolicyKeyAlgorithmToApi(templateAlgorithm);
+          return { value: apiAlgorithm, label: apiAlgorithm };
+        })
+      : certKeyAlgorithms.map((option) => ({ value: option.value as string, label: option.label }));
 
-    let allowedSubjectAttributeTypes: CertSubjectAttributeType[] = [];
-    let shouldShowSubjectSection = false;
+    let allowedSubjectAttributeTypes: CertSubjectAttributeType[];
     if (templateData.subject && templateData.subject.length > 0) {
```
sheensantoscapadngan: same here... seems like it's not consistent with the rule?

sheensantoscapadngan (approved):

PR #7187, opened 2026-07-07
fix(secret-blind-index): skip soft-deleted projects in helper queries

jaydeep-pipaliya:
## What

Three helper queries in \`secret-blind-index-dal.ts\` — \`countOfSecretsWithNullSecretBlindIndex\`, and two find variants — leftJoin \`project_environments\` with \`andOnNull(Environment.deleteAfter)\` but never touch \`projects\`. Soft-deleting a project does not soft-delete its environments, so blind-index maintenance helpers keep counting / returning secret rows against projects sitting in the delete grace window.

## Fix

Adds an active-project leftJoin \`andOnNull(Project.deleteAfter)\` after each env join. The existing \`where({ projectId })\` predicate scopes correctly; and if the project has been soft-deleted, env.projectId is nulled out by the env-side \`andOnNull\`, causing the project join to fail so the row is effectively excluded from the aggregation.

Same shape as the previously shipped webhook (#6970), integration (#6983), secret-rotation-v2 (#7002), dynamic-secret (#7065), snapshot (#7066), folder-commit (#7067), honey-token (#7088), secret-import (#7089), secret-approval-request (#7090, merged), access-approval-request (#7105), access-approval-policy (#7106), secret-approval-policy (#7107), reminder (#7124), app-connection (#7125), secret-approval-request-secret (#7126), secret-folder-version (#7155), secret-version v1 (#7156), offline-usage-report (#7157), secret-v2-bridge/secret-version (#7167), project-folder-grant (#7168), and telemetry (#7169) fixes.

## Checklist

### DIFF
--- a/backend/src/services/secret-blind-index/secret-blind-index-dal.ts
+++ b/backend/src/services/secret-blind-index/secret-blind-index-dal.ts
@@ -13,17 +13,14 @@ export const secretBlindIndexDALFactory = (db: TDbClient) => {
   const countOfSecretsWithNullSecretBlindIndex = async (projectId: string, tx?: Knex) => {
     try {
       const doc = await (tx || db.replicaNode())(TableName.Secret)
-        .leftJoin(TableName.SecretFolder, `${TableName.SecretFolder}.id`, `${TableName.Secret}.folderId`)
-        .leftJoin(TableName.Environment, function joinActiveEnvForFolder() {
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Both right — leftJoin + andOnNull just nulls the joined columns, it doesn't drop the row. Pushed 86ce1de that converts all three helpers to innerJoin on SecretFolder/Environment/Project with andOnNull for the env deleteAfter and a whereNull(Project.deleteAfter) after the project join. Matches the canonical pattern from org-product-stats-dal.ts and the folder-commit fix in #7067.

PR #7200, opened 2026-07-07
fix: dashboard field empty issue with import paths

akhilmhdh:
## Context

On the Secret Overview page, a secret that has a real value in an environment could render as **EMPTY** when that same key is also brought in by an import (normal or replicated) whose value is empty. The environment comparison grid correctly showed a green "present" tick, but opening the per-env view or the expanded multi-env compare showed no value.

## Reproduction

1. In env A (e.g. `dev`), create a secret `TEST` with an empty value.
2. In env B (e.g. `staging`), create `TEST` with a real value, and add an import of env A at the same path.
3. Open the Secret Overview. Env B shows a green tick for `TEST`, but the per-env view and the expanded compare show EMPTY.

After the fix, env B shows the real local value in both views while the tick stays green. An environment that only receives the key via an empty import still correctly shows as empty.

## Screenshots

Screenshot 2026-07-08 at 12 37 47 AM

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretTableRow.tsx
+++ b/frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/SecretTableRow.tsx
@@ -307,7 +307,9 @@ export const SecretTableRow = ({
             secretPath={secretPath}
             isVisible={isSecretVisible || isSingleEnvSecretsVisible}
             secretName={secretKey}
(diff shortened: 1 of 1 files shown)


scott-ray-wilson (approved):
LGTM, verified locally

PR #7207, opened 2026-07-08
feat: support JSON and plain text variables in CF workers sync

mathnogueira:
## Context

Secret name conflicts with non-secret variables in Cloudflare would cause the sync to fail. Now, it saves the secret as the original type of the variable: secret, plain_text or JSON.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
+++ b/backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
@@ -68,8 +68,10 @@ const getSecretKeys = async (secretSync: TCloudflareWorkersSyncWithCredentials):
 
   const secrets = secretsResponse.data.result.map((s) => ({ key: s.name, type: s.type }));
 
+  const SYNCABLE_BINDING_TYPES = new Set(["plain_text", "json"]);
+
   const nonSecretBindings = (settingsResponse.data.result.bindings || [])
-    .filter((b) => b.type !== "secret_text")
+    .filter((b) => SYNCABLE_BINDING_TYPES.has(b.type))
     .map((b) => ({ key: b.name, type: b.type }));
 
   const secretKeySet = new Set(secrets.map((s) => s.key));
--- a/backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
+++ b/backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
@@ -60,9 +60,14 @@ const throwOnUndeployedVersionError = (err: unknown) => {
   throw err;
 };
 
-const getSecretKeys = async (
+type TCloudflareBindings = {
+  secrets: TCloudflareSecretMetadata[];
+  nonSecretBindings: TCloudflareSecretMetadata[];
+};
+
+const getCloudflareBindings = async (
   secretSync: TCloudflareWorkersSyncWithCredentials
-): Promise<TCloudflareSecretMetadata[]> => {
+): Promise<TCloudflareBindings> => {
   const {
     destinationConfig,
     connection: {
@@ -94,14 +99,14 @@ const getSecretKeys = async (
 
   const SYNCABLE_BINDING_TYPES = new Set(["plain_text", "json"]);
 
+  const secretKeySet = new Set(secrets.map((s) => s.key));
   const nonSecretBindings = (settingsResponse.data.result.bindings || [])
-    .filter((b) => SYNCABLE_BINDING_TYPES.has(b.type))
+    .filter((b) => SYNCABLE_BINDING_TYPES.has(b.type) && !secretKeySet.has(b.name))
     .map((b) => ({ key: b.name, type: b.type }));
 
-  const secretKeySet = new Set(secrets.map((s) => s.key));
-  const uniqueNonSecretBindings = nonSecretBindings.filter((v) => !secretKeySet.has(v.key));
-
-  return [...secrets, ...uniqueNonSecretBindings];
+  // We need to know both types because cloudflare prevents secrets and variables (plaintext or JSON)
+  // to have the same name.
+  return { secrets, nonSecretBindings };
 };
 
 export const CloudflareWorkersSyncFns = {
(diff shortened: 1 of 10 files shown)


mathnogueira:
@greptile review again

mathnogueira:
@greptile review this once more

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
```diff
@@ -33,27 +38,48 @@ const throwOnUndeployedVersionError = (err: unknown) => {
...
+  const SYNCABLE_BINDING_TYPES = new Set(["plain_text", "json"]);
+
+  const nonSecretBindings = (settingsResponse.data.result.bindings || [])
+    .filter((b) => SYNCABLE_BINDING_TYPES.has(b.type))
+    .map((b) => ({ key: b.name, type: b.type }));
+
+  const secretKeySet = new Set(secrets.map((s) => s.key));
+  const uniqueNonSecretBindings = nonSecretBindings.filter((v) => !secretKeySet.has(v.key));
+
+  return [...secrets, ...uniqueNonSecretBindings];
```
adilsitos: I notice that we merge everything here and return it inside the array, and on the syncSecrets method we create a json with unique keys on line 95, couldn't this break the sync and overwrite secrets as non secret value?

adilsitos: I did a validation and the scenario I added before is not possible because cloudflare does not allow it, but maybe it is still worth changing it on the code so we don't get wrong conclusions by just lookin at this. Does this make sense?

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
```diff
@@ -65,11 +91,25 @@ export const CloudflareWorkersSyncFns = {
...
+    const existingSecrets = await getSecretKeys(secretSync);
+    const existingSecretsMap = Object.fromEntries(existingSecrets.map(({ key, type }) => [key, type]));
     const secretMapKeys = new Set(Object.keys(secretMap));
 
+    const bindingEntries: Array<[string, { value: string }]> = [];
+    const secretEntries: Array<[string, { value: string }]> = [];
+
+    for (const [key, val] of Object.entries(secretMap)) {
+      const existingType = existingSecretsMap[key];
+      if (existingType && existingType !== "secret_text") {
```
adilsitos: nit: `secret_text` could be a constant and reused

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
```diff
@@ -82,26 +122,117 @@ export const CloudflareWorkersSyncFns = {
...
+        const updatedBindings = settingsData.result.bindings.map((binding) => {
+          if (binding.type !== "secret_text" && updatedBindingMap[binding.name] !== undefined) {
+            const newValue = updatedBindingMap[binding.name].value;
+            if (binding.type === "json") {
+              try {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                return { ...binding, json: JSON.parse(newValue) };
+              } catch {
+                throw new BadRequestError({
+                  message: `"${binding.name}" already exists in cloudflare as a JSON variable: the value you provided is not valid JSON`
```
adilsitos: I think the right message here should be that the json is not valid right? Because The only problem I see is if the `JSON.parse` raises an error.

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
```diff
@@ -82,26 +122,117 @@ export const CloudflareWorkersSyncFns = {
           }
         );
       }
+      /* eslint-enable no-await-in-loop */
+
+      if (bindingEntries.length > 0) {
```
adilsitos, line 153: I would change the order that the requests happen here. I noticed that we first update the secret values (what is the current behavior) and later update the json/text values. The json case can raise an exception, which would make this sync partial correct. Maybe it is better validate that all the json values are fine and then perform the request to update the secrets and the json/text

adilsitos: This would not happen on the first interaction, but if someone changes the value inside infisical to an invalid json, this can break

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
adilsitos, line 1: Ignore comment placement. 

Not sure if we discussed it, but this will change the behavior of the sync. Previously we were just supporting `secrets`, but this will also allow the change of JSON/Text env variables (no problem). But, the values will continue to have this type, so maybe we could add a message on the UI explaining that we don't change the type, of those secrets. If I'm not mistaken, we do this for vercel, explaining the sync behavior. 

Image

adilsitos, line 1: Ignore comment placement 

Another thing that we could also have is an option to the user to select in which type the secrets will be saved. We could have a dropdown here. 

Image

I think this can be even more helpful because the behavior of the sync will change. I'm afraid of users that have this sync can lose environments because of this change

adilsitos (changes requested):

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
mathnogueira: I wanted to add the first part just to make sure user understands why an invalid JSON is failing, otherwise they might not know this is happening because CF already has a variable with the same name.

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
```diff
@@ -33,27 +60,53 @@ const throwOnUndeployedVersionError = (err: unknown) => {
...
+      `${IntegrationUrls.CLOUDFLARE_WORKERS_API_URL}/client/v4/accounts/${accountId}/workers/scripts/${destinationConfig.scriptId}/secrets`,
+      { headers }
+    ),
+    request.get<{
+      result: { bindings: Array<{ type: string; name: string; text?: string }> };
+    }>(
+      `${IntegrationUrls.CLOUDFLARE_WORKERS_API_URL}/client/v4/accounts/${accountId}/workers/scripts/${destinationConfig.scriptId}/settings`,
+      { headers }
+    )
+  ]);
```
mathnogueira, line 96: Oh no, we need the second request even if `shouldSyncNonSecretBindings` is false. We need to know the variable keys that are plaintext and JSON to filter from the secrets, otherwise sync will fail with:

> Name `EXISTING_VARIABLE` already exists.

Example:

| SOURCE | KEY | Type  | 
|---|---|---|
| Infisical  | URL  | secret  | 
| Infisical  | PASSWORD  | secret  |   
| Infisical | USERNAME | secret |
| Cloudflare  | WORKER_PAYLOAD  | JSON  |  
| Cloudflare | USERNAME | PlainText |


If I don't call the second request, I don't know CF has a `USERNAME`. If I try to sync `URL`, `PASSWORD`, `USERNAME`, it will fail because I'm trying to create a secret called `USERNAME` while there is a Plaintext variable called `USERNAME`. 

So I have to filter out `USERNAME` to ensure the sync will work.

frontend/src/components/secret-syncs/forms/SecretSyncOptionsFields/CloudflareWorkersSyncOptionsFields.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <Controller
+        name="syncOptions.syncNonSecretBindings"
+        control={control}
+        render={({ field: { value, onChange }, fieldState: { error } }) => (
+          <Field className="mb-4">
+            <Field orientation="horizontal">
+              <FieldContent>
+                <Label htmlFor="sync-non-secret-bindings">Sync plaintext and JSON variables</Label>
+                <FieldDescription>
+                  When enabled, Infisical will also sync plaintext and JSON variable bindings in
```
adilsitos, line 38: I think it is also good to say that they wont be able to create json/text from inifiscal, only update the json/text values they have on cf.

backend/src/services/secret-sync/cloudflare-workers/cloudflare-workers-fns.ts
adilsitos, line 96: nit: maybe the second request should be required only if the shouldSyncNonSecretBindings is true? Otherwise we area calling it and not using any of it.

adilsitos (approved):

PR #7223, opened 2026-07-08
feat: add metrics for secrets

mathnogueira:
## Context

Add new metrics for the secret management product.

## Metrics (grafana syntax)

1. Rotation failure rate

```
A: sum by (type) (ceil(increase(infisical_secret_rotation_outcome_count_total[$__range])))
B: sum by (type) (ceil(increase(infisical_secret_rotation_outcome_count_total{outcome="failure"}[$__range])))
C: B/A*100
```

2. Slow secret reads
``` 
histogram_quantile(0.95, sum by (le) (rate(infisical_secret_operation_duration_bucket{operation="read"}[10m])))
```

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -3831,14 +3855,14 @@ export const secretV2BridgeServiceFactory = ({
...
-    getSecretByName,
-    getSecrets,
+    createSecret: withSecretMetrics(createSecret, { duration: "write", write: "create" }),
+    deleteSecret: withSecretMetrics(deleteSecret, { duration: "delete", write: "delete" }),
+    updateSecret: withSecretMetrics(updateSecret, { duration: "write", write: "update" }),
+    createManySecret: withSecretMetrics(createManySecret, { duration: "write", write: "create" }),
+    updateManySecret: withSecretMetrics(updateManySecret, { duration: "write", write: "update" }),
+    deleteManySecret: withSecretMetrics(deleteManySecret, { duration: "delete", write: "delete" }),
+    getSecretByName: withSecretMetrics(getSecretByName, { duration: "read" }),
+    getSecrets: withSecretMetrics(getSecrets, { duration: "read" }),
```
mathnogueira, line 3864: Instead of changing the code, I felt using a decorator would be better.

backend/src/db/migrations/20260709200205_dynamic_secret_leases_expire_at_index.ts
```diff
@@ -0,0 +1,25 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasIndex = await knex.schema.hasTable(TableName.DynamicSecretLease).then(async (exists) => {
+    if (!exists) return false;
```
akhilmhdh: Isn't this incorrect because - 
1. Table doesn't exist => hasIndex = false
2. if !hasIndex = true => it does alter table.

🤔 Can we just use the async await syntax than the then syntax. It feels hard to read as well.

backend/src/db/migrations/20260709200205_dynamic_secret_leases_expire_at_index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  if (!hasIndex) {
+    await knex.schema.alterTable(TableName.DynamicSecretLease, (t) => {
+      t.index("expireAt", "dynamic_secret_leases_expire_at_index");
+    });
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(TableName.DynamicSecretLease, (t) => {
```
akhilmhdh: Missing table check here

backend/src/lib/telemetry/metrics.ts
```diff
@@ -545,4 +600,58 @@ export const registerInfrastructureMetrics = (db: Knex) => {
...
+    description:
+      "Dynamic secret leases past expiry with status 'Failed to delete', grouped by provider. Alert on any value > 0 sustained 60m.",
+    unit: "{credential}"
+  });
+
+  orphanedGauge.addCallback(async (result) => {
+    if (!isTelemetryEnabled()) return;
+    try {
+      const rows: { type: string; count: string }[] = await db("dynamic_secret_leases")
+        .join("dynamic_secrets", "dynamic_secret_leases.dynamicSecretId", "dynamic_secrets.id")
```
akhilmhdh: I would recommend to use the db.ref().withSchema pattern and enums like other DAL calls we do

backend/src/lib/telemetry/metrics.ts
```diff
@@ -545,4 +600,58 @@ export const registerInfrastructureMetrics = (db: Knex) => {
...
+  const rotationStatusGauge = meter.createObservableGauge("infisical.secret_rotation.status", {
+    description:
+      "Current secret rotation counts by type and status. Alert on failure ratio > 50% per type with >= 5 total rotations.",
+    unit: "{rotation}"
+  });
+
+  rotationStatusGauge.addCallback(async (result) => {
+    if (!isTelemetryEnabled()) return;
+    try {
+      const rows: { type: string; rotationStatus: string; count: string }[] = await db("secret_rotations_v2")
```
akhilmhdh: Likewise here as well

backend/src/lib/telemetry/metrics.ts
akhilmhdh, line 597: I just checked and this function seems to be registered before the migration. So for a fresh instance this would break right? Because the migration didn't ran and the tables are not created.

I would recommend to move the secret management ones to seperate function like `registerSecretManagementMetrics` and keep it after the migration ran. Also i would recommend after this is registered register a set ex in redis to avoid doing this for all containers.

backend/src/lib/telemetry/metrics.ts
```diff
@@ -545,4 +600,58 @@ export const registerInfrastructureMetrics = (db: Knex) => {
     result.observe(pool.numFree?.() ?? 0, { "db.pool.state": "free" });
     result.observe(pool.numPendingAcquires?.() ?? 0, { "db.pool.state": "pending" });
   });
+
+  // Secret rotation status: GROUP BY type, rotationStatus. Backs "A rotation type is broken" alert.
+  const rotationStatusGauge = meter.createObservableGauge("infisical.secret_rotation.status", {
```
akhilmhdh: So i had a chat with @victorvhs017 on this now. There are two issues i am seeing here.
1. This is a massive table scan query and on top of that it's running each 60s
2. This is running for each container as well. 

So @victorvhs017 suggested we don't need to do this code wise and instead just wire it up by ourself in our internal system. I think it makes sense because these are things are easy to catch where otels shines is ones not easy to query.

akhilmhdh (changes requested):

akhilmhdh (commented):
One question: Don't you think dynamic secret lease outcome like success or failure must be recorded

akhilmhdh (approved):

PR #7229, opened 2026-07-09
docs(supabase-sync): describe behavior at Supabase's 100-secret cap

### LINKED ISSUE #7204: Docs: Supabase secret sync — behavior when hitting Supabase's hardcoded 100-secret limit is undocumented
Path: /integrations/secret-syncs/supabase (https://infisical.com/docs/integrations/secret-syncs/supabase)

Supabase enforces a hard limit of **100 secrets per project**, which is hardcoded on their side and cannot be raised. The Supabase secret sync docs don't describe what happens when this limit comes into play.

Could you clarify (and ideally document) the behavior for these cases:

1. **Syncing more than 100 secrets**: If the Infisical source path contains more than 100 secrets, what does the sync do? Fail outright, sync the first 100 and drop the rest, or something else?
2. **Target already at the cap**: If the Supabase project is already at 100 secrets (including secrets not managed by Infisical), how does the sync handle new additions — hard error, partial write, retry loop?
3. **Error surfacing**: Where does this failure show up (sync status in the UI, logs, webhook/alert)? Is it distinguishable from other sync errors?
4. **Overwrite vs. import modes**: Does the behavior differ depending on the configured sync mode (e.g. overwriting destination secrets vs. importing existing ones)?

We're running against this exact limit in production, so understanding the failure mode matters for us before relying on the sync. Thanks!

jaydeep-pipaliya:
Fixes #7204.

## What

Adds a Warning callout to the Supabase Sync guide that documents what happens when the source or destination reaches Supabase's hard cap of 100 secrets per project. Content is derived directly from \`backend/src/services/secret-sync/supabase/supabase-sync-fns.ts\` so the docs match the code.

Covers each of the four questions in the issue:

- **More than 100 source secrets:** Infisical batches writes in groups of 100 (\`chunkArray(toCreate, 100)\`). Batches that would push the destination over 100 fail at the Supabase API level; earlier batches in the same run may have already written successfully.
- **Destination already at 100:** The first create batch fails immediately with the same error path.
- **Where the failure surfaces:** \`syncStatus = failed\`, \`lastSyncMessage\` carries the Supabase error, and auto-sync retries on its normal schedule. The error names the first secret key in the failing batch, not every affected key.
- **Deletion mode interaction:** Because the create phase runs before the delete phase in \`syncSecrets\`, disabling \"Disable Secret Deletion\" does not free up capacity in the same sync run. If the destination is at cap and you want to add new secrets, either lower the source below 100 first or delete the unwanted destination secrets manually.

Also mentions that Supabase's reserved \`SUPABASE_*\` keys are skipped by the sync but still count against the 100-cap.

## Checklist

- [ ] I have added tests (docs-only)

### DIFF
--- a/docs/integrations/secret-syncs/supabase.mdx
+++ b/docs/integrations/secret-syncs/supabase.mdx
@@ -7,6 +7,17 @@ description: "Learn how to configure a Supabase Sync for Infisical."
 
 - Create a [Supabase Connection](/integrations/app-connections/supabase)
 
+<Warning>
+    Supabase enforces a hard cap of **100 secrets per project** that cannot be raised. When a sync would push the destination past that cap, the affected batch fails and the sync is reported as failed.
+
+    - **More than 100 source secrets:** Infisical batches writes to Supabase in groups of 100. Individual batches that would push the destination over 100 fail on the Supabase API call, and the sync status flips to failed. Earlier batches in the same run may have written successfully before the failing batch.
+    - **Destination already at 100:** The first create batch fails immediately with the same error path.
+    - **Where the failure surfaces:** The sync's `syncStatus` becomes `failed`, `lastSyncMessage` records the Supabase error, and auto-sync retries on its normal schedule. The error names the first secret key in the failing batch, not every affected key.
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Both right — the SUPABASE_INTERNAL_SECRETS guard only fires in the delete/remove loops, not on writes. Rewrote the reserved-keys note in b33e0c3 to describe the real behaviour (skipped on delete only) and to recommend renaming any colliding source secrets before enabling the sync.

PR #7232, opened 2026-07-09
fix: cert issuance validation

sheensantoscapadngan:
## Context
Opened up validation for cert issuance flow

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/certificate-router.ts
+++ b/backend/src/server/routes/v1/certificate-router.ts
@@ -37,7 +37,7 @@ type CertificateServiceResponse = TCertificateIssuanceResponse | Omit<TCertifica
 // Subject attributes are the requester's own certificate fields and flow into a structured
 // distinguished name (no string concatenation), so there's no injection risk. We don't constrain
 // their format here (wildcards like *.example.com are valid); only bound the length.
-const subjectAttributeField = z.string().trim().max(256);
+const subjectAttributeField = z.string().trim().max(255);
 
 const extractCertificateData = (
   data: CertificateServiceResponse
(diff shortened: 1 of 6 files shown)


backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -136,12 +140,12 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
...
-              organizationalUnit: validateTemplateRegexField.nullish(),
-              country: validateTemplateRegexField.nullish(),
-              state: validateTemplateRegexField.nullish(),
-              locality: validateTemplateRegexField.nullish(),
+              commonName: subjectAttributeField.nullish(),
+              organization: subjectAttributeField.nullish(),
+              organizationalUnit: subjectAttributeField.nullish(),
+              country: subjectAttributeField.nullish(),
+              state: subjectAttributeField.nullish(),
+              locality: subjectAttributeField.nullish(),
```
x032205, line 148: Claude pointed this out:

Loosening these to `max(256)` is fine for the internal CA (createDistinguishedName escapes via x509.Name), but issueCertificate/orderCertificate also route to external CAs that build the DN by unescaped concat (e.g. aws-pca: `CN=${commonName}` joined with ", " and passed as the string `name` to Pkcs10...create, which parses it as RFC-4514). So `example.com,O=EvilCorp` injects an extra O RDN. The old validator blocked commas, so this reintroduces DN injection. Better to escape at the external-CA builders (route them through createDistinguishedName) rather than loosen input.

lb-vn (dismissed):

carlosmonastyrski (approved):

PR #7235, opened 2026-07-09
fix: check connection id to define if sync is duplicated or not

mathnogueira:
## Context

Two syncs pointing to AWS SM in the same region, but different accounts would be flagged as duplicates. This includes the connection id as part of the duplication detection.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/app-connection/aws/aws-connection-fns.ts
+++ b/backend/src/services/app-connection/aws/aws-connection-fns.ts
@@ -228,63 +228,24 @@ export const validateAwsConnectionCredentials = async (appConnection: TAwsConnec
   return appConnection.credentials;
 };
 
-const extractAccountIdFromArn = (roleArn: string): string | null => {
-  const match = roleArn.match(/^arn:aws[^:]*:iam::(\d{12}):/);
-  return match ? match[1] : null;
-};
+export const getAwsAccountId = async (appConnection: TAwsConnectionConfig): Promise<string | null> => {
+  try {
+    const awsConfig = await getAwsConnectionConfig(appConnection);
 
-// eslint-disable-next-line no-bitwise
-const extractAccountIdFromAccessKey = (accessKeyId: string): string | null => {
-  if (accessKeyId.length < 16) return null;
+    const stsEndpoint =
+      appConnection.method === AwsConnectionMethod.AssumeRole ? appConnection.credentials.stsEndpoint : undefined;
+    const stsRegion = getStsSigningRegion(stsEndpoint);
 
-  const trimmed = accessKeyId.slice(4);
-  const base32Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567";
+    const sts = new STSClient({
+      region: stsRegion.region ?? awsConfig.region,
+      credentials: awsConfig.credentials,
+      ...(stsEndpoint && stsRegion.isValidEndpoint ? { endpoint: stsEndpoint } : {})
+    });
 
-  const bits: number[] = [];
-  for (const char of trimmed) {
-    const val = base32Chars.indexOf(char.toUpperCase());
-    if (val === -1) return null;
-    for (let i = 4; i >= 0; i -= 1) {
-      // eslint-disable-next-line no-bitwise
-      bits.push((val >> i) & 1);
-    }
+    const response = await sts.send(new GetCallerIdentityCommand({}));
+    return response.Account ?? null;
+  } catch (error) {
+    logger.error(error, "Failed to retrieve AWS account ID via STS");
+    return null;
   }
-
-  const bytes: number[] = [];
-  for (let i = 0; i < 6 * 8 && i < bits.length; i += 8) {
-    let byte = 0;
-    for (let j = 0; j < 8 && i + j < bits.length; j += 1) {
-      // eslint-disable-next-line no-bitwise
-      byte = (byte << 1) | bits[i + j];
-    }
-    bytes.push(byte);
-  }
-
-  if (bytes.length < 6) return null;
-
-  let num = 0n;
-  for (const byte of bytes) {
-    // eslint-disable-next-line no-bitwise
-    num = (num << 8n) | BigInt(byte);
-  }
-  const mask = 0x7fffffffff80n;
-  // eslint-disable-next-line no-bitwise
-  const accountId = ((num & mask) >> 7n).toString();
-
-  return accountId.padStart(12, "0");
-};
-
-export const extractAwsAccountId = (connection: {
-  method: string;
-  credentials: { roleArn?: string; accessKeyId?: string };
-}): string | null => {
-  if (connection.method === AwsConnectionMethod.AssumeRole && connection.credentials.roleArn) {
-    return extractAccountIdFromArn(connection.credentials.roleArn);
-  }
-
-  if (connection.method === AwsConnectionMethod.AccessKey && connection.credentials.accessKeyId) {
-    return extractAccountIdFromAccessKey(connection.credentials.accessKeyId);
-  }
-
-  return null;
 };
--- a/backend/src/services/app-connection/aws/aws-connection-fns.ts
+++ b/backend/src/services/app-connection/aws/aws-connection-fns.ts
@@ -227,3 +227,64 @@ export const validateAwsConnectionCredentials = async (appConnection: TAwsConnec
 
   return appConnection.credentials;
 };
+
+const extractAccountIdFromArn = (roleArn: string): string | null => {
+  const match = roleArn.match(/^arn:aws[^:]*:iam::(\d{12}):/);
+  return match ? match[1] : null;
+};
+
+// eslint-disable-next-line no-bitwise
+const extractAccountIdFromAccessKey = (accessKeyId: string): string | null => {
+  if (accessKeyId.length < 16) return null;
+
+  const trimmed = accessKeyId.slice(4);
+  const base32Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ234567";
+
+  const bits: number[] = [];
+  for (const char of trimmed) {
+    const val = base32Chars.indexOf(char.toUpperCase());
+    if (val === -1) return null;
+    for (let i = 4; i >= 0; i -= 1) {
+      // eslint-disable-next-line no-bitwise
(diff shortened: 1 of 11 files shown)


backend/src/services/secret-sync/secret-sync-maps.ts
```diff
@@ -215,7 +215,13 @@ const defaultDuplicateCheck: DestinationDuplicateCheckFn = () => true;
 
 export const DESTINATION_DUPLICATE_CHECK_MAP: Record<SecretSync, DestinationDuplicateCheckFn> = {
   [SecretSync.AWSParameterStore]: defaultDuplicateCheck,
-  [SecretSync.AWSSecretsManager]: defaultDuplicateCheck,
+  [SecretSync.AWSSecretsManager]: (_existingConfig, _newConfig, existingConnectionId, newConnectionId) => {
+    // TODO: we should also check the AWS account id here, otherwise
+    // two connections to the same account would cause issues.
+    // If assume role: get account id from ARN
+    // If credentials, use AWS_ACCESS_ID to infer the account id
+    return existingConnectionId === newConnectionId;
```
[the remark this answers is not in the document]
mathnogueira: This is documented in the TODO, will be done later.

mathnogueira:
@greptile review again

backend/src/services/app-connection/aws/aws-connection-fns.ts
```diff
@@ -227,3 +227,64 @@ export const validateAwsConnectionCredentials = async (appConnection: TAwsConnec
...
+  for (const byte of bytes) {
+    // eslint-disable-next-line no-bitwise
+    num = (num << 8n) | BigInt(byte);
+  }
+  const mask = 0x7fffffffff80n;
+  // eslint-disable-next-line no-bitwise
+  const accountId = ((num & mask) >> 7n).toString();
+
+  return accountId.padStart(12, "0");
+};
```
adilsitos: Instead of computing the accountId locally, maybe we could call AWS using sts with a `GetCallerIdentity` and get the account, something like this: 


```suggestion
export const getAwsConnectionAccountId = async (connection: TAwsConnection): Promise<string | null> => {
  try {
    const awsConfig = await getAwsConnectionConfig(buildAwsConnectionConfig(connection, connection.credentials));

    const stsEndpoint =
      connection.method === AwsConnectionMethod.AssumeRole ? connection.credentials.stsEndpoint : undefined;
    const stsRegion = getStsSigningRegion(stsEndpoint);

    const sts = new STSClient({
      region: stsRegion.region ?? awsConfig.region,
      sha256: CustomAWSHasher,
      credentials: awsConfig.credentials,
      ...(stsEndpoint && stsRegion.isValidEndpoint
        ? { endpoint: stsEndpoint }
        : { useFipsEndpoint: crypto.isFipsModeEnabled() })
    });

    const identity = await sts.send(new GetCallerIdentityCommand({}));

    return identity.Account ?? null;
  } catch (error) {
    logger.error(error, `Failed to resolve AWS account ID for connection [connectionId=${connection.id}]`);
    return null;
  }
};

```

adilsitos: maybe we can use this approach to both assume role and access key

backend/src/services/secret-sync/secret-sync-maps.ts
```diff
@@ -235,7 +258,7 @@ export const DESTINATION_DUPLICATE_CHECK_MAP: Record<SecretSync, DestinationDupl
   [SecretSync.Render]: defaultDuplicateCheck,
   [SecretSync.Flyio]: defaultDuplicateCheck,
   [SecretSync.TriggerDev]: defaultDuplicateCheck,
-  [SecretSync.GitLab]: (existingConfig, newConfig) => {
+  [SecretSync.GitLab]: async ({ existingConfig, newConfig }) => {
```
adilsitos: was this lint? is there a reason we also changed the gitlab sync to be async?

frontend/src/hooks/api/secretSyncs/useDuplicateDestinationCheck.ts
adilsitos, line 1: ignore comment placement. 

I noticed that the duplicate check only happens when the destination has a path, but several other syncs only use a project name and don't have the concept of a path, which allow the user to generate a sync even if their org does not allow it.

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -280,29 +282,48 @@ export const secretSyncServiceFactory = ({
...
+          throw new NotFoundError({ message: `App connection with ID "${connId}" not found` });
         }
+        const credentials = await decryptAppConnectionCredentials({
+          orgId: conn.orgId,
+          encryptedCredentials: conn.encryptedCredentials,
+          kmsService,
+          projectId: conn.projectId
+        });
+        return { ...conn, credentials } as TAppConnection;
+      };
```
adilsitos: I think we might need to check the permission of the user on the app connections we will be checking. if the app connection is from the project, we already check it, but if the app connection that the sync is using if from the org, we should check it the permission. 

I think we can decrypt and validate the destination, but we should not return the projectID in case they don't have access. 

Image

adilsitos (changes requested):

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -280,29 +282,48 @@ export const secretSyncServiceFactory = ({
     try {
       const existingSyncs = await secretSyncDAL.findByDestinationAndOrgId(destination, actor.orgId);
 
-      const duplicates = existingSyncs.filter((sync) => {
-        if (sync.id === excludeSyncId) {
-          return false;
+      const decryptConnection = async (connId: string): Promise<TAppConnection> => {
```
adilsitos: I think we can memoize this as well, since some syncs can share the same connId. 


```suggestion
     const connectionCache = new Map<string, Promise<TAppConnection>>();
      const decryptConnection = (connId: string): Promise<TAppConnection> => {
        const cached = connectionCache.get(connId);
        if (cached) return cached;

        const connectionPromise = (async () => {
          const conn = await appConnectionDAL.findById(connId);
          if (!conn || conn.orgId !== actor.orgId) {
            throw new NotFoundError({ message: `App connection with ID "${connId}" not found` });
          }
          const credentials = await decryptAppConnectionCredentials({
            orgId: conn.orgId,
            encryptedCredentials: conn.encryptedCredentials,
            kmsService,
            projectId: conn.projectId
          });
          return { ...conn, credentials } as TAppConnection;
        })();

        connectionCache.set(connId, connectionPromise);
        return connectionPromise;
      };
```

backend/src/services/secret-sync/secret-sync-maps.ts
mathnogueira: DestinationDuplicateCheckFn is async now (to allow us to get the app connection info), so this had to become async too to match the function type.

backend/src/services/secret-sync/secret-sync-maps.ts
```diff
@@ -211,11 +213,40 @@ export const SECRET_SYNC_SKIP_FIELDS_MAP: Record<SecretSync, string[]> = {
...
+
+  const existingAwsConn = existingConn as TAwsConnection;
+  const newAwsConn = newConn as TAwsConnection;
+
+  const [existingAccountId, newAccountId] = await Promise.all([
+    getAwsAccountId(buildAwsConnectionConfig(existingAwsConn, existingAwsConn.credentials)),
+    getAwsAccountId(buildAwsConnectionConfig(newAwsConn, newAwsConn.credentials))
+  ]);
+
+  if (!existingAccountId || !newAccountId) return false;
```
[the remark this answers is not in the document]
mathnogueira: This is true, but changing this to `true` would prevent users from creating a new sync with a new credential in case the other sync is broken due to an expired credential in the AppConnection, even if that expired connection is not used by a conflicting sync.

The UX could be worse and the duplicate error would actually ofuscate the real problem. Will not apply this.

backend/src/services/secret-sync/secret-sync-service.ts
mathnogueira: CleanShot 2026-07-10 at 14 38 42

CleanShot 2026-07-10 at 14 39 47@2x

backend/src/services/secret-sync/secret-sync-maps.ts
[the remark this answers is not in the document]
mathnogueira: Plus, this is not a huge risk as this is a legit user with access to the AWS account.

frontend/src/hooks/api/secretSyncs/useDuplicateDestinationCheck.ts
adilsitos: I will create a ticket for this and attach a video.

backend/src/services/secret-sync/secret-sync-maps.ts
```diff
@@ -211,11 +213,44 @@ export const SECRET_SYNC_SKIP_FIELDS_MAP: Record<SecretSync, string[]> = {
...
+const defaultDuplicateCheck: DestinationDuplicateCheckFn = async () => true;
+const awsDuplicateCheck: DestinationDuplicateCheckFn = async ({ existingSync, newSync, decryptConnection }) => {
+  if (!newSync.connectionId) return true;
+
+  const existingKeySchema = (existingSync.syncOptions?.keySchema as string) ?? "";
+  const newKeySchema = (newSync.syncOptions?.keySchema as string) ?? "";
+
+  // Mismatched keySchemas can still conflict (e.g. sync A with keySchema "INFISICAL_{{secretKey}}" and secret HOST
+  // vs sync B without keySchema and secret INFISICAL_HOST will write to the same destination key).
+  const hasKeySchemaConflict = !existingKeySchema || !newKeySchema || existingKeySchema === newKeySchema;
```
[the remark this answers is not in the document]
mathnogueira: This case is covered and it doesn't trigger a duplication. Only if A or B has an empty keySchema and the other doesn't. 

Will not work on it.

adilsitos (approved):

PR #7250, opened 2026-07-10
feat(roles): add personal override action on secrets resource

adilsitos:
## Context

This adds a new action `personal-override` which allow users that have the `describe` and the new action to create personal overrides without having access to the actual secret

### Scenario 1: user is a member (has create, delete, edit and describe) on secrets can still add overrides. 
- Create a user which is a member 
- make sure that he can still add overrides 

### Scenario 2: user has `describe` and `personal-override` action 
- create a user that has only `describe` and `personal-override`
- check that he can create personal override
- validate that they don't have access to the shared secrets 

### scenario 3: user has only `describe`
- create a user that has only `describe`
- validate that they cannot add a personal override

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/permission/permission-fns.test.ts
+++ b/backend/src/ee/services/permission/permission-fns.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, test } from "vitest";
 
 import { conditionsMatcher } from "@app/lib/casl";
 
-import { expandLegacyForbidActions } from "./permission-fns";
+import { expandLegacyForbidActions, throwIfMissingSecretPersonalOverridePermission } from "./permission-fns";
 import {
   ProjectPermissionActions,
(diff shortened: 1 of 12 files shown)


varonix0 (changes requested):
We still allow editing and deleting personal overrides, even if the user holding the role only has access `Describe Secret` as an example. looks like we're correctly blocking editing a personal override on the frontend, but on the backend we still allow it.

Deleting personal overrides is allowed no matter what, even if the user doesn't have the Delete permission, or the personal overrides permission

adilsitos:
@greptile review it again and update the summary

varonix0 (approved):

PR #7252, opened 2026-07-10
feat(code-signing): prevent signers from approving their own signing requests

bernie-g:
## Context

Enforces that a code-signing requester can never approve their own signing request, and that an admin cannot pre-approve signing for themselves. Both are unconditional (no policy setting), so self-approval is blocked by design rather than by approval-policy configuration — the same model as PAM access requests.

## Steps to verify the change

1. On a signer with an approval policy, add yourself as an eligible approver, submit a signing request, and confirm you cannot approve it (API returns 403; the UI hides the approve/reject action on your own request). Requesting is still allowed and being listed as an approver is expected — you just can't approve your own request.
2. As an admin, confirm you are not offered as a grantee in the Pre-approve modal, and that pre-approving yourself via the API is rejected.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/signer/signer-policy-service.ts
+++ b/backend/src/services/signer/signer-policy-service.ts
@@ -8,7 +8,7 @@ import {
   ResourcePermissionSignerActions,
   ResourcePermissionSub
 } from "@app/ee/services/permission/resource-permission";
-import { BadRequestError, ForbiddenRequestError, NotFoundError } from "@app/lib/errors";
+import { BadRequestError, NotFoundError } from "@app/lib/errors";
 import { ms } from "@app/lib/ms";
 
 import {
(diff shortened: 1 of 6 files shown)


backend/src/services/signer/signer-policy-service.ts
```diff
@@ -620,6 +620,12 @@ export const signerPolicyServiceFactory = ({
     if (!dto.granteeUserId && !dto.granteeIdentityId) {
       throw new BadRequestError({ message: "granteeUserId or granteeIdentityId is required." });
     }
+    if (
```
carlosmonastyrski: Pre-approve signing access is a valid scenario

carlosmonastyrski (approved):

PR #7259, opened 2026-07-11
fix(membership-identity): guard identityDAL.findById result before reading fields

### LINKED ISSUE #7255: PATCH identity-membership role update 500s — TypeError reading 'projectId' in updateMembership (v0.162.0)
## Description
On self-hosted **v0.162.0**, updating a **machine identity's** project membership role via the API returns **500 Internal Server Error**. The backend throws a `TypeError` before the role change is applied.

## Steps to reproduce
1. Create a machine identity (Universal Auth).
2. Add it to a project as `member`: `POST /api/v1/projects/{projectId}/identity-memberships/{identityId}` with `{ "roles": [{ "role": "member" }] }` — succeeds (200).
3. Change its role to read-only: `PATCH /api/v1/projects/{projectId}/identity-memberships/{identityId}` with `{ "roles": [{ "role": "viewer" }] }`.

## Expected
Membership role updated to `viewer` (200).

## Actual
`500 { "statusCode": 500, "message": "Something went wrong", "error": "InternalServerError" }`

Backend log:
```
TypeError: Cannot read properties of undefined (reading 'projectId')
    at ... updateMembership ... :196
    at ... handler ... :163
```

## Also tried (all 500)
- Endpoints: `/api/v1/projects/{id}/identity-memberships/{id}`, `/api/v1/projects/{id}/memberships/identities/{id}`, `/api/v2/workspace/{id}/identity-memberships/{id}`
- Bodies: `{roles:[{role}]}`, `{roles:[{role,isTemporary:false}]}`, plus adding `projectId` / `workspaceId` / `identityId` to the body.
- `POST` re-create with `role: viewer` returns `400 "Identity ... is missing organization"` (separate guard).

## Impact
A machine identity's project role cannot be changed after creation via the API — it's stuck at the role assigned at membership-creation time. Blocks moving agents to least-privilege (read-only) roles.

## Environment
- Infisical self-hosted **v0.162.0** (Docker), Postgres 14, machine-identity (Universal Auth) caller with org-admin.

jaydeep-pipaliya:
Fixes #7255.

## What

Five sites across the project and org membership-identity factories fetch \`identityDAL.findById(dto.selector.identityId)\` and then immediately read \`identityDetails.projectId\` or \`identityDetails.orgId\` without checking whether the lookup returned a row. If it returns \`undefined\` for any reason (identity gone, stale id, cache race), the request 500s with:

\`\`\`
TypeError: Cannot read properties of undefined (reading 'projectId')
\`\`\`

which is what the reporter of #7255 hit trying to PATCH a machine identity's project role on v0.162.0.

## Fix

Adds a \`NotFoundError\` guard on all five sites:

- \`project-membership-identity-factory.onCreateMembershipIdentityGuard\`
- \`project-membership-identity-factory.onUpdateMembershipIdentityGuard\`
- \`project-membership-identity-factory.onDeleteMembershipIdentityGuard\`
- \`org-membership-identity-factory.onUpdateMembershipIdentityGuard\`
- \`org-membership-identity-factory.onDeleteMembershipIdentityGuard\`

The 500 now becomes a clean 404 that names the missing identity id, so operators can see what went wrong instead of getting a generic \`InternalServerError\`.

The \`membership-identity-service\` layer above still ensures a valid membership exists before running these guards, so the NotFoundError is a defense-in-depth net rather than a routine path — but it converts a hard crash into a diagnosable error either way.

## Checklist

- [ ] I have added tests (defensive guard; hard to reproduce the exact identityDetails-undefined state without a test double)

### DIFF
--- a/backend/src/services/membership-identity/org/org-membership-identity-factory.ts
+++ b/backend/src/services/membership-identity/org/org-membership-identity-factory.ts
@@ -7,7 +7,7 @@ import {
   validatePrivilegeChangeOperation
 } from "@app/ee/services/permission/permission-fns";
 import { TPermissionServiceFactory } from "@app/ee/services/permission/permission-service-types";
-import { BadRequestError, InternalServerError, PermissionBoundaryError } from "@app/lib/errors";
+import { BadRequestError, InternalServerError, NotFoundError, PermissionBoundaryError } from "@app/lib/errors";
 import { requestMemoKeys } from "@app/lib/request-context/memo-keys";
 import { requestMemoize } from "@app/lib/request-context/request-memoizer";
 import { TIdentityDALFactory } from "@app/services/identity/identity-dal";
@@ -147,6 +147,9 @@ export const newOrgMembershipIdentityFactory = ({
     const identityDetails = await requestMemoize(requestMemoKeys.identityFindById(dto.selector.identityId), () =>
       identityDAL.findById(dto.selector.identityId)
     );
+    if (!identityDetails) {
+      throw new NotFoundError({ message: `Identity with ID '${dto.selector.identityId}' not found` });
+    }
     if (identityDetails.projectId) {
       throw new BadRequestError({ message: "Failed to create organization membership for a project scoped identity" });
     }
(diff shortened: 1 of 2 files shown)


jaydeep-pipaliya:
Greptile P1 right — I patched five sites but missed onCreateMembershipIdentityGuard in the org factory, which reads identityDetails.orgId on the very next line and throws the same TypeError. Pushed da845ba adding the same NotFoundError guard so the create-org-membership path is covered too.

PR #7261, opened 2026-07-11
fix(secret-approval-request): guard user/env/project lookups in notification path

jaydeep-pipaliya:
## What

Both \`generateSecretApprovalRequest\` and \`generateSecretApprovalRequestV2Bridge\` run a Slack / Microsoft Teams workflow-notification block after the approval request is committed to the DB. They fetch \`user\`, \`env\`, and \`project\` via \`requestMemoize\` + DAL.findById and then read \`user.email\`, \`env.name\`, \`project.name\` without any null check. If any of the three come back \`undefined\` (deleted user between commit and notify, a slug that doesn't resolve, project gone), the notification block throws:

\`\`\`
TypeError: Cannot read properties of undefined (reading 'email' / 'name')
\`\`\`

and the caller sees a **500** for a request that has **already committed** — leaving the approval request stuck without an ack path.

## Fix

Early-return guards on both notification blocks. If \`user\`, \`env\`, or \`project\` is undefined, we return the already-committed \`secretApprovalRequest\` (preserving each function's declared return shape) and simply skip the workflow notification. The DB write is not affected.

Same defensive shape as the \`identityDetails\` guard fixes in #7259 (membership-identity factories) and #7260 (identity-service). Read-then-use lookups need a null guard.

## Checklist

- [ ] I have added tests (defensive guard on the notification-side branch)

### DIFF
--- a/backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
+++ b/backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
@@ -1679,9 +1679,6 @@ export const secretApprovalRequestServiceFactory = ({
 
     const env = await projectEnvDAL.findOne({ slug: environment, projectId });
     const user = await requestMemoize(requestMemoKeys.userFindById(actorId), () => userDAL.findById(actorId));
-    if (!user || !env) {
-      return secretApprovalRequest;
-    }
 
     const projectPath = `/organizations/${actorOrgId}/projects/secret-management/${projectId}`;
     const approvalPath = `${projectPath}/approval`;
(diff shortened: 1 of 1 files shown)


jaydeep-pipaliya:
Both right — the early-return skipped `sendApprovalEmailsFn` (which drives the approver emails + in-app SECRET_CHANGE_REQUEST notifications) and the PostHog telemetry too, leaving the approval request committed but with no one pinged.

Pushed 3835a9e that narrows the guard: only the `triggerWorkflowIntegrationNotification` block is now wrapped in `if (user && env [&& project])`. `sendApprovalEmailsFn` runs unconditionally (it fetches its own project internally and doesn't touch the three lookups), and telemetry uses `user?.username ?? user?.email ?? actorId` so the distinctId falls back cleanly. Approvers get pinged even when a lookup misses; only the Slack/Teams workflow notification is skipped.

PR #7279, opened 2026-07-14
feat: allow multiple projects to be selected in project grants

mathnogueira:
## Context

We got some feedback that it would be useful if it would be possible to select multiple envs and projects at once instead of one by one. 

Important to notice that this will ignore duplicate grants (will not add them again), but this will not error out in the UI. The main reason is because we error out in that scenario it becomes annoying to select multiple envs/paths and multiple projects, especially if a grant already exists for any of the env/path/project combinations. 

So it makes UX better to ignore those errors and make them no-op instead.

## Screenshots

CleanShot 2026-07-14 at 15 00 47

CleanShot 2026-07-14 at 21 10 42

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/project-folder-grant/project-folder-grant-service.ts
+++ b/backend/src/services/project-folder-grant/project-folder-grant-service.ts
@@ -175,7 +175,8 @@ export const projectFolderGrantServiceFactory = ({
     const grants = await projectFolderGrantDAL.listBySourceProject(sourceProjectId);
 
     const folderIds = grants.map((g) => g.sourceFolderId);
-    const folderPaths = folderIds.length > 0 ? await folderDAL.findSecretPathByFolderIds(sourceProjectId, folderIds) : [];
+    const folderPaths =
+      folderIds.length > 0 ? await folderDAL.findSecretPathByFolderIds(sourceProjectId, folderIds) : [];
 
     return grants.map((g, i) => ({
       ...g,
--- a/backend/src/services/project-folder-grant/project-folder-grant-service.ts
+++ b/backend/src/services/project-folder-grant/project-folder-grant-service.ts
@@ -172,7 +172,15 @@ export const projectFolderGrantServiceFactory = ({
       ProjectPermissionSub.ProjectFolderGrant
     );
 
-    return projectFolderGrantDAL.listBySourceProject(sourceProjectId);
+    const grants = await projectFolderGrantDAL.listBySourceProject(sourceProjectId);
+
+    const folderIds = grants.map((g) => g.sourceFolderId);
+    const folderPaths = folderIds.length > 0 ? await folderDAL.findSecretPathByFolderIds(sourceProjectId, folderIds) : [];
+
+    return grants.map((g, i) => ({
+      ...g,
+      secretPath: folderPaths[i]?.path ?? "/"
+    }));
   };
 
   const listGrantsForTargetProject = async ({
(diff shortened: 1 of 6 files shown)


mathnogueira:
@greptile review again

mathnogueira:
@greptile review once more

backend/src/services/project-folder-grant/project-folder-grant-service.ts
```diff
@@ -172,7 +172,16 @@ export const projectFolderGrantServiceFactory = ({
...
-    return projectFolderGrantDAL.listBySourceProject(sourceProjectId);
+    const grants = await projectFolderGrantDAL.listBySourceProject(sourceProjectId);
+
+    const folderIds = grants.map((g) => g.sourceFolderId);
+    const folderPaths =
+      folderIds.length > 0 ? await folderDAL.findSecretPathByFolderIds(sourceProjectId, folderIds) : [];
+
+    return grants.map((g, i) => ({
+      ...g,
+      secretPath: folderPaths[i]?.path ?? "/"
```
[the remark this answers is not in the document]
adilsitos: I think it is good to have it, even if they cannot see the folder. This way we make sure that we don't have a case where someone finds a vulnerability in another place that provides access to several paths and they can use it here.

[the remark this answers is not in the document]
mathnogueira: Well, this makes sense. We could censor it somehow, but I don't think it would be very useful to see a grant where you cannot see which folder it's granting access to.

frontend/src/pages/project/SettingsPage/components/CrossProjectSharingSection/ShareSecretsSheet.tsx
adilsitos, line 1: I saw we are deduping the request in case we add the same environment + folder + destination, but this is possible on the UI. can't we also dedupe that there? 

Image

frontend/src/pages/project/SettingsPage/components/CrossProjectSharingSection/CrossProjectSharingSection.tsx
adilsitos, line 1: question: I think this was discussed before, but just asking it here as well. In scenarios that users have a lot of environments this might be overwhealming, maybe an approach here would be to just show the environments that have the permission. For me it is fine both ways, but maybe it is worth checking if this is fine (for long names it is working good)

Image

frontend/src/pages/project/SettingsPage/components/CrossProjectSharingSection/ShareSecretsSheet.tsx
```diff
@@ -26,129 +25,432 @@ import {
...
+      }
+      throw err;
+    }
+  };
+
+  const removeGrant = async (dto: TDeleteProjectFolderGrantDTO) => {
+    await apiRequest.delete(`/api/v1/project-folder-grants/${dto.grantId}`, {
+      params: { sourceProjectId: dto.sourceProjectId }
+    });
+  };
```
adilsitos: is there a reason we changed this to direct api calls here instead of using the hook useDeleteProjectFolderGrant? The hooks was also invalidating some queries, maybe this would be necessary as well?

frontend/src/pages/project/SettingsPage/components/CrossProjectSharingSection/ShareSecretsSheet.tsx
```diff
@@ -26,129 +25,432 @@ import {
...
+      if (next.length === 0) return [{ environment: "", secretPaths: [] }];
+      return next;
+    });
+  };
+
+  const handleAddGroup = () => {
+    setGroups((prev) => [...prev, { environment: "", secretPaths: [] }]);
+  };
+
+  const postGrant = async (dto: TCreateProjectFolderGrantDTO) => {
```
adilsitos: is there a reason we changed this to direct api calls here instead of using the hook useCreateProjectFolderGrant? The hooks was also invalidating some queries, maybe this would be necessary as well?

adilsitos (commented):
just some questions and QOL

frontend/src/pages/project/SettingsPage/components/CrossProjectSharingSection/CrossProjectSharingSection.tsx
mathnogueira: Makes sense, seems that some projects have more than 15 envs

adilsitos (approved):

PR #7287, opened 2026-07-14
feat: add domainComponents to PKI cert fields

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -146,6 +148,7 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
               country: subjectAttributeField.nullish(),
               state: subjectAttributeField.nullish(),
               locality: subjectAttributeField.nullish(),
+              domainComponents: z.array(domainComponentSchema).optional(),
```
sheensantoscapadngan: why is this optional but the above properties are nullish?

backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -367,6 +369,10 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
           certificateRequestForService[field] = attributes[field] ?? undefined;
         }
       }
+      // Domain components are multi-valued; only include when the request explicitly provides a non-empty list.
+      if (attributes?.domainComponents && attributes.domainComponents.length > 0) {
+        certificateRequestForService.domainComponents = attributes.domainComponents;
```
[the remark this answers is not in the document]
sheensantoscapadngan: up !

backend/src/services/certificate-common/certificate-csr-utils.ts
```diff
@@ -100,6 +106,36 @@ export const extractCertificateRequestFromCSR = (csr: string): TCertificateReque
...
+  if (!injectsDefault) return undefined;
+
+  return createDistinguishedName({
+    commonName: csrSubject.commonName ?? request.commonName,
+    organization: csrSubject.organization ?? request.organization,
+    ou: csrSubject.ou ?? request.organizationalUnit,
+    country: csrSubject.country ?? request.country,
+    province: csrSubject.province ?? request.state,
+    locality: csrSubject.locality ?? request.locality,
+    domainComponents: csrSubject.domainComponents
```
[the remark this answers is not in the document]
sheensantoscapadngan: why is the handling of this not consistent with the other properties?

backend/src/services/certificate-common/certificate-csr-utils.ts
```diff
@@ -100,6 +106,26 @@ export const extractCertificateRequestFromCSR = (csr: string): TCertificateReque
...
+  const csrSubject = extractDnParts(new x509.Pkcs10CertificateRequest(csr).subjectName);
+
+  return createDistinguishedName({
+    commonName: csrSubject.commonName ?? request.commonName,
+    organization: csrSubject.organization ?? request.organization,
+    ou: csrSubject.ou ?? request.organizationalUnit,
+    country: csrSubject.country ?? request.country,
+    province: csrSubject.province ?? request.state,
+    locality: csrSubject.locality ?? request.locality,
+    domainComponents: csrSubject.domainComponents
```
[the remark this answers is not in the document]
sheensantoscapadngan: valid ?

docs/documentation/platform/pki/settings/policies.mdx
```diff
@@ -74,6 +75,15 @@ For each attribute, configure enforcement:
   </Accordion>
 </AccordionGroup>
 
+<Note>
+  Domain Component (DC) is matched independently from the Common Name. A subject can carry more than
+  one domain component (for example `CN=host, DC=auth, DC=lucidgov, DC=app`), and each value is
+  checked against the Domain Component rule individually, similar to Subject Alternative Names. To
+  scope certificates to a domain, add a Common Name rule for the hostname and a separate Domain
+  Component rule for the domain labels (for the example above, allow `auth`, `lucidgov`, and `app`,
+  or a pattern such as `*`).
```
sheensantoscapadngan: I am unsure if we should frame it this way because technically the use of DC is not so common anymore, since people combine that into the common name. This might give the impression that people should be using DC for "google.com". We're really just accommodating the folks that still use DC but not necessarily ask them to move to DC

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -1116,6 +1117,7 @@ export const certificateV3ServiceFactory = ({
         const processResult = await processSelfSignedCertificate({
           certificateRequest: {
             ...certificateRequest,
+            domainComponents: certificateRequestWithDefaults.domainComponents,
             validity: { ttl: resolvedTtl }
           },
```
sheensantoscapadngan, line 1129: why is that only the `domainComponent` field is being passed from `certificateRequestWithDefaults`? is this a bug

backend/src/services/certificate-common/certificate-issuance-utils.ts
```diff
@@ -375,7 +376,8 @@ export const generateSelfSignedCertificate = async ({
     ou: certificateSubject.organizational_unit,
     country: certificateSubject.country,
     province: certificateSubject.state_or_province_name,
-    locality: certificateSubject.locality_name
+    locality: certificateSubject.locality_name,
+    domainComponents: certificateRequest.domainComponents
```
sheensantoscapadngan, line 385: also, why is it that `domainComponent` is the only one from `certificateRequest` whereas everything else came from `certificateSubject`

sheensantoscapadngan (approved):

backend/src/services/certificate-authority/certificate-authority-fns.test.ts
```diff
@@ -237,3 +241,31 @@ describe("buildCrlDistributionPointUrls", () => {
...
+    const dn = createDistinguishedName({ commonName: "example.com", organization: "Acme" });
+    expect(dn).toBe("O=Acme, CN=example.com");
+  });
+
+  it("should emit one DC RDN per domain component, in the given order, after the CN", () => {
+    const dn = createDistinguishedName({
+      commonName: "auth-AD-MANAGER02-CA",
+      domainComponents: ["app", "example", "auth"]
+    });
+    expect(dn).toBe("CN=auth-AD-MANAGER02-CA, DC=app, DC=example, DC=auth");
```
tmccombs, line 256: I'm pretty sure this should be "DC=app, DC=example, DC=auth, CN=auth-AD-MANAGER02-CA".

The order of the DN should be from most general to most specific.

backend/src/services/certificate-authority/certificate-authority-fns.ts
tmccombs, line 1: I think that createDistinguishedName needs to be updated to something like:
```ts
export const createDistinguishedName = (parts: TDNParts) => {
  // Build JSON array for x509.Name - the library handles all RFC 4514 escaping
  const jsonName: Array<{ [type: string]: string[] }> = [];
  if (parts.country) jsonName.push({ C: [parts.country] });
  if (parts.organization) jsonName.push({ O: [parts.organization] });
  if (parts.ou) jsonName.push({ OU: [parts.ou] });
  if (parts.province) jsonName.push({ ST: [parts.province] });
  // DC is multi-valued and ordered; emit one RDN per value in the given order (after CN, AD-style).
  if (parts.domainComponents) {
    for (const dc of parts.domainComponents) {
      if (dc) jsonName.push({ DC: [dc] });
    }
  }
  if (parts.commonName) jsonName.push({ CN: [parts.commonName] });
  if (parts.locality) jsonName.push({ L: [parts.locality] });

  // Create Name object from JSON and convert to properly escaped string
  const name = new x509.Name(jsonName);
  return name.toString();
};
```
Note that I moved inserting the DC before inserting the CN.

Because the DC entries should come before the CN entry. (Also, I think that locality should come before commonName, and probably domainComponents as well, but I'm not sure).

backend/src/services/certificate-policy/certificate-policy-service.ts
```diff
@@ -394,6 +397,56 @@ export const certificatePolicyServiceFactory = ({
...
 
+    const domainComponentPolicy = subjectPolicies?.find(
+      (attrPolicy) => attrPolicy.type === CertSubjectAttributeType.DOMAIN_COMPONENT
+    );
+    const requestDomainComponents = request.domainComponents ?? [];
+    const dcFieldName = CertSubjectAttributeType.DOMAIN_COMPONENT;
+    if (domainComponentPolicy) {
+      if (!options?.skipRequired && domainComponentPolicy.required && domainComponentPolicy.required.length > 0) {
+        for (const requiredValue of domainComponentPolicy.required) {
+          const hasMatch = requestDomainComponents.some(
```
tmccombs, line 408: This is probably better than nothing, but ideally, the constraint would take the order into account as well. 

This currently only compares each component to an list of allowed, required, or denied values. 

This means that if you have something like `required = ["com", "example"]`, then that would allow both "DC=example, DC= com" (example.com) and "DC=com,DC=example" (com.example), which is probably not what you want. 

I think most commonly you would want to either:
- Require that the list of components *exactly* matches a list of components: say, exactly "DC=com,DC=example"
- Require that the list of components starts with a prefix, so that "DC=com,DC=example,DC=a,DC=b" would match the prefix `["com", "example"]`
- Require that the list of components matches a list with wildcards, like `["com", "example", "*"]`

I can think of two ways of expressing that:

1. Make the required/allowed/deny lists lists of lists, so for example:
    ```terraform
    required = [
      # exact match
      ["com", "example", "www"],
      # wildcard for single component
      ["app", "example", "*"],
      # wildcard for multiple components, could use something else, but probably need a specific value
      ["internal", "**"],
    ]
    ```
2. add a new "domain" synthetic constraint (that conflicts with commonName and domainComponent) that is a combination of commonName, and domainComponent, and is parsed into it's componets, so that something like `**.example.com` would match anything that had "DC=com,DC=example" followed by any number of DC and CN components, And `*.example.com` would match `DC=com,DC=example` followed by any CN (and no more DC components), `www.example.com` would match exactly "DC=com,DC=example,CN=www`, `*-a.example.com` would match the exact disinguishedNames fo `["com", "example"]` and a common name that matches the wildcard pattern "*-a", etc.

backend/src/services/certificate-authority/certificate-authority-fns.ts
tmccombs: Maybe.. The comment above says "RFC 4514 Distinguished Name string", but [RFC 4514 section 2.1](https://datatracker.ietf.org/doc/html/rfc4514#section-2.1) says that the string order should be " starting with the last element of the sequence and moving  backwards toward the first" compared to the binary representation, which means it would be most specific to most general, and so it would look like `CN=auth-AD-MANAGER02-CA,DC=auth,DC=example,DC=app`. But maybe the places this string is used the order doesn't matter between different types of attributes? Although for the values within DC, it certainly would.

PR #7301, opened 2026-07-15
fix(webauthn): support hardware security keys on FIPS instances

victorvhs017:
## Context

Hardware security keys (e.g. YubiKey) could be registered as passkeys but then fail at MFA sign-in on FIPS-enabled instances. `@simplewebauthn/server` defaults to offering EdDSA (`-8`) first; Ed25519-capable keys pick it, but OpenSSL's FIPS provider rejects Ed25519 verification (`Invalid keyData`). Registration succeeds; authentication does not.

This PR:
- Pins registration to **ES256 (`-7`) and RS256 (`-257`)** only — both FIPS-compatible and mandatory for FIDO2 authenticators
- **Removes the platform-authenticator pre-check** that blocked hardware keys from registering
- **Adds a warning** in the passkey manager for existing security-key credentials that may need re-registration

## Steps to verify the change

- [ ] Register a new YubiKey/hardware security key on a FIPS instance — enrollment and MFA sign-in both succeed
- [ ] Register a platform passkey (Touch ID / Windows Hello) — still works
- [ ] Open passkey manager with an existing hardware key — warning is shown
- [ ] Re-register an affected security key after removing the old credential — MFA works

You can use chrome tools to generate a key:

image

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/webauthn/webauthn-service.ts
+++ b/backend/src/services/webauthn/webauthn-service.ts
@@ -91,7 +91,7 @@ export const webAuthnServiceFactory = ({
       // cannot be verified under FIPS mode (OpenSSL FIPS provider rejects the key
       // import with "Invalid keyData"), so the credential enrolls and then always
       // fails MFA. ES256 is mandatory for FIDO2 authenticators, so nothing is
-      // excluded in practice;
+      // excluded in practice
       supportedAlgorithmIDs: [-7, -257],
       excludeCredentials: existingCredentials.map((cred) => ({
         id: cred.credentialId,
--- a/backend/src/services/webauthn/webauthn-service.ts
+++ b/backend/src/services/webauthn/webauthn-service.ts
@@ -63,6 +63,13 @@ export const webAuthnServiceFactory = ({
   const RP_NAME = "Infisical";
   const RP_ID = new URL(appCfg.SITE_URL || "http://localhost:8080").hostname;
   const ORIGIN = appCfg.SITE_URL || "http://localhost:8080";
+  // ES256/RS256 only. The library default also offers/accepts EdDSA (-8), which
+  // Ed25519-capable authenticators (e.g. modern YubiKeys) pick first, but Ed25519
+  // cannot be verified under FIPS mode (the OpenSSL FIPS provider rejects the key
+  // import with "Invalid keyData"), so such a credential enrolls and then always
+  // fails MFA. ES256 is mandatory for FIDO2 authenticators, so nothing is excluded
+  // in practice.
+  const SUPPORTED_COSE_ALGORITHM_IDS = [-7, -257];
   /**
    * Generate registration options for a new passkey
    * This is the first step in passkey registration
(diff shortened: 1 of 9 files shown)


frontend/src/pages/user/PersonalSettingsPage/components/SecuritySection/PasskeyManagerDialog.tsx
```diff
@@ -44,6 +47,17 @@ export const PasskeyManagerDialog = ({ isOpen, onOpenChange }: Props) => {
...
+  // Hardware security keys registered before the server pinned ES256/RS256 may hold an
+  // EdDSA credential that some instances (FIPS mode) cannot verify at sign-in. We can't
+  // inspect the credential's algorithm client-side, so surface a soft notice whenever a
+  // security-key credential is present. Platform and phone passkeys always report
+  // "internal"/"hybrid" transports and are unaffected.
+  const hasSecurityKeyCredential = credentials.some(
+    (credential) =>
+      !credential.transports?.length ||
+      !credential.transports.some((transport) => transport === "internal" || transport === "hybrid")
+  );
```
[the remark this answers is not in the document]
victorvhs017: Known limitation. The API intentionally doesn't expose credential algorithms, and a schema addition seemed too much for an alert.

The test is conditional ("if signing in with your security key fails"), and we can come back to it if we start to get usage friction.

backend/src/services/webauthn/webauthn-service.ts
```diff
@@ -63,6 +63,13 @@ export const webAuthnServiceFactory = ({
   const RP_NAME = "Infisical";
   const RP_ID = new URL(appCfg.SITE_URL || "http://localhost:8080").hostname;
   const ORIGIN = appCfg.SITE_URL || "http://localhost:8080";
+  // ES256/RS256 only. The library default also offers/accepts EdDSA (-8), which
+  // Ed25519-capable authenticators (e.g. modern YubiKeys) pick first, but Ed25519
+  // cannot be verified under FIPS mode (the OpenSSL FIPS provider rejects the key
+  // import with "Invalid keyData"), so such a credential enrolls and then always
+  // fails MFA. ES256 is mandatory for FIDO2 authenticators, so nothing is excluded
+  // in practice.
+  const SUPPORTED_COSE_ALGORITHM_IDS = [-7, -257];
```
akhilmhdh, line 73: Should we do a check for fips mode and disable it just for them?

victorvhs017: We could, but if a user goes from a non-FIPS to a FIPS image, all existing hardware keys created using the EdDSA algorithm would stop working.

What I did in the last commit was make the alert message conditional on FIPS, so it won't show on non-FIPS images.

akhilmhdh (approved):

PR #7302, opened 2026-07-16
feat: add support for HMAC keys in KMS

carlosmonastyrski:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -2561,7 +2561,8 @@ export const KMS = {
     projectId: "The ID of the project to create the key in.",
     name: "The name of the key to be created. Must be slug-friendly.",
     description: "An optional description of the key.",
-    encryptionAlgorithm: "The algorithm to use when performing cryptographic operations with the key.",
+    algorithm: "The cryptographic algorithm of the key (e.g. aes-256-gcm, RSA_4096, HMAC_SHA_256).",
+    encryptionAlgorithm: "Deprecated: use 'algorithm' instead. Retained as an alias for backwards compatibility.",
     type: "The type of key to be created, either encrypt-decrypt or sign-verify, based on your intended use for the key.",
     isExportable:
       "Whether the raw key material can be exported after creation. When set to false, the key can never be exported regardless of permissions. This cannot be changed after creation."
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -2630,6 +2630,15 @@ export const KMS = {
     data: "The data in string format to be verified (base64 encoded). For data larger than 1MB you must first create a digest of the data and then pass the digest in the data parameter.",
     signature: "The signature to be verified (base64 encoded).",
     isDigest: "Whether the data is already digested or not."
+  },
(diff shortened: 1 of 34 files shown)


backend/src/server/routes/v1/cmek-router.ts
```diff
@@ -106,6 +114,17 @@ export const registerCmekRouter = async (server: FastifyZodProvider) => {
...
+            data.keyUsage === KmsKeyUsage.GENERATE_VERIFY_MAC &&
+            !Object.values(HmacAlgorithm).includes(data.encryptionAlgorithm as HmacAlgorithm)
+          ) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: `encryptionAlgorithm must be a valid HMAC algorithm. Valid options are: ${Object.values(
+                HmacAlgorithm
+              ).join(", ")}`
+            });
+          }
```
sheensantoscapadngan, line 133: It seems that with the introduction of HMAC, calling the field "encryption algorithm" is even more inappropriate now (especially in conjunction with "signing"). Perhaps we can rename this to be "key algorithm" instead or just "algorithm". Thoughts? We'll add a new property of course instead of breaking the APIs

backend/src/server/routes/v1/cmek-router.ts
```diff
@@ -919,6 +947,114 @@ export const registerCmekRouter = async (server: FastifyZodProvider) => {
     }
   });
```
sheensantoscapadngan, line 970: IGNORE PLACEMENT

nit but maybe we can just omit the MAC and data field from here? doesn't seem like it serves any purpose since it's just the same user input as the previous step
Image

frontend/src/context/ProjectPermissionContext/types.ts
```diff
@@ -44,6 +44,8 @@ export enum ProjectPermissionCmekActions {
   Decrypt = "decrypt",
   Sign = "sign",
   Verify = "verify",
+  GenerateMac = "generate-mac",
+  VerifyMac = "verify-mac",
   Rotate = "rotate",
   ExportPrivateKey = "export-private-key"
```
sheensantoscapadngan, line 50: have we mentioned this in the permissions doc already ?

backend/src/lib/api-docs/constants.ts
```diff
@@ -2561,7 +2561,8 @@ export const KMS = {
     projectId: "The ID of the project to create the key in.",
     name: "The name of the key to be created. Must be slug-friendly.",
     description: "An optional description of the key.",
-    encryptionAlgorithm: "The algorithm to use when performing cryptographic operations with the key.",
+    algorithm: "The cryptographic algorithm of the key (e.g. aes-256-gcm, RSA_4096, HMAC_SHA_256).",
+    encryptionAlgorithm: "Deprecated: use 'algorithm' instead. Retained as an alias for backwards compatibility.",
```
sheensantoscapadngan, line 2565: let's hide this from the API spec please

sheensantoscapadngan (approved):

PR #7318, opened 2026-07-16
feat(search): metadata search 

adilsitos:
## Context
This adds a way to search all secrets in all folders using metadata 

## Screenshots

https://github.com/user-attachments/assets/8747bbb3-a654-484c-98d3-3fd36345ca52

image

## Steps to verify the change
- check the video 

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


adilsitos:
@greptile review again and update the summary

backend/package.json
```diff
@@ -66,7 +66,7 @@
     "cipher-base": "1.0.5",
     "sha.js": "2.4.12",
     "jws": "^4.0.1",
-    "qs": "^6.14.2",
+    "qs": "$qs",
```
mathnogueira, line 69: any reason for this?

mathnogueira (approved):

backend/package.json
adilsitos: What changed in this PR is only the spelling. qs became a direct dependency here, and npm won't let you override a package you directly depend on unless the override spec exactly matches the direct spec. Using $qs resolves the override to whatever version we declare in dependencies, so the two stay in sync automatically instead of us having to keep two literals matching by hand.

adilsitos: Talked with @mathnogueira we could also remove the override, since it is the same version, but this was already placed before, so we will keep it avoid problems.

PR #7319, opened 2026-07-16
fix(project): chunk-delete environments during project hard delete

victorvhs017:
## Context

Project hard-delete already chunked `secret_versions_v2` ahead of the final cascade. Projects with thousands of environments could still hit enormous single-transaction cascades (folders → secrets → …), causing timeouts and failed cleanup jobs.

This adds a batched environment delete step (batch size 100) between secret-version pruning and the final project cascade, mirroring the existing pattern with per-batch `statement_timeout`, inter-batch sleep, and logging of `environmentsPruned`.

## Steps to verify the change

1. Soft-delete a project with many environments.
2. Wait for / trigger the project cleanup worker after `deleteAfter` expires.
3. Confirm the project is hard-deleted without timeout errors.
4. Check logs for `[environmentsPruned=N]` alongside `[versionsPruned=...]`.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/project-env/project-env-dal.ts
+++ b/backend/src/services/project-env/project-env-dal.ts
@@ -3,7 +3,7 @@ import { Knex } from "knex";
 import { TDbClient } from "@app/db";
 import { TableName, TProjectEnvironments } from "@app/db/schemas";
 import { DatabaseError } from "@app/lib/errors";
-import { buildFindFilter, ormify, TFindFilter, TFindOpt } from "@app/lib/knex";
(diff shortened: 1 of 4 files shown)


scott-ray-wilson (changes requested):
looks clean but one concern flagged by claude locally:

this moves the env cascade from the untimed final transaction to under a 30s `statement_timeout`, and there are several un-indexed relations

So seems like this might make us more prone to timing out now that it's bound on such a short timeout?

scott-ray-wilson (approved):
LGTM

PR #7335, opened 2026-07-18
feat(pam): connection test

lb-vn:
## Context

Connection test for postgres, ssh, windows.

Paired with gateway PR: https://github.com/Infisical/cli/pull/318

Runs on account creation, blocks if fail.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -269,6 +280,50 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
...
+
+    if (!gatewayId) {
+      throw new BadRequestError({ message: "A gateway must be attached to this account." });
+    }
+
+    const cd = connectionDetails as { host: string; port: number };
+    const request = buildRequest(connectionDetails, credentials);
+
+    try {
+      await testConnectionWithGateway(
```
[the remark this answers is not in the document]
x032205: I think this is fine

backend/src/ee/services/gateway-v2/gateway-v2-fns.ts
```diff
@@ -0,0 +1,41 @@
...
+export const testConnectionWithGateway = async (
+  targetHost: string,
+  targetPort: number,
+  gatewayId: string,
+  gatewayV2Service: TGatewayDep,
+  request: Record<string, unknown>,
+  timeoutMs: number,
+  signal?: AbortSignal
+): Promise<void> => {
+  const [host] = await verifyHostInputValidity({ host: targetHost, isGateway: true, isDynamicSecret: false });
```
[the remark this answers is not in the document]
x032205: I think this is fine, as they must still have create account or edit account perm

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -269,6 +280,50 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
...
+        gatewayId,
+        gatewayV2Service,
+        request,
+        CONNECTION_TEST_TIMEOUT_MS
+      );
+    } catch (err) {
+      throw new BadRequestError({
+        message: `Connection test failed: ${(err as Error).message || "unable to connect to the target"}`
+      });
+    }
```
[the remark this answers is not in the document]
x032205: I think this is fine, since these are provided by the account creator/editor anyways. It would be difficult to rewrite errors for all account and error types

backend/src/ee/services/gateway-v2/gateway-v2-fns.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+  if (!platform) throw new BadRequestError({ message: "Unable to connect to gateway" });
+
+  await withGatewayV2Proxy(
+    async (proxyPort) => {
+      const response = await callTestConnection({ port: proxyPort, body: request, timeoutMs, signal });
+      if (!response.ok) throw new Error(response.errorMessage);
+    },
+    {
+      protocol: GatewayProxyProtocol.Discovery,
```
sheensantoscapadngan: why is this discovery?

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -269,6 +280,50 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
     };
   };
 
+  // throws to block create/update when the account can't reach/authenticate its target
```
sheensantoscapadngan, line 286: IGNORE PLACEMENT

do we already handle the case when gateway is outdated? we don't want to force every single person to update their gateway instance...

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -519,6 +585,28 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
...
+      gatewayId !== undefined ||
+      gatewayPoolId !== undefined ||
+      templateId !== undefined
+    ) {
+      const effectiveConnectionDetails = connectionDetails
+        ? validateConnectionDetails(accountType, connectionDetails)
+        : validateConnectionDetails(accountType, await decrypt(projectId, existing.encryptedConnectionDetails));
+
+      // only test with credentials supplied in this request to prevent exfiltration
+      const testCredentials = credentials ? validateCredentials(accountType, credentials) : {};
```
bernie-g: I think if we do a connection test with empty credentials like this it'll just fail. I get why you're doing this but I don't think this will work

backend/src/lib/gateway-v2/test-connection-rpc.ts
```diff
@@ -0,0 +1,77 @@
...
+import { logger } from "@app/lib/logger";
+
+export type TestConnectionResponse = { ok: true; status: number } | { ok: false; status: number; errorMessage: string };
+
+export const TEST_CONNECTION_RPC_TIMEOUT_MS = 60_000;
+
+const MAX_RPC_RESPONSE_BYTES = 1024 * 1024;
+
+// calls the gateway's /v1/test-connection handler over the local relay proxy with an opaque request body
+export const callTestConnection = (args: {
```
bernie-g: We now have this function basically duplicated in five places. Instead of adding a 6th one would you mind combining all these into a shared util function? I see it in ssh-rpc.ts, winrm-rpc.ts, adcs-rpc.ts, pkcs11-rpc.ts, and test-connection-rpc.ts.

backend/src/ee/services/gateway-v2/gateway-v2-fns.ts
x032205: Because this was previously used solely for discovery and adapted to be more general and support connection tests too. I thought renaming it wouldn't be beneficial because:
1. It would introduce breaking changes (not too severe since discovery is also new, but still a problem)
2. "Discovery" actually still fits as a name here. It's the best name I could think of, because names like "Platform" or something are too general. Discovery encompasses probing, testing connections, port sweeps, etc

backend/src/ee/services/pam-account/pam-account-service.ts
x032205: Added a capability check. now only new gateway do block

backend/src/ee/services/pam-account/pam-account-service.ts
x032205: If you trace the code path it'll lead to only doing a TCP probe. So if no credentials are provided then it just does a simple probe. The reason we do this is to avoid a bad actor exfiltrating credentials by pointing to a compromised host

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -0,0 +1,73 @@
...
+      database: cd.database,
+      sslEnabled: cd.sslEnabled,
+      sslRejectUnauthorized: cd.sslRejectUnauthorized,
+      sslCertificate: cd.sslCertificate
+    };
+  },
+  [PamAccountType.SSH]: (_connectionDetails, credentials) => {
+    const creds = credentials as { authMethod: string; username: string; password?: string; privateKey?: string };
+
+    if (creds.authMethod === PamSshAuthMethod.Certificate || !isCredentialConfigured(PamAccountType.SSH, credentials)) {
```
bernie-g: From Claude:

 PATCH /pam/accounts/ssh/:id  {"connectionDetails": {"host": "127.0.0.1", "port": 2223}}
  → 400 "Connection test failed: invalid auth method: "

  The update path substitutes {} for missing credentials, and isCredentialConfigured(SSH, {}) returns true vacuously (no authMethod means no secret fields are "applicable"), so the builder sends mode: "ssh" with no auth method
  instead of falling back to the TCP probe. Postgres degrades correctly with the same input; SSH doesn't. UI is unaffected since the form always sends authMethod.

  Suggested fix: pass null instead of {} when the request has no credentials, and do the credentials === null || !isCredentialConfigured(...) → TCP fallback once in assertConnectionOk so builders only ever see real credentials.

bernie-g (approved):

PR #7338, opened 2026-07-19
feat: added new flags in place

akhilmhdh:
## Context

This PR adds more checks and flag in place for the new pricing setup

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/ee/services/license/license-types.ts
```diff
@@ -104,6 +107,21 @@ export type TFeatureSet = {
...
+  // Max secret syncs allowed org-wide. null (default) means uncapped; a number enforces the cap at
+  // creation time.
+  secretSyncLimit: null;
+  // Max internal CAs allowed org-wide (external CAs are not capped). null (default) means uncapped; a
+  // number enforces the cap at creation time.
+  maxInternalCas: null;
+  // Max PAM accounts allowed org-wide. null (default) means uncapped; a number enforces the cap at
+  // creation time.
+  maxPamAccounts: null;
+  // Product gate flags. null (default) is ignored; an explicit `true` blocks the product's primary
```
scott-ray-wilson: think this should be `false`

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -290,6 +293,37 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
       ResourcePermissionSub.PamResource
     );
 
+    const plan = await licenseService.getPlan(ctx.actorOrgId);
+    // The `pam` gate flag is ignored when null; an explicit `true` blocks PAM account creation.
```
scott-ray-wilson: same thing, think this should be `false`

backend/src/services/certificate-authority/certificate-authority-service.ts
```diff
@@ -308,9 +312,31 @@ export const certificateAuthorityServiceFactory = ({
       subject(ProjectPermissionSub.CertificateAuthorities, { name })
     );
 
+    const plan = await licenseService.getPlan(actor.orgId);
+    // The `certManager` gate flag is ignored when null; an explicit `true` blocks creating any CA
```
scott-ray-wilson: same here, should be `false`

backend/src/services/identity/identity-service.ts
```diff
@@ -379,8 +379,10 @@ export const identityServiceFactory = ({
       const deletedIdentity = await identityDAL.deleteById(id);
       await licenseService.updateSubscriptionOrgMemberCount(identityOrgMembership.scopeOrgId);
       usageMeteringService.emit(identityOrgMembership.scopeOrgId, MaxIdentities.key);
-      // Deleting the identity cascades its project + group memberships, so the secret-manager meter changes too.
+      // Deleting the identity cascades its project + group memberships, so the secret-manager and PAM
```
scott-ray-wilson: nit: a lot of comments updated of expanding secrets-manger to secrets-manager and PAM; these will be dated as soon as we add other products with metering, might make more generic? just flagging, no need to address

frontend/src/hooks/api/subscriptions/types.ts
```diff
@@ -84,4 +84,8 @@ export type SubscriptionPlan = {
   honeyTokens: boolean;
   honeyTokenLimit: number;
   secretsBrokering: boolean;
+  // Product gate flags. When explicitly true, the product's primary creation flow is blocked and an
```
scott-ray-wilson: same, should be false

frontend/src/layouts/PkiManagerLayout/PkiManagerLayout.tsx
```diff
@@ -1,22 +1,37 @@
...
       {assumedPrivilegeDetails && <AssumePrivilegeModeBanner />}
       {isDashboard && <CertManagerInstanceBanner />}
       <div className="flex-1 overflow-x-hidden overflow-y-auto bg-bunker-800 px-12 pt-10 pb-4">
         <Outlet />
       </div>
+      <UpgradePlanModal
+        isOpen={isUpgradeModalOpen}
+        onOpenChange={setIsUpgradeModalOpen}
+        text="Certificate Manager is not available on your current plan. Upgrade to continue using it."
+      />
```
scott-ray-wilson, line 34: should we actually prevent them from navigating into the product and do this on the product overview?

frontend/src/pages/organization/BillingV2Page/components/RemoveProductModal.tsx
scott-ray-wilson, line 1: nothing tells me I have a pending/scheduled cancellation 

Image

scott-ray-wilson, line 1: I also cannot un-cancel, still shows removed

Image

scott-ray-wilson, line 1: ignore file placement: 

Claude flagged this locally

Backend commitment validation is looser than the frontend's

buildCheckoutItems` (license-v2-service.ts:353-357) accepts a commitment for any dimension priced on both cadences with a *known* kind, and `KNOWN_PRICE_KINDS` includes `"metered"` (line 81). The frontend deliberately excludes metered dims from commitment (`isCommittable`: `!dim.meteredAnnual`, ActivateView.tsx:84) and now ignores server-reported `committed` on metered dims (`dimAnnualCommitted`, billing-v2-format.ts:102). A direct API call can therefore commit a metered-annual dimension, creating a subscription state the UI refuses to render as committed. Suggest requiring the **annual** price to be `per_unit` in `dimensionHasBothCadences`.

Also minor: when `cadence !== "annual"` with commitments, the thrown message ("plan has both monthly and annual pricing") describes the wrong failure; a cadence-specific message would help API consumers.

backend/src/ee/services/license/license-service.ts
```diff
@@ -556,6 +575,12 @@ export const licenseServiceFactory = ({
     actorAuthMethod,
     actorOrgId
   }: TCreateOrgPortalSession) => {
+    if (envConfig.DISABLE_LICENSE_V1_CLOUD) {
+      throw new BadRequestError({
+        message: "We're currently updating our license system. Please check back again later."
+      });
+    }
```
scott-ray-wilson, line 582: think this is intended but this doesn't just stop new checkouts but prevents existing users form editing

backend/src/db/migrations/utils/env-config.ts
scott-ray-wilson, line 1: ignore file placement: when I activated pam I had to hard refresh to be able to use it as feature cache wasn't invalidated:

1. **Frontend**: in `billingV2/mutations.tsx`, also invalidate `subscriptionQueryKeys.getOrgSubsription(orgId)` in the subscription-changing mutations (add product, checkout complete, cancel, resume, commitment change), so the gates lift without a reload.
2. **Backend**: delete `KeyStorePrefixes.LicenseCloudPlan(orgId)` wherever v2 billing calls `invalidateEntitlements`. The cleanest single-point fix is inside `licenseClient.invalidateEntitlements` itself, which covers all seven call sites.

scott-ray-wilson, line 1: ignore file placement:

I think the meter tracking for pam is incorrect, I've added users and it doesn't show anything; seems like nothing in `pam-membership-service.ts` was updated

scott-ray-wilson (changes requested):

backend/src/ee/services/license/license-types.ts
akhilmhdh: Hmm! I set it as false to avoid self hosting. But ya maidul told me it's ok to break it for folks

frontend/src/layouts/PkiManagerLayout/PkiManagerLayout.tsx
akhilmhdh: Good question

frontend/src/pages/organization/BillingV2Page/components/RemoveProductModal.tsx
akhilmhdh: I have switched to new logic where it gets cancelled immediately.

akhilmhdh: This is ok! The license server does the full validation on this including checks on commitment based etc. Metering is not supported as of now as well as none of the resources uses it.

scott-ray-wilson (approved):
LGTM

PR #7359, opened 2026-07-21
feat(agent-proxy): last used timestamp for proxied services

saifsmailbox98:
## Context

Surfaces a "last used" timestamp on each proxied service, so you can see when an agent last had a request brokered through the proxy for it. Until now the proxy logged per-request activity only to its own local stream (Infisical never stored any of it), so there was no server-side signal of whether a proxied service was actually being used. The agent proxy now reports usage back to the platform, and the secrets dashboard shows a relative "Used 5m ago" on each proxied service.

**Data model:** new nullable `lastUsedAt` column on `proxied_service`.

**Backend:** new `POST /proxied-services/:serviceId/report-usage` endpoint (identity-token only) that stamps `lastUsedAt`. The write is monotonic and clamps a future timestamp to `now()`, so an out-of-order or clock-skewed report can't move it backward or into the future. Because the write is skipped when the stored value is already newer, `updatedAt` isn't churned on no-op reports. No audit log entry — it fires roughly once a minute per active service.

**Permissions:** new `ReportUsage` action on the `ProxiedServices` subject, added to the Admin role and exposed in the project role editor. This lets a least-privilege proxy machine identity be granted just usage reporting, without `Proxy` or `Edit`.

**Frontend:** the proxied service row renders the relative time with an exact-timestamp tooltip (mirrors the identity "last used" pattern), renders nothing when never used, and keeps the multi-env matrix clean (last-used shows in the per-environment expansion).

**Reporting identity:** usage is reported by the proxy's own machine identity, never the agent's. The agent token lives in the untrusted execution environment by design, so letting it report would let a compromised agent forge usage; the proxy is the trusted observer.

Companion CLI PR: https://github.com/Infisical/cli/pull/322

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/routes/v1/proxied-service-router.ts
+++ b/backend/src/ee/routes/v1/proxied-service-router.ts
@@ -202,6 +202,26 @@ export const registerProxiedServiceRouter = async (server: FastifyZodProvider) =
   });
 
   server.route({
+    method: "POST",
+    url: "/:serviceId/report-usage",
+    config: { rateLimit: writeLimit },
+    onRequest: verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN]),
+    schema: {
+      hide: true,
+      tags: [ApiDocsTags.ProxiedServices],
+      description: "Report that the agent proxy brokered a request for this service",
+      params: z.object({ serviceId: z.string().uuid() }),
+      response: {
+        200: z.object({ success: z.boolean() })
+      }
+    },
+    handler: async (req) => {
+      await server.services.proxiedService.reportUsage({ serviceId: req.params.serviceId }, req.permission);
+      return { success: true };
+    }
+  });
+
+  server.route({
     method: "DELETE",
     url: "/:serviceId",
     config: { rateLimit: writeLimit },
--- a/backend/src/ee/services/proxied-service/proxied-service-service.ts
+++ b/backend/src/ee/services/proxied-service/proxied-service-service.ts
@@ -31,6 +31,7 @@ import {
   TProxiedServiceCredentialInput,
   TProxiedServiceDashboardCountDTO,
   TProxiedServiceDashboardListDTO,
+  TReportProxiedServiceUsageDTO,
   TUpdateProxiedServiceDTO
 } from "./proxied-service-types";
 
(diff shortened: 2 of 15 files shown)


backend/src/ee/services/proxied-service/proxied-service-service.ts
```diff
@@ -650,13 +651,47 @@ export const proxiedServiceServiceFactory = ({
...
+        environment: service.environmentSlug,
+        secretPath: resolvedSecretPath
+      })
+    );
+
+    // Clamp a future timestamp (e.g. proxy clock skew) to now; stampLastUsed enforces monotonicity.
+    const now = new Date();
+    const stamp = usedAt < now ? usedAt : now;
+    await proxiedServiceDAL.stampLastUsed(serviceId, stamp);
+    return { lastUsedAt: stamp };
```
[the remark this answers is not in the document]
saifsmailbox98: Valid observation but no real impact — the only caller (the agent proxy) ignores the response body, so this value is never consumed. Can drop the returned lastUsedAt entirely to remove the ambiguity.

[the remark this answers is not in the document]
saifsmailbox98: Already done — reportUsage returns nothing now and the route responds with an empty object. No rule needed.

backend/src/ee/routes/v1/proxied-service-router.ts
```diff
@@ -201,6 +201,30 @@ export const registerProxiedServiceRouter = async (server: FastifyZodProvider) =
...
+    config: { rateLimit: writeLimit },
+    onRequest: verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN]),
+    schema: {
+      hide: true,
+      tags: [ApiDocsTags.ProxiedServices],
+      description: "Report that the agent proxy brokered a request for this service",
+      params: z.object({ serviceId: z.string().uuid() }),
+      body: z.object({ usedAt: z.string().datetime() }),
+      response: {
+        200: z.object({})
```
akhilmhdh: I response can be a simple success: true

backend/src/ee/services/proxied-service/proxied-service-service.ts
```diff
@@ -650,13 +651,44 @@ export const proxiedServiceServiceFactory = ({
...
+    ForbiddenError.from(permission).throwUnlessCan(
+      ProjectPermissionProxiedServiceActions.ReportUsage,
+      subject(ProjectPermissionSub.ProxiedServices, {
+        environment: service.environmentSlug,
+        secretPath: resolvedSecretPath
+      })
+    );
+
+    // Clamp a future timestamp (proxy clock skew) to now.
+    const now = new Date();
```
akhilmhdh: I don't think we need to be this precise or detail. 😅

akhilmhdh: Also i think we should dedupe this on both side. That is here with a redis key - otherwise we would be bashing this and causing cpu lock spike

akhilmhdh (changes requested):

akhilmhdh (dismissed):

akhilmhdh (approved):

PR #7361, opened 2026-07-21
improvement(secret-syncs): allow user defined regions on replication policy for gcp

adilsitos:
## Context
We were just allowing users to select one region as replication when creating a new sync, but we should allow more than one. This PR also updates the docs with this information and move the screenshots to the new v3 component 

### Check compatibility with older version (using main branch)
- create a new sync on gcp that has the scope global and has a location ID
- checkout to this branch 
- check that nothing breaks and you can update the sync to have more regions 

### Create a new sync with scope global and several locations
- check that the new type of sync works fine 

Both of the previous scenarios you can do using the terraform, so the TF provider is also validated 

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


mathnogueira (dismissed):

adilsitos:
@greptile review the latest changes on the UI

mathnogueira (approved):

PR #7364, opened 2026-07-21
feat: add a instance version update check and display

scott-ray-wilson:
## Context

This PR adds a cron to get the latest release version once a week to display an update in the UI for self-hosted instances that aren't airgapped

## Screenshots

CleanShot 2026-07-21 at 12 05 06@2x

## Steps to verify the change

**Setup** (make the dev stack look like an internet-connected self-hosted instance on an old version):

1. In root `.env`, add `INFISICAL_PLATFORM_VERSION=0.150.0` and `VITE_INFISICAL_PLATFORM_VERSION=v0.150.0`, and ensure `LICENSE_SERVER_KEY` / `LICENSE_SERVER_V2_SERVICE_KEY` are NOT set (they mark the instance as cloud, which disables the check).
2. `docker compose -f docker-compose.dev.yml up -d --force-recreate backend frontend`

**Backend**

3. `docker logs infisical-dev-api 2>&1 | grep update-check` shows `cron[instance-update-check]: registered (pattern="0 0 * * 0")` and, within ~30s, `refreshed latest platform version [latestVersion=<latest>]` (boot fetch; the cron also fires a catch-up run for the missed Sunday slot, so no need to wait for a real Sunday).
4. `docker exec infisical-dev-redis redis-cli GET update-check-latest-version` returns `<latest>`; `TTL` is ~1209600 (14 days).
5. `curl -s localhost:4000/api/v1/admin/config | jq .config.latestAvailableVersion` returns `"<latest>"` (comparison is server-side; the field is only set when newer than the running version).

**UI** (http://localhost:8080, any logged-in user)

6. Navbar, left of the plan badge: info dot + "New version available"; hover shows `v0.150.0 → v<latest>`; click opens the Infisical upgrade tool (https://upgrade.infisical.com/). Nothing is rendered when up to date.
7. Help dropdown (`?` icon): `Version: 0.150.0 (v<latest> available)`.

**Gating** (apply each to `.env`, recreate backend, observe)

8. `DISABLE_UPDATE_CHECK=true`: the only log line is `cron[instance-update-check]: disabled` (no outbound request at all), `latestAvailableVersion` null, indicator gone.
9. `INFISICAL_PLATFORM_VERSION=<latest>`: check still runs (logs appear) but the field is null and the indicator is gone, since the instance is up to date.
10. Remove `INFISICAL_PLATFORM_VERSION` (dev default): check fully disabled, no version UI anywhere.
11. Offline-license auto-disable: `LICENSE_KEY=$(printf '{"license":{},"signature":"x"}' | base64)` (offline keys are detected structurally, so a fake one drives the real path; a boot WARN "offline license verification failed" is expected). Confirm: the only update-check log line is `cron[instance-update-check]: disabled`, `latestAvailableVersion` null even with a previously cached Redis value, indicator gone.

**Tests**: `cd backend && npm run test:unit` covers semver parse/compare and the full gating matrix (cloud, flag, offline license, hash/missing version) in `update-check-fns.test.ts`.

Air-gapped failure mode: the fetch fails quietly (debug log only, no cron retries, no failed-run noise), the field stays null, and the UI shows nothing.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/update-check/update-check-queue.ts
+++ b/backend/src/services/update-check/update-check-queue.ts
@@ -59,14 +59,15 @@ export const updateCheckServiceFactory = ({ cronJob, keyStore }: TUpdateCheckSer
   };
 
   const init = () => {
-    if (!isEnabled) return;
-
     cronJob.register({
       name: CronJobName.InstanceUpdateCheck,
       pattern: "0 0 * * 0", // weekly on Sunday at midnight UTC
       runHashTtlS: 3 * 24 * 60 * 60,
+      enabled: isEnabled,
       handler: refreshLatestVersion
     });
+
+    if (!isEnabled) return;
 
     // also refresh once at boot so a fresh instance is not blind until the next Sunday fire
     void keyStore
--- a/backend/src/services/update-check/update-check-queue.ts
+++ b/backend/src/services/update-check/update-check-queue.ts
@@ -1,0 +1,97 @@
+import { getLicenseKeyConfig } from "@app/ee/services/license/license-fns";
+import { LicenseType } from "@app/ee/services/license/license-types";
+import { KeyStorePrefixes, KeyStoreTtls, TKeyStoreFactory } from "@app/keystore/keystore";
+import { getConfig } from "@app/lib/config/env";
+import { request } from "@app/lib/config/request";
+import { CronJobName, TCronJobFactory } from "@app/lib/cron/cron-job";
+import { logger } from "@app/lib/logger";
+
+import { isVersionNewer, parseSemanticVersion } from "./update-check-fns";
+
+const GITHUB_LATEST_RELEASE_URL = "https://api.github.com/repos/Infisical/infisical/releases/latest";
+
+type TUpdateCheckServiceFactoryDep = {
+  cronJob: TCronJobFactory;
+  keyStore: Pick<TKeyStoreFactory, "getItem" | "setItemWithExpiry">;
+};
+
+export type TUpdateCheckServiceFactory = ReturnType<typeof updateCheckServiceFactory>;
+
+export const updateCheckServiceFactory = ({ cronJob, keyStore }: TUpdateCheckServiceFactoryDep) => {
+  const appCfg = getConfig();
+
+  const licenseKeyConfig = getLicenseKeyConfig();
+  const hasOfflineLicense = licenseKeyConfig.isValid && licenseKeyConfig.type === LicenseType.Offline;
+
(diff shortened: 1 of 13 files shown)


scott-ray-wilson:
@greptile review

backend/src/services/update-check/update-check-queue.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const init = () => {
+    if (!isEnabled) return;
+
+    cronJob.register({
+      name: CronJobName.InstanceUpdateCheck,
+      pattern: "0 0 * * 0", // weekly on Sunday at midnight UTC
+      runHashTtlS: 3 * 24 * 60 * 60,
+      handler: refreshLatestVersion
+    });
```
victorvhs017: The cronJob.register has an "enabled" param we can use here.

frontend/src/layouts/OrganizationLayout/components/NavBar/Navbar.tsx
```diff
@@ -699,6 +701,16 @@ export const Navbar = () => {
                 <div className="flex items-center gap-2 px-3 py-1.5 text-xs text-muted">
                   <Info className="size-3.5" />
                   Version: {envConfig.PLATFORM_VERSION}
+                  {serverConfig.latestAvailableVersion && (
+                    <a
+                      href="https://github.com/Infisical/infisical/releases"
```
victorvhs017: Should we redirect the user to Docker instead? The release notes can be too much for the regular user.

Maybe the upgrade tool? https://upgrade.infisical.com/

frontend/src/layouts/OrganizationLayout/components/NavBar/VersionBadge.tsx
```diff
@@ -16,9 +18,33 @@ export const VersionBadge = () => {
...
-      {`v${version}`}
-    </Badge>
+    <Tooltip>
+      <TooltipTrigger asChild>
+        <a
+          href="https://github.com/Infisical/infisical/releases"
+          target="_blank"
+          rel="noopener noreferrer"
+          className="mt-1 mr-2 hidden items-center gap-x-1.5 text-xs text-muted transition-colors hover:text-accent md:inline-flex"
+        >
```
victorvhs017, line 40: Same here

victorvhs017 (changes requested):

victorvhs017 (approved):
LGTM!

PR #7370, opened 2026-07-21
feat: azure auto rotate app connection credentials

mathnogueira:
## Context

Add Vault App Connection credential rotations. This enable the credential rotation for the remaining 4 Azure App Connections. This is only valid for `client secret` auth method.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/app-connection/azure-app-configuration/azure-app-configuration-connection-fns.ts
+++ b/backend/src/services/app-connection/azure-app-configuration/azure-app-configuration-connection-fns.ts
@@ -103,10 +103,11 @@ export const validateAzureAppConfigurationConnectionCredentials = async (
       };
 
     case AzureAppConfigurationConnectionMethod.ClientSecret:
-      const { tenantId, clientId, clientSecret } = inputCredentials as {
+      const { tenantId, clientId, clientSecret, clientSecretKeyId } = inputCredentials as {
         tenantId: string;
         clientId: string;
         clientSecret: string;
+        clientSecretKeyId?: string;
       };
 
       try {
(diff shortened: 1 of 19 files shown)


backend/src/services/app-connection/azure-app-configuration/azure-app-configuration-connection-fns.ts
adilsitos, line 1: ignore comment placement.

To rotate the credentials, we need more permissions. From what I noticed, we need to have one of those: 

Image 

Can you add these into the docs so the users know which permissions are required?

adilsitos, line 1: ignore placement.

the rotation on app connections don't have a way to validate that they were successful. The rotation can fail and the user never knows, there is no UI indication. I think we can create a ticket and let the team know, so we can fix this in another time, what do you think? 

A health check button to check if the creadential is valid would also be really great.

adilsitos, line 1: ignore comment placement.

This step on the docs was really hard to do, because there isn't those options on the current UI (the permissions here seem to be the ones from azure key vault). Can we update this screenshot and the text as well? 

https://infisical.com/docs/integrations/app-connections/azure-app-configuration#azure-app-configuration-permissions

adilsitos: This is also required on the azure dns in case they want to support rotation

adilsitos: this is also required for azure key vault

adilsitos (changes requested):
Left some recommendations regarding the docs. I think it is being hard to follow. Also, I left a suggestion for us to create in another time. The rotations are working fine!

adilsitos (approved):
LGTM. Thanks for updating the docs!

PR #7371, opened 2026-07-21
improvement(identity-access-token): cache org membership in revocation verdict

victorvhs017:
## Context

Every identity access token validation previously called `orgDAL.findEffectiveOrgMembership` after the revocation check, adding a Postgres read to a high-frequency auth path. Revocation verdicts were already cached in Redis (`identity-revocation-verdict:{identityId}:{fingerprint}`, ~10 min TTL with jitter), but membership was re-queried on every request regardless.

This PR folds org membership into the revocation verdict path. On a cache miss (or stale version), `assertTokenIsNotRevoked` now checks revocation markers **and** membership, then caches the result as allow or deny. Stamped allow hits skip both the revocation DAL and the membership query. New org-scoped revocation markers (`scope: orgId`) reject tokens issued before deactivate/remove for that org only — tokens scoped to other orgs in a hierarchy stay valid. `membershipIdentityService` calls `revokeTokensForIdentityOrgMembership` when an identity org membership is deactivated or deleted, bumping `identity-revocation-version:{identityId}` so cached allows are invalidated.

`findEffectiveOrgMembership` is also optimized to try a direct membership row lookup first before falling back to the group-membership path, reducing cost on the cold-path recheck.

## Steps to verify the change

1. Authenticate with a machine identity token twice in quick succession; confirm the second request does not call `findEffectiveOrgMembership` (cached allow at `identity-revocation-verdict:{identityId}:{fingerprint}`).
2. Deactivate an identity's org membership (`isActive: false`); confirm existing tokens for that org are rejected and `identity-revocation-version:{identityId}` increments.
3. Remove an identity from an org; confirm the same org-scoped rejection behavior.
4. In a sub-org hierarchy, confirm tokens whose JWT `orgId` differs from the revoked org still authenticate.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/e2e-test/routes/v1/identity-access-token-redesign.spec.ts
+++ b/backend/e2e-test/routes/v1/identity-access-token-redesign.spec.ts
@@ -131,8 +131,7 @@ const deleteOrgIdentityMembership = async (identityId: string) => {
   // The real removal paths (membershipIdentityService.deleteMembership,
   // identityV1.deleteIdentity sub-org branch) bump the identity's revocation
   // version so cached "allow" verdicts are rechecked once against Postgres. This
-  // helper mutates the row directly, so mirror that bump here — otherwise the
-  // cache-gated membership check keeps serving the stale allow until the TTL.
+  // helper mutates the row directly, so mirror that bump here.
   await testRedis.incr(KeyStorePrefixes.IdentityRevocationVersion(identityId));
 };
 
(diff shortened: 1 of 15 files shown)


akhilmhdh (commented):
One question: So when a membership becomes inactive we mark it as well. Are we not invalidating when they become back active or just accepting that it will bring it back after cache expires

victorvhs017:
> One question: So when a membership becomes inactive we mark it as well. Are we not invalidating when they become back active or just accepting that it will bring it back after cache expires

I added the invalidation when the machine identity becomes a member of a group or organization again. This added more version invalidation spots, and if we are bumping the version more, we'll have more cache misses.

I don't think this membership process is done as frequently to reduce the cache effectiveness, but if we observe a high number of calls for this membership query or the revocation query, we should split this into two different cache key sets.

akhilmhdh (approved):

PR #7376, opened 2026-07-22
fix(pki): base certificate renewal eligibility on the certificate's enrollment method

saifsmailbox98:
## Context

Renewal and auto-renewal eligibility were gated on the certificate **profile's** `enrollmentType`. Under the applications flow a single profile can expose multiple enrollment methods (e.g. API + ACME), so that field no longer reflects how a given certificate was actually issued. An API-issued cert under a profile labeled `acme`/`est`/`scep` was wrongly rejected (`ACME certificates cannot be auto-renewed`) even though Infisical holds its private key and can renew it.

Eligibility is now judged per-certificate: the enrollment method recorded on the cert's own `certificate_request`, combined with the existing "Infisical holds the private key" check. Protocol-enrolled certs (where the client owns the key) stay correctly blocked, and disabling auto-renewal is now allowed regardless of enrollment method.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/certificate-v3/certificate-v3-service.ts
+++ b/backend/src/services/certificate-v3/certificate-v3-service.ts
@@ -2301,10 +2301,7 @@ export const certificateV3ServiceFactory = ({
     });
   };
 
-  // Blocks renewal for certificates enrolled via an external/protocol method (ACME/EST/SCEP), where the client
-  // owns the private key and drives its own renewal. Judged by the certificate's OWN enrollment record, never the
-  // profile's enrollmentType (a legacy label that drifts under the application flow). Only an explicit non-API
-  // value blocks; a missing signal defers to the stored-private-key check, so older certs are not wrongly blocked.
+  // Only an explicit non-API enrollment blocks; a missing signal defers to the key-presence check.
   const $assertCertificateEnrollmentRenewable = async (
     certId: string,
     { context, tx }: { context: "renewal" | "auto-renewal"; tx?: Knex }
--- a/backend/src/services/certificate-v3/certificate-v3-service.ts
+++ b/backend/src/services/certificate-v3/certificate-v3-service.ts
@@ -1,5 +1,6 @@
 import { ForbiddenError, subject } from "@casl/ability";
 import { randomUUID } from "crypto";
+import { Knex } from "knex";
 import RE2 from "re2";
 
 import { ActionProjectType, ResourceType, TCertificates } from "@app/db/schemas";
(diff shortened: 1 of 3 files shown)


backend/src/services/certificate/certificate-dal.ts
```diff
@@ -555,6 +555,25 @@ export const certificateDALFactory = (db: TDbClient) => {
...
   };
 
+  // Returns how THIS certificate was actually enrolled, read from its originating certificate_requests row
+  // (earliest wins). Deliberately does NOT fall back to the profile's enrollmentType: that field is a legacy
+  // per-profile label that drifts under the application flow (a profile created as "acme" can still issue API
+  // certs), which is exactly what misjudges renewability. Returns null when there is no request row, so callers
+  // can defer to another signal (the stored-private-key check) rather than block on a stale label.
+  const getRequestEnrollmentTypeByCertId = async (certId: string, tx?: Knex): Promise<string | null> => {
+    try {
+      const row = (await (tx || db.replicaNode())(TableName.CertificateRequests)
```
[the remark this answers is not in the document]
saifsmailbox98: Valid, switched to a primary read so replica lag can't turn a missing request row into a false eligible. The described bypass can't actually happen today (protocol-enrolled certs store no secret, so the key check already blocks them) — this just keeps the guard sound for the future case it's meant to protect.

saifsmailbox98:
@greptile review

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -2355,15 +2369,10 @@ export const certificateV3ServiceFactory = ({
...
-
-        if (profile.enrollmentType !== EnrollmentType.API) {
-          throw new ForbiddenRequestError({
-            message:
-              "Certificate is not eligible for renewal: Only certificates issued from an API enrollment profile can be renewed through this endpoint"
-          });
-        }
       }
 
+      await $assertCertificateEnrollmentRenewable(originalCert.id, { context: "renewal", tx });
```
carlosmonastyrski: Should we use an enum for these values of "context"?

saifsmailbox98: Rather than an enum, I dropped the shared helper and inlined the check at both call sites, so the `context` values are gone entirely. Reads like the other eligibility checks there now.

carlosmonastyrski (approved):

PR #7384, opened 2026-07-22
feat: add expire alert for machine identity authentication

Thiago-AS:
## Context

This PR introduces a generic **Alerts** system that allows users to receive proactive notifications when resources require attention.

The first supported alert provider covers **identity credential expiration**. Users can now be notified before a Universal Auth client secret expires, instead of only discovering the issue after authentication starts failing.

Alerts can be scoped to an **organization** or **project**, and optionally to a specific **identity**. Users can configure how far in advance they want to be notified (for example, 30 days or 1 week before expiration) and deliver notifications through one or more channels:

- Email
- Slack
- Webhook
- PagerDuty

## Screenshots

image

image

image

### EMAIL

image

### WEBHOOK

```
{
  "specversion": "1.0",
  "type": "com.infisical.identity.authentication.expiration",
  "source": "/alerts/b59581a5-2f67-4fa1-93db-954e07154642",
  "id": "217235bd-6d1e-4a12-87e9-79842889ece7",
  "time": "2026-07-22T18:15:24.503Z",
  "datacontenttype": "application/json",
  "subject": "identity.authentication.expiry",
  "data": {
    "alert": {
      "id": "b59581a5-2f67-4fa1-93db-954e07154642",
      "name": "Org Expire",
      "resourceType": "identity.authentication",
      "condition": "1d"
    },
    "items": [
      {
        "id": "ua-client-secret:de98e305-384c-4411-8be7-22cd4aac57df",
        "title": "Org Identity",
        "fields": [
          {
            "label": "Secret Name",
            "value": "30 min secret"
          },
          {
            "label": "Secret Type",
            "value": "Universal Auth Client Secret"
          },
          {
            "label": "Expires",
            "value": "July 22, 2026 at 06:44 PM UTC"
          }
        ]
      }
    ],
    "metadata": {
      "totalItems": 1,
      "viewUrl": "http://localhost:8080/organizations/e0e5e42d-dff4-4209-bdbb-401f6faa3ca2/identities/3afeff21-dfd7-44bb-a03d-caf8960e2a70"
    }
  }
}
```

### SLACK

image

### Happy path

- Create an organization-scoped alert with an Email channel.
  - Create a Universal Auth client secret that expires within the configured window.
  - Verify the email is delivered and an `alert_history` record is created.
- Repeat using a project-scoped alert that targets a single identity.
  - Only that identity's expiring credential should trigger the alert.
- Configure all four notification channels on a single alert and verify each one delivers successfully.

### Deduplication

- Let an alert fire once, then immediately trigger another processing run.
  - No duplicate notification should be sent.
- Enable `dailyReminder`.
  - The notification should be sent again during the next daily run.

### Authorization and security

- As a user from Organization A, attempt to `GET`, `PATCH`, or `DELETE` an alert belonging to Organization B by calling the API directly.
  - The request should return `403` or `404` without leaking resource information.
- Create a channel with a webhook signing secret.
  - Retrieve the alert and verify the secret is never returned.
  - The response should only expose flags such as `hasSigningSecret: true`.
- Attempt to configure a Webhook or Slack URL that points to:
  - `http://`
  - `http://169.254.169.254`
  - `http://localhost`
  - All should be rejected.
- Attempt to add a recipient (user or group) that is not a member of the organization or project.
  - The request should fail validation.
- Remove a recipient user from the organization or project, then trigger the alert.
  - They should no longer receive notifications.
  - Their recipient mapping should be automatically pruned.

### Cleanup

- Delete an identity with an alert scoped to it.
  - The associated alerts and channels should be removed.
- Soft-delete a project.
  - Its alerts should stop being processed and no external notifications should be sent.

### Failure handling

- Configure a Webhook endpoint that consistently returns HTTP 500.
  - The delivery should be marked as failed.
  - The failure should be logged.
  - Because delivery never succeeded, the notification should be retried on the next processing run.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking Change
- [ ] Documentation
- [ ] Chore

---

## Checklist

- [x] Tested locally.
- [ ] Documentation updated (if needed).
- [ ] `CLAUDE.md` files updated (if needed).
- [ ] Contributing guide reviewed.


backend/src/services/alert/alert-engine.ts
```diff
@@ -0,0 +1,196 @@
...
+                sendLimit(() => definition.send({ channelId: channel.id, config, payload, recipient, deps }))
+              )
+            );
+            const failures = results.filter((result) => !result.success);
+            if (failures.length === results.length) {
+              return { ...base, success: false, error: failures.map((f) => f.error).join("; ") };
+            }
+            if (failures.length > 0) {
+              logger.warn(
+                `Alert ${channel.channelType} channel had partial delivery failures [alertId=${alert.id}] [channelId=${channel.id}]: ${failures
```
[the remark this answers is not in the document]
Thiago-AS: This is the intended behavior for now. As long as at least one recipient successfully receives the email, the delivery is marked as successful.

backend/src/services/alert/channels/alert-channel-registry.ts
```diff
@@ -0,0 +1,44 @@
...
+    type: AlertChannelType.SLACK,
+    directed: false,
+    secretFields: ["webhookUrl"],
+    configSchema: SlackChannelConfigSchema,
+    send: sendSlackNotification
+  },
+  [AlertChannelType.WEBHOOK]: {
+    type: AlertChannelType.WEBHOOK,
+    directed: false,
+    secretFields: ["signingSecret"],
```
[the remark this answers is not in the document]
Thiago-AS: The signing secret is the secret, not the URL, following the same model as PKI channels

backend/src/db/migrations/20260714120000_add-alerts.ts
```diff
@@ -0,0 +1,173 @@
...
+      t.foreign("alertHistoryId").references("id").inTable(TableName.AlertHistory).onDelete("CASCADE");
+      t.foreign("channelId").references("id").inTable(TableName.AlertChannel).onDelete("SET NULL");
+      // Serves the dedup lookup: join on alertHistoryId, filter status=success + targetId IN (...).
+      t.index(["alertHistoryId", "targetId"]);
+      t.index("channelId");
+    });
+
+    await createOnUpdateTrigger(knex, TableName.AlertHistoryTarget);
+  }
+}
```
victorvhs017, line 152: We could add a history cleanup to the daily cleanup job. AWS keeps a 30-day history; it seems like a good start, and we can tweak as required.

Image

backend/src/server/routes/v1/alert-router.ts
```diff
@@ -0,0 +1,197 @@
...
+        resourceId: z.string().nullable().optional(),
+        eventType: z.string().min(1),
+        condition: z.unknown().optional(),
+        enabled: z.boolean().optional(),
+        projectId: z.string().nullable().optional(),
+        channels: z.array(CreateChannelInputSchema).min(1).max(MAX_CHANNELS_PER_ALERT)
+      }),
+      response: { 200: z.object({ alert: AlertResponseSchema }) }
+    },
+    onRequest: verifyAuth([AuthMode.JWT]),
```
victorvhs017: Any reason why we couldn't add machine identity here? I can see this as a very common terraform provider resource in the future, and that would require it.

This is also true for the other endpoints here.

backend/src/db/migrations/20260714120000_add-alerts.ts
```diff
@@ -0,0 +1,173 @@
...
+      t.text("description").nullable();
+      t.string("resourceType").notNullable();
+      t.string("resourceId").nullable();
+      t.string("eventType").notNullable();
+      t.string("triggerType").notNullable();
+      t.jsonb("condition").nullable();
+      t.boolean("enabled").notNullable().defaultTo(true);
+      t.uuid("orgId").notNullable();
+      t.string("projectId").nullable();
+      t.uuid("createdByUserId").nullable();
```
victorvhs017: Following the last comment, this would mean that we should update this to actorId, to be extensible. (same thing for all the other fields with the same name here)

backend/src/server/routes/v1/alert-router.ts
```diff
@@ -0,0 +1,197 @@
...
+    handler: async (req) => {
+      const alert = await server.services.alert.createAlert({
+        ...req.body,
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId
+      });
+      return { alert };
+    }
```
victorvhs017, line 107: We are missing audit logs here

backend/src/server/routes/v1/alert-router.ts
```diff
@@ -0,0 +1,197 @@
...
+      const alert = await server.services.alert.updateAlert({
+        alertId: req.params.alertId,
+        ...req.body,
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId
+      });
+      return { alert };
+    }
```
victorvhs017, line 258: We are missing audit logs here.

backend/src/server/routes/v1/alert-router.ts
```diff
@@ -0,0 +1,197 @@
...
+    handler: async (req) => {
+      const alert = await server.services.alert.deleteAlert({
+        alertId: req.params.alertId,
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId
+      });
+      return { alert };
+    }
```
victorvhs017, line 295: Same here

backend/src/services/alert/channels/alert-channel-pagerduty-fns.ts
```diff
@@ -0,0 +1,100 @@
...
+        isAxiosErrorRetryable,
+        { channelId: ctx.channelId, channelLabel: "PagerDuty" }
+      )
+    )
+  );
+
+  const errors = results.filter((result) => !result.success && result.error).map((result) => result.error);
+  if (errors.length > 0) {
+    return { success: false, error: errors.join("; ") };
+  }
```
victorvhs017: If one item fails, all targets would be recorded as failed.

Also, we don't split the error by target, so we can't know which one failed.

If we are not adding errors per target, we need to know in some other way. The pagerduty api returns the dedup_key

Image

backend/src/services/alert/alert-engine.ts
```diff
@@ -0,0 +1,197 @@
...
+                  .map((f) => f.error)
+                  .join("; ")}`
+              );
+            }
+            return { ...base, success: true };
```
victorvhs017, line 183: We need to build an alarm history, even if it's just a table in the alarm sheet. We need to know when the alarm was sent, to whom, and if it was delivered or not, and if not, why not.

This is super important because that's the only place an administrator can check to validate "why we didn't act if the alarm is set up?"

Losing an alarm can be critical.

I would also argue that, as a guaranteed deliver, we send the alarms as in-app notifications. So even if the channel failed, we alarmed the user somehow.

backend/src/services/alert/channels/alert-channel-retry-fns.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const isAxiosErrorRetryable = (err: unknown): boolean => {
+  const axiosErr = err as AxiosError;
+  const status = axiosErr.response?.status;
+  if (status === 429) return true;
+  if (status && status >= 500) return true;
+  if (axiosErr.code && RETRYABLE_NETWORK_ERRORS.includes(axiosErr.code)) return true;
+  if (axiosErr.message?.toLowerCase().includes("timeout")) return true;
+  return false;
+};
```
victorvhs017: Our request already has the `axiosRetry` setup; wouldn't this be redundant?

Image

backend/src/services/alert/providers/identity-credential-alert-provider.ts
```diff
@@ -0,0 +1,266 @@
...
+    const alertBefore = (alert.condition as { alertBefore?: string } | null)?.alertBefore;
+
+    return {
+      alert: {
+        id: alert.id,
+        name: alert.name,
+        orgId: alert.orgId,
+        resourceType: alert.resourceType,
+        ...(alertBefore ? { condition: alertBefore } : {}),
+        viewUrl
```
victorvhs017, line 158: We need to add the projectId here too, right? Or the webhook payload would always set the source as org

backend/src/services/alert/providers/identity-credential-alert-dal.ts
```diff
@@ -0,0 +1,115 @@
...
+    try {
+      const row = (await (tx || db.replicaNode())(TableName.Project)
+        .where({ id: projectId })
+        .select("type")
+        .first()) as { type: string } | undefined;
+      return row?.type ?? null;
+    } catch (error) {
+      throw new DatabaseError({ error, name: "GetProjectType" });
+    }
+  };
```
victorvhs017, line 140: We need to check if the project is soft-deleted. Do we have this query already in the project DAL?

backend/src/services/alert/providers/identity-credential-alert-provider.ts
```diff
@@ -0,0 +1,266 @@
...
+      case ProjectType.PAM:
+        return `${siteUrl}/organizations/${orgId}/pam`;
+      default:
+        return `${siteUrl}/organizations/${orgId}/projects/${projectType}/${projectId}`;
+    }
+  };
+
+  const buildViewUrl = async (alert: TAlertContext): Promise<string> => {
+    const appCfg = getConfig();
+    const siteUrl = appCfg.SITE_URL ?? "https://app.infisical.com";
```
victorvhs017: nit: we can remove the `?? "https://app.infisical.com"`, the app wouldn't work without a SITE_URL

backend/src/services/alert/providers/identity-credential-alert-provider.ts
```diff
@@ -0,0 +1,266 @@
...
+  };
+
+  const targetId = (target: TIdentityCredentialTarget): string => `${target.credentialType}:${target.id}`;
+
+  const assertResourceInScope = async (input: {
+    orgId: string;
+    projectId?: string | null;
+    resourceId?: string | null;
+  }): Promise<void> => {
+    if (!input.resourceId) return;
```
victorvhs017, line 186: I think we should throw here until we support organization, and project, wide identity alerts in the UI. Right now, you can create via the API alerts that aren’t bound to a specific identity. They won’t appear anywhere in the UI, but the daily cron will still enqueue and run them.

backend/src/services/alert/alert-engine.ts
victorvhs017, line 1: Not specific only to this file, but to the whole due targets logic.

I can see two potential issues:

1 - We schedule the job at one time (00:00), but when the job runs, the secret has already expired.

For example:

- I have an alarm set to 1 day before
- The secret expires at 00:05, the cron job ran at 00:00 and scheduled the job.
- For some reason, the job took too long to execute, and it executed at 00:06
- That alert would be lost because the secret is expired

2 - What does "remind me X days before" actually mean for us? Right now we are not reminding the user 1 day before; we are doing it related to when the job runs.

For example:

- I have an alarm set to 1 day before
- The secret expires at 14:00
- The job executes at 00:00
- The alarm is going off 14 hours before the secret expires, not 1 day before.

I understand that to actually match it, we would need to run the cron every minute, but we can keep the current daily cron and certify that the alarm will run AT LEAST X days before.

Another option, and possibly the best one here, would be to use the job `delay` property https://api.docs.bullmq.io/classes/v5.Job.html#delay; then we could send the alarm at the exact time before the expiration.

victorvhs017, line 1: Another consideration is regarding how we separate the scheduling and the alarm.

In the cron, we enqueue all active alarms, but we only validate whether those alarms will send anything when the job runs. In a scenario where we have tens of thousands of alarms, that could mean a good amount of jobs basically doing no-op and only taking Redis resources on the queue.

One example of no-op job: 30d alert, secret alerted yesterday -> We only do the dedup when the job runs.

Why not do this whole process in the cron and only enqueue jobs that we are certain will do something?

backend/src/services/alert/alert-engine.ts
```diff
@@ -0,0 +1,197 @@
...
+        const dueTargets = targets.filter((target) => !alertedSet.has(`${channel.id}:${provider.targetId(target)}`));
+        const cap = definition?.maxTargetsPerRun;
+        if (cap && dueTargets.length > cap) {
+          logger.info(
+            `Alert ${channel.channelType} channel capped at ${cap} targets this run; ${dueTargets.length - cap} deferred to the next run [alertId=${alert.id}] [channelId=${channel.id}]`
+          );
+          return { channel, dueTargets: dueTargets.slice(0, cap) };
+        }
+        return { channel, dueTargets };
+      })
```
victorvhs017, line 94: With the cap, we could lose alarms, right?

If the secret expired tomorrow and the alarm never happened because it's always after the cap limit, would the next run on the next day be a no-op because the secret expired?

backend/src/services/alert/alert-queue.ts
victorvhs017, line 1: Reminder: after merging this PR we need to update the cron job list in the infrastructure repo.

backend/src/services/alert/alert-service.ts
```diff
@@ -0,0 +1,408 @@
...
+            createdByUserId: alert.createdByUserId
+          },
+          cipher.encryptor,
+          tx
+        );
+        // eslint-disable-next-line no-await-in-loop -- one shared tx connection; writes must be serial
+        await alertChannelMembershipDAL.insertMany([{ alertId: alert.id, channelId: channel.id }], tx);
+      }
+    }
+  };
```
victorvhs017, line 319: With this reconciliation, in a scenario where the alert already happened, if I remove an email channel, for example, and add another email channel, I would be alerted again, right? Because the dedupe uses the channelId.

Is this expected or should we not alert?

frontend/src/hooks/api/alerts/mutations.ts
victorvhs017, line 1: Unrelated to the file.

Regarding permissions. I see that we are using the permissions from the resource of the alert. But I think we should have an alerts dedidcated permissions; it could be an action, as we have for tokens:

Image

frontend/src/pages/organization/IdentityDetailsByIDPage/components/IdentityAlertDetail.tsx
victorvhs017, line 1: We are missing documentation for this alert feature.

victorvhs017 (changes requested):

backend/src/db/migrations/20260714120000_add-alerts.ts
Thiago-AS: Yeah, sounds good

backend/src/server/routes/v1/alert-router.ts
Thiago-AS: Will add audit logs!

backend/src/services/alert/alert-engine.ts
Thiago-AS: we do, we have AlertHistory and AlertHistoryTarget, both works for history tracking and deduping.

backend/src/services/alert/providers/identity-credential-alert-provider.ts
Thiago-AS: Makes sense!

backend/src/services/alert/providers/identity-credential-alert-dal.ts
Thiago-AS: I'm already filtering soft deleted project on the alert side, this is mainly used to build the viewUrl.

backend/src/services/alert/alert-engine.ts
Thiago-AS: The main reason is that we want the cron to stay as lightweight as possible. Cron jobs are time-bound and are expected to run on a predictable schedule. If we move all the filtering, deduplication, and business logic into the cron, it becomes slower and more likely to overrun or miss its schedule as the number of alarms grows.

The cron's responsibility should just be triggering work. The actual processing, deduplication, and decision of whether an alarm should be sent belongs in the worker job, which is designed for retries, scaling, and resilience. A no-op job is generally a much cheaper failure mode than a cron that can't keep up with its schedule.

backend/src/services/alert/alert-engine.ts
victorvhs017: Where are we showing this in the UI?

victorvhs017: Oh, by table I meant a table on the alarm screen!

Thiago-AS: Oh, you mean in the UI? Not for now. Maidul wants to keep it simple and only add more when we actually have customers asking for it. So no top-level alert management or any kind of filtering for alerts yet.

backend/src/services/alert/channels/alert-channel-retry-fns.ts
Thiago-AS: The axiosRetry config lives in createRequestClient, but the alert channels don't use that client. They send requests through safeRequest (our SSRF-safe Axios wrapper), which doesn't have any built-in retry logic.

backend/src/services/alert/alert-engine.ts
Thiago-AS: 1. Fixed by anchoring the alert window to the cron fire time instead of the worker's clock. The cron now stamps each job with a `scheduledAt` timestamp, which gets passed all the way down to the query. The lower bound is now `expiresAt > asOf`, where `asOf` is the cron fire time.

2. Since the cron only runs once a day, the only way to guarantee "at least X before expiration" is to start scanning one polling interval early. The upper bound is now `asOf + alertBefore + 1 day`.

   That means alerts will fire somewhere between **X and X+1 days** before expiration, guaranteeing they're never sent later than X. For example, with a 1-day alert and a secret expiring on day D at 14:00, it'll be picked up by the D-1 00:00 run and fire about 38 hours before expiration instead of 14 hours. The email still says "expiring within 1 day"; the extra day is just an internal scan buffer to make sure we never undershoot. We also extended the dedup window accordingly so each alert still only fires once.

> Another option, and possibly the best one here, would be to use BullMQ's `delay` property so we could send the alert at the exact time before the expiration.

I agree, but I think that would add a lot of complexity that we should avoid for now. There's no clear roadmap for this feature yet, and moving to delayed jobs would fundamentally change the model. Alerts would become per-secret instead of per-rule, we'd have to schedule and cancel jobs whenever a secret is created, revoked, its TTL changes, or an alert is edited, and we'd lose the nice self-healing property of recomputing everything from the database on each daily run.

backend/src/services/alert/alert-engine.ts
Thiago-AS: The cap only applies to PagerDuty (it's 10, since it opens one incident per target). Email, Slack, and webhooks batch everything into a single message, so they're never capped.

For PagerDuty though, I added a order by on expireAt so we make sure we alarm the most urgent ones. I would leave the cap as of now, I don't think we will run into an issue with a machine identity with more than 10 UA.

backend/src/services/alert/alert-service.ts
Thiago-AS: Dedup keys off channelId. If you remove an email channel and add a new one, reconciliation creates a brand new channel with a new ID (since the incoming payload has no ID, it goes through the create path). That new channel has no alert history, so it gets notified again.

I'd say that's expected and fine. Dedup is intentionally scoped per channel because each channel is an independent destination. A brand new channel has never been notified before, so sending it notifications is consistent. It's the same as adding a Slack channel to an alert that has already fired via email - you'd expect the new channel to receive notifications too. It only re-fires for targets that are still actually due (within the alert window and not expired), so it isn't resurrecting old alerts.

backend/src/services/alert/channels/alert-channel-retry-fns.ts
victorvhs017: safeRequest still uses our custom axios request as the base and we don't overwrite the retry:

image

backend/src/services/alert/alert-engine.ts
victorvhs017: Could we make a concession here on the in-app notification?

Just fire a notification alongside the alert, and that's it. Would this add too much extra complexity?

backend/src/services/alert/alert-engine.ts
victorvhs017: Yep, I think that works for version 1! Should be enough to catch some tight expiration dates related to when the cron runs.

victorvhs017: Let's at least add some observability here so we know how many of these are actually sending alarms.

If the percentage is too big, we'll probably want to revisit it and update to something like multiple job levels. The cron enqueues a planner, and this planner fans out the actual alert jobs.

backend/src/services/alert/channels/alert-channel-retry-fns.ts
Thiago-AS: Yeah, nice catch. I missed that. Let me update the current retry logic to handle 429 responses as well, for some reason it only retry > 500.

backend/src/services/alert/alert-engine.ts
Thiago-AS: Agreed to keep as it is for now 👍

backend/src/services/alert/alert-engine.ts
```diff
@@ -0,0 +1,208 @@
...
+  smtpService
+}: TAlertEngineDep) => {
+  const runAlert = async (alert: TAlerts, opts?: { asOf?: Date }) => {
+    const provider = alertProviderRegistry.get(alert.resourceType);
+    if (!provider) {
+      logger.warn(`No alert provider registered for resource type '${alert.resourceType}' [alertId=${alert.id}]`);
+      return;
+    }
+
+    const targets = await provider.findDueTargets({
```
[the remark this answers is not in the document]
Thiago-AS: that's intended.

frontend/src/hooks/api/alerts/mutations.ts
Thiago-AS: Since we are tying to a specific resource for now, we decided to not add top level permissions/product. So for now we will keep the identity permissions to handle that (if we expand, a specific alert management will be needed).

backend/src/services/alert/alert-history-dal.ts
```diff
@@ -0,0 +1,131 @@
...
+            }))
+          );
+        }
+
+        return history;
+      });
+    } catch (error) {
+      throw new DatabaseError({ error, name: "CreateWithTargets" });
+    }
+  };
```
victorvhs017, line 54: When the status is failed, we could save the reason in the DB, as we do for secret syncs. This helps a lot, and customers can solve most things without the need for a support ticket to check logs.

backend/src/services/alert/channels/alert-channel-slack-fns.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  return {
+    text: escapeSlackText(`${headerText} - ${payload.summary}`),
+    blocks: [{ type: "header", text: { type: "plain_text", text: headerText, emoji: true } }],
+    attachments: [{ color, blocks: attachmentBlocks }]
+  };
+};
+
+const triggerSlackWebhook = async (url: string, payload: TSlackPayload): Promise<void> => {
+  await safeRequest.post(url, payload, {
```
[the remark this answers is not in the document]
victorvhs017: The debug log is ok, but the real issue is here:

Image

I validated in CloudWatch, and we should redact the Slack webhooks here.

We could add this to Claude.MD, too.

backend/src/services/membership-user/membership-user-service.ts
```diff
@@ -554,6 +557,12 @@ export const membershipUserServiceFactory = ({
           additionalPrivilegeDAL,
           approvalPolicyDAL
         });
+        // Prune the user's alert-channel recipient rows across the whole org (org- and
+        // project-scoped channels) since leaving the org drops all their access.
+        await alertChannelRecipientDAL.deleteUsersRecipientsByScope(
+          { userIds: [dto.selector.userId], orgId: dto.permission.orgId },
+          tx
+        );
```
victorvhs017: We prune the recipients, but what about the case when all recipients from a channel are removed? That channel will not be removed and will linger in the DB, right? If we have alerts where the only channels are these empty channels, would we be scheduling alarms for no one?

If we don't allow users to create alarms without recipients, we should clean up these empty channels and alerts from the DB.

docs/documentation/platform/identities/machine-identities.mdx
```diff
@@ -85,6 +86,91 @@ Once revoked, the token is rejected on subsequent authenticated requests and can
...
+<Note>
+  The alert watches the **client secret**, not the access tokens minted from it. Access tokens are
+  short-lived by design and are renewed programmatically. See
+  [Universal Auth](/documentation/platform/identities/universal-auth) for the difference between
+  Client Secret TTL, Access Token TTL, and Access Token Max TTL.
+</Note>
+
+### Create an Alert
+
+Open the machine identity you want to monitor and find the **Alert** row in its details. Click the bell icon to create one.
```
victorvhs017, line 104: Can we add a screenshot here?

frontend/src/views/Alerts/AlertForm.tsx
```diff
@@ -0,0 +1,450 @@
...
+        });
+        createNotification({ text: "Successfully created alert", type: "success" });
+      }
+      onComplete();
+    } catch (err) {
+      const message =
+        (err as { response?: { data?: { message?: string } } })?.response?.data?.message ??
+        "Failed to save alert";
+      createNotification({ text: message, type: "error" });
+    }
```
victorvhs017: We already handle the error notification on the global error handler; we can drop this to avoid the duplicated toast.

backend/src/services/alert/alert-principal-scope-fns.ts
```diff
@@ -0,0 +1,54 @@
...
+      projectId,
+      userIds,
+      groupIds
+    });
+    return { userIds: new Set(effectiveUserIds), groupIds: new Set(effectiveGroupIds) };
+  }
+
+  const inScopeUserIds = new Set<string>();
+  if (userIds.length) {
+    const memberships = await orgDAL.findMembership({ $in: { actorUserId: userIds }, scopeOrgId: orgId });
```
victorvhs017: orgDAL.findMembership doesn't filter deactivated users, nor users who were invited but didn't accept the invite yet.

They would receive the alerts.

backend/src/services/user/user-service.ts
```diff
@@ -566,6 +570,11 @@ export const userServiceFactory = ({
 
     const user = await userDAL.deleteById(userId);
 
+    await alertChannelRecipientDAL.deleteByPrincipals({
+      principalType: AlertPrincipalType.USER,
+      principalIds: [userId]
+    });
```
victorvhs017: We could make this atomic to avoid orphans. Same for group delete.

frontend/src/views/Alerts/AlertForm.tsx
victorvhs017, line 1: I think this button is too hidden:

Image

For PKI, we have a whole card + list:

Image

victorvhs017, line 1: This is related to the last comment. I understand we'll only have one option now, and you can only have one possible configuration; that's why a table wouldn't make sense, but looking at the form:

Image

All the trigger options seem like they should be clickable, and you could change them, but they're not, again because we only have this combination.

I feel like a more extensible design would be hinting that we expect to have more options in the future.

This could be done by having the table as we have for PKI (what's the issue with allowing multiple alerts to the same machine identity?). And also with selects that only have one option for the triggers, as we have for the machine identity auth templates, for example:

Image

But with the cooler container you have for each trigger.

victorvhs017, line 1: If we are using numbered steps, a V3 updated example is the secret sync sheet:

Image

victorvhs017, line 1: It's not clear here that we are talking about Universal Auth Tokens. The same is valid for the scope; it seems like we are watching all authentications.

Image

frontend/src/views/Alerts/AlertForm.tsx
```diff
@@ -0,0 +1,450 @@
...
+            </FixedField>
+
+            <Controller
+              control={control}
+              name="eventType"
+              render={({ field: { value } }) => (
+                <FixedField label="Event">
+                  <Badge variant="warning">{ALERT_EVENT_TYPE_LABELS[value]}</Badge>
+                  <span className="text-xs text-muted">
+                    Fires as the authentication nears expiration
```
victorvhs017: I'm not sure about the wording here. What do you think about these?

- Notifies before a credential expires
- Triggers before credential expiration

frontend/src/views/Alerts/ChannelCard.tsx
victorvhs017, line 1: The select should open the options above the field if we don't have more space at the bottom, instead of extending the height of the sheet.

Image

victorvhs017, line 1: A test button for each channel would be great, so the user doesn't have to wait until the next day to see if their webhook/slack/pagerduty integration is working. Even more important because we don't show failures.

Image

victorvhs017, line 1: Unrelated to the file:

Image

I feel like 1 and 2 are saying the same thing. Could we remove that header in 1 and start with the description?

victorvhs017, line 1: 1 - Do we align text in the center for all our emails? If not the actual pattern, could we justify it?

2 - I think we could specify that it's the machine identity's alert settings

Image

frontend/src/views/Alerts/AlertForm.tsx
Thiago-AS: I talked with Scott about this. Right now, since we only have a 1:1 relationship and an identity can only have one alert, having a whole list card for it feels like overkill. We decided to keep it where it is for now.

Thiago-AS: > All the trigger options seem like they should be clickable, and you could change them, but they're not, again because we only have this combination.
> 
> I feel like a more extensible design would be hinting that we expect to have more options in the future.

In this case, we want these to be fixed because the alert is created from within the resource itself. The only thing that could potentially change is the event, but to keep the experience consistent, we decided to leave it as is for now.

If we ever introduce a top level Alerts page where users can create alerts independently of a resource, then what you described is exactly how we'd implement it, with those options being selectable instead of fixed.

backend/src/services/alert/alert-history-dal.ts
Thiago-AS: I had that before, but since we don't have a way to manage alert history or errors yet, I removed it to avoid leaving dead code. I added logs instead, since we'd still need to check this kind of thing manually either way. 

When we decide to improve the alerts system overall, we can definitely add that. For now, though, it wouldn't provide much value, so I'd keep just the logs.

backend/src/services/membership-user/membership-user-service.ts
Thiago-AS: This is already a no-op. I don't think we should remove something as a side effect. Deleting an admin's alert configuration as a result of an unrelated member removal is an implicit destructive cascade. But to make the behavior more explicit, I'll disable the channel when there are no recipients available.

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -1016,6 +1040,11 @@ export const superAdminServiceFactory = ({
       scope: AccessScope.Organization,
       id: membershipId
     });
+
+    if (organizationMembership?.actorUserId) {
+      await alertChannelRecipientDAL.pruneOutOfScopeRecipients({ userIds: [organizationMembership.actorUserId] });
+    }
+
```
victorvhs017, line 1054: This one should not be in a transaction with the delete above

frontend/src/views/Alerts/AlertForm.tsx
victorvhs017: I still think we need to add somewhere in this form that we are talking about Universal Auth Client Secrets specifically.

frontend/src/views/Alerts/AlertDetail.tsx
victorvhs017, line 1: The "Enabled" toggle has a different background color compared to the bg.

Image

Image

frontend/src/views/Alerts/AlertDetail.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const handleDeleteAlert = async () => {
+    if (!existingAlert) return;
+
+    try {
+      await deleteAlert.mutateAsync({ alertId: existingAlert.id });
+      createNotification({ text: "Successfully deleted alert", type: "success" });
+      handlePopUpToggle("deleteAlert", false);
+    } catch {
+      createNotification({ text: "Failed to delete alert", type: "error" });
+    }
```
victorvhs017: This would also be duplicated notification

backend/src/server/routes/v1/alert-router.ts
```diff
@@ -0,0 +1,277 @@
...
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
+    handler: async (req) =>
+      server.services.alertChannelTest.testChannel({
+        ...req.body,
+        actor: req.permission.type,
+        actorId: req.permission.id,
+        actorAuthMethod: req.permission.authMethod,
+        actorOrgId: req.permission.orgId
+      })
```
victorvhs017: I think we should audit-log the tests too, as they may be alerting people, and an admin may want to know who triggered that.

victorvhs017 (changes requested):

frontend/src/hooks/api/alerts/types.ts
victorvhs017, line 1: Unrelated to the file.

We need to update the metrics docs here: https://infisical.com/docs/self-hosting/guides/monitoring-telemetry#available-metrics

victorvhs017 (changes requested):

docs/self-hosting/guides/monitoring-telemetry.mdx
victorvhs017, line 1: nit: there's a scrollbar here for some reason:

Image

docs/self-hosting/guides/monitoring-telemetry.mdx
```diff
@@ -854,6 +854,45 @@ Per-provider observability for the audit-log stream feature (Datadog, Splunk, Cu
...
+      - `outcome` (string): one of the closed enum values below
+
+    **Outcomes**:
+
+    | Outcome | Meaning |
+    |---------|---------|
+    | `delivery_success` | Every channel the run touched delivered. |
+    | `delivery_partial` | At least one channel failed, so the notification was dropped on that channel. Dedup is per channel and target, so the next run retries only the failed channels. |
+    | `delivery_failed` | No channel delivered — the whole run dropped. |
+    | `alert_not_found` | The alert was deleted (or its project soft-deleted) between enqueue and run. |
```
victorvhs017: nit: the project soft-deletion is backend-only; it's not something transparent to the user in the UI like the environment soft delete. We can leave only delete.

victorvhs017 (changes requested):

docs/self-hosting/guides/monitoring-telemetry.mdx
Thiago-AS: I don't know how to remove that lol

victorvhs017 (approved):

PR #7385, opened 2026-07-22
feat(pam): connection test for rest of accounts

lb-vn:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -18,50 +31,196 @@ export type TestConnectionRequest =
...
+    case PamAccountType.MySQL:
+    case PamAccountType.MsSQL: {
+      const cd = connectionDetails as {
+        database: string;
+        sslEnabled?: boolean;
+        sslRejectUnauthorized?: boolean;
+        sslCertificate?: string;
+      };
+      const c = creds as { authMethod?: string; username: string; password?: string } | null;
+      if (!c || (accountType === PamAccountType.MsSQL && c.authMethod !== "sql-login")) return tcp(host, port);
```
saifsmailbox98, line 100: since we moved to the gateway shouldn't the other auth methods be also validated via the gateway correctly?

or is that something we'll do later?

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -657,13 +670,22 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
         : validateConnectionDetails(accountType, await decrypt(projectId, existing.encryptedConnectionDetails));
 
       // only test with credentials supplied in this request to prevent exfiltration
```
saifsmailbox98, line 672: why shouldn't we test again, like for example if a username is updated?

right now, if you update something like username, it just saves it without testing it.

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -18,50 +31,196 @@ export type TestConnectionRequest =
...
+        ldapTlsServerName?: string;
+      };
+      const c = creds as { username: string; password?: string } | null;
+      if (!c) return tcp(host, port);
+      return {
+        host,
+        port,
+        request: {
+          mode: TestConnectionMode.LDAP,
+          username: c.username,
```
saifsmailbox98: Does this work if the username is just sAMAccountName? can we use `qualifyUsernameWithDomain` here?


Discovery also imports them like that, so you'd find later when adding the password

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -18,50 +31,196 @@ export type TestConnectionRequest =
...
+          mode: TestConnectionMode.SSH,
+          authMethod: c.authMethod,
+          username: c.username,
+          password: c.password,
+          privateKey: c.privateKey
+        }
+      };
+    }
+    case PamAccountType.Windows:
+      return tcp(host, port);
```
saifsmailbox98, line 216: can we use `winrm.Ping` in the gateway for RDP?

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -18,50 +31,196 @@ export type TestConnectionRequest =
...
+          sslEnabled: cd.sslEnabled,
+          sslRejectUnauthorized: cd.sslRejectUnauthorized,
+          sslCertificate: cd.sslCertificate
+        }
+      };
+    }
+    case PamAccountType.Kubernetes: {
+      const cd = connectionDetails as { sslRejectUnauthorized?: boolean; sslCertificate?: string };
+      const c = creds as { authMethod: string; serviceAccountToken?: string } | null;
+      if (!c || c.authMethod !== "service-account-token" || !c.serviceAccountToken) return tcp(host, port);
```
saifsmailbox98: even gateway auth can be checked, right? since the gateway is the same.

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-connection-test.ts
x032205: We deferred NTLM and Kerberos auth due to client library complications on the Go side. Will likely tackle later

backend/src/ee/services/pam-account/pam-account-service.ts
x032205: We only test when password is updated to prevent a bad actor from pointing the account to some compromised host and having credentials be sent to it that they originally didn't set

saifsmailbox98: are we prioritizing exfiltration prevention over connection testing here? is that the right thing to do? Also, isn't having edit access a privileged access? 

Someone with the ability to add new gateway and edit accounts can also create a fake gateway to get the credentials.

backend/src/ee/services/pam-account/pam-account-connection-test.ts
x032205: Sadly winrm is different protocol from RDP so it's likely not going to work

backend/src/ee/services/pam-account/pam-account-connection-test.ts
saifsmailbox98: ok

saifsmailbox98 (approved):

PR #7395, opened 2026-07-23
fix(project): let org admins break-glass into projects they only belong to via a group

Thiago-AS:
## Context

An org admin with `organization-admin-console` + `access-all-projects` couldn't grant themselves admin access to a project if they were already a project member **through a group**.

This happened because of two issues:

1. **The "Grant Admin" button never showed up.** The All Projects page was checking `isMember`, which includes both direct and group membership. So if you only had access through a group, the UI treated you as already joined and hid the button, even though you didn't have a direct role.
2. **Leaving the project didn't work either.** `leaveProject` only checks direct project members, so group-only members got the confusing error: `"You are not a member of this project"`.

**This PR changes that:**

- `searchProjects` now returns both `isMember` and `isDirectMember`. The UI now uses `isDirectMember` to decide whether to show the `Grant Admin` button.
  - Group-only members with `access-all-projects` will now see both the `Joined` badge and the `Grant Admin` button.
  - Direct members will just see `Joined`.
  - Users with no access will still see `Join as Admin`.
- `leaveProject` now returns a proper error for group-only members, explaining that their access comes from a group and needs to be removed there instead.

## Screenshots

image

Screenshot 2026-07-23 at 16 23 16

## Steps to verify

1. Log in as an org admin.
2. Create a group (**Organization → Access Control → Groups**) and add a second test user to it.
3. Create a project and add the group under **Project → Access Control → Groups** with the **Member** role.
4. Log in as the group-only test user:
   - **Without** `access-all-projects`: the project shows as `Joined`, and trying to leave returns the new group-access error.
   - **With** `access-all-projects`: the project shows `Joined` + `Grant Admin`. Clicking `Grant Admin` gives the user a direct Admin role, and the row updates to just `Joined`.
5. Verify a real non-member (no direct or group access) still sees `Join as Admin`, and the deprecated leave flow still returns `"You are not a member of this project"`.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format.
- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/group/user-group-membership-dal.ts
+++ b/backend/src/ee/services/group/user-group-membership-dal.ts
@@ -57,9 +57,14 @@ export const userGroupMembershipDALFactory = (db: TDbClient) => {
 
   /**
    * Given a set of [userIds], returns the subset that have access to project [projectId]
-   * through a (non-pending) group membership, i.e. they belong to a group that is itself
-   * a member of the project. Used to validate approvers/bypassers who are project members
-   * via a group rather than directly.
+   * through a group membership, i.e. they belong to a group that is itself
+   * a member of the project.
+   *
+   * Note: this intentionally does NOT filter on `isPending`. The permission engine
+   * (`permission-dal.getPermission`) grants project access purely by group membership
+   * without considering `isPending`, and the flag is no longer cleared for accepted
+   * users (the un-pend transition was removed with the signup-flow simplification).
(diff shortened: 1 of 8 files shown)


scott-ray-wilson (approved):
LGTM

PR #7398, opened 2026-07-23
fix(ui): show grant-privileges action for org custom roles

Thiago-AS:
## Context

Custom org roles couldn't grant the **Grant Privileges** action for **Machine Identity Management** or **Group Management** from the UI, even though the API accepted it just fine.

The issue was that the org role builder reuses `GeneralPermissionPolicies`, a component originally built for project roles. That component hides the `grant-privileges` action because it's deprecated at the project level (it was replaced by `assign-role` / `assign-additional-privileges`).

The problem is that org and project permissions use the same string values (`"identity"`, `"groups"`, `"grant-privileges"`), so the project-specific "hide deprecated actions" logic was also running for org roles. As a result, the UI was hiding a valid permission that the backend still supports and requires.

At the organization level, `grant-privileges` is **not** deprecated. It's still the permission the backend checks when assigning roles to org identities and groups, and there's no replacement, so it should always be visible.

## Steps to verify

1. In the UI, go to **Organization Access Control → Roles** and create or edit a custom role.
2. Add a **Machine Identity Management** policy and open the actions dropdown. Verify that **Grant Privileges** is available.
3. Repeat the same for **Group Management** and verify that **Grant Privileges** is available there as well.
4. Save the role, reopen it, and confirm the permission was persisted.
5. Verify project roles still behave correctly by opening a project custom role and confirming the deprecated **Grant Privileges (Legacy)** and **Secrets Read** actions remain hidden unless the role already contains them.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the conventional commit format (`type(scope): short description`).
- [x] Tested locally.
- [ ] Updated docs (if needed).
- [ ] Updated `CLAUDE.md` files (if needed).
- [ ] Read the contributing guide.

### DIFF
--- a/frontend/src/pages/organization/ProductSettingsPage/project-templates/EditProjectTemplateSection/components/ProjectTemplateEditRoleForm.tsx
+++ b/frontend/src/pages/organization/ProductSettingsPage/project-templates/EditProjectTemplateSection/components/ProjectTemplateEditRoleForm.tsx
@@ -13,6 +13,7 @@ import { slugSchema } from "@app/lib/schemas";
 import { AddPoliciesButton } from "@app/pages/project/RoleDetailsBySlugPage/components/AddPoliciesButton";
 import {
   GeneralPermissionPolicies,
+  PermissionScope,
   TPermissionAction
 } from "@app/pages/project/RoleDetailsBySlugPage/components/GeneralPermissionPolicies";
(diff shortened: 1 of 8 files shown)


scott-ray-wilson (approved):
LGTM

PR #7403, opened 2026-07-23
fix: webhook skipped-payload marking and username generation loop guard

PedroHenrique0713:
## Summary

Two fixes in Infisical backend.

### 1. Skipped webhook payloads incorrectly marked as success (`webhook-fns.ts`)
When `getWebhookPayload` returned null for unhandled event types, the map returned `undefined`. `Promise.allSettled` wrapped this as `{status:'fulfilled'}`, causing webhooks to be marked `lastStatus: "success"` in the DB despite never being triggered. Fix: filter out null payloads before `Promise.allSettled`.

### 2. `normalizeUsername` potential infinite loop (`user-fns.ts`)
The `do...while` loop had no max retry bound. Under pathological conditions (exhausted namespace from pre-filled usernames), the registration endpoint would hang forever. Fix: `for` loop with `MAX_RETRIES = 20`, throws on exhaustion.

### DIFF
--- a/backend/src/services/user/user-fns.ts
+++ b/backend/src/services/user/user-fns.ts
@@ -6,12 +6,14 @@ import { TUserDALFactory } from "@app/services/user/user-dal";
 export const normalizeUsername = async (username: string, userDAL: Pick<TUserDALFactory, "findOne">) => {
   let attempt: string;
   let user;
+  const MAX_RETRIES = 20;
 
-  do {
+  for (let retries = 0; retries < MAX_RETRIES; retries++) {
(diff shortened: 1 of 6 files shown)


PedroHenrique0713:
Closing this one. I opened a stack of PRs in this repo today and, because each branch was cut from the previous one instead of from the base, they overlap: this PR carries the commits of the earlier ones as well. I am consolidating the work in #7400 and will resubmit the remaining fixes individually, on top of the base branch, once that one has been reviewed. Sorry for the noise.

PR #7413, opened 2026-07-24
feat: add SCEP support for Microsoft Intune

carlosmonastyrski:
## Context

Adds Microsoft Intune as a SCEP challenge-validation backend, and an option to have the SCEP RA certificate signed by the profile's CA.

### Changes
- **New `microsoft-intune` app connection** (client-secret auth): validates each device's Intune challenge via the Intune SCEP API and reports issuance success/failure back to Intune.
- **Intune challenge type** on SCEP enrollment: pick the Microsoft Intune challenge and a connection instead of a shared/dynamic password. Cert-based renewal is forced off for Intune so validation runs on every request.
- **"Sign RA certificate with the CA" option**: self-signed by default; enabling it (internal CAs only, for now) issues the RA from the profile's CA so it chains to the CA root, which strict clients like Apple and Intune require.
- **Fix**: `validateAlgorithmCompatibility` now matches both stored signature-algorithm orderings (`SHA256-RSA` and `RSA-SHA256`).

### Notes
- Approval policies are not supported on Intune SCEP profiles (Intune issues synchronously)

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,103 @@
...
+      kmsService: deps.kmsService
+    })) as TMicrosoftIntuneConnectionCredentials;
+
+    const graphToken = await getMicrosoftEntraToken(credentials, MicrosoftEntraTokenResource.Graph);
+    const serviceUri = await discoverScepValidationServiceUri(graphToken);
+    const intuneToken = await getMicrosoftEntraToken(credentials, MicrosoftEntraTokenResource.Intune);
+
+    cached = { serviceUri, intuneToken };
+    return cached;
+  };
```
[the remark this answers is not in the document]
carlosmonastyrski: The cache lives inside a single `intuneDelegatedHandler` call, which is per request (validate then report, a few seconds), while the token is valid for ~1 hour, so it never expires mid use.

backend/src/ee/services/pki-scep/pki-scep-service.ts
```diff
@@ -496,20 +488,24 @@ export const pkiScepServiceFactory = ({
...
+    await validationHandler
+      .reportIssued?.({
+        transactionId: parsed.transactionId,
+        csrDer,
+        certificateDer: result.issuedCertDer,
+        validationConnectionId: scepConfig.validationConnectionId
+      })
+      .catch((reportErr) =>
+        logger.error(reportErr, `Failed to report SCEP issuance success [transactionId=${parsed.transactionId}]`)
+      );
```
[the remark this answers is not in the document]
carlosmonastyrski: Leaving as is. Hard failing the request on a `successNotification` error would cause an infinite issue then notify-fail then retry then issue loop whenever the failure is deterministic (for example Intune rejecting the payload), minting a new cert on every retry. The current log and continue behavior causes at most one bounded Intune re-push cycle, which is the safer tradeoff.

backend/src/ee/services/pki-scep/pki-scep-service.ts
```diff
@@ -456,37 +448,62 @@ export const pkiScepServiceFactory = ({
...
-        actor: { type: ActorType.SCEP_ACCOUNT, metadata: { profileId: profile.id } },
-        event: {
-          type: EventType.SCEP_ENROLLMENT,
-          metadata: { ...auditMetadata, status: "pending" as const }
-        }
-      });
+      if (scepConfig.challengeType === ScepChallengeType.MICROSOFT_INTUNE) {
+        const failReason =
+          "Certificate issuance requires approval and cannot complete synchronously for Microsoft Intune.";
+        void logEnrollmentEvent({ status: "failure", failReason });
```
sheensantoscapadngan: let's use enums please

backend/src/ee/services/pki-scep/pki-scep-service.ts
```diff
@@ -456,37 +448,62 @@ export const pkiScepServiceFactory = ({
...
+          error: err instanceof Error ? err.message : "Certificate issuance failed",
+          validationConnectionId: scepConfig.validationConnectionId
+        })
+        .catch((reportErr) =>
+          logger.error(reportErr, `Failed to report SCEP issuance failure [transactionId=${parsed.transactionId}]`)
+        );
+      throw err;
+    }
 
     if (result.status === "pending") {
```
sheensantoscapadngan: and here

backend/src/ee/services/pki-scep/pki-scep-fns.ts
```diff
@@ -59,6 +118,131 @@ export const generateRaCertificate = async (
...
+    return buildRaCertificate(slug);
+  }
+
+  const ca = await deps.certificateAuthorityDAL.findByIdWithAssociatedCa(caId);
+  const caType = (ca?.externalCa?.type as CaType) ?? CaType.INTERNAL;
+  if (!isScepRaCaSigningSupported(caType)) {
+    logger.warn(
+      `SCEP RA CA-signing requested for unsupported CA type '${caType}'; using a self-signed RA [caId=${caId}]`
+    );
+    return buildRaCertificate(slug);
```
sheensantoscapadngan: shouldn't we be failing HARD here?

backend/src/ee/services/pki-scep/pki-scep-service.ts
```diff
@@ -496,18 +513,46 @@ export const pkiScepServiceFactory = ({
...
+          transactionId: parsed.transactionId,
+          recipientNonce: parsed.senderNonce,
+          failInfo: ScepFailInfo.BadRequest
+        });
       }
+    } else {
+      await notifyIssued()?.catch((reportErr) =>
+        logger.error(reportErr, `Failed to report SCEP issuance success [transactionId=${parsed.transactionId}]`)
+      );
+    }
```
sheensantoscapadngan: Can we not just check whether `notify issued` was indeed defined for the provider and call it if it is, and then log an error that's not necessarily specific to Intune so that this applies for others in the future? It feels weird that right now we're building this to be generic because we're using separate providers, but at the same time the calling code itself has references to Intune

backend/src/services/app-connection/app-connection-maps.ts
```diff
@@ -165,5 +166,6 @@ export const APP_CONNECTION_PLAN_MAP: Record<AppConnection, AppConnectionPlanTyp
   [AppConnection.Cloud66]: AppConnectionPlanType.Regular,
   [AppConnection.LiteLLM]: AppConnectionPlanType.Regular,
   [AppConnection.Fireworks]: AppConnectionPlanType.Regular,
-  [AppConnection.NutanixPrismCentral]: AppConnectionPlanType.Regular
+  [AppConnection.NutanixPrismCentral]: AppConnectionPlanType.Regular,
+  [AppConnection.MicrosoftIntune]: AppConnectionPlanType.Regular
```
sheensantoscapadngan: let's make intune an enterprise app connection!

backend/src/services/pki-application/pki-application-enrollment-service.ts
```diff
@@ -631,14 +695,41 @@ export const pkiApplicationEnrollmentServiceFactory = ({
...
+      certificateAuthorityDAL
+    });
+
+    const existingScepConfig = junction.scepConfigId
+      ? await scepEnrollmentConfigDAL.findById(junction.scepConfigId)
+      : null;
+    // Regenerating the RA breaks devices already trusting it, so only do it on first enable or a signing change.
+    const shouldGenerateRaCert =
+      isFirstCreate || (existingScepConfig ? existingScepConfig.signRaWithCa !== signRaWithCa : false);
+    let raCert: Awaited<ReturnType<typeof generateAndEncryptScepRaCertificate>> | null = null;
```
sheensantoscapadngan: In order to simplify, maybe we can make it so that people can't flip the sign RA with CA flag for existing profiles within an application? Like make people either disable SCEP and re-enroll SCEP or remove profile and re-add profile.

docs/integrations/app-connections/microsoft-intune.mdx
```diff
@@ -0,0 +1,87 @@
...
+  - `scep_challenge_provider`
+
+## Setup Microsoft Intune Connection in Infisical
+
+<Tabs>
+  <Tab title="Infisical UI">
+    <Steps>
+      <Step title="Navigate to App Connections">
+        In your **Certificate Manager** project, go to **Settings** and open the **App Connections** tab, then click **Add Connection**.
+      </Step>
```
sheensantoscapadngan, line 56: project mentioned

docs/integrations/app-connections/microsoft-intune.mdx
```diff
@@ -0,0 +1,87 @@
...
+- An App Registration in Azure with a Client Secret and the API permissions below.
+
+#### Required Azure permissions
+
+Set the API permissions of your Azure application to include the following **Application permissions**, then grant admin consent:
+
+- Microsoft Graph
+  - `Application.Read.All`
+- Microsoft Intune
+  - `scep_challenge_provider`
```
sheensantoscapadngan: for this one, maybe you can attach screenshots?

frontend/src/helpers/appConnections.ts
```diff
@@ -74,6 +74,7 @@ import { HerokuConnectionMethod } from "@app/hooks/api/appConnections/types/hero
 import { KempLoadMasterConnectionMethod } from "@app/hooks/api/appConnections/types/kemp-loadmaster-connection";
 import { LaravelForgeConnectionMethod } from "@app/hooks/api/appConnections/types/laravel-forge-connection";
 import { LiteLLMConnectionMethod } from "@app/hooks/api/appConnections/types/litellm-connection";
```
sheensantoscapadngan, line 76: wait is this the Intune logo? should we do Microsoft instead? 

Image

frontend/src/pages/cert-manager/ApplicationDetailsByIDPage/components/ConfigureEnrollmentModal.tsx
```diff
@@ -971,44 +1057,86 @@ const ScepPanel = ({
...
+                  available for internal certificate authorities.
                 </p>
               </div>
-              <Switch
-                variant="project"
-                id="scep-include-ca"
-                checked={field.value}
-                onCheckedChange={field.onChange}
-              />
+              {raCaSigningSupported ? (
```
sheensantoscapadngan: Do we really have to introduce this new API property? Is this not something that we can deduce via the CA type field?

frontend/src/pages/cert-manager/ApplicationDetailsByIDPage/components/ConfigureEnrollmentModal.tsx
```diff
@@ -971,44 +1057,86 @@ const ScepPanel = ({
...
                 variant="project"
-                id="scep-cert-renewal"
+                id="scep-include-ca"
                 checked={field.value}
                 onCheckedChange={field.onChange}
               />
             </div>
           )}
         />
+        {!isIntune && (
```
sheensantoscapadngan, line 1166: In the backend, I think we should be failing when `allowCertBasedRenewal` is true, but then the challenge is Intune (rather than silently falling back to disallowing it), because people won't necessarily be using the UI when configuring SCEP. They can do it via Terraform as well.

backend/src/db/migrations/20260721143512_add-scep-intune-and-ra-ca-signing.ts
```diff
@@ -0,0 +1,45 @@
...
+    );
+    const hasSignRaWithCa = await knex.schema.hasColumn(TableName.PkiScepEnrollmentConfig, "signRaWithCa");
+
+    await knex.schema.alterTable(TableName.PkiScepEnrollmentConfig, (t) => {
+      if (hasValidationConnectionId) {
+        t.dropForeign(["validationConnectionId"]);
+        t.dropIndex("validationConnectionId");
+        t.dropColumn("validationConnectionId");
+      }
+      if (hasSignRaWithCa) {
```
sheensantoscapadngan, line 40: IGNORE PLACEMENT

Let's add a hyperlink or something to point people in the right direction here in terms of where to create Intune app connections... right now it feels like a dead-end 
Image

sheensantoscapadngan: I think it would also be best if we attached a link somewhere here to the documentation so that people understand what they'd have to configure on the Intune side of things and how it all generally connects.

frontend/src/helpers/appConnections.ts
carlosmonastyrski: Changed to another one 👍 (it'll be the best if we don't use the azure one so we avoid misunderstandings)

docs/documentation/platform/pki/applications/enrollment-methods/scep.mdx
```diff
@@ -63,6 +63,18 @@ Infisical supports two challenge authentication modes for SCEP enrollment:
...
     - The password is hashed before storage and cannot be retrieved after creation.
   </Tab>
+  <Tab title="Microsoft Intune">
+    The challenge password is validated directly with Microsoft Intune instead of against a password stored in Infisical. Intune generates a unique challenge for each device it enrolls, and Infisical asks Intune to confirm that the challenge is valid before issuing the certificate.
+
+    Use this mode when devices are managed by Microsoft Intune. It removes the need to distribute a shared password and lets Intune track certificate issuance per device.
+
+    - Requires a **Microsoft Intune** [App Connection](/integrations/app-connections/microsoft-intune) with permission to validate SCEP requests. Microsoft Intune is an enterprise App Connection.
+    - Infisical reports issuance success and failure back to Intune for each request.
+
```
sheensantoscapadngan, line 73: as part of this PR, I think we should also publish a guide specifically for Microsoft Intune and how to configure SCEP end-to-end

Image

backend/src/services/pki-application/pki-application-enrollment-service.ts
```diff
@@ -707,12 +709,20 @@ export const pkiApplicationEnrollmentServiceFactory = ({
...
       certificateAuthorityDAL
     });
 
+    // External CAs always issue asynchronously and Intune has no pending state, so every enrollment
+    // would be reported to Intune as a failure.
+    if (isIntune && caType !== CaType.INTERNAL) {
+      throw new BadRequestError({
+        message: `Microsoft Intune validation requires an internal certificate authority. This profile uses ${CERTIFICATE_AUTHORITIES_TYPE_MAP[caType]}, which issues asynchronously and cannot complete within a SCEP request.`
+      });
+    }
```
sheensantoscapadngan, line 727: It's confusing how CA type is returned by the `resolveSCEPRaSigning` when this can just be something that we can query here in the same level...  I know that at the end of the day the SCEP RA is issued by the exact same Certificate Authority of the profile, but it's not necessarily the case, right, because what if they set Sign RA with CA to be false?

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -17,41 +19,74 @@ import { IScepValidationHandler, TScepValidationHandlerDeps } from "./scep-valid
...
+          kmsService: deps.kmsService
+        })) as TMicrosoftIntuneConnectionCredentials;
+
+        const graphToken = await getMicrosoftEntraToken(credentials, MicrosoftEntraTokenResource.Graph);
+        const serviceUri = await discoverScepValidationServiceUri(graphToken.accessToken);
+        const intuneToken = await getMicrosoftEntraToken(credentials, MicrosoftEntraTokenResource.Intune);
+
+        return { serviceUri, intuneToken: intuneToken.accessToken, expiresAt: intuneToken.expiresAt };
+      }
+    })
```
sheensantoscapadngan: Are we storing this right now as plain text? If so, we should make it encrypted the same way that we cache our secrets right now given that read access to Redis seems pretty common

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+  return Buffer.from(digest).toString("hex").toUpperCase();
+};
+
+export const intuneDelegatedHandler = (deps: TScepValidationHandlerDeps): IScepValidationHandler => {
+  const acquireIntuneAccess = async (connectionId: string): Promise<TIntuneAccess> => {
+    const connection = await deps.appConnectionDAL.findById(connectionId);
+    if (!connection || connection.app !== AppConnection.MicrosoftIntune) {
+      throw new BadRequestError({ message: "The configured Microsoft Intune connection could not be found." });
+    }
+
```
sheensantoscapadngan, line 42: IGNORE PLACEMENT

doesn't seem necessary for users to know about the alternative case...
Image

sheensantoscapadngan:
@greptileai review this once...

sheensantoscapadngan:
@codex review this

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
+import * as x509 from "@peculiar/x509";
+
+import { KeyStorePrefixes } from "@app/keystore/keystore";
+import { withCache } from "@app/lib/cache/with-cache";
+import { BadRequestError } from "@app/lib/errors";
+import { AppConnection } from "@app/services/app-connection/app-connection-enums";
+import { decryptAppConnectionCredentials } from "@app/services/app-connection/app-connection-fns";
```
sheensantoscapadngan, line 7: IGNORE PLACEMENT

maybe you can link access policy here since it might not be too obvious what you're referring to? also... first mention 

Image

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+      const { serviceUri, intuneToken } = await resolveIntune(ctx.validationConnectionId);
+      const result = await intuneValidateScepRequest({
+        intuneAccessToken: intuneToken,
+        serviceUri,
+        transactionId: ctx.transactionId,
+        certificateRequest: ctx.csrDer.toString("base64")
+      });
+
+      return result.allowed ? { allowed: true } : { allowed: false, error: result.errorDescription };
+    },
```
sheensantoscapadngan, line 139: IGNORE PLACEMENT

I recall seeing this somewhere else as well but is it correct to refer to it as a "root" when technically it can be pointed to an intermediate CA? 

Image

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const toThumbprint = async (cert: x509.X509Certificate) => {
+  const digest = await cert.getThumbprint("SHA-1");
+  return Buffer.from(digest).toString("hex").toUpperCase();
+};
+
+export const intuneDelegatedHandler = (deps: TScepValidationHandlerDeps): IScepValidationHandler => {
+  const acquireIntuneAccess = async (connectionId: string): Promise<TIntuneAccess> => {
+    const connection = await deps.appConnectionDAL.findById(connectionId);
+    if (!connection || connection.app !== AppConnection.MicrosoftIntune) {
```
sheensantoscapadngan, line 39: IGNORE PLACEMENT

Image
It's technically allowed if they define the self-signed CA in Intune though right?

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+      keyStore: deps.keyStore,
+      key: cacheKey,
+      ttlSeconds: intuneAccessTtlSeconds,
+      fetcher: async () => {
+        const credentials = (await decryptAppConnectionCredentials({
+          orgId: connection.orgId,
+          projectId: connection.projectId,
+          encryptedCredentials: connection.encryptedCredentials,
+          kmsService: deps.kmsService
+        })) as TMicrosoftIntuneConnectionCredentials;
```
sheensantoscapadngan: IGNORE PLACEMENT

Image

Do they have to know about this? We don't want to overwhelm them with information that they'll be fine without.

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+      }
+    })
+      .then((access) => {
+        for (const key of intuneAccessCache.keys()) {
+          if (key !== cacheKey && key.startsWith(connectionKeyPrefix)) intuneAccessCache.delete(key);
+        }
+        intuneAccessCache.set(cacheKey, access);
+        return access;
+      })
+      .finally(() => intuneAccessInFlight.delete(cacheKey));
```
sheensantoscapadngan: Wouldn't we want to toggle this on by default when Microsoft Intune is selected? For me, it gets disabled when I change it to Microsoft Intune...

Image

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+    return acquisition;
+  };
+
+  const resolveIntune = async (connectionId?: string | null) => {
+    if (!connectionId) {
+      throw new BadRequestError({
+        message: "SCEP enrollment is configured for Microsoft Intune validation but no connection is selected."
+      });
+    }
+
```
sheensantoscapadngan, line 122: IGNORE PLACEMENT

The input is expecting a.cer file, but following your instructions in the doc will actually result in the creation of a .pem file. Some users may be lost here
Image

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+    return acquireIntuneAccess(connectionId);
+  };
+
+  return {
+    supportsPendingIssuance: false,
+    requiresIssuanceNotification: true,
+    validateRequest: async (ctx) => {
```
sheensantoscapadngan, line 129: IGNORE PLACEMENT

What is the deployment channel value here that should be set? I don't think we mentioned it in the doc? 
Image

backend/src/ee/services/pki-scep/challenge/scep-intune-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+        notification: {
+          transactionId: ctx.transactionId,
+          certificateRequest: ctx.csrDer.toString("base64"),
+          certificateThumbprint: await toThumbprint(cert),
+          certificateSerialNumber: cert.serialNumber,
+          certificateExpirationDateUtc: cert.notAfter.toISOString(),
+          issuingCertificateAuthority: cert.issuer
+        }
+      });
+    },
```
sheensantoscapadngan, line 157: IGNORE PLACEMENT OF COMMENT

Were you able to test this out using an intermediate CA because it kind of feels like you only really have to add the root certificate here rather than everything in the chain which is what the docs mention? 

Image

sheensantoscapadngan:
2 bugs:

1. validity not being respected
2. EKUs not being respected as well

docs/documentation/platform/pki/guides/applications/microsoft-intune-scep.mdx
```diff
@@ -65,24 +65,34 @@ The result is a request flow of: device asks Intune for a profile, device sends
...
+
+        Navigate to **Certificate Authorities**, open the CA behind your profile, and in the **CA Certificates** section select **Download CA Certificate**. If that CA is an intermediate, also select **Download CA Certificate Chain** to obtain its issuers.
+
+        <Note>
+          Infisical exports PEM files (`cert.pem` and `chain.pem`). Intune requires DER-encoded `.cer` files, so convert each certificate before uploading it:
 
-        Navigate to **Certificate Authorities**, open the CA behind your profile, and in the **Certificates** section choose **Download CA Certificate**. If your CA is an intermediate, download the full chain and repeat the next step for each certificate in it.
+          ```bash
+          openssl x509 -in cert.pem -outform der -out ca-root.cer
+          ```
```
sheensantoscapadngan: we can just tell folks to change the extension

docs/documentation/platform/pki/guides/applications/microsoft-intune-scep.mdx
```diff
@@ -77,11 +77,7 @@ The request flow is: the device retrieves its profile from Intune, submits the S
         Navigate to **Certificate Authorities**, open the CA behind your profile, and in the **CA Certificates** section select **Download CA Certificate**.
 
         <Note>
-          Infisical exports `cert.pem`. Intune requires a DER-encoded `.cer` file, so convert it before uploading:
-
-          ```bash
-          openssl x509 -in cert.pem -outform der -out ca.cer
-          ```
+          Infisical exports the certificate as `cert.pem`, and Intune's upload field expects a `.cer` extension. Rename it to `cert.cer` before uploading. No conversion is required, because a `.cer` file carries Base-64 encoded X.509 data, which is what `cert.pem` already contains.
```
sheensantoscapadngan: I don't think we have to justify here why there's no conversion. Users don't really have to care about that.

backend/src/services/certificate-common/certificate-issuance-utils.ts
```diff
@@ -412,20 +422,13 @@ export const generateSelfSignedCertificate = async ({
...
-                    throw new BadRequestError({
-                      message: `Unsupported Subject Alternative Name type: ${san.type as string}`
-                    });
+                const generalNameType = CERT_SUBJECT_ALTERNATIVE_NAMES[san.type]?.generalNameType;
+                if (!generalNameType) {
+                  throw new BadRequestError({
+                    message: `Unsupported Subject Alternative Name type: ${san.type as string}`
+                  });
                 }
+                return { type: generalNameType as "dns" | "ip" | "email" | "url" | "upn", value: san.value };
```
sheensantoscapadngan: can we not use an enum here for the casting?

docs/documentation/platform/pki/guides/applications/microsoft-intune-scep.mdx
```diff
@@ -125,34 +125,24 @@ The request flow is: the device retrieves its profile from Intune, submits the S
         | **Root Certificate** | The Trusted certificate profile created above |
         | **SCEP Server URLs** | The **SCEP endpoint URL** copied from Infisical |
 
-        Set the subject name format and subject alternative name according to what your devices require. Intune seals both into its challenge and validates the device's request against them.
+        Set the subject name format and subject alternative name to what your devices require, then complete the remaining fields.
 
-        Not every field on this profile reaches Infisical. Intune transmits the device's request together with a validation result for the challenge, so any value it does not place in that request must be configured on the issuing side:
+        Intune treats the remaining settings in two different ways. Some travel to Infisical inside the device's request, while others are policy that Intune enforces on its own side and never transmits:
```
sheensantoscapadngan: how come Intune never throws an error when the issued certificate does not match the properties defined in the Intune profile?

docs/documentation/platform/pki/guides/applications/microsoft-intune-scep.mdx
```diff
@@ -125,34 +125,24 @@ The request flow is: the device retrieves its profile from Intune, submits the S
...
         </Warning>
 
         ![Intune SCEP certificate profile settings including the Infisical SCEP Server URL](/images/platform/pki/integrations/microsoft-intune/intune-scep-profile-settings.png)
 
-        <Warning>
-          Whatever you choose must be permitted by the [certificate policy](/documentation/platform/pki/settings/policies) behind your profile. Anything the policy disallows causes issuance to fail after Intune has already approved the challenge.
-        </Warning>
+        <Note>
+          Fields vary by platform. Windows profiles add **Hash algorithm** and **Key storage provider**. Apple profiles add **Deployment Channel**, which has to agree with **Certificate type**: Device Channel stores device certificates in the system keychain, User Channel stores user certificates in the user keychain, and under User Channel the certificate type can only be **User**. The channel cannot be changed after the profile is deployed.
```
sheensantoscapadngan: I was able to update the channel from User to Device with no issues... or is that different from Certificate Type?
image

docs/docs.json
```diff
@@ -152,6 +152,7 @@
                       "integrations/app-connections/litellm",
                       "integrations/app-connections/ssh",
                       "integrations/app-connections/mssql",
+                      "integrations/app-connections/microsoft-intune",
```
sheensantoscapadngan, line 155: IGNORE PLACEMENT

Looks like we didn't make any changes to the docs to mention that we already support UPN and the any-purpose extended key usage, etc. . I think we listed out explicitly somewhere, and that list will become outdated now

docs/docs.json
```diff
@@ -898,7 +899,8 @@
                           "documentation/platform/pki/guides/applications/jboss-certbot",
                           "documentation/platform/pki/guides/applications/windows-server-acme",
                           "documentation/platform/pki/guides/applications/gloo-mesh",
-                          "documentation/platform/pki/guides/applications/jamf-pro-scep"
+                          "documentation/platform/pki/guides/applications/jamf-pro-scep",
```
sheensantoscapadngan, line 902: IGNORE PLACEMENT

The UI here feels off because only the extended key usage is misaligned. 
Image

sheensantoscapadngan:
@codex @greptileai review this one more

docs/documentation/platform/pki/guides/applications/microsoft-intune-scep.mdx
carlosmonastyrski: Yup, improved the explanation there

docs/documentation/platform/pki/guides/applications/microsoft-intune-scep.mdx
carlosmonastyrski: CleanShot 2026-07-29 at 15 44 52@2x
"Intended purpose" not part of the CSR

sheensantoscapadngan (dismissed):

sheensantoscapadngan (approved):

PR #7426, opened 2026-07-25
fix(agent-proxy): let user tokens report proxied-service usage

saifsmailbox98:
## Context

The proxied-service `report-usage` endpoint only accepted machine-identity tokens (`verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN])`), while every other proxied-service route accepts both a user JWT and a machine identity. The new local `agent-proxy run` mode brokers with the developer's own user token, so its best-effort usage reports were rejected with `403` and a proxied service's "last used" timestamp never updated for local runs.

- `report-usage` now accepts `AuthMode.JWT` in addition to `IDENTITY_ACCESS_TOKEN`, matching the other proxied-service routes.
- The `ReportUsage` action is granted to the **member** and **viewer** default project roles (admin already had it), so any user who can see a proxied service can report its usage. Reporting only debounce-stamps a "last used" timestamp, so it's safe to allow for read-oriented roles.

Companion CLI change: https://github.com/Infisical/cli/pull/335

## Steps to verify the change

1. As a non-admin (member) user, run `infisical secrets agent-proxy run --env <e> --path <p> -- <cmd>` against a project with an enabled proxied service, and make a brokered request through it.
2. Confirm the service's "last used" updates in the dashboard. Previously the CLI logged a `403` ("You are not allowed to access this resource") from `report-usage` and the timestamp never advanced for local runs.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


eeshsaxena (commented):
The viewer half of the permission change looks wrong to me.

`reportUsage` is a write. It resolves the service, then does `throwUnlessCan(ProjectPermissionProxiedServiceActions.ReportUsage, ...)` and, for the report that claims the debounce key, writes the usage record. `buildViewerPermissionRules` is `Read` on every other subject in that function, so granting `ReportUsage` there means a read-only project member can write usage for any proxied service in the project.

Letting user tokens report usage is the point of the PR and the `AuthMode.JWT` change makes sense for that, but the member grant already covers the humans who are meant to be using the proxy. I would leave viewer on `Read` alone.

akhilmhdh (approved):

PR #7436, opened 2026-07-27
fix: diff API spec against the merge base instead of production

Thiago-AS:
## Context

The breaking API check previously compared every PR against **live production** (`app.infisical.com`). Since production reflects the latest deployment—not the current `main` branch—any legitimate breaking change merged into `main` caused every subsequent PR to fail until the next deployment. As a result, unrelated PRs showed the same failures, and the check gradually lost credibility.

### What changed

- **Baseline is now `HEAD^1`** (the `main` commit the PR is merged onto). Results are now reproducible—re-running the workflow later produces the same verdict.
- Added **`publish-openapi-spec.yml`**, which publishes the OpenAPI spec cache on every push to `main`, keyed by the `backend/` git tree hash. This is required because PR workflows can only restore caches created on the default branch.
- Added a shared composite action: **`generate-openapi-spec`**.
- The check is now **advisory only**. It never fails the workflow. Instead, it:
  - Posts a PR comment when breaking changes are detected.
  - Updates that comment once the breaking changes are resolved.
  - Always publishes the full changelog in the job summary.

Expected steady state:

- **~5 minutes** on a cache hit
- **~10 minutes** on a cache miss

## Steps to verify

This PR triggers the workflow on itself because the `paths` filter now includes both the workflow and the composite action.

1. **Quiet path** (this run)
   - No backend changes, so the base and head specs should be identical.
   - Expected result:
     - ✅ Workflow succeeds
     - ✅ No PR comment
     - ✅ Changelog appears in the job summary
     - First run is a cache miss, so both specs are generated.

2. **Breaking path**
   - Push a commit that adds a required field to any request body under `backend/src/server/routes/`.
   - Expected result:
     - ✅ Workflow still succeeds
     - ✅ PR comment is posted with the breaking-change warning
     - ✅ Baseline spec is restored from cache (cache hit), reducing runtime by roughly half.
     
image

3. **Resolved path**
   - Revert the breaking commit.
   - Expected result:
     - ✅ Existing PR comment is updated to indicate the issue was resolved, including the resolving commit SHA.
     
   image

### Post-merge verification

The following behavior can only be verified after merging:

- `publish-openapi-spec.yml` runs on every push to `main`.
- The next backend PR should restore the baseline spec from cache, logging a **cache hit** and skipping baseline spec generation entirely.

## Type

- [x] fix

## Checklist

- [x] Title follows the Conventional Commits format
- [x] Tested locally
- [ ] Updated documentation (if needed)
- [ ] Updated `CLAUDE.md` files (if needed)

### DIFF
--- a/.github/actions/generate-openapi-spec/action.yml
+++ b/.github/actions/generate-openapi-spec/action.yml
@@ -1,0 +1,109 @@
+name: "Generate OpenAPI spec"
+description: "Builds the backend image from a checkout, boots it, and dumps its OpenAPI spec to a file."
+
+inputs:
+  source-path:
+    description: "Path to the repository checkout to build from."
+    required: false
+    default: "."
+  output:
+    description: "Path to write the normalized OpenAPI spec to."
+    required: true
+  port:
+    description: "Host port to bind the API container to."
+    required: false
+    default: "4000"
+  container-name:
+    description: "Name for the API container."
+    required: false
+    default: "infisical-api"
+
+runs:
+  using: "composite"
+  steps:
+    - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+    - name: Build the API image
+      uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
+      with:
+        context: ${{ inputs.source-path }}/backend
+        load: true
+        tags: ${{ inputs.container-name }}
+        cache-from: type=gha,scope=openapi-spec
+        cache-to: type=gha,mode=max,scope=openapi-spec
+
+    - name: Create the database
+      shell: bash
+      run: |
+        DB_NAME="$(echo '${{ inputs.container-name }}' | tr '-' '_')"
+        echo "OPENAPI_SPEC_DB=$DB_NAME" >> "$GITHUB_ENV"
+
+        # `compose up -d` returns before postgres accepts connections. With a warm
+        # image layer cache the build above no longer absorbs that delay, so wait.
+        for _ in $(seq 1 30); do
+          if docker compose -f docker-compose.dev.yml exec -T db pg_isready -U infisical -q; then
+            break
+          fi
+          sleep 2
+        done
+
+        docker compose -f docker-compose.dev.yml exec -T db \
+          psql -U infisical -d postgres -c "CREATE DATABASE \"$DB_NAME\"" \
+          || echo "Database $DB_NAME already exists, reusing it."
+
+    - name: Boot the API
+      shell: bash
+      run: |
+        docker run --name "${{ inputs.container-name }}" -d -p ${{ inputs.port }}:4000 \
+          -e DB_CONNECTION_URI="postgres://infisical:infisical@172.17.0.1:5432/${OPENAPI_SPEC_DB}?sslmode=disable" \
+          -e REDIS_URL=redis://172.17.0.1:6379 \
+          -e JWT_AUTH_SECRET=something-random \
+          -e ENCRYPTION_KEY=4bnfe4e407b8921c104518903515b218 \
+          -e TELEMETRY_ENABLED=false \
+          -e SECRET_SCANNING_GIT_APP_ID=793712 \
+          -e SECRET_SCANNING_PRIVATE_KEY=some-random \
+          -e SECRET_SCANNING_WEBHOOK_SECRET=some-random \
+          "${{ inputs.container-name }}"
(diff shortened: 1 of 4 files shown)


.github/actions/generate-openapi-spec/action.yml
```diff
@@ -0,0 +1,110 @@
...
+        docker run --name "${{ inputs.container-name }}" -d -p ${{ inputs.port }}:4000 \
+          --add-host=host.docker.internal:host-gateway \
+          -e DB_CONNECTION_URI="postgres://infisical:infisical@host.docker.internal:5432/${OPENAPI_SPEC_DB}?sslmode=disable" \
+          -e REDIS_URL=redis://host.docker.internal:6379 \
+          -e JWT_AUTH_SECRET=something-random \
+          -e ENCRYPTION_KEY=4bnfe4e407b8921c104518903515b218 \
+          -e TELEMETRY_ENABLED=false \
+          -e SECRET_SCANNING_GIT_APP_ID=793712 \
+          -e SECRET_SCANNING_PRIVATE_KEY=some-random \
+          -e SECRET_SCANNING_WEBHOOK_SECRET=some-random \
```
victorvhs017: Do we really need these envs here?

.github/workflows/publish-openapi-spec.yml
victorvhs017, line 1: What happens to the API check on an open PR after we merge another PR? Does the action re-run?

During the time between the first check and the PR being approved, we could have a bunch of other PRs merged, and the API check would be stale, looking at the "old main"

.github/actions/generate-openapi-spec/action.yml
```diff
@@ -0,0 +1,110 @@
...
+    - uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+    - name: Build the API image
+      uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
+      with:
+        context: ${{ inputs.source-path }}/backend
+        load: true
+        tags: ${{ inputs.container-name }}
+        cache-from: type=gha,scope=openapi-spec
+        cache-to: type=gha,mode=max,scope=openapi-spec
```
victorvhs017: We run this on every PR (twice on a miss) and on the main merge. We should only cache on the main merge, because of how GitHub scopes this cache access: A PR can read caches written on the default branch, but caches a PR writes are visible only to that PR

https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache

.github/actions/generate-openapi-spec/action.yml
victorvhs017, line 1: Another reason to avoid caching too much is the 10GB shared quota.

We need to check where we can reuse the cache between all the actions, or this improvement would not perform as well due to all the cache misses from the LRU behavior.

Current cache output:

Image

victorvhs017 (changes requested):

.github/actions/generate-openapi-spec/action.yml
Thiago-AS: not really I think, I just kept the old ones, will remove.

.github/workflows/publish-openapi-spec.yml
Thiago-AS: Nope, a merge to main doesn't trigger anything for open PRs. pull_request only fires on opened, synchronize, and reopened, and a manual re-run just replays the original event, so it doesn't refresh the baseline either. It only updates on the next push to the branch, when HEAD^1 resolves to the current main.

The staleness is in the safe direction. The head spec is built from that same older main, so the diff still isolates this PR. Someone else merging can't make this PR suddenly report breaking changes, which is the property the production baseline didn't have.

The only thing a stale run can miss is an interaction where main changed something this PR also touches and only the combination is breaking. The proper fix is a merge queue. merge_group runs the check against main + PR at merge time, which is exactly the case we're missing. 

I don't think it's worth adopting just for this though. It's a repo-wide workflow change, serializes merges behind full CI runs, and we'd be doing it to close a small gap in an advisory, non-blocking check (for now). The other option is re-running every open PR after every merge to main, which means a full head build per PR for every merge.

I also made the report print the head SHA the baseline was paired with, so an old comment doesn't look like it's reflecting the current state.

victorvhs017: Agreed, I don't think it's easy to happen, even more so without raising a merge conflict after merging into main, which would cause a new commit in the PR, and the check would run again.

victorvhs017 (approved):

PR #7447, opened 2026-07-27
feat(pam): redis account type

lb-vn:
## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/pam-account/pam-account-connection-test.ts
+++ b/backend/src/ee/services/pam-account/pam-account-connection-test.ts
@@ -154,13 +154,14 @@ export const buildGatewayConnectionTest = async (
         sslCertificate?: string;
       };
       const c = creds as { username?: string; password?: string } | null;
+      if (!c) return tcp(host, port);
       return {
         host,
         port,
         request: {
           mode: TestConnectionMode.Redis,
-          username: c?.username,
-          password: c?.password,
+          username: c.username,
+          password: c.password,
           sslEnabled: cd.sslEnabled,
           sslRejectUnauthorized: cd.sslRejectUnauthorized,
           sslCertificate: cd.sslCertificate
--- a/backend/src/ee/services/pam-account/pam-account-connection-test.ts
+++ b/backend/src/ee/services/pam-account/pam-account-connection-test.ts
@@ -12,6 +12,7 @@ import { extractGatewayTarget, isCredentialConfigured, qualifyUsernameWithDomain
 export enum TestConnectionMode {
   SQL = "sql",
   MongoDB = "mongodb",
+  Redis = "redis",
   LDAP = "ldap",
   Kubernetes = "kubernetes",
   SSH = "ssh",
@@ -35,6 +36,14 @@ export type TestConnectionRequest =
       username: string;
       password?: string;
       authSource: string;
+      sslEnabled?: boolean;
+      sslRejectUnauthorized?: boolean;
+      sslCertificate?: string;
+    }
+  | {
+      mode: TestConnectionMode.Redis;
+      username?: string;
+      password?: string;
       sslEnabled?: boolean;
       sslRejectUnauthorized?: boolean;
       sslCertificate?: string;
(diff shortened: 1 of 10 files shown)


backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -310,6 +310,46 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+    connectionDetails: z.object({
+      host: z.string().trim().min(1).max(255),
+      port: z.coerce.number().int().min(1).max(65535),
+      sslEnabled: z.boolean(),
+      sslRejectUnauthorized: z.boolean(),
+      sslCertificate: optionalTrimmedString
+    }),
+    credentials: z.object({
+      username: z.string().trim().min(1).max(256),
+      password: z
```
saifsmailbox98, line 326: What is the intended behavior for passwords? Are we making them required or keeping optional?

Keeping them optional could mean that if the resource is also publicly exposed then someone can access it directly after getting the real host from Infisical. But why would they keep it publicly exposed.

Keeping is optional is probably the right approach, as private resources can't be reached directly but can be configured without password.


Right now there are some inconsistencies:
- UI makes them required
- In API it is optional
- The "Unavailable" badge is shown on the account with "This account can't be used yet: no password is set."
- The launch button is disabled
- But the CLI allows session creation

docs/documentation/platform/pam/accounts/redis.mdx
```diff
@@ -0,0 +1,67 @@
...
+    | **Port** | Redis port (default: 6379) |
+    | **SSL Enabled** | Enable SSL/TLS for the connection |
+    | **Reject Unauthorized** | Reject connections with invalid certificates (only if SSL enabled) |
+    | **SSL Certificate** | Custom CA certificate (only if SSL enabled) |
+  </Step>
+  <Step title="Enter credentials">
+    | Field | Description |
+    |-------|-------------|
+    | **Username** | Redis ACL username. Use `default` for the default user (also works with a legacy `requirepass` password on Redis 6+) |
+    | **Password** | Redis password |
```
saifsmailbox98: Update this if changing the password logic (read the other comment about password first)

docs/cli/commands/pam.mdx
```diff
@@ -90,7 +90,7 @@ infisical pam access <path> [flags]
 Based on the account's template, the CLI starts the appropriate local proxy or credential helper.
 
 <AccordionGroup>
-  <Accordion title="Databases (PostgreSQL, MySQL, SQL Server, MongoDB)">
+  <Accordion title="Databases (PostgreSQL, MySQL, SQL Server, MongoDB, Redis)">
```
saifsmailbox98: Can we have an accordion for Redis and not put it under database?

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -12,6 +12,7 @@ import { extractGatewayTarget, isCredentialConfigured, qualifyUsernameWithDomain
 export enum TestConnectionMode {
   SQL = "sql",
   MongoDB = "mongodb",
+  Redis = "redis",
```
saifsmailbox98, line 15: IGNORE COMMENT PLACEMENT

Will web access be added as part of a separate ticket?

saifsmailbox98 (changes requested):

backend/src/ee/services/pam-account/pam-account-schemas.ts
x032205: making optional

backend/src/ee/services/pam-account/pam-account-connection-test.ts
x032205: Web access wasn't part of the ticket so it wasn't added. I'm not sure if we even need it (at least for now)

saifsmailbox98 (dismissed):

saifsmailbox98 (approved):

PR #7453, opened 2026-07-28
feat: spacelift sync

mathnogueira:
## Context

Adds support for spacelift secret syncs. It supports both Variables and File mounts.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


mathnogueira:
@greptile review again

backend/src/services/secret-sync/spacelift/spacelift-sync-schemas.ts
```diff
@@ -0,0 +1,82 @@
...
+        message: "File path is required when config type is file mount",
+        path: ["mountPath"]
+      });
+    }
+  });
+
+const SpaceliftSyncOptionsConfig: TSyncOptionsConfig = { canImportSecrets: true };
+
+const SpaceliftSyncOptionsSchema = z.object({
+  writeOnly: z.boolean().optional().default(false).describe(SecretSyncs.ADDITIONAL_SYNC_OPTIONS.SPACELIFT.writeOnly)
```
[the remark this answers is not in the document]
mathnogueira: will keep as is, because that's the default option for spacelift.

docs/integrations/app-connections/spacelift.mdx
```diff
@@ -0,0 +1,61 @@
...
+    Open the downloaded file and extract the `api_key_id` and `api_key_secret` values.
+  </Step>
+</Steps>
+
+## Setup Spacelift Connection in Infisical
+
+<Steps>
+  <Step title="Navigate to App Connections">
+    Navigate to the **Integrations** tab in the desired project, then select **App Connections**.
+    ![App Connections Tab](/images/app-connections/general/add-connection.png)
```
adilsitos: It was recently added the revamped image, you can use it here. 

```suggestion
    ![App Connections Tab](/images/app-connections/general/revamped-app-connection-tab.png)
```

docs/integrations/app-connections/spacelift.mdx
```diff
@@ -0,0 +1,61 @@
...
+  </Step>
+  <Step title="Add Connection">
+    Select the **Spacelift Connection** option from the connection options modal.
+    ![Select Spacelift Connection](/images/app-connections/spacelift/spacelift-app-connection-select.png)
+  </Step>
+  <Step title="Input Credentials">
+    Enter your Spacelift API URL, API Key ID, and API Key Secret in the provided fields and
```
adilsitos, line 53: maybe we can have a note saying that the API URL is the base URL you have on your spacelifit account

docs/integrations/secret-syncs/spacelift.mdx
```diff
@@ -0,0 +1,184 @@
...
+        - **Config Type**: Choose how secrets are stored in the Spacelift context:
+            - **Environment Variables**: Each secret is synced as an individual environment variable (default).
+            - **File Mount**: Secrets are mounted as files under `/mnt/workspace/`.
+        - **File Format** *(File Mount only)*: Choose the file format for mounted secrets:
+            - **.env File**: All secrets are bundled into a single `.env` file at the specified path (e.g., `/mnt/workspace/secrets.env`).
+            - **One Secret Per File**: Each secret is stored as a separate file. The filename is the secret key and the content is the secret value (e.g., `/mnt/workspace/secrets/MY_SECRET`).
+        - **File Path** *(.env File only)*: The path for the mounted `.env` file relative to `/mnt/workspace/` (e.g., `secrets.env`). Required.
+        - **Directory Path** *(One Secret Per File only)*: The directory under `/mnt/workspace/` where secret files are created (e.g., `secrets/`). Optional -- leave empty to mount files directly in `/mnt/workspace/`.
+
+        ![Configure Destination](/images/secret-syncs/spacelift/spacelift-sync-destination.png)
```
adilsitos: nit: I think this screenshot is not updated. It doesn't have the config type

backend/src/services/app-connection/spacelift/index.ts
adilsitos, line 1: ignore comment placement 

not sure if it was requested this, but usually when it is not possible to retrieve the value, we force the overwrite to make things simpler. But if this was a request, no problem for me. 

Image

docs/integrations/app-connections/spacelift.mdx
```diff
@@ -0,0 +1,61 @@
...
+  <Step title="Set a Name">
+    Set a name for the API key, choose the type **Secret** and click **Continue**.
+
+    ![Set API Key Name](/images/app-connections/spacelift/spacelift-set-api-key-name.png)
+  </Step>
+  <Step title="Configure Roles, Space, and Create">
+    Select the roles **Space reader** and **Space writer**, choose the Space that will be managed by this credential, then click **Add** and **Create API Key**. A file will be downloaded containing your credentials.
+
+    ![Select Roles](/images/app-connections/spacelift/spacelift-select-roles.png)
+  </Step>
```
adilsitos, line 35: only these roles do not work to update the secrets. To update a context we need to create a custom role and attach the permission to `edit context`. I've attached a video for reference. 

https://github.com/user-attachments/assets/e9ceae49-3732-4f54-b682-a634e0be0e67

For reference, this is the error message I received: 

Image

backend/src/services/secret-sync/spacelift/spacelift-sync-fns.ts
```diff
@@ -0,0 +1,327 @@
...
+    const { apiKeyId, apiKeySecret } = secretSync.connection.credentials;
+    const { contextId, configType, mountPath, fileMountFormat } = secretSync.destinationConfig;
+    const writeOnly = secretSync.syncOptions?.writeOnly ?? false;
+
+    const jwt = await authenticateSpacelift(instanceUrl, apiKeyId, apiKeySecret);
+
+    if (configType === SpaceliftConfigType.FileMount) {
+      const basePath = mountPath ?? "";
+
+      if (fileMountFormat === SpaceliftFileMountFormat.SecretPerFile) {
```
adilsitos, line 335: I feel like this `if` statement here could be a separate method, it will be easier to follow the implementation

backend/src/services/secret-sync/spacelift/spacelift-sync-fns.ts
```diff
@@ -0,0 +1,327 @@
...
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+
+      try {
+        await addContextConfigElement(instanceUrl, jwt, contextId, key, secretMap[key].value, writeOnly);
+      } catch (error) {
+        throw new SecretSyncError({ error, secretKey: key });
+      }
+    }
```
adilsitos: secrets that end with a `\` are supported on infisical, but they are not UNIX valid. We can remove them from the sync process (I would not recommend to remove the `\`, so this can cause more confusion IMO). We should also add a warning when the user is setting the sync, so they know about this limitation 

Image

adilsitos: I also noticed we are inserting each secret one at a time, don't they have a bulk way of doing this?

backend/src/services/secret-sync/spacelift/spacelift-sync-fns.ts
adilsitos: also noticed that this pattern exists in several methods, maybe a refactor can make it easier to understand.

backend/src/services/secret-sync/spacelift/spacelift-sync-fns.ts
adilsitos: question: in the case of `.env` files the error does not happen, but maybe we should remove this as well? 

image

backend/src/services/secret-sync/spacelift/spacelift-sync-fns.ts
```diff
@@ -0,0 +1,327 @@
...
+      const filePath = basePath;
+      const hasSecrets = Object.keys(secretMap).length > 0;
+
+      if (hasSecrets) {
+        const envContent = secretMapToEnvFileContent(secretMap);
+        const encoded = Buffer.from(envContent).toString("base64");
+        await addContextConfigElement(instanceUrl, jwt, contextId, filePath, encoded, writeOnly, "FILE_MOUNT");
+      } else if (!secretSync.syncOptions.disableSecretDeletion) {
+        await deleteContextConfigElement(instanceUrl, jwt, contextId, filePath);
+      }
```
adilsitos: currently we can have secrets that have `=` and `#` so this can break the parsing that happens on the dotenv and confuse break this sync, I think we can also remove the secrets that fall under this so the current parsing approach does not break. 

Example: 
create a secret called: `test=adilson`

Image

create a secret called: `#comment`
sync works, but this can be considered a comment (?)

Image

adilsitos (changes requested):
Some problems with the roles/permissions. Also, I think we need to not consider some secrets during the sync process, otherwise it will fail. Let me know if you have any questions.

docs/integrations/app-connections/spacelift.mdx
mathnogueira: oh I updated it in one place and forgot the other.

adilsitos (approved):

PR #7460, opened 2026-07-28
fix(secret-expansion): resolve nested relative refs in referenced environment

### LINKED ISSUE #7459: backend-go: nested relative secret reference resolves in the requesting environment instead of the referenced one
### Product

Secrets Management

### Describe the bug

In the Go sidecar (`backend-go`), a **relative** secret reference that appears inside a secret reached through an **absolute** reference is resolved against the *requesting* board's environment/path instead of the referenced secret's own environment/path.

Given `dev` and `prod` that both define `HOST`:

| env | key | value |
|---|---|---|
| dev | `HOST` | `dev-host` |
| dev | `API_URL` | `${prod.URL}` |
| prod | `HOST` | `prod-host` |
| prod | `URL` | `https://${HOST}/api` |

Reading `API_URL` from `dev` returns:

- Node (`/api/v4/secrets`): `https://prod-host/api`
- Go sidecar: `https://dev-host/api`

The Go expander never even issues a read for `prod:/:HOST` — it resolves `${HOST}` from its `localLookup` map, which only ever contains secrets from the requested board (plus its imports).

### Root cause

`SecretExpander.expandValue` carries no notion of which environment/path the value it is expanding belongs to:

https://github.com/Infisical/infisical/blob/main/backend-go/internal/services/secretmanager/secret/expansion.go#L203-L218

```go
if len(parts) == 1 {
    // Relative reference: ${KEY}
    found = true // Always replace relative refs (with empty if not found)
    if sec := e.localLookup[parts[0]]; sec != nil {
```

`localLookup` is built once in `NewSecretExpander` from the secrets of the requested board. When the expander recurses into a secret fetched via `FetchAbsoluteSecrets` (a different env/path), single-part references inside that value still hit `localLookup`.

Node does the opposite — it pushes the *referenced* secret's own `environment` / `secretPath` onto the expansion stack, so subsequent single-part refs resolve there:

https://github.com/Infisical/infisical/blob/main/backend/src/services/secret-v2-bridge/secret-reference-fns.ts#L427-L459

```ts
const node = {
  value: referencedSecretValue,
  secretPath: referencedSecretPath,
  environment: referencedSecretEnvironmentSlug,
  ...
};
...
stack.push({ ...node, visitedSecrets: newVisitedSecrets });
```

and the relative branch then reads from that popped context:

```ts
if (entities.length === 1) {
  const [secretKey] = entities;
  const referredValue = await fetchSecret(environment, secretPath, secretKey);
```

### Impact

Both backends serve `/api/v4/secrets`, and the Go sidecar is shadow-compared against Node in `backend/src/server/plugins/go-sidecar-shadowing.ts`, so this shows up as a `value_mismatch`.

Beyond the mismatch, the resolved value is wrong in a way that crosses an environment boundary: the value returned for a `prod` reference is composed from `dev` data. Anyone who can write a secret in `dev` can change what a `${prod....}` reference expands to, as long as the prod secret contains a relative reference. The permission callback (`CanAccessAbsolute`) is not bypassed — prod's own value is still access-checked — but the substituted content silently comes from the lower environment.

It also fails closed in the opposite direction: if the key does not exist on the requesting board, the reference expands to empty rather than reading the correct value from the referenced environment.

### To Reproduce

Add to `backend-go/internal/services/secretmanager/secret/expansion_test.go`:

```go
func TestExpand_NestedRelativeUsesReferencedEnv(t *testing.T) {
	secrets := []*ProcessedSecret{
		newTestSecret("HOST", "dev-host", "dev", "/"),
		newTestSecret("API_URL", "${prod.URL}", "dev", "/"),
	}

	var requested []string
	opts := ExpandOpts{
		FetchAbsoluteSecrets: func(refs []AbsoluteSecretRef) []*ProcessedSecret {
			var out []*ProcessedSecret
			for _, ref := range refs {
				requested = append(requested, ref.CacheKey())
				switch {
				case ref.Env == "prod" && ref.Key == "URL":
					out = append(out, newTestSecret("URL", "https://${HOST}/api", "prod", "/"))
				case ref.Env == "prod" && ref.Key == "HOST":
					out = append(out, newTestSecret("HOST", "prod-host", "prod", "/"))
				}
			}
			return out
		},
	}

	NewSecretExpander(secrets, opts).Expand()

	t.Logf("refs requested: %v", requested)
	assert.Equal(t, "https://prod-host/api", secrets[1].Value)
}
```

Result on `main`:

```
expansion_test.go: refs requested: [prod:/:URL]
    Error:      Not equal:
                expected: "https://prod-host/api"
                actual  : "https://dev-host/api"
```

Note `prod:/:HOST` is never requested.

### Expected behavior

A relative reference inside a secret reached through an absolute reference resolves against that secret's own environment and path, matching Node. In the example above, `dev`'s `API_URL` expands to `https://prod-host/api`.

### Additional context

The existing test `TestExpand_AbsoluteReference_WithNestedRelative` currently asserts the divergent behavior, which is why this has not surfaced — its `FetchAbsoluteSecrets` mock returns the same secret regardless of which refs are requested, so the prod-side key is never exercised.

I have a fix ready and will open a PR shortly.

### Deployment Type

Self-hosted / Cloud (backend-go sidecar)

suvvvv:
## Context

Closes #7459.

In `backend-go`, a relative reference (`${KEY}`) that appears inside a secret reached through an absolute reference (`${env.path.KEY}`) was resolved against the **requesting** board's environment/path rather than the referenced secret's own environment/path.

With `HOST` defined in both `dev` and `prod`, and `prod`'s `URL = https://${HOST}/api`, reading `${prod.URL}` from `dev` returned `https://dev-host/api` instead of `https://prod-host/api`. The expander never issued a read for `prod:/:HOST` at all — `${HOST}` was served from `localLookup`, which only holds secrets from the requested board and its imports.

Node resolves this correctly: `recursivelyExpandSecret` pushes the referenced secret's own `environment`/`secretPath` onto the expansion stack, so nested single-part refs resolve there. Since both backends serve `/api/v4/secrets` and the Go sidecar is shadow-compared against Node in `go-sidecar-shadowing.ts`, this surfaced as a `value_mismatch` — and produced a value for a `prod` reference that was partly composed from `dev` data.

### What changed

- `expandValue` now carries the environment/path that owns the value being expanded, plus an `isLocal` flag. Single-part refs use the import-aware `localLookup` only while expanding secrets from the requested board; inside a secret fetched from elsewhere they resolve as `{ownerEnv, ownerPath, KEY}` and are fetched like any other absolute ref.
- Substitution of a reference is deferred while its resolved value still has unfetched dependencies. Writing a partially expanded value into the parent would erase the environment context needed to resolve what remains — the reference is left in place so the next pass can resolve it with the full context.
- `Expand()` ends with a final pass that substitutes whatever is left, so missing or permission-denied refs still collapse to an empty string exactly as before.

`TestExpand_AbsoluteReference_WithNestedRelative` asserted the old behavior and has been updated. Its `FetchAbsoluteSecrets` mock returned the same secret regardless of which refs were requested, so the prod-side key was never exercised — that is why the bug went unnoticed. The mock now responds per-ref, and the test asserts both the expanded value and that `prod:/:LOCAL_VAR` was actually requested.

## Steps to verify the change

```bash
cd backend-go
make test-unit   # all packages pass
make lint        # 0 issues
```

To see the old behavior, check out `main` and run the repro in the linked issue — it returns `https://dev-host/api` and never requests `prod:/:HOST`.

New/updated tests:

- `TestExpand_AbsoluteReference_WithNestedRelative` — nested relative ref resolves in the referenced env; prod's `LOCAL_VAR` wins over the same-named key in `dev`.
- `TestExpand_AbsoluteReference_NestedRelativeNoLocalFallback` — when the key is absent in the referenced env, it expands to empty rather than silently falling back to the requesting board.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

## Note on a related mismatch

While tracing this I noticed a second, separate Node/Go divergence: Node was deliberately changed in `1e3a28c71f` ("keep literal `${ref}` when referenced secret does not exist"), while Go still substitutes an empty string for missing refs. I kept this PR to the environment-resolution bug and left Go's existing missing-ref behavior untouched. Happy to follow up with that in a separate PR if you'd like it aligned.

### DIFF
--- a/backend-go/internal/services/secretmanager/secret/expansion.go
+++ b/backend-go/internal/services/secretmanager/secret/expansion.go
@@ -45,6 +45,7 @@ type SecretExpander struct {
 	fullyExpanded  []bool
 	deniedRefs     map[string]struct{} // cacheKey -> denied by permission check
 	requestedRefs  map[string]struct{} // cacheKey -> already requested (avoid infinite loop)
+	deferred       bool                // a substitution was postponed pending a fetch
 }
 
 // NewSecretExpander creates an expander from priority-ordered secrets.
@@ -120,12 +121,13 @@ func (e *SecretExpander) Expand() {
 		}
 	}
 
-	// Everything fetchable has been fetched. Substitute what remains, including
-	// references that were deferred while their dependencies were still pending.
-	for i := range e.fullyExpanded {
-		e.fullyExpanded[i] = false
+	// A reference is left unsubstituted only when its resolved value still had
+	// unfetched dependencies, so the final substitution pass is needed only in
+	// that case. Running it unconditionally would hand every value a second
+	// depth budget and defeat maxExpansionDepth.
+	if e.deferred {
+		e.expandPass(true)
 	}
-	e.expandPass(true)
 
 	e.replaceDeniedRefs()
 	e.replaceNotFoundRefs()
@@ -292,6 +294,7 @@ func (e *SecretExpander) expandValue(
 		// resolve it with the correct environment context; substituting a
 		// partially expanded value here would strip that context away.
 		if pending && !final {
+			e.deferred = true
 			continue
 		}
 
--- a/backend-go/internal/services/secretmanager/secret/expansion.go
+++ b/backend-go/internal/services/secretmanager/secret/expansion.go
@@ -74,7 +74,7 @@ func NewSecretExpander(secrets []*ProcessedSecret, opts ExpandOpts) *SecretExpan
 // It iteratively expands references, fetching absolute refs as needed.
 func (e *SecretExpander) Expand() {
 	for {
-		needed := e.expandPass()
+		needed := e.expandPass(false)
 		if len(needed) == 0 {
 			break
 		}
@@ -120,6 +120,13 @@ func (e *SecretExpander) Expand() {
 		}
 	}
 
+	// Everything fetchable has been fetched. Substitute what remains, including
+	// references that were deferred while their dependencies were still pending.
+	for i := range e.fullyExpanded {
+		e.fullyExpanded[i] = false
+	}
+	e.expandPass(true)
+
 	e.replaceDeniedRefs()
 	e.replaceNotFoundRefs()
 }
@@ -141,7 +148,12 @@ func (e *SecretExpander) HasDeniedRefs() bool {
 
 // expandPass expands all references in current values.
 // Returns list of absolute refs that are needed but not yet available.
-func (e *SecretExpander) expandPass() []AbsoluteSecretRef {
+//
+// When final is false, a reference whose resolved value still depends on
+// un-fetched secrets is left untouched so the next pass can re-resolve it with
+// the full environment context. When final is true, anything still unresolved
+// (missing or permission-denied) collapses to an empty string.
+func (e *SecretExpander) expandPass(final bool) []AbsoluteSecretRef {
 	neededRefs := make(map[string]AbsoluteSecretRef)
 
 	for i, sec := range e.secrets {
@@ -149,7 +161,7 @@ func (e *SecretExpander) expandPass() []AbsoluteSecretRef {
 			continue
 		}
 
-		expanded, needed := e.expandValue(sec.Value, make(map[string]struct{}), 0)
+		expanded, needed := e.expandValue(sec.Value, sec.Environment, sec.SecretPath, true, final, make(map[string]struct{}), 0)
 		sec.Value = expanded
 
 		if len(needed) == 0 && !hasReferences(expanded) {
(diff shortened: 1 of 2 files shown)


suvvvv:
Good catch on the `maxExpansionDepth` regression — confirmed and fixed in 95d79b2.

I probed it with a 25-link local chain (`K00 -> K01 -> ... -> K24`), comparing `main` against the PR:

| | `K00` expands to | levels |
|---|---|---|
| `main` | `L00-...-L11-${K12}` | 12 |
| PR (before fix) | `L00-...-L23-END` | 24 |

Exactly as described: the final pass cleared `fullyExpanded` for every secret and re-ran expansion, so a value that had already hit the cap got a second depth budget and its chain kept going.

The fix narrows the pass to what it was actually for. It exists only to substitute references that were postponed while their resolved value still had unfetched dependencies, so it now runs only when a substitution was in fact deferred (`e.deferred`). A purely local chain never defers, so it takes exactly one pass and is bounded as before — the probe now returns `L00-...-L11-${K12}`, byte-identical to `main`.

Clearing `fullyExpanded` also turned out to be unnecessary and I dropped it: a secret is marked fully expanded only when no references remain (`len(needed) == 0 && !hasReferences(expanded)`), so it has nothing left to substitute. Anything deferred still has references and is therefore never marked, which means it is picked up by the final pass anyway.

Added `TestExpand_MaxDepth_NotExtendedByLaterPasses` asserting the exact capped output, since the existing `TestExpand_MaxDepth` only checked `Contains(value, "levelA-")` and passed either way.

`make test-unit` and `make lint` (0 issues) both pass.

PR #7468, opened 2026-07-29
fix(ui): normalize iOS input font sizing

andrewhuhh:
## Context

On narrow viewports, the shared v3 `Input` and `TextArea` currently switch from `text-sm` to `text-base` so iOS Safari does not zoom when a control receives focus. Because that exception lives inside each component and applies to every mobile browser, the controls can appear larger than surrounding typography even when Safari zoom prevention is not needed.

This change makes the behavior systematic:

- `Input` and `TextArea` now use `text-sm` at every viewport size, with no component-level mobile font override.
- The global Tailwind theme explicitly defines the normal `text-xs`, `text-sm`, and `text-base` scale as 12px, 14px, and 16px.
- Application bootstrap detects iOS and iPadOS, including iPads presenting a desktop-class user agent, and marks the document with `data-platform="ios"`.
- On detected iOS devices, the same global scale shifts to 14px, 16px, and 18px. This keeps form controls at Safari's 16px focus threshold while preserving their relationship to labels, descriptions, and surrounding text.

Non-iOS typography and desktop component behavior remain unchanged.

Linear: [DES-59](https://linear.app/infisical/issue/DES-59/normalize-mobile-input-font-resizing)

## Screenshots

Visual evidence has not been captured yet.

Outstanding capture requests:

CleanShot 2026-07-29 at 09 38 34@2x
Infisical  Networking
(no zoom on iphone)

## Steps to verify the change

1. In desktop Chromium, open the `Generic/Input` and `Generic/TextArea` Storybook stories and confirm both controls compute to `text-sm` (14px).
2. Confirm the document root does not receive `data-platform="ios"` and `text-xs`, `text-sm`, and `text-base` compute to 12px, 14px, and 16px.
3. On iOS Safari, open the same stories and confirm the document root receives `data-platform="ios"`.
4. Confirm `text-xs`, `text-sm`, and `text-base` compute to 14px, 16px, and 18px on iOS.
5. Focus the input and textarea and confirm Safari does not zoom the viewport.
6. Exercise default, disabled, error, prefilled, and autofill states, plus textarea growth.
7. Confirm desktop Chromium and non-iOS Safari retain their previous typography and control behavior.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -235,6 +235,9 @@
   --color-blue-900: #00367d;
 }
 
+/* iOS Safari zooms the viewport when focusing an input whose font-size is below 16px.
+   Bumping the type scale on detected iOS devices keeps form controls at/above that
+   threshold to prevent the zoom while preserving their relationship to surrounding text. */
 html[data-platform="ios"] {
   --text-xs: 0.875rem;
   --text-sm: 1rem;
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -66,6 +66,9 @@
   --breakpoint-dashboard: 1400px;
 
   /* Fonts */
+  --text-xs: 0.75rem;
+  --text-sm: 0.875rem;
+  --text-base: 1rem;
   --font-alliance:
     "Alliance No. 2", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial,
     sans-serif;
@@ -230,6 +233,12 @@
   --color-blue-700: #0053bf;
   --color-blue-800: #004299;
   --color-blue-900: #00367d;
+}
+
+html[data-platform="ios"] {
+  --text-xs: 0.875rem;
+  --text-sm: 1rem;
+  --text-base: 1.125rem;
 }
 
 @utility min-table-row {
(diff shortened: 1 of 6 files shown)


andrewhuhh:
@greptile review

frontend/src/lib/fn/platform.ts
```diff
@@ -0,0 +1,9 @@
+const isIos =
+  /iPad|iPhone|iPod/.test(navigator.userAgent) ||
+  (navigator.userAgent.includes("Macintosh") && navigator.maxTouchPoints > 1);
```
scott-ray-wilson: can we move this into `initializePlatform` or add optional chaining as it is accessing navigator which doesn't exist in every context; not sure if its problematic now but good practice for future proofing

frontend/src/index.css
```diff
@@ -232,6 +235,12 @@
   --color-blue-900: #00367d;
 }
 
+html[data-platform="ios"] {
```
scott-ray-wilson, line 241: nit lets add a short comment explaining why we do this for future maintainers

scott-ray-wilson (changes requested):

frontend/src/lib/fn/platform.ts
andrewhuhh: ooo

scott-ray-wilson (approved):

PR #7470, opened 2026-07-29
improvement(auth): improve verification code input

andrewhuhh:
## Context

Improves the verification-code experience for onboarding and the other flows that share the same input.

Previously, the code was represented by six independent inputs. That made selection, replacement, partial paste, keyboard navigation, browser autofill, and accessibility behave differently from a normal text field. A completed code also required an explicit first Verify click, while the Verify button appeared and disappeared with field completeness.

This change:

- replaces `react-code-input` with one native input presented as segmented code cells;
- preserves normal input behavior for typing, Backspace, partial and complete paste, `Cmd/Ctrl + A`, arrow keys, Home/End, pointer selection, and one-time-code autofill;
- makes clicking a populated cell select its digit and renders a visible caret for collapsed selections;
- uses one state-aware outline spanning the complete selected range instead of native text-selection coloring;
- automatically submits only the first complete code entry;
- keeps a Retry action visible after the first attempt, disabled until the code is complete and enabled for subsequent manual verification;
- selects the complete code after a failed verification so it can be replaced immediately; and
- removes the obsolete dependency, global styles, and per-consumer submit-label overrides.

Related work: [PLATFOR-591](https://linear.app/infisical/issue/PLATFOR-591/improve-the-onboarding-token-input) and [PLATFOR-592](https://linear.app/infisical/issue/PLATFOR-592/make-the-token-space-automatically-submit-a-filled-token).

## Screenshots

CleanShot 2026-07-30 at 12 31 29@2x
CleanShot 2026-07-30 at 12 32 00@2x

## Steps to verify the change

1. Open the email verification step and type a six-digit code; confirm focus advances visually and the first completed fill submits once automatically.
2. Delete and re-enter digits after that first attempt; confirm no further automatic submissions occur and Retry remains visible.
3. Confirm Retry is disabled for an incomplete code, enabled for a complete code, and pending state prevents duplicate submission.
4. Paste a partial or complete code at different caret positions and confirm the digits replace/distribute like one input.
5. Use `Cmd/Ctrl + A`, typing, Backspace, arrows, Shift + arrows, Home, and End; confirm selection and replacement follow native input behavior.
6. Click populated and empty cells and drag across adjacent cells; confirm populated cells select their digit and one continuous outline spans the selected range.
7. Submit an invalid code; confirm focus stays in the field, all digits are selected, the outline uses danger styling, and typing replaces the complete code.
8. Verify six-digit codes use `inputmode="numeric"` and `autocomplete="one-time-code"`; verify eight-character recovery codes retain text input behavior.
9. Repeat representative checks in signup, MFA verification/setup, and the personal-settings email-change flow.
10. Check the segmented layout and selection outline at narrow and standard viewports.

Validation completed:

- `git diff --check`
- `jq empty frontend/package.json frontend/package-lock.json`
- private-file verification
- local browser checks for typing, paste, selection/replacement, pointer cell selection, drag selection, caret placement, first-fill auto-submit, persistent Retry behavior, and invalid-code recovery before the final shared-outline adjustment

Remaining verification gap: the final shared-outline geometry and responsive/error visuals have not been rerun because switching the shared preview from PLATFOR-547 requires explicit migration-risk approval. No package-manager, full frontend lint, or full frontend type-check command was run.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


andrewhuhh:
@greptile review

andrewhuhh:
@greptile review

backend/src/services/auth-token/auth-token-service.test.ts
scott-ray-wilson, line 1: ask @akhilmhdh about this change; even if the error code is more correct this is a breaking api change. In this case I don't imagine there would be much impact but still should verify with him

frontend/src/components/v3/platform/VerificationCode/VerificationCodeInput.tsx
scott-ray-wilson, line 1: nit but I'm not sure we should flash white on the outline when moving between fields? it's so fast it is kind of distracting

scott-ray-wilson, line 1: looks a little weird in the email update page; also shows a button but I never have the opportunity to press it

https://github.com/user-attachments/assets/ffb1d387-8651-4e1d-a0dc-ed7cacdd6174

scott-ray-wilson, line 1: same here on enable mfa 

https://github.com/user-attachments/assets/d55056a8-03a0-49ba-81c4-e11ea54e3239

scott-ray-wilson (dismissed):

frontend/src/components/v3/platform/VerificationCode/VerificationCodeInput.tsx
andrewhuhh: tbh this is more a problem with the call site that the input, will change the call site

backend/src/services/auth-token/auth-token-service.test.ts
andrewhuhh: can revert the contract here, since the count is now clientside. forgot to revert this 😮‍💨

scott-ray-wilson (approved):
lgtm

PR #7476, opened 2026-07-29
fix(secret-manager): resolve nested relative refs in the referenced environment

### LINKED ISSUE #7459: backend-go: nested relative secret reference resolves in the requesting environment instead of the referenced one

Manibharadwaj:
## Context

Fixes #7459.

In `backend-go`, a **relative** reference (`${KEY}`) inside a value reached through an **absolute** reference (`${env.path.KEY}`) was resolved against the *requesting* board instead of the referenced one.

`localLookup` is built once in `NewSecretExpander` from the requested board's secrets, and `expandValue` carried no notion of which board the value it was expanding belonged to — so single-part references inside a fetched foreign value still hit `localLookup`.

Given `dev` and `prod` that both define `HOST`:

| env | key | value |
|---|---|---|
| dev | `HOST` | `dev-host` |
| dev | `API_URL` | `${prod.URL}` |
| prod | `HOST` | `prod-host` |
| prod | `URL` | `https://${HOST}/api` |

Reading `API_URL` from `dev`:

- Node (`/api/v4/secrets`): `https://prod-host/api`
- Go sidecar before this PR: `https://dev-host/api` — and `prod:/:HOST` was never even requested

Beyond the `value_mismatch` that `go-sidecar-shadowing.ts` reports, the resolved value crossed an environment boundary: anyone who can write a secret in `dev` could change what a `${prod....}` reference expands to, as long as the prod secret contained a relative reference. `CanAccessAbsolute` was not bypassed — prod's own value was still access-checked — but the substituted content silently came from the lower environment.

### The fix

`expandValue` now carries an `expansionCtx` identifying the board a value belongs to. This mirrors what the Node implementation already does in `secret-reference-fns.ts` — it pushes the referenced secret's own `environment` / `secretPath` onto the expansion stack, and the relative branch reads from that popped context. Relative references in a foreign context resolve through `absoluteLookup` for that env/path, requesting the value if it hasn't been fetched yet.

Two supporting changes make that correct and terminating:

1. **Substitution of a foreign value is deferred while it still has pending references.** Otherwise a partially expanded value gets inlined into the requesting board, where its remaining relative references would resolve against the wrong environment on the next pass — reintroducing the same bug one level up.
2. **Absolute refs that were fetched and turned out to be missing or denied now resolve to empty during expansion**, rather than only in the final `replaceNotFoundRefs` / `replaceDeniedRefs` cleanup pass. Without this, a permanently-unresolvable inner ref would keep the deferral in (1) pending forever and leave a literal `${prod.URL}` in the output. Final output for those cases is unchanged; the cleanup passes are retained as a safety net.

The local-board branch is untouched, so import priority ordering (`localLookup`, first-occurrence-wins) behaves exactly as before.

### Note for reviewers

I changed one existing test, `TestExpand_AbsoluteReference_WithNestedRelative`. It asserted `prefix-local-resolved-suffix` — that is, it encoded the bug, asserting that a `dev` value leaks into a `prod` reference. Its `FetchAbsoluteSecrets` stub also ignored the `refs` argument and returned the same secret for every request, which hid the fact that the second fetch was never issued. The stub is now ref-aware and the assertion reflects the referenced environment's value.

One deliberate divergence from Node that this PR does **not** change: Node leaves a literal `${REF}` when a referenced secret does not exist, while Go substitutes an empty string. That is pre-existing Go behavior and out of scope here.

## Steps to verify the change

```
cd backend-go
go test -race ./internal/services/secretmanager/...
make lint
```

The repro from the issue is covered by `TestExpand_NestedRelativeResolvesInReferencedEnv`, which also asserts that `prod:/:HOST` is actually requested.

Verification performed:

| Check | Result |
|---|---|
| Issue's exact repro | passes; `prod:/:HOST` now requested |
| 5 of the new tests reverted against `main` | all 5 fail — they are genuine regression tests |
| `go test -race ./internal/...` (cold cache) | all packages pass |
| `golangci-lint run --build-tags integration` (cold cache) | 0 issues |
| `gofmt -l`, `go vet` | clean |

New test coverage:

- `TestExpand_NestedRelativeResolvesInReferencedEnv` — the reported case
- `TestExpand_NestedRelativeInReferencedPath` — same, across a non-root secret path
- `TestExpand_NestedRelativeMissingInReferencedEnv` — must not fall back to the requesting env's value
- `TestExpand_NestedRelativeDeniedInReferencedEnv` — permission denial in the referenced env is tracked and resolves empty
- `TestExpand_NestedRelativeCycleAcrossEnvs` — self-cycle reached through an absolute ref
- `TestExpand_NestedRelativeMutualCycleAcrossEnvs` — mutual `prod ↔ staging` recursion, with a fetch-count guard proving expansion converges

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally
- [ ] Updated docs (if needed) — not needed, no user-facing behavior documented changes
- [ ] Updated CLAUDE.md files (if needed) — not needed, no architectural change

### DIFF
--- a/backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
+++ b/backend-go/internal/server/api/secretmanager/secret/get_secret_by_name.go
@@ -164,6 +164,8 @@ func (h *Handler) getSecretByName(ctx context.Context, opts *getSecretByNameInte
 		}
 
 		expander := secretsvc.NewSecretExpander(allSecrets, secretsvc.ExpandOpts{
+			BoardEnv:  opts.Environment,
+			BoardPath: opts.SecretPath,
 			CanAccessAbsolute: func(ref secretsvc.AbsoluteSecretRef, tags []string) bool {
 				return checker.CanReadSecretValue(ref.Env, ref.Path, ref.Key, tags)
 			},
--- a/backend-go/internal/server/api/secretmanager/secret/list_secrets.go
+++ b/backend-go/internal/server/api/secretmanager/secret/list_secrets.go
@@ -102,6 +102,8 @@ func (h *Handler) listSecrets(ctx context.Context, opts *listSecretsInternalOpts
 		}
 
 		expander := secretsvc.NewSecretExpander(filteredPool, secretsvc.ExpandOpts{
+			BoardEnv:  opts.Environment,
+			BoardPath: opts.SecretPath,
 			CanAccessAbsolute: func(ref secretsvc.AbsoluteSecretRef, tags []string) bool {
 				return checker.CanReadSecretValue(ref.Env, ref.Path, ref.Key, tags)
 			},
--- a/backend-go/internal/services/secretmanager/secret/expansion.go
+++ b/backend-go/internal/services/secretmanager/secret/expansion.go
@@ -33,6 +33,18 @@ type ExpandOpts struct {
 
 const maxExpansionDepth = 10
 
+// expansionCtx identifies which board a value being expanded belongs to.
+// Relative references (${KEY}) resolve against this context, so a value pulled
+// in through an absolute reference resolves its own relative references in the
+// referenced environment/path rather than in the requesting board.
+type expansionCtx struct {
+	// local marks the requested board (plus its imports), whose secrets are
+	// resolved through localLookup with import priority ordering.
+	local bool
+	env   string
+	path  string
+}
+
 var interpolationRegex = regexp.MustCompile(`\${([a-zA-Z0-9-_.]+)}`)
 
 // SecretExpander handles secret reference expansion.
@@ -149,7 +161,7 @@ func (e *SecretExpander) expandPass() []AbsoluteSecretRef {
 			continue
 		}
 
-		expanded, needed := e.expandValue(sec.Value, make(map[string]struct{}), 0)
+		expanded, needed := e.expandValue(sec.Value, expansionCtx{local: true}, make(map[string]struct{}), 0)
 		sec.Value = expanded
 
 		if len(needed) == 0 && !hasReferences(expanded) {
@@ -179,7 +191,7 @@ func (e *SecretExpander) expandPass() []AbsoluteSecretRef {
 // This happens because the outer ${A} resolves to the inner expansion result,
 // where the cycle was detected and replaced with empty string. This is intentional
 // behavior matching the Node.js implementation.
-func (e *SecretExpander) expandValue(value string, visited map[string]struct{}, depth int) (string, []AbsoluteSecretRef) {
+func (e *SecretExpander) expandValue(value string, ctx expansionCtx, visited map[string]struct{}, depth int) (string, []AbsoluteSecretRef) {
 	if depth > maxExpansionDepth {
 		return value, nil
 	}
@@ -200,15 +212,16 @@ func (e *SecretExpander) expandValue(value string, visited map[string]struct{}, 
 		var secretID string
 		var found bool
 
-		if len(parts) == 1 {
-			// Relative reference: ${KEY}
+		switch {
+		case len(parts) == 1 && ctx.local:
+			// Relative reference on the requested board: ${KEY}
 			found = true // Always replace relative refs (with empty if not found)
 			if sec := e.localLookup[parts[0]]; sec != nil {
 				secretID = sec.Environment + ":" + sec.SecretPath + ":" + sec.Secret.Key
 
 				if _, cyclic := visited[secretID]; !cyclic {
 					visited[secretID] = struct{}{}
-					expandedValue, moreNeeded := e.expandValue(sec.Value, visited, depth+1)
+					expandedValue, moreNeeded := e.expandValue(sec.Value, ctx, visited, depth+1)
 					delete(visited, secretID)
 					resolvedValue = expandedValue
 					neededRefs = append(neededRefs, moreNeeded...)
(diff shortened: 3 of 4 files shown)


Manibharadwaj:
Thanks both — I looked into each finding rather than taking them at face value. One was a real regression, one I could not reproduce but hardened anyway. Both addressed in 170e1d6.

### Codex P2 (import fallback for self-absolute refs) — confirmed, real regression

Reproduced exactly as described. `dev /` importing `staging /`, `ROOT = ${dev.B}`, `dev.B = x-${IMPORT_ONLY}`:

- on `main`: `x-from-import`
- with the first commit: `x-` ❌

Treating every fetched absolute secret as foreign dropped the board's import fallback whenever the absolute reference pointed back at the requested board itself.

Fixed by giving `ExpandOpts` the requested board's identity (`BoardEnv` / `BoardPath`, wired at both call sites from `opts.Environment` / `opts.SecretPath`). `contextFor` returns the **local** context when a fetched secret resolves back to that board, so `localLookup`'s import fallback and priority ordering still apply. Covered by `TestExpand_SelfAbsoluteRefKeepsImportFallback`, which fails without the fix.

Worth noting for context: Node's relative branch always goes through `fetchSecret(environment, secretPath, key)`, which reads the folder via `findByFolderId` with no import fallback — so Go and Node already differ here at the top level. I've kept Go's existing import semantics rather than changing them in a bugfix PR.

### Greptile P1 (depth limit loses foreign context) — could not reproduce, hardened regardless

I tried four escalating constructions to get a foreign chain past `maxExpansionDepth` to leak a requesting-environment value: a >10-deep foreign chain; the same with the tail key also defined locally; an independent longer chain to force extra passes; and a sibling absolute ref in the same value to force a pass *after* the depth-limited collapse. In every case the result was a literal `${K11}`, never the local value.

The reason is the deferral this PR adds: a foreign value is never partially inlined while it still has pending references, so the depth-limited collapse always lands on the terminal round, with nothing left to trigger a subsequent local pass. Round count is bounded by chain depth, so no chain can outlive the depth cap to produce that extra pass.

That said, the residual literal `${KEY}` was ugly and the invariant was implicit rather than enforced. The depth guard now blanks relative references in a foreign context (absolute refs are left alone, since they carry their own context), so the class is closed structurally instead of resting on that reasoning. `TestExpand_DepthLimitDoesNotLeakForeignRelativeRefs` asserts both that no local value appears and that no unresolved reference escapes to the requesting board.

### Verification

`go test -race ./internal/...` (cold cache) all green, `golangci-lint run --build-tags integration` 0 issues, `gofmt -l` and `go vet` clean.

Manibharadwaj:
Hi @akhilmhdh — is this something you'd want?

It fixes nested relative references not resolving across environments, but it is a fairly large change to the reference-resolution path, and I would rather not have it sit here if the approach isn't the one you'd take.

Happy to any of the following, whichever is most useful:

- split it into a smaller, more targeted fix
- rework it if you'd prefer a different approach
- close it if this is something you'd rather handle internally

No rush, just did not want it sitting open indefinitely if it isn't wanted.

PR #7480, opened 2026-07-29
fix: replace silent 200 returns with proper errors in org membership DELETE endpoints

### LINKED ISSUE #7478: DELETE /api/v2/organizations/{organizationId}/memberships/{membershipId} returns 200 but membership is not deleted (self-hosted v0.162.14)
### Product

### Describe the bug

# DELETE organization membership API returns 200 OK but does not remove membership

## Description

The DELETE API endpoint for removing an organization membership returns `HTTP 200 OK`, but the membership is not actually removed.

After the successful DELETE request, the user is still present in the organization members list.

## Environment

- Deployment: Self-hosted Docker Compose deployment
- Infisical image: `infisical/infisical:latest`
- Infisical version: `v0.162.14`

Docker image information:
INFISICAL_PLATFORM_VERSION=v0.162.14
NODE_VERSION=22.22.0

## API endpoint

DELETE /api/v2/organizations/{organizationId}/memberships/{membershipId}

Response

The API returns:

HTTP/1.1 200 OK

Response body:

(empty)

PowerShell output:

StatusCode        : 200
StatusDescription : OK
Content           : {}
RawContent        : HTTP/1.1 200 OK

## Request example

```powershell
Invoke-WebRequest `
    -Uri "https://192.168.71.88/api/v2/organizations/{organizationId}/memberships/{membershipId}" `
    -SkipCertificateCheck `
    -Method Delete `
    -Headers @{
        "Authorization" = "Bearer <access_token>"
    }

### Expected behavior

Actual behavior

After the DELETE request completes successfully, the membership still exists.

The user is still returned by:

GET /api/v1/organizations/{organizationId}/members

with the same membership ID.

Expected behavior

The membership should be removed from the organization and should no longer appear in the organization members list.

### Screenshots

_No response_

### Deployment Type

Self-hosted

### Additional context

Additional information
The same API instance successfully handles other organization/project membership operations.
The request is sent through Traefik reverse proxy, but the backend logs confirm that the request reaches the application.
The issue appears to affect only the DELETE membership operation.

Could you please verify if this endpoint is working correctly in self-hosted deployments on version v0.162.14?

antan2002:
## Description

Fixes two endpoints returning HTTP 200 with empty body instead of a proper error when the deletion can't proceed:

- `DELETE /:organizationId/memberships/:membershipId`
- `DELETE /:organizationId/memberships` (bulk)

## Root Cause

1. **Auth guard silent return**: Both endpoints accept `API_KEY` and `IDENTITY_ACCESS_TOKEN` auth modes in `verifyAuth`, but the handler silently returns `undefined` (→ empty 200) when `req.auth.actor !== ActorType.USER`. API-key-authenticated requests succeed with a 200 but no deletion occurs.

2. **No-op on mismatched data**: If the `membershipId` doesn't exist or doesn't belong to the org, the DAL's DELETE affects 0 rows. The handler destructures `[undefined]` from the empty array and returns a response missing required `OrgMembershipsSchema` fields (`id`, `scope`, etc.), which gets silently stripped.

## Changes

| File | Change |
|------|--------|
| `organization-router.ts` (`deleteOrgMembership`) | Replace silent `return` with `BadRequestError`; add `NotFoundError` null-check after deletion |
| `organization-router.ts` (`bulkDeleteOrgMemberships`) | Replace silent `return` with `BadRequestError` |

## Testing

- [x] JWT user: delete works as before
- [x] API KEY / IDENTITY token: now returns 400 with `BadRequestError` instead of empty 200
- [x] Invalid membership ID: now returns 404 with `NotFoundError` instead of empty 200

Closes #7478

### DIFF
--- a/backend/src/server/routes/v2/organization-router.ts
+++ b/backend/src/server/routes/v2/organization-router.ts
@@ -8,6 +8,7 @@ import { addAuthOriginDomainCookie } from "@app/server/lib/cookie";
 import { GenericResourceNameSchema } from "@app/server/lib/schemas";
 import { getTelemetryDistinctId } from "@app/server/lib/telemetry";
 import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
+import { BadRequestError, NotFoundError } from "@app/lib/errors";
 import { ActorType, AuthMode } from "@app/services/auth/auth-type";
 import { sanitizedOrganizationSchema } from "@app/services/org/org-schema";
 import { PostHogEventTypes } from "@app/services/telemetry/telemetry-types";
@@ -262,7 +263,7 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
     },
     onRequest: verifyAuth([AuthMode.JWT, AuthMode.API_KEY, AuthMode.IDENTITY_ACCESS_TOKEN]),
     handler: async (req) => {
-      if (req.auth.actor !== ActorType.USER) return;
+      if (req.auth.actor !== ActorType.USER) throw new BadRequestError({ message: "Auth type not supported for this operation" });
 
       const membership = await server.services.org.deleteOrgMembership({
         userId: req.permission.id,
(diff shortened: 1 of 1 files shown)


antan2002:
@maidul98 Would you mind taking a look when you get a chance? I've also applied for the Full Stack Engineer role — happy to discuss how this fix fits into the broader codebase.

PR #7485, opened 2026-07-29
fix: 500 when listing certificates for a certificate profile

molchalih:
getCertificatesByProfile selects cn, but the certificates table has commonName, so GET /api/v1/cert-manager/certificate-profiles/:id/certificates always fails with 42703 column certificates.cn does not exist. The search filter hits the same thing. Aliased back to cn so the response shape is unchanged for the route schema, the types and the frontend.


chrisbennight:
Just flagging I ran into this today, would love to see this get merged or otherwise addressed - i've been working around it with https://infisical.com/docs/api-reference/endpoints/certificates/search endpoint, but would love to have it just work.

PR #7494, opened 2026-07-30
improvement(redis): add throttled structured logging for connection errors

victorvhs017:
## Context

Adds throttled pino logging on all Redis clients (URL, Sentinel, Cluster): `ERROR` at most once per 10s on failure, one `INFO` on recovery (skips initial connect). Docs clarify Redis is required persistent storage (queues, locks, rate limits), not just cache.

## Steps to verify the change

1. Stop Redis while API runs → structured `Redis connection error` within ~10s, ≤1 line per 10s.
2. Start Redis → single `Redis connection restored` log.
3. `GET /api/status` still 200 during outage.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/lib/config/redis.ts
+++ b/backend/src/lib/config/redis.ts
@@ -21,7 +21,7 @@ export type TRedisConfigKeys = Partial<{
 
 const REDIS_ERROR_LOG_THROTTLE_MS = 10_000;
 
-const attachConnectionLogging = (client: Redis | Cluster) => {
+const attachConnectionLogging = (client: Redis | Cluster, name: string) => {
   let lastLoggedAt = 0;
   let isDown = false;
 
@@ -31,7 +31,7 @@ const attachConnectionLogging = (client: Redis | Cluster) => {
 
     lastLoggedAt = now;
     isDown = true;
-    logger.error({ err }, `Redis connection error [error=${err.message}]`);
+    logger.error({ err, redisClient: name }, `Redis connection error [client=${name}] [error=${err.message}]`);
   });
 
   client.on("ready", () => {
@@ -39,13 +39,13 @@ const attachConnectionLogging = (client: Redis | Cluster) => {
     if (!isDown) return;
     isDown = false;
     lastLoggedAt = 0;
-    logger.info("Redis connection restored");
+    logger.info({ redisClient: name }, `Redis connection restored [client=${name}]`);
   });
 
   return client;
 };
 
-export const buildRedisFromConfig = (cfg: TRedisConfigKeys) => {
+export const buildRedisFromConfig = (cfg: TRedisConfigKeys, name = "unknown") => {
   if (cfg.REDIS_URL) {
     return attachConnectionLogging(
       new Redis(cfg.REDIS_URL, {
@@ -58,7 +58,8 @@ export const buildRedisFromConfig = (cfg: TRedisConfigKeys) => {
           }
           return false;
         }
-      })
+      }),
+      name
     );
   }
 
@@ -81,7 +82,8 @@ export const buildRedisFromConfig = (cfg: TRedisConfigKeys) => {
             return false;
           }
         }
-      })
+      }),
+      name
     );
   }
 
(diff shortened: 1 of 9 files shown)


docs/self-hosting/configuration/requirements.mdx
```diff
@@ -62,9 +65,19 @@ Redis is utilized for session management and background tasks in Infisical.
 Redis requirements:
 
 - Use Redis versions 6.x or 7.x. We advise upgrading to at least Redis 6.2.
-- Redis Cluster mode is currently not supported; use Redis Standalone, with or without High Availability (HA).
+- Redis Standalone (with or without High Availability), Redis Sentinel, and Redis Cluster are all
```
Thiago-AS: Are we fully cluster supported?

We have `mget(keys)` and `del(keys)` which can read multiple keys that do not belong to the same hash slot and throw `CROSSSLOT`

We also have `getKeysByPattern` that issues an scan so ioredis routes it to a random node and can result in a silent partial keyspace

backend/src/lib/config/redis.ts
```diff
@@ -17,59 +19,94 @@ export type TRedisConfigKeys = Partial<{
...
-        username: cfg.REDIS_USERNAME,
-        password: cfg.REDIS_PASSWORD,
-        tls: cfg?.REDIS_CLUSTER_ENABLE_TLS ? {} : undefined,
+const attachConnectionLogging = (client: Redis | Cluster, name: string) => {
+  let lastLoggedAt = 0;
+  let isDown = false;
+
+  client.on("error", (err: Error) => {
+    const now = Date.now();
+    if (now - lastLoggedAt < REDIS_ERROR_LOG_THROTTLE_MS) return;
```
Thiago-AS: Should we have the throttle per error type other than per client, this  can end up hiding the actual root cause. 

This channel can receive command level fatal errors as well. If a `NOAUTH`, `WRONGPASS`, or `CROSSSLOT` error happens within 10 seconds of an `ECONNREFUSED`, it gets dropped without leaving any trace.

backend/src/lib/config/redis.ts
```diff
@@ -17,59 +19,94 @@ export type TRedisConfigKeys = Partial<{
...
-    return new Redis.Cluster(cfg.REDIS_CLUSTER_HOSTS, {
-      dnsLookup: cfg.REDIS_CLUSTER_AWS_ELASTICACHE_DNS_LOOKUP_MODE
-        ? (address, callback) => callback(null, address)
-        : undefined,
-      retryDelayOnClusterDown: 300,
-      redisOptions: {
-        username: cfg.REDIS_USERNAME,
-        password: cfg.REDIS_PASSWORD,
-        tls: cfg?.REDIS_CLUSTER_ENABLE_TLS ? {} : undefined,
+const attachConnectionLogging = (client: Redis | Cluster, name: string) => {
```
Thiago-AS: Btw, `redis-bus` and `event-bus-service` is duplicating the `redis` instead of calling `buildRedisFromConfig`, so they wont get named handlers or throtling.

backend/src/lib/config/redis.ts
```diff
@@ -17,59 +19,94 @@ export type TRedisConfigKeys = Partial<{
...
+  let lastLoggedAt = 0;
+  let isDown = false;
+
+  client.on("error", (err: Error) => {
+    const now = Date.now();
+    if (now - lastLoggedAt < REDIS_ERROR_LOG_THROTTLE_MS) return;
+
+    lastLoggedAt = now;
+    isDown = true;
```
Thiago-AS, line 73: It's set on any error, including non connection ones, so a later routine ready can log "Redis connection restored" even though there was never a connection outage.

Maybe trigger that on "close" ?

Thiago-AS (changes requested):

docs/self-hosting/configuration/requirements.mdx
victorvhs017: Great check! I remember some customers using it. I'll create a ticket so we can fix these points and remove that part from the doc now.

Thiago-AS (approved):

PR #7499, opened 2026-07-31
feat: add PKI windows and linux post-sync command support

carlosmonastyrski:
## Context

Delivering a certificate file does not make a service use it. Nginx and IIS keep serving the old certificate until something reloads them, so today the last step is manual.

This adds an optional post-sync command to the Linux Server and Windows Server sync destinations, run on the target host by the Gateway once per sync run that delivers at least one file.

The command is a template using the same `{{placeholder}}` syntax as dynamic secrets. Each value is substituted as a quoted shell literal, so a common name carrying shell metacharacters cannot break out and execute as code. Text that is not a documented placeholder is left exactly as written, so a command calling helm, jq, or a Go template keeps working.

`{{certificatePath}}` and `{{commonName}}` name one specific certificate, so a command using either is confined to a sync with a single certificate linked. That is enforced when the command changes, when the linked set changes, and again at run time for a set that grew on its own, where the run fails rather than resolving the placeholder to an arbitrary certificate.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/pki-sync/pki-sync-post-sync-command-fns.ts
+++ b/backend/src/services/pki-sync/pki-sync-post-sync-command-fns.ts
@@ -1,0 +1,216 @@
+import RE2 from "re2";
+
+import { logger } from "@app/lib/logger";
+
+import { PkiSyncStatus } from "./pki-sync-enums";
+import { PkiSyncError } from "./pki-sync-errors";
+
+export enum PostSyncCommandVariable {
+  CertificatePath = "certificatePath",
+  CertificateDirectory = "certificateDirectory",
+  CertificateFiles = "certificateFiles",
+  CommonName = "commonName",
+  Pkcs12Password = "pkcs12Password"
+}
+
+// Matches a {{name}} placeholder.
+const PLACEHOLDER_PATTERN = new RE2("\\{\\{\\s*([a-zA-Z0-9_]+)\\s*\\}\\}", "g");
+
+const isPostSyncCommandVariable = (name: string): name is PostSyncCommandVariable =>
+  (Object.values(PostSyncCommandVariable) as string[]).includes(name);
+
+const SINGLE_CERTIFICATE_VARIABLES: PostSyncCommandVariable[] = [
+  PostSyncCommandVariable.CertificatePath,
+  PostSyncCommandVariable.CommonName
+];
+
+export const formatPostSyncCommandVariables = (variables: PostSyncCommandVariable[]): string =>
+  variables.map((variable) => `{{${variable}}}`).join(", ");
+
+export const findSingleCertificatePostSyncCommandVariables = (command?: string): PostSyncCommandVariable[] => {
+  if (!command) return [];
+
+  const used = new Set<string>();
+  command.replace(PLACEHOLDER_PATTERN, (match: string, name: string) => {
+    used.add(name);
+    return match;
+  });
+
+  return SINGLE_CERTIFICATE_VARIABLES.filter((variable) => used.has(variable));
+};
+
+export const POST_SYNC_COMMAND_MAX_LENGTH = 2048;
+
+export const POST_SYNC_COMMAND_TIMEOUT_MS = 30_000;
+
+const MAX_CAPTURED_OUTPUT_CHARS = 1000;
+
+const MAX_FAILURE_DETAIL_CHARS = 120;
+
+const REDACTED_PLACEHOLDER = "[REDACTED]";
+
+export type TPostSyncCommandContext = {
+  certificatePath?: string;
+  commonName?: string;
+  certificateDirectory: string;
+  certificateFiles: string[];
+  pkcs12Password?: string;
+};
+
+export type TPostSyncCommandResult = {
+  status: PkiSyncStatus.Succeeded | PkiSyncStatus.Failed;
+  exitCode?: number;
+  durationMs: number;
+  output?: string;
+  failureDetail?: string;
+  error?: string;
+};
+
+export type TPostSyncCommandExecutionResult = { stdout: string; stderr: string; exitCode: number };
+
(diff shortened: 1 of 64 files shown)


backend/src/services/pki-sync/windows-server/windows-server-pki-sync-fns.ts
```diff
@@ -192,6 +231,8 @@ export const windowsServerPkiSyncFactory = ({
           gatewayDeps
         );
 
+        paths.forEach((deliveredPath) => deliveredPaths.add(deliveredPath));
```
bernie-g, line 234: ignore comment placement

We don't have custom roles right now so this isn't an issue right now, but when we do add them I think the UX for users who have the ability to edit cert syncs but not the post sync command is weird. We still render the field and I think the form submission will fail 100% of the time

backend/src/services/pki-sync/pki-sync-post-sync-command-fns.ts
```diff
@@ -0,0 +1,216 @@
...
+  const values = new Map<PostSyncCommandVariable, string | undefined>([
+    [PostSyncCommandVariable.CertificatePath, context.certificatePath],
+    [PostSyncCommandVariable.CertificateDirectory, context.certificateDirectory],
+    [PostSyncCommandVariable.CertificateFiles, context.certificateFiles.join("\n")],
+    [PostSyncCommandVariable.CommonName, context.commonName],
+    [PostSyncCommandVariable.Pkcs12Password, context.pkcs12Password]
+  ]);
+
+  return command.replace(PLACEHOLDER_PATTERN, (match: string, name: string) =>
+    isPostSyncCommandVariable(name) ? toShellLiteral(values.get(name) ?? "") : match
```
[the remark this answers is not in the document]
carlosmonastyrski: The policies and the approval workflows should cover these cases, as the only way to achieve this in the first place is by having no real checks/validations for the certificates issued by the application

backend/src/services/pki-sync/pki-sync-queue.ts
```diff
@@ -617,8 +619,16 @@ export const pkiSyncQueueFactory = ({
         }
       }
 
-      if (certSyncFailureCount > 0) {
-        syncMessage = `${certSyncFailureCount} certificate(s) failed to sync to the destination`;
+      postSyncCommandResult = syncResult.postSyncCommand;
```
bernie-g, line 625: Can we do this earlier so the audit log always has the command no matter what, even if earlier code throws

backend/src/services/pki-sync/pki-sync-queue.ts
bernie-g, line 377: Since we retry certificate syncs multiple times in case of errors, maybe we should add a note in the docs that this command may run multiple times in case of an error

frontend/src/pages/project/RoleDetailsBySlugPage/components/ProjectRoleModifySection.utils.tsx
```diff
@@ -159,7 +159,8 @@ const PkiSyncPolicyActionSchema = z.object({
   [ProjectPermissionPkiSyncActions.Delete]: z.boolean().optional(),
   [ProjectPermissionPkiSyncActions.SyncCertificates]: z.boolean().optional(),
   [ProjectPermissionPkiSyncActions.ImportCertificates]: z.boolean().optional(),
-  [ProjectPermissionPkiSyncActions.RemoveCertificates]: z.boolean().optional()
```
bernie-g, line 162: ignore comment placement

I wonder if it would be useful to have a dry run button that tests the command with dummy data. Thoughts?

frontend/src/pages/cert-manager/PkiSyncDetailsByIDPage/PkiSyncDetailsByIDPage.tsx
```diff
@@ -136,6 +137,7 @@ const PageContent = () => {
                     <PkiSyncDetailsSection pkiSync={pkiSync} />
                     <PkiSyncDestinationSection pkiSync={pkiSync} />
                     <PkiSyncFieldMappingsSection pkiSync={pkiSync} />
+                    <PkiSyncPostSyncCommandSection pkiSync={pkiSync} />
```
bernie-g, line 140: ignore comment placement

Curious, why did we decide to run the command once per sync instead of once per cert?

frontend/src/pages/project/RoleDetailsBySlugPage/components/ProjectRoleModifySection.utils.tsx
carlosmonastyrski: Not sure if it can be done with just dummy data, the majority of the scenarios this would be a service restart or a binding to a new certificate path, if we let a user execute the command with random data, then it'll most likely mess up their instance/process/service. For this V1 we'll be allowing only to set the command and relying on the admin who is configuring it to treat it as a sensitive command that needs to be configured correctly

frontend/src/pages/cert-manager/PkiSyncDetailsByIDPage/PkiSyncDetailsByIDPage.tsx
carlosmonastyrski: For simplicity and how other vendors do this exact same thing. In a future if it's required we'll be adding a more complex structure support where you can define a block that'll be executed once per certificate

backend/src/services/pki-sync/pki-sync-queue.ts
carlosmonastyrski: There is a FAQ for it, but added a bit more visibility in the docs

bernie-g (approved):

PR #7507, opened 2026-08-01
fix: delete alerts when deleting identity

Thiago-AS:
## Context

Deleting an identity left its alerts behind.

- `deleteAlertsForResource(...)` handles scoped removals (leaving a project or org) and only removes alerts from that scope.
- `deleteAlertsForDeletedResource(...)` handles hard deletes and removes every alert for the resource, regardless of scope.

Both use the same internal helper, so channel cleanup, dedup state, and history cleanup stay consistent.

## Steps to verify

1. Start the local stack and run migrations.
2. Create an identity and an expiration alert.
3. Delete the identity through both the v1 and v2 APIs. All alerts for that identity should be gone.
4. Verify the `alerts`, `alert_channels`, `alert_recipients`, and `alert_history` tables no longer contain records for that identity.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/alert/alert-service.ts
victorvhs017: Between the find and the actual delete, we could have new alerts added.

victorvhs017 (changes requested):

victorvhs017 (changes requested):
One thing I just realized is that we are not clearing orphaned alerts that already exist before this is merged.

victorvhs017 (approved):

PR #7510, opened 2026-08-03
feat(pam): redis web access

saifsmailbox98:
## Context

Adds web access for Redis PAM accounts, so users can run Redis commands from the browser instead of going through the CLI. Sessions are recorded and the reason/MFA gates apply, same as the other account types.

PAM-391. Stacked on #7447.

**Terminal:** input is tokenized the way `redis-cli` does it (quotes, `\n`, `\t`, `\xHH`), and replies are formatted the same way (`(nil)`, `(integer) N`, quoted strings, numbered arrays, and `(error)` for a failure inside an array, which is what an `EXEC` reply carries when a queued command fails). Control bytes in replies are rendered printable so a stored value cannot repaint another user's terminal, with newline and tab passed through so multi-line replies like `INFO` stay readable.

**Reply size:** a reply is bounded as it arrives, and the connection is dropped if a single command's reply passes 8MB. Without that, a large value is fully materialized before anything can act on it, which for a 128MB value means roughly 450MB of memory and a fatal out-of-memory that no `try/catch` can reach, ending every session on the pod. Replies under the limit still print their first 256KB followed by a truncation notice.

**Blocked commands:** `SUBSCRIBE`, `PSUBSCRIBE`, `SSUBSCRIBE`, `MONITOR`, `SELECT`, `RESET`, `CLIENT` and `HELLO` are rejected with a message pointing at the CLI. They either push frames the terminal has no way to render or escape from, stop the server replying at all, or leave the connection in a state the client no longer tracks correctly.

**Session end reasons are now visible.** They were written to the terminal and then covered by the disconnected overlay, so a session that ended from its reply size, an idle timeout, or revoked approval all looked identical. The reason now appears on the card that replaces the terminal, for every account type.

## Steps to verify the change

1. Create a Redis account with a gateway, launch a session, click **Connect in Browser**
2. Run `PING`, `SET k v`, `GET k`, `STRLEN k`, `GET missing`, `RPUSH`/`LRANGE`, `HSET`/`HGETALL`, `CONFIG GET maxmemory`, `INFO`, and an unknown command; check each reply type and that an error does not end the session
3. `SET k "a\nb"` then `GET k`, and `SET k "he said \"hi\""` then `GET k`, for the escape round trip
4. `SUBSCRIBE test` and `SELECT 1` should both be refused with the session still alive
5. `MULTI`, `SET k v`, `LPUSH k x`, `GET k`, `EXEC`; expect `(error)` at position 2 with position 3 still shown
6. `EVAL "return string.rep('A', 400*1024)" 0`; expect the first 256KB plus a truncation notice, and the next command to still work
7. `SETRANGE big 12582911 x` then `GET big`; expect the session to end with "The reply was too large to read safely"

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/pam-web-access/redis/pam-redis-formatter.ts
```diff
@@ -0,0 +1,70 @@
...
+    const ch = input[i];
+
+    if (inQuote) {
+      if (ch === inQuote) {
+        tokens.push(current);
+        current = "";
+        inQuote = null;
+      } else {
+        current += ch;
+      }
```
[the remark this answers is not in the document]
saifsmailbox98: Valid, this was genuinely broken. I've added escape handling inside quotes, so `SET key "hello\"world"` now comes through as 3 tokens instead of 4. Added a few test cases for it as well.

backend/src/ee/services/pam-web-access/redis/pam-redis-formatter.ts
```diff
@@ -0,0 +1,70 @@
...
+  let inQuote: "'" | '"' | null = null;
+
+  for (let i = 0; i < input.length; i += 1) {
+    const ch = input[i];
+
+    if (inQuote) {
+      if (ch === inQuote) {
+        tokens.push(current);
+        current = "";
+        inQuote = null;
```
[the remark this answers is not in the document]
saifsmailbox98: This is the same thing greptile flagged just above on this file, so the escape handling fix covers both.

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
```diff
@@ -0,0 +1,164 @@
...
+export const handleRedisSession = async (
+  ctx: TSessionContext,
+  params: { connectionDetails: Record<string, unknown>; credentials: Record<string, unknown> }
+): Promise<TSessionHandlerResult> => {
+  const { socket, relayPort, resourceName, sessionId, sendMessage, sendSessionEnd, isNearSessionExpiry, onCleanup } =
+    ctx;
+  const connectionDetails = params.connectionDetails as { host: string; port: number };
+  const credentials = params.credentials as { username?: string };
+
+  const redisClient = new Redis({
```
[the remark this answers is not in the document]
saifsmailbox98: I think this one's a false positive. I tested it with an ACL user that has `-info`, and ioredis just skips the ready check when INFO isn't permitted and emits ready anyway, so the session still connects fine.

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
```diff
@@ -0,0 +1,164 @@
...
+    prompt
+  });
+
+  logger.info({ sessionId }, "Redis web access session established");
+
+  // Sequential message processing to prevent concurrent command issues
+  let processingPromise = Promise.resolve();
+
+  socket.on("message", (rawData: Buffer | ArrayBuffer | Buffer[]) => {
+    processingPromise = processingPromise
```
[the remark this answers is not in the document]
saifsmailbox98: This is real, but it isn't specific to redis. None of the handlers here have backpressure and there's no central frame limit either, so ssh and the data explorer behave the same way. Feels like something we should do centrally rather than in this one handler, so I'm leaving it out of this PR.

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  const tokens = tokenizeRedisInput(trimmed);
+  if (tokens.length === 0) {
+    return { output: "", shouldClose: false };
+  }
+
+  const [command, ...args] = tokens;
+
+  try {
+    const result = await redisClient.call(command, ...args);
```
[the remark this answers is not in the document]
saifsmailbox98: Valid, and the cleanup part was the real problem. `quit()` was queueing behind the blocked command and never returning, so the client and its tunnel were staying open. I've added a 30s deadline on commands plus a forced disconnect on cleanup. Not adding response size limits for now, that needs a bit more thought.

backend/src/ee/services/pam-web-access/redis/pam-redis-formatter.ts
```diff
@@ -0,0 +1,77 @@
+import RE2 from "re2";
+
+const escapeBackslash = new RE2(/\\/, "g");
+const escapeQuote = new RE2(/"/, "g");
+
+const escapeRedisString = (str: string): string => str.replace(escapeBackslash, "\\\\").replace(escapeQuote, '\\"');
```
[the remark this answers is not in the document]
saifsmailbox98: Valid, control bytes were going straight through to xterm. They are escaped into printable form now, same as redis-cli does, and applied to error messages too.

backend/src/ee/services/pam-web-access/redis/pam-redis-formatter.test.ts
```diff
@@ -0,0 +1,162 @@
+import { describe, expect, it } from "vitest";
+
+import { formatRedisReply, tokenizeRedisInput } from "./pam-redis-formatter";
```
bernie-g, line 3: ignore comment placement

Image
Can we remove the host/port in the parenthesis since it's in every prompt already

backend/src/ee/services/pam-web-access/redis/pam-redis-formatter.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+describe("tokenizeRedisInput", () => {
+  it("splits simple command", () => {
+    expect(tokenizeRedisInput("SET key value")).toEqual(["SET", "key", "value"]);
+  });
+
+  it("handles double-quoted strings", () => {
+    expect(tokenizeRedisInput('SET key "hello world"')).toEqual(["SET", "key", "hello world"]);
+  });
```
bernie-g, line 12: ignore comment placement

Image
We seem to not be able to render newlines

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
```diff
@@ -0,0 +1,198 @@
...
+  } catch (err) {
+    try {
+      redisClient.disconnect();
+    } catch {
+      /* ignore */
+    }
+    throw err;
+  }
+
+  const prompt = `${connectionDetails.host}:${connectionDetails.port}> `;
```
bernie-g, line 152: ignore comment placement

I think we are missing a maximum response length cap. We should also add a max execution time cap

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
```diff
@@ -0,0 +1,198 @@
...
+// and its relay tunnel open. Bound the graceful close, then force the socket shut.
+const CLEANUP_QUIT_TIMEOUT_MS = 2_000;
+
+const callWithDeadline = async (redisClient: Redis, command: string, args: string[]): Promise<unknown> => {
+  let timer: NodeJS.Timeout | undefined;
+  try {
+    return await Promise.race([
+      redisClient.call(command, ...args),
+      new Promise<never>((_, reject) => {
+        timer = setTimeout(
```
bernie-g, line 48: ignore comment placement

Running `SUBSCRIBE test` crashes the entire backend

bernie-g (changes requested):

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
saifsmailbox98: Blocked the pubsub commands and a few others that put the connection into a mode ioredis can't follow.

backend/src/ee/services/pam-web-access/redis/pam-redis-formatter.test.ts
saifsmailbox98: Fixed `\n` being stored as a literal `n` and `INFO` collapsing into one line. Newline and tab pass through now.

backend/src/ee/services/pam-web-access/redis/pam-redis-formatter.test.ts
saifsmailbox98: Removed

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
saifsmailbox98: Per-command timeout was already there at 30s. Added a 256KB response cap, and a 15s timeout on connection setup since that had none.

backend/src/ee/services/pam-web-access/redis/pam-redis-session-handler.ts
```diff
@@ -0,0 +1,236 @@
...
+    return {
+      output: `(error) ${command.toUpperCase()} is not supported in web access, use the Infisical CLI for this\n`,
+      shouldClose: false
+    };
+  }
+
+  try {
+    const result = await callWithDeadline(redisClient, command, args);
+    const formatted = formatRedisReply(result);
+    if (formatted.length > MAX_REPLY_BYTES) {
```
bernie-g, line 93: This MAX_REPLY_BYTES is pretty superficial here since the full unbounded response could already get loaded into memory in `result`

bernie-g: This will still OOM the backend I think

docs/documentation/platform/pam/accounts/redis.mdx
```diff
@@ -38,22 +38,39 @@ Redis accounts let you manage access to your Redis databases. Users connect thro
...
+    3. Click **Connect in Browser**
 
-The command outputs the local port to connect to. Then use `redis-cli` or any Redis client:
-```bash
-redis-cli -h 127.0.0.1 -p <port>
-```
+    Type `quit` or `exit` to end the session.
 
-No password needed since the proxy handles authentication.
+    Web Access is built for interactive use. Very long replies are shortened, and commands that stream results, such as `SUBSCRIBE` and `MONITOR`, are available through the CLI.
```
bernie-g:
```suggestion
    Web Access is built for interactive use. Very long replies are shortened, and commands that stream results, such as `SUBSCRIBE` and `MONITOR`, are only available through the CLI.
```

bernie-g (dismissed):

bernie-g (dismissed):

bernie-g (approved):

PR #7514, opened 2026-08-03
feat(telemetry): report PKI certificate issuance from profile-based flows

saifsmailbox98:
## Context

PKI metrics were broken in two separate ways:

1. **The dashboards pointed at an event that no longer exists.** `Issue PKI Certificate` was renamed to `Issue PKI Certificate aggregated` back in May, so every PKI dashboard silently read zero. Already fixed on the PostHog side — screenshots below.

2. **Most certificate issuance never emitted an event at all.** The modern path logged `Certificate Request Created` instead of an issuance, and ACME, EST and SCEP logged nothing. Only **5 orgs** looked like they were issuing certificates. **54** actually were.

This PR fixes the second one. Every issuance flow now emits `Issue PKI Certificate`, tagged with the application, profile and enrollment method, so issuance can be broken down by ACME vs EST vs SCEP vs API and attributed to a PKI application.

## Screenshots

**Infisical PKI dashboard** — empty → MAU ~40, WAU ~14-21, DAU 1-10

| Before | After |
|---|---|
| PKI dashboard before | PKI dashboard after |

**PKI Activation Rate (30d)** — `0%` → `0.1%`

| Before | After |
|---|---|
| PKI activation rate before | PKI activation rate after |

**PKI Activation Funnel** — step 3: `0 orgs` → `21 orgs`

| Before | After |
|---|---|
| PKI activation funnel before | PKI activation funnel after |

These stay low until this ships, since the orgs on the modern path still aren't emitting an issuance event.

<details>
<summary><b>Implementation notes</b></summary>

**Where the event is emitted**

| Flow | Reported from |
|---|---|
| API / UI — issue, sign | `certificate-v3-service` |
| ACME, EST, SCEP | `certificate-v3-service` — all three converge on `signCertificateFromProfile` |
| Self-signed | `certificate-v3-service` |
| Renewal — API and auto-renew cron | `certificate-v3-service` |
| External CAs that finish in one job — ACME, Azure AD CS, ADCS, AWS ACM, AWS PCA, Venafi TPP | `certificate-issuance-queue` |
| DigiCert, GoDaddy | their own polling processors |

Two things make that split necessary:

- `orderCertificate` always returns `PENDING` for external CAs, so instrumenting it would have counted orders rather than issuances. The queue reports instead.
- DigiCert and GoDaddy can end that queue job at `PENDING_VALIDATION`, with the certificate only arriving later. The queue tracks whether a certificate actually exists by the end of the job and stays silent when it doesn't; the two processors report at the point they attach the certificate. (This is tracked in the handler rather than re-read from the DB — a re-read hits the read replica moments after the write, so replica lag could drop the event.)

All six call sites go through one shared reporter, `certificate-common/certificate-telemetry-fns.ts`, rather than assembling the payload each time.

**Event properties:** `applicationId`, `profileId`, `enrollmentType`, `operation` (`issue` / `sign` / `order` / `renew`). The last two are registered in `AGGREGATION_BREAKDOWN_DIMENSIONS` so the hourly aggregation keeps them as flat values instead of histogramming them into `{"acme": 12}`. `enrollmentType` is read from the profile on the async path, since ACME and SCEP land there for external CAs.

**PKI applications:** `Created` / `Deleted` now carry `applicationId` (previously only `orgId`, so an application couldn't be joined to anything), and `PKI Application Updated` was added — it wasn't tracked at all.

**Left alone:** legacy CA, subscriber and deprecated certificate routes keep their existing events. `SignCert` has had zero events in 90 days but its emit sites are on working routes, so removing it buys nothing. Every emit is wrapped in try/catch — analytics must never fail an issuance.

**Tests:** unit tests assert the event fires with the right properties on the synchronous issue and sign paths. The async queue and the two processors are not covered.

</details>

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/certificate-authority/certificate-issuance-queue.ts
+++ b/backend/src/services/certificate-authority/certificate-issuance-queue.ts
@@ -441,9 +441,8 @@ export const certificateIssuanceQueueFactory = ({
       );
     };
 
-    // DigiCert and GoDaddy can finish this job at PENDING_VALIDATION, with the certificate only
-    // arriving later in their polling processors. Those processors report issuance themselves, so
-    // this job must not report anything when it leaves the request pending.
+    // DigiCert and GoDaddy attach the certificate later in their processors, so a pending order
+    // must not be reported as issued. Tracked here rather than re-read: the replica lags the write.
     let certificateExistsAfterThisJob = true;
 
     try {
@@ -1216,8 +1215,6 @@ export const certificateIssuanceQueueFactory = ({
       }
 
       if (certificateExistsAfterThisJob) {
-        // ACME and SCEP fall back to this async path when the profile is backed by an external CA,
-        // so the enrollment method has to come from the profile rather than being assumed to be API.
         const telemetryProfile = profileId ? await certificateProfileDAL?.findById(profileId) : undefined;
 
         await reportCertificateIssued({
--- a/backend/src/services/certificate-authority/certificate-issuance-queue.ts
+++ b/backend/src/services/certificate-authority/certificate-issuance-queue.ts
@@ -441,6 +441,11 @@ export const certificateIssuanceQueueFactory = ({
       );
     };
 
+    // DigiCert and GoDaddy can finish this job at PENDING_VALIDATION, with the certificate only
+    // arriving later in their polling processors. Those processors report issuance themselves, so
+    // this job must not report anything when it leaves the request pending.
+    let certificateExistsAfterThisJob = true;
+
     try {
       logger.info(`Processing certificate issuance job for [certificateId=${certificateId}] [caId=${caId}]`);
 
@@ -955,12 +960,14 @@ export const certificateIssuanceQueueFactory = ({
               `DigiCert order issued immediately (pre-validated domains), attached certificate [certificateRequestId=${certificateRequestId}] [certificateId=${attachedCertificateId}]`
             );
           } catch (finaliseError) {
+            certificateExistsAfterThisJob = false;
             logger.error(
               finaliseError,
               `DigiCert immediate finalisation failed, will be retried by polling queue [certificateRequestId=${certificateRequestId}]`
             );
           }
         } else {
+          certificateExistsAfterThisJob = false;
           await setPending(`DigiCert is processing the request — order #${digicertResult.metadata.digicert.orderId}`);
           logger.info(
             `DigiCert order placed, awaiting validation [certificateRequestId=${certificateRequestId}] [orderId=${digicertResult.metadata.digicert.orderId}]`
@@ -1043,6 +1050,7 @@ export const certificateIssuanceQueueFactory = ({
           return;
         }
 
+        certificateExistsAfterThisJob = false;
         await setPending(
           `GoDaddy is processing the request — certificate ${godaddyResult.metadata.godaddy.certificateId}`
         );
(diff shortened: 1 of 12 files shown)


backend/src/services/certificate-authority/certificate-issuance-queue.ts
```diff
@@ -1201,6 +1206,34 @@ export const certificateIssuanceQueueFactory = ({
...
+      // is counted alongside the synchronous paths in certificate-v3-service.
+      try {
+        const project = await projectDAL.findById(ca.projectId);
+        if (project?.orgId) {
+          // ACME and SCEP fall back to this async path when the profile is backed by an external CA,
+          // so the enrollment method has to come from the profile rather than being assumed to be API.
+          const telemetryProfile = profileId ? await certificateProfileDAL?.findById(profileId) : undefined;
+
+          await telemetryService.sendPostHogEvents({
+            event: PostHogEventTypes.IssueCert,
```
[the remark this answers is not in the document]
saifsmailbox98: Valid. GoDaddy always falls through and DigiCert does too without `immediateCertificateId`, so pending validation was getting counted as issued. Gated it on the certificate actually existing after the job, and the DigiCert/GoDaddy processors now report at the point they attach the cert.

backend/src/services/certificate-authority/digicert/digicert-certificate-authority-processor.ts
```diff
@@ -181,6 +187,17 @@ export const processDigiCertPendingValidationRequest = async (
...
     );
+
+    await reportCertificateIssued({
+      telemetryService: deps.telemetryService,
+      projectDAL: deps.projectDAL,
+      projectId: request.projectId,
+      profileId: request.profileId,
+      applicationId: request.applicationId,
+      enrollmentType: EnrollmentType.API,
+      operation: parsed.digicert.isRenewal ? "renew" : "order"
```
carlosmonastyrski: let's use an enum for these values, maybe in the future we rename them or need to change them and would be easier to extend/maintain with that inside an enum 🙏

backend/src/services/certificate-common/certificate-telemetry-fns.ts
```diff
@@ -0,0 +1,53 @@
...
+  operation,
+  profileId,
+  applicationId,
+  orgId,
+  distinctId,
+  projectDAL
+}: TReportCertificateIssuedDTO) => {
+  try {
+    const resolvedOrgId = orgId || (projectDAL ? (await projectDAL.findById(projectId))?.orgId : undefined);
+    if (!resolvedOrgId) return;
```
carlosmonastyrski: Should we add a warn log or something here? feels off that in a cert manager event we can not find the orgId 🤔

backend/src/services/certificate-authority/digicert/digicert-certificate-authority-processor.ts
saifsmailbox98: yeah makes sense, made it an enum in `certificate-constants.ts`

backend/src/services/certificate-common/certificate-telemetry-fns.ts
saifsmailbox98: added a warn. only way to hit it is the project getting deleted mid issuance

carlosmonastyrski (approved):

PR #7517, opened 2026-08-03
fix(identity-auth): use stored accessTokenTTL as fallback in update TTL validation

### LINKED ISSUE #7516: Updating only accessTokenMaxTTL is wrongly rejected in 8 identity auth methods (stored maxTTL used as TTL fallback)
### Product

Platform/Machine identities

### Describe the bug

The update flows for 8 identity auth methods share a copy-pasted TTL validation:

```ts
if (
  (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL) > 0 &&
  (accessTokenTTL || identityGcpAuth.accessTokenMaxTTL) > (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL)
) {
  throw new BadRequestError({ message: "Access token TTL cannot be greater than max TTL" });
}
```

On the left side of the comparison, when `accessTokenTTL` is omitted from the request, the fallback is the stored `accessTokenMaxTTL`. It should be the stored `accessTokenTTL` — the check is supposed to compare the effective TTL against the effective max TTL.

The result: **lowering only `accessTokenMaxTTL` is rejected even when the stored TTL is well below the new max.**

Affected update flows (left-side fallback is `accessTokenMaxTTL`):

- `identity-gcp-auth-service.ts` (`updateGcpAuth`)
- `identity-kubernetes-auth-service.ts` (`updateKubernetesAuth`)
- `identity-oidc-auth-service.ts` (`updateOidcAuth`)
- `identity-jwt-auth-service.ts` (`updateJwtAuth`)
- `identity-azure-auth-service.ts` (`updateAzureAuth`)
- `identity-aws-auth-service.ts` (`updateAwsAuth`)
- `identity-token-auth-service.ts` (`updateTokenAuth`)
- `identity-ua-service.ts` (`updateUniversalAuth`)

The five newer auth methods already do this correctly with `accessTokenTTL` as the left-side fallback: `identity-alicloud-auth`, `identity-ldap-auth`, `identity-oci-auth`, `identity-spiffe-auth`, `identity-tls-cert-auth`. The differential between the two groups is what makes the intended semantics unambiguous.

### To Reproduce

1. Create a machine identity with Universal Auth: `accessTokenTTL = 3600` (1h), `accessTokenMaxTTL = 2592000` (30d).
2. `PATCH /api/v1/auth/universal-auth/identities/:identityId` with body `{ "accessTokenMaxTTL": 604800 }` (7d), without re-sending `accessTokenTTL`.
3. Response is `400 Access token TTL cannot be greater than max TTL`, because the check compares the stored **max** TTL (30d) against the new max (7d) instead of the stored TTL (1h).

### Expected behavior

The update succeeds: the effective TTL after the update (1h, unchanged) is below the new max TTL (7d). Re-sending `accessTokenTTL: 3600` alongside the max-TTL change makes the same request pass, which shows the stored-TTL comparison is what was intended.

### Deployment Type

Self-hosted

### Additional context

Minor related cleanup: `identity-azure-auth-service.ts` names the row variable `identityGcpAuth` (copy-paste from the GCP service).

I have a fix ready and will open a PR referencing this issue.

omlahore:
## Context

Fixes #7516.

The update flows for 8 identity auth methods validate that the new access token TTL does not exceed the new max TTL. When `accessTokenTTL` is omitted from the request, the left side of the comparison falls back to the stored `accessTokenMaxTTL` instead of the stored `accessTokenTTL`:

```ts
(accessTokenTTL || identityGcpAuth.accessTokenMaxTTL) > (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL)
```

So lowering only `accessTokenMaxTTL` is rejected with `Access token TTL cannot be greater than max TTL` even when the stored TTL is far below the new max. Example: TTL=1h, maxTTL=30d; `PATCH { accessTokenMaxTTL: 7d }` compares 30d > 7d and 400s, although 1h < 7d is valid.

The five newer auth methods (alicloud, ldap, oci, spiffe, tls-cert) already use the stored `accessTokenTTL` as the fallback. This PR aligns the remaining eight (gcp, kubernetes, oidc, jwt, azure, aws, token-auth, universal-auth) with them. One-word change per service.

Also renames the copy-pasted `identityGcpAuth` variable in the Azure auth service to `identityAzureAuth`.

## Screenshots

N/A (backend validation change).

## Steps to verify the change

1. Create a machine identity with Universal Auth: `accessTokenTTL: 3600`, `accessTokenMaxTTL: 2592000`.
2. `PATCH /api/v1/auth/universal-auth/identities/:identityId` with `{ "accessTokenMaxTTL": 604800 }` only.
3. Before: 400 `Access token TTL cannot be greater than max TTL`. After: 200, maxTTL updated.
4. Invalid updates still rejected: with stored TTL 3600, `{ "accessTokenMaxTTL": 60 }` fails as before, and `{ "accessTokenTTL": 7200, "accessTokenMaxTTL": 3600 }` fails as before.

Covered by a regression test: `backend/src/services/identity-ua/identity-ua-service.test.ts`. On `main` the first case fails with `Access token TTL cannot be greater than max TTL`; with this fix all four pass. The other three cases assert the guards this must not weaken (lowering max TTL below the stored TTL still rejects, an explicit TTL above an explicit max TTL still rejects, max TTL 0 still means unlimited) and pass on `main` too.

Universal auth is the representative case; the other seven services have the identical expression. `npm run test:unit`, `npm run type:check` and eslint are clean.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally (unit test + type:check + eslint)

### DIFF
--- a/backend/src/services/identity-aws-auth/identity-aws-auth-service.ts
+++ b/backend/src/services/identity-aws-auth/identity-aws-auth-service.ts
@@ -497,7 +497,7 @@ export const identityAwsAuthServiceFactory = ({
 
     if (
       (accessTokenMaxTTL || identityAwsAuth.accessTokenMaxTTL) > 0 &&
-      (accessTokenTTL || identityAwsAuth.accessTokenMaxTTL) > (accessTokenMaxTTL || identityAwsAuth.accessTokenMaxTTL)
+      (accessTokenTTL || identityAwsAuth.accessTokenTTL) > (accessTokenMaxTTL || identityAwsAuth.accessTokenMaxTTL)
     ) {
       throw new BadRequestError({ message: "Access token TTL cannot be greater than max TTL" });
     }
--- a/backend/src/services/identity-azure-auth/identity-azure-auth-service.ts
+++ b/backend/src/services/identity-azure-auth/identity-azure-auth-service.ts
@@ -393,11 +393,11 @@ export const identityAzureAuthServiceFactory = ({
       });
     }
 
-    const identityGcpAuth = await identityAzureAuthDAL.findOne({ identityId });
+    const identityAzureAuth = await identityAzureAuthDAL.findOne({ identityId });
 
     if (
-      (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL) > 0 &&
-      (accessTokenTTL || identityGcpAuth.accessTokenMaxTTL) > (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL)
+      (accessTokenMaxTTL || identityAzureAuth.accessTokenMaxTTL) > 0 &&
+      (accessTokenTTL || identityAzureAuth.accessTokenTTL) > (accessTokenMaxTTL || identityAzureAuth.accessTokenMaxTTL)
     ) {
       throw new BadRequestError({ message: "Access token TTL cannot be greater than max TTL" });
     }
@@ -445,7 +445,7 @@ export const identityAzureAuthServiceFactory = ({
       return extractIPDetails(accessTokenTrustedIp.ipAddress);
     });
 
-    const updatedAzureAuth = await identityAzureAuthDAL.updateById(identityGcpAuth.id, {
+    const updatedAzureAuth = await identityAzureAuthDAL.updateById(identityAzureAuth.id, {
       tenantId,
       resource,
       allowedServicePrincipalIds,
--- a/backend/src/services/identity-gcp-auth/identity-gcp-auth-service.ts
+++ b/backend/src/services/identity-gcp-auth/identity-gcp-auth-service.ts
@@ -444,7 +444,7 @@ export const identityGcpAuthServiceFactory = ({
 
     if (
       (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL) > 0 &&
-      (accessTokenTTL || identityGcpAuth.accessTokenMaxTTL) > (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL)
+      (accessTokenTTL || identityGcpAuth.accessTokenTTL) > (accessTokenMaxTTL || identityGcpAuth.accessTokenMaxTTL)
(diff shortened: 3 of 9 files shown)


omlahore:
Added a regression test: `backend/src/services/identity-ua/identity-ua-service.test.ts`

On `main`, one of its four cases fails with the exact bug:

```
FAIL  updateUniversalAuth access token TTL validation >
      lowering only accessTokenMaxTTL is accepted when the stored TTL still fits
AssertionError: promise rejected "BadRequest: Access token TTL cannot be
                greater than max TTL" instead of resolving
```

With the fix applied all four pass. The other three cover the guards this must not weaken: lowering max TTL below the stored TTL is still rejected, an explicit TTL above an explicit max TTL is still rejected, and a max TTL of 0 still means unlimited. Those three pass on `main` as well, so the test is not just agreeing with the change.

Universal auth is the representative case here; the other seven services carry the identical expression. It follows the existing service factory unit test pattern with mocked DALs and a CASL ability, same shape as `service-token-service.test.ts`. `test:unit`, `type:check` and eslint are clean.

The fix itself is unchanged: one word per service across 8 files, aligning gcp, kubernetes, oidc, jwt, azure, aws, token-auth and universal-auth with alicloud, ldap, oci, spiffe and tls-cert, which already use the stored `accessTokenTTL` as the fallback.

@maidul98

omlahore:
@adkah this has been open since 3 Aug with no review. It fixes #7516: the update flows for eight identity auth methods compare the incoming access token TTL against the stored `accessTokenMaxTTL` rather than the stored `accessTokenTTL`, so lowering only `accessTokenMaxTTL` on an existing identity is rejected even when the values are valid.

The fix is the same one-line fallback in each of the eight services, no behaviour change when `accessTokenTTL` is supplied. All checks pass.

PR #7532, opened 2026-08-04
improvement: revise identity alert ui

scott-ray-wilson:
## Context

This PR revamps the identity alert UI

## Screenshots
CleanShot 2026-08-04 at 16 30 31@2x

CleanShot 2026-08-04 at 16 30 34@2x

CleanShot 2026-08-04 at 16 30 46@2x

CleanShot 2026-08-04 at 16 30 49@2x

CleanShot 2026-08-04 at 16 30 52@2x

## Header button (org identity page)
- [ ] No alert: `Alert` button shows (no dot), sits left of Options, opens the Add Alert sheet
- [ ] Details card no longer shows the old Alert row

## Add Alert sheet
- [ ] Order: Name (prefilled "Secret expiration alert") → Description → watching banner → Enabled → Condition → Repeat daily → Channels
- [ ] Banner: ClockAlert icon, info tint, top-aligned, bold identity name
- [ ] Condition: number input validates 1-90, error shows beneath
- [ ] Repeat daily renders as an org-tinted checkbox
- [ ] Add channel: all 4 types appear; channel cards have no Name field; trash only when 2+ channels
- [ ] Email: recipients select works, required; Slack/Webhook/PagerDuty: config fields required
- [ ] Send test fires from an unsaved channel
- [ ] Submitting with no channels shows "At least one channel is required"
- [ ] Create Alert → success toast, sheet closes

## Popover (alert exists)
- [ ] Button gains green dot (gray when disabled)
- [ ] Popover: name, Active/Disabled badge, "7 days before client secret expiry", "Notifies Email, Slack", comfortable width
- [ ] Enabled switch toggles inline: toast, badge and dot flip, popover stays open
- [ ] Edit Details closes popover, opens prefilled Edit Alert sheet (existing secrets show "leave blank to keep")
- [ ] Remove Alert: v3 dialog, requires typing the alert name, spinner while deleting, button returns to dot-less state

## Project identity page
- [ ] Project-managed identity: Alert button before Permission Audit, project-yellow accents, alert is project-scoped
- [ ] Org-managed identity: read-only popover (no toggle/edit/remove), button hidden entirely when no alert exists

## Permissions & regressions
- [ ] User without identity edit permission: create button / toggle / edit / remove all disabled
- [ ] Pre-existing alert (made before this change) still renders and survives an edit round-trip (custom channel names preserved)

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


frontend/src/views/Alerts/AlertAction.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+    try {
+      await updateAlert.mutateAsync({ alertId: existingAlert.id, enabled });
+      createNotification({ text: `Alert ${enabled ? "enabled" : "disabled"}`, type: "success" });
+    } catch {
+      // MutationCache reports request errors globally.
+    }
+  };
+
+  const handleDeleteAlert = async () => {
```
Thiago-AS, line 110: The input is never cleared after deletion.

Create an alert with X name. Delete it, then create another alert and accept the default name. Open Remove Alert. The input is prefilled with the old name.

frontend/src/views/Alerts/AlertAction.tsx
```diff
@@ -0,0 +1,285 @@
...
+              <Badge variant={existingAlert.enabled ? "success" : "neutral"}>
+                {existingAlert.enabled ? "Active" : "Disabled"}
+              </Badge>
+            </div>
+            <p className="text-sm text-muted">{formatConditionSummary(existingAlert)}</p>
+            {channelSummary && <p className="text-sm text-muted">Notifies {channelSummary}</p>}
+          </div>
+          {!readOnly &&
+            renderPermissionGate((isAllowed) => (
+              <>
```
Thiago-AS: when access is denied, the gate wraps this in a tooltip trigger using asChild, which can only clone a single element. the fragment will lose the ref and break the Access restricted tooltip.

frontend/src/views/Alerts/AlertAction.tsx
```diff
@@ -0,0 +1,285 @@
...
+// stored condition is missing or malformed.
+const formatConditionSummary = (alert: TAlert): string => {
+  const days = parseAlertBeforeDays(alert.condition?.alertBefore);
+  if (days === null) {
+    return ALERT_EVENT_TYPE_LABELS[alert.eventType as AlertEventType] ?? alert.eventType;
+  }
+  return `${days} day${days === 1 ? "" : "s"} before client secret expiry`;
+};
+
+const formatChannelSummary = (alert: TAlert): string =>
```
Thiago-AS, line 67: We are ignoring channels being enabled here, so even if I have the channel disabled, it shows that it will notify the channel.

frontend/src/views/Alerts/AlertForm.tsx
```diff
@@ -215,209 +179,150 @@ export const AlertForm = ({
...
+            </FieldContent>
+          </Field>
 
-          <section className="flex flex-col gap-4">
-            <SectionHeader title="Trigger" />
+          <Alert className="items-start [&>svg]:mt-0.5 [&>svg]:text-info">
+            <ClockAlertIcon />
+            <AlertDescription>
+              <span>
+                Watching <span className="font-medium text-foreground">{watchTarget}</span> ·
```
Thiago-AS, line 217: The idea here was to have reusable components that people can just plug into, with the labels automatically derived from the source, type, and event. No need to change anything now, just mentioning it so we avoid creating a separate alert form for every new alert type.

frontend/src/views/Alerts/ChannelsField.tsx
```diff
@@ -28,7 +28,8 @@ type Props = {
 
 const buildNewChannel = (channelType: AlertChannelType): TChannelForm => ({
   channelType,
-  name: "",
+  // The name input was dropped from the design; new channels are named after their type.
```
Thiago-AS: I'm okay with that, but can we at least add some kind of uniqueness for now? Say someone adds three webhooks and they're all named "Webhook". That's not great if they import them into Terraform, since they'll all end up with the same name and will be hard to manage because the endpoint is a secret.

Thiago-AS (changes requested):

Thiago-AS (dismissed):

Thiago-AS (approved):

PR #7535, opened 2026-08-05
improvement(pki): show disabled certificate profiles for API enrollment

saifsmailbox98:
## Context

The certificate request dropdown on an Application only listed profiles that already had API enrollment configured, so a profile attached to the Application but not configured for API just silently vanished from the list with no explanation. Now every attached profile is listed, and the ones without API enrollment render as disabled rows with "API enrollment not configured" under them, plus a hint pointing at the Application's Settings tab. The empty state is narrowed to the case it actually describes: no profiles attached to the Application at all.

**Cache invalidation:** found while testing the above. The dropdown's items come from the `certificate-profiles / list` query scoped by `applicationId`, but the Application mutations only invalidated `pkiApplicationKeys.all`. With the global 60s `staleTime`, attaching a profile and returning to the request sheet within that window served the stale list, so the newly attached profile was missing (and a detached one lingered) until the query went stale and remounted. `invalidateAll` in `pkiApplications/mutations.ts` now also invalidates `["certificate-profiles", "list"]`, which covers attach, detach, and every enrollment set/clear path since they all route through it. Same cross-domain invalidation the reverse direction already does in `certificateProfiles/mutations.tsx` and `certificates/mutations.tsx`.

## Screenshots

Certificate profile dropdown showing a profile disabled because API enrollment is not configured

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/hooks/api/pkiApplications/mutations.ts
+++ b/frontend/src/hooks/api/pkiApplications/mutations.ts
@@ -5,6 +5,7 @@ import { apiRequest } from "@app/config/request";
 import { approvalGrantQuery } from "../approvalGrants/queries";
 import { approvalPolicyQuery } from "../approvalPolicies/queries";
 import { approvalRequestQuery } from "../approvalRequests/queries";
+import { certificateProfileKeys } from "../certificateProfiles/queries";
 import { pkiApplicationKeys } from "./queries";
 import {
   TAddPkiApplicationMemberDTO,
@@ -31,7 +32,7 @@ const BASE_URL = "/api/v1/cert-manager/applications";
 
 const invalidateAll = (qc: ReturnType<typeof useQueryClient>) => {
   qc.invalidateQueries({ queryKey: pkiApplicationKeys.all });
-  qc.invalidateQueries({ queryKey: ["certificate-profiles", "list"] });
+  qc.invalidateQueries({ queryKey: certificateProfileKeys.lists() });
 };
 
 export const useCreatePkiApplication = () => {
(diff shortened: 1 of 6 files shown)


frontend/src/hooks/api/pkiApplications/mutations.ts
```diff
@@ -29,8 +29,10 @@ import {
 
 const BASE_URL = "/api/v1/cert-manager/applications";
 
-const invalidateAll = (qc: ReturnType<typeof useQueryClient>) =>
+const invalidateAll = (qc: ReturnType<typeof useQueryClient>) => {
   qc.invalidateQueries({ queryKey: pkiApplicationKeys.all });
+  qc.invalidateQueries({ queryKey: ["certificate-profiles", "list"] });
```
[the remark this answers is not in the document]
saifsmailbox98: Not really an issue. The profile list is mounted only on the Settings tab, which is exactly where attach/detach and enrollment changes happen. Member changes are on a different tab, so nothing extra is fetched. Only app rename/delete makes one extra call, keeping it shared since most mutations here need it.

frontend/src/hooks/api/pkiApplications/mutations.ts
```diff
@@ -29,8 +29,10 @@ import {
 
 const BASE_URL = "/api/v1/cert-manager/applications";
```
carlosmonastyrski, line 31: Ignore comment placement
Let's at least sort it, so enabled options are shown first, otherwise it's a bit messy to use

Image

frontend/src/hooks/api/pkiApplications/mutations.ts
carlosmonastyrski: shouldn't we just use "certificateProfileKeys"?

carlosmonastyrski (changes requested):

carlosmonastyrski (approved):

PR #7536, opened 2026-08-05
feat: scope code-signing approvals to a signing situation

carlosmonastyrski:
## Context

Today an approval to sign code is all-or-nothing: once someone is approved, they can sign anything until their signature count or time window runs out. This change lets you tie an approval to the exact situation it was granted for, so you can approve "this specific installer, built by this command, on this build machine, as this account" instead of "100 signatures". Anything you leave blank stays unrestricted, so you choose how tight it is. The IP address is filled in from the request itself, since that is the one parameter a client cannot observe.

A request's scope is fixed once it is open, so wrong parameters mean rejecting it and opening a new one rather than editing it under review. For a pipeline that signs something different every run, the signing clients choose what to leave out up front with `approval.exclude_scope_fields`, usually `data_hash`, since timestamping changes the digest between builds.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/server/routes/v1/signer-routers/signing-router.ts
```diff
@@ -54,6 +81,7 @@ export const registerSignerSigningRouter = async (server: FastifyZodProvider) =>
       const result = await server.services.pkiSigner.sign({
         signerId: req.params.signerId,
         ...req.body,
+        ipAddress: req.realIp,
```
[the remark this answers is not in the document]
carlosmonastyrski: It's deliberate. `docs/self-hosting/configuration/envars.mdx` says outright that leaving it unset means Infisical *"trusts all forwarded-IP headers (legacy behavior, preserved for backwards compatibility)"*, warns that setting it *"prevents IP allowlist bypass via spoofed proxy headers"*, and states plainly that leaving it unset *"is only safe when Infisical is not reachable directly from the internet."*

docs/documentation/platform/pki/code-signing/approvals.mdx
```diff
@@ -82,11 +82,102 @@ Open the Signer's **Approvals** tab and click the pencil icon on the **Approval
...
     You can combine both. For example: `maxSignings=10` with `signing window=1h` issues access good for at most 10 sign calls within one hour of approval, whichever comes first.
 
+    [Scoping](#scoping-an-approval) is not configured here. Whoever opens a request or pre-approval chooses, per request, which signing parameters to lock it to.
   </Step>
 </Steps>
 
 Press **Save policy**. The new policy applies to **new** requests; existing active access keeps its original terms until it expires or is revoked.
 
+<Warning>
+  **The signing window starts when the request is opened, not when it is approved.** A request that asks for 8 hours and waits 3 hours for an approver leaves 5 usable hours, and one approved after 8 hours is already expired. Set the window wide enough to cover how long approvals usually take.
```
sheensantoscapadngan: It seems like our description right now is implying that the signing window is after the approval is granted. Why is it that we start counting down even when the request hasn't been approved yet? 

Image

docs/documentation/platform/pki/code-signing/approvals.mdx
```diff
@@ -82,11 +82,102 @@ Open the Signer's **Approvals** tab and click the pencil icon on the **Approval
...
+
+The remaining parameters (command, signing application, checksum, hostname, OS username) are reported by the signing client. With the [PKCS#11 module](/documentation/platform/pki/code-signing/pkcs11-module) they are captured from the real signing process, so they describe what is happening on well-behaved machines and give approvers useful context. Someone calling the [Sign API](/api-reference/endpoints/code-signing/signers/sign) directly, however, can report whatever values they like. Treat them as identification, not as a security boundary, and reach for **IP address** and **data digest** when you need a limit that holds regardless of how the caller behaves.
+
+<Tip>
+  For release pipelines, the strongest combination is a **data digest** scope on the artifact you just built plus a signature count of **1**. The approval then covers exactly one signature over exactly one file, whatever tooling is used.
+</Tip>
```
sheensantoscapadngan: I don't think we have to explicitly call this out... do other platforms frame it this way as well?

docs/documentation/platform/pki/code-signing/approvals.mdx
```diff
@@ -82,11 +82,102 @@ Open the Signer's **Approvals** tab and click the pencil icon on the **Approval
...
+SCOPE_HASH=$(echo -n "$DIGEST" | xxd -r -p | sha256sum | cut -d' ' -f1)
+
+curl -s -X POST "$INFISICAL_URL/api/v1/cert-manager/signers/$SIGNER_ID/requests" \
+  -H "Authorization: Bearer $TOKEN" \
+  -H "Content-Type: application/json" \
+  -d "{
+    \"justification\": \"Release v2.4.0 Windows installer\",
+    \"requestedSignings\": 1,
+    \"scope\": { \"dataHash\": \"$SCOPE_HASH\" }
+  }"
```
sheensantoscapadngan: I think it's a lot easier for users to understand docs that are more geared towards showing the UI as opposed to the API invocations. If people really do care about the API invocations, then we can easily just link them to the API reference

docs/documentation/platform/pki/code-signing/approvals.mdx
```diff
@@ -82,11 +82,102 @@ Open the Signer's **Approvals** tab and click the pencil icon on the **Approval
...
+
+<Tip>
+  For release pipelines, the strongest combination is a **data digest** scope on the artifact you just built plus a signature count of **1**. The approval then covers exactly one signature over exactly one file, whatever tooling is used.
+</Tip>
+
+### Pre-qualifying a build artifact
+
+A pipeline can ask for approval of one specific artifact before it signs. Build first, hash the artifact, then open a request scoped to that digest: the approval it produces cannot sign anything else, so the approver is signing off on a file rather than on a window of time.
+
+The digest to declare is the SHA-256 of **the bytes your sign call submits**, because that is what Infisical hashes to check the scope. The recipe below therefore applies when the pipeline posts the file itself to the Sign API:
```
sheensantoscapadngan: This might give people the impression that they would have to do this, even though our modules automatically handle this right?

docs/documentation/platform/pki/code-signing/approvals.mdx
```diff
@@ -217,6 +314,9 @@ In-flight sign calls already in progress complete normally; revocation is checke
   <Accordion title="What's the difference between Expired and Revoked?">
     **Expired** is automatic: the window ended or the signature count was reached. **Revoked** is explicit: an Administrator (or the requester via cancel) ended the access before its natural expiry. Either way, signing under it is no longer possible. The status difference shows up in the audit log.
   </Accordion>
+  <Accordion title="My sign call is denied even though I have active access.">
+    If the error mentions your access being scoped to specific signing parameters, the request was scoped and your sign call does not match one of the declared values. Open the request to see the scope, and compare it against what you are signing with now: a different machine, tool, command, IP address, or payload is enough to be denied even when signatures remain. Send the declared command, signing application, hostname or OS username in `clientMetadata` on the sign call, and to sign in a different situation open a new request describing it.
```
sheensantoscapadngan: what `clientMetadata` on the sign call?

frontend/src/hooks/api/signers/types.ts
```diff
@@ -77,6 +78,58 @@ export const getSignerStatusBadgeVariant = (status: SignerStatus) => {
...
+
+export const codeSigningScopeFieldDescriptions: Record<CodeSigningScopeField, string> = {
+  [CodeSigningScopeField.Command]: "The exact signing command that must be issued.",
+  [CodeSigningScopeField.SigningApplication]:
+    "The signing tool executable, as its filename appears on the machine.",
+  [CodeSigningScopeField.SigningApplicationHash]: "Checksum of the signing tool binary.",
+  [CodeSigningScopeField.Hostname]: "The machine the signing requests come from.",
+  [CodeSigningScopeField.OsUsername]: "The operating system account running the signing tool.",
+  [CodeSigningScopeField.IpAddress]: "The address sign calls must come from.",
+  [CodeSigningScopeField.DataHash]: "SHA-256 hash of the payload being signed."
```
bernie-g, line 109: Why are we calling this DataHash in some places and Data Digest in others? Can we standardize on one

frontend/src/pages/cert-manager/components/ScopeFieldsFormSection.tsx
```diff
@@ -0,0 +1,271 @@
...
+import {
+  CodeSigningScopeField,
+  codeSigningScopeFieldDescriptions,
+  codeSigningScopeFieldLabels,
+  codeSigningScopeFieldMatchRules,
+  MONOSPACED_SCOPE_FIELDS,
+  TCodeSigningScope
+} from "@app/hooks/api/signers";
+
+const MAX_COMMAND_LENGTH = 2048;
```
bernie-g, line 43: Is 2048 enough here? I feel like signing commands will exceed this?

frontend/src/pages/cert-manager/components/ScopeFieldsFormSection.tsx
```diff
@@ -0,0 +1,271 @@
...
+const SHA256_HEX = /^[a-fA-F0-9]{64}$/;
+
+const validateScopeValue = (field: CodeSigningScopeField, value: string): string | undefined => {
+  switch (field) {
+    case CodeSigningScopeField.SigningApplicationHash:
+    case CodeSigningScopeField.DataHash:
+      return SHA256_HEX.test(value) ? undefined : "Must be a 64-character SHA-256 hex digest";
+    case CodeSigningScopeField.IpAddress:
+      return z.string().ip().safeParse(value).success ? undefined : "Must be a valid IP address";
+    case CodeSigningScopeField.Command:
```
bernie-g, line 54: Should we actually just check the command hash instead? Then we don't have to worry about length and also people passing in secrets/sensitive args that don't get redacted

backend/src/services/signer/signer-policy-service.ts
bernie-g, line 224: This isnt a change from your PR but I think we should set the start time for the request to when it's approved rather than when the request is opened. Would you mind changing this

backend/src/server/routes/v1/signer-routers/signing-router.ts
```diff
@@ -28,11 +28,38 @@ export const registerSignerSigningRouter = async (server: FastifyZodProvider) =>
...
-            tool: z.string().max(128).optional(),
-            hostname: z.string().max(256).optional(),
-            reportedIp: z.string().max(64).optional()
+            tool: z
+              .string()
+              .max(128)
+              .optional()
+              .describe("The signing application. Compared against a 'signingApplication' scope on the approval."),
+            hostname: z.string().max(256).optional().describe("The machine the sign call is made from."),
+            reportedIp: z
```
bernie-g, line 37: What's the point of this if we use the IP address that the call comes from anyway? Also is this even used anywhere

backend/src/server/routes/v1/signer-routers/signing-router.ts
```diff
@@ -28,11 +28,38 @@ export const registerSignerSigningRouter = async (server: FastifyZodProvider) =>
...
+              .string()
+              .max(256)
+              .optional()
+              .describe("The operating system account running the signing tool."),
+            signingApplicationHash: z
+              .string()
+              .max(64)
+              .optional()
+              .describe("SHA-256 checksum of the signing application binary."),
+            moduleVersion: z
```
bernie-g, line 53: I don't think module version is used anywhere. Correct me if I'm wrong

bernie-g (changes requested):

frontend/src/hooks/api/signers/types.ts
```diff
@@ -77,6 +78,58 @@ export const getSignerStatusBadgeVariant = (status: SignerStatus) => {
...
+  [CodeSigningScopeField.SigningApplicationHash]: "Signing Application Checksum",
+  [CodeSigningScopeField.Hostname]: "Hostname",
+  [CodeSigningScopeField.OsUsername]: "OS Username",
+  [CodeSigningScopeField.IpAddress]: "IP Address",
+  [CodeSigningScopeField.DataHash]: "Data Digest"
+};
+
+export const codeSigningScopeFieldDescriptions: Record<CodeSigningScopeField, string> = {
+  [CodeSigningScopeField.Command]: "The exact signing command that must be issued.",
+  [CodeSigningScopeField.SigningApplication]:
```
bernie-g, line 103: ignore comment placement

When you request signing and it fails we automatically scope to every single possible context including the hash of the thing being signed, which from what I understand can have timestamps or other data that changes frequently so in theory the digest could change on the very next build and then the approval would fail, right? I feel like we should have a way to change the scope of the request that's sent or maybe make it so the approver can loosen the requirements? Let me know your thoughts

frontend/src/pages/cert-manager/components/ScopeFieldsFormSection.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+export const SigningScopeSchema = z
+  .array(
+    z.object({
+      field: z.nativeEnum(CodeSigningScopeField),
+      value: z.string().trim()
+    })
+  )
+  .superRefine((entries, ctx) => {
+    const seen = new Set<CodeSigningScopeField>();
```
bernie-g, line 73: ignore comment placement

Image
Slightly unrelated to this PR but can we reorder this UI to show the most recent approval requests at the top? That's also how we do it for PAM

bernie-g (changes requested):

frontend/src/hooks/api/signers/types.ts
bernie-g: Was this resolved? I see we still refer to it as both

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4714,10 +4715,20 @@ interface GetPkiSigningOperationsEvent {
   type: EventType.GET_PKI_SIGNING_OPERATIONS;
   metadata: {
     signerId: string;
+    signerName: string;
     count: number;
```
bernie-g, line 4728: ignore comment placement

Image
Why can you not select "No limit" here if the policy lets you set no maximum limit

bernie-g (approved):

PR #7540, opened 2026-08-05
feat: allow change policy to apply to machine identities

mathnogueira:
## Context

Until now, every action on secrets made by a machine identity would skip any change policy. With this feature, now machine identities also generate change requests if a policy is in place. 

To prevent this from breaking existing integrations, existing policies allow machine identities to bypass the policy and also allow new ones to be configured to allow machine identities to bypass the policy.

## Steps to verify the change

1. Existing policies are updated to have bypass for machine identities = true
2. All 3 write operations on secrets are bypassed if bypass is true
3. All 3 write operations on secrets create change requests if bypass is false
4. Operations done by user still act the same

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


mathnogueira:
@greptile review again

backend/src/db/migrations/20260805140000_add-committer-identity-to-approval-requests.ts
```diff
@@ -0,0 +1,28 @@
...
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasTable = await knex.schema.hasTable(TableName.SecretApprovalRequest);
+  if (hasTable) {
+    const hasColumn = await knex.schema.hasColumn(TableName.SecretApprovalRequest, "committerIdentityId");
+    if (!hasColumn) {
+      await knex.schema.alterTable(TableName.SecretApprovalRequest, (tb) => {
+        tb.uuid("committerIdentityId").nullable();
+        tb.foreign("committerIdentityId").references("id").inTable(TableName.Identity).onDelete("SET NULL");
```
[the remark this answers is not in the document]
mathnogueira: The same happens to users, so it's consistent with existing behavior.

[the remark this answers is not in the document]
mathnogueira: update the grade of the PR

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
```diff
@@ -1683,7 +1712,7 @@ export const secretApprovalRequestServiceFactory = ({
         notification: {
           type: TriggerFeature.SECRET_APPROVAL,
           payload: {
-            userEmail: user.email as string,
+            userEmail: user?.email ?? "machine-identity",
```
adilsitos: wouldn't it make sense for us to have a different field where we say the name of the machine identity in this case?

backend/src/ee/routes/v1/deprecated-secret-approval-policy-router.ts
adilsitos, line 1: ignore comment placement. I think some e2e test could be added for this feature, just so we make sure that it is working as expected and if any change happens in the future we can be aware.

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
```diff
@@ -2211,7 +2243,7 @@ export const secretApprovalRequestServiceFactory = ({
     void telemetryService
       .sendPostHogEvents({
         event: PostHogEventTypes.SecretApprovalRequestSubmitted,
-        distinctId: user.username ?? user.email ?? actorId,
+        distinctId: user?.username ?? user?.email ?? actorId,
```
adilsitos, line 2250: maybe we could also start to save the type, it may be easier for us in case we need to check these events on posthog (e.g. all events for machine identities)

backend/src/ee/services/secret-approval-policy/secret-approval-policy-fns.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Acts as a TypeScript type guard: when it returns `true`, downstream code
+ * can safely access `policy.id`, `policy.name`, etc.
+ */
+export const shouldApplyPolicy = <T extends { bypassForMachineIdentities: boolean }>(
+  policy: T | undefined,
+  actorType: ActorType
+): policy is T => {
+  if (!policy) return false;
+  if (actorType === ActorType.IDENTITY && policy.bypassForMachineIdentities) return false;
```
adilsitos, line 28: I think we can break a scenario here, because we are only accounting for `ActorType.IDENTITY` but now we allow anything that is authenticated using JWT, so maybe it is possible that an actor type of any other type that used to have a valid response to be blocked. Maybe we can check all the actor types and return true in that case.

backend/src/ee/services/secret-replication/secret-replication-service.ts
```diff
@@ -450,8 +451,8 @@ export const secretReplicationServiceFactory = ({
               destinationFolder.environmentSlug,
               destinationFolder.path
             );
-            // this means it should be a approval request rather than direct replication
-            if (policy && actor === ActorType.USER) {
```
adilsitos, line 454: (got this with claude, but seems to make sense).

469 [correctness] The replication worker now admits identities via shouldApplyPolicy but still inserts `committerUserId: actorId`, killing the job with no user-visible error; line 746 was left on the old `policy && actor === ActorType.USER` check, so the two branches disagree.

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
adilsitos, line 1: on line 330 we still check this: `committerUserID != actorId`, and ignores commiterIdentityID

backend/src/services/secret-v2-bridge/secret-v2-bridge-fns.ts
adilsitos: still hard codes actorId

backend/src/services/secret/secret-service.ts
adilsitos: still hardcodes actorID (but also checks for machine identities)

backend/src/ee/routes/v1/deprecated-secret-approval-policy-router.ts
adilsitos: I belive this are some good scenarios: 
- user secret approval 
- machine identity secret approval with bypass
- machine identity secret approval without bypass -> blocks 
- other type of actor -> continues to work fine

adilsitos, line 1: ignore comment placement. 

Image

Do we have a plan to add this in the future for machine identities as well? Maybe if we have and we don't have a ticket, it might be good to add it.

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
adilsitos: We also check if the user has an admin role there, maybe right now this does not make sense anymore. I believe we could have a specific role in some cases, but this would be a different task as well IMO, maybe worth checking on our channel.

adilsitos (changes requested):
Added some comments regarding places we might need to change

backend/src/ee/routes/v1/deprecated-secret-approval-policy-router.ts
mathnogueira: Not right now. We want to have the solution inlined with the KPI Certificate approvals, there it only allow any machine id to bypass. It's either all machine identities or none.

backend/src/db/migrations/20260805140000_add-committer-identity-to-approval-requests.ts
```diff
@@ -0,0 +1,28 @@
+import { Knex } from "knex";
```
varonix0: Let's combine the two migrations into one

varonix0 (approved):

adilsitos (approved):

PR #7541, opened 2026-08-05
feat: removed all license v2 read mode and old code

akhilmhdh:
## Context

This PR removes the following things as part of cleanup
1. Removes all dual read setup we had for license server
2. Removed unused old license code
3. Standardized and removed unused environment variables as well. We are keeping `LICENSE_V2_SERVICE_KEY` to avoid breaking prod

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/ee/services/license/license-service.ts
```diff
@@ -222,61 +164,29 @@ export const licenseServiceFactory = ({
...
-      currentPlan = v1Plan;
+    // Serve from the License Server, projected into the legacy plan shape so getPlan callers are unchanged.
+    if (!licenseClient) {
+      throw new BadRequestError({ message: "License Server client is not configured" });
     }
+    const entitlements = await licenseClient.getEntitlements({ id: rootOrgId, name: org.name, slug: org.slug });
+    if (!entitlements) {
+      throw new BadRequestError({ message: "License Server entitlements are unavailable" });
+    }
+    const currentPlan = projectV2ToFeatureSet(getDefaultOnPremFeatures(), entitlements);
```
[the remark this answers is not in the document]
akhilmhdh: This is ok. We will be bringing this state back later

backend/src/services/org/org-service.ts
```diff
@@ -831,6 +820,10 @@ export const orgServiceFactory = ({
       };
     });
 
+    // After commit: the org is gone, so stop it billing. Never throws, so a License Server outage can't
+    // fail a deletion that already succeeded, and it stays out of the transaction.
+    await licenseService.cancelOrgSubscription(orgId);
```
Thiago-AS: This will fire for sub-orgs as well, shouldn't be guarded by `deletedOrg.customerId`? From what I saw we always have customerId null, is that the case?

Thiago-AS (changes requested):

backend/src/services/org/org-service.ts
akhilmhdh: Customer id is not generated anymore from our side. It's something only license server do. This will trigger it and it will fail but it will not cause any problem as it has try catch inside it. But let me think we if i can get rid of that as well

Thiago-AS (approved):

PR #7545, opened 2026-08-05
fix(dialog): constrain modal content height

andrewhuhh:
## Context

The v3 `Dialog` primitive did not constrain its height to the viewport, so long forms and dynamically growing content could overflow the visible modal area. Individual consumers then owned inconsistent scrolling behavior, footer actions could move out of reach, and legacy `overflow-visible` overrides prevented sticky footers from using the dialog as their scroll boundary.

This change:

- caps v3 `DialogContent` at `calc(100dvh - 2rem)` and makes it a viewport-safe flex/scroll container;
- uses `w-[calc(100%-2rem)] max-w-lg` as the responsive width contract, preserving viewport gutters while allowing existing `max-w-*` callsite overrides to widen dense dialogs as intended;
- introduces `DialogBody` as the flexible, scroll-owning middle section so headers and footers remain available;
- standardizes the sticky dialog footer treatment and aligns Alert Dialog footer padding;
- removes conflicting `overflow-visible` overrides from v3 Dialog consumers so sticky positioning remains relative to the dialog;
- updates the Share Secret flow to use the shared scroll body;
- updates the bulk-tag secrets dialog (the ENG-5510 downstream validation case) so its table owns bounded scrolling;
- defaults the shared v3 `FilterableSelect` and `CreatableSelect` menus to a fixed `document.body` portal, while preserving explicit caller overrides, so menus do not expand or clip inside Dialog and Sheet scroll containers;
- marks v3 React Select portals and keeps their pointer/focus interactions inside the active Dialog interaction flow;
- migrates the remaining affected certificate-manager member dialog from the inline v2 Filterable Select to the v3 portalled wrapper;
- removes the redundant Duplicate Organization Role refetch so the dialog opens directly at its final form/footer height instead of swapping from a short loading placeholder.

The width audit covered all 147 direct v3 `DialogContent` callsites and the two class-forwarding wrappers. Existing `max-w-*` overrides are compatible with the new contract; no direct callsite applies a conflicting fixed width, Dialog-level minimum width, viewport-defeating maximum, or inline width style.

Related: [ENG-5486](https://linear.app/infisical/issue/ENG-5486/add-max-h-for-dialog-modal-components)

## Screenshots

Viewport-constrained dialog with anchored footer
Dialog footer and bounded content behavior
CleanShot 2026-08-06 at 16 15 07

## Steps to verify the change

1. Run the frontend type-check, ESLint on the changed frontend files, and Prettier check.
2. Open the **Scrollable Content** Dialog story and confirm the middle section scrolls while the header and footer remain available.
3. Open **Organization → Secret Sharing → Share a Secret**, expand the long-form options, and confirm the dialog remains viewport-safe.
4. In Secret Manager, select many secrets and open **Add Tags**. Confirm the table scrolls, the footer remains anchored, and the tag selector menu opens without changing the dialog height.
5. Open **Move Secrets** at a short viewport height and confirm its footer sticks to the dialog rather than the page viewport.
6. Open v3 `FilterableSelect` and `CreatableSelect` controls inside Dialogs (for example, an invite-member flow and **Add Tags**). Confirm each menu portals above the dialog, does not expand or clip inside the scrolling content, and selecting an option does not dismiss the dialog.
7. Open **Organization → Access Management → Identities → Create** and open the Role Filterable Select. Confirm the menu escapes the Sheet's scrolling body without changing the Sheet height and remains selectable.
8. Open a v3 Filterable Select outside an overlay and a consumer with an explicit portal target, if available. Confirm normal positioning still works and the explicit caller configuration takes precedence over the shared defaults.
9. In Certificate Manager, open the application member dialog and confirm the migrated Users, Machine Identities, Groups, and Role selects remain usable within the constrained dialog.
10. Open **Certificate Manager → Applications → application Settings → Configure enrollment methods**. Confirm the `max-w-3xl` dialog renders at its intended width without horizontal clipping or an unnecessary horizontal scrollbar, while retaining viewport gutters on a narrow screen.
11. At the organization level, open **Access Management → Roles → Duplicate Role**. Confirm the complete form and footer appear immediately without a loading-placeholder height jump.
12. Spot-check another migrated Dialog consumer that previously set `overflow-visible` (for example, Add Member, a project-template member dialog, or a KMIP configuration dialog). At a short viewport height, confirm content scrolls within the dialog and the footer remains anchored to the dialog boundary.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/components/v3/generic/Dialog/Dialog.tsx
+++ b/frontend/src/components/v3/generic/Dialog/Dialog.tsx
@@ -55,7 +55,7 @@ function DialogContent({
       <DialogPrimitive.Content
         data-slot="dialog-content"
         className={cn(
-          "fixed top-[50%] left-[50%] z-50 grid translate-x-[-50%] translate-y-[-50%] gap-4 rounded-lg border border-border bg-popover p-6 text-foreground shadow-lg duration-200 outline-none data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
+          "fixed top-[50%] left-[50%] z-50 flex max-h-[calc(100dvh-2rem)] thin-scrollbar translate-x-[-50%] translate-y-[-50%] flex-col gap-4 overflow-y-auto rounded-lg border border-border bg-popover p-6 text-foreground shadow-lg duration-200 outline-none has-data-[slot=dialog-footer]:pb-0 data-[state=open]:animate-in data-[state=open]:fade-in-0 data-[state=open]:zoom-in-95",
           DIALOG_CONTENT_WIDTH_CLASSNAME,
           className
         )}
@@ -88,7 +88,17 @@ function DialogHeader({ className, ...props }: React.ComponentProps<"div">) {
   return (
     <div
       data-slot="dialog-header"
-      className={cn("flex flex-col gap-2 text-left", className)}
+      className={cn("flex shrink-0 flex-col gap-2 text-left", className)}
+      {...props}
+    />
+  );
+}
+
+function DialogBody({ className, ...props }: React.ComponentProps<"div">) {
+  return (
+    <div
+      data-slot="dialog-body"
+      className={cn("min-h-0 thin-scrollbar min-w-0 flex-1 overflow-y-auto", className)}
       {...props}
     />
   );
@@ -98,7 +108,10 @@ function DialogFooter({ className, ...props }: React.ComponentProps<"div">) {
   return (
     <div
       data-slot="dialog-footer"
-      className={cn("flex flex-row flex-wrap justify-end gap-2", className)}
+      className={cn(
+        "sticky bottom-0 z-10 -mx-6 flex shrink-0 flex-row flex-wrap justify-end gap-2 rounded-b-lg border-t border-border bg-container p-4",
+        className
+      )}
       {...props}
     />
   );
@@ -130,6 +143,7 @@ function DialogDescription({
 export {
   Dialog,
   DIALOG_CONTENT_WIDTH_CLASSNAME,
+  DialogBody,
   DialogClose,
   DialogContent,
   DialogDescription,
--- a/frontend/src/components/v3/generic/Dialog/Dialog.tsx
+++ b/frontend/src/components/v3/generic/Dialog/Dialog.tsx
@@ -6,6 +6,11 @@ import { cn } from "../../utils";
 
 const DIALOG_CONTENT_WIDTH_CLASSNAME = "w-lg max-w-[calc(100%-2rem)]";
 
+const isAllowedOutsideInteraction = (target: EventTarget | null) =>
+  Boolean(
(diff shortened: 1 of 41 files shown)


andrewhuhh:
@greptile review

frontend/src/components/v3/generic/Dialog/Dialog.tsx
scott-ray-wilson, line 1: in certain places it looks like inline styling is causing the modal to transition height when opened with these footer changes (duplicate role at org level is an example)

https://github.com/user-attachments/assets/d892476c-f4ba-49db-bf9d-0386347e2adf

scott-ray-wilson, line 1: actually introduced in a previous PR but the tables in some modals now have a undesired scroll

https://github.com/user-attachments/assets/fc69ff33-2ad1-4ebe-aef7-fddcca9a73e5

scott-ray-wilson (dismissed):
looks very nice, two comments

frontend/src/components/v3/generic/Dialog/Dialog.tsx
andrewhuhh: oh hm

andrewhuhh: gonna make a note of this and have codex do a check pass, this is interestin

andrewhuhh: to be fixed in its' own PR https://linear.app/infisical/issue/PLATFOR-639/fix-unnecessary-horizontal-scrollbars-in-dialog-tables

andrewhuhh:
+ new thing to fix: consider live regressions where existing dialogs may have dense horizontal content, which our new width limits for dialogs (copied over from AlertDialog) may break via causing overflow scroll

frontend/src/pages/organization/RoleByIDPage/components/DuplicateOrgRoleModal.tsx
scott-ray-wilson, line 1: Image

duplicating role now fails; this should of been caught by typescript but guess we have incorrect mappings

scott-ray-wilson (dismissed):

frontend/src/pages/organization/RoleByIDPage/components/DuplicateOrgRoleModal.tsx
andrewhuhh: gg

andrewhuhh: let me take a closer look at this

andrewhuhh: fixed

scott-ray-wilson (approved):
lgtm

PR #7550, opened 2026-08-05
fix: enforce domain component order in certificate policies

carlosmonastyrski:
## Context

Domain components form an ordered chain in a distinguished name, so `DC=corp,DC=example,DC=com` and the same labels in any other order name different entities, but a certificate policy matched each DC label independently and accepted either. A `domain_component` rule now holds ordered sequences (`allowed: [["corp","example","com"]]`) and is enforced position by position, with per-component wildcards, and a rule may list several acceptable sequences.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/e2e-test/routes/v1/certificate-policy-domain-components.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+};
+
+const createPolicy = async (name: string, subject: TSubjectRule[]) =>
+  testServer.inject({
+    method: "POST",
+    url: "/api/v1/cert-manager/certificate-policies",
+    headers: { authorization: `Bearer ${jwtAuthToken}` },
+    body: { projectId, name, subject }
+  });
+
```
sheensantoscapadngan, line 37: IGNORE PLACEMENT

perhaps we open the value field with an empty value like the following instead of `*` so that the placeholder guide is displayed? otherwise, user would feel lost initially until after they clear up the field 

Image

backend/e2e-test/routes/v1/certificate-policy-domain-components.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+const getPolicy = async (policyId: string) =>
+  testServer.inject({
+    method: "GET",
+    url: `/api/v1/cert-manager/certificate-policies/${policyId}?projectId=${projectId}`,
+    headers: { authorization: `Bearer ${jwtAuthToken}` }
+  });
+
+const subjectOf = (payload: string) =>
+  (JSON.parse(payload) as { certificatePolicy: { subject: TSubjectRule[] } }).certificatePolicy.subject;
+
```
sheensantoscapadngan, line 47: IGNORE PLACEMENT

Image

I think it makes a lot more sense to have it as just one row instead of two rows for each field, like what we have right now in production... perhaps we can get Scott's thoughts on how to make that work with the current layout? it's confusing that the actual value is below...

backend/e2e-test/routes/v1/certificate-policy-domain-components.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+const domainComponentRule = (subject: TSubjectRule[]) => subject.find((rule) => rule.type === "domain_component");
+
+describe("Certificate policy domain component rules", async () => {
+  const createdPolicyIds: string[] = [];
+
+  beforeAll(async () => {
+    projectId = await createProject();
+  });
+
```
sheensantoscapadngan, line 56: IGNORE PLACEMENT

outside the scope of this PR but perhaps we can include it if its minor... but maybe we don't allow the user to click Continue if the signature algorithm and key algorithm here aren't selected, or if all the required fields aren't filled up. It was a bit odd that I was only able to find out that something was wrong with the form at the very last step. 

Image

backend/e2e-test/routes/v1/certificate-policy-domain-components.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+        headers: { authorization: `Bearer ${jwtAuthToken}` }
+      });
+    }
+
+    await testServer.inject({
+      method: "DELETE",
+      url: `/api/v1/projects/${projectId}`,
+      headers: { authorization: `Bearer ${jwtAuthToken}` }
+    });
+  });
```
sheensantoscapadngan, line 72: IGNORE PLACEMENT

also outside the scope of this PR but i'm not seeing any highlighted fields? 
Image

backend/e2e-test/routes/v1/certificate-policy-domain-components.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  const trackPolicy = (payload: string) => {
+    const { certificatePolicy } = JSON.parse(payload) as { certificatePolicy: { id: string } };
+    createdPolicyIds.push(certificatePolicy.id);
+    return certificatePolicy.id;
+  };
+
```
sheensantoscapadngan, line 79: IGNORE PLACEMENT

hmmm "template policy" terminology seems off 
Image

sheensantoscapadngan (approved):

PR #7554, opened 2026-08-05
feat(secrets): redesign move destination picker

andrewhuhh:
## Context

The move dialog made source and destination locations hard to compare, nested folders appeared as flat paths, and creating a destination required leaving the workflow first.

This change redesigns the shared move destination picker on current `main`:

- replaces the legacy creatable select with the v3 `Combobox`;
- uses fixed stacked source and destination location rows separated by a **Move to** divider aligned near the first fifth;
- uses consistent absolute paths in the fields, keeps project root permanently available, labels it with the project name in the picker, and still submits `/`;
- gives existing nested folders tree guides plus ancestor context during search;
- accepts typed single-level or nested paths, clears stale selections when the typed path changes, and creates only the missing segments in each applicable environment;
- keeps overwrite controls with the actions while preserving direct-parent permission checks, incomplete multi-environment handling, cyclic-move prevention, query invalidation, and secret-rotation moves.

Related ticket: [SECRETS-496](https://linear.app/infisical/issue/SECRETS-496/add-a-way-to-create-new-folder-in-the-move-secrets-dialog)

## Screenshots

The move dialog presents source and destination as fixed stacked rows separated by a **Move to** divider aligned near the first fifth:

Move dialog with fixed stacked location rows

The same row structure remains intact on narrower viewports:

Move dialog with fixed stacked rows at a narrow viewport

## Steps to verify the change

1. Open a Secrets Management project overview, select secrets or a folder, and open the move dialog.
2. Confirm the top row shows the current environment and absolute folder path while the bottom row provides the destination environment and folder controls.
3. Resize the viewport and confirm the rows remain stacked with the **Move to** divider aligned near the first fifth and no Source/Destination headings or directional arrow.
4. Open the destination picker and confirm project root remains visible while searching and is labeled with the project name.
5. Search for an existing nested destination and confirm its indentation, tree guides, and ancestor path are visible; select it and confirm the field renders the absolute path. Then type a different path and confirm the prior selection clears and Move becomes disabled.
6. Type a unique nested path such as `apps/new service`; confirm it normalizes to `apps/new-service` and appears as a **New Folder** option.
7. Create it and confirm all missing path segments are created only in the applicable environments, then selected without moving resources until the explicit move action.
8. With a path that exists in only some applicable environments, confirm creation fills only the missing environments before selection.
9. Try a destination inside a selected folder and confirm creation is disabled with the cyclic-move reason. Verify a user lacking create permission sees the affected environment named instead.
10. Exercise single- and multi-environment moves, including overwrite confirmation and a rotation-only selection.
11. Run `cd frontend && npm run lint && npm run type:check`.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


andrewhuhh:
@greptile review

PR #7592, opened 2026-08-07
feat: allow to edit cert details on manual renewals

carlosmonastyrski:
## Context

Manual certificate renewal now lets you change the certificate at the same time. Previously renewing a certificate produced an identical copy with a new key, so if a service had picked up an extra hostname or needed a longer validity, you had to issue a brand new certificate and wire it up from scratch. Now the renewal wizard opens with everything prefilled from the current certificate and you change only what should differ: the names it covers, its subject details, how long it lasts, or its key usages. You also choose how the key pair is handled, which is the part that matters most in practice: you can keep the existing key so anything pinned to it keeps working, generate a fresh one, or supply your own signing request. Anything you do not touch carries over untouched, including metadata and any destinations the certificate is being synced to, and the certificate profile's policy still validates the result, so a renewal can never produce something the profile would have rejected in the first place.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -2444,926 +2032,89 @@ export const certificateV3ServiceFactory = ({
...
-          basicConstraints: { isCA: true, pathLength: originalCert.pathLength ?? undefined }
+    if (metadata && projectPermission) {
+      ForbiddenError.from(projectPermission).throwUnlessCan(
+        ProjectPermissionCertificateActions.Edit,
+        subject(ProjectPermissionSub.Certificates, {
+          commonName: certificate.commonName,
+          altNames: certificate.altNames?.split(",").map((s) => s.trim()),
+          serialNumber: certificate.serialNumber,
+          metadata
         })
```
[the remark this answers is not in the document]
carlosmonastyrski: Not quite: `assertCanEditCertificate` already authorizes against the certificate's **persisted** metadata, which it reads from `resourceMetadataDAL`. The check at 2061–2066 is an *additional* gate on the replacement metadata, so the caller has to satisfy both the current and the resulting state. Removing it would be the actual bug.

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateRenewalModal.tsx
```diff
@@ -15,66 +155,419 @@ type Props = {
...
+  const stepKeys = useMemo<RenewalStepKey[]>(() => {
+    if (keySource === CertificateRenewalKeySource.Csr) return ["setup", "csr"];
+    const keys: RenewalStepKey[] = ["setup"];
+    if (constraints.shouldShowSubjectSection || constraints.shouldShowSanSection)
+      keys.push("subject");
+    keys.push("options");
+    return keys;
+  }, [keySource, constraints.shouldShowSubjectSection, constraints.shouldShowSanSection]);
+
+  const { step, setStep, currentStepKey, goBack, goNext, onFormInvalid } = useWizardSteps({
```
saifsmailbox98, line 360: ignore comment placement

if you add a SAN row and leave it blank, you can move to the next step, but it gives this error: fix the highlighted fields, but nothing is highlighted.

can we validate each step?

backend/src/services/certificate/certificate-dal.ts
```diff
@@ -753,21 +785,27 @@ export const certificateDALFactory = (db: TDbClient) => {
...
       if (result.caId) {
         caType = internalCaId ? "internal" : "external";
       }
 
-      return { ...rest, caType } as TCertificateWithRequestDetails;
+      return {
+        ...rest,
+        caType,
+        hasPrivateKey: Boolean(privateKeyRef),
+        keyUsages: normalizeKeyUsagesForResponse(rest.keyUsages),
```
saifsmailbox98: doesn't this change what the API returns?

also check how they are being rendered now on the cert details page

backend/src/server/routes/v1/certificate-router.ts
saifsmailbox98, line 1188: should we log what actually changed here?

saifsmailbox98 (approved):

PR #7594, opened 2026-08-07
fix(ui): correct tooltip behavior in select menus

andrewhuhh:
## Context

The policy-detail tooltip in the Add Machine Identity sheet could render behind the open Additional Privileges select menu. The shared React Select portal used an effectively unbounded `z-index: 99999`, which placed it above the v3 tooltip overlay.

This change:

- moves the shared v3 React Select menu portal to overlay tier `60`, above sheets and dialogs but below tooltips
- enforces non-hit-testable tooltip content inside modal layers so Radix cannot override it with inline pointer-event behavior
- adds a Storybook composition with an open select menu and tooltip to document the expected overlay order

This affects shared v3 React Select menus and tooltips without changing selection behavior, data, permissions, or APIs.

[PLATFOR-641](https://linear.app/infisical/issue/PLATFOR-641/tooltip-opens-behind-dropdown)

## Screenshots

No shareable visual artifact is attached yet. Outstanding capture:

- **State:** Add Machine Identity sheet with the Additional Privileges menu open and a policy info tooltip visible above the menu.
- **Prerequisites:** A Secret Management project and permission to open Project Access Control.
- **Navigation:** Project Access Control → Machine Identities → Add Machine Identity to Project.
- **Interaction:** Open Additional Privileges and hover a policy row's info icon.
- **Expected result:** The tooltip renders above the select menu and remains non-hit-testable.

## Steps to verify the change

1. Open a Secret Management project's Project Access Control page.
2. Select **Machine Identities** and open **Add Machine Identity to Project**.
3. Open the **Additional Privileges** select.
4. Hover each policy row's info icon from multiple directions.
5. Confirm the tooltip renders above the select menu and does not prevent interaction with the menu.
6. Run `make reviewable-ui`.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/components/v3/generic/ReactSelect/FilterableSelect.stories.tsx
+++ b/frontend/src/components/v3/generic/ReactSelect/FilterableSelect.stories.tsx
@@ -1,11 +1,15 @@
 import { useState } from "react";
+import { components, OptionProps } from "react-select";
 import type { Meta, StoryObj } from "@storybook/react-vite";
+import { InfoIcon } from "lucide-react";
 
 import { Field, FieldDescription, FieldError, FieldLabel } from "../Field";
+import { Tooltip, TooltipContent, TooltipTrigger } from "../Tooltip";
 import { FilterableSelect } from "./FilterableSelect";
 
 type Option = { label: string; value: string };
 type RegionOption = Option & { region: string };
+type PolicyOption = Option & { description: string };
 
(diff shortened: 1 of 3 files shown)


scott-ray-wilson (approved):
LGTM

PR #7610, opened 2026-08-10
fix(secret-sync): honour disableSecretDeletion in the Azure DevOps sync

### LINKED ISSUE #7609: Azure DevOps sync ignores the Disable Secret Deletion option
### Product

Secret Sync

### Describe the bug

`disableSecretDeletion` is enforced individually by each secret sync rather than centrally in `secret-sync-fns.ts`. Of the 44 destinations with a `*-sync-fns.ts`, 43 read the flag somewhere in their sync path. `azure-devops` is the only one whose sync functions never reference it at all:

```console
$ grep -rn "disableSecretDeletion" backend/src/services/secret-sync/azure-devops/
$ 
```

The comparable destinations all guard their prune step, in one of a few shapes:

```ts
// vercel, 1password, databricks, ... 
if (secretSync.syncOptions.disableSecretDeletion) return;

// aws-parameter-store, aws-secrets-manager
if (syncOptions.disableSecretDeletion) return { createdSecretKeys, updatedSecretKeys, deletedSecretKeys };

// gcp
if (secretSync.syncOptions.disableSecretDeletion) continue;
```

### Why it matters here specifically

`azure-devops-sync-fns.ts` builds its payload from the Infisical secret map alone, then PUTs the whole variable group:

```ts
const variables: Record<string, { value: string; isSecret: boolean }> = {};
for (const [key, secret] of Object.entries(secretMap)) {
  if (secret?.value !== undefined) {
    variables[key] = { value: secret.value, isSecret: true };
  }
}
...
await request.put(url, {
  name: groupName,
  description: groupName,
  type: "Vsts",
  variables,
  variableGroupProjectReferences: [...]
}, { ... });
```

Nothing reads the existing group's variables first, so whatever the destination already had is not carried into the request body. The [Update variable group reference](https://learn.microsoft.com/en-us/rest/api/azure/devops/distributedtask/variablegroups/update?view=azure-devops-rest-7.1) describes the body field as "Sets variables contained in the variable group", which reads as replacement rather than merge on a PUT.

I could not verify the delete-on-omit behaviour against a real Azure DevOps organisation, so I am flagging rather than asserting it. Either reading is still a bug:

- If the PUT replaces, a user who enables "Disable Secret Deletion" still loses every variable in the group that is not managed by Infisical, which is the exact scenario the option exists to prevent.
- If the PUT merges, then this destination has no prune step at all, and the option is offered in the UI while doing nothing here.

### To Reproduce

1. Create an Azure DevOps secret sync and enable **Disable Secret Deletion** in the sync options.
2. Add a variable directly in the destination variable group that does not exist in the Infisical environment.
3. Trigger a sync.
4. Check whether the manually added variable is still present.

### Expected behavior

Same as every other destination: when `disableSecretDeletion` is set, a sync should not remove destination secrets that are absent from Infisical.

### Additional context

Checked `azure-entra-id-scim` too, since it also has no reference to the flag in its sync functions, but it does not appear to have a prune step, so I do not think it is affected.

The wider observation is that the flag is reimplemented 43 separate times in at least seven different shapes. Enforcing it once around the prune step would make this class of gap impossible rather than relying on each new destination remembering. Happy to open a PR for the `azure-devops` case on its own, or for the centralisation if that is wanted, but the latter touches every destination so it seemed worth asking first.

omlahore:
## Context

Fixes #7609

`disableSecretDeletion` is enforced individually by each secret sync rather than centrally. Of the 44 destinations with a `*-sync-fns.ts`, 43 read it somewhere in their sync path. `azure-devops` was the only one that never did:

```console
$ grep -rn "disableSecretDeletion" backend/src/services/secret-sync/azure-devops/
$
```

That matters here because the destination sets the variable group's whole variable collection rather than patching it. The body was built from the Infisical secret map alone, so any variable already in the group that Infisical does not manage was dropped on every sync, whether or not the user had enabled the option.

The fix is small because `$getEnvGroupId` was already fetching each group's `variables` and throwing them away. It now returns them, and `syncSecrets` seeds the request body with the unmanaged ones when deletion is disabled.

## The secret round-trip, since this is the part worth checking

Azure DevOps does not return secret variable values on a GET; they come back as `null`. So the obvious worry with carrying variables through is that writing `null` back would blank a real secret.

It does not: the API treats a `null` value on update as "keep the existing one". That is the behaviour the Terraform provider relies on, and it is what made [microsoft/terraform-provider-azuredevops#1433](https://github.com/microsoft/terraform-provider-azuredevops/issues/1433) behave the way it does. I have called this out in a comment next to the merge so the next reader does not have to rediscover it.

## Steps to verify the change

1. Create an Azure DevOps secret sync and enable **Disable Secret Deletion**.
2. Add a variable directly in the destination variable group that does not exist in the Infisical environment. Add both a plain and a secret one.
3. Trigger a sync.
4. Before: both are gone. After: both survive, and the secret keeps its value.
5. Turn the option off and sync again to confirm pruning still happens.

I could not run step 4 myself, since I do not have an Azure DevOps organisation to sync into. The change is verified by `type:check` and `eslint --max-warnings 0`, and by reading the API contract, but someone with a real org should confirm the round-trip before this merges.

## Type

- [x] Fix

## Checklist

- [x] Tested locally (`npm run type:check`, `eslint --max-warnings 0`); not exercised against a live Azure DevOps organisation, see above
- [ ] Updated docs (not needed, this makes an existing documented option work as described)

## Two things deliberately left out of scope

**Pruning ignores `keySchema` here.** With the option off, this destination still drops every unmanaged variable, where other destinations only prune keys matching their schema. That is arguably a second bug, but fixing it changes behaviour for existing users, so it seemed wrong to fold in silently.

**The flag is reimplemented 43 times in seven different shapes.** Enforcing it once around the prune step would make this class of gap structurally impossible. Happy to do that instead if preferred, though it touches every destination and felt like a separate conversation.

### DIFF
--- a/backend/src/services/secret-sync/azure-devops/azure-devops-sync-fns.ts
+++ b/backend/src/services/secret-sync/azure-devops/azure-devops-sync-fns.ts
@@ -73,10 +73,10 @@ export const azureDevOpsSyncFactory = ({ kmsService, appConnectionDAL }: TAzureD
 
     for (const group of response.data.value) {
       if (group.name === environmentName) {
-        return { groupId: group.id.toString(), groupName: group.name };
+        return { groupId: group.id.toString(), groupName: group.name, variables: group.variables ?? {} };
       }
     }
-    return { groupId: "", groupName: "" };
+    return { groupId: "", groupName: "", variables: {} };
   };
 
   const syncSecrets = async (secretSync: TAzureDevOpsSyncWithCredentials, secretMap: TSecretMap) => {
@@ -94,7 +94,11 @@ export const azureDevOpsSyncFactory = ({ kmsService, appConnectionDAL }: TAzureD
 
     const { accessToken, orgName, isOAuth } = await getConnectionAuth(secretSync);
 
-    const { groupId, groupName } = await $getEnvGroupId(
+    const {
+      groupId,
+      groupName,
+      variables: existingVariables
+    } = await $getEnvGroupId(
       accessToken,
       orgName,
       secretSync.destinationConfig.devopsProjectId,
@@ -103,6 +107,20 @@ export const azureDevOpsSyncFactory = ({ kmsService, appConnectionDAL }: TAzureD
     );
 
     const variables: Record<string, { value: string; isSecret: boolean }> = {};
+
+    // The update below sets the group's whole variable collection, so anything left out of the
+    // request body is dropped. Carry over the variables Infisical does not manage when deletion
+    // is disabled, otherwise this destination would remove them regardless of the option.
+    // Azure DevOps returns secret values as null, and treats a null value on update as "keep the
+    // existing one", so round-tripping a secret variable here does not clear it.
+    if (secretSync.syncOptions.disableSecretDeletion) {
+      for (const [key, variable] of Object.entries(existingVariables)) {
+        if (!(key in secretMap)) {
+          variables[key] = variable;
+        }
+      }
+    }
+
     for (const [key, secret] of Object.entries(secretMap)) {
       if (secret?.value !== undefined) {
         variables[key] = { value: secret.value, isSecret: true };


omlahore:
@mathnogueira tagging you because you have most of the recent commits under `backend/src/services/secret-sync`, rather than guessing at who owns this.

All seven workflows on this PR are sitting at `action_required`, so none of them has run: backend tests, DB schema validation, the API breaking-change check, lint and the BDD suite. The green marks are Socket, Greptile, CLA, GitGuardian and Snyk, which are third-party apps that bypass that gate, so they do not tell you the code was tested.

Could someone approve the run? The change makes the Azure DevOps sync honour `disableSecretDeletion`, so it stops removing destination secrets a user explicitly asked it to leave alone.

PR #7615, opened 2026-08-11
fix: return inputs from PATCH `/dynamic-secrets`

mathnogueira:
## Context

We used to return the `inputs` on the response of the `PATH /dynamic-secrets` but it got removed. This reintroduces it as this is a required field for the terraform provider.

But it requires user to have the `ReadRootCredentials` permission, otherwise we keep returning the `inputs` as `null` to prevent leaking those values to unauthorized identities.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


mathnogueira:
@greptile review it again

varonix0 (approved):

PR #7634, opened 2026-08-11
fix(logging): change health check logs from info to debug level

### LINKED ISSUE #7392: /api/status logs a full event-loop histogram on every request, making health probes ~45% of log volume
## Describe the bug

`GET /api/status` writes two `logger.info` lines on **every** request — a summary line and a full `JSON.stringify(histogram, null, 2)` dump of the event-loop delay histogram:

https://github.com/Infisical/infisical/blob/main/backend/src/server/routes/index.ts

```ts
handler: async () => {
  const cfg = getConfig();
  const serverCfg = await getServerCfg();

  const meanLagMs = histogram.mean / 1e6;
  const maxLagMs = histogram.max / 1e6;
  const p99LagMs = histogram.percentile(99) / 1e6;

  logger.info(`Event loop stats - Mean: ...`);
  logger.info(`Raw event loop stats: ${JSON.stringify(histogram, null, 2)}`);
  ...
```

`/api/status` is the natural health endpoint, so in any orchestrated deployment it is called by a liveness/readiness probe on a fixed interval. In our Kubernetes deployment (the official `infisical-standalone` Helm chart) the readiness probe is `GET /api/status` with `period=5s`, which turns this into a continuous log stream.

Introduced in #3282 (merged 2025-03-20), still present in `v0.162.12`.

## Impact

Measured on a single-replica self-hosted instance, `infisical/infisical:v0.162.10`, one hour of container logs:

| | |
|---|---|
| `/api/status` requests | 720/h (readiness probe, every 5s) |
| Log records produced by them | 1440/h — ~34,500/day |
| Bytes of the `Raw event loop stats` dump alone | 826 KB/h |
| Total log bytes for the container | 1.81 MB/h |
| **Share of the container's log volume** | **~45%** |

That is ~20 MB/day, ~0.6 GB/month of log ingest per replica for a single message. On a hosted log backend this is a real line item, and it buries the lines that actually matter — the same hour contained only 347 `getPlan` lines and 173 `injectPermission` lines.

There is no way to turn it off: the calls are unconditional, not gated by `LOG_LEVEL` or any env var.

## The logged values are also not usable as-is

The histogram is created once at startup and never `.reset()`, so every field is cumulative over the process lifetime:

```json
{
  "count": 8000542,
  "min": 376832,
  "max": 5553258495,
  "mean": 20116550.6745398,
  "percentiles": { "50": 20103168, "99.99998807907104": 5549064192 }
}
```

- `count: 8000542` × 20 ms resolution ≈ 44.4 h — exactly the pod's age. Nothing is ever windowed.
- `mean` (20.1 ms) and `p50`/`p75` sit on the sampling-resolution floor, i.e. the loop is idle. They will read ~20 ms forever regardless of load.
- `max: 5.55 s` is one stall that happened at some unknown point in the last 44 hours. It is frozen in `max` and in the top percentiles for the rest of the process's life and cannot decay.

So the numbers cannot show a trend, cannot be alerted on, and become less informative the longer the process runs — while costing half the log volume.

## Suggested fix

Any of these would resolve it; (1) is the smallest:

1. Demote both lines to `logger.debug`. One-line change, keeps the diagnostic available for anyone who raises `LOG_LEVEL`.
2. Remove them. The same data is already exported properly when telemetry is enabled: `initTelemetryInstrumentation` registers `RuntimeNodeInstrumentation`, which emits `nodejs.eventloop.delay.{min,max,mean,stddev,p50,p90,p99}` as real time-series via the Prometheus/OTLP exporter — windowed, alertable, and not tied to how often a probe hits the health endpoint.
3. If the logging is worth keeping, move it out of the request handler onto a fixed timer and `histogram.reset()` after each emit, so the values describe the last interval rather than the whole process lifetime.

Happy to open a PR for (1) if that is the direction you'd prefer.

## Environment

- Infisical `v0.162.10` (verified the code is unchanged in `v0.162.12`)
- Self-hosted, `infisical-standalone` Helm chart 1.10.0, Kubernetes

vedgharat:
## Context

Fixes #7392 — the `/api/status` health check endpoint was logging event-loop performance statistics at `info` level on every request. Since health checks are typically polled frequently (by load balancers, orchestrators, and uptime monitors), this produced excessive log output in production, drowning out more meaningful logs and increasing log storage/egress costs.

These event-loop stats (`Mean`, `Max`, `p99` lag, and the raw histogram) are only useful for performance debugging and are not actionable during normal operation. This PR moves both log statements from `logger.info` to `logger.debug`, so they only appear when debug logging is explicitly enabled.

**Behavior before:** Every health check request emitted two `info`-level logs (event loop stats summary + full raw histogram).
**Behavior after:** Both statements log at `debug` level — silent in normal production logging, visible only when debug logging is enabled.

## Screenshots

N/A (logging-only change, no UI/UX impact)

## Steps to verify the change

1. Confirm the two changed log statements in `backend/src/server/routes/index.ts` now use `logger.debug(...)`.
2. `npm run type:check` — no new type errors introduced by this change.
3. `npm run lint` — passes on the changed file.
4. `npm run test:unit` — 995 tests pass; the 56 failing test files are pre-existing and caused by two packages (`tsyringe`, `@peculiar/asn1-pkcs8`) not declared in `package.json`, unrelated to this change.
5. With logging at default level, the `/api/status` endpoint no longer emits event-loop stat logs; they appear only when debug logging is enabled.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the conventional commit format: `fix(logging): change health check logs from info to debug level`
- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/index.ts
+++ b/backend/src/server/routes/index.ts
@@ -4136,13 +4136,13 @@ export const registerRoutes = async (
       const maxLagMs = histogram.max / 1e6;
       const p99LagMs = histogram.percentile(99) / 1e6;
 
-      logger.info(
+      logger.debug(
         `Event loop stats - Mean: ${meanLagMs.toFixed(2)}ms, Max: ${maxLagMs.toFixed(2)}ms, p99: ${p99LagMs.toFixed(
           2
         )}ms`
       );
 
-      logger.info(`Raw event loop stats: ${JSON.stringify(histogram, null, 2)}`);
+      logger.debug(`Raw event loop stats: ${JSON.stringify(histogram, null, 2)}`);
 
       return {
         date: new Date(),


vedgharat:
Can anyone review and approve this?

PR #7646, opened 2026-08-12
feat(agent-policies): broker credentials on the intersection of agent and user policies

saifsmailbox98:
## Context

Today, building an agent means using agent proxies and proxied services, and the agent's own machine identity carries the permissions. Nobody checks what the *human* using that agent is allowed to do, so access ends up governed by an allowlist on the agent's side, and every permission is configured twice: proxy access on the proxied service for the user, plus secret-read access for the agent proxy's identity.

This replaces that with an intersection. An agent gets its own policies, a person gets theirs, and a brokered request has to satisfy both. The agent builder can be sure the agent cannot do more than it was built for, and a person can't borrow an agent to exceed what they hold themselves.

There is no set arithmetic and nothing is precomputed. A request is allowed when it matches at least one rule on an agent policy naming that agent **and** at least one rule on a user policy naming that user, so a user rule of `GET` narrows an agent rule of `Any` on the same host. That comparison is only meaningful per request, at the proxy.

**Agent Proxy:** a new first-class resource in Networking, alongside gateways and relays. It enrols through the existing `resource_auth_methods` registry (token only, no AWS arm) and authenticates as `AGENT_PROXY_ACCESS_TOKEN` / `ActorType.AGENT_PROXY`, the same shape relays and KMIP servers already use, so it needs no machine identity of its own and `tokenVersion` revokes it. It accepts inbound connections by design, so no relay is involved. Gateways were the obvious home for this and don't fit: they accept no inbound connections and pin one target per connection in the client cert, which a forward proxy can't express.

**Policies:** agent policies carry the agents they apply to, a target, the credentials to broker, and rules; user policies carry users, a target and rules. Both are project-scoped. A target seeds the credential slots and the rules, so the author only picks a secret per slot: env var names, placeholder values and injection surfaces are all derived and never asked for. Writing a credential requires ReadValue on the referenced secret, the check `proxied-service-service.ts` already performs, so nobody can attach a value they can't read. Rules are unordered, since ordering changes nothing when every rule is an allow.

**Sessions:** an agent calls `POST /agent-sessions` with a user's email (the agent is trusted to have verified it, which is all a Slack bot gets) and receives a session token, the placeholder env vars, and the MITM root CA. The token has no expiry; revocation is structural instead, because `resolveSession` re-checks the agent flag, project membership and policies on every call, so losing any of them ends the session on the next refresh. Only an enrolled agent proxy may redeem a session.

**Identities:** `isAgent` on machine identities. Only an agent may start a session on someone's behalf, and only agents appear in a policy's agent selector.

**Audit:** every brokered, blocked and passed-through request is recorded against both the agent and the user, which is the part that makes "the agent acted as Saif" answerable.

**Frontend:** an Agent Proxies tab under Networking (create, enrollment token, revoke, delete, health), a project-level Agent Proxy page with Agent Policies and User Policies tabs, create/edit sheets for both, and the Agent toggle on the machine identity form. Credentials are chosen with a browse-secrets popover modelled on the secret reference wizard.

Proxied services and today's agent proxy are untouched.

Companion CLI PR: https://github.com/Infisical/cli/pull/361

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


saifsmailbox98:
lisbon

PR #7663, opened 2026-08-14
feat(ui): migrate project roles access control to v3

andrewhuhh:
## Context

Project role access control still mixed legacy v2 overlays, form controls, dialogs, permission-tree components, and page headers into otherwise migrated Access Control surfaces. This made role management and permission editing inconsistent with the current v3 interaction and accessibility contracts.

This change migrates the ENG-5544-owned project role experience and its adjacent shared permission surfaces to v3 while preserving existing behavior:

- keeps role search, sorting, pagination, built-in/custom distinctions, plan gates, permission-gated actions, and project-type-specific subjects;
- adds keyboard-operable sortable headers and role rows while retaining independent row actions;
- uses v3 role details, edit, rename, and typed-delete flows with pending state and project-type-aware mutations;
- uses the canonical v3 `PageHeader` on role details and keeps the permission editor read-only without edit access;
- migrates policy template selection, conflict preview, and Vault policy import to v3 Dialog, Sheet, Field, Accordion, select, and semantic feedback compositions;
- migrates the shared permission tree, permission simulation, folder/role nodes, access-denied states, and permission wrappers to the v3 contracts;
- replaces the legacy v2 restricted-access banner with the v3 platform component and adds component coverage for it.

The change preserves dirty/save state, validation, imported-policy previews, and existing permission behavior. It is frontend-only: no backend APIs, schemas, authorization rules, routes, or dependencies are changed.

Related ticket: [ENG-5544](https://linear.app/infisical/issue/ENG-5544/migrate-project-roles-access-control-surface-to-v3)

## Screenshots

CleanShot 2026-08-19 at 16 34 50@2x
CleanShot 2026-08-19 at 16 34 59@2x
CleanShot 2026-08-19 at 16 35 07@2x
CleanShot 2026-08-19 at 16 35 19@2x
CleanShot 2026-08-19 at 16 35 26@2x
CleanShot 2026-08-19 at 16 35 36@2x

## Steps to verify the change

1. Run `make reviewable-ui` from the repository root.
2. Open a project with custom roles and navigate to `/organizations/{orgId}/projects/secret-management/{projectId}/access-management?selectedTab=roles`.
3. Verify search, sort, pagination, built-in/custom badges, keyboard row navigation, and permission-gated row actions.
4. Open a custom role and verify the v3 header, edit/rename validation, duplicate-role compatibility, and typed delete confirmation.
5. Change role permissions, confirm dirty/save state, save, reload, and verify the persisted permissions.
6. Add a policy template and import a Vault policy; verify conflict handling and both previews before saving.
7. Repeat with restricted permissions and across applicable project types to verify read-only behavior, access-denied states, and project-type-specific subjects.

## Type

- [ ] Fix
- [x] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [x] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format (scope is optional, e.g., `fix: prevent crash on sync` or `fix(api): handle null response`).
- [x] Tested locally


andrewhuhh:
@greptile review

frontend/src/pages/project/RoleDetailsBySlugPage/components/EditProjectRoleDialog.tsx
scott-ray-wilson, line 1: I feel like it should be anchored right if it is this order or we should swap their order? 

Image

frontend/src/pages/project/AccessControlPage/components/ProjectRoleListTab/components/ProjectRoleList/ProjectRoleList.tsx
scott-ray-wilson, line 1: hmmm I'm not sure how I feel about the dual table; why not just use a column or badge to indicate?

Image

frontend/src/components/v3/platform/AccessRestrictedBanner.tsx
scott-ray-wilson, line 1: why did we add this back with stories? we replaced this in my recent PR 🤔

frontend/src/components/v3/platform/SecretManagerResources/SecretManagerResources.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+            resource.color.tileClassName
+          )}
+        >
+          <resource.Icon className="size-4" />
+        </span>
+        <div className="min-w-0 flex-1">
+          <div className="flex flex-wrap items-baseline gap-2">
+            <span className="text-sm font-medium text-foreground">{resource.name}</span>
+            <span className="font-mono text-xs text-accent">{resource.slug}</span>
+            <span className="font-mono text-xs text-muted">{resource.permissionSubject}</span>
```
scott-ray-wilson: where are we using this component and it seems overly verbose with name, slug and permission subject? 

Image

frontend/src/pages/project/RoleDetailsBySlugPage/components/EditProjectRoleDialog.tsx
scott-ray-wilson, line 1: I'm not sure this is related to anything in this PR or recent changes but the autofill text is now displayed over the input text instead of replacing it? 

Image

scott-ray-wilson, line 1: if I delete a role with a dirty form I still see the discard changes confirmation 

https://github.com/user-attachments/assets/56779d3c-0a88-468b-820d-2268eca075ac

scott-ray-wilson:
**Cert-manager custom roles: this PR makes them un-editable and un-deletable, while Duplicate can still mint new ones**

The migration newly threads `projectType` into the role mutations on the details page: policy save (`RolePermissionsSection.tsx`), delete (`RoleDetailsBySlugPage.tsx`), and rename (`EditProjectRoleDialog.tsx`). For Certificate Manager projects that reroutes PATCH/DELETE to `/api/v1/cert-manager/access/roles/:roleId`, whose handlers unconditionally throw `CERT_MANAGER_CUSTOM_ROLE_ERROR` (`backend/src/server/routes/v1/cert-manager-access-routers/roles-router.ts`). At merge-base these calls omitted `projectType` and hit the generic `/api/v1/projects/:projectId/roles` endpoints, which have no project-type guard, so all three actions worked.

**Is this state actually reachable? Yes:**

1. Legacy custom roles exist as data. The stub router only landed in May 2026; before that, cert-manager projects used the generic roles surface where custom role creation worked. No migration cleaned those rows up, and the generic v1 endpoints still accept custom-role writes for cert projects today (API/Terraform).
2. There is an in-product click path. `MemberRoleDetailsSection.tsx:245` and `IdentityRoleDetailsSection.tsx:257` link any assigned role to `/roles/$roleSlug`, so a member or identity still assigned a legacy custom role routes you to the details page even though the roles list hides custom roles in cert-manager.
3. The page itself anticipates the state: `RoleDetailsBySlugPage.tsx:212` renders the "Custom roles act as Member in Certificate Manager" alert for exactly `isCertManager && isCustomRole`, with Save/Edit/Delete still enabled next to it.

**Why it matters despite being an edge population:** the details page was the only cleanup path for these legacy roles (the list hides them), and delete now permanently 400s with the button enabled. Meanwhile the details-page Duplicate item is gated only by `isCustomRole` + create permission (no `!isCertManager` gate, unlike `ProjectRoleList`) and omits `projectType`, so it succeeds via the generic endpoint: one legacy role can create new invisible custom roles that then can't be saved, renamed, or deleted. That combination is a one-way door.

**Suggested fix, either direction resolves it:**

- Keep legacy roles operable: drop the `projectType` arg from these three mutations so they stay on the generic endpoints, or
- Drop support deliberately: disable Save/Edit/Delete for cert-manager custom roles and add the missing `!isCertManager` gate to the details-page Duplicate item.

Longer term it may be worth resolving `projectType` inside the roles hooks (and including it in the query keys) so call sites can't disagree; `DuplicateProjectRoleModal`'s `useGetProjectRoleBySlug` currently shares the `['roles', { projectId, roleSlug }]` cache key with the page's typed observers while fetching a different URL.

scott-ray-wilson (dismissed):

frontend/src/pages/project/RoleDetailsBySlugPage/components/EditProjectRoleDialog.tsx
andrewhuhh: not worth solving here imo, this needs to be a structural change from sheet which i've got a ticket for

frontend/src/pages/project/AccessControlPage/components/ProjectRoleListTab/components/ProjectRoleList/ProjectRoleList.tsx
andrewhuhh: i found an old ticket of yours saying we should turn this into dual tables re; github does thi

andrewhuhh: https://linear.app/infisical/issue/DX-41/split-custom-roles-vs-default-roles-into-two-different-tables

frontend/src/pages/project/RoleDetailsBySlugPage/components/EditProjectRoleDialog.tsx
andrewhuhh: wtf

andrewhuhh: fixing

frontend/src/components/v3/platform/AccessRestrictedBanner.tsx
andrewhuhh: o good point wtf

andrewhuhh: oh i remmeber now, it was part of a larger effort to enrich the platform storybook section. not sure why exactly it was added back tho

frontend/src/components/v3/platform/SecretManagerResources/SecretManagerResources.stories.tsx
andrewhuhh: this is a const store of all secrets management microservices. there were a few places recreating this list + icons + colors indiviudally, so i moved these to a store + updated callsites to reference these. will make future feature additions sync much more easily to expected sites

andrewhuhh: tbh dont know if these belong in storybook

scott-ray-wilson (approved):
LGTM!

PR #7694, opened 2026-08-18
feat(pki): windows/linux PKI Syns pre-fetch commands

carlosmonastyrski:
## Context

Adds preflight checks to Linux and Windows certificate syncs. A preflight check is a command configured on the sync that runs on the destination host before any certificate is delivered, so a sync stops early when the host is not ready instead of pushing files somewhere that cannot use them. The same check also runs once a day on its own, surfacing a host that quietly breaks before the next renewal depends on it, with the reason shown on the sync's status. Commands are capped at 10 seconds so a hung check cannot stall a sync, and they support the same variables as post-sync commands.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


docs/documentation/platform/pki/applications/certificate-syncs/linux-server.mdx
```diff
@@ -111,6 +115,75 @@ PKCS#12
...
+</Steps>
+
+### Available Variables
+
+<PreflightCheckVariables />
+
+### Length Limit
+
+A check is limited to 2048 characters, counted after the variables are substituted. For anything longer, put the logic in a script on the server and call that script.
+
```
sheensantoscapadngan: is this a technical limit on Linux?

frontend/src/components/pki-syncs/forms/CreatePkiSyncForm.tsx
```diff
@@ -134,11 +136,14 @@ const getFormTabs = (
...
     baseTabs.push({
-      name: "Post-Sync Command",
-      key: "postSyncCommand",
-      fields: ["syncOptions.postSyncCommand"] as FieldPath<TPkiSyncForm>[]
+      name: "Commands",
+      key: "hostCommands",
+      fields: [
+        "syncOptions.preflightCommand",
+        "syncOptions.postSyncCommand"
+      ] as FieldPath<TPkiSyncForm>[]
```
sheensantoscapadngan: IGNORE PLACEMENT OF COMMENT

should we refer to this as "pre-flight" or just a "health-check" in general? wondering what terminology would be more intuitive for users given that this does also run once a day rather than only during certificate sync ops @carlosmonastyrski @bernie-g 
Image

frontend/src/components/pki-syncs/forms/CreatePkiSyncForm.tsx
```diff
@@ -175,9 +180,11 @@ export const CreatePkiSyncForm = ({
 
   const { syncOption } = usePkiSyncOption(destination);
   const canSetPostSyncCommand = useCanSetPostSyncCommand(applicationId);
+  const canSetPreflightCommand = useCanSetPreflightCommand(applicationId);
   const FORM_TABS = getFormTabs(
     destination,
-    Boolean(syncOption?.canRunPostSyncCommand) && canSetPostSyncCommand
+    (Boolean(syncOption?.canRunPreflightCommand) && canSetPreflightCommand) ||
```
sheensantoscapadngan: IGNORE PLACEMENT

Seems pretty important for users to be able to test pre-flight checks from here, no? Otherwise, how do people know that what they inputted indeed works? 
Image

frontend/src/components/pki-syncs/forms/CreatePkiSyncForm.tsx
```diff
@@ -175,9 +180,11 @@ export const CreatePkiSyncForm = ({
 
   const { syncOption } = usePkiSyncOption(destination);
   const canSetPostSyncCommand = useCanSetPostSyncCommand(applicationId);
+  const canSetPreflightCommand = useCanSetPreflightCommand(applicationId);
```
sheensantoscapadngan: IGNORE PLACEMENT

Is it important for people to know when the last time a check ran?

frontend/src/components/pki-syncs/forms/CreatePkiSyncForm.tsx
```diff
@@ -374,11 +381,17 @@ export const CreatePkiSyncForm = ({
...
             {currentKey === "mappings" && <PkiSyncFieldMappingsFields destination={destination} />}
-            {currentKey === "postSyncCommand" && (
-              <PkiSyncPostSyncCommandFields
-                destination={destination}
-                canEditCommand={canSetPostSyncCommand}
-              />
+            {currentKey === "hostCommands" && (
+              <div className="flex flex-col gap-8">
+                <PkiSyncPreflightCommandFields
+                  destination={destination}
```
sheensantoscapadngan, line 388: IGNORE PLACEMENT

this view is starting to become overwhelming... could we break this up into smaller sections/components? might need Scott's help here
Image

frontend/src/components/pki-syncs/forms/CreatePkiSyncForm.tsx
```diff
@@ -79,12 +81,12 @@ const STEP_META: Record<
...
-    short: "Command after sync",
-    subtitle: "Optionally run a command on the host after certificates are written.",
-    rightLabel: "POST-SYNC COMMAND",
+  hostCommands: {
+    short: "Commands on the host",
+    subtitle: "Optionally check the host before the sync, and act on it afterward.",
+    rightLabel: "COMMANDS",
     rightDescription:
-      "The gateway runs your command on the destination host once the run's files are in place, so a service can reload and pick up the new certificate. It only runs when the sync delivers a file, and a failure marks the sync failed."
+      "The preflight check runs first and stops the sync if the host is not ready, so a bad host never receives a certificate. The post-sync command runs afterward, once the run's files are in place, so a service can reload and pick up the new certificate. It only runs when the sync delivers a file. Either command failing marks the sync failed."
```
sheensantoscapadngan: IGNORE PLACEMENT

Seems like it would be helpful for people to be able to trigger the health checks manually as well, no? 
Image

backend/src/db/migrations/20260811151204_add-pki-sync-health-check.ts
```diff
@@ -0,0 +1,82 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+const HEALTH_CHECK_DISCOVERY_INDEX = "idx_pki_syncs_health_check_discovery";
+const MIGRATION_TIMEOUT = 60 * 60 * 1000; // 60 minutes
+const MIGRATION_LOCK_TIMEOUT = 30 * 1000; // 30 seconds
+
```
bernie-g, line 8: ignore comment placement

Image

This collapsable sidebar looks a bit weird to me - can we vertically center the labels in their sections maybe?

backend/src/db/migrations/20260811151204_add-pki-sync-health-check.ts
```diff
@@ -0,0 +1,82 @@
...
+const HEALTH_CHECK_COLUMNS = ["lastHealthCheckRanAt", "lastHealthCheckStatus", "lastHealthCheckMessage"] as const;
```
bernie-g, line 9: ignore comment placement

Image

It's unclear to me what `Capped at 15 seconds.` is referring to (I haven't read the code yet)

bernie-g: Also I think this entire caption sounds a bit AI. Can we rephrase it to be more clear and direct with plain english

bernie-g: Same with the sidebar comment

backend/src/db/migrations/20260811151204_add-pki-sync-health-check.ts
```diff
@@ -0,0 +1,82 @@
...
+
+type HealthCheckColumn = (typeof HEALTH_CHECK_COLUMNS)[number];
+
+const findExistingColumns = async (knex: Knex): Promise<Set<HealthCheckColumn>> => {
+  const present = new Set<HealthCheckColumn>();
```
bernie-g, line 14: ignore comment placement

Image

Getting this error when trying to edit a health check command after already adding a certificate. Seems like a bug

bernie-g: Also for the creation flow what do you think about moving the commands section to after the certificates section so we mostly don't run into this error anymore?

backend/src/db/migrations/20260811151204_add-pki-sync-health-check.ts
```diff
@@ -0,0 +1,82 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+const HEALTH_CHECK_DISCOVERY_INDEX = "idx_pki_syncs_health_check_discovery";
+const MIGRATION_TIMEOUT = 60 * 60 * 1000; // 60 minutes
+const MIGRATION_LOCK_TIMEOUT = 30 * 1000; // 30 seconds
+
+const HEALTH_CHECK_COLUMNS = ["lastHealthCheckRanAt", "lastHealthCheckStatus", "lastHealthCheckMessage"] as const;
+
```
bernie-g, line 10: Image
Can we add an alert for failed health checks? Possibly failed post-sync commands too? Otherwise I'm not sure how anyone will find out about the daily health check failing

backend/src/services/pki-sync/pki-sync-health-check-queue.ts
```diff
@@ -0,0 +1,456 @@
...
+      });
+      if (cleared) {
+        logger.info(
+          `cron[${CronJobName.PkiSyncHealthCheck}]: host recovered, cleared the reported failure [syncId=${pkiSync.id}]`
+        );
+      }
+      return;
+    }
+
+    if (result.exitCode === undefined && !isFinalAttempt) {
```
bernie-g: According to Claude, this will make the health check retry if the command times out, which probably isn't a good idea if the command has already run

backend/src/services/pki-sync/pki-sync-dal.ts
```diff
@@ -331,8 +331,88 @@ export const pkiSyncDALFactory = (db: TDbClient) => {
...
+        lastHealthCheckMessage: outcome.message
+      });
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Record health check outcome - PKI Sync" });
+    }
+  };
+
+  const $whereMessageStartsWithAny = (builder: Knex.QueryBuilder, prefixes: string[]) => {
+    prefixes.forEach((prefix) => {
+      void builder.orWhereRaw(`left("lastSyncMessage", ?) = ?`, [prefix.length, prefix]);
```
bernie-g: This seems sus, can we just store the separate status's in their own columns

backend/src/services/pki-sync/pki-sync-dal.ts
```diff
@@ -331,8 +331,88 @@ export const pkiSyncDALFactory = (db: TDbClient) => {
...
+        lastHealthCheckRanAt: outcome.ranAt,
+        lastHealthCheckStatus: outcome.status,
+        lastHealthCheckMessage: outcome.message
+      });
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Record health check outcome - PKI Sync" });
+    }
+  };
+
+  const $whereMessageStartsWithAny = (builder: Knex.QueryBuilder, prefixes: string[]) => {
```
bernie-g: ignore comment placement

**From claude:**
The guard that blocks {{commonName}}/{{certificatePath}} on multi-certificate syncs counts only explicitly-linked certificates from certificate_syncs, but subscriber-backed syncs get their certificates from the subscriber
  instead, so the count is always 0 and the check silently passes.

  Later, when that subscriber has a second certificate issued, the health check throws and every sync aborts permanently, with no change to the sync's own config to explain it.

backend/src/db/migrations/20260811151204_add-pki-sync-health-check.ts
carlosmonastyrski: Added a global notification for any PKI Sync failure, the custom alerts will be added in a separate PR

backend/src/services/pki-sync/pki-sync-health-check-queue.ts
```diff
@@ -0,0 +1,507 @@
...
+
+        const outcome = await $processHealthCheck(job.data.syncId, isFinalAttempt).catch((err: unknown) => {
+          logger.error(
+            { err, syncId: job.data.syncId },
+            `cron[${CronJobName.PkiSyncHealthCheck}]: check could not be run [syncId=${job.data.syncId}] [attempt=${job.attemptsStarted}]`
+          );
+          throw err;
+        });
+
+        if (outcome.skipped && isFinalAttempt) {
```
bernie-g: If the outcome is skipped then isFinalAttempt would be false so we'd never record the health check being skipped, right? I think this condition is wrong

bernie-g (approved):

PR #7695, opened 2026-08-18
feat: split the seat count dal from get plan

akhilmhdh:
## Context

This PR removes the seat count dal query from the get plan. This is to improve the performance of getPlan and also to do this aggregation when it's really needed.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/ee/services/license/license-fns.ts
```diff
@@ -203,14 +201,16 @@ export const setupLicenseRequestWithStore = (
...
+  licenseService: Pick<TLicenseServiceFactory, "getPlan" | "getOrgSeatUsage">,
   orgId: string,
   type?: UserAliasType
 ) => {
   const plan = await licenseService.getPlan(orgId);
   const isEnterpriseBypass = plan?.slug === "enterprise" && !plan?.enforceIdentityLimit;
+  if (isEnterpriseBypass || !plan?.identityLimit) return;
 
-  if (!isEnterpriseBypass && plan?.identityLimit && plan.identitiesUsed >= plan.identityLimit) {
+  const { identitiesUsed } = await licenseService.getOrgSeatUsage(orgId);
```
Thiago-AS: We need to pass TX here. All places that call `throwOnPlanSeatLimitReached` runs inside a transaction otherwise we might run into connection pool deadlock.

backend/src/ee/services/license/license-service.ts
```diff
@@ -246,38 +258,15 @@ export const licenseServiceFactory = ({
...
         data: { currentPlan: v1Plan }
-      } = await licenseServerCloudApi.request.get<{ currentPlan: TFeatureSet }>(
-        `/api/license-server/v1/customers/${org.customerId}/cloud-plan`
-      );
-      currentPlan = v1Plan;
+      } = await licenseServerCloudApi.request.get<{
+        currentPlan: TFeatureSet & { membersUsed?: number; identitiesUsed?: number };
+      }>(`/api/license-server/v1/customers/${org.customerId}/cloud-plan`);
+      const { membersUsed, identitiesUsed, ...planWithoutUsage } = v1Plan;
+      currentPlan = planWithoutUsage;
```
Thiago-AS, line 265: We are removing the block that compared the license server’s stored identitiesUsed against the live count and PATCHed quantity/quantityIdentities when they didn’t match. That ran on every plan cache miss and was basically the self healing loop.

The only PATCH left is in updateSubscriptionOrgMemberCount, and all 12 of its call sites are in org-membership, identity, SSO, or SCIM paths. None of them are triggered by project membership or project deletion.

So moving a user into their first Secret Manager project, or deleting their last one, changes the billable count.

Before, that drift would get fixed the next time the cache expired. Now, it just sticks around indefinitely. Is that intended?

backend/src/ee/services/license/license-service.ts
```diff
@@ -222,8 +220,22 @@ export const licenseServiceFactory = ({
     }
   };
 
+  const getOrgSeatUsage = async (orgId: string, tx?: Knex): Promise<TOrgSeatUsage> => {
+    let scopedOrgId: string | null = null;
+    if (instanceType === InstanceType.Cloud) {
```
Thiago-AS, line 225: Are we changing for self hosted to be capped on instance and not org? Is this intended?

A self-hosted instance with 3 orgs × 5 identities and identityLimit: 10 used to allow each org up to 10; it now caps the whole instance at 10.

backend/src/ee/services/license/license-service.ts
```diff
@@ -471,7 +454,7 @@ export const licenseServiceFactory = ({
     if (refreshCache) {
       await refreshPlan(rootOrgId);
     }
-    const plan = await getPlan(rootOrgId, projectId);
+    const plan = getPlan(rootOrgId, projectId);
```
Thiago-AS: Why no await here?

frontend/src/pages/organization/AccessManagementPage/components/OrgMembersTab/components/OrgMembersSection/OrgMembersSection.tsx
```diff
@@ -80,20 +77,7 @@ export const OrgMembersSection = () => {
...
   const { mutateAsync: updateOrgMembership } = useUpdateOrgMembership();
 
-  const isMoreIdentitiesAllowed = subscription?.identityLimit
-    ? subscription.identitiesUsed < subscription.identityLimit
-    : true;
-
-  const isEnterprise = subscription?.slug === SubscriptionPlanTypes.Enterprise;
-
   const handleAddMemberModal = () => {
-    if (!isMoreIdentitiesAllowed && !isEnterprise) {
```
Thiago-AS, line 90: This at limit user now fills in the whole invite modal and gets a server error instead of the upgrade prompt

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentitySection.tsx
```diff
@@ -196,14 +189,6 @@ const IdentitySectionContent = ({ view = "identities" }: Props) => {
                   <Button
                     variant={isSubOrganization ? "sub-org" : "org"}
                     onClick={() => {
-                      if (!isMoreIdentitiesAllowed && !isEnterprise) {
```
Thiago-AS, line 199: Same here, error instead of upgrade modal.

backend/src/services/identity-v2/identity-service.ts
```diff
@@ -119,11 +119,14 @@ export const identityV2ServiceFactory = ({
     const plan = await licenseService.getPlan(dto.permission.orgId);
     const isEnterpriseBypass = plan?.slug === "enterprise" && !plan?.enforceIdentityLimit;
 
-    if (!isEnterpriseBypass && plan?.identityLimit && plan.identitiesUsed >= plan.identityLimit) {
-      // limit imposed on number of identities allowed / number of identities used exceeds the number of identities allowed
-      throw new BadRequestError({
-        message: "Failed to create identity due to identity limit reached. Upgrade plan to create more identities."
-      });
+    if (!isEnterpriseBypass && plan?.identityLimit) {
+      const { identitiesUsed } = await licenseService.getOrgSeatUsage(dto.permission.orgId);
```
Thiago-AS, line 123: Should we add here? We were using the cache plan. This might cause some perf issues 🤔

Thiago-AS (changes requested):

backend/src/ee/services/license/license-service.ts
akhilmhdh: This is not the reconciliation loop anymore. This is old license server loop, the new loop is a queue trigger that happens on getplan or any change

backend/src/services/identity-v2/identity-service.ts
akhilmhdh: The present system is the one causing performance issue. It calls it for all the getPlan even when you don't need it. Billable unit aggregation is a bit heavy especially the new ones. It should not run for getPlan

frontend/src/pages/organization/AccessManagementPage/components/OrgIdentityTab/components/IdentitySection/IdentitySection.tsx
akhilmhdh: We will be improving this experience later. There is another initiative for that.

frontend/src/pages/organization/AccessManagementPage/components/OrgMembersTab/components/OrgMembersSection/OrgMembersSection.tsx
akhilmhdh: We will be improving this experience later. There is another initiative for that.

backend/src/ee/services/license/license-service.ts
akhilmhdh: Yes! This is the expected behavior.

backend/src/ee/services/license/license-service.ts
akhilmhdh: Doesn't matter much. It's returned and awaited at the router level. Both are same I assume

Thiago-AS: Yeah, assuming everything that reads this is async we guchi, just thinking that it can silently break in case someone calls it sync.

akhilmhdh: We. do have lint check for that. But updating this.

Thiago-AS (approved):

PR #7703, opened 2026-08-18
feat(gateway-pools): load balancing

bernie-g:
## Context

Gateway pools picked a gateway at random, so one could end up carrying far more work than the others while its peers sat idle. Pools now send each new job to whichever gateway is least busy, and retry on another gateway when one can't be reached.

Needs the matching CLI change so gateways report how busy they are: Infisical/cli#370

## Steps to verify the change

1. Put two or more gateways in a pool and point an app connection or Kubernetes auth config at it.
2. Drive traffic through and confirm it spreads instead of landing on one gateway.
3. Kill a gateway mid-traffic: requests keep succeeding and the logs show a retry on another member.
4. Load one gateway up (e.g. a few PAM CLI sessions) and confirm new work goes to the quieter one.
5. Use a wrong password on the target and confirm it fails straight away, without being retried elsewhere.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/e2e-test/mocks/keystore.ts
+++ b/backend/e2e-test/mocks/keystore.ts
@@ -79,6 +79,9 @@ export const mockKeyStore = (): TKeyStoreFactory => {
     },
     claimLeastLoaded: async (keys, baseOccupancies) => {
       if (keys.length === 0) return 0;
+      if (keys.length !== baseOccupancies.length) {
+        throw new Error("claimLeastLoaded: baseOccupancies must have one entry per key");
+      }
       let bestIdx = 0;
       let bestTotal: number | null = null;
       keys.forEach((key, i) => {
--- a/backend/e2e-test/mocks/keystore.ts
+++ b/backend/e2e-test/mocks/keystore.ts
@@ -138,14 +138,6 @@ export const mockKeyStore = (): TKeyStoreFactory => {
     hashGet: async (key, field) => {
       return hashStore[key]?.[field] ?? null;
     },
-    hashGetAll: async (key) => {
-      return { ...(hashStore[key] ?? {}) };
-    },
-    hashDelete: async (key, field) => {
-      if (!hashStore[key]?.[field]) return 0;
-      delete hashStore[key][field];
-      return 1;
-    },
     pgIncrementBy: async () => {
       return 1;
     },
--- a/backend/e2e-test/mocks/keystore.ts
+++ b/backend/e2e-test/mocks/keystore.ts
@@ -19,7 +19,7 @@ export const mockKeyStore = (): TKeyStoreFactory => {
     },
     setExpiry: async () => 0,
     ttl: async () => -1,
-    setItemWithExpiry: async (key, value) => {
+    setItemWithExpiry: async (key, _expiryInSeconds, value) => {
       store[key] = value;
       return "OK";
     },
@@ -77,8 +77,15 @@ export const mockKeyStore = (): TKeyStoreFactory => {
     incrementByAndRefreshExpiryIfUnderLimit: async () => {
       return 1;
     },
-    decrementByOrDelete: async () => {
-      return 0;
+    decrementByOrDelete: async (key) => {
+      const current = typeof store[key] === "string" ? parseInt(store[key] as string, 10) : 0;
+      const next = current - 1;
+      if (next <= 0) {
+        delete store[key];
+        return 0;
+      }
+      store[key] = String(next);
+      return next;
     },
     incrementByWithExpiry: async (key, value) => {
       const current = typeof store[key] === "string" ? parseInt(store[key] as string, 10) : 0;
@@ -111,8 +118,22 @@ export const mockKeyStore = (): TKeyStoreFactory => {
     hashGet: async (key, field) => {
       return hashStore[key]?.[field] ?? null;
     },
+    hashGetAll: async (key) => {
+      return { ...(hashStore[key] ?? {}) };
+    },
+    hashDelete: async (key, field) => {
+      if (!hashStore[key]?.[field]) return 0;
+      delete hashStore[key][field];
+      return 1;
+    },
     pgIncrementBy: async () => {
       return 1;
+    },
+    getItemsPrimary: async (keys) => {
+      return keys.map((key) => {
+        const value = store[key];
+        return typeof value === "string" ? value : null;
+      });
     },
     getItems: async (keys) => {
       const values = keys.map((key) => {
--- a/backend/e2e-test/mocks/keystore.ts
+++ b/backend/e2e-test/mocks/keystore.ts
@@ -77,6 +77,23 @@ export const mockKeyStore = (): TKeyStoreFactory => {
     incrementByAndRefreshExpiryIfUnderLimit: async () => {
       return 1;
     },
+    claimLeastLoaded: async (keys, baseOccupancies) => {
+      if (keys.length === 0) return 0;
+      let bestIdx = 0;
+      let bestTotal: number | null = null;
+      keys.forEach((key, i) => {
+        const reserved = typeof store[key] === "string" ? parseInt(store[key] as string, 10) : 0;
+        const total = baseOccupancies[i] + reserved;
+        if (bestTotal === null || total < bestTotal) {
+          bestTotal = total;
+          bestIdx = i + 1;
+        }
+      });
+      const chosen = keys[bestIdx - 1];
+      const current = typeof store[chosen] === "string" ? parseInt(store[chosen] as string, 10) : 0;
+      store[chosen] = String(current + 1);
+      return bestIdx;
+    },
     decrementByOrDelete: async (key) => {
       const current = typeof store[key] === "string" ? parseInt(store[key] as string, 10) : 0;
       const next = current - 1;
--- a/backend/src/ee/routes/v2/gateway-router.ts
+++ b/backend/src/ee/routes/v2/gateway-router.ts
@@ -3,7 +3,7 @@ import z from "zod";
 import { GatewaysV2Schema } from "@app/db/schemas";
 import { GATEWAYS } from "@app/lib/api-docs";
 import { zodBuffer } from "@app/lib/zod";
-import { gatewayLoadReportLimit, readLimit, writeLimit } from "@app/server/config/rateLimiter";
+import { gatewayMetricsReportLimit, readLimit, writeLimit } from "@app/server/config/rateLimiter";
 import { slugSchema } from "@app/server/lib/schemas";
 import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
 import { AuthMode } from "@app/services/auth/auth-type";
@@ -68,25 +68,25 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
   // gateway cannot look alive just by reporting load.
   server.route({
     method: "POST",
-    url: "/load",
+    url: "/metrics",
     config: {
-      rateLimit: gatewayLoadReportLimit
+      rateLimit: gatewayMetricsReportLimit
     },
     schema: {
-      operationId: "gatewayLoadReport",
+      operationId: "gatewayMetricsReport",
       body: z.object({
-        activeChannels: z.number().int().min(0).max(1_000_000).describe(GATEWAYS.LOAD_REPORT.activeChannels)
+        activeChannels: z.number().int().min(0).max(1_000_000).describe(GATEWAYS.METRICS_REPORT.activeChannels)
       }),
       response: {
         200: z.object({
-          gatewayId: z.string().uuid().describe(GATEWAYS.LOAD_REPORT.gatewayId),
-          activeChannels: z.number().int().describe(GATEWAYS.LOAD_REPORT.activeChannels)
+          gatewayId: z.string().uuid().describe(GATEWAYS.METRICS_REPORT.gatewayId),
+          activeChannels: z.number().int().describe(GATEWAYS.METRICS_REPORT.activeChannels)
         })
       }
     },
     onRequest: verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN, AuthMode.GATEWAY_ACCESS_TOKEN]),
     handler: async (req) => {
-      return server.services.gatewayV2.reportLoad({
+      return server.services.gatewayV2.reportMetrics({
         orgPermission: req.permission,
         activeChannels: req.body.activeChannels
       });
--- a/backend/src/ee/routes/v2/gateway-router.ts
+++ b/backend/src/ee/routes/v2/gateway-router.ts
@@ -1,8 +1,9 @@
 import z from "zod";
 
 import { GatewaysV2Schema } from "@app/db/schemas";
+import { GATEWAYS } from "@app/lib/api-docs";
 import { zodBuffer } from "@app/lib/zod";
-import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
+import { gatewayLoadReportLimit, readLimit, writeLimit } from "@app/server/config/rateLimiter";
 import { slugSchema } from "@app/server/lib/schemas";
 import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
 import { AuthMode } from "@app/services/auth/auth-type";
@@ -61,34 +62,34 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
     }
   });
 
-  // Deliberately separate from /heartbeat: that one probes back through the relay and writes to the
-  // database, which is far too costly to run at the cadence pool selection needs. This only touches
-  // the load tracker's Redis keys.
+  // Separate from /heartbeat because of what its handler costs, not how it is called: a heartbeat
+  // makes the platform dial back through the relay to probe liveness and then writes to Postgres,
+  // which is far too expensive at the cadence pool selection needs. This one only writes a Redis key,
+  // and deliberately does not touch heartbeat, so a gateway cannot look alive by reporting load.
   server.route({
     method: "POST",
     url: "/load",
     config: {
-      rateLimit: writeLimit
+      rateLimit: gatewayLoadReportLimit
     },
     schema: {
       operationId: "gatewayLoadReport",
       body: z.object({
-        activeChannels: z.number().int().min(0).max(1_000_000)
+        activeChannels: z.number().int().min(0).max(1_000_000).describe(GATEWAYS.LOAD_REPORT.activeChannels)
       }),
       response: {
         200: z.object({
-          message: z.string()
+          gatewayId: z.string().uuid().describe(GATEWAYS.LOAD_REPORT.gatewayId),
+          activeChannels: z.number().int().describe(GATEWAYS.LOAD_REPORT.activeChannels)
         })
       }
     },
     onRequest: verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN, AuthMode.GATEWAY_ACCESS_TOKEN]),
     handler: async (req) => {
-      await server.services.gatewayV2.reportLoad({
+      return server.services.gatewayV2.reportLoad({
         orgPermission: req.permission,
         activeChannels: req.body.activeChannels
       });
-
-      return { message: "Successfully reported gateway load" };
     }
   });
 
--- a/backend/src/ee/routes/v2/gateway-router.ts
+++ b/backend/src/ee/routes/v2/gateway-router.ts
@@ -62,10 +62,10 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
     }
   });
 
-  // Separate from /heartbeat because of what its handler costs, not how it is called: a heartbeat
-  // makes the platform dial back through the relay to probe liveness and then writes to Postgres,
-  // which is far too expensive at the cadence pool selection needs. This one only writes a Redis key,
-  // and deliberately does not touch heartbeat, so a gateway cannot look alive by reporting load.
+  // Heartbeat is expensive: the platform dials back through the relay to check the gateway is
+  // really there, then writes to Postgres. Every gateway reports load every 10 seconds, which is
+  // far too often for that. This one just writes a Redis key. It does not touch heartbeat, so a
+  // gateway cannot look alive just by reporting load.
   server.route({
     method: "POST",
     url: "/load",
--- a/backend/src/ee/routes/v2/gateway-router.ts
+++ b/backend/src/ee/routes/v2/gateway-router.ts
@@ -61,6 +61,37 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
     }
   });
 
+  // Deliberately separate from /heartbeat: that one probes back through the relay and writes to the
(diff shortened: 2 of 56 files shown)


backend/src/ee/services/gateway-pool/gateway-pool-service.ts
```diff
@@ -257,42 +258,130 @@ export const gatewayPoolServiceFactory = ({
...
+      }
+    }
+
+    if (!selected) selected = pickRandomGateway(eligible);
+    if (!selected)
+      throw new BadRequestError({ message: unavailableMessage ?? "Gateway pool has no healthy gateways." });
+
+    // Bookkeeping must never be able to fail a selection, whatever the tracker does internally.
+    try {
+      await loadTracker?.reserve(selected.id);
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -61,6 +61,37 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
...
 
+  // Deliberately separate from /heartbeat: that one probes back through the relay and writes to the
+  // database, which is far too costly to run at the cadence pool selection needs. This only touches
+  // the load tracker's Redis keys.
+  server.route({
+    method: "POST",
+    url: "/load",
+    config: {
+      rateLimit: writeLimit
+    },
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -980,6 +981,27 @@ export const gatewayV2ServiceFactory = ({
...
+    orgPermission,
+    activeChannels
+  }: {
+    orgPermission: OrgServiceActor;
+    activeChannels: number;
+  }) => {
+    const gateway =
+      orgPermission.type === ActorType.GATEWAY
+        ? await gatewayV2DAL.findById(orgPermission.id)
+        : await gatewayV2DAL.findOne({ orgId: orgPermission.orgId, identityId: orgPermission.id });
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -980,6 +981,27 @@ export const gatewayV2ServiceFactory = ({
...
 
+  const reportLoad = async ({
+    orgPermission,
+    activeChannels
+  }: {
+    orgPermission: OrgServiceActor;
+    activeChannels: number;
+  }) => {
+    const gateway =
+      orgPermission.type === ActorType.GATEWAY
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -61,6 +61,37 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
...
+    schema: {
+      operationId: "gatewayLoadReport",
+      body: z.object({
+        activeChannels: z.number().int().min(0).max(1_000_000)
+      }),
+      response: {
+        200: z.object({
+          message: z.string()
```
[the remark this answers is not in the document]
bernie-g: Fixed

usefactorapp:
FINDING 1

Redis-backed load tracking governs selection for 500+ connection fan-out paths

The repository migrates diverse connection providers to a centralized `GatewayPoolService` that uses a Redis-backed `GatewayLoadTracker` for load-aware selection. Logic like `claimLeastLoaded` and `trackReservationRelease` now manages concurrency across all connection types (SSH, LDAP, SQL), introducing a repository-wide dependency on the performance of load-tracking hash keys. Factor found no validation evidence for this critical selection logic within the changed paths.

Evidence:
- Symbol: `backend.src.ee.services.gateway-pool.gateway-pool-service.selectGatewayFromPool`
- Symbol: `backend.src.lib.gateway-v2.gateway-load-tracker.claimLeastLoaded`
- Discovery IR reports 2247 boundary crossings with no validation coverage for the centralized selection path

Agent Prompt:
Trace the lifecycle of a connection request starting from `AppConnectionService.createAppConnection` in `backend/src/services/app-connection/app-connection-service.ts` down through `GatewayPoolService.selectGatewayFromPool` and into `GatewayLoadTracker.claimLeastLoaded`. Investigate how Redis keys like `reservationKey` and `reportedKey` are used in `backend/src/lib/gateway-v2/gateway-load-tracker.ts` to maintain state. Explain how `runWithPoolFailover` handles transport failures vs. application-level errors and cite relevant file evidence. Do not modify code.


FINDING 2

AsyncLocalStorage attempt context decoupled from 2200+ deep cross-service execution chains

The new retry logic in `gateway-retry.ts` depends on a `GatewayAttemptContext` managed via `AsyncLocalStorage`. This context allows services deep in the call stack to call `markAttemptTransportFailure` without requiring signature changes across thousands of symbols. This decoupling is essential for the `isGatewayTransportFailure` check, which distinguishes between connectivity issues and application errors in the failover mechanism.

Evidence:
- Symbol: `backend.src.lib.gateway-v2.gateway-attempt-context.runGatewayAttempt`
- Symbol: `backend.src.lib.gateway-v2.gateway-retry.isAttemptRetryable`
- The state is propagated implicitly using `AsyncLocalStorage` storage in `backend/src/lib/gateway-v2/gateway-attempt-context.ts`

Agent Prompt:
Trace the implementation and usage of `runGatewayAttempt` in `backend/src/lib/gateway-v2/gateway-attempt-context.ts`. Identify where `markAttemptTransportFailure` is called by connection providers and how `gateway-retry.ts` consumes this state via `isGatewayTransportFailure` to decide on retries. Explain the mechanism by which the `AsyncLocalStorage` state is persisted across the service boundaries indicated by the discovery metrics. Do not modify code.


FINDING 3

Unified ALPN-based protocol multiplexing standardizes disparate tunneling execution paths

Disparate connection providers for SSH, SQL, and HTTP are being refactored to use a unified `withGatewayV2Proxy` execution wrapper. This wrapper implements protocol multiplexing through a `protocolToAlpn` mapping, effectively standardizing how tunnels are established across the repository. This change moves protocol-specific connectivity logic into a shared infrastructure layer, centralizing blast radius for tunneling failures.

Evidence:
- Symbol: `backend.src.lib.gateway-v2.gateway-v2.protocolToAlpn`
- Symbol: `backend.src.lib.gateway-v2.gateway-v2.withGatewayV2Proxy`
- Refactoring of `executeSshCommandViaGateway`, `requestWithHCVaultGateway`, and `executeWinRMGatewayOperation` to this shared wrapper

Agent Prompt:
Inspect `backend/src/lib/gateway-v2/gateway-v2.ts` and focus on the `withGatewayV2Proxy` function. Determine how it maps internal protocol requests to ALPN identifiers via `protocolToAlpn` and how the `relayServer` handles lifecycle cleanup. Trace a specific consumer like `executeSshCommandViaGateway` in `ssh-connection-fns.ts` and explain how it passes credentials and configuration through this standardized tunnel. Do not modify code.

backend/e2e-test/mocks/keystore.ts
```diff
@@ -19,7 +19,7 @@ export const mockKeyStore = (): TKeyStoreFactory => {
     },
```
x032205: ignore placement

I ran a test with 1 old gateway (pre-PR) and one new gateway (post-PR). The old gateway had 10 sessions running, with the new one having 0. The results were that the old gateway would still get connections randomly, even when the new gateway has significantly less load.

I understand that with the current architecture this problem is hard to avoid, and I'm not sure if we can even come up with a solution, but adding the comment anyways.

I saw a comment from Sheen on the other PR that proposes keeping track of load on the backend rather than the gateway which could potentially fix this issue.

backend/src/keystore/keystore.ts
```diff
@@ -447,6 +458,45 @@ export const keyStoreFactory = (
     return Number(result);
   };
 
+  // Choosing and claiming has to be one round trip. Done as separate read and write calls, every
+  // concurrent selection reads the same minimum before any of them claims it and they all pile onto
+  // the same gateway, which is the stampede the reservation exists to prevent.
+  const CLAIM_LEAST_LOADED_SCRIPT = `
```
lb-vn, line 480: Claude:

Different gateway ids hash to different slots, so this returns CROSSSLOT on redis cluster and we silently fall back to random. Should we hash tag the reservation key?

> Redis Cluster splits the keyspace into 16384 slots and each node owns a subset. A Lua script runs on one node, so Redis requires every key it touches to live in the same slot; if they don't it refuses with a CROSSSLOT error rather than running.

backend/src/keystore/keystore.ts
```diff
@@ -447,6 +458,45 @@ export const keyStoreFactory = (
...
+      local total = tonumber(ARGV[i]) + reserved
+      -- strict less-than, so ties fall to the caller's order (pre-shuffled for a random tie-break)
+      if bestTotal == nil or total < bestTotal then
+        bestTotal = total
+        bestIdx = i
+      end
+    end
+    if bestIdx == 0 then return 0 end
+    redis.call("INCR", KEYS[bestIdx])
+    redis.call("EXPIRE", KEYS[bestIdx], ttl)
```
lb-vn: Claude:

EXPIRE on every claim keeps pushing the ttl out, so a dead pod's reservations never expire. Maybe we only set it when INCR returns 1?

backend/src/keystore/keystore.ts
```diff
@@ -447,6 +458,45 @@ export const keyStoreFactory = (
...
+    return bestIdx
+  `;
+
+  /** Returns the 1-based index of the claimed key, or 0 when no keys were given. */
+  const claimLeastLoaded = async (
```
lb-vn, line 503: Claude:

If baseOccupancies is shorter than keys the script reads the ttl as an occupancy. Worth a length guard.

backend/src/lib/gateway-v2/gateway-v2.ts
```diff
@@ -183,6 +191,18 @@ export const setupRelayServer = async ({
           // Stage 2: Establish mTLS connection to gateway through the relay
           const gatewayConn = await createGatewayConnection(relayConn, gateway, protocol);
 
+          // The caller may have given up while the two handshakes were in flight. Its "close" event
+          // has already fired by now, so attaching the teardown listeners below would never run and
+          // the channel would stay counted for the life of the pod.
+          if (clientConn.destroyed) {
```
lb-vn, line 286: Claude:

The channel is already up here, so skipping markAttemptTunnelEstablished leaves the attempt retryable after the gateway may have reached the target. Should we mark it above this check?

backend/src/server/routes/index.ts
```diff
@@ -573,6 +574,10 @@ export const registerRoutes = async (
   const cronJob = cronJobFactory({ redis, redlock });
   cronJob.start();
 
+  // Reached from the gateway proxy layer in src/lib, which has no DI, so it is installed as a module
+  // singleton rather than threaded through every call site.
+  initGatewayLoadTracker(keyStore);
```
lb-vn: Claude:

We never call shutdown() on this, so a restarting pod leaves its published load and reservations behind. Wire it into onClose?

backend/src/services/hsm-connector/hsm-connector-routing.ts
```diff
@@ -70,7 +73,15 @@ export const hsmConnectorRoutingFactory = ({ gatewayV2Service, gatewayPoolServic
...
     /* eslint-disable no-await-in-loop */
     for (let attempt = 0; attempt <= PKCS11_POOL_RETRY_LIMIT; attempt += 1) {
-      const gatewayId = await pickGateway(args.connector, triedGatewayIds);
+      let gatewayId;
+      try {
+        gatewayId = await pickGateway(args.connector, triedGatewayIds);
+      } catch (err) {
+        // Once every capable member has been tried, the accumulated PKCS#11 failure is the useful
+        // error. Reporting "no HSM-capable gateway" instead points operators at the wrong problem.
+        if (!lastError) throw err;
```
lb-vn, line 82: Claude:

This loop retries on any error, unlike runWithPoolFailover which only retries transport failures. Is pkcs11 safe to replay?

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
```diff
@@ -535,9 +535,7 @@ export const pamWebAccessServiceFactory = ({
 
       relayServer = await setupRelayServer({
         protocol: isRdp ? GatewayProxyProtocol.PamRdpBrowser : GatewayProxyProtocol.Pam,
-        relayHost: certs.relayHost,
-        relay: certs.relay,
-        gateway: certs.gateway,
+        ...certs,
```
lb-vn: ignore placement, this belongs on line 463 within the same file

Web access never goes through runWithPoolFailover. Is that deliberate?

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -61,6 +62,37 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
     }
   });
 
+  // Separate from /heartbeat because of what its handler costs, not how it is called: a heartbeat
+  // makes the platform dial back through the relay to probe liveness and then writes to Postgres,
+  // which is far too expensive at the cadence pool selection needs. This one only writes a Redis key,
+  // and deliberately does not touch heartbeat, so a gateway cannot look alive by reporting load.
```
lb-vn: i think we can simplify this to just say something like "heartbeat does a full e2e liveness test through the relay which is too expensive for a high-frequency reporting function". When I read this comment yesterday it completely went over my head because of its weird opening and jargon, but when I was told by Bernie the issue with using `/heartbeat` in a Slack thread I understood immediately. Clearly theres some issue with wording here imo

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -61,6 +62,37 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
...
+        activeChannels: z.number().int().min(0).max(1_000_000).describe(GATEWAYS.LOAD_REPORT.activeChannels)
+      }),
+      response: {
+        200: z.object({
+          gatewayId: z.string().uuid().describe(GATEWAYS.LOAD_REPORT.gatewayId),
+          activeChannels: z.number().int().describe(GATEWAYS.LOAD_REPORT.activeChannels)
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.IDENTITY_ACCESS_TOKEN, AuthMode.GATEWAY_ACCESS_TOKEN]),
```
lb-vn, line 87: Is IDENTITY_ACCESS_TOKEN needed here?

backend/src/ee/services/gateway-pool/gateway-pool-selection-fns.ts
```diff
@@ -0,0 +1,35 @@
...
+ * are on different scales: a non-reporting member is counted only on channels this platform opened,
+ * which is strictly lower than a reporting peer's true occupancy. Comparing them would pull traffic
+ * onto un-upgraded gateways for a whole rollout, so such a pool gets no load awareness at all.
+ */
+export const isPoolComparable = <T extends TSelectableGateway>(
+  candidates: T[],
+  scores: Map<string, TGatewayScore>
+): boolean => {
+  const reportedCount = candidates.filter((c) => scores.get(c.id)?.reported).length;
+  return reportedCount === 0 || reportedCount === candidates.length;
```
lb-vn: i may be mistaken but isn't having a `reportedCount === 0` check wrong here? I think this would count a pool full of old gateways as comparable, no?

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -980,6 +981,34 @@ export const gatewayV2ServiceFactory = ({
...
+  }) => {
+    let gateway;
+    if (orgPermission.type === ActorType.GATEWAY) {
+      gateway = await gatewayV2DAL.findById(orgPermission.id);
+    } else {
+      // Same check heartbeat makes: an identity whose gateway permission was revoked must not keep
+      // submitting occupancy values and steering pool routing.
+      await $validateIdentityAccessToGateway(orgPermission.orgId, orgPermission.id, orgPermission.authMethod);
+      gateway = await gatewayV2DAL.findOne({ orgId: orgPermission.orgId, identityId: orgPermission.id });
+    }
```
lb-vn, line 1005: is there cases where identities report load on behalf of gateways? Seems weird to me

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -980,6 +981,34 @@ export const gatewayV2ServiceFactory = ({
...
+
+    if (!gateway || gateway.orgId !== orgPermission.orgId) {
+      throw new NotFoundError({ message: `Gateway with ID '${orgPermission.id}' not found` });
+    }
+
+    // Deliberately no heartbeat side effect: liveness stays with /heartbeat so a gateway cannot look
+    // healthy purely by reporting load, and so this stays a single Redis write.
```
lb-vn: I don't understand this comment. Is it saying we're purposefully not updating liveness because it belongs with heartbeat? What's the point of the comment? I think it should be clarified or removed

backend/src/server/config/rateLimiter.ts
```diff
@@ -38,6 +38,19 @@ export const writeLimit: RateLimitOptions = {
   keyGenerator: (req) => req.realIp
 };
 
+// Gateways report load every 10s (6/min each). Keyed by the reporting gateway rather than by IP,
+// because the write limiter's IP key is shared: ~100 gateways behind one NAT address would exhaust
+// a 200/min quota on load reports alone, start getting 429s, and their entries would go stale.
+export const gatewayLoadReportLimit: RateLimitOptions = {
+  timeWindow: 60 * 1000,
+  hook: "preValidation",
+  max: 30,
```
lb-vn: Why do we allow 30 per minute when there's only 6 calls per minute? Seems very generous. Maybe like 10 at most?

lb-vn (changes requested):

backend/src/keystore/keystore.ts
bernie-g: Fixed

backend/src/keystore/keystore.ts
bernie-g: Fixed

backend/src/keystore/keystore.ts
bernie-g: Fixed

backend/src/lib/gateway-v2/gateway-v2.ts
bernie-g: Fixed

backend/src/server/routes/index.ts
bernie-g: Done

backend/src/services/hsm-connector/hsm-connector-routing.ts
bernie-g: This is pre-existing, this PR doesn't add this or change this

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
bernie-g: Fixed

backend/src/ee/routes/v2/gateway-router.ts
bernie-g: Done

backend/src/ee/routes/v2/gateway-router.ts
bernie-g: Yup, this is for legacy machine identity gateways

backend/src/ee/services/gateway-pool/gateway-pool-selection-fns.ts
bernie-g: Correct yea. This function is just checking if the gateways are the same type. Since a pool of all old gateways is the same type, this function returns true

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
bernie-g: It's just for gateways that use the legacy machine identity auth. We could potentially stop supporting it but we haven't yet for other features I don't think

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
bernie-g: Reworded it

backend/src/server/config/rateLimiter.ts
bernie-g: Set it to 10

backend/src/ee/services/gateway-pool/gateway-pool-selection-fns.ts
lb-vn: But technically old gateways aren't comparable right? Only new ones are. Old ones should always fall back to a random selection rather than a comparison

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
lb-vn: @bernie-g I think we can drop this support entirely, since technically only new gateways can report load anyways right? This means we can remove the identity token from the auth array in the endpoint too I think

backend/src/ee/services/gateway-pool/gateway-pool-selection-fns.ts
bernie-g: Renamed

backend/src/server/routes/index.ts
```diff
@@ -4259,6 +4264,9 @@ export const registerRoutes = async (
   }
 
   server.addHook("onClose", async () => {
+    // Drops this pod's published counts, so a rolling deploy does not leave its already-dead
+    // channels counting against those gateways for the freshness window.
```
lb-vn: I think this comment is outdated after recent changes. Can likely be removed

backend/src/server/routes/index.ts
```diff
@@ -577,6 +578,10 @@ export const registerRoutes = async (
   const cronJob = cronJobFactory({ redis, redlock });
   cronJob.start();
```
lb-vn, line 579: IGNORE PLACEMENT

Claude is still pointing out that in the CLI repo we wait 10 seconds until the first metrics report, which causes the selection to be random until that first 10 seconds is up. Whats the verdict on just sending a metrics report right away?

lb-vn (changes requested):

backend/src/server/routes/index.ts
bernie-g: Removed

backend/src/server/routes/index.ts
bernie-g: Changed it to send the metrics immediately

lb-vn (approved):

PR #7707, opened 2026-08-18
fix: use Redis stream for telemetry batching events

Thiago-AS:
## Context

The PostHog aggregated-event buffer previously wrote one Redis key per occurrence (`telemetry-event-*`). `TelemetryAggregatedEvents` discovered these with `getKeysByPattern` for each `(event type, bucket)`: `8 × 30 = 240` full keyspace scans every 10 minutes. On a ~440k-key keyspace, this used ~73s of Redis CPU per run and scaled linearly with the number of aggregated events.

Buffered events now use one Redis stream per `(event type, bucket)`:

`telemetry-agg-stream:<event>:<bucket>`

The cron reads shards by key, so runtime scales with pending events instead of keyspace size. No more `SCAN`.

Aggregation behavior is unchanged: same bucketing, grouping, breakdowns, and `<event> aggregated` payload.

### Other changes

- **Bounded storage:** `MAXLEN ~ 100k` per shard on writes and a 30-minute `XTRIM MINID` at the start of every run.
- **At-least-once delivery:** shards are trimmed only after `postHog.flush()` succeeds.
- **Orphan cleanup:** every visited shard gets a 1-hour `EXPIRE` refresh. Active shards stay alive; abandoned shards expire automatically.
- Drain limit of **50k entries** per shard per run.
- `BUCKET_CONCURRENCY` reduced from **5 → 2**.
- `groupIdentify` deduplication is now shared across the entire run.
- Added unit tests in `telemetry-service.test.ts`.

### End-to-end

1. Start Redis and the backend:

```bash
docker compose -f docker-compose.dev.yml up -d
```

2. Enable telemetry and configure a PostHog capture endpoint. Restart the backend.

3. Generate aggregated events:

```bash
for i in $(seq 1 20); do
  curl -s \
    -H "Authorization: Bearer $TOKEN" \
    "http://localhost:8080/api/v3/secrets/raw?workspaceSlug=<slug>&environment=<env>" \
    >/dev/null
done
```

4. Verify streams exist and legacy keys do not:

```bash
docker compose -f docker-compose.dev.yml exec redis \
  redis-cli --scan --pattern 'telemetry-agg-stream:*'

docker compose -f docker-compose.dev.yml exec redis \
  redis-cli --scan --pattern 'telemetry-event-*'
# expect no results
```

5. Wait for the `*/10` cron (or trigger `processAggregatedEvents`) and verify:

- Backend logs show `Processed N events from bucket <id> for <event>`.
- PostHog receives one `<event> aggregated` event per `(distinctId, org, breakdown)`.
- The drained shard has `XLEN = 0`.
- The shard TTL is ~3600 seconds.

6. Confirm Redis no longer performs scans:

```bash
redis-cli INFO commandstats | grep -E 'cmdstat_(scan|xrange|xtrim)'
```

`cmdstat_scan` should stop increasing during cron runs.

### Failure path

Stop the PostHog endpoint and let the cron run:

- The shard is **not** trimmed.
- The error is logged.
- The next successful run republishes the batch.
- Entries older than 30 minutes are still removed by the age trim.


backend/src/services/telemetry/telemetry-service.ts
```diff
@@ -50,6 +50,12 @@ const TELEMETRY_BUCKET_NAMES = Array.from(
   (_, i) => `bucket-${i.toString().padStart(2, "0")}`
 );
 
+const TELEMETRY_EVENT_STREAM_BATCH_SIZE = 10_000;
+const TELEMETRY_EVENT_STREAM_COLLECT_CEILING = 50_000;
+const TELEMETRY_EVENT_STREAM_MAX_ENTRIES = 100_000;
+const TELEMETRY_EVENT_STREAM_RETENTION_MS = 30 * 60 * 1000;
+const TELEMETRY_EVENT_STREAM_KEY_TTL_SECONDS = 60 * 60;
```
victorvhs017, line 64: We need to add a way to observe/alarm the health of this process to validate these limits. It's ok to lose some events sometimes, but if we are constantly dropping, we would need to tweak these values.

backend/src/services/telemetry/telemetry-service.ts
```diff
@@ -319,17 +323,19 @@ To opt into telemetry, you can set "TELEMETRY_ENABLED=true" within the environme
...
-        eventKey,
-        KeyStoreTtls.TelemetryAggregatedEventInSeconds,
-        JSON.stringify({
-          distinctId: event.distinctId,
-          event: event.event,
-          properties: event.properties,
-          organizationId: event.organizationId,
-          ...(resolvedOrgName ? { organizationName: resolvedOrgName } : {})
-        })
+      await keyStore.streamAdd(
```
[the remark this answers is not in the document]
victorvhs017: I think we should pipeline an expire in the xadd here, not only because we guarantee that we won't have any lingering stream, it doesn't matter the edge case, but also if we are running the app in development and the telemetry is disabled, we never add the expire

Image

backend/src/services/telemetry/telemetry-service.ts
```diff
@@ -461,137 +467,175 @@ To opt into telemetry, you can set "TELEMETRY_ENABLED=true" within the environme
...
+
+      if (collected === 0 || !lastId) return 0;
 
-      // Clean up processed data for this bucket
-      await keyStore.deleteItemsByKeyIn(bucketKeys);
+      await publishAggregatedEvents(eventType, events, orgsIdentifyAttempted);
 
-      logger.info(`Processed ${bucketEventsParsed.length} events from bucket ${bucketId} for ${eventType}`);
-      return bucketEventsParsed.length;
+      await postHog.flush();
```
victorvhs017: I iterated over this report some times and tested it to validate. It seems to be a real issue:

---

`flush()` doesn't deliver here, so the `streamTrim` two lines down can delete events that were never sent.

`capture()` enqueues asynchronously (`addPendingPromise(prepareEventMessage(...).then(...))`), so the queue is still empty when `_flush()` runs and it returns on `if (!queue.length) return`. Delivery actually happens later, via the SDK's own 10s `flushInterval`. Normal traffic still arrives, which is why this is invisible, but the ordering guarantee this line is here to provide doesn't exist. If the process restarts inside that window, or PostHog is failing, the shard is already trimmed and the batch is gone rather than retried next tick.

Separately, the capture loop above never awaits, so nothing drains between enqueues and the queue's 1000-entry cap evicts the oldest. 1200 aggregated groups in, 1000 out, shard trimmed regardless.

**To reproduce**, save as `backend/repro.js` and run `cd backend && node repro.js`:

<details>
<summary>repro.js</summary>

```js
const http = require("http");
const zlib = require("zlib");
const { PostHog } = require("posthog-node");

let requests = 0, delivered = 0;
const server = http.createServer((req, res) => {
  const chunks = [];
  req.on("data", (c) => chunks.push(c));
  req.on("end", () => {
    requests += 1;
    const buf = Buffer.concat(chunks);
    const raw = (req.headers["content-encoding"] || "").includes("gzip") ? zlib.gunzipSync(buf).toString() : buf.toString();
    try { delivered += (JSON.parse(raw).batch || []).length; } catch {}
    res.writeHead(200, { "content-type": "application/json" });
    res.end('{"status":1}');
  });
});
const HOST = "http://127.0.0.1:9099";

(async () => {
  if (process.argv[2] === "child") {
    const ph = new PostHog("phc_test", { host: process.env.PH_HOST });
    ph.capture({ event: "x", distinctId: "d", properties: {} });
    await ph.flush();
    process.exit(0); // what a restart right after the trim looks like
  }
  await new Promise((r) => server.listen(9099, "127.0.0.1", r));

  await new Promise((r) => require("child_process")
    .fork(__filename, ["child"], { env: { ...process.env, PH_HOST: HOST } }).on("exit", r));
  console.log(`1) capture -> await flush() -> exit   requests reaching PostHog: ${requests}`);

  requests = 0; delivered = 0;
  const ph = new PostHog("phc_test", { host: HOST });
  for (let i = 0; i < 1200; i += 1) ph.capture({ event: "y", distinctId: `g-${i}`, properties: {} });
  await ph.flush();
  await new Promise((r) => setTimeout(r, 20000));
  await ph.shutdown();
  console.log(`2) 1200 captures, no await between them   delivered: ${delivered} / 1200`);
  server.close();
  process.exit(0);
})();
```

</details>

```
1) capture -> await flush() -> exit   requests reaching PostHog: 0
2) 1200 captures, no await between them   delivered: 1000 / 1200
```

**Where the 1000 comes from.** The default lives in `@posthog/core`, which `posthog-node@5.28.1` pins exactly (`"@posthog/core": "1.23.3"`, not a range) and does not override:

- Default: [`posthog-core-stateless.ts#L152-L156`](https://github.com/PostHog/posthog-js/blob/92c030b/packages/core/src/posthog-core-stateless.ts#L152-L156) → `this.maxQueueSize = Math.max(this.flushAt, options.maxQueueSize ?? 1000)`, so `Math.max(20, 1000)` with no options passed.
- Drop-oldest: [`posthog-core-stateless.ts#L895-L903`](https://github.com/PostHog/posthog-js/blob/92c030b/packages/core/src/posthog-core-stateless.ts#L895-L903) → `queue.shift()` then `_logger.info('Queue is full, the oldest event is dropped.')`. FIFO, so the discarded message is the oldest, and it's logged at `info`, which is why nothing surfaces.
- The option's JSDoc: [`types.ts#L10-L13`](https://github.com/PostHog/posthog-js/blob/92c030b/packages/core/src/types.ts#L10-L13). `maxQueueSize` isn't mentioned on posthog.com/docs/libraries/node at all, so the source is the only reference.

Confirm against what's installed rather than upstream:

```
cd backend && grep -n "maxQueueSize\|Queue is full" node_modules/@posthog/core/dist/posthog-core-stateless.js
# 95:  this.maxQueueSize = Math.max(this.flushAt, options.maxQueueSize ?? 1000);
# 477: if (queue.length >= this.maxQueueSize) {
# 479:   this._logger.info('Queue is full, the oldest event is dropped.');
```

The client is constructed with only `{ host }` at `telemetry-service.ts:151-153`, so the 1000 default applies.

**Fix needs both parts.** `maxQueueSize` above the 50k collection ceiling for the eviction, and `captureImmediate()` for the ordering, since it's the only path in this version that awaits the request and propagates its failure. Raising `maxQueueSize` alone still leaves the trim outrunning the send.

victorvhs017, line 660: **Non-blocking**

I went over this report a few times, and all tests seem to be true. It's an internal behavior in the SDK, and it doesn't block the PR, but I think it's worth raising here.

Passing `maxQueueSize` might be a small one-line improvement here if we don't want to go deeper. We didn't have this issue before because we had awaits in the loop.

The other issues are not regressions, more on the production-hardening side.

----

`flush()` doesn't deliver here, so the `streamTrim` two lines down can delete events that were never sent.

`capture()` enqueues asynchronously (`addPendingPromise(prepareEventMessage(...).then(...))`), so the queue is still empty when `_flush()` runs and it returns on `if (!queue.length) return`. Delivery actually happens later, via the SDK's own 10s `flushInterval`. Normal traffic still arrives, which is why this is invisible, but the ordering guarantee this line is here to provide doesn't exist. If the process restarts inside that window, or PostHog is failing, the shard is already trimmed and the batch is gone rather than retried next tick.

Separately, the capture loop above never awaits, so nothing drains between enqueues and the queue's 1000-entry cap evicts the oldest. 1200 aggregated groups in, 1000 out, shard trimmed regardless.

**To reproduce**, save as `backend/repro.js` and run `cd backend && node repro.js`:

<details>
<summary>repro.js</summary>

```js
const http = require("http");
const zlib = require("zlib");
const { PostHog } = require("posthog-node");

let requests = 0, delivered = 0;
const server = http.createServer((req, res) => {
  const chunks = [];
  req.on("data", (c) => chunks.push(c));
  req.on("end", () => {
    requests += 1;
    const buf = Buffer.concat(chunks);
    const raw = (req.headers["content-encoding"] || "").includes("gzip") ? zlib.gunzipSync(buf).toString() : buf.toString();
    try { delivered += (JSON.parse(raw).batch || []).length; } catch {}
    res.writeHead(200, { "content-type": "application/json" });
    res.end('{"status":1}');
  });
});
const HOST = "http://127.0.0.1:9099";

(async () => {
  if (process.argv[2] === "child") {
    const ph = new PostHog("phc_test", { host: process.env.PH_HOST });
    ph.capture({ event: "x", distinctId: "d", properties: {} });
    await ph.flush();
    process.exit(0); // what a restart right after the trim looks like
  }
  await new Promise((r) => server.listen(9099, "127.0.0.1", r));

  await new Promise((r) => require("child_process")
    .fork(__filename, ["child"], { env: { ...process.env, PH_HOST: HOST } }).on("exit", r));
  console.log(`1) capture -> await flush() -> exit   requests reaching PostHog: ${requests}`);

  requests = 0; delivered = 0;
  const ph = new PostHog("phc_test", { host: HOST });
  for (let i = 0; i < 1200; i += 1) ph.capture({ event: "y", distinctId: `g-${i}`, properties: {} });
  await ph.flush();
  await new Promise((r) => setTimeout(r, 20000));
  await ph.shutdown();
  console.log(`2) 1200 captures, no await between them   delivered: ${delivered} / 1200`);
  server.close();
  process.exit(0);
})();
```

</details>

```
1) capture -> await flush() -> exit   requests reaching PostHog: 0
2) 1200 captures, no await between them   delivered: 1000 / 1200
```

**Where the 1000 comes from.** The default lives in `@posthog/core`, which `posthog-node@5.28.1` pins exactly (`"@posthog/core": "1.23.3"`, not a range) and does not override:

- Default: [`posthog-core-stateless.ts#L152-L156`](https://github.com/PostHog/posthog-js/blob/92c030b/packages/core/src/posthog-core-stateless.ts#L152-L156) → `this.maxQueueSize = Math.max(this.flushAt, options.maxQueueSize ?? 1000)`, so `Math.max(20, 1000)` with no options passed.
- Drop-oldest: [`posthog-core-stateless.ts#L895-L903`](https://github.com/PostHog/posthog-js/blob/92c030b/packages/core/src/posthog-core-stateless.ts#L895-L903) → `queue.shift()` then `_logger.info('Queue is full, the oldest event is dropped.')`. FIFO, so the discarded message is the oldest, and it's logged at `info`, which is why nothing surfaces.
- The option's JSDoc: [`types.ts#L10-L13`](https://github.com/PostHog/posthog-js/blob/92c030b/packages/core/src/types.ts#L10-L13). `maxQueueSize` isn't mentioned on posthog.com/docs/libraries/node at all, so the source is the only reference.

Confirm against what's installed rather than upstream:

```
cd backend && grep -n "maxQueueSize\|Queue is full" node_modules/@posthog/core/dist/posthog-core-stateless.js
# 95:  this.maxQueueSize = Math.max(this.flushAt, options.maxQueueSize ?? 1000);
# 477: if (queue.length >= this.maxQueueSize) {
# 479:   this._logger.info('Queue is full, the oldest event is dropped.');
```

The client is constructed with only `{ host }` at `telemetry-service.ts:151-153`, so the 1000 default applies.

**Fix.** The `flush()` already on this line is the only call in this version that reports a failed delivery. `captureImmediate()` POSTs but swallows the error in `.catch((err) => console.error(err))`, and `shutdown()` swallows it in `logFlushError`; I tested all three against a 500 endpoint and only `flush()` with a non-empty queue rejects. So it just needs the queue populated when it runs: yield once after the capture loop so the deferred enqueues land, and pass `maxQueueSize` above the 50k collection ceiling. That also keeps the 100-per-POST batching, which matters, since a per-event send would turn ~2k POSTs into ~200k.

victorvhs017 (changes requested):

backend/src/lib/telemetry/metrics.ts
```diff
@@ -668,6 +688,64 @@ export const recordAlertDispatchOutcomeMetric = (params: { resourceType: string;
...
+  });
+
+export const recordTelemetryAggregationBacklogMetric = (params: { eventType: string; backlog: number }) =>
+  safely(() => {
+    if (!isTelemetryEnabled()) return;
+    telemetryAggregationBacklogHistogram.record(params.backlog, {
+      "telemetry_aggregation.event_type": params.eventType
+    });
+  });
+
```
victorvhs017: I feel like calling this telemetry aggregation is confusing with our own OTEL metrics. This is from an old pattern in the code (see the TELEMETRY_ENABLED env), but at least for the metrics we can name it more accordingly, and then when we look at the dashboards we know that they're talking about posthog/product analytics and not about our own aot collector

backend/src/lib/telemetry/telemetry-attributes.ts
```diff
@@ -19,6 +19,8 @@ export const INFISICAL_CORE_METER_ATTRIBUTES = [
   "audit_log.actor_type",
   "audit_log.backend",
   "audit_log.drop_reason",
+  "telemetry_aggregation.event_type",
+  "telemetry_aggregation.drop_reason",
```
victorvhs017: same here

victorvhs017 (changes requested):

victorvhs017 (approved):

PR #7717, opened 2026-08-19
fix: enforce name constraints, path length, and EKU on identity TLS cert auth

Thiago-AS:
## Context

Chain validation for TLS cert auth was ignoring three RFC 5280 rules, and each one is a way to authenticate as an identity you shouldn't:

- **Name constraints** — if you hold a sub-CA under the pinned anchor that's restricted to a namespace, you could mint a leaf for any name and auth as any identity pinned to that anchor. Now honored, including constraints on the anchor itself (operators often pin a constrained sub-CA directly, not the root). Subtree matching is delegated to `pkijs` instead of hand-rolled.
- **`pathLenConstraint`** — a CA capping how many intermediates may sit below it was not enforced. Now it is.
- **Extended Key Usage** — a leaf issued for `serverAuth` or `codeSigning` authenticated fine, so anyone with a server cert under the CA could impersonate the machine identity. Now rejected unless the EKU includes `clientAuth` or `anyExtendedKeyUsage`. Certs with no EKU stay accepted (same as OpenSSL/Go), so existing setups don't break.

Name constraint checking only kicks in when something on the path actually asserts constraints, so chains without them validate exactly as before.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [x] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
```diff
@@ -1,22 +1,156 @@
...
+ *
+ * A certificate that omits the extension is unconstrained and stays accepted, which is how OpenSSL
+ * and Go read a missing EKU and what keeps existing leaves that carry no EKU working. Only a
+ * certificate that explicitly enumerates purposes and leaves client authentication out is rejected.
+ */
+export const permitsClientAuth = (cert: TNativeX509): boolean => {
+  const usages = new x509.X509Certificate(cert.raw).getExtension(x509.ExtendedKeyUsageExtension)?.usages;
+  if (!usages) return true;
+  return usages.some((oid) => oid === CLIENT_AUTH_EKU_OID || oid === ANY_PURPOSE_EKU_OID);
+};
```
victorvhs017, line 504: `permitsClientAuth` parses with `@peculiar/x509` on every login now. If Node's crypto.X509Certificate accepted the cert but peculiar throws (stricter ASN.1), we return 500 instead of 401. 

We could wrap and treat parse failure as deny instead of internal server error.

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
```diff
@@ -1,22 +1,156 @@
...
+
+const pkiCryptoEngine = new CryptoEngine({ name: "identity-tls-cert-auth", crypto: webcrypto as Crypto });
 
 const isWithinValidityWindow = (cert: TNativeX509, at: Date): boolean =>
   new Date(cert.validFrom) <= at && at <= new Date(cert.validTo);
 
+const isSelfIssued = (cert: TNativeX509): boolean => cert.subject === cert.issuer;
+
+const basicConstraintsPathLength = (cert: TNativeX509): number | undefined =>
+  new x509.X509Certificate(cert.raw).getExtension(x509.BasicConstraintsExtension)?.pathLength;
```
victorvhs017: Same thing here

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
```diff
@@ -1,22 +1,156 @@
...
+      initialExcludedSubtreesSet: anchorConstraints?.excludedSubtrees ?? []
+    },
+    pkiCryptoEngine
+  );
+
+  if (result) return null;
+  if (resultCode === PERMITTED_SUBTREE_VIOLATION_CODE || resultCode === EXCLUDED_SUBTREE_VIOLATION_CODE) {
+    return { ok: false, reasonCode: "name_constraint_violation" };
+  }
+  return { ok: false, reasonCode: "ca_verification_failed" };
```
victorvhs017: Same pattern as EKU: Node validates the chain, then pkijs re-checks for name constraints. pkijs can fail on chains Node accepts (e.g. Ed25519 → resultCode: -1), but we map that to generic "ca_verification_failed" with no log, so a valid chain looks broken.

I think it's worth a warn log with resultCode/message.

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-service.ts
```diff
@@ -177,6 +190,18 @@ export const identityTlsCertAuthServiceFactory = ({
...
+          throw new UnauthorizedError({
+            message: CHAIN_FAILURE_MESSAGES.issuer_client_auth_usage_not_allowed,
+            detail: {
+              reasonCode: "issuer_client_auth_usage_not_allowed",
+              identityId: identity.id,
+              orgId: identity.orgId,
+              identityName: identity.name
+            }
+          });
+        }
```
victorvhs017: We are adding CA EKU to single-hop and full constraint checks in chain mode. We should have the same  anchor name constraints and CA validity checks for single-hops

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
victorvhs017, line 504: We only do `permitsClientAuth` at login time. We could check this when we attach/update the cert and raise the issue before the user gets a login error.

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
```diff
@@ -62,80 +203,143 @@ export const verifyClientCertificateChain = ({
...
+    for (const path of orderedPaths) {
+      const pathLengthFailure = enforcePathLength(path);
+      // eslint-disable-next-line no-await-in-loop -- stop at the first path that validates
+      const failure = pathLengthFailure ?? (await enforceNameConstraints(path, now));
+      if (!failure) return { ok: true };
+      constraintFailure ??= failure;
+    }
+  } catch {
+    return { ok: false, reasonCode: "ca_verification_failed" };
+  }
```
victorvhs017: The try around the whole path loop means a throw on path 1 skips paths 2..n and returns "ca_verification_failed", which fights the multi-path design. 

We could move the catch inside the loop (failed path + continue) and log the error.

victorvhs017 (changes requested):

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
Thiago-AS: it only runs on issuers that already cleared permitsClientAuth, so an unparseable cert can't reach it

backend/src/services/identity-tls-cert-auth/identity-tls-cert-auth-fns.ts
```diff
@@ -79,43 +355,107 @@ const nameConstraintsOf = (cert: Certificate): NameConstraints | null => {
...
+  const anchorState = pathConstraints[pathConstraints.length - 1];
+  const anchorConstraints = anchorState.status === "present" ? anchorState.constraints : null;
 
-  const engine = new CertificateChainValidationEngine({
-    trustedCerts: [anchor],
-    // pkijs takes the chain end-entity last.
-    certs: parsedPath.slice(0, -1).reverse(),
-    checkDate: now
+  const evaluate = async (path: Certificate[]): Promise<TVerificationFailure | null> => {
+    const { result, resultCode, resultMessage } = await new NameConstraintsEngine(path).verify(
```
victorvhs017, line 424: `pkijs` also runs certificate `policyMappings` logic, which can crash or fail on intermediates in 3+ cert chains. Valid logins and real namespace violations both end up as generic `ca_verification_failed`, even though we don't validate policy mappings at all, only name constraints.

victorvhs017 (changes requested):

victorvhs017 (dismissed):

victorvhs017 (approved):

PR #7731, opened 2026-08-20
fix(identity): optimize lockout lookup to prevent full keyscan that caused slow machine identity page load

lobocv:
## Context

Listing project machine identities ran one Redis keyspace scan per identity to compute lockout status, so the endpoint's cost scaled with both page size and total Redis key volume. This replaces that with a single batched `MGET` per page.

Universal auth stores its lockout state under a key whose last segment is the client ID, and Postgres already holds that client ID on a table the list query already joins. So instead of discovering the key by scanning, the lookup derives it and reads every identity on the page in one round trip. LDAP auth keys its lockout by the submitted directory username, which is never persisted here, so those rows still fall back to the pattern scan. Only universal auth and LDAP have lockout configuration columns, so every other identity now does zero Redis work instead of a full sweep.

`LOCKOUT_CAPABLE_AUTH_METHODS` and `isExactlyResolvable` make that split explicit rather than implicit. The scan this replaced was method-agnostic — whatever wrote a lockout key produced a badge automatically — and an enumerated list is not. Any method added to the constant without an exact-key rule still returns correct results through the scan fallback, so the read path cannot silently stop seeing keys the write path produces.

Measured on real Redis at a 1,000,020-key keyspace:

| Operation | 20 keys | 1,000,020 keys |
|---|---|---|
| Old: pattern scan, one identity | 2 ms | ~98 ms |
| Old: pattern scan × 20 (one page) | ~20 ms | ~1950 ms |
| New: single `MGET` of 20 exact keys | 1 ms | 1–2 ms |

The old path scales linearly with keyspace size; the new path is flat.

### Scope notes for reviewers

- **This speeds up three endpoints, not one.** `membershipIdentity.listMemberships` also backs the org identity-membership listing (`identity-org-membership-router.ts:336`) and the cert-manager identity listing (`cert-manager-access-routers/identities-router.ts:56`). All three get the batched lookup. No caller's response shape changes.
- **The project listing now fails open on Redis errors; the out-of-scope endpoints still fail closed.** If Redis is unavailable this path logs and omits the lockout badge rather than failing the page, which is a deliberate improvement over the previous 500. Enforcement is unaffected — the login gate still fails closed — so only the informational badge degrades. The inconsistency with the org-level and detail endpoints is a scope boundary, not an oversight.
- **`universalAuthClientId` is deliberately not serialised.** It now flows from the DAL into the service layer to build the Redis key, and is kept out of the API by the response schemas selecting explicit fields. The e2e test asserts its absence rather than assuming it, so a future `.passthrough()` cannot leak it unnoticed.
- **One intentional behaviour change:** orphaned lockout keys no longer produce badges. A key still inside its TTL for an auth method the identity no longer has — an LDAP key after LDAP auth was revoked, or a universal auth key from a previous attach with a different client ID — used to surface as a badge. It no longer does.

### Deliberately out of scope

Each is the same class of problem and worth its own ticket:

1. The org-level identity listing (`identity-service.ts:538`, `identity-v2/identity-service.ts:575`) and the single-identity detail paths (`identity-service.ts:344`, `identity-v2/identity-service.ts:425`) loop the singular helper the same way. Their DALs already join the universal auth table, so each is the same small change.
2. `clearLockouts` scans the whole keyspace via `keyStore.deleteItems({ pattern })`, backing the admin clear-lockout endpoints.
3. Under a Redis Cluster deployment, a multi-key `MGET` without hash tags returns `CROSSSLOT`. This path catches and fails open, so badges would vanish silently. Pre-existing rather than introduced here — `getKeysByPattern` only scans the primary node, and three other call sites already do multi-key `getItems`, including on the license hot path — but it wants a cluster-safe batched read.
4. The cohesion refactor: one module owning the key format, the state type, both login paths and both clear paths, removing the `parts[3]` string parsing and the duplicated lockout state type.
5. The listing ignores the `orderBy` / `orderDirection` the frontend sends, and `LIMIT`/`OFFSET` without a stable order can repeat or skip rows across pages.

## Screenshots

Not applicable — no UI change.

## Steps to verify the change

No schema change, so no migration is involved.

1. Lint and types, from the repo root:
   ```
   make reviewable-api
   ```
2. Unit tests — `identity-fns.test.ts` covers the batching logic against a mocked key store (empty page, non-lockout-capable methods, exact-key hit, counting-only record, one `MGET` for a 20-row page, LDAP fallback, missing-client-ID fallback, both methods locked out, malformed value, Redis error); `membership-identity-service.test.ts` covers the service wiring:
   ```
   cd backend && npx vitest run -c vitest.unit.config.mts src/services/identity/identity-fns.test.ts src/services/membership-identity/membership-identity-service.test.ts
   ```
3. End-to-end regression test. This is the one that matters: a mocked key store cannot prove the derived key matches the key the login path writes, so this drives three real failed logins through the universal auth login endpoint and asserts the listing endpoint reports the resulting lockout. It also asserts `universalAuthClientId` is absent from the response.

   Local setup: copy `.env.test.example` to `.env.test` at the repo root, then `docker compose -f docker-compose.dev.yml up -d db redis`.
   ```
   cd backend && npx vitest run -c vitest.e2e.config.mts e2e-test/routes/v1/identity-lockout-listing.spec.ts
   ```
   Two caveats for anyone running this on an Apple Silicon host: the harness cannot boot natively there, because `@infisical/quic` resolves a native binding from `@infisical/quic-darwin-arm64`, which is not published — run it in a linux container as CI does. And the suite's `DROP SCHEMA CASCADE` teardown needs `max_locks_per_transaction` above the default 64 for this schema's object count; the `db-test` service in `docker-compose.dev.yml` already sets 512.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally
- [x] Updated docs (if needed) — not needed; no user-facing behaviour or API change
- [x] Updated CLAUDE.md files (if needed) — not needed; no new service, pattern, or architectural shift

### DIFF
--- a/backend/e2e-test/routes/v1/identity-lockout-listing.spec.ts
+++ b/backend/e2e-test/routes/v1/identity-lockout-listing.spec.ts
@@ -1,6 +1,10 @@
 import { IdentityAuthMethod, OrgMembershipRole, ProjectMembershipRole } from "@app/db/schemas";
 import { seedData1 } from "@app/db/seed-data";
 
+type TIdentityMembershipsResponse = {
+  identityMemberships: { identity: { id: string; activeLockoutAuthMethods: string[] } }[];
+};
+
 describe("Project identity listing lockout indicators", async () => {
   test("reports an active universal auth lockout and hides the client id", async () => {
     let identityId: string | undefined;
@@ -62,10 +66,11 @@ describe("Project identity listing lockout indicators", async () => {
       });
       expect(listRes.statusCode).toBe(200);
 
-      const row = listRes.json().identityMemberships.find((m) => m.identity.id === identity.id);
+      const { identityMemberships } = JSON.parse(listRes.payload) as TIdentityMembershipsResponse;
+      const row = identityMemberships.find((m) => m.identity.id === identity.id);
       expect(row).toBeDefined();
-      expect(row.identity.activeLockoutAuthMethods).toContain(IdentityAuthMethod.UNIVERSAL_AUTH);
-      expect(row.identity).not.toHaveProperty("universalAuthClientId");
+      expect(row?.identity.activeLockoutAuthMethods).toContain(IdentityAuthMethod.UNIVERSAL_AUTH);
+      expect(row?.identity).not.toHaveProperty("universalAuthClientId");
     } catch (err) {
       failure = err instanceof Error ? err : new Error(String(err));
     } finally {
@@ -95,7 +100,8 @@ describe("Project identity listing lockout indicators", async () => {
       headers: { authorization: `Bearer ${jwtAuthToken}` }
     });
     expect(listRes.statusCode).toBe(200);
-    listRes.json().identityMemberships.forEach((m) => {
+    const { identityMemberships } = JSON.parse(listRes.payload) as TIdentityMembershipsResponse;
+    identityMemberships.forEach((m) => {
       expect(m.identity.activeLockoutAuthMethods).toEqual([]);
     });
   });
--- a/backend/e2e-test/routes/v1/identity-lockout-listing.spec.ts
+++ b/backend/e2e-test/routes/v1/identity-lockout-listing.spec.ts
@@ -3,65 +3,89 @@ import { seedData1 } from "@app/db/seed-data";
 
 describe("Project identity listing lockout indicators", async () => {
   test("reports an active universal auth lockout and hides the client id", async () => {
-    const createRes = await testServer.inject({
-      method: "POST",
-      url: "/api/v1/identities",
-      body: { name: "lockout-listing-mi", role: OrgMembershipRole.Admin, organizationId: seedData1.organization.id },
-      headers: { authorization: `Bearer ${jwtAuthToken}` }
-    });
-    expect(createRes.statusCode).toBe(200);
-    const { identity } = createRes.json();
+    let identityId: string | undefined;
+    let failure: Error | undefined;
 
-    const attachRes = await testServer.inject({
-      method: "POST",
-      url: `/api/v1/auth/universal-auth/identities/${identity.id}`,
-      body: { lockoutEnabled: true, lockoutThreshold: 3, lockoutDurationSeconds: 300 },
-      headers: { authorization: `Bearer ${jwtAuthToken}` }
-    });
-    expect(attachRes.statusCode).toBe(200);
-    const { clientId } = attachRes.json().identityUniversalAuth;
-
-    await testServer.inject({
-      method: "POST",
-      url: `/api/v1/projects/${seedData1.project.id}/memberships/identities/${identity.id}`,
-      body: { roles: [{ role: ProjectMembershipRole.Member }] },
-      headers: { authorization: `Bearer ${jwtAuthToken}` }
-    });
-
-    const secretRes = await testServer.inject({
-      method: "POST",
-      url: `/api/v1/auth/universal-auth/identities/${identity.id}/client-secrets`,
-      body: { description: "test" },
-      headers: { authorization: `Bearer ${jwtAuthToken}` }
-    });
-    expect(secretRes.statusCode).toBe(200);
-
-    for (let i = 0; i < 3; i += 1) {
-      // eslint-disable-next-line no-await-in-loop
-      await testServer.inject({
+    try {
+      const createRes = await testServer.inject({
         method: "POST",
-        url: "/api/v1/auth/universal-auth/login",
-        body: { clientId, clientSecret: "definitely-the-wrong-secret" }
+        url: "/api/v1/identities",
+        body: {
+          name: "lockout-listing-mi",
+          role: OrgMembershipRole.Admin,
+          organizationId: seedData1.organization.id
+        },
+        headers: { authorization: `Bearer ${jwtAuthToken}` }
       });
+      expect(createRes.statusCode).toBe(200);
+      const { identity } = createRes.json();
+      identityId = identity.id;
+
+      const attachRes = await testServer.inject({
+        method: "POST",
+        url: `/api/v1/auth/universal-auth/identities/${identity.id}`,
+        body: { lockoutEnabled: true, lockoutThreshold: 3, lockoutDurationSeconds: 300 },
+        headers: { authorization: `Bearer ${jwtAuthToken}` }
+      });
+      expect(attachRes.statusCode).toBe(200);
+      const { clientId } = attachRes.json().identityUniversalAuth;
+
+      const membershipRes = await testServer.inject({
+        method: "POST",
+        url: `/api/v1/projects/${seedData1.project.id}/memberships/identities/${identity.id}`,
+        body: { roles: [{ role: ProjectMembershipRole.Member }] },
+        headers: { authorization: `Bearer ${jwtAuthToken}` }
+      });
+      expect(membershipRes.statusCode).toBe(200);
+
+      const secretRes = await testServer.inject({
+        method: "POST",
+        url: `/api/v1/auth/universal-auth/identities/${identity.id}/client-secrets`,
+        body: { description: "test" },
+        headers: { authorization: `Bearer ${jwtAuthToken}` }
+      });
+      expect(secretRes.statusCode).toBe(200);
+
+      for (let i = 0; i < 3; i += 1) {
+        // eslint-disable-next-line no-await-in-loop
+        await testServer.inject({
+          method: "POST",
+          url: "/api/v1/auth/universal-auth/login",
+          body: { clientId, clientSecret: "definitely-the-wrong-secret" }
+        });
+      }
+
+      const listRes = await testServer.inject({
+        method: "GET",
+        url: `/api/v1/projects/${seedData1.project.id}/memberships/identities?limit=100`,
+        headers: { authorization: `Bearer ${jwtAuthToken}` }
+      });
+      expect(listRes.statusCode).toBe(200);
+
+      const row = listRes.json().identityMemberships.find((m) => m.identity.id === identity.id);
+      expect(row).toBeDefined();
+      expect(row.identity.activeLockoutAuthMethods).toContain(IdentityAuthMethod.UNIVERSAL_AUTH);
+      expect(row.identity).not.toHaveProperty("universalAuthClientId");
+    } catch (err) {
+      failure = err instanceof Error ? err : new Error(String(err));
+    } finally {
+      if (identityId) {
+        const deleteRes = await testServer.inject({
+          method: "DELETE",
+          url: `/api/v1/identities/${identityId}`,
+          headers: { authorization: `Bearer ${jwtAuthToken}` }
+        });
+        // Only report a cleanup failure when nothing else already failed, so it never
+        // masks the assertion this test exists to catch.
+        if (!failure && deleteRes.statusCode !== 200) {
+          failure = new Error(`cleanup DELETE /api/v1/identities/${identityId} returned ${deleteRes.statusCode}`);
+        }
+      }
     }
 
-    const listRes = await testServer.inject({
-      method: "GET",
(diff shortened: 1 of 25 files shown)


backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,93 @@ export const getIdentityActiveLockoutAuthMethods = async (
   return Array.from(activeLockoutAuthMethods);
 };
 
+export type TIdentityLockoutLookup = {
```
lobocv, line 34: Add a comment to this type to clarify why it is needed (confirm my understanding as well)

As far as I understand, this acts as a proxy for the Redis key, defining all the metadata required in order to format the key without actually passing the key itself.

backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,93 @@ export const getIdentityActiveLockoutAuthMethods = async (
...
+  try {
+    return (JSON.parse(raw) as { lockedOut?: boolean }).lockedOut === true;
+  } catch {
+    return false;
+  }
+};
+
+// Universal auth keys its lockout by client id, which the caller already has, so it is the only
+// lockout-capable method resolvable by exact key. Every other lockout-capable method falls
+// through to the pattern scan below.
```
lobocv, line 58: Add a comment here that we should refactor this library in the future so that we can do exact lookups on any identity + auth method so we prevent doing large keyscans that hammer redis and slow down page load times.

backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,93 @@ export const getIdentityActiveLockoutAuthMethods = async (
...
+  identities: TIdentityLockoutLookup[],
+  keyStore: TLockoutKeyStore
+): Promise<Record<string, IdentityAuthMethod[]>> => {
+  const result: Record<string, IdentityAuthMethod[]> = {};
+  if (!identities.length) return result;
+
+  const add = (identityId: string, method: IdentityAuthMethod) => {
+    if (!result[identityId]) result[identityId] = [];
+    if (!result[identityId].includes(method)) result[identityId].push(method);
+  };
```
lobocv, line 90: Add a one-liner comment explaining what the add function is doing.

backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,106 @@ export const getIdentityActiveLockoutAuthMethods = async (
...
+      key: KeyStorePrefixes.IdentityLockoutState(
+        el.id,
+        IdentityAuthMethod.UNIVERSAL_AUTH,
+        el.universalAuthClientId as string
+      )
+    }));
+
+  if (exactLookups.length) {
+    try {
+      const values = await keyStore.getItems(exactLookups.map((el) => el.key));
```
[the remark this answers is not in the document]
lobocv: Fixed in 773f1fa421, though with one correction to the premise.

The cross-slot hazard is real: `getItems` maps straight to `mget`, and since these keys are spread by identity id, any page holding more than one universal auth identity would trip `CROSSSLOT` under `REDIS_CLUSTER_HOSTS` and the surrounding catch would swallow it into "no badges for the whole page".

But the previous per-identity path was not accurate under cluster either. It went through `getKeysByPattern`, which calls `primaryRedis.scan` — and `SCAN` against one node only walks that node's slots, so it already missed keys living elsewhere. The honest regression is narrower than "these states could previously be retrieved": the old path's *reads* were single-key `GET`s and so were correctly routed, meaning it degraded to partial results, whereas the batched read degraded to none.

The suggested fix does not work as written, though. ioredis routes a pipeline by its first key, so pipelining individual `GET`s hits the same cross-slot wall. Hash-tagging the keys would genuinely fix it, but it changes the key format the login path writes and would orphan every lockout already in Redis.

What landed instead is a retry: catch `CROSSSLOT` and re-read the same keys as concurrent single-key `GET`s, which Cluster routes independently. That costs N commands but still one round trip of latency, against the N full keyspace scans this path replaced, and it is free on single-node Redis because the retry never fires. Covered by a new unit test.

This is deliberately a hedge rather than a real fix. A properly cluster-aware batched read belongs in the keystore, not here — the same `getItems`/`mget` pattern is already used by `license-service.ts:317`, `project-events-sse-service.ts:135`, and `telemetry-service.ts:472`, including on the license hot path, so fixing it in one caller would leave the platform gap and set an inconsistent precedent. There is also no clustered environment in CI, so anything more ambitious would ship unverified. We will do it properly when this moves to Go.

backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,106 @@ export const getIdentityActiveLockoutAuthMethods = async (
...
+  // above) still needs a pattern lookup.
+  const scanLookups = identities.filter((el) =>
+    el.authMethods.some((method) => LOCKOUT_CAPABLE_AUTH_METHODS.includes(method) && !isExactlyResolvable(el, method))
+  );
+
+  await Promise.all(
+    scanLookups.map(async (el) => {
+      try {
+        const methods = await getIdentityActiveLockoutAuthMethods(el.id, keyStore);
+        methods.forEach((method) => add(el.id, method as IdentityAuthMethod));
```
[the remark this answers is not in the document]
lobocv: Good catch — fixed in 773f1fa421, applying the suggestion.

Confirming the analysis: the scan matches every key under `lockout:identity:<id>:*` and reported each one, so an identity with LDAP attached would surface a badge for a method since revoked, or for a universal auth key from an earlier attach whose client id had changed. The exact-key read had already correctly found nothing for the *current* client id, and the scan then overrode that with a stale one.

Worth naming why this was inconsistent rather than simply wrong: dropping orphaned keys was intended, but it only took effect for identities resolved exclusively through the exact-key path. Any identity that also hit the scan path kept surfacing them. The filter makes the behaviour uniform.

Two deferred items fall out of this for free — the redundant double lookup for identities holding both universal auth and LDAP (the scan's `universal-auth` result is now skipped, since the exact read is authoritative for it), and the unvalidated `IdentityAuthMethod` cast, which `el.authMethods.includes(authMethod)` now rejects at runtime for any unrecognised key segment.

One tradeoff, recorded deliberately: if the exact-key read fails and fails open, the scan will no longer backfill `universal-auth` where it previously might. That is the right call — the exact key is the authoritative answer, so inferring from a keyspace sweep when it is unavailable is worse than showing no badge.

Added two unit tests, both verified to fail without the change: one for the changed-client-id case, one for a method the identity no longer holds. The ten pre-existing tests pass unmodified.

backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,136 @@ export const getIdentityActiveLockoutAuthMethods = async (
...
+const readLockoutStates = async (keyStore: TLockoutKeyStore, keys: string[]) => {
+  try {
+    return await keyStore.getItems(keys);
+  } catch (err) {
+    if (!(err instanceof Error) || !err.message.includes("CROSSSLOT")) throw err;
+    return Promise.all(keys.map((key) => keyStore.getItem(key)));
+  }
+};
+
+export const getActiveLockoutAuthMethodsForIdentities = async (
```
Thiago-AS, line 77: This helper is only wired into `membershipIdentity.listMemberships`. The two higher traffic identity listings still do the per row scan it replaces:

- **identity-service.ts:533** (searchOrgIdentities, the org Identities page)
- **identity-v2/identity-service.ts:569** (enrichWithLockouts, the v2 identity search)

Both still `Promise.all` a per row `getIdentityActiveLockoutAuthMethods`, while `getKeysByPattern` does an unbounded SCAN over the whole keyspace. So an org listing with 100 identities per page issues 100 full keyspace sweeps.

Both call sites are close to ready. `identity-org-dal.ts` (searchIdentities) and `identity-v2/identity-membership-dal.ts` (searchIdentitiesV2) already left-join IdentityUniversalAuth and call buildAuthMethods. They just don't select clientId. It's the same three-line DAL change made here for `membership-identity-dal.ts`

There's also a behavioral split now. The new path filters scan results to auth methods the identity still holds, so an identity with an orphaned lockout key shows a badge on the org page but no badge on the project page. Two endpoints, two answers for the same identity.

backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,136 @@ export const getIdentityActiveLockoutAuthMethods = async (
...
+  if (exactLookups.length) {
+    try {
+      const values = await readLockoutStates(
+        keyStore,
+        exactLookups.map((el) => el.key)
+      );
+      values.forEach((raw, idx) => {
+        if (isLockedOut(raw)) add(exactLookups[idx].identityId, IdentityAuthMethod.UNIVERSAL_AUTH);
+      });
+    } catch (err) {
```
Thiago-AS, line 110: This swaps the previous behavior, where a Redis error propagates and the request fails, for a page wide silent omission.

Two things concern me:

- The response can't express "unknown". `activeLockoutAuthMethods: []` means "not locked out" to every consumer. `IdentityTable.tsx` renders the Locked out: … tooltip only for a non-empty array, so during a Redis blip an admin checking why an identity can't authenticate sees a clean row and concludes lockout isn't the cause. For this field, a failed request may be a better answer than a confidently wrong one.
- The blast radius is the whole page, not one row. The old per-identity call meant one bad read affected one identity. This catch wraps the single MGET covering every UA identity on the page, so one error blanks them all at once.

backend/e2e-test/routes/v1/identity-lockout-listing.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  test("reports no lockout for an identity that has never failed a login", async () => {
+    const listRes = await testServer.inject({
+      method: "GET",
+      url: `/api/v1/projects/${seedData1.project.id}/memberships/identities?limit=100`,
+      headers: { authorization: `Bearer ${jwtAuthToken}` }
+    });
+    expect(listRes.statusCode).toBe(200);
+    const { identityMemberships } = JSON.parse(listRes.payload) as TIdentityMembershipsResponse;
+    identityMemberships.forEach((m) => {
+      expect(m.identity.activeLockoutAuthMethods).toEqual([]);
```
Thiago-AS: This might cause flakyness. This assertion is scoped to whatever happens to be in the shared seed project rather than to data the test owns, which gives it two failure modes that have nothing to do with the code under test.

backend/src/services/identity/identity-fns.ts
```diff
@@ -24,6 +25,136 @@ export const getIdentityActiveLockoutAuthMethods = async (
...
+// clustered deployments accurate, since Cluster routes each read independently: it costs N
+// commands but still one round trip of latency, against the N full keyspace scans this path
+// replaced. Hash-tagging the keys would avoid the retry entirely, but it would change the key
+// format the login path writes and orphan every lockout already held in Redis, so a properly
+// cluster-aware batched read is left to the Go rewrite.
+const readLockoutStates = async (keyStore: TLockoutKeyStore, keys: string[]) => {
+  try {
+    return await keyStore.getItems(keys);
+  } catch (err) {
+    if (!(err instanceof Error) || !err.message.includes("CROSSSLOT")) throw err;
```
Thiago-AS: How reliable is using the error message here? Is there any other way we can assert other than the message?

backend/e2e-test/routes/v1/identity-lockout-listing.spec.ts
lobocv: Good catch! This is one of the hardest things about end to end tests. I didn't realize it was using a shared fixture. It will become very important as we built out more tests to have to have fixture generators rather than a shared fixtures. The costs is that it takes tests longer to provision fixtures for every test case. It's all a balance..

ill see what I can do here...

lobocv:
Closing this as it is solved in another PR

PR #7738, opened 2026-08-20
improvement(secrets): unify overview navigation and query state

andrewhuhh:
## Context

This PR implements the navigation and filter-state portion of the Secrets v3 migration tracked by [SECRETS-534](https://linear.app/infisical/issue/SECRETS-534/migrate-secrets-area-to-v3). [SECRETS-559](https://linear.app/infisical/issue/SECRETS-559/unify-secrets-overview-navigation-and-query-state) unifies the cohesive work from SECRETS-535, SECRETS-537, and SECRETS-543.

Previously, environment selection, search, tags, and resource filters were split between local state, localStorage, and transient one-shot URL ingestion. Folder rows and breadcrumbs also reset or restored transient filter history, which could lose view state during navigation. Filter navigation could additionally interrupt pending batch changes or clear bulk selections.

This change keeps shareable navigation state—environment slugs, secret path, and resource type filters—in the URL while keeping sensitive search text and tag slugs in local state. Existing links containing search or tag parameters are applied once and immediately stripped with replacement navigation. The overview resolves stored environment preferences before its first enabled query, normalizes inaccessible environments, preserves unrelated safe query parameters through folder and breadcrumb transitions, and keeps search, tag, batch-mode, and bulk-selection state intact during view-only filtering. Quick-search result navigation now carries only path, environment, and resource type in the URL. The change also corrects nested dropdown trigger composition and adds accessible names to icon-only controls.

## Steps to verify the change

1. Install frontend dependencies using the repository's supported workflow.
2. Run the focused state contracts:
   `cd frontend && node --test --import tsx src/pages/secret-manager/OverviewPage/overviewSearchState.test.ts src/pages/secret-manager/OverviewPage/components/ResourceSearchInput/resourceSearchState.test.ts`
3. Run the repository frontend validation gate:
   `/Users/andrewhuang/Documents/Agent\ Workspace/skills/validate-infisical-frontend/scripts/frontend-validation-control gate --worktree "$PWD"`
4. In a project with multiple accessible environments, open Secrets Overview and verify:
   - selecting environments updates the canonical environment URL parameters;
   - resource type filters update `filterBy`, while search text and tag slugs do not remain in the URL;
   - incoming links with `search` or `tags` apply the filter once and then remove those parameters;
   - quick-search results navigate to the correct path and environment without exposing resource identifiers in the URL;
   - folder rows and breadcrumbs preserve local search/tag filters and unrelated safe query parameters;
   - search, filtering, and pagination continue to work with batch mode enabled and with bulk selections active;
   - a stored environment preference is used by the first enabled overview query;
   - an inaccessible environment slug is normalized safely.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the conventional-commit format.
- [x] Tested locally

### DIFF
--- a/frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
+++ b/frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
@@ -250,10 +250,12 @@ import {
 } from "./components";
 import {
   hasOverviewScopeChanged,
+  hasSensitiveOverviewSearchState,
   normalizeOverviewEnvironments,
+  parseOverviewTags,
   resolveOverviewEnvironmentSlugs,
   serializeOverviewResourceFilter,
-  serializeOverviewTags,
+  stripSensitiveOverviewSearchState,
   updateOverviewSecretPath
 } from "./overviewSearchState";
 
@@ -337,7 +339,8 @@ const OverviewPageContent = () => {
     ) ?? false;
   const isProjectV3 = currentProject?.version === ProjectVersion.V3;
   const projectSlug = currentProject?.slug as string;
-  const searchFilter = routerSearch.search;
+  const [searchFilter, setSearchFilter] = useState(routerSearch.search);
+  const [tagFilter, setTagFilter] = useState(() => parseOverviewTags(routerSearch.tags));
   const secretPath = (routerSearch?.secretPath as string) || "/";
   const { subscription } = useSubscription();
   const { mutateAsync: importVaultSecrets } = useImportVaultSecrets();
@@ -375,20 +378,10 @@ const OverviewPageContent = () => {
         nextFilter[rowType] = true;
       });
 
-    if (routerSearch.tags) nextFilter[RowType.Secret] = true;
+    if (Object.keys(tagFilter).length > 0) nextFilter[RowType.Secret] = true;
 
     return nextFilter;
-  }, [routerSearch.filterBy, routerSearch.tags]);
-
-  const tagFilter = useMemo(
-    () =>
-      (routerSearch.tags ?? "").split(",").reduce<Record<string, boolean>>((acc, tag) => {
-        const tagSlug = tag.trim();
-        if (tagSlug) acc[tagSlug] = true;
-        return acc;
-      }, {}),
-    [routerSearch.tags]
-  );
+  }, [routerSearch.filterBy, tagFilter]);
 
   const [selectedEntries, setSelectedEntries] = useState<{
     // selectedEntries[name/key][envSlug][resource]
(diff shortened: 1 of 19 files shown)


andrewhuhh:
@greptile review

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
scott-ray-wilson, line 1: search/filtering with batch mode enables now breaks 

https://github.com/user-attachments/assets/468225cd-33c2-4e34-aa4d-10156e5e5f65

scott-ray-wilson: same thing with bulk selection, I can no longer bulk select and search/filter

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
```diff
@@ -457,85 +479,64 @@ const OverviewPageContent = () => {
...
-  // Apply one-shot deep-link inputs to local filters, then strip them from the URL in a SINGLE
-  // navigate. These arrive from notifications, legacy dashboard bookmarks (`tags`, `filterBy`),
-  // or the secret reference tree (`environments` + `search`). Handling them in one effect/navigate
-  // (rather than two racing effects) guarantees every param is cleared after it's applied. That
-  // matters most for `environments`: re-selecting the same environment from the reference tree
-  // changes the param again and re-fires this effect instead of being a no-op. Runs reactively
-  // (not mount-only) because the tree is rendered inside this page, so navigating from a node
-  // updates the params without remounting.
+  const initializedEnvironmentProjectId = useRef<string | null>(null);
   useEffect(() => {
```
scott-ray-wilson, line 488: this fires post mount so we actually fetch the overview and then the single stored env 

Image

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
scott-ray-wilson, line 1: I think you could argue either way but do you think each filter change should be a push? the back button ends up being a lot of walks back instead of a single press

https://github.com/user-attachments/assets/cbefe6b5-0c95-452a-a13e-2557645d0d40

scott-ray-wilson (dismissed):

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
andrewhuhh: on it

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
andrewhuhh: fixed

andrewhuhh:
@greptile review

scott-ray-wilson (approved):
LGTM

PR #7744, opened 2026-08-21
feat(encryption): implement root encryption key rotation

victorvhs017:
## Context

Self-hosted operators had no supported way to rotate `ENCRYPTION_KEY`, the env var that wraps the in-database root key protecting every secret. A leaked key stayed valid indefinitely, and compliance programs that require periodic rotation had no first-class path.

This adds a staged rotation workflow: generating a key is inert until a pod boots with the new value. On first boot, KMS trial-decrypts pending/retained rows, promotes the new key into the fixed sentinel row (keeping older app versions bootable), and keeps one superseded copy for rolling restarts. Secrets are not re-encrypted; only the root-key wrapper changes.

A legacy-key snapshot (`kms_legacy_encryption_keys`) decouples the pre-KMS tier (`project_bots`, `user_encryption_keys`, blind indexes, `org_bots`) from `process.env`, so env rotation does not strand those rows. Admin API routes and a Server Console panel expose status, generate/discard/complete flows, KEK fingerprint history, and straggler warnings via `lastResolvedAt`. A weekly cron GC retires abandoned pending rotations (30 days) and superseded keys after `KMS_ROOT_KEY_RETENTION_DAYS` (default 7), unless recent boot evidence says a straggler may still need the old key.

## Screenshots

image

image

image

image

## Steps to verify the change

1. Run migrations: `cd backend && npm run migration:latest-dev && npm run generate:schema`
2. Start the stack and open **Admin → Encryption** as a server admin. Confirm active fingerprint, empty pending/retained state, and history.
3. **Generate new key** — copy the one-time key and fingerprint. Verify `GET /api/v1/admin/encryption/status` shows `pendingRotation` and `blockers` includes `rotation-pending`.
4. **Discard** the pending key — confirm it disappears from status and the DB has no unactivated `kms_root_config` rows.
5. Generate again, set `ENCRYPTION_KEY` to the new value on one instance, and restart it. Confirm logs show promotion, status moves pending → retained, and secrets still decrypt.
6. Restart a second instance still on the old key — it should boot and log a superseded-key warning; `lastResolvedAt` should update on the retained row.
7. **Remove previous key** in the UI (or `POST /api/v1/admin/encryption/rotations/:id/complete`). If `lastResolvedAt` is within the last hour, confirm the acknowledgment prompt; with `acknowledged: true`, the retained row is deleted and history gets `retiredAt`.
8. Confirm `kms_legacy_encryption_keys` is seeded on boot and legacy-tier reads work after env rotation (`infisical.legacy_root_key.usage` metric increments on use).

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/lib/crypto/cryptography/crypto.ts
```diff
@@ -282,7 +283,9 @@ const cryptographyFactory = () => {
         data: string,
         appCfgOverride?: Pick<TEnvConfig, "ROOT_ENCRYPTION_KEY" | "ENCRYPTION_KEY">
       ) => {
-        const appCfg = appCfgOverride || getConfig();
+        // Snapshot over environment, so an ENCRYPTION_KEY rotation cannot strand these rows. It holds
+        // the same values, so ciphertext is unchanged. The fallback covers migrations and early boot.
+        const appCfg = getLegacyEncryptionSnapshot() || appCfgOverride || getConfig();
```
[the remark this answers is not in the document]
victorvhs017: The code that requires the legacy key will be removed after an observation period.

victorvhs017:
@greptileai re-review

backend/src/services/encryption-key-rotation/encryption-key-rotation-service.ts
```diff
@@ -0,0 +1,320 @@
...
+   * deletion that did not happen.
+   */
+  const $retireRetainedKey = async (tx: Knex, row: TKmsRootConfig) => {
+    const deleted = await kmsRootConfigDAL.deleteById(row.id, tx);
+    if (!deleted) return false;
+
+    if (row.kekFingerprint) {
+      const history = await kmsKekHistoryDAL.findHistory(tx);
+      const entry = history.find((e) => e.kekFingerprint === row.kekFingerprint && !e.retiredAt);
+      if (entry) await kmsKekHistoryDAL.updateById(entry.id, { retiredAt: new Date() }, tx);
```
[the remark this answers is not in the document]
victorvhs017: @greptileai A duplicate fingerprint can't occur. An operator cannot rotate "back to" a previous key, because they never choose the key.

victorvhs017:
@greptileai re-review

backend/src/server/routes/v1/admin-router.ts
```diff
@@ -637,6 +639,166 @@ export const registerAdminRouter = async (server: FastifyZodProvider) => {
...
+        })
+      }
+    },
+    onRequest: (req, res, done) => {
+      verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN])(req, res, () => {
+        verifySuperAdmin(req, res, done);
+      });
+    },
+    handler: async (req, res) => {
+      const rotation = await server.services.encryptionKeyRotation.createRotation({
```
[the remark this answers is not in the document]
victorvhs017: This will be done in another PR.

backend/src/@types/knex.d.ts
```diff
@@ -296,12 +296,18 @@ import {
   TKmipServers,
   TKmipServersInsert,
   TKmipServersUpdate,
+  TKmsKekHistory,
```
akhilmhdh, line 299: What does Kek means 🤔 ?

backend/src/lib/crypto/legacy-key/legacy-key.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+export type TLegacyKeyMaterial = {
+  // Post-FIPS-relabel, what the legacy tier resolves today. Writes always use these.
+  current: TLegacyKeySnapshot;
+  // Pre-relabel. Decrypt-only fallback: the relabel can drop a key existing rows were written under.
+  original?: TLegacyKeySnapshot;
+};
+
+let legacyKeyMaterial: TLegacyKeyMaterial | null = null;
```
akhilmhdh, line 18: This feels a bit weird pattern

backend/src/services/encryption-key-rotation/encryption-key-rotation-service.ts
```diff
@@ -0,0 +1,344 @@
...
+
+    const staged = await kmsRootConfigDAL.transaction(async (tx) => {
+      await tx.raw("SELECT pg_advisory_xact_lock(?)", [PgSqlLock.KmsRootKeyInit]);
+
+      // Read under the lock so the warning describes the key that will actually be removed on promotion.
+      const [retained] = await kmsRootConfigDAL.findRetained(tx);
+
+      const stillPending = await kmsRootConfigDAL.findPending(tx);
+      if (stillPending.length) {
+        if (!supersede) {
```
akhilmhdh: Took me quite sometime to understand supersede meant. May be `replacePending` would be more straight forward

backend/src/db/schemas/kms-kek-history.ts
```diff
@@ -0,0 +1,22 @@
+// Code generated by automation script, DO NOT EDIT.
+// Automated by pulling database and generating zod schema
+// To update. Just run npm run generate:schema
+// Written by akhilmhdh.
+
+import { z } from "zod";
+
+import { TImmutableDBKeys } from "./models";
+
+export const KmsKekHistorySchema = z.object({
```
akhilmhdh, line 10: I guess this is where we would need a system level audit log like GitLab

backend/src/server/routes/v1/admin-router.ts
```diff
@@ -637,6 +639,166 @@ export const registerAdminRouter = async (server: FastifyZodProvider) => {
     }
   });
 
+  server.route({
```
akhilmhdh, line 641: I think we need to think about the API design. I think something like this makes more easier to understand.

Give that there can only be three states i think we can omit the ids and follow the states

```
GET    /api/v1/admin/encryption/root-key              → { active, staged, expiring, canRotate }
POST   /api/v1/admin/encryption/root-key 
GET    /api/v1/admin/encryption/root-key/rotations  → history, paginated
```

For delete it can be explicit state
```
DELETE /api/v1/admin/encryption/root-key/pending
DELETE /api/v1/admin/encryption/root-key/previous
```

akhilmhdh: Supersede with query params in a POST feels weird. I think all parameters in a POST should be in a body

backend/src/lib/api-docs/constants.ts
```diff
@@ -3973,3 +3948,43 @@ export const SECRET_VALIDATION_RULES = {
...
+      "When an instance last started using the previous key. Recent means one is probably still running and would fail to restart if the key were removed. Null only means none has started since the rotation, not that none exists.",
+    history:
+      "Every key this instance has used, oldest last. Retained after a key is removed, so a restored backup can be matched to an archived key."
+  },
+  CREATE: {
+    description:
+      "Generate a new value for ENCRYPTION_KEY. The key is returned once and never stored, and nothing changes until an instance starts with it.",
+    supersede:
+      "Replace an already-generated key that has not been applied yet. The replaced key stops working immediately.",
+    fingerprint:
```
akhilmhdh: I think we can call it as label then. I feel like fingerprint may be harder to explain to folks

docs/self-hosting/guides/rotating-the-encryption-key.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+[`ENCRYPTION_KEY`](/self-hosting/configuration/envars#param-encryption-key) protects every secret in your
+instance. Rotating it periodically limits how much a leaked key is worth, and some compliance programs
+require it on a fixed schedule.
+
+Rotating is a two-part process by design: Infisical generates a key, and **the rotation only takes effect
+when an instance first starts with the new value**. Generating a key changes nothing, so you can generate
+one, think about it, and discard it with no consequences.
+
+<Note>
```
akhilmhdh, line 19: This feels a bit misleading.. First question in my mind would be `so is my secret safe?` and feels like rotation is not a good answer

docs/self-hosting/guides/rotating-the-encryption-key.mdx
```diff
@@ -0,0 +1,102 @@
...
+  Your secrets are never re-encrypted. Only a small internal wrapper changes, so a rotation is instant
+  regardless of how many secrets you store, and there is no read-only window.
+</Note>
+
+## Before you start
+
+- You need **server admin** access.
+- If your instance uses an [HSM](/self-hosting/guides/hsm-integration), there is no environment variable
+  to rotate. Rotate the key on the HSM instead.
+- Decide where you will keep the **old** key. See [After the rotation](#after-the-rotation).
```
akhilmhdh: The HSM part i think we can say it out as first section with a banner

docs/self-hosting/guides/rotating-the-encryption-key.mdx
```diff
@@ -0,0 +1,102 @@
...
+
+    ![Remove encryption key panel in the Server Console](/images/self-hosting/guides/rotating-encryption-key/remove-key.png)
+
+    <Note>
+      Rotating again before removing the previous key removes it for you, at the moment the new key is
+      first used. Infisical warns you about this when it generates the key. Only one previous key is ever
+      kept, so an instance two rotations behind will not be able to restart.
+    </Note>
+
+    <Note>
```
akhilmhdh: I think we can shorten this more like saying that our visibility is on restarted once and the server admin has to make sure all the instances are updated with new key

docs/self-hosting/guides/rotating-the-encryption-key.mdx
```diff
@@ -0,0 +1,102 @@
...
+      Infisical can only see instances that have *started* on the previous key. If one did so recently, it
+      asks you to confirm before removing the key, since you are the only one who can tell whether that
+      instance has since moved onto the new key. It cannot see an instance that has been running since
+      before the rotation and has not restarted. If such an instance restarts after the key is gone, it
+      fails to start with an error naming the key it needs; setting that key again brings it back.
+    </Note>
+  </Step>
+</Steps>
+
+## After the rotation
```
akhilmhdh, line 94: I was thinking what if we only show a delete banner and option to delete the key like after a day. This can be user configurable as well by an environment variable.

akhilmhdh (changes requested):

backend/src/@types/knex.d.ts
victorvhs017: Key Encryption Key! From KEK and DEK abbreviations.

backend/src/services/encryption-key-rotation/encryption-key-rotation-service.ts
victorvhs017: I couldn't find a better word and didn't want to make it too big. I'm renaming it!

backend/src/db/schemas/kms-kek-history.ts
victorvhs017: Yes! That's why I brought the discussion to the team. But it will probably be a follow-up PR.

backend/src/lib/crypto/legacy-key/legacy-key.ts
victorvhs017: It’s a boot-time in-memory cache, like ROOT_ENCRYPTION_KEY, but stored in a shared module because the crypto layer (which kms-service depends on, not the other way around) needs to read the legacy key snapshot without every caller passing keys through.

Here:
image

Also, this will be deleted in the future, so I went for convenience instead of a bigger refactoring.

backend/src/lib/api-docs/constants.ts
victorvhs017: Yep, I think "label" works!

backend/src/server/routes/v1/admin-router.ts
victorvhs017: - I agree on the supersede query param, updated that!
- history being paginated is a good catch
- I'm changing the status to root-key

---

For the POST, I think /rotations makes more sense because the POST isn't rotating anything. it stages a candidate key in kms_root_config table, and the rotation only actually happens when the user sets the environment variable. So what the POST creates is a rotation attempt, not a new root key, which is why /rotations reads better to me than /root-key.

On the canRotate, I think we can drop it and blockers both. It's just adding complexity to the response, and the user can derive that from encryptionStrategy and staged.

I think we should keep the IDs on the DELETE. Imagine the use case where we have more than one admin and one of them starts a new rotation while you're deleting. I agree that's a big edge case, and the rotation is not a routine operation, but we should have all the guardrails here.

Deleting by rotatioId makes things explicit. We could move it to label, actually, instead of rotation ID, which is something that the user has access to in the table.

---

What do you think about this option:

```
GET    /api/v1/admin/encryption/root-key
POST   /api/v1/admin/encryption/root-key/rotations
GET    /api/v1/admin/encryption/root-key/rotations?offset=&limit=
DELETE /api/v1/admin/encryption/root-key/staged?label=
DELETE /api/v1/admin/encryption/root-key/expiring?label=&force=

GET /encryption/root-key response:

{ rootKey: {
    encryptionStrategy: string | null,
    active:   { label: string | null, activatedAt: Date },
    staged:   { label: string, createdAt: Date } | null,
    expiring: { label: string | null, supersededAt: Date, lastResolvedAt: Date | null } | null
} }

```

I updated the code to this one.

docs/self-hosting/guides/rotating-the-encryption-key.mdx
victorvhs017: Updated the docs here.

docs/self-hosting/guides/rotating-the-encryption-key.mdx
victorvhs017: We have that yellow warning from the screenshots in the description, which also has a button.

I don't think it's worth making it configurable because it only shows up in the server console.

backend/src/lib/crypto/legacy-key/legacy-key.ts
akhilmhdh: Makes sense

backend/src/server/routes/v1/admin-router.ts
akhilmhdh: Wow! This looks much better. Good one. One change would be DELETE query param can be also a body.

It's recommended to use query params to only GET request. Most of the others ones can be just body itself.

backend/src/services/encryption-key-rotation/encryption-key-rotation-service.ts
```diff
@@ -0,0 +1,366 @@
...
+        });
+      }
+
+      logger.info(
+        `Encryption key rotation completed, previous key removed [retainedKeyId=${retained.id}] [label=${retained.kekLabel ?? "unknown"}]`
+      );
+    });
+  };
+
+  const runGarbageCollection = async () => {
```
akhilmhdh: We should rename this to be more specific like `removeInactiveKeys`

akhilmhdh (commented):
Just two mores and let me see if i can improve the docs wording a bit more. Doing one last sanity check.

backend/src/server/routes/v1/admin-router.ts
victorvhs017: Makes sense! Updating the code

akhilmhdh (approved):

PR #7759, opened 2026-08-21
feat: add RFC 8693 token exchange grant

Thiago-AS:
## Context

Middleware acting on behalf of users (MCP servers, AI agents, internal developer portals, API gateways) currently has to authenticate using a shared machine identity. This loses user attribution in the audit log and grants every user the full set of permissions assigned to that shared identity.

This change adds support for the RFC 8693 Token Exchange grant to the existing OAuth token endpoint. Trusted middleware can present a user's token issued by the organization's OIDC identity provider and receive a short-lived Infisical access token for that same user. The issued token carries the user's actual permissions and is fully attributed to them in the audit log. No refresh token is issued, and delegated tokens cannot access account self-management endpoints.

OAuth applications now explicitly declare their supported grant types. Existing applications and API clients that do not specify any grant types continue using the authorization code flow by default, so this change is fully backward compatible.

Token Exchange applications introduce one additional configuration field: the audience that the organization's identity provider issues for that middleware. This prevents tokens minted for unrelated applications from being exchanged.

The trust anchor is the organization's OIDC SSO configuration rather than per-application issuer settings. As a result, enabling Token Exchange, updating its audience, or rotating a Token Exchange application's client secret all require SSO edit permissions. Deleting a client, rotating its secret, or removing the Token Exchange grant immediately revokes any delegated tokens it has previously issued.

## Steps to verify the change

1. In the UI, create an OAuth application with the **Token Exchange** flow. Verify that the **Audience** field is required and that the redirect URI fields are hidden.
2. Verify that the SSO page displays the "Applications depend on this issuer" callout once a Token Exchange application exists.
3. Verify that successful exchanges appear in the organization audit log, attributed to the end user, with the acting application included in the metadata.
4. Run the end-to-end flow below.

### Testing end to end locally

A complete walkthrough is available in `sso.md` under **Testing OAuth Token Exchange (RFC 8693)**. The abbreviated flow is:

```bash
make up-dev-oidc     # Starts the default stack plus Keycloak with the seeded Infisical realm
make seed-dev-oidc   # Creates the oidc org, admin@oidc.com / password123!, verified domain, and active OIDC configuration
```

The seeded realm also includes a second client, `infisical-mcp`, which represents the middleware.

Sign in to http://localhost:8080/ as `admin@oidc.com`, navigate to **Organization Settings → OAuth Applications**, create an application using the **Token Exchange** flow, set the audience to `infisical-mcp`, and copy the generated client credentials.

Then run:

```bash
CLIENT_ID="<paste>"
CLIENT_SECRET="<paste>"

SUBJECT_TOKEN=$(curl -s -X POST \
  http://localhost:8088/realms/infisical/protocol/openid-connect/token \
  -d grant_type=password -d scope=openid \
  -d client_id=infisical-mcp -d client_secret=infisical-mcp-client-secret \
  -d username=admin@oidc.com -d password='password123!' \
  | python3 -c 'import sys,json; print(json.load(sys.stdin)["id_token"])')

curl -s -X POST http://localhost:8080/api/v1/oauth/token \
  -u "$CLIENT_ID:$CLIENT_SECRET" \
  -d grant_type=urn:ietf:params:oauth:grant-type:token-exchange \
  -d subject_token="$SUBJECT_TOKEN" \
  -d subject_token_type=urn:ietf:params:oauth:token-type:id_token \
  | python3 -m json.tool
```

Use the `id_token`, not the access token. The seeded Keycloak realm does not include an audience mapper, so its access tokens do not contain an `aud` claim.

Finally, validate the issued token:

* `GET /api/v1/oauth/validate` returns `200`.
* `GET /api/v1/user/me/totp` returns `403`, confirming delegated tokens cannot access account management endpoints.
* Reading a secret in the `oidc` organization succeeds.

`sso.md` also documents the recommended negative test cases, including cross-application replay, invalid signatures, supplying a `scope`, SSO disabled, MFA required, unregistered grants, and the alias requirement for users who have never signed in through SSO.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/oauth-client/oauth-client-service.ts
```diff
@@ -359,10 +581,188 @@ export const oauthClientServiceFactory = ({
...
+      },
+      event: {
+        type: EventType.OAUTH_CLIENT_TOKEN_EXCHANGE,
+        metadata: {
+          clientDbId: client.id,
+          clientId: client.clientId,
+          clientName: client.name,
+          subjectUserId: user.id,
+          subjectExternalId: subject,
+          tokenVersionId: tokenSession.id
```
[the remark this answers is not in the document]
Thiago-AS: We already have this information on the actor field.

frontend/src/pages/organization/SettingsPage/components/OrgOauthClientsTab/OauthClientModal.tsx
```diff
@@ -81,60 +134,78 @@ export const OauthClientModal = ({
...
       if (isEditing && editingClient) {
         await updateOauthClient({
+          ...payload,
           clientDbId: editingClient.id,
-          name,
-          description: description || null,
-          redirectUris: parsedRedirectUris,
-          requirePkce
+          description: form.description || null,
+          tokenExchangeAudience: isTokenExchange ? form.tokenExchangeAudience : null
```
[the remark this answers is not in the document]
Thiago-AS: Thats the correct behavior.

frontend/src/pages/organization/SettingsPage/components/OrgOauthClientsTab/OauthClientModal.tsx
```diff
@@ -144,121 +266,292 @@ export const OauthClientModal = ({
...
+                          className="data-[orientation=horizontal]:group-data-[style=underline]/tabs-list:flex-1"
+                        >
+                          Authorization code
+                        </TabsTrigger>
+                        <TabsTrigger
+                          value={OauthClientFlow.TokenExchange}
+                          className="data-[orientation=horizontal]:group-data-[style=underline]/tabs-list:flex-1"
+                        >
+                          Token exchange
+                        </TabsTrigger>
```
scott-ray-wilson: because these aren't just sections of configuration but mutually exclusive options I don't think this should be tabs - I think it should either be choice cards or simply a select with the two options 

Image

frontend/src/pages/organization/SettingsPage/components/OrgOauthClientsTab/OauthClientModal.tsx
scott-ray-wilson, line 1: I think we should use an alert dialog here or disable close on click outside because a user might accidentally press outside and the modal closes 

Image

scott-ray-wilson: also a little alert warning triangle in the alert would be cute :3

scott-ray-wilson, line 1: In your steps to verify you mentioned:

> Verify that the SSO page displays the "Applications depend on this issuer" callout once a Token Exchange application exists.

but I don't see any?

Image

Image

scott-ray-wilson, line 1: we should make this have a confirmation input to delete (andrew created a re-usable component to make this easy!)

Image

backend/e2e-test/routes/v1/oauth-token-exchange.spec.ts
```diff
@@ -0,0 +1,614 @@
...
+  test("a TTL longer than the instance lifetime is capped, not honoured", async () => {
+    const { client, token } = await issueAccessToken(90 * 24 * 60 * 60);
+
+    expect(client.accessTokenTTL).toBe(90 * 24 * 60 * 60);
+    expect(token.statusCode).toBe(200);
+    const { expires_in: expiresIn } = JSON.parse(token.payload) as { expires_in: number };
+    expect(expiresIn).toBeLessThan(client.accessTokenTTL);
+
+    await deleteClient(client.id);
+  });
```
scott-ray-wilson: should this be an error rather than silently shortening the ttl?

backend/e2e-test/routes/v1/oauth-token-exchange.spec.ts
scott-ray-wilson, line 1: ignore file placement:

a revoked delegated token surfaces as **404** `{"error": "Session not found"}` on every endpoint, including `GET /api/v1/oauth/validate`, because revocation deletes the session row and the shared `fnValidateJwtIdentity` throws `NotFoundError` for a missing session. That's uniform with regular user JWTs, but `/oauth/validate` is the endpoint middleware will code against to decide "re-exchange now", and a 404 there is easy to misread as a wrong URL rather than a revoked token (I did exactly that while testing). Would it be worth having validate map the dead-session case to 401, or a 200 `{"active": false}` in the RFC 7662 style? If keeping the auth path uniform is the priority, a line in the token-exchange docs telling middleware to treat any 4xx from validate as "re-exchange" would cover it.

backend/src/server/routes/v1/secret-sharing-router.ts
scott-ray-wilson, line 1: I wonder if we shouldn't add this here; reason being is this isn' permission based - it's org membership exclusively so there's no real intention behind giving it access or not - we should probably add an actual share secret permission to the org scope but I wonder if until then we should remove oauth support here?

backend/src/services/oauth-client/oauth-client-service.ts
```diff
@@ -332,37 +574,236 @@ export const oauthClientServiceFactory = ({
...
+      throw new OauthTokenError({
+        code: OauthTokenErrorCode.ServerError,
+        message: "Failed to create a session for the user this token identifies."
+      });
+
+    // A revocation can land between authenticating the client and inserting the session above, deleting
+    // nothing because the session did not exist yet. Every withdrawal path commits its client write
+    // before deleting sessions, so re-reading the client here catches it -- on the primary, since the
+    // replica can still be serving the pre-revocation row.
+    const currentClient = await oauthClientDAL.findByIdOnPrimary(client.id);
```
scott-ray-wilson: via local claude review:

The comment's premise is false. It says "every withdrawal path commits its client write before deleting sessions," but all three withdrawal paths run the client write and `revokeSessionsByUserAgent` inside one `oauthClientDAL.transaction` (`updateOauthClient:380`, `deleteOauthClient:411`, `rotateOauthClientSecret:436`), so the write and the sweep commit atomically, never write-first. That reopens the narrow race the re-read is meant to close:
>
> 1. The exchange inserts its session at line 720 (own connection, auto-commits).
> 2. A concurrent withdrawal transaction has already executed its `DELETE ... WHERE userAgent = <tag>` (which did not see that session, since the insert committed after the DELETE's scan) but has not yet COMMITted its client UPDATE.
> 3. The exchange's `findByIdOnPrimary` at 735 is a non-locking SELECT, so under READ COMMITTED it reads the still-uncommitted old client row; `hasClientAuthorityChanged` is false and the token is signed at 747.
> 4. The withdrawal COMMITs. Its DELETE already ran, so the exchange's session survives and the token stays valid up to `accessTokenTTL`.
>
> `origin/main`'s delete path committed the client removal before sweeping, which does not have this gap; wrapping the write and sweep in one transaction introduced it. Fix: commit the client write in its own transaction, then sweep sessions after commit (restores the write-first ordering the comment already assumes), or make the re-read a locking read (`FOR UPDATE`) so it blocks on the uncommitted withdrawal and then sees the new row. Separately, the in-transaction sweep is an unbounded tenant-sized DELETE, which the write-first-then-chunked-sweep shape also resolves.

backend/src/services/oauth-client/oauth-token-error.ts
```diff
@@ -0,0 +1,106 @@
...
+// `UnauthorizedError` and `ForbiddenRequestError` both mean a grant we will not act on (a bad
+// authorization code, refresh token or subject token, or a subject we cannot resolve to a usable
+// account), so they take whichever code that grant owes. Both are needed because the exchange refuses an
+// unusable subject in two vocabularies: its own membership checks raise the first, while
+// `getOrgPermission` raises the second for the same kind of refusal, and mapping only one would answer
+// half of them with an unexplained 500. Client authentication throws `invalid_client` explicitly instead,
+// because there it is the client that failed rather than the token. Anything else is a bug or an outage,
+// so it becomes a generic `server_error`: its message was written for a different envelope and could
+// carry internals.
+export const toOauthTokenError = (error: unknown, grantType?: OauthGrantType): OauthTokenError => {
```
scott-ray-wilson, line 90: `jsonwebtoken`'s `TokenExpiredError`/`JsonWebTokenError` from `validateRefreshToken` fall through to `server_error` (HTTP 500), so an expired or malformed `refresh_token` returns 500 instead of RFC 6749's 400 `invalid_grant`.

backend/src/services/oauth-client/oauth-client-service.ts
```diff
@@ -206,22 +423,51 @@ export const oauthClientServiceFactory = ({
...
 
     return { client: sanitizeOauthClient(updatedClient), clientSecret };
   };
 
+  const assertGrantEnabled = (client: TOauthClients, grantType: OauthGrantType) => {
+    if (!getGrantTypes(client).includes(grantType)) {
+      throw new OauthTokenError({
+        code: OauthTokenErrorCode.UnauthorizedClient,
+        message: `This application is not registered for the '${grantType}' grant type.`
+      });
```
scott-ray-wilson: do we need to include `authorize/info` and `authorize/consent` in `isOauthTokenRoute,`? only `/token` is schema-exempt so they wouldn't get the rfc expected error response

backend/src/server/routes/v1/oauth-router.ts
```diff
@@ -369,55 +439,141 @@ export const registerOAuthRouter = async (server: FastifyZodProvider) => {
...
+        const grantType = (req.body as { grant_type?: unknown } | undefined)?.grant_type;
+        const isKnownGrantType =
+          typeof grantType === "string" && Object.values(OauthGrantType).includes(grantType as OauthGrantType);
+
+        throw new OauthTokenError({
+          code: isKnownGrantType ? OauthTokenErrorCode.InvalidRequest : OauthTokenErrorCode.UnsupportedGrantType,
+          message: isKnownGrantType
+            ? describeValidationError(req.validationError)
+            : `Unsupported 'grant_type'. Supported values are: ${Object.values(OauthGrantType).join(", ")}`
+        });
```
scott-ray-wilson, line 511: claude said: Missing `grant_type` returns `unsupported_grant_type` - Use `invalid_request` for a missing parameter.

backend/src/server/routes/v1/oauth-router.ts
```diff
@@ -369,55 +439,141 @@ export const registerOAuthRouter = async (server: FastifyZodProvider) => {
     config: {
       rateLimit: authRateLimit
     },
+    attachValidation: true,
```
scott-ray-wilson, line 449: not sure if pressing but parser errors bypass `attachValidation` and escape the RFC contract.

backend/src/services/oauth-client/oauth-client-service.ts
```diff
@@ -332,37 +574,236 @@ export const oauthClientServiceFactory = ({
...
+    if (!tokenSession)
+      throw new OauthTokenError({
+        code: OauthTokenErrorCode.ServerError,
+        message: "Failed to create a session for the user this token identifies."
+      });
+
+    // A revocation can land between authenticating the client and inserting the session above, deleting
+    // nothing because the session did not exist yet. Every withdrawal path commits its client write
+    // before deleting sessions, so re-reading the client here catches it -- on the primary, since the
+    // replica can still be serving the pre-revocation row.
```
scott-ray-wilson: from claude:

The re-read guard here rests on a premise that is false. The comment above says "every withdrawal path commits its client write before deleting sessions," but all three withdrawal methods run the client write and `revokeSessionsByUserAgent` inside one transaction (`updateOauthClient:380`, `deleteOauthClient:411`, `rotateOauthClientSecret:436`), so the write and the sweep commit atomically, never write-first.
>
> That leaves a cross-transaction race. The exchange inserts its session at line 720 (its own connection, auto-committed) and then re-reads the client here at 735 (a plain SELECT, no transaction). If a concurrent withdrawal has already run its `DELETE ... WHERE userAgent = <tag>` (which did not see this session, since the insert committed after the DELETE's scan) but has not yet committed its client UPDATE, then under READ COMMITTED this re-read returns the still-uncommitted old client row, `hasClientAuthorityChanged` is false, and the token is signed at 747. The withdrawal then commits; its DELETE already ran, so this session survives and the token stays valid up to `accessTokenTTL`.
>
> Fix: commit the client write in its own transaction first, then sweep sessions after commit (this restores the write-first ordering the comment already assumes), or make this re-read a locking `SELECT ... FOR UPDATE` so it blocks on the uncommitted withdrawal and then observes the new row. The first option also lets the sweep move out of the withdrawal transaction, where it is currently an unbounded tenant-sized DELETE.

frontend/src/pages/organization/SettingsPage/components/OrgOauthClientsTab/OauthClientModal.tsx
```diff
@@ -40,101 +60,203 @@ const isValidRedirectUri = (uri: string) => {
...
+export const OauthClientModal = ({ popUp, handlePopUpClose, onCreated }: Props) => {
+  const isSheetOpen = Boolean(popUp?.clientForm?.isOpen);
+  const [editingClient, setEditingClient] = useState<TOauthClient | undefined>();
   const isEditing = Boolean(editingClient);
 
+  const { currentOrg } = useOrganization();
+  const { data: oidcConfig, isPending: isOidcConfigPending } = useGetOIDCConfig(
+    currentOrg?.id ?? ""
+  );
+  const hasActiveOidcSso = Boolean(oidcConfig?.isActive);
```
scott-ray-wilson, line 191: nit: this incorrectly shows users without permission that they don't have sso configured when it's actually they don't have the permission to read the config

backend/src/services/oauth-client/oauth-client-types.ts
```diff
@@ -2,19 +2,63 @@ import { z } from "zod";
...
+// Marks a token as carrying the user's authorization unnarrowed, unlike the consented scopes an
+// authorization-code token carries. A positive marker on purpose: absence has to keep meaning zero
+// permissions, so a dropped claim fails closed.
+export enum OauthDelegationMode {
+  Full = "full"
+}
+
+export const DEFAULT_OAUTH_ACCESS_TOKEN_TTL_SECONDS = 86400;
+export const MIN_OAUTH_ACCESS_TOKEN_TTL_SECONDS = 60;
+export const MAX_OAUTH_ACCESS_TOKEN_TTL_SECONDS = 90 * 24 * 60 * 60;
```
scott-ray-wilson, line 39: total nit but not sure we should have constants in a types file

scott-ray-wilson: there's already schemas below :dead

docs/documentation/platform/oauth-applications/authorization-code.mdx
scott-ray-wilson, line 1: should we add photos here? 

Image

scott-ray-wilson (changes requested):

frontend/src/pages/organization/SettingsPage/components/OrgOauthClientsTab/OauthClientModal.tsx
Thiago-AS: I ended up removing that, couldn't really find a nice way to display something in that screen (forgot to remove from the PR descp)

Thiago-AS: What do you think?

backend/src/server/routes/v1/secret-sharing-router.ts
Thiago-AS: Yeah, there are a couple of places like that. It’s more related to the loosened permission than to OAuth. I suggest we keep because otherwise we kill the token exchange behavior, I already restricted for administration write operations, so the rest shouldn't be a worry. WDYT?

docs/documentation/platform/oauth-applications/authorization-code.mdx
Thiago-AS: I was avoiding images until we have a clear guideline from Andrian on how to do so, those images get outdated REALLY fast, and do more harm than good.

backend/src/services/oauth-client/oauth-client-types.ts
Thiago-AS: Yeah, we’re not really following that very rigorously, lol.

frontend/src/pages/organization/SettingsPage/components/OrgOauthClientsTab/OauthClientModal.tsx
scott-ray-wilson: I think it would be a good idea, even just an info alert in or above the card would suffice

Thiago-AS: Working on it!

backend/src/server/routes/v1/secret-sharing-router.ts
scott-ray-wilson: I would think we should default to not adding oauth in these case since it's ambiguous - better to assume no access than full when it can't be derived - I doubt people would be using oauth for these end points anyway and they should be quick adds to patch if requested where we can properly handle them?

scott-ray-wilson (approved):
LGTM

PR #7762, opened 2026-08-21
feat(permissions): folder rbac

adilsitos:
## Context
This introduces a new permission into the system, the RBAC for folders. More info on the design doc.

## Screenshots

image

image

image

## Steps to verify the change

There are several scenarios, so my recommendation is to create a folder with: secrets, dynamic secrets, rotations, imports and (if you want) honey tokens. Use an account which is project admin and another account that you can keep changing their permissions, so it is easier to validate how the permissions are being evaluated. 

### Scenario 1: The user has folder RBAC permission in two different folders
- expected: the permissions from one folder don't modify the permissions of another folder

### Scenario 2: Check the cache is working correctly
- To do this it is recommended to enable the debug mode on knex, so you sure the cache is happening correctly (you can add `DEBUG=knex:query` on the .env)
- expected: if the marker is still up (15 seconds) the cache should responde directly. If the marker is not valid, we need to just compute the fingerprint 

### Scenario 3: Delete folders with RBAC as a member user 
- Define your test user to have the role member
- Define permissions to a folder to a specific user (it can be any user)
- expected: the member user can remove the folder 

### Scenario 4: Move folders with RBAC as a member user 
- Define your test user to have the role member
- Define permissions to a folder to a specific user (it can be any user)
- expected: the member user can move the folder

### Scenario 5: Project admins don't show on the rbac sheet
- expected: when opening the RBAC sheet, make sure that project admins (identities or users) don't appear there.

### Scenario 6: Check all the defined permissions defined 
- Keep changing the permission using the project admin to validate that the targeted user is getting the desired permissions (this is the most exhaustive one)

### Scenario 7: define a temporary full access. 
- try to define a user with full access role using temporary permissions
- expected: fail. A user with temporary full access can create definitive permissions for himself. This would be a vulnerability  

### Scenario 8: user has folder RBAC permission and become project admin 
- expected:  the folder RBAC permissions are not evaluated, he can see everything that a project admin can

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/folder-permission/folder-permission-service.ts
```diff
@@ -0,0 +1,204 @@
...
+      const dbError = error instanceof DatabaseError ? (error.error as { code?: string }) : null;
+      if (dbError?.code === DatabaseErrorCode.UniqueViolation) throw alreadyExistsError();
+      throw error;
+    }
+  };
+
+  const updateFolderGrant = async (dto: TUpdateFolderGrantDTO) => {
+    const { projectId, environmentSlug, secretPath, target } = dto;
+    const folder = await resolveFolder(projectId, environmentSlug, secretPath, secretFolderDAL);
+    await assertManageFolderAccess(dto.permission, projectId, folder, permissionService);
```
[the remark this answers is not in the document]
adilsitos: @veria-ai makes sense. I will send a patch that does not allow the Full access to be temporary, only permanent

backend/src/services/folder-permission/folder-permission-fns.ts
```diff
@@ -0,0 +1,184 @@
...
+// Access can be inherited from a group, so this cannot be a membership lookup on
+// actorUserId/actorIdentityId: a group-derived actor has no such row and would be rejected despite
+// holding real access to the project.
+const assertTargetMembership = async (
+  orgId: string,
+  projectId: string,
+  target: TFolderGrantActor,
+  folderPermissionDAL: Pick<TFolderPermissionDALFactory, "hasProjectAccess">
+) => {
+  const hasAccess = await folderPermissionDAL.hasProjectAccess({
```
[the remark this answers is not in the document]
adilsitos: veria-ai this was fixed

backend/src/services/folder-permission/folder-permission-service.ts
```diff
@@ -0,0 +1,213 @@
...
+  additionalPrivilegeDAL,
+  secretFolderDAL,
+  permissionService,
+  folderPermissionDAL
+}: TFolderPermissionServiceFactoryDep) => {
+  const createFolderGrant = async (dto: TCreateFolderGrantDTO) => {
+    const { projectId, environmentSlug, secretPath, target } = dto;
+    assertFullAccessIsPermanent(dto.role, Boolean(dto.type?.isTemporary));
+
+    const folder = await resolveFolder(projectId, environmentSlug, secretPath, secretFolderDAL);
```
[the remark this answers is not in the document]
adilsitos: veria-ai this was fixed

backend/src/services/folder-permission/folder-permission-service.ts
```diff
@@ -0,0 +1,255 @@
...
+        folderRBACAccess: folderAccess ? toFolderGrant(folderAccess, projectId, folder) : null
+      })),
+      totalCount
+    };
+  };
+
+  const listActorFolderGrants = async (dto: TListActorFolderGrantsDTO) => {
+    const { projectId, target } = dto;
+    const callerPermission = await getFolderAccessPermission(dto.permission, projectId, permissionService);
+    assertReadActorGrantsAccess(callerPermission, target);
```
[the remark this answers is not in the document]
adilsitos: the other permissions use the `Member.read` and they can see it, so we should not have a problem here. Let me know what you think @akhilmhdh

backend/src/services/folder-permission/folder-permission-fns.ts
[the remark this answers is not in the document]
adilsitos: this was fixed, can you check again?

adilsitos:
@greptile review it again and update the summary.

backend/src/services/folder-permission/folder-permission-fns.ts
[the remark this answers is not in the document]
adilsitos: this is just not covering cases where the user is not added as pending true. This is not happening anymore when the account already exists, if you think it is important to also check this use case, let me know @akhilmhdh

backend/src/ee/services/permission/folder-roles.ts
```diff
@@ -0,0 +1,228 @@
...
+export const SECRET_FOLDER_ROLE_PERMISSIONS: Record<SecretFolderRole, RawRuleOf<MongoAbility<ProjectPermissionSet>>[]> =
+  {
+    [SecretFolderRole.List]: folderListPermissions,
+    [SecretFolderRole.Read]: folderReadPermissions,
+    [SecretFolderRole.Edit]: folderEditPermissions,
+    [SecretFolderRole.Manage]: folderManagePermissions,
+    [SecretFolderRole.FullAccess]: folderFullAccessPermissions
+  };
+
+const buildFolderScopedDenyRules = () => {
```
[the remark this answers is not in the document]
adilsitos: we discussed this and it is fine to not have these ones now. cc: @akhilmhdh 

edit: actually, we discussed about the ProjectFolderGrant, sending a fix to deny the commit operations.

backend/src/services/folder-permission/folder-permission-service.ts
```diff
@@ -0,0 +1,273 @@
...
+
+    return { folderAccess: toFolderGrant(grant, projectId, folder) };
+  };
+
+  const deleteFolderGrant = async (dto: TDeleteFolderGrantDTO) => {
+    const { projectId, environmentSlug, secretPath, target } = dto;
+    await assertFolderRbacLicensed(dto.permission.orgId);
+    const callerPermission = await getFolderAccessPermission(dto.permission, projectId, permissionService);
+    const folder = await resolveFolder(projectId, environmentSlug, secretPath, secretFolderDAL);
+    assertManageFolderAccess(callerPermission, folder);
```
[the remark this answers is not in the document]
varonix0: We generally need better default roles yes, but I think blocking self-revocation is limiting users more than necessary

backend/src/ee/services/permission/permission-service.ts
```diff
@@ -465,6 +490,54 @@ export const permissionServiceFactory = ({
...
+        expiry: FOLDER_PERMISSION_VERSION_TTL
+      });
+    };
+
+  const $getFolderScopedPrivileges = async (
+    projectId: string,
+    actor: ActorType.USER | ActorType.IDENTITY,
+    actorId: string
+  ): Promise<TCachedFolderScopedPrivileges> =>
+    requestMemoize(requestMemoKeys.folderScopedPrivileges({ projectId, actor, actorId }), () =>
```
akhilmhdh: One note: Does this create key if you don't have any folder permission? Also can't we join together this as one key with permission layer because this would be evaluated with permission together always, so doesn't make sense to be in two keys.

backend/src/server/routes/v1/identity-project-membership-router.ts
```diff
@@ -561,4 +595,225 @@ export const registerIdentityProjectMembershipRouter = async (server: FastifyZod
...
+          identities: SanitizedFolderAccessIdentitySchema.array().describe(FOLDER_ACCESS.LIST_IDENTITIES.identities),
+          identitiesWithoutAccess: SanitizedFolderAccessIdentitySchema.array().describe(
+            FOLDER_ACCESS.LIST_IDENTITIES.identitiesWithoutAccess
+          ),
+          totalCount: z.number().describe(FOLDER_ACCESS.LIST_IDENTITIES.totalCount),
+          totalCountWithoutAccess: z.number().describe(FOLDER_ACCESS.LIST_IDENTITIES.totalCountWithoutAccess)
+        })
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
```
akhilmhdh: Question: I see we haven't exposed the user endpoint but the identity ones have been exposed. I think it should be both not exposed or exposed.

backend/src/services/folder-permission/folder-permission-dal.ts
```diff
@@ -0,0 +1,405 @@
...
+  // Memberships are tracked by createdAt, not updatedAt: every login bumps Membership.updatedAt
+  // (lastLoginTime), which would bust the roster on any member's sign-in.
+  const getFolderAccessRosterFingerprint = async (
+    { projectId, orgId, actorType }: TProjectScope & { actorType: TFolderActorType },
+    tx?: Knex
+  ) => {
+    try {
+      const conn = tx || db.replicaNode();
+      const { groupTable } = $actorColumns(actorType);
+      const countAndMax = (column: string) => conn.raw(`count(*) || ':' || coalesce(max(??)::text, '')`, [column]);
```
akhilmhdh: NIT: Wow! This is quite hard to understand. But makes sense in complexity - did you check the index usage of this?

backend/src/services/folder-permission/folder-permission-service.ts
```diff
@@ -0,0 +1,395 @@
...
+  permissionService: Pick<TPermissionServiceFactory, "getProjectPermission" | "invalidateProjectFolderPermissionCache">;
+  folderPermissionDAL: Pick<
+    TFolderPermissionDALFactory,
+    | "findProjectUserRoster"
+    | "findProjectIdentityRoster"
+    | "getFolderAccessRosterFingerprint"
+    | "hasProjectAccess"
+    | "isProjectAdmin"
+  >;
+  licenseService: Pick<TLicenseServiceFactory, "getPlan">;
```
akhilmhdh, line 61: [Ignore comment placement]

This temporary access button feels weird to me with the other sheet components. Feels like misplaced

Image

backend/src/services/folder-permission/folder-permission-service.ts
```diff
@@ -0,0 +1,395 @@
...
+  keyStore: Pick<TKeyStoreFactory, "getItem" | "setItemWithExpiry">;
+};
+
+type TFolderActorType = ActorType.USER | ActorType.IDENTITY;
+
+type TProjectScope = { projectId: string; orgId: string };
+
```
akhilmhdh, line 68: Admin badge may be taking more attention that it should be 

Image

akhilmhdh (changes requested):
One small confusion in folder permission. Let's catch up on that.

backend/src/ee/services/permission/permission-service.ts
adilsitos: The problem of joining the two is that we want to have different cache policies, where the folder permission outlives the permission roles one. This was a design decision, we can discuss this if you want. 

Regarding creating the keys, an empty user is creating an empty key on redis, but the memory footprint of this should be small. Do you have a concern with this on scale? We can remove it, but in this case we will be hitting the db every time a request that does not have permission calls the endpoint. I think it is more of a tradeoff here we can discuss.

backend/src/server/routes/v1/identity-project-membership-router.ts
adilsitos: I checked the endpoints, and they are with `hide=false`, can you point the place you found this?

backend/src/services/folder-permission/folder-permission-dal.ts
adilsitos: Working on a resolution to make this simpler

backend/src/services/folder-permission/folder-permission-service.ts
adilsitos: we can discuss this. This was one of the latest night changes we had on the UI yesterday.

backend/src/services/folder-permission/folder-permission-service.ts
adilsitos: Do you think if we decrease its size it gets better?

backend/src/ee/services/permission/permission-fns.ts
```diff
@@ -438,6 +450,139 @@ const handlebarsClient = (() => {
...
+  actorId: string,
+  {
+    additionalPrivilegeDAL,
+    secretFolderDAL
+  }: {
+    additionalPrivilegeDAL: Pick<TAdditionalPrivilegeDALFactory, "findFolderScopedPrivileges" | "transaction">;
+    secretFolderDAL: Pick<TSecretFolderDALFactory, "findSecretPathByFolderIds">;
+  }
+): Promise<TCachedFolderScopedPrivileges> => {
+  return additionalPrivilegeDAL.transaction(async (tx) => {
```
akhilmhdh: I don't think we need a transaction here. We can just open it by normally. If you wanted to avoid read replica instead of a transaction you can just use the primary node in the DAL

backend/src/ee/services/permission/permission-fns.ts
```diff
@@ -438,6 +450,139 @@ const handlebarsClient = (() => {
...
+    const rows = await additionalPrivilegeDAL.findFolderScopedPrivileges({ projectId, actorId, actorType: actor }, tx);
+    if (!rows.length) return { privileges: [] };
+
```
akhilmhdh: [Ignore comment placement]

Update environment is missing invalidate cache

akhilmhdh (approved):

varonix0 (approved):

PR #7768, opened 2026-08-21
feat: allow creation of cert profiles from inside applications

sheensantoscapadngan:
## Context
Allow creation of cert profiles from within applications

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/pages/cert-manager/ApplicationDetailsByIDPage/components/ApplicationRequestsTab.tsx
+++ b/frontend/src/pages/cert-manager/ApplicationDetailsByIDPage/components/ApplicationRequestsTab.tsx
@@ -1,6 +1,7 @@
 import { useNavigate, useParams } from "@tanstack/react-router";
 
 import { CertificateRequestsSection } from "../../CertificateRequestsPage/components/CertificateRequestsSection";
+import { ApplicationTab } from "../application-tabs";
 
 type Props = {
   applicationId: string;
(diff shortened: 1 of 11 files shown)


frontend/src/pages/cert-manager/ApplicationDetailsByIDPage/components/ApplicationRequestsTab.tsx
```diff
@@ -1,6 +1,7 @@
 import { useNavigate, useParams } from "@tanstack/react-router";
```
carlosmonastyrski, line 1: Ignore comment placement

Should we also add it here on the application creation modal?

Image

sheensantoscapadngan: for this PR, I made it so that certificate profiles field is not shown during application creation if none exists yet. I think for people bootstrapping, we'd want to have them think about creating certificate profiles when they're already inside an application since the first concern would be "I first want a container for storing these X certificates" and then after that would be "now I need to issue certificates so its time to create a certificate profile"

sheensantoscapadngan: flow would generally be the same even for ones with existing certificate profiles. create the container first and then think about the actual definition of the certificate you want to put inside the container

carlosmonastyrski (approved):

PR #7769, opened 2026-08-21
feat(pam): dedicated folder views

lb-vn:
## Context

Clicking on a folder redirects you to its own dedicated page, reflected in the URL as well. Searching for accounts at the root level still works.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/const/routes.ts
+++ b/frontend/src/const/routes.ts
@@ -405,7 +405,11 @@ export const ROUTE_PATHS = Object.freeze({
     ),
     AccountsPage: setRoute(
       "/organizations/$orgId/pam/accounts",
-      "/_authenticate/_inject-org-details/_org-layout/organizations/$orgId/pam/_pam-layout/accounts"
+      "/_authenticate/_inject-org-details/_org-layout/organizations/$orgId/pam/_pam-layout/accounts/"
+    ),
+    FolderPage: setRoute(
+      "/organizations/$orgId/pam/accounts/$folderId",
+      "/_authenticate/_inject-org-details/_org-layout/organizations/$orgId/pam/_pam-layout/accounts/$folderId"
     ),
     TemplatesPage: setRoute(
       "/organizations/$orgId/pam/templates",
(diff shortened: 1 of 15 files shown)


frontend/src/const/routes.ts
```diff
@@ -405,7 +405,11 @@ export const ROUTE_PATHS = Object.freeze({
...
     AccountsPage: setRoute(
       "/organizations/$orgId/pam/accounts",
-      "/_authenticate/_inject-org-details/_org-layout/organizations/$orgId/pam/_pam-layout/accounts"
+      "/_authenticate/_inject-org-details/_org-layout/organizations/$orgId/pam/_pam-layout/accounts/"
+    ),
+    FolderPage: setRoute(
+      "/organizations/$orgId/pam/accounts/$folderId",
+      "/_authenticate/_inject-org-details/_org-layout/organizations/$orgId/pam/_pam-layout/accounts/$folderId"
     ),
     TemplatesPage: setRoute(
```
sheensantoscapadngan, line 414: IGNORE PLACEMENT

Now that we can actually go inside a folder, would it make sense for us to also have a more native way to display all of these? Could we not have an approvals tab, a permissions tab, etc. from this main view as well? these all seem pretty hidden now no?

Image

frontend/src/pages/pam/PamAccountsPage/components/AccountsBreadcrumb.tsx
```diff
@@ -0,0 +1,41 @@
...
+  BreadcrumbList,
+  BreadcrumbPage,
+  BreadcrumbSeparator
+} from "@app/components/v3";
+
+type Props = {
+  orgId: string;
+  // Omitted at the root, where the folder icon is the current location.
+  folderName?: string;
+};
```
sheensantoscapadngan, line 17: IGNORE PLACEMENT

when you're working from within a specific folder view, it would seem misleading that you're able to switch  a folder during creation no? 
Image

sheensantoscapadngan (approved):

PR #7772, opened 2026-08-22
fix: resolved dev server reloading

akhilmhdh:
## Context

In development sometimes the dev server just reloads the full page and this would happen when you visit each page. This was due to two things
1. Vite saves the cache in `.vite` inside node modules. When you do a docker compose rebuild, that cache is lost and the frontend cache links gets broken. This is now avoided by having a volume cache to hold it like a CDN
2. Vite would do a full reload when a new dependency is found. The crypto, some icons, and packages are found later stage by Vite, and this causes the page to reload. I have added those ones as optimizeDeps so the first scanner picks it up. The cold start is not high to be even noticed.

## Steps to verify the change

1. Remove the present dev image for frontend
2. Now do a docker compose up and go through pages
3. Do a rebuild or compose down and it should still work fine.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,13 @@ up-prod:
 down:
 	docker compose -f docker-compose.dev.yml down
 
+# Wipes the persisted Vite dep-prebundle cache. Reach for this when the dev server serves
+# stale or broken /node_modules/.vite/deps chunks; it re-optimizes on the next `make up-dev`.
+COMPOSE_PROJECT_NAME ?= $(notdir $(CURDIR))
+clear-frontend-cache:
+	docker compose -f docker-compose.dev.yml rm -sf frontend
+	docker volume rm -f $(COMPOSE_PROJECT_NAME)_frontend_vite_cache
+
 reviewable-ui:
 	cd frontend && \
 	npm run lint:fix && \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -16,6 +16,13 @@ up-prod:
 down:
 	docker compose -f docker-compose.dev.yml down
 
+# Wipes the persisted Vite dep-prebundle cache. Reach for this when the dev server serves
+# stale or broken /node_modules/.vite/deps chunks; it re-optimizes on the next `make up-dev`.
+COMPOSE_PROJECT_NAME ?= $(notdir $(CURDIR))
+clear-frontend-cache:
+	docker compose -f docker-compose.dev.yml rm -sf frontend
+	docker volume rm -f $(COMPOSE_PROJECT_NAME)_frontend_vite_cache
+
```
scott-ray-wilson, line 28: I think this breaks if the folder name is uppercase which I often do for worktrees based off tickets, ie PLATFOR-532 - because docker lowercases it

scott-ray-wilson (changes requested):

scott-ray-wilson (approved):

PR #7777, opened 2026-08-24
fix(identity-lockout): replace Redis SCAN with sorted-set index for list reads

victorvhs017:
## Context

Project membership and identity list endpoints call `getIdentityActiveLockoutAuthMethods` once per row to populate `activeLockoutAuthMethods`. The old layout stored one Redis string key per auth method/slug and discovered them with `getKeysByPattern` (SCAN). A `limit=1000` memberships list walked the entire keyspace up to 1000 times.

Lockout state now uses a **sorted-set index + per-slug item keys**:

- **Item key** (unchanged): `lockout:identity:{identityId}:{authMethod}:{slug}` — JSON `{ lockedOut, failedAttempts }` with native Redis TTL
- **Index key** (new): `lockout:identity:{identityId}` — ZSET of *locked* slugs only, scored by lockout deadline (`{authMethod}:{slug}` members)

List reads do one `ZRANGEBYSCORE` per identity (score ≥ now), bounded by active lockouts. Login reads/writes the known item key directly. Admin clears enumerate the index on the **primary** (`sortedSetMembersPrimary`) so replica lag cannot hide a lockout being removed.

Only `lockedOut: true` states are indexed; failure counters stay off the index so LDAP username sprays cannot inflate list reads. Writes atomically upsert item + index via a Lua script (`setIndexedItemWithExpiry`), prune expired index members, and extend (never shorten) the index TTL. LDAP login now validates username (max 255) and password (max 1024).

## Steps to verify the change

1. After triggering a lockout, confirm Redis layout:
   ```bash
   redis-cli GET "lockout:identity:<identityId>:ldap-auth:<slug>"
   redis-cli ZRANGEBYSCORE "lockout:identity:<identityId>" $(date +%s)000 +inf WITHSCORES
   ```
2. List project identity memberships (`GET /api/v1/projects/{projectId}/memberships/identities?limit=1000`) and confirm `activeLockoutAuthMethods` is correct. Monitor Redis — no `SCAN` on lockout paths; one `ZRANGEBYSCORE` per identity row.
3. Spray failed LDAP logins with varied usernames and confirm the index stays empty (`ZCARD lockout:identity:<identityId>` = 0) until a real lockout occurs.
4. Admin-clear lockouts for an auth method and confirm both item keys and index members are removed; Universal Auth lockouts on the same identity are left intact.
5. Confirm successful login clears the item key and index member together.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/keystore/keystore.ts
+++ b/backend/src/keystore/keystore.ts
@@ -163,8 +163,11 @@ export const KeyStorePrefixes = {
   LicenseUsageReconcileMarker: (orgId: string) => `license-usage-reconcile-${orgId}` as const,
   LicenseUsageLastReported: (orgId: string, featureKey: string) =>
     `license-usage-last-reported-${orgId}-${featureKey}` as const,
-  IdentityLockoutStateHash: (identityId: string) => `lockout:identity:${identityId}` as const,
-  IdentityLockoutStateField: (authMethod: string, slug: string) => `${authMethod}:${slug}` as const,
+  IdentityLockoutState: (identityId: string, authMethod: string, slug: string) =>
+    `lockout:identity:${identityId}:${authMethod}:${slug}` as const,
+  // Sorted set of the identity's *locked* auth methods, scored by when each lockout ends.
+  IdentityLockoutIndex: (identityId: string) => `lockout:identity:${identityId}` as const,
+  IdentityLockoutMember: (authMethod: string, slug: string) => `${authMethod}:${slug}` as const,
 
   TelemetryAggregatedEventStream: (event: string, bucketId: string) =>
     `telemetry-agg-stream:${event}:${bucketId}` as const,
@@ -305,8 +308,19 @@ export type TKeyStoreFactory = {
   hashGet: (key: string, field: string) => Promise<string | null>;
   hashGetAll: (key: string) => Promise<Record<string, string>>;
   hashGetAllPrimary: (key: string) => Promise<Record<string, string>>;
-  hashSetFieldWithMinExpiry: (key: string, field: string, value: string, expiryInSeconds: number) => Promise<void>;
-  hashDeleteFields: (key: string, fields: string[]) => Promise<number>;
+  // sorted-set indexed items: item key gets native TTL; optional index member is scored by the same
+  // deadline and pruned on write.
+  setIndexedItemWithExpiry: (arg: {
+    indexKey: string;
+    member: string;
+    itemKey: string;
+    value: string;
+    expiryInSeconds: number;
+    indexed: boolean;
+  }) => Promise<void>;
+  deleteIndexedItems: (arg: { indexKey: string; members: string[]; itemKeys: string[] }) => Promise<void>;
+  sortedSetRangeByScore: (key: string, min: string | number, max: string | number) => Promise<string[]>;
+  sortedSetMembersPrimary: (key: string) => Promise<string[]>;
   // pg
   pgIncrementBy: (key: string, dto: { incr?: number; expiry?: string; tx?: Knex }) => Promise<number>;
   pgGetIntItem: (key: string, prefix?: string) => Promise<number | undefined>;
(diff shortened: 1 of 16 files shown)


backend/src/services/identity/identity-fns.ts
```diff
@@ -1,29 +1,113 @@
...
+
+// Named after the hash key and field so the login paths serialise per lockout rather than per identity.
+export const identityLockoutLockKey = (identityId: string, authMethod: string, slug: string) =>
+  KeyStorePrefixes.IdentityLockoutLock(`${lockoutHashKey(identityId)}:${lockoutField(authMethod, slug)}`);
+
+export const getIdentityLockoutState = async (
+  { identityId, authMethod, slug }: { identityId: string; authMethod: string; slug: string },
+  keyStore: Pick<TKeyStoreFactory, "hashGet">
+): Promise<TIdentityLockoutState | undefined> => {
+  const raw = await keyStore.hashGet(lockoutHashKey(identityId), lockoutField(authMethod, slug));
```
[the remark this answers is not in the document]
victorvhs017: That's expected, the migration overhead is not worth it for such a short TTL

backend/src/services/identity/identity-fns.ts
```diff
@@ -1,29 +1,113 @@
...
+  if (!raw) return undefined;
```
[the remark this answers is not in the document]
victorvhs017: @greptileai That's expected, the migration overhead is not worth it for such a short TTL

backend/src/keystore/keystore.ts
```diff
@@ -532,6 +533,105 @@ export const keyStoreFactory = (
...
+        local take = math.min(total, batch)
+        local offset = 0
+        local nextPosition = tonumber(scanned[1])
+
+        local listpack = total > batch
+        if listpack then
+          -- A listpack-encoded hash ignores both the cursor and COUNT and returns every field, so
+          -- HSCAN cannot advance on its own and the window has to be tracked by hand.
+          offset = position % total
+        end
```
lobocv: The listpack stuff is a bit concerning from an edge-case perspective. We can't really trust this works without some integration tests. Is it possible to write some integration tests for this `HASH_SET_FIELD_WITH_MIN_EXPIRY` function, using small keys and large hashes to hopefully test the listpack behaviour?

backend/src/keystore/keystore.ts
```diff
@@ -532,6 +533,105 @@ export const keyStoreFactory = (
 
   const hashGet = async (key: string, field: string) => primaryRedis.hget(key, field);
 
+  const hashGetAll = async (key: string) => pickPrimaryOrSecondaryRedis(primaryRedis, redisReadReplicas).hgetall(key);
+
+  const hashGetAllPrimary = async (key: string) => primaryRedis.hgetall(key);
+
+  // A hash carries one TTL for the whole key, so the expiry can only ever grow: shortening it would
```
lobocv: I think there might be a simpler solution by using a redis sorted set. Each identity can keep all it's auth method lockouts in one sorted set, where the TTL is the sorting key.

Then, getting all the lockouts for an identity is simply reading the sorted set for the key (the key just is the identity).
On each write we could delete any keys in the sorted set that have expired. We could also do this on read, but I think it's safer on write, although it means we will be storing a bit more keys if there are no writes. That could be consumer side logic though... for the currently usecase (auth lockout) we can do it on write as it should be safe (we have lots of logins so it should keep the set pruned). In other cases we can decide to do it on read.

To get the lockout state for an individual (identity, authMethod) you just first read the sorted set and use it as an index to look up the redis key of the actual auth lockout. The auth lockout can have a TTL as well and it's okay if they are cleaned up separately as long as they timestamp they have it the same (which it should be because they'll be written together).

What do you think? I really like that solution because it's much easier to test and understand and doesn't have as many edge cases / complex scan/reap logic

victorvhs017: Great catch! This solves the complexity I had with the reap process, and because we don't need the count for the list endpoint, it works perfectly to list the locked items. The tradeoff is just adding more keys, but it should be fine, as the issue is the SCAN process itself, not the number of keys we have.

backend/src/keystore/keystore.ts
victorvhs017: outdated as we moved for sorted sets

victorvhs017:
@greptileai re-review

backend/src/keystore/keystore.ts
```diff
@@ -164,10 +164,10 @@ export const KeyStorePrefixes = {
...
     `license-usage-last-reported-${orgId}-${featureKey}` as const,
   IdentityLockoutState: (identityId: string, authMethod: string, slug: string) =>
-    `lockout:identity:${identityId}:${authMethod}:${slug}` as const,
-  IdentityLockoutStateByMethodPattern: (identityId: string, authMethod: string) =>
-    `lockout:identity:${identityId}:${authMethod}:*` as const,
-  IdentityLockoutStatePattern: (identityId: string) => `lockout:identity:${identityId}:*` as const,
+    `lockout:identity:{${identityId}}:${authMethod}:${slug}` as const,
+  // Sorted set of the identity's *locked* auth methods, scored by when each lockout ends.
+  IdentityLockoutIndex: (identityId: string) => `lockout:identity:{${identityId}}` as const,
+  IdentityLockoutMember: (authMethod: string, slug: string) => `${authMethod}:${slug}` as const,
```
lobocv, line 170: the keystore package is generic, right? It shouldn't know about lockout stuff. That's a consumer of keystore and builds on top of it. I think this should live in the identity-service.ts

but since it seems like it's already been like this, we can leave it. The bug is urgent and I dont want to hold the review up for it

backend/src/keystore/keystore.ts
```diff
@@ -532,6 +545,79 @@ export const keyStoreFactory = (
...
+    else
+      local current = redis.call('TTL', KEYS[1])
+      local wanted = tonumber(ARGV[3])
+      if current < 0 or current < wanted then
+        redis.call('EXPIRE', KEYS[1], wanted)
+      end
+    end
+
+    return 1
+  `;
```
lobocv, line 573: nit:
This is a bit confusing to me. Do we need to write this as a script? Can we not just expose the sorted set operations and move the composition of them into the lockout library code?  

I'm not sure what ARG[5] (index) means? This seems to be functionality specific to the lockout feature and I'm not sure it should be in a keystore library..

But this may just be how the code is structured and I don't want to force a large refactor either. My gut tells me that the package structure is wrong.

At the very least, can we clear up what this script is doing with some comments?

backend/src/keystore/keystore.ts
```diff
@@ -532,6 +545,79 @@ export const keyStoreFactory = (
...
+    if redis.call('ZCARD', KEYS[1]) == 0 then
+      redis.call('DEL', KEYS[1])
+    end
+    return 1
+  `;
+
+  const deleteIndexedItems: TKeyStoreFactory["deleteIndexedItems"] = async ({ indexKey, members, itemKeys }) => {
+    if (!members.length && !itemKeys.length) return;
+    await primaryRedis.eval(DELETE_INDEXED_ITEMS, itemKeys.length + 1, indexKey, ...itemKeys, ...members);
+  };
```
lobocv, line 614: nit: This too seems like application logic. The need to have a sorted set index and separate keys for actual values is something the lockout code needs, i'm not sure it's the right place to put it into the generic keystore lib

Not blocking, just an observation.

backend/src/services/identity/identity-fns.ts
```diff
@@ -1,29 +1,105 @@
 import { IdentityAuthMethod } from "@app/db/schemas";
 import { KeyStorePrefixes, TKeyStoreFactory } from "@app/keystore/keystore";
 
+export type TIdentityLockoutState = {
+  lockedOut: boolean;
+  failedAttempts: number;
+};
+
+const lockoutIndexKey = (identityId: string) => KeyStorePrefixes.IdentityLockoutIndex(identityId);
```
lobocv, line 9: KeyStorePrefixes.IdentityLockoutIndex should be defined in this package, not keystore. keystore shouldn't know anything about lockout ideally.

lobocv (approved):
Looks good @victorvhs017 !
I hope you don't mind. I left some comments just on observations of tech debt that I see. I'm not asking you to clean it up because it's probably a larger refactor, and I know this bug is pretty urgent. Just wanted to note some of these ideas down so we could reference them back in the future when we talk about rewriting some of this stuff and the new architecture.

PR #7784, opened 2026-08-24
feat: show renewed certificates as Renewed

carlosmonastyrski:
## Context

Renewal issues a replacement instead of modifying the original, so the inventory showed two Active certificates under the same name. Status is now four mutually exclusive buckets with precedence revoked > expired > renewed > active: a superseded certificate reports Renewed and drops out of the active and expiring-soon counts, while staying valid until its own expiry, since renewal deliberately does not revoke what it replaces. Also adds server-side validation of the status filter, a Renewed saved view, and one shared derived-status helper behind the table, details page, and CSV export.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/docs/documentation/platform/pki/concepts/certificate-lifecycle.mdx
+++ b/docs/documentation/platform/pki/concepts/certificate-lifecycle.mdx
@@ -33,19 +33,7 @@ Install issued certificates on the systems that need them — web servers, load 
 
 Replace a certificate before it expires to maintain trust and availability. Renewal can reuse the same key pair or rotate to a new one. In Infisical, renewal can be **server-driven** (Infisical auto-renews and syncs to external systems) or **client-driven** (an agent or workload initiates renewal when key material needs to stay under client control).
 
-### What happens to the original certificate
-
-Renewal issues a new certificate with a new serial number instead of changing the existing one, because the certificate already installed on your systems has to keep working until the replacement is deployed. Infisical links the two so the relationship is visible from either side:
-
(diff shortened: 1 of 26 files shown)


docs/documentation/platform/pki/concepts/certificate-lifecycle.mdx
```diff
@@ -33,6 +33,20 @@ Install issued certificates on the systems that need them — web servers, load
...
+- The original reports a status of **Renewed** and links to its replacement under **Renewed by**.
+- The replacement reports **Active** and links back under **Renewed from**.
+
+A renewed certificate no longer counts as active in the inventory or on the dashboard and is left out of expiry alerts, and any Certificate Sync attached to it moves to the replacement, so the replacement is what gets pushed to your destinations. It stays cryptographically valid until it expires, which is what gives you a window to roll the replacement out. If the original needs to stop being trusted before then, revoke it.
+
+<Note>
+  Renewal never revokes the original certificate. Any client still holding it continues to trust it
+  until it expires or you revoke it.
+</Note>
+
```
sheensantoscapadngan, line 37: I think this page should remain high-level. It feels off that we're only elaborating it for renewal, but all the others are pointing to a dedicated page altogether. Shouldn't we keep it consistent?

sheensantoscapadngan (approved):

PR #7788, opened 2026-08-24
fix: honor machine identity auth on org membership endpoints

Thiago-AS:
## Context

1. Machine identity requests were silently doing nothing. All three routes include IDENTITY_ACCESS_TOKEN in verifyAuth, but each handler started with if (req.auth.actor !== ActorType.USER) return;.

2. Unknown membership IDs returned 500 instead of 404. deleteOrgMembershipsFn returned [], then the route tried to read membership.status from undefined and threw a TypeError. Passing a user ID where a membership ID is expected is an obvious way to hit this.

3. Writes could affect non-user memberships. Identity and group org memberships share the memberships table. The read lookups filter by actorUserId, but the write lookups did not. For example, passing an identity's org membership ID to PATCH with a metadata body triggered identityMetadataDAL.delete({ userId: null, orgId }). Knex renders that as where "userId" is null, which deleted every machine identity's metadata in the org.

4. Adds a `delete-user-org-membership` audit event. There was no audit trail for user org
  membership removal at all, only a PostHog event. Emitted once per removed member, matching
  `REMOVE_PROJECT_MEMBER`.

## Steps to verify the change

1. In the UI, go to **Access Control > Machine Identities > Create**, and select the **Admin** role. Open **Universal Auth**, copy the **Client ID**, and add a **Client Secret**.

2. Get a token:

```bash
curl -s -X POST http://localhost:8080/api/v1/auth/universal-auth/login \
  -H 'Content-Type: application/json' \
  -d '{"clientId":"<id>","clientSecret":"<secret>"}' | jq -r .accessToken
```

3. Invite a throwaway user and get its membership ID from `GET /api/v2/organizations/$ORG/memberships`. Then run:

```bash
BASE=http://localhost:8080/api/v2/organizations/$ORG/memberships
A="Authorization: Bearer $TOKEN"

# 200 + membership body; the row is actually deleted
# (previously returned 200 + an empty body without deleting the row)
curl -i -X DELETE "$BASE/$MEMBERSHIP_ID" -H "$A"

# 404 with the requested ID
# (previously returned 500)
curl -i -X DELETE "$BASE/00000000-0000-4000-8000-000000000000" -H "$A"

# 200; isActive is set to false
curl -i -X PATCH "$BASE/$MEMBERSHIP_ID" -H "$A" \
  -H 'Content-Type: application/json' -d '{"isActive":false}'
```

4. Verify non-user membership rejection. Take an `id` from `GET /api/v2/organizations/$ORG/identity-memberships` (this `id` is the membership row ID) and confirm that both requests return **404**. Also confirm that `identity_metadata` still contains its rows:

```bash
curl -i -X DELETE "$BASE/$IDENTITY_MEMBERSHIP_ID" -H "$A"
curl -i -X PATCH "$BASE/$IDENTITY_MEMBERSHIP_ID" -H "$A" \
  -H 'Content-Type: application/json' -d '{"metadata":[{"key":"x","value":"y"}]}'
```

5. Check the audit log under **Organization > Audit Logs**. There should be one **Delete User Org Membership** entry for each removed member, with `identity` as the actor and metadata containing `membershipId` and a non-null `userId`. The rejected 404 requests should not produce any audit log entries.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/server/routes/v2/organization-router.ts
+++ b/backend/src/server/routes/v2/organization-router.ts
@@ -211,6 +211,21 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
         ...req.body
       });
 
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        orgId: req.params.organizationId,
+        event: {
+          type: EventType.UPDATE_USER_ORG_MEMBERSHIP,
+          metadata: {
+            membershipId: membership.id,
+            userId: membership.actorUserId as string,
+            role: req.body.role,
+            isActive: req.body.isActive,
+            metadataKeys: req.body.metadata?.map(({ key }) => key)
+          }
+        }
+      });
+
       if (req.body.role) {
         void server.services.telemetry.sendPostHogEvents({
           event: PostHogEventTypes.OrgMembershipRoleUpdated,
--- a/backend/src/server/routes/v2/organization-router.ts
+++ b/backend/src/server/routes/v2/organization-router.ts
@@ -1,6 +1,7 @@
 import { z } from "zod";
 
 import { OrgMembershipsSchema, OrgMembershipStatus, ProjectMembershipsSchema, ProjectsSchema } from "@app/db/schemas";
+import { EventType } from "@app/ee/services/audit-log/audit-log-types";
 import { ApiDocsTags, ORGANIZATIONS } from "@app/lib/api-docs";
 import { getConfig } from "@app/lib/config/env";
 import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
@@ -200,10 +201,9 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
     },
     onRequest: verifyAuth([AuthMode.JWT, AuthMode.IDENTITY_ACCESS_TOKEN]),
     handler: async (req) => {
-      if (req.auth.actor !== ActorType.USER) return;
-
       const membership = await server.services.org.updateOrgMembership({
-        userId: req.permission.id,
+        actor: req.permission.type,
+        actorId: req.permission.id,
         actorAuthMethod: req.permission.authMethod,
         orgId: req.params.organizationId,
         membershipId: req.params.membershipId,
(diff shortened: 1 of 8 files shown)


backend/src/server/routes/v2/organization-router.ts
```diff
@@ -262,16 +262,27 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
...
       });
 
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        orgId: req.params.organizationId,
+        event: {
+          type: EventType.DELETE_USER_ORG_MEMBERSHIP,
+          metadata: {
+            membershipId: membership.id,
+            userId: membership.actorUserId as string
```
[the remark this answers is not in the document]
Thiago-AS: That's only the metadata, the actor is still readable.

backend/src/server/routes/v2/organization-router.ts
```diff
@@ -322,16 +333,30 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
...
+        actorId: req.permission.id,
         actorAuthMethod: req.permission.authMethod,
         orgId: req.params.organizationId,
         membershipIds: req.body.membershipIds,
         actorOrgId: req.permission.orgId
       });
 
+      for (const membership of memberships) {
+        // eslint-disable-next-line no-await-in-loop
+        await server.services.auditLog.createAuditLog({
```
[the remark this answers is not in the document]
Thiago-AS: No need for now, following the code pattern we have on project membership deletion.

backend/src/services/org/org-service.ts
```diff
@@ -1180,29 +1181,38 @@ export const orgServiceFactory = ({
...
+      throw new NotFoundError({ message: `Organization membership with ID '${membershipId}' not found` });
+
     const [deletedMembership] = await deleteOrgMembershipsFn({
       orgMembershipIds: [membershipId],
       orgId,
       orgDAL,
       projectKeyDAL,
       userAliasDAL,
       licenseService,
-      userId,
```
victorvhs017, line 1205: Non-blocking: Not directly related to this PR, but flagging it here as it showed up in the tests; deleting org memberships here doesn't delete the `identity_metadata` row, leaving it orphaned.

backend/src/server/routes/v2/organization-router.ts
```diff
@@ -200,17 +201,31 @@ export const registerOrgRouter = async (server: FastifyZodProvider) => {
...
 
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        orgId: req.params.organizationId,
+        event: {
+          type: EventType.UPDATE_USER_ORG_MEMBERSHIP,
+          metadata: {
+            membershipId: membership.id,
+            userId: membership.actorUserId as string,
+            role: req.body.role,
```
victorvhs017, line 222: We should also send the roleID here, as the name can be modified in the future, making it hard to track.

victorvhs017 (changes requested):

backend/src/server/routes/v2/organization-router.ts
Thiago-AS: We use the role slug in every audit log, which is why I kept it that way. The only place we use the ID is when logging the role entity itself, it's best to keep it that way to maintain consistency across entities.

victorvhs017: That's fine, I'll create a follow-up so we can validate all audit logs to check if we are passing trackable data

victorvhs017 (approved):

backend/src/services/org/org-service.ts
Thiago-AS: I though this was intended, so if the user gets invited back it keeps the identity metadata 🤔

PR #7790, opened 2026-08-24
feat(pam): postgres account discovery

lb-vn:
## Context

Adds account discovery for postgres

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
+++ b/backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
@@ -8,7 +8,6 @@ import { logger } from "@app/lib/logger";
 
 import { PamAccountType, PamPostgresAuthMethod } from "../../pam/pam-enums";
 import { executeWithGateway, sweepReachableTargets } from "../pam-discovery-fns";
-import { expandTargets } from "../pam-discovery-targets";
 import {
   TDiscoveredAccount,
   TDiscoveryCredentialAccount,
@@ -20,7 +19,6 @@ import {
 const QUERY_TIMEOUT_MS = 20 * 1000;
 const SCAN_CONCURRENCY = 32;
 const SWEEP_DIAL_TIMEOUT_MS = 3 * 1000;
-const MAX_SWEEP_TARGETS = 65536;
 const MAX_ACCOUNTS_PER_HOST = 2000;
 const MAX_ACCOUNTS_PER_SCAN = 50000;
 
@@ -89,7 +87,7 @@ const toPostgresAccount = (account: TDiscoveryCredentialAccount): TPostgresAccou
   };
 };
 
-// an IAM token is minted per host/port/user pair, which a sweep across arbitrary targets cannot do, so only password accounts can drive a scan
+// an IAM token is minted per host/port/user, so it can't be reused across targets
 const isUsableAccount = (account: TPostgresAccount) => Boolean(account.password);
 
 export const postgresDiscoveryFactory: TPamDiscoveryFactory = ({
@@ -99,7 +97,7 @@ export const postgresDiscoveryFactory: TPamDiscoveryFactory = ({
   gatewayV2Service
 }) => {
   const accounts = credentialAccounts.map(toPostgresAccount);
-  const config = configuration as { cidrRanges: string[] };
+  const config = configuration as { hosts: string[] };
 
   const enumerateInstance = (host: string, port: number, account: TPostgresAccount) =>
     executeWithGateway(host, port, gatewayId, gatewayV2Service, async (proxyPort) => {
@@ -225,26 +223,26 @@ export const postgresDiscoveryFactory: TPamDiscoveryFactory = ({
       });
     }
 
-    const targets = expandTargets(config.cidrRanges);
+    const targets = [...new Set(config.hosts.map((host) => host.trim()).filter(Boolean))];
     const usablePorts = [...new Set(accounts.filter(isUsableAccount).map((a) => a.port))];
+    const candidatePairs = targets.flatMap((host) => usablePorts.map((port) => ({ host, port })));
 
-    const sweepTargets = targets.flatMap((host) => usablePorts.map((port) => ({ host, port })));
-    if (sweepTargets.length > MAX_SWEEP_TARGETS) {
-      throw new BadRequestError({
-        message: `Scan expands to ${sweepTargets.length} host-port combinations, exceeding the limit of ${MAX_SWEEP_TARGETS}. Reduce the target ranges or the number of distinct credential ports.`
-      });
-    }
-    const open = await sweepReachableTargets(sweepTargets, gatewayId, gatewayV2Service, SWEEP_DIAL_TIMEOUT_MS, signal);
+    const open = await sweepReachableTargets(
+      candidatePairs,
+      gatewayId,
+      gatewayV2Service,
+      SWEEP_DIAL_TIMEOUT_MS,
+      signal
+    );
 
-    // scan every reachable host:port pair, plus any pair a credential account names outright
-    const instancesToScan = sweepTargets.filter(
+    const instancesToScan = candidatePairs.filter(
       ({ host, port }) => open.has(`${host}:${port}`) || accounts.some((a) => a.host === host && a.port === port)
     );
 
     if (!instancesToScan.length) {
       throw new BadRequestError({
         message:
-          "No hosts were reachable on the credential ports in the target range. Check the targets, that the gateway can reach them, and that PostgreSQL is listening."
+          "None of the targets were reachable on the credential accounts' ports. Check the hosts, that the gateway can reach them, and that PostgreSQL is listening."
       });
     }
 
--- a/backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
+++ b/backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
@@ -112,6 +112,8 @@ export const postgresDiscoveryFactory: TPamDiscoveryFactory = ({
           user: account.username,
           password: account.password,
           connectionTimeoutMillis: QUERY_TIMEOUT_MS,
+          statement_timeout: QUERY_TIMEOUT_MS,
+          query_timeout: QUERY_TIMEOUT_MS,
           ssl: account.sslEnabled
             ? {
                 rejectUnauthorized: account.sslRejectUnauthorized,
--- a/backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
+++ b/backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
@@ -24,7 +24,35 @@ const MAX_SWEEP_TARGETS = 65536;
 const MAX_ACCOUNTS_PER_HOST = 2000;
 const MAX_ACCOUNTS_PER_SCAN = 50000;
 
-const ENUMERATION_QUERY = `SELECT rolname FROM pg_roles WHERE rolcanlogin ORDER BY rolname LIMIT ${MAX_ACCOUNTS_PER_HOST}`;
+// one row past the cap, so a capped instance is detectable and can be excluded from reconciliation
+const ENUMERATION_QUERY = `SELECT rolname FROM pg_roles WHERE rolcanlogin ORDER BY rolname LIMIT ${MAX_ACCOUNTS_PER_HOST + 1}`;
+
+// pg surfaces auth/TLS/connection failures with driver-level detail; map the ones a user can act on and keep the
+// raw error to the log, so the run's machine errors stay stable and free of database internals
+const describeConnectionError = (err: unknown): string => {
+  const code = (err as { code?: string })?.code;
+  switch (code) {
+    case "28P01":
+    case "28000":
+      return "Authentication failed for the credential account";
+    case "3D000":
+      return "The credential account's database does not exist on this instance";
+    case "53300":
+      return "The instance refused the connection because it is out of connection slots";
+    case "ECONNREFUSED":
+      return "Connection refused";
+    case "ETIMEDOUT":
+    case "ENETUNREACH":
+    case "EHOSTUNREACH":
+      return "Timed out connecting to the instance";
+    case "DEPTH_ZERO_SELF_SIGNED_CERT":
+    case "SELF_SIGNED_CERT_IN_CHAIN":
+    case "UNABLE_TO_VERIFY_LEAF_SIGNATURE":
+      return "TLS verification failed. Add the instance's CA certificate to the credential account, or disable certificate verification.";
+    default:
+      return "Could not enumerate roles on this instance";
+  }
+};
 
 const TRAILING_HYPHENS_REGEX = new RE2(/-+$/);
 
@@ -73,13 +101,8 @@ export const postgresDiscoveryFactory: TPamDiscoveryFactory = ({
   const accounts = credentialAccounts.map(toPostgresAccount);
   const config = configuration as { cidrRanges: string[] };
 
-  const orderAccountsForHost = (host: string) => [
-    ...accounts.filter((a) => a.host === host),
-    ...accounts.filter((a) => a.host !== host)
-  ];
-
-  const enumerateHost = (host: string, account: TPostgresAccount) =>
-    executeWithGateway(host, account.port, gatewayId, gatewayV2Service, async (proxyPort) => {
+  const enumerateInstance = (host: string, port: number, account: TPostgresAccount) =>
+    executeWithGateway(host, port, gatewayId, gatewayV2Service, async (proxyPort) => {
       const db = knex({
         client: "pg",
         connection: {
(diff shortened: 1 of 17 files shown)


lb-vn:
RE2 job failing is a false positive

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+const QUERY_TIMEOUT_MS = 20 * 1000;
+const SCAN_CONCURRENCY = 32;
+const SWEEP_DIAL_TIMEOUT_MS = 3 * 1000;
+const MAX_SWEEP_TARGETS = 65536;
+const MAX_ACCOUNTS_PER_HOST = 2000;
+const MAX_ACCOUNTS_PER_SCAN = 50000;
+
+// one row past the cap, so a capped instance is detectable and can be excluded from reconciliation
+const ENUMERATION_QUERY = `SELECT rolname FROM pg_roles WHERE rolcanlogin ORDER BY rolname LIMIT ${MAX_ACCOUNTS_PER_HOST + 1}`;
+
```
sheensantoscapadngan, line 27: We should be able to automatically fill up the target input with the hosts of the selected credential accounts, right? 
Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+      return "Timed out connecting to the instance";
+    case "DEPTH_ZERO_SELF_SIGNED_CERT":
+    case "SELF_SIGNED_CERT_IN_CHAIN":
+    case "UNABLE_TO_VERIFY_LEAF_SIGNATURE":
+      return "TLS verification failed. Add the instance's CA certificate to the credential account, or disable certificate verification.";
+    default:
+      return "Could not enumerate roles on this instance";
+  }
+};
+
```
sheensantoscapadngan, line 54: The selected postgres accounts already have a gateway. Why is it that we still define a gateway over here? Seems a bit redundant? what if the gateway in the account does not match the selected gateway?
Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+  host: string;
+  port: number;
+  database: string;
+  sslEnabled: boolean;
+  sslRejectUnauthorized: boolean;
+  sslCertificate?: string;
+  username: string;
+  password?: string;
+};
+
```
sheensantoscapadngan, line 67: Does it make sense to do a network sweep of Postgres instances even when the actual credential accounts are linked or scoped to the specific host and port that's defined within them? 

Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+    port: connectionDetails.port,
+    database: connectionDetails.database,
+    sslEnabled: Boolean(connectionDetails.sslEnabled),
+    sslRejectUnauthorized: connectionDetails.sslRejectUnauthorized !== false,
+    sslCertificate: connectionDetails.sslCertificate,
+    username: credentials.username,
+    password: credentials.authMethod === PamPostgresAuthMethod.AwsIam ? undefined : credentials.password
+  };
+};
+
```
sheensantoscapadngan, line 89: When you import a discovered PAM account, you're importing it under a specific database. Right now, by default, we use whatever the scanning credential is set to. The discovery process that we execute is able to fetch all the accounts for the entire instance. I don't think we can assume implicitly that because the discovery account was set to find actual accounts within `postgres`, we're supposed to set the imported ones to have `postgres` as well because they could very well be used with different databases altogether (like `infisical`).

Does it make sense for users to be able to override the database field value during the import process? 

Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+// an IAM token is minted per host/port/user pair, which a sweep across arbitrary targets cannot do, so only password accounts can drive a scan
+const isUsableAccount = (account: TPostgresAccount) => Boolean(account.password);
+
```
sheensantoscapadngan, line 92: Continuing on the above comment regarding the imported database field value, what would happen then if I wanted to use this staged account across two different databases, like let's say 'postgres' and 'infisical'. Can I do import twice? We should be able to handle that as well right? seems pretty common

Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+    signal: AbortSignal
+  ): Promise<{
+    accounts: TDiscoveredAccount[];
+    error?: TDiscoveryMachineError;
+    machine: string;
+    complete: boolean;
+  }> => {
+    const machine = `${host}:${port}`;
+    if (signal.aborted) return { accounts: [], machine, complete: false };
+
```
sheensantoscapadngan, line 149: In terms of general UX, shouldn't we be asking them to input the passwords as part of the import step? 

Image

I would think that after importing an account, you should be able to just make use of it automatically. In the future, we will have support for being able to automatically rotate the actual password of the account, but for now I feel like it would also make sense for us to have password input.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+    const candidates = [
+      ...accounts.filter((a) => a.host === host && a.port === port),
+      ...accounts.filter((a) => a.host !== host || a.port !== port)
+    ].filter(isUsableAccount);
+
```
sheensantoscapadngan, line 154: Gateway was empty after I imported the PAM account from staging. Is this intentional? Shouldn't we set it to the gateway value that was used to discover that entity? 
Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+        discovered.push(...result.accounts);
+        if (result.complete) scannedAccountMachines.push(result.machine);
+      }
+    }
+    if (droppedInstances) {
+      logger.warn(
+        `PAM PostgreSQL discovery dropped instances at the per-scan limit [instances=${droppedInstances}] [limit=${MAX_ACCOUNTS_PER_SCAN}]`
+      );
+    }
+
```
sheensantoscapadngan, line 270: Could we have used the exact same sheet for PAM account creation here? 

Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+      });
+
+      try {
+        const result = await db.raw<{ rows: { rolname: string }[] }>(ENUMERATION_QUERY);
+        return result.rows.map((row) => row.rolname);
+      } finally {
+        await db.destroy();
+      }
+    });
+
```
sheensantoscapadngan, line 134: After a postgres account is imported, the "staged account" becomes empty and yet checking the scan history shows there being one account. 

Image

Image

I think there's value in seeing references to the imported entities. Perhaps we can have it so that stale accounts, instead, would be discovered accounts. From there, you can see all the imported ones and all the stale ones as well, but then it would be more like a status, a badge or something. 
Image

Right now it just seems like it's a mystery. Although the accounts discovered are already within Infisical, there is no way for me to consolidate that

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,283 @@
...
+                rejectUnauthorized: account.sslRejectUnauthorized,
+                servername: host,
+                ...(account.sslCertificate ? { ca: account.sslCertificate } : {})
+              }
+            : false
+        },
+        acquireConnectionTimeout: QUERY_TIMEOUT_MS,
+        pool: { min: 0, max: 1 }
+      });
+
```
sheensantoscapadngan, line 126: When you say "are skipped", do you mean you can't define it as one of the selected accounts? hmm current wording might be misleading

Image

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: I guess that could be a decent UX addition, but this just directly copies linux discovery. We'd need some custom component to make the autocomplete work

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: This is intentional and follows every other discovery type. You can select multiple credential accounts so there could be a conflict of which gateway to choose from. Because of this we decided to just let users select a specific gateway for discovery making things simple. It's also possible that you may use a generic set of admin credentials across different networks/environments.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: I think we should remove the CIDR range support but keep the IPs/hosts input. You may want to scan hosts that are not directly linked with your credential account.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: This would break our discovery model. I think for this to work we need to take one of two paths:
1. Allow a single postgres account to store multiple databases (and let users select which to connect to?)
2. Implement account duplication so that the user can import an account and then duplicate it and change the database

If we allow multi-importing, things like tracking staleness will probably get super complex

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: I don't think we need to do this, especially not in this PR. Every discovery source faces this same problem, and I think it's perfectly fine until we add reconsiliation.

Once we add reconciliation, we can customize how it functions. Either ask admins for the password right away, or take the same password as the credential account, or use a hardcoded password. There will be lots of flexibility in how we can build this.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: Right now we're just following what every other discovery source does and use that simple import modal. Not sure what else you'd be customizing besides the target location and the template though, so the sheet seems unnecessary.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: So the first thing you pointed out is intentional. We just state how many accounts a scan found, and if it finds 1 it'll always show 1 even if that 1 is imported. This is ok.

As for surfacing discovered accounts, this would be a bigger change to think about likely outside of this PR, as this would affect our other 2 discovery sources. I think right now users can just go to the accounts page and see every imported and stale account with badges just fine. The only missing functionality is seeing the account for a specific discovery source. We can take this to Slack to discuss

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: Originally you could select them but they'd just be skipped. I've pushed a change that does not let you select them for discovery at the API level. Filtering them out of the dropdown would require adding a new dedicated endpoint that decrypts part of the row and just adds too much overhead so I skipped doing that.

I also reworded the docs to say that IAM accounts "cannot be used" rather than "are skipped".

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: This is intentional, all discovery sources are like this. The gateway is supposed to come from the template so we don't touch overrides by default.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: Not sure how we'd do this. If we allow someone to set database in the import modal, it would be weird since different selected accounts may need to access different databases. And adding staged account row editing also seems weird because its against how discovery functions: accounts are untouched and match the source until after import. I think this can be tackled in a separate initiative, right now nothing stops someone from just editing the database after import.

This may be something to look into when looking at reconciliation for importing staged accounts. Rules for setting a password and database would be great.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
sheensantoscapadngan: Why will there be any conflict over which gateway to choose when the credential accounts themselves already have an attached gateway? couldn't we just to go through each linked discovery account and check one by one using their respective gateways? 

I know we do this for the other discovery type and so this is us evaluating whether we wanna go through with that pattern moving forward or not

sheensantoscapadngan: Let's say you have a PAM account, which is defined inside private network 1, which has a gateway. Let's refer to that as gateway 1. Now, would it even make sense for you to use that credential to search for PAM accounts that are in a different network altogether where Gateway 2 exists? are we expecting people to use it like that?

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
sheensantoscapadngan: Most people would hit this problem, though no? If it will indeed break our discovery model, then it would make sense to rethink that instead of just accepting that we can't support common use-cases

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
sheensantoscapadngan: yeah I know that what we're following as part of this implementation is how we did the previous discovery sources, but I don't think we should use that as a basis for justifying the decisions here. We're still in the early stages in terms of this discovery feature, and so now would be the perfect time to make certain adjustments if we're seeing that it doesn't cover what we intend it to cover. 

Why couldn't we just define the gateway value right away with the one that was used to discover the account?

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
sheensantoscapadngan: I'm more so questioning whether the free text input for targets even makes sense in the first place. A Postgres username, password, and gateway trio is only valid for one specific Postgres instance. Does it even make sense for you to use that to try and randomly test multiple other instances? 

Again, I know that this isn't how we did all the other discovery sources, and so I'm bringing this up as a way to rethink what we have implemented.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
sheensantoscapadngan: I think this is a notable UX improvement we can make now

I don't see why we necessarily have to leave it for later though. We can spend some time getting this in the right shape now instead of accumulating debt to revisit... that would just lead to redundant work no?

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
sheensantoscapadngan: Wouldn't now be a better time then because we only have two discovery sources to revisit?

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: Username + password can absolutely be re-used across multiple instances. Like I pointed out in my other comment someone could have a generic admin credential that they use across multiple instances for discovery purposes.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: > go through each linked discovery account...

And what happens if an account with credentials postgres:postgres technically works as a credential, but it's gateway points to the wrong network? The point is to allow one account (regardless of gateway) to be able to try it's credentials against different networks/machines. All we take is the username+password pair, that's it.

> would it even make sense for you to use that credential to search for PAM accounts...

Depends on if the admin expects that account's username+password combination to work. If it's an account with a unique password within private network 1, then it would be useless to use it for discovery within private network 2. Cross-network only works when you have a credential pair that works across different networks and instances.

Also there's a technical limitation to constantly iterating every gateway. At the moment we open one gateway channel and then iterate through passwords through that channel. The alternative would be opening and closing channels for every host and every login attempt on that host.

MOST IMPORTANTLY I think there's a real risk to taking gateways straight from accounts. Currently you choose ONE gateway per discovery source, but the alternative is using MULTIPLE based on which account credential works. What if the host we're scanning is localhost and the wrong gateway gets selected purely because of the credential order? Now we've discovered accounts on the completely wrong instance because we assumed the gateway.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: Your example does not specifically apply to discovery. I can say the exact same thing about a regular Postgres account:

what would happen then if I wanted to use this ~~staged~~ account across two different databases, like let's say 'postgres' and 'infisical'.
image

The problem isn't with discovery, it's that we built Postgres accounts so that they only store ONE database each. This means that to address the problem, we're not rethinking discovery but instead rethinking the Postgres account model itself.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: There is no new debt to accumulate since every discovery source already works like this.

Adding this would require thinking about a mechanism that allows us to input the passwords of potentially 100+ accounts during the import step. How do we do this? Do we stack password modals one-by-one?

Also, what even guarantees that the platform admin setting up discovery KNOWS the passwords? In most cases, the team owning the resource has that information. What if the admin is just setting up discovery pointing to that team's environment and using that team's account for discovery credentials. With the current system the platform admin just imports the staged accounts into that team's folder and they take over and set the credentials themselves. Adding a hard block on import until passwords are provided now shifts that responsibility to the platform admin.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: > Why couldn't we just define the gateway value right away with the one that was used to discover the account?

Because when we first made discovery, me, the reviewer, and anyone else involved in the initiative agreed that you'd simply set that gateway on a template and then choose that template for every imported account.

The flow is as such:
- Platform admin creates a discovery source called `us-east-prod-postgres`
- Platform admin creates a postgres template called `us-east-prod` and sets the same gateway as the discovery source
- Platform admin runs discovery, selects all staged accounts, and imports them with the newly created template
- Done

Automatically setting an account override just sounds wrong. These are not supposed to be used for managing large amounts of accounts, these are for individual overrides in niche usecases.

Additionally, platform admins are the only people who can BOTH edit the discovery sources and account templates. They cannot (by default) see accounts that they imported and their gateway overrides. So the permissioning system also favors the original model.

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
lb-vn: As long as it's in a separate ticket and PR then sure. I don't see a reason to put it into this one and hold off release. This would require discussions and cause delays. We don't even know what direction to take here yet

backend/src/ee/services/pam-discovery/postgres/postgres-discovery-factory.ts
```diff
@@ -0,0 +1,281 @@
+import slugify from "@sindresorhus/slugify";
```
sheensantoscapadngan, line 1: IGNORE PLACEMENT

we can remove the "dependencies" column for Postgres right? 
Image

sheensantoscapadngan (approved):

PR #7792, opened 2026-08-24
improvement(admin): remove unused SAML/OIDC trust-email toggles and wire LDAP trust

victorvhs017:
## Context

The Server Admin Console exposed three “trust emails” toggles (SAML, LDAP, OIDC), but only org-level **SSO enforcement** actually controlled SAML/OIDC first-login verification. The SAML/OIDC server flags were stored and editable but never read by login flows, and the OIDC enable path even referenced “trust OIDC emails” as an SMTP workaround that did not exist in code.

This PR removes the unused `trustSamlEmails` and `trustOidcEmails` columns, API fields, and admin UI controls. **Trust LDAP emails** remains as the single instance-level bypass for LDAP provisioning. That flag is now wired into `ldap-config-service` (`skipEmailVerification` when `serverCfg.trustLdapEmails` is true). SAML and OIDC continue to skip verification only when the organization enforces SSO for a verified domain.

A migration drops the two removed columns. `updateServerCfg` treats PATCH bodies that only contain removed fields as a no-op so older clients do not hit Knex empty-update errors. Docs and admin copy are updated to match.

## Screenshots

image

## Steps to verify the change

1. Confirm `super_admin` no longer has `trustSamlEmails` / `trustOidcEmails`; `trustLdapEmails` remains.
2. Open `/login/admin` → General: only **Trust LDAP emails** is shown; save toggles successfully.
3. **LDAP:** With trust disabled, new LDAP user gets email verification on first login; with trust enabled, verification is skipped (without requiring SSO enforcement).
4. **SAML/OIDC:** First-login verification is skipped only when the org enforces SSO; no server-admin bypass exists.
5. **OIDC enable:** With broken SMTP, enabling OIDC returns the updated error (no mention of trusting OIDC emails).
6. **Backward compat:** `PATCH /api/v1/admin/config` with only `trustSamlEmails` / `trustOidcEmails` returns 200 without error.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [x] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/ldap-config/ldap-config-service.ts
```diff
@@ -504,7 +504,7 @@ export const ldapConfigServiceFactory = ({
     // When the org enforces SSO, the verified domain + IdP are authoritative, so we skip the
     // separate email-verification step (the email-domain ownership check above already proves the
     // org owns this domain, and password signup is blocked for enforced domains).
-    const skipEmailVerification = Boolean(organization.authEnforced);
+    const skipEmailVerification = Boolean(organization.authEnforced) || Boolean(serverCfg.trustLdapEmails);
```
Thiago-AS, line 504: On `main`, the login path does not reads `trustLdapEmails`. The setting has been inert since the auth revamp, but it still shows up as a working switch in the admin UI.

That means any instance where `super_admin.trustLdapEmails` is still `true`, set by an admin during the period when the setting did nothing, will start skipping LDAP email verification as soon as this deploys. Is that intended?

backend/src/ee/services/ldap-config/ldap-config-service.ts
```diff
@@ -504,7 +504,7 @@ export const ldapConfigServiceFactory = ({
     // When the org enforces SSO, the verified domain + IdP are authoritative, so we skip the
```
Thiago-AS: This has become stale then, worth changing or removing:

> LDAP-only orgs can't reach authEnforced at all, which is why the instance-wide flag has to exist.

backend/src/services/super-admin/super-admin-service.ts
```diff
@@ -474,6 +474,13 @@ export const superAdminServiceFactory = ({
       envOverridesUpdated = true;
     }
 
+    // A caller that only sends fields this API no longer recognises (the removed trustSamlEmails /
+    // trustOidcEmails switches, for instance) leaves nothing to write. Knex rejects an empty
+    // update, so treat it as a no-op instead of failing the request.
+    if (!Object.values(updatedData).some((value) => value !== undefined)) {
```
Thiago-AS, line 477: I'd go with 400 over 200. Zod strips unknown keys, so a `PATCH /api/v1/admin/config` with only `trustSamlEmails` parses to `{}` and returns 200 with a config body. 

If we want to start deprecating stuff, we need to let a client that still relies on the field know, so 400 makes the removal visible to the callers that need to know, otherwise we won't ever be able to fully remove that.

Thiago-AS (changes requested):

backend/src/ee/services/ldap-config/ldap-config-service.ts
victorvhs017: That should be expected behavior for users who have that setting turned on. We broke that during the auth revamp. For SAML and OIDC it was moved to the org settings, but LDAP is in a limbo.

Thiago-AS (approved):

PR #7801, opened 2026-08-25
fix(webhook): allow self hosted users to increase timeout and change job behavior

adilsitos:
## Context
The previous Webhook job was being scheduled and if a new change happened while a previous Webhook was running, this new one would be lost.  Added a `replace:true` so if a change arrives while a previous change was enqueued, we just removes the old one and updates it with the fresh data. Also added `keepLastIfActive:true` so if a change arrive while a job is delivering this is enqueued to run after the job finishes. 

This also allows self hosted users to define the expiration time they want for the webhook

## Steps to verify the change
- Create a server that has a high delay (around 14 seconds)
- Create a webhook pointing to this new server
- create secrets (the webhook needs to match the path of these secrets)
- change the secrets several times. Make sure that at the end only 2 requests were made to the server with delay (replace took action, so only the fresh one was sent.  keepLastActive also took action, since the first one was also sent).
- Play around with the admin and change the timeout there. 

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/lib/config/env.ts
+++ b/backend/src/lib/config/env.ts
@@ -483,6 +483,18 @@ const envSchema = z
     PAM_AWS_SECRET_ACCESS_KEY: zpStr(z.string().optional()),
     /* ----------------------------------------------------------------------------- */
 
+    /* Webhooks ----------------------------------------------------------------------------- */
+    WEBHOOK_TRIGGER_TIMEOUT_MS: z.coerce
+      .number()
+      .int()
+      .min(1000)
+      .max(40 * 1000)
+      .default(15 * 1000)
+      .describe(
+        "Wall-clock ceiling for a single outbound webhook delivery before it is aborted. Capped at 40s because the webhook test endpoint delivers inline in the request path."
+      ),
+    /* ----------------------------------------------------------------------------- */
+
     /* App Connections ----------------------------------------------------------------------------- */
     ALLOW_INTERNAL_IP_CONNECTIONS: zodStrBool.default("false"),
 
@@ -1068,6 +1080,16 @@ export const overwriteSchema: {
         key: "DISABLE_PUBLIC_SECRET_SHARING",
         description:
           "Disable creation of unauthenticated public secret shares (the /share-secret page). Set to 'true' to block public sharing."
+      }
+    ]
+  },
+  webhooks: {
+    name: "Webhooks",
+    fields: [
+      {
+        key: "WEBHOOK_TRIGGER_TIMEOUT_MS",
+        description:
+          "How long, in milliseconds, to wait for a webhook receiver to respond before the delivery is aborted. Defaults to 15000. Must be between 1000 and 40000."
       }
     ]
   }
--- a/backend/src/services/secret/secret-queue.ts
+++ b/backend/src/services/secret/secret-queue.ts
@@ -140,6 +140,7 @@ export type TGetSecrets = {
 
 const MAX_SYNC_SECRET_DEPTH = 5;
 const SYNC_SECRET_DEBOUNCE_INTERVAL_MS = 3000;
+const WEBHOOK_DEBOUNCE_INTERVAL_MS = 1000;
 
 export const uniqueSecretQueueKey = (environment: string, secretPath: string) =>
   `secret-queue-dedupe-${environment}-${secretPath}`;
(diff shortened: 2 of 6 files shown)


frontend/src/pages/secret-manager/SettingsPage/components/WebhooksTab/AddWebhookForm.tsx
victorvhs017, line 273: We need to update this placeholder, too

backend/src/services/webhook/webhook-fns.ts
```diff
@@ -460,6 +460,7 @@ export const fnTriggerWebhook = async ({
       } as TWebhookPayloads;
       const payload = getWebhookPayload(formattedEvent);
       if (!payload) return;
+      logger.info({ hookId: hook.id }, "Triggering webhook");
```
victorvhs017, line 463: It's hard to match this log with the started and finished logs, so tracking the job we are talking about can be hard.

These are the other 2:

logger.info({ environment, secretPath, projectId }, "Secret webhook job started");
logger.info({ environment, secretPath, projectId }, "Secret webhook job ended");

backend/src/lib/config/env.ts
```diff
@@ -483,6 +483,18 @@ const envSchema = z
...
+    WEBHOOK_TRIGGER_TIMEOUT_MS: z.coerce
+      .number()
+      .int()
+      .min(1000)
+      .max(40 * 1000)
+      .default(15 * 1000)
+      .describe(
+        "Wall-clock ceiling for a single outbound webhook delivery before it is aborted. Capped at 40s because the webhook test endpoint delivers inline in the request path."
+      ),
+    /* ----------------------------------------------------------------------------- */
```
victorvhs017, line 496: Because this is an environment variable with a default, this can be a bit misleading. Because there's no env set.

Image

backend/src/services/secret/secret-queue.ts
victorvhs017, line 1: I think we should update the concurrency for the SecretWebhook queue. `concurrency: 1` is the default, and in my tests, a webhook in an unrelated path was delayed until the webhook on another path finished; this is even worse when we increase the timeout.

We can put { concurrency: 5 }

victorvhs017, line 720: Another thing I got in my tests, which is unrelated to this change, but I think we should fix as part of improving this flow, is that these retries are not working for webhooks.

backend/src/services/secret/secret-queue.ts
```diff
@@ -703,10 +704,14 @@ export const secretQueueFactory = ({
         }
       },
       {
-        jobId: `secret-webhook-${environment}-${projectId}-${secretPath}`,
+        deduplication: {
+          id: `secret-webhook-${environment}-${projectId}-${secretPath}`,
+          keepLastIfActive: true,
+          replace: true
```
victorvhs017, line 710: I checked prod Redis:` ZCARD bull:secret-webhook:failed` is 5. It means that when we move to this new deduplication, BullMQ starts tracking the jobs in a different way, and those 5 legacy ones that never cleared on fail will be in Redis forever with no TTL.

We could clean up on boot, using the same pattern as legacy repeatable jobs in queue-service.ts

victorvhs017 (changes requested):

backend/src/lib/config/env.ts
adilsitos: Do you think we should just increase the timeout and don't have this option? I don't see any problem in increasing it to 30 seconds, but as Akhi said, this can be problematic on cloud instances. I thought about adding this so they can change it on their side.

victorvhs017: After testing, I think this can be problematic because it will hold one of our 5 concurrency slots for the jobs (after we update it) for that long, and if we fix the retry, it would be (40 + backoff) * 5, so 200s for a webhook that is timing out.

adilsitos: Gotcha. Maybe if we just fix the retry on failed this would already work for their use case, right? Because, the problem is the cold start, so a first try can break, but the next one would not get a cold start and nothing bad would happen. Makes sense?

PR #7817, opened 2026-08-25
feat(abac): add trim suffix 

adilsitos:
## Context

add trim suffix for abac 

## Steps to verify the change
- go into the organization and define the metadata of a user (e.g. `namespace = myapp-pr-2`
- go into a secrets management project and create a project role like this: `/{{ trimSuffix identity.metadata.namespace '-pr-*' }}`
- log in with this other user (or assume privileges)
- validate that you can just go into the folders that you have permission using your metadata. 

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/permission/permission-fns.ts
```diff
@@ -423,17 +424,46 @@ const expandLegacyForbidActions = <T extends RawRuleOf<MongoAbility<ProjectPermi
...
   return textStr.startsWith(prefix) ? textStr.substring(prefix.length) : textStr;
 };
 
+const hbsTrimSuffix = (text: string, suffix: string) => {
+  const textStr = String(text || "");
+  if (!textStr) return textStr;
+
+  if (typeof suffix !== "string" || !suffix) return textStr;
+
+  if (!picomatch.scan(suffix).isGlob) {
```
varonix0: question: why are we doing the extra checks on glob paths?

i would think that:
```
{{ hbsTrimSuffix "/service-**-operator-staging" "-staging" }}
```
works just like nay other string 🤔

varonix0: wondering if this is gonna be hard to undo down the road if we don't want this behavior

mathnogueira: This is an optimization, otherwise it would need to go through the loop on line 455 to remove static prefixes. (can be from 2 to 4 times slower).

adilsitos: Yep. If we know that the type is not glob, we can trim it right away (if we are going to support glob)

backend/src/ee/services/permission/permission-fns.ts
```diff
@@ -423,17 +424,46 @@ const expandLegacyForbidActions = <T extends RawRuleOf<MongoAbility<ProjectPermi
...
+  let isSuffixMatch: (input: string) => boolean;
+  try {
+    isSuffixMatch = picomatch(suffix, { dot: true });
+  } catch {
+    return textStr;
+  }
+
+  for (let i = textStr.length; i >= 0; i -= 1) {
+    if (isSuffixMatch(textStr.slice(i))) return textStr.slice(0, i);
+  }
```
mathnogueira: Well, this works, but requires a lot of tests to find the shortest path. Tried a few approaches: regexes, using `picomatch.scan`, etc. But I cannot remove this loop unfortunately.

Using picomatch's RE directly makes this 40 to 60% faster, but I'm not comfortable with using it directly because I know picomatch runs some validations before running the regex and this can prevent security issues know to Regex (picomatch is not RE2 compliant)

adilsitos: we have a maximum of 256. But the other problem is that this can generate unexpected behavior if configured wrong. we can just leave the first case (without picomatch) and that is it.

mathnogueira (dismissed):

backend/src/ee/services/permission/permission-fns.ts
mathnogueira: > we can just leave the first case (without picomatch) and that is it.

We can't, user is requesting a way to strip dynamic content, we need a glob matching.

backend/src/ee/services/permission/permission-fns.ts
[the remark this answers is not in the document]
adilsitos: this was fixed @veria-ai

mathnogueira (approved):

PR #7820, opened 2026-08-26
feat(pki): expose the latest renewal of a certificate so agents can follow renewals

carlosmonastyrski:
## Context

Adds an `orderId` column that groups every certificate in a renewal lineage, backfilled from existing `renewedFrom` links and inherited on renewal across all certificate authority types, and exposes a derived `latestRenewalCertificateId` on `GET /v1/cert-manager/certificates/:id`.
That field resolves the newest usable certificate in the requested certificate's own renewal chain in a single query, so a client no longer walks the chain one request at a time; it only ever names a descendant of the certificate you asked about, and never one that is revoked or expired. `orderId` is internal and omitted from every certificate response. Also surfaces the certificate ID in the details Overview card and documents the agent behaviour.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/db/migrations/20260824173418_add-certificate-order-id.ts
```diff
@@ -0,0 +1,37 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  if (await knex.schema.hasColumn(TableName.Certificate, "orderId")) return;
+
+  await knex.schema.alterTable(TableName.Certificate, (t) => {
+    t.uuid("orderId").notNullable().defaultTo(knex.raw("gen_random_uuid()"));
+    t.index("orderId");
```
sheensantoscapadngan: shouldn't we be indexing this using the concurrent approach so that it doesn't block production once we deploy

backend/src/services/certificate/certificate-dal.ts
```diff
@@ -750,6 +744,45 @@ export const certificateDALFactory = (db: TDbClient) => {
...
+      const result = await (tx || db.replicaNode()).raw<{ rows: { id: string }[] }>(
+        `
+        WITH RECURSIVE descendants AS (
+          SELECT id, "createdAt", status, "notAfter", 1 AS depth
+          FROM ??
+          WHERE "renewedFromCertificateId" = ? AND "orderId" = ?
+          UNION ALL
+          SELECT c.id, c."createdAt", c.status, c."notAfter", d.depth + 1
+          FROM ?? c
+          INNER JOIN descendants d ON c."renewedFromCertificateId" = d.id
```
sheensantoscapadngan: Why do we need recursion here? Can we not just order by order ID and fetch the latest one? or am I misunderstanding this?

backend/src/db/migrations/20260824173418_add-certificate-order-id.ts
carlosmonastyrski: Right now it won't be able to block it since the table size is not that big yet, but added it as a safe guard 👍

sheensantoscapadngan (approved):

PR #7827, opened 2026-08-26
fix: resolve provisioning identifiers through SSO user aliases

Thiago-AS:
## Context

Provisioning systems driven by an IdP often only know users by the identifier the IdP asserts (a UPN like for example) rather than by their mailbox. Infisical already stores that identifier: OIDC/SAML/LDAP login writes it to user_aliases.externalId. But only login read it back. Every invite, removal, and membership lookup resolved people by users.username alone, so provisioning by an IdP identifier silently created a duplicate account (or 403'd on an SSO-enforced org) and the real user never got the access.

This makes provisioning resolve the same identifier login already does. users.username still equals the user's real email, so nothing about mail delivery, invites by email, or SCIM changes.

Repercussions worth knowing:

- Applies to every org with SSO aliases, not opt-in. It only engages when an identifier matches no username.
- Duplicate aliases now fail loudly. Two users sharing an externalId in one org returns 400 instead of silently creating a shadow user. Picking one would be a guess about which human an identifier names.
- First login adopts a placeholder instead of duplicating it. Someone provisioned by identifier before they ever log in gets their placeholder row rewritten to the asserted mailbox, so the access it was granted is not stranded. Refused for anything a human has claimed (accepted, email-verified, password set), anything already bound to an IdP, ghosts, and anything without a membership in the logging-in org. OIDC only; SAML and LDAP are deliberately unchanged.
- Removal is symmetric. Anything provisionable by identifier is removable by it, and a partial or ambiguous match rejects the whole request rather than removing some people.
- Fixes a pre-existing bug: removing one person named by both their email and their username returned 400.

## Steps to verify the change

make up-dev-oidc      # brings up Keycloak with a pre-seeded realm
make seed-dev-oidc    # verified domain + active OIDC config on an `oidc` org

1. Log in at http://localhost:8080/api/v1/sso/oidc/login?orgSlug=oidc as john / password123!. Confirm one users row and one user_aliases row holding the Keycloak sub.
2. Org Access Control > Invite, and enter that sub instead of the email. The member list should show the existing user, and no second row should appear keyed on the identifier.
3. Invite an identifier that matches no alias. It should fall through to a placeholder exactly as before.
4. Add the user to a project, then DELETE /api/v1/projects/:projectId/memberships with the identifier in usernames. The membership should go; the account, org membership, and alias should remain.
5. Adoption: seed a users row whose username is a Keycloak sub, give it an org membership and a project membership, then log in as that user. The same row should be rewritten to the real email, with the project membership intact and no second account.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/user-alias/user-alias-dal.ts
```diff
@@ -1,13 +1,42 @@
...
+   */
+  const findBySsoExternalIds = async (
+    { externalIds, orgIds, aliasTypes }: { externalIds: string[]; orgIds: string[]; aliasTypes: string[] },
+    tx?: Knex
+  ) => {
+    try {
+      return await (tx || db.replicaNode())(TableName.UserAliases)
+        .whereIn("orgId", orgIds)
+        .whereIn("aliasType", aliasTypes)
+        .whereIn("externalId", externalIds)
```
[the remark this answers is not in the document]
Thiago-AS: Aliases are always created against the account whose username is the email the IdP asserted, and only after we've verified the org owns that domain, so an unverified alias isn't a binding to some random account. The flag just tracks whether the user later finished the separate email-verification step. 

Staleness in the isStaleSsoAlias sense isn't something you can read off the row either: it's a mismatch between the alias and a live assertion, and this path has no assertion to compare against. Filtering on isEmailVerified would also break SCIM, which creates aliases with isEmailVerified: false and never flips it until the user logs in, so every SCIM-provisioned user would stop resolving by their IdP identifier before first login, which is exactly the case this exists for. And for the attack to pay off you'd need to already control an account at an org-verified address that the org's IdP asserts for the target subject, at which point SSO login and SCIM would bind to that account anyway.

backend/src/ee/services/oidc/oidc-config-service.ts
```diff
@@ -305,6 +311,27 @@ export const oidcConfigServiceFactory = ({
...
+            userAliasDAL,
+            orgDAL,
+            tx
+          });
+
+          if (adoption) {
+            newUser = adoption.user;
+            adoptedFromUsername = adoption.adoptedFromUsername;
+          }
+        }
```
scott-ray-wilson, line 333: we should reject the login when the matching provisioned placeholder has an inactive org membership:  the inactive row may be adopted and mutated, or, if adoption is declined for tenant isolation, this creates a second user with a fresh active membership. I reproduced the latter locally and the new account could select the organization. Returning `null` from adoption isn’t enough—we need to distinguish “no placeholder” from “inactive placeholder” and fail the login.

scott-ray-wilson (changes requested):

backend/src/ee/services/oidc/oidc-config-service.ts
Thiago-AS: Nice catch!

scott-ray-wilson (approved):
LGTM

PR #7837, opened 2026-08-26
fix: make approvals check for validation rules

mathnogueira:
## Context

Until now, if a change policy is in place and a new approval is created, validation rules would be ignored. So invalid secrets could be added to the project. This enforces the validation rules on change request creations.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
+++ b/backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
@@ -1840,6 +1840,8 @@ export const secretApprovalRequestServiceFactory = ({
       project.secretDetectionIgnoreValues || []
     );
 
+    const secretsToValidate: { key: string; value?: string; secretId?: string }[] = [];
+
     // for created secret approval change
     const createdSecrets = data[SecretOperations.Create];
     if (createdSecrets && createdSecrets?.length) {
@@ -1853,13 +1855,7 @@ export const secretApprovalRequestServiceFactory = ({
       if (secrets.length)
         throw new BadRequestError({ message: `Secret already exists: ${secrets.map((el) => el.key).join(",")}` });
 
-      await secretValidationRuleService.validateSecrets({
-        projectId,
-        environment,
-        envId: folder.envId,
-        secretPath,
-        secrets: createdSecrets.map((s) => ({ key: s.secretKey, value: s.secretValue }))
-      });
+      secretsToValidate.push(...createdSecrets.map((s) => ({ key: s.secretKey, value: s.secretValue })));
 
       commits.push(
         ...createdSecrets.map((createdSecret) => ({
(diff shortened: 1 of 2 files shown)


mathnogueira:
@greptile review this

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
```diff
@@ -1913,6 +1924,16 @@ export const secretApprovalRequestServiceFactory = ({
             ).values()
           ];
 
+          if (upsertSecrets.length) {
+            await secretValidationRuleService.validateSecrets({
```
lobocv: Is it common in typescript to have side-effects like this? I was confused why there is no value being returned for the validateSecrets call. When I looked at the implementation, it seems like it validates and throws an exception, which would stop the code in its tracks over here. I guess I'm just used to Go having explicit error handling, so it's a bit strange to me now.

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
```diff
@@ -1913,6 +1924,16 @@ export const secretApprovalRequestServiceFactory = ({
...
+              projectId,
+              environment,
+              envId: folder.envId,
+              secretPath,
+              secrets: upsertSecrets.map((s) => ({ key: s.secretKey, value: s.secretValue }))
```
lobocv: We're doing two validations, one for creates and one for updates. There is a failure mode that one part will succeed and the other one fails validation. This leads the system in an inconsistent state and the user has to figure it out.

Would it make more sense to do a single validation at the start of the function (for all secrets involved)? It's generally good practice to put all your validation up front before you do any mutations

lobocv (dismissed):
Approving but I left one comment for you to consider

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
mathnogueira: Yes, it's common. At the end it will be handled by our global exception handler. But I agree with you, Go's approach is easier to follow and more explicit. I can't wait until we start migrating things to Go 😄

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
mathnogueira: This will not lead to inconsistent state because this is only populating the `commits` array. We only write the state after we have all commits (line 2186). Until then, it's building the state and validating it.

But you got a good point there, we should aim for a single validation after we have the state built.

lobocv (approved):
Looks good! thanks for streamlining validation!

PR #7843, opened 2026-08-27
improvement: add port number to Splunk provider

Thiago-AS:
## Context

- Add port number support to audit log stream Splunk provider.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/ee/services/audit-log-stream/splunk/splunk-provider-schemas.ts
```diff
@@ -30,6 +32,7 @@ export const SplunkProviderCredentialsSchema = z.object({
         ctx.addIssue({ code: "custom", message: "Invalid hostname" });
       }
     }),
+  port: z.number().int().min(1).max(65535).optional(),
```
victorvhs017, line 35: If you don't pass the port in a PATCH now, the value will be reset, and you will be using the default port.

victorvhs017 (changes requested):

backend/src/ee/services/audit-log-stream/splunk/splunk-provider-schemas.ts
Thiago-AS: It should never happen, because the UI always load the port, but let me add a guard like the others.

victorvhs017 (approved):

PR #7853, opened 2026-08-27
fix: pki ux patches

sheensantoscapadngan:
## Context
This PR adds support for the following:
- real-time policy feedback during certificate issuance and renewal
image

- improved certificate authorities page design
image

- added on-click navigation for table rows

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
```diff
@@ -82,7 +82,7 @@ export const CaSection = () => {
   };
```
carlosmonastyrski, line 82: Image

Maybe we can ask @scott-ray-wilson for his feedback on this "missing field" component

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
```diff
@@ -82,7 +82,7 @@ export const CaSection = () => {
...
```
carlosmonastyrski, line 83: Image
Looks a bit off with long values, maybe we can use tooltips or a better format?

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
```diff
@@ -82,7 +82,7 @@ export const CaSection = () => {
...
 
   return (
```
carlosmonastyrski, line 84: Image
same here, maybe a hover card where deny values are shown with a red color 🤔

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
```diff
@@ -82,7 +82,7 @@ export const CaSection = () => {
...
-    <Card className="mb-6">
+    <Card>
```
carlosmonastyrski, line 85: Image
Let's also add it on the TTL max value please

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
```diff
@@ -82,7 +82,7 @@ export const CaSection = () => {
...
       <CardHeader>
```
carlosmonastyrski, line 86: Image
DC fields look a bit off, only shows it on the first one, maybe we could show it below the last one? 🤔 or break it into how each should look like?

carlosmonastyrski: Image
same with the error, as soon as I start typing it's showing me this error when it should show it only when I try to continue (otherwise I'll see it spamming every time I type until I get all DC fields there)

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
```diff
@@ -82,7 +82,7 @@ export const CaSection = () => {
...
         <CardTitle>
```
carlosmonastyrski, line 87: Image
If I define two requires for a field, it's not populating it when I click on add missing fields

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
carlosmonastyrski: Image
Also if I define many field policies, then this gets too long

sheensantoscapadngan: removing this altogether

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
sheensantoscapadngan: added proper spacing. i think red would be mistaken as an "error" message

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
carlosmonastyrski, line 82: Do you think it'll be a good idea to show these in a better structure like we do on PKI Syncs?

Image

Image

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
carlosmonastyrski, line 83: tooltips positions are a bit off

Image

Image

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
```diff
@@ -82,7 +82,7 @@ export const CaSection = () => {
   };
 
   return (
-    <Card className="mb-6">
```
carlosmonastyrski, line 85: if I have two requires for DC, the enforcement breaks

Image

Image

Image

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
carlosmonastyrski, line 86: DC deny information is not displayed

Image

Image

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
carlosmonastyrski, line 87: Let's show the TTL maximum as  "years/months/days/hours" instead of "y/m/d/h"
Image

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
sheensantoscapadngan: wait what do you mean? this is intentional. what you're seeing in the error message is the expected allowed values for all three DCs combined

frontend/src/pages/cert-manager/CertificateAuthoritiesPage/components/CaSection.tsx
sheensantoscapadngan: Addressed this. It is now visible if the number of DC matches the one defined in the policy itself.

carlosmonastyrski (approved):

PR #7857, opened 2026-08-28
fix(auth): pool https agents and fix gateway sni

victorvhs017:
## Context

Kubernetes TokenReview (machine identity auth and the gateway Kubernetes executor) built a new `https.Agent` on every call. That re-parses the cluster CA into a fresh OpenSSL trust store and forces a full TLS handshake each time — TLS session cache is per-agent, so there was nothing to reuse. Gateway-tunneled reviews are addressed to `localhost`, but SNI still used `getKubernetesServerName` (undefined for IPs; `lastIndexOf(":")` also truncated IPv6). Node then checked the cert against `localhost` and handshake failed.

Callers that cannot go through `safeRequest` now share `getSharedHttpsAgent`, which uses the existing pinned-agent pool (in-process `Map`, cap 200, LRU via insertion order, `keepAlive`, `maxSockets` 50 / `maxFreeSockets` 10). Cache key is CA + `rejectUnauthorized` + `servername` (plus protocol/IP pin when used from `safeRequest`). Eviction calls `agent.destroy()` so idle keep-alive sockets are not stranded for the process lifetime. Direct TokenReview still omits SNI for IP hosts (RFC 6066); the gateway path uses `getKubernetesHostname` so identity is the cluster host, including IPs.

## Steps to verify the change

1. Direct Kubernetes identity login against a hostname API server. Confirm TokenReview succeeds and repeat logins reuse a connection (one TLS handshake, not one per login).
2. Same login with the API server as an IPv4 or IPv6 address (no SNI on the direct path; cert matched on IP SANs).
3. Login with token review through a gateway: hostname cluster, then an IP-hosted cluster. Handshake must use the Kubernetes host, not `localhost`.
4. Rotate the cluster CA on an identity and login again — a new agent must be used (CA is part of the pool key).

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/resource-auth-method/kubernetes-gateway-executor.ts
+++ b/backend/src/ee/services/resource-auth-method/kubernetes-gateway-executor.ts
@@ -1,8 +1,7 @@
-import https from "node:https";
-
 import { request as httpRequest } from "@app/lib/config/request";
 import { GatewayHttpProxyActions, GatewayProxyProtocol } from "@app/lib/gateway";
 import { withGatewayV2Proxy } from "@app/lib/gateway-v2/gateway-v2";
+import { getSharedHttpsAgent } from "@app/lib/validator/safe-request";
 
 import { TGatewayV2ConnectionDetails } from "../gateway-v2/gateway-v2-types";
 import { TKubernetesRequestExecutor } from "./kubernetes-auth-fns";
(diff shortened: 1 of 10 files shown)


victorvhs017:
@greptileai re-review

akhilmhdh (approved):

PR #7860, opened 2026-08-28
fix(api): thread tx through DAL reads in createProject to prevent connection-pool deadlock

Jaimin2687:
## Context

`createProject` in `project-service.ts` opens a database transaction (line 283) and, inside it, was making several DAL read calls that never received the `tx` handle. Each stray call checks out a **second connection** from the pool (`ormify` resolves `(tx || db.replicaNode())`, so an omitted `tx` always opens a new one). The transaction itself already holds one, so a single `createProject` request with a template that includes users, groups, and identities could tie up **8+ connections simultaneously**.

`DB_POOL_MAX` defaults to 10. Two concurrent template-based project creations can exhaust all connections — every request is blocked waiting for a connection that no other request can release. The application deadlocks and every API call (not just project creation) hangs until the process is killed.

### What was leaking

**Direct DAL calls missing `tx`:**

| Line | Call | Fix |
|------|------|-----|
| 387 | `userDAL.find({ $in: { username: … } })` | add `{ tx }` options arg |
| 466 | `groupDAL.find({ orgId: …, $in: { slug: … } })` | add `{ tx }` options arg |
| 524 | `identityDAL.find({ orgId: …, $in: { id: … } })` | add `{ tx }` options arg |
| 658 | `userDAL.findUserEncKeyByUserId(actorId)` | add `, tx` second arg |

**Service call performing unthreaded DB reads inside the transaction:**

| Line | Call | Fix |
|------|------|-----|
| 315 | `projectTemplateService.findProjectTemplateByName(...)` | Hoisted before the transaction — it runs `projectTemplateDAL.findOne` + 3 `findByTemplateId` calls internally, all without `tx`. The lookup is read-only and depends only on inputs available before the transaction. |

Beyond pool exhaustion, these reads also ran outside the transaction boundary, so they could return stale data that the subsequent writes depend on — a correctness issue on top of the availability one.

## Screenshots

N/A — backend-only change, no UI impact.

### Reproduction (before the fix)

1. Set `DB_POOL_MAX=5` in your `.env` to shorten the fuse.
2. Create an org and a project template that includes at least one user, one group, and one identity.
3. Fire 6 concurrent `POST /api/v1/workspace` requests using that template:
   ```bash
   for i in $(seq 1 6); do
     curl -s -X POST http://localhost:8080/api/v1/workspace \
       -H "Authorization: Bearer $TOKEN" \
       -H "Content-Type: application/json" \
       -d "{\"projectName\":\"pool-test-$i\",\"template\":\"your-template\"}" &
   done
   wait
   ```
4. All requests hang. Subsequent `GET` requests to any endpoint also hang. The process must be killed.

### Verification (after the fix)

1. Same setup, same concurrency.
2. All 6 requests complete (or fail cleanly on duplicate slugs).
3. Unrelated API calls continue to respond normally under load.

### Code review

- Confirm the `findProjectTemplateByName` call now sits **above** the `transaction()` call, not inside it.
- Grep the transaction block and confirm every `await` on a DAL method inside it now ends with `tx` or `{ tx }`. No call inside this transaction should resolve against the bare `db` pool.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/project/project-service.ts
+++ b/backend/src/services/project/project-service.ts
@@ -280,6 +280,21 @@ export const projectServiceFactory = ({
       throw new ForbiddenRequestError({ message: "You don't have permission to create a project" });
     }
 
+    let projectTemplate: Awaited<ReturnType<typeof projectTemplateService.findProjectTemplateByName>> | null = null;
+
+    switch (template) {
+      case InfisicalProjectTemplate.Default:
+        projectTemplate = null;
+        break;
+      default:
+        projectTemplate = await projectTemplateService.findProjectTemplateByName(template, {
+          id: actorId,
+          orgId: organization.id,
+          type: actor,
+          authMethod: actorAuthMethod
+        });
+    }
+
     const results = await (trx || projectDAL).transaction(async (tx) => {
       await tx.raw("SELECT pg_advisory_xact_lock(?)", [PgSqlLock.CreateProject(organization.id)]);
 
@@ -303,21 +318,6 @@ export const projectServiceFactory = ({
             message: "KMS does not belong in the organization"
           });
         }
-      }
-
-      let projectTemplate: Awaited<ReturnType<typeof projectTemplateService.findProjectTemplateByName>> | null = null;
-
-      switch (template) {
-        case InfisicalProjectTemplate.Default:
-          projectTemplate = null;
-          break;
-        default:
-          projectTemplate = await projectTemplateService.findProjectTemplateByName(template, {
-            id: actorId,
-            orgId: organization.id,
-            type: actor,
-            authMethod: actorAuthMethod
-          });
       }
 
       const slug = projectSlug || slugify(`${workspaceName}-${alphaNumericNanoId(4)}`);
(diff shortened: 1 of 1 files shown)


Jaimin2687:
Good catch — both reviewers are correct that `findProjectTemplateByName` (line 315) performs 4 DAL reads without `tx` inside the transaction.

Addressed in the force-pushed commit (`e6c8f28`): moved the entire template lookup **before** the transaction opens. The lookup is read-only and depends only on inputs available before the transaction (`template`, `actorId`, `organization.id`, `actor`, `actorAuthMethod`), so it does not need transactional consistency with the writes that follow.

This follows the pattern from `CODE_QUALITY.md`: _"read and validate, do the expensive and external work, then open a short transaction for only the writes that must be atomic together."_

The transaction block now contains zero DAL calls that bypass `tx`.

Jaimin2687:
Hi team, circling back on this connection pool deadlock fix now that the holidays are clear. I know threading tx through the DAL is a sensitive architectural change. Let me know if you need me to write any specific load tests or adjust the implementation to help get this reviewed. Happy to help.

PR #7866, opened 2026-08-28
feat(secret-sync): add Daytona app connection and secret sync

lobocv:
## Context

Closes [SECRETS-567](https://linear.app/infisical/issue/SECRETS-567/add-daytona-secret-sync-and-app-connection).

[Daytona](https://www.daytona.io/) runs sandboxes that consume organization-scoped secrets. Customers keeping those secrets in Infisical had no way to get them into Daytona, so they copied values by hand. This adds an app connection holding a Daytona API key, and a secret sync that pushes an Infisical environment and folder into that organization's secrets.

## What Daytona's API supports

Base URL `https://app.daytona.io/api`, bearer-authenticated with an API key carrying `manage:secrets`.

| Operation | Endpoint |
|---|---|
| List | `GET /secret/paginated?cursor=&limit=` |
| Create | `POST /secret` |
| Update value | `PATCH /secret/{secretId}` |
| Delete | `DELETE /secret/{secretId}` |

Three properties of that API drive the design:

- **Values are write-only.** A list returns `id`, `name` and an opaque `placeholder`, never the value. So the sync cannot import (`canImportSecrets: false`) and cannot skip a write because the value is unchanged. Every run rewrites every key.
- **Names are immutable** and constrained to `^[a-zA-Z_][a-zA-Z0-9_-]*$`. A rename is a delete plus a create, which the existing add-and-prune loop already produces.
- **API keys are organization-scoped, and nothing scopes below the organization.** None of Daytona's API paths contains a project, workspace, team or namespace concept, and the secret endpoints take no scope parameter.

## Design notes

**`destinationConfig` is empty.** The connection pins the organization, so there is nothing to configure. One consequence: `checkDuplicateDestination` returns early on an empty config, so duplicate detection does not apply to Daytona even for orgs that enable `blockDuplicateSecretSyncDestinations`. Making it apply means changing that early return in shared code, which belongs in its own PR.

**Credentials validate against `GET /secret/paginated`, not the key's own details.** Reading `GET /api-keys/current` to inspect its `permissions[]` would plausibly require `manage:api_keys`, so a key holding only `manage:secrets` could be rejected as invalid. Listing secrets tests exactly the capability the sync needs, and a 403 still yields a message naming `manage:secrets` and where to grant it.

**Invalid key names fail the run rather than being skipped.** Every key is checked before any write, so a run either writes everything or writes nothing instead of leaving Daytona half updated. The error names every offending key and the rule.

**`hosts` is never sent.** Daytona injects an opaque placeholder into the sandbox and substitutes the real value in its egress proxy, but only for hosts on that secret's allowed hosts list. Infisical never sets or changes that list: create sends name and value, update sends only the value, so an allowlist set in Daytona survives every later sync. Documented on the sync page.

## Screenshots

CleanShot 2026-09-02 at 9 06 06 AM@2x

CleanShot 2026-09-02 at 9 08 25 AM@2x

## Steps to verify the change

Manual, against a real Daytona account:

1. Create a Daytona API key with `manage:secrets`. Confirm the connection saves, and that a key without that permission is rejected with a message naming it.
2. Create a sync onto a folder holding several secrets and run it. Confirm each appears in Daytona, and that a sandbox resolves one of the values.
3. Change a value in Infisical, re-sync, confirm Daytona shows the new value.
4. Delete a secret in Infisical, re-sync, confirm it is removed. Repeat with **Disable secret deletion** on and confirm it survives.
5. Add a secret whose key breaks the name rule, such as `MY.KEY`. Confirm the run fails and the error names both the key and the rule.
6. Set allowed hosts on a synced secret in Daytona, re-sync, and confirm the list is still there.
7. Delete the sync with removal enabled and confirm the secrets are gone from Daytona.

## Out of scope, deliberately

- **Self-hosted Daytona.** The base URL is a constant in `IntegrationUrls`, as with Devin and Ona. Requests go through `safeRequest`, so an optional `instanceUrl` can follow if a customer asks.
- **`rsync` added to the Vale vocabulary.** Unrelated to Daytona, but required: adding `Daytona` changes the lint configuration, which escalates `lint-docs-branch` from the branch's `.mdx` files to the whole site, surfacing a latent pre-existing failure in `cli/commands/pam.mdx`.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
```diff
@@ -0,0 +1,187 @@
...
+import { TSecretMap } from "@app/services/secret-sync/secret-sync-types";
+
+import { TDaytonaListSecretsResponse, TDaytonaSecret, TDaytonaSyncWithCredentials } from "./daytona-sync-types";
+
+const DAYTONA_SECRET_NAME_PATTERN = /^[a-zA-Z_][a-zA-Z0-9_-]*$/;
+const DAYTONA_SECRET_NAME_RULE =
+  "Daytona secret names must start with a letter or underscore and contain only letters, digits, hyphens and underscores.";
+
+const DAYTONA_PAGE_SIZE = 100;
+const DAYTONA_MAX_PAGES = 100;
```
lobocv: This is too high, that's 10k secrets... let's limit to something more reasonable such as 10

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const DAYTONA_MAX_RETRIES = 3;
+const DAYTONA_BASE_RETRY_DELAY_MS = 500;
+
+const sleep = (ms: number) =>
+  new Promise<void>((resolve) => {
+    setTimeout(resolve, ms);
+  });
+
+const withDaytonaRetry = async <T>(fn: () => Promise<T>): Promise<T> => {
```
lobocv: Do we have any common http retry util funcs that we could use? This seems like it could be generic. Is there even a place to put such code? If not, lets keep this as it is.

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      try {
+        if (existing) {
+          // Daytona never returns a secret's value, so there is no way to skip an unchanged write.
+          await updateDaytonaSecretValue(apiKey, existing.id, value);
+        } else {
+          await createDaytonaSecret(apiKey, key, value);
+        }
+      } catch (error) {
+        if (error instanceof AxiosError && error.response?.status === 409) {
```
lobocv: Does typescript have any constants we can use for the http codes  instead of the magic numbers / literals ?

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
lobocv: Lowered to 10, so 1,000 secrets at the 100-per-page size. Fixed in 657af331da.

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
lobocv: Yes, and there is a place: `src/lib/retry/index.ts`. It already exports `retryWithBackoff`, but I did not reuse it directly because it retries on **any** thrown error. For Daytona that would retry the rejected-credential, missing-permission and name-collision paths, whose whole purpose is to fail fast with an actionable message rather than after several pointless round trips.

So I added `retryOnRateLimit` alongside it: generic, 429-only, exponential backoff, configurable retry count and base delay. Daytona now uses it and its private copy is gone. Fixed in 657af331da.

Worth flagging separately: `travis-ci`, `render`, `trigger-dev` and `ona` each carry their own near-identical 429 helper. Migrating those four is mechanical but touches working syncs, so I kept it out of this PR rather than widening it. Happy to do it as a follow-up.

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
lobocv: axios exports an `HttpStatusCode` enum, and axios is already the HTTP client throughout this code. Switched the `429` / `409` / `403` / `401` literals to `HttpStatusCode.TooManyRequests` / `.Conflict` / `.Forbidden` / `.Unauthorized`, in both the sync and the connection validator, and swapped `instanceof AxiosError` for `isAxiosError` while I was in there. Fixed in 657af331da.

There is also an `HttpStatusCodes` enum in `server/plugins/error-handler.ts`, but it is local to that file and not exported, so the axios one is the better fit here.

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
```diff
@@ -0,0 +1,168 @@
...
+  );
+
+  return secrets;
+};
+
+const createDaytonaSecret = (apiKey: string, name: string, value: string) =>
+  retryOnRateLimit(() =>
+    request.post(
+      `${IntegrationUrls.DAYTONA_API_URL}/secret`,
+      { name, value },
```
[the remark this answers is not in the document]
lobocv: Is this a problem because the `IntegrationUrls.DAYTONA_API_URL` could potentially be empty? What is this hosts field? I dont see us passing it in

backend/src/services/app-connection/daytona/daytona-connection-fns.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  // Listing secrets is the capability every Daytona sync needs, so validating against it proves both
+  // that the key works and that it carries manage:secrets. Reading the key's own details would not.
+  try {
+    await request.get(`${IntegrationUrls.DAYTONA_API_URL}/secret/paginated`, {
+      params: { limit: 1 },
+      headers: getDaytonaAuthHeaders(apiKey)
+    });
+  } catch (error: unknown) {
+    logger.error({ error }, "Failed to validate Daytona connection");
```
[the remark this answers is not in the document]
lobocv: Good catch, this was real. Redaction in `lib/logger/logger.ts` matches by key name to depth 3 (`*.*.${key}`) and the list does not include `Authorization`, while an Axios error carries the key at `error.config.headers.Authorization` — depth 4, wrong key name. So the raw error would have shipped a working Daytona API key to the log sink.

Removed the log line entirely in ad16606c2f. Every branch below it already throws a `BadRequestError` naming what went wrong, so nothing diagnostic is lost. Left a comment so it does not get added back.

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    cursor = data.nextCursor ?? undefined;
+    if (!cursor) return secrets;
+  }
+
+  logger.warn(
+    `Daytona secret listing stopped at the ${DAYTONA_MAX_PAGES} page cap; the destination may hold more secrets than were read`
+  );
+
+  return secrets;
```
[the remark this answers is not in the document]
lobocv: Agreed, and this was the more dangerous half of the cap problem: a truncated listing makes keys past the cap look *absent*, so reconciliation would recreate them and skip deletions it owed, all while reporting success.

Now throws in ad16606c2f rather than returning a partial inventory.

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
```diff
@@ -0,0 +1,168 @@
...
+import { TSecretMap } from "@app/services/secret-sync/secret-sync-types";
+
+import { TDaytonaListSecretsResponse, TDaytonaSecret, TDaytonaSyncWithCredentials } from "./daytona-sync-types";
+
+const DAYTONA_SECRET_NAME_PATTERN = /^[a-zA-Z_][a-zA-Z0-9_-]*$/;
+const DAYTONA_SECRET_NAME_RULE =
+  "Daytona secret names must start with a letter or underscore and contain only letters, digits, hyphens and underscores.";
+
+const DAYTONA_PAGE_SIZE = 100;
+const DAYTONA_MAX_PAGES = 10;
```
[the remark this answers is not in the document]
lobocv: Same root cause as the P1 above, and fixed by the same change in ad16606c2f: the listing now throws when it hits the cap instead of returning a partial inventory, so a stale secret can no longer survive past page 10 while the run reports success.

backend/src/services/secret-sync/daytona/daytona-sync-schemas.ts
```diff
@@ -0,0 +1,54 @@
...
+const DaytonaSyncDestinationConfigSchema = z.object({
+  organizationName: z
+    .string()
+    .trim()
+    .min(1, "Organization name is required")
+    .max(255)
+    .describe(SecretSyncs.DESTINATION_CONFIG.DAYTONA.organizationName)
+});
+
+const DaytonaSyncOptionsConfig: TSyncOptionsConfig = { canImportSecrets: false };
```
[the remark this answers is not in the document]
lobocv: Correct, and worth fixing. The generic `keySchema` validator in `secret-sync-schemas.ts` allows `/`, so `{{environment}}/{{secretKey}}` saved cleanly and then failed on every run — the worst shape for this, since the sync looks healthy until it does not.

Fixed in ad16606c2f. `CreateDaytonaSyncSchema` and `UpdateDaytonaSyncSchema` now refine on the schema's literal characters: each placeholder is substituted with a token that is itself valid, so only the literals are under test, and the result is checked against `DAYTONA_SECRET_NAME_PATTERN` (now shared with the sync rather than duplicated). Resolved names are still checked at sync time, since `{{secretKey}}` can expand to anything.

Verified against the parser:

```
ACCEPT  no key schema
ACCEPT  {{secretKey}}
ACCEPT  INFISICAL_{{secretKey}}
ACCEPT  {{environment}}_{{secretKey}}
ACCEPT  _{{secretKey}}
REJECT  {{environment}}/{{secretKey}}   -> syncOptions.keySchema
REJECT  1{{secretKey}}                  -> syncOptions.keySchema
```

The leading-digit case was not in the report but fails for the same reason, so it is covered too.

docs/integrations/secret-syncs/daytona.mdx
```diff
@@ -0,0 +1,153 @@
...
+                    "name": "my-daytona-sync",
+                    "projectId": "3c90c3cc-0d44-4b50-8888-8dd25736052a",
+                    "description": "an example sync",
+                    "connectionId": "3c90c3cc-0d44-4b50-8888-8dd25736052a",
+                    "environment": "dev",
+                    "secretPath": "/my-secrets",
+                    "isEnabled": true,
+                    "syncOptions": {
+                        "initialSyncBehavior": "overwrite-destination",
+                        "autoSyncEnabled": true,
```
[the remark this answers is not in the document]
lobocv: Confirmed against the schema: `GenericCreateSecretSyncFieldsSchema` defines a top-level `isAutoSyncEnabled` (defaulting to `true`), and neither `isEnabled` nor `syncOptions.autoSyncEnabled` exists. Zod strips both, so a user copying the sample and setting them to `false` would have got an auto-syncing sync anyway.

Fixed in ad16606c2f, in both the request and response samples.

Note the same wrong fields appear in `docs/integrations/secret-syncs/devin.mdx`, which is where I copied the sample from, so it is likely wrong across several sync docs. Out of scope here, but worth a sweep.

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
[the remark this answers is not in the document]
lobocv: The premise here is right, and it is worth stating precisely because it is not the usual "secret in an env var" case. Daytona sets the sandbox env var to an opaque *placeholder*, not the value, and their egress proxy substitutes the real value on the way out. So `hosts` is not incidental hardening — it is the mechanism that makes the placeholder design worth anything. Omit it and any code in the sandbox can send the placeholder to a host it chooses and have Daytona hand back plaintext.

Two things keep this a deliberate default rather than a hole:

**We match Daytona's own default.** A secret created in their UI without `hosts` behaves identically. This sync does not make a secret less restricted than creating it by hand.

**We never send `hosts`, on create or update.** The `PATCH` body is `{ value }` only, so a per-secret allowlist set in Daytona survives every subsequent sync. That is the escape hatch for anyone who wants restrictions: set them once in Daytona, and Infisical keeps writing values without touching them.

The reason not to require an allowlist in Infisical is that we would have exactly one field for the whole sync, applied uniformly to every key in the folder. `DB_PASSWORD` and `ANTHROPIC_API_KEY` want different allowlists, so a single sync-wide list is wrong for most secrets in it and would also overwrite anything set per-secret in Daytona. That is a worse outcome than inheriting their default and leaving per-secret control where the per-secret knob actually lives.

An **optional** sync-wide `hosts` field is a reasonable future addition for the uniform case, and I am happy to add it if you want it. I would not make it required.

One caveat I have not been able to verify without a live account, and which I have added to the manual test plan: this rests on `PATCH` treating omitted fields as unchanged. Their `UpdateSecret` schema is all-optional so that is the expected semantics, but if it instead resets omitted fields, every sync run would silently wipe a hand-set allowlist — which would turn this from a deliberate default into a real bug. Worth confirming before merge.

docs/integrations/secret-syncs/daytona.mdx
adilsitos, line 1: ignore comment placement.

There is a snippet called `SecretSyncBrowser` under the `docs` folder. You also need to add a reference to daytona there. Otherwise it won't show on the page below 

Image

docs/integrations/app-connections/daytona.mdx
adilsitos, line 1: ignore comment placement.

A similar snippet for app connections also exist, you also need to update it there.

docs/integrations/app-connections/daytona.mdx
```diff
@@ -0,0 +1,90 @@
+---
+title: "Daytona connection"
```
adilsitos, line 2: You can just leave `Daytona`.

docs/integrations/app-connections/daytona.mdx
```diff
@@ -0,0 +1,90 @@
...
+                Complete the form with:
+                - A **name** for the connection (for example `daytona-prod`)
+                - An optional **description**
+                - Your **Daytona API key**
+            </Step>
+            <Step title="Connection created">
+                Infisical validates the key against Daytona and confirms it carries `manage:secrets`. Once it passes, the connection is ready to use in your project.
+            </Step>
+        </Steps>
+    </Tab>
```
adilsitos, line 60: We are missing the screenshots here. It is super helpful for users to have a screenshot to see where they need to go into the external platform and also inside Infisical. 

If you don't use any app for screenshot, I highly recommend shottr. Super quick

docs/integrations/secret-syncs/daytona.mdx
adilsitos, line 1: We also need the screenshots of all the steps here.

docs/integrations/app-connections/daytona.mdx
```diff
@@ -0,0 +1,90 @@
...
+                Select **Add Connection** and choose **Daytona** from the list of available connections.
+            </Step>
+            <Step title="Fill out the connection form">
+                Complete the form with:
+                - A **name** for the connection (for example `daytona-prod`)
+                - An optional **description**
+                - Your **Daytona API key**
+            </Step>
+            <Step title="Connection created">
+                Infisical validates the key against Daytona and confirms it carries `manage:secrets`. Once it passes, the connection is ready to use in your project.
```
adilsitos: nit: maybe we don't need to say that we confirm that it carries the `manage:secrets`, this might be too technical to leave here. But we can leave it here as well, no problem.

docs/integrations/app-connections/daytona.mdx
```diff
@@ -0,0 +1,90 @@
+---
+title: "Daytona connection"
+description: "Learn how to configure a Daytona connection for Infisical."
+---
+
+[Daytona](https://www.daytona.io/) runs sandboxes that read organization-scoped secrets at runtime. Infisical connects to Daytona with an **API key**, which is used to push secrets into a Daytona organization through the [Daytona secret sync](/integrations/secret-syncs/daytona).
```
adilsitos: We are trying to only use the bolding when we reference something on the UI (e.g. On **add service** ...)

```suggestion
[Daytona](https://www.daytona.io/) runs sandboxes that read organization-scoped secrets at runtime. Infisical connects to Daytona with an API key, which is used to push secrets into a Daytona organization through the [Daytona secret sync](/integrations/secret-syncs/daytona).
```

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/DaytonaSyncFields.tsx
adilsitos, line 1: Maybe I'm missing something, but If this field will stay only inside Infisical I don't see the gain of defining it. What was the rationale?  

Image

docs/integrations/secret-syncs/daytona.mdx
adilsitos: Since this is a new class of sync (au sandbox), you can even create a new section. I saw that this is inside SAAS now, but maybe from a marketing perspective, it is better to have its own section. Good to check with @adkah is this makes sense.

docs/integrations/secret-syncs/daytona.mdx
```diff
@@ -0,0 +1,159 @@
...
+<Note>
+    Daytona secret names must start with a letter or an underscore, and can contain only letters, digits, hyphens, and underscores. A sync that includes a key Daytona rejects fails and names the offending key, so no secrets are written. Use a **key schema** to reshape names that do not fit.
+
+    The key schema itself is checked against the same rule when you create or update the sync, so a schema containing a slash is rejected up front rather than failing on every run.
+</Note>
+
+<Note>
+    Infisical does not set the **allowed hosts** list on the secrets it creates, so they are unrestricted in the same way as a secret created in Daytona without one. Daytona substitutes the real value for its placeholder on requests to any host.
+
+    To restrict a secret, set its allowed hosts in Daytona. This sync only ever writes the value, so a list you set there is preserved on later syncs.
```
adilsitos: If we want to have this, we can define this as a step on the sync (on the destination step). When we perform the creation, we attach this into the secrets. 

If we prefer to not add this option on the destination, in my opinion we can remove the section we say that a user can change the allowed hosts on daytona directly. The philosophy of secret syncs is that you should not touch the secrets on the external platform. Even though you can do this, we can omit this recommendation from here.

backend/src/lib/retry/index.ts
```diff
@@ -41,3 +43,25 @@ export const retryWithBackoff = async <T>(fn: () => Promise<T>, maxRetries = 3,
...
+  for (let attempt = 0; ; attempt += 1) {
+    try {
+      return await fn();
+    } catch (error) {
+      const isRateLimited = isAxiosError(error) && error.response?.status === HttpStatusCode.TooManyRequests;
+      if (!isRateLimited || attempt >= maxRetries) throw error;
+      await delay(baseDelay * 2 ** attempt);
+    }
+  }
+};
```
adilsitos: I believe we can remove it. The request already has a way of performing retries inside the library. You can check it on: `backend/src/lib/config/request.ts` (line 34)

If we want to have a custom one, you can define the client in a different way and pass a custom retry, but I don't think it this is required. 

Maybe this is also something we can discuss and tackle in a different time, I agree that this is a bit hidden of the implementation

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
```diff
@@ -0,0 +1,164 @@
...
+import { getDaytonaAuthHeaders } from "@app/services/app-connection/daytona";
+import { IntegrationUrls } from "@app/services/integration-auth/integration-list";
+import { SecretSyncError } from "@app/services/secret-sync/secret-sync-errors";
+import { matchesSchema } from "@app/services/secret-sync/secret-sync-fns";
+import { TSecretMap } from "@app/services/secret-sync/secret-sync-types";
+
+import { DAYTONA_SECRET_NAME_PATTERN, DAYTONA_SECRET_NAME_RULE } from "./daytona-sync-schemas";
+import { TDaytonaListSecretsResponse, TDaytonaSecret, TDaytonaSyncWithCredentials } from "./daytona-sync-types";
+
+const DAYTONA_PAGE_SIZE = 100;
```
adilsitos: I saw that the maximum is 200 [here](https://www.daytona.io/docs/openapi.json), maybe we could increase it? Or you see a problem with this?

backend/src/services/app-connection/daytona/daytona-connection-fns.ts
```diff
@@ -0,0 +1,67 @@
...
+  };
+};
+
+export const validateDaytonaConnectionCredentials = async (config: TDaytonaConnectionConfig) => {
+  const { apiKey } = config.credentials;
+
+  // Listing secrets is the capability every Daytona sync needs, so validating against it proves both
+  // that the key works and that it carries manage:secrets. Reading the key's own details would not.
+  try {
+    await request.get(`${IntegrationUrls.DAYTONA_API_URL}/secret/paginated`, {
```
adilsitos: Optional: if we want to easily support self host in the future, we would need to change the `request` to use a `safeRequest`. We could change this now as well, but this would be totally optional honestly

docs/integrations/secret-syncs/daytona.mdx
adilsitos: Checking how this works, I think it makes a lot of sense. Imagine a scenario that you want to do this, you would need to update each secret by hand (or create a script to update those). 

Maybe we can have this as an optional field, where if it is empty you don't add the allowed hosts (current behavior). We can also check this field to make sure that this is a host. I would also advice to check that we don't receive an IP (daytona alreayd validate this as well). You can use the `blockLocalAndPrivateIpAddresses` helper for this.

docs/integrations/app-connections/daytona.mdx
adilsitos, line 1: ignore comment placement. We should also add the docs related to the APIs, so users know that this can be created using only the infisical api. 

syncs for reference: https://infisical.com/docs/api-reference/endpoints/secret-syncs/1password/list
app connections for reference: https://infisical.com/docs/api-reference/endpoints/app-connections/1password/list

adilsitos (changes requested):

docs/integrations/app-connections/daytona.mdx
lobocv: Agreed, we should leave that out

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/DaytonaSyncFields.tsx
lobocv: I figured it would be helpful for the user to see the config and know that it belongs to a specific Daytona org. It's just for informational purposes because the org is encoded into the API key. If you don't think this is the right place to put it I can remove it. 
The other property this adds was that it allows us to check for duplicate configs, because otherwise the config is empty and we can't check duplication.

adilsitos: Hmm gotcha. One thing that you can do (if it is possible) is after the user select the the connection perform a request to daytona and show the options of orgs the user has. If this is possible. this can live inside the app-connection-fns file of daytona.

docs/integrations/secret-syncs/daytona.mdx
lobocv: I spoke with @varonix0 and we decided since we don't really have a solid pattern for per-secret provider-specific configuration, we will leave out the hosts for now and let the user set it in Daytona. We won't ever change what the user sets so the next sync won't remove their hosts

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/DaytonaSyncFields.tsx
lobocv: I decided to remove the Organization Name and disable secret sync dedupe for Daytona. It's not possible to get the list of organizations because the API key is scoped to one organization. So at best we would only return the current organization. The parameter was to enable deduplication but we can just disable it globally for Daytona.

docs/integrations/app-connections/daytona.mdx
lobocv: I added screenshots and docs for the app connection flow, see the screenshot in the PR description. Thanks!

backend/src/services/app-connection/daytona/daytona-connection-fns.ts
lobocv: Daniel and I agreed that we won't support hosts right now. They can configure hosts in daytona and we will respect it and never change it

docs/integrations/app-connections/daytona.mdx
lobocv: I tested it locally and the API docs are generating. I think it comes from the zod schema?

backend/src/services/secret-sync/daytona/daytona-sync-fns.ts
lobocv: Good catch! updated to 200

backend/src/lib/retry/index.ts
lobocv: Oh nice, then ill remove this!

docs/integrations/secret-syncs/daytona.mdx
```diff
@@ -0,0 +1,165 @@
...
+title: "Daytona sync"
+description: "Learn how to configure a Daytona sync for Infisical."
+---
+
+**Prerequisites:**
+- Set up and add secrets to [Infisical Cloud](https://app.infisical.com)
+- Create a [Daytona connection](/integrations/app-connections/daytona)
+
+<Note>
+    Daytona secret names must start with a letter or an underscore, and can contain only letters, digits, hyphens, and underscores. A sync that includes a key Daytona rejects fails and names the offending key, so no secrets are written. Use a **key schema** to reshape names that don't fit.
```
adilsitos: This is something I noticed on the first review and forgot to mention. In my opinion we can remove the `Use a **key schema** to reshape names that don't fit.` Because the user need to define a key schema in a specific way to make it work, and we can have some customer saying that they have key schemas and the sync is still failing.

backend/src/services/secret-sync/secret-sync-maps.ts
```diff
@@ -334,7 +338,12 @@ export const DESTINATION_DUPLICATE_CHECK_MAP: Record<SecretSync, DestinationDupl
   [SecretSync.HasuraCloud]: defaultDuplicateCheck,
   [SecretSync.Qovery]: defaultDuplicateCheck,
   [SecretSync.Cloud66]: defaultDuplicateCheck,
-  [SecretSync.Spacelift]: defaultDuplicateCheck
+  [SecretSync.Spacelift]: defaultDuplicateCheck,
+  // Daytona has an empty destination config, so checkDuplicateDestination returns early and never
+  // reaches this. Kept accurate rather than defaulted: a Daytona API key is bound to one organization
+  // and nothing scopes below it, so two syncs sharing a connection do write the same secrets.
+  [SecretSync.Daytona]: async ({ existingSync, newSync }) =>
```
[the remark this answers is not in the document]
adilsitos: @lobocv this is actually right and it is breaking the behavior. If you define the product setting that you can only have secrets into a specific path, this is bypassing that configuration. You can define a new map (i.e. syncs that don't have destination), otherwise this check is never called. 

Feel free to call me if necessary.

Cursor approach: 
```
Daytona’s destination config is an empty object — a Daytona API key is bound to one org, so the connection **is** the destination. The duplicate checker compared `connectionId` correctly, but two guards treated `{}` as “form not filled in yet” and returned before it ran:

- **Backend** `checkDuplicateDestination` early-returned on empty `destinationConfig`
- **Frontend** `useDuplicateDestinationCheck` required at least one config field before calling the API

Result: two Daytona syncs on the same connection were never flagged, including when “Block duplicate secret sync destinations” was on.

**Fix:** mark Daytona as connection-scoped. The empty-config short-circuit still applies to destinations that use the default checker. For Daytona, the backend runs the map (same `connectionId` = duplicate), and the frontend checks once a connection is selected.

Destinations with real config fields are unchanged — an incomplete form still does not fire the check.
```

docs/integrations/app-connections/daytona.mdx
adilsitos: it was missing the files on the `api-reference`. It is all good now 💯

backend/src/services/secret-sync/secret-sync-maps.ts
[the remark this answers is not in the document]
adilsitos: image
for reference. I was able to create a sync to the same destination, using the same app connection

docs/integrations/secret-syncs/daytona.mdx
lobocv: yeah that makes sense. Really they need to change their secret names to match what is accepted by in Daytona. By the key schema i'm guessing they mean the templating string. That's not going to be flexible enough to make a difference. Ill remove it

adilsitos (approved):

PR #7867, opened 2026-08-28
feat: allow one LDAP connection to back many Linux and Windows syncs

carlosmonastyrski:
## Context

Today, pushing certificates to a fleet of servers means creating one connection per machine, each holding its own copy of the same credential, and updating every one of them whenever that credential is rotated.

This change lets a single LDAP connection stand in for the whole domain instead: it supplies both the credential and the directory, so one connection can serve as many syncs as you have servers. Each sync simply names the machine it delivers to, choosing from a list read live from the directory or typing an address that is not listed, and every sync shares the one credential. Rotate the service account's password once on the connection and every sync picks it up on its next run, with nothing else to touch.

Existing SSH and WinRM connections keep working exactly as before, and the new option only appears for Linux Server and Windows Server syncs when an LDAP connection is selected.

## Type

- [ ] Fix
- [x] Feature
- [x] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


carlosmonastyrski:
@veria-ai please review this PR

carlosmonastyrski:
@veria-ai please review this PR

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4549,6 +4549,8 @@ interface CreatePkiSyncEvent {
     destination: string;
     applicationId?: string;
     connectionId?: string;
+    connectionName?: string;
```
bernie-g, line 4552: ignore comment placement

Image

Do we plan on supporting other LDAP providers in the future? Otherwise this could just be hardcoded to active directory and we rename the entire feature to AD

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4549,6 +4549,8 @@ interface CreatePkiSyncEvent {
...
+    targetHost?: string;
     hasCredentials?: boolean;
     hasPostSyncCommand?: boolean;
```
bernie-g, line 4555: ignore comment placement

Image
Can we validate that the target host is reachable when saving? For example I used the test AD instance with a gateway running on my local machine and it suggested the two domains that are only reachable through the AD's DNS so they both fail when running the sync but I'm able to select and save either of them. I think it'd be better UX for the user if we test reachability at save time

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -6,6 +6,7 @@ import { promisify } from "util";
 import { z } from "zod";
 
 import { BadRequestError } from "@app/lib/errors";
+import { netbiosFromDomainFqdn } from "@app/lib/ldap/ldap-search-fns";
 
 import { GcpServiceAccountAuthMethod, PamAccountType, PamPostgresAuthMethod, PamSshAuthMethod } from "../pam/pam-enums";
 import { getApplicablePolicies, PamPolicyDescriptorSchema } from "../pam/pam-policies";
```
bernie-g, line 12: ignore comment placement

Image

If the target host is unreachable can we make this error a bit better? Something like `Could not resolve the target host "EC2AMAZ-8V4780K.corp.example.com" from Gateway "test".` Also ignore the screenshot looking weird, I think that's CleanShot being wonky

backend/src/ee/services/audit-log/audit-log-types.ts
bernie-g: Oh nvm I didn't realize LDAP was an existing app connection type. I think in that case we need to add checks in this PR to make sure the provider type is active directory in case we add other LDAP providers in the future

backend/src/lib/ldap/ldap-search-fns.test.ts
```diff
@@ -0,0 +1,188 @@
+import { EventEmitter } from "node:events";
+
+import {
```
bernie-g, line 3: From claude:

Permission bypass via connection swap (pki-sync-service.ts:253): changing only connectionId between two LDAP connections retargets delivery to a different machine's domain without set-target-host; the gate only compares
     host/port/ssl. The PR's own command gating already treats a connection swap as retargeting.

backend/src/lib/ldap/ldap-search-fns.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  buildDomainBaseDN,
+  escapeLdapFilterValue,
+  getLdapAttribute,
+  getLdapAttributeBuffer,
+  netbiosFromDomainFqdn,
+  searchLdap
+} from "./ldap-search-fns";
```
bernie-g, line 10: From claude:

  3. No aggregate concurrency bound (keystore.ts:61): the 3-job cap moved to per-(connection, host); a fleet-wide renewal can push hundreds of concurrent sessions through one gateway and DC.

bernie-g (changes requested):

frontend/src/components/pki-syncs/forms/schemas/base-pki-sync-schema.ts
```diff
@@ -9,6 +11,37 @@ export const HostCommandSchema = z
...
+  .optional();
+
+export const PkiSyncTargetHostSchema = z
+  .string()
+  .trim()
+  .max(253, "Target host is too long")
+  .refine(
+    (value) =>
+      !value ||
+      /^[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/.test(
```
bernie-g, line 34: Can we use RE2 here please

carlosmonastyrski: RE2 is only for backend, closing this comment

backend/src/ee/services/permission/default-roles.ts
```diff
@@ -871,7 +872,8 @@ const buildApplicationAdminPermissionRules = () => {
       ResourcePermissionPkiSyncActions.ImportCertificates,
       ResourcePermissionPkiSyncActions.RemoveCertificates,
       ResourcePermissionPkiSyncActions.SetPostSyncCommand,
-      ResourcePermissionPkiSyncActions.SetHealthCheckCommand
+      ResourcePermissionPkiSyncActions.SetHealthCheckCommand,
+      ResourcePermissionPkiSyncActions.SetTargetHost
```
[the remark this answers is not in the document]
carlosmonastyrski: Intended

bernie-g (approved):

PR #7875, opened 2026-08-28
fix: collapse signer approvals into a single setup card until a policy exists

carlosmonastyrski:
## Context

When a signer has no approval policy and no requests, the Approvals tab now shows one card with an "Add Policy" button instead of two cards whose only affordance was an edit pencil. Once a policy or any request exists, the usual two-card split returns unchanged.

## Type

Open a signer with no policy, go to Approvals, click "Add Policy", add a step and save. Both cards appear.

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/pages/cert-manager/SignerDetailPage/components/SignerApprovalPolicyTab.tsx
+++ b/frontend/src/pages/cert-manager/SignerDetailPage/components/SignerApprovalPolicyTab.tsx
@@ -1,10 +1,15 @@
 import { useLayoutEffect, useMemo, useRef, useState } from "react";
-import { PencilIcon, UsersIcon } from "lucide-react";
+import { PencilIcon, PlusIcon, UsersIcon } from "lucide-react";
 
 import {
+  Button,
   Card,
   CardContent,
+  CardDescription,
+  CardHeader,
+  CardTitle,
   Empty,
+  EmptyContent,
   EmptyDescription,
   EmptyHeader,
   EmptyTitle,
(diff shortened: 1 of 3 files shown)


frontend/src/pages/cert-manager/SignerDetailPage/components/SignerApprovalPolicyTab.tsx
scott-ray-wilson, line 1: unrelated to your changes but we should have a badge here, it says 1 on the sidebar but it doesn't tell me which tab so I was confused at first when I saw an empty table 

Image

scott-ray-wilson, line 1: unrelated to your changes but creating a policy says edit in the sheet header

Image

scott-ray-wilson, line 1: unrelated to your changes but we should hide table headers when empty state 

Image

scott-ray-wilson, line 1: not sure if this scenario even has an actual path but if so, when there are requests but no policy I would say we should do the button full width instead of the empty component (keep it how it is for the stand alone version though, just think it doesn't look good squished)

Image

scott-ray-wilson (approved):

frontend/src/pages/cert-manager/SignerDetailPage/components/SignerApprovalPolicyTab.tsx
carlosmonastyrski: Addressing these in a separate PR 👍

PR #7876, opened 2026-08-28
fix(frontend): keep empty value out of key when pasting env vars

eeshsaxena:
## Context

When pasting env content into the Create Secret form, `getKeyValue` splits each line into a key and value on the first delimiter. It returned the **whole string as the key** whenever the delimiter was the last character, because `hasValueAfterDelimiter` was false in that case and the function fell back to `{ key: pastedContent.trim(), value: "" }`.

So an empty-valued line — which is common in `.env` files — was mis-parsed:

```
"KEY="    -> { key: "KEY=", value: "" }     (before)   ❌
"TOKEN:"  -> { key: "TOKEN:", value: "" }   (before)   ❌
```

Pasting `KEY=` created a secret literally named `KEY=` (delimiter included) instead of `KEY` with an empty value.

The fix drops the `hasValueAfterDelimiter` fallback: split on the delimiter whenever one is present, and only treat the whole string as the key when there is no delimiter at all. A trailing delimiter then yields an empty value, as expected.

```
"KEY="    -> { key: "KEY", value: "" }       (after)   ✔
"KEY=v"   -> { key: "KEY", value: "v" }       (unchanged)
"A=B=C"   -> { key: "A", value: "B=C" }        (unchanged, first delimiter wins)
"KEY"     -> { key: "KEY", value: "" }         (unchanged, no delimiter)
```

## Steps to verify the change

In the Secrets Overview, open Create Secret and paste a line with an empty value, e.g. `EMPTY=`. The key field is `EMPTY` (not `EMPTY=`) and the value is empty. Added unit tests in `frontend/src/helpers/parseEnvVar.test.ts` covering trailing-delimiter, no-delimiter, multi-delimiter, and empty-input cases.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally (ran `getKeyValue` against the cases above; results match the tests).

_Disclosure: this change was written with a coding agent; I reviewed it and can explain every line._

### DIFF
--- a/frontend/src/helpers/parseEnvVar.test.ts
+++ b/frontend/src/helpers/parseEnvVar.test.ts
@@ -1,0 +1,32 @@
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+
+import { getKeyValue } from "./parseEnvVar";
+
+describe("getKeyValue", () => {
+  const delimiters = ["=", ":"];
+
+  it("splits a key and value on the first delimiter", () => {
+    assert.deepEqual(getKeyValue("KEY=value", delimiters), { key: "KEY", value: "value" });
+    assert.deepEqual(getKeyValue("TOKEN: abc", delimiters), { key: "TOKEN", value: "abc" });
+  });
+
+  it("keeps only the first delimiter as the split point", () => {
+    assert.deepEqual(getKeyValue("A=B=C", delimiters), { key: "A", value: "B=C" });
+  });
+
+  it("returns an empty value when the delimiter is at the end", () => {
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Small but user-facing: empty-valued lines are common in pasted .env files, and today they create a secret whose key includes the trailing = or : rather than an empty value. The fix only changes the trailing-delimiter case; all other splits (normal key=value, first-delimiter-wins for key=a=b, and no-delimiter keys) are unchanged and covered by the added tests.

PR #7879, opened 2026-08-29
feat(pam): account password viewing with access requests

lb-vn:
## Context

Adds the ability to view account passwords and also integrates with our approval workflows. You can request approval for viewing a password if you're an operator. Also adds the new "Operator" role, which lets you view password.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -22,6 +22,7 @@ import { readLimit, writeLimit } from "@app/server/config/rateLimiter";
 import { slugSchema } from "@app/server/lib/schemas";
 import { getTelemetryDistinctId } from "@app/server/lib/telemetry";
 import { withRoutePrefix } from "@app/server/lib/with-route-prefix";
+import { isUserSessionAuth } from "@app/server/plugins/auth/inject-identity";
```
sheensantoscapadngan, line 26: IGNORE PLACEMENT

what is this auth method field? do we need this here? 
Image

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -745,6 +751,71 @@ export const registerPamAccountRouter = async (server: FastifyZodProvider) => {
     }
   });
 
+  server.route({
+    method: "POST",
```
sheensantoscapadngan, line 770: isn't this supposed to be a GET request by convention?

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -745,6 +751,71 @@ export const registerPamAccountRouter = async (server: FastifyZodProvider) => {
...
+          .max(64)
+          .optional()
+          .describe("A verified MFA session ID. Required when the template requires MFA.")
+      }),
+      response: {
+        200: z.object({
+          accountType: z.string().describe("The account's platform type"),
+          credentials: z.record(z.unknown()).describe("The account's stored credentials, secrets included")
+        })
+      }
```
sheensantoscapadngan, line 792: shouldn't we be using a discriminatedUnion here? that way it's very clear what the shape of the return is going to be for each account type?

backend/src/ee/routes/v1/pam-routers/pam-resource-role-router.ts
```diff
@@ -13,6 +13,12 @@ const DEFAULT_RESOURCE_ROLES = [
     isDefault: true,
     description: "Full control of accounts, folders, sessions, and memberships"
   },
+  {
+    slug: PamResourceRole.Operator,
+    name: "Operator",
+    isDefault: true,
+    description: "Launch sessions and view account credentials; cannot approve requests or change configuration"
```
sheensantoscapadngan: unsure if we should even mention the approve requests part here...seems like we can just keep it as "Launch sessions and view account credentials" ?

backend/src/ee/routes/v1/pam-routers/pam-resource-role-router.ts
```diff
@@ -13,6 +13,12 @@ const DEFAULT_RESOURCE_ROLES = [
     isDefault: true,
     description: "Full control of accounts, folders, sessions, and memberships"
   },
+  {
+    slug: PamResourceRole.Operator,
```
sheensantoscapadngan, line 17: IGNORE PLACEMENT

We're still showing an error here when the account requires MFA to view the credentials...
Image

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -745,6 +751,71 @@ export const registerPamAccountRouter = async (server: FastifyZodProvider) => {
...
+        projectId: req.internalPamProjectId,
+        actorEmail: isUserSessionAuth(req.auth) ? (req.auth.user.email ?? "") : "",
+        reason: req.body.reason,
+        mfaSessionId: req.body.mfaSessionId,
+        actorId: req.permission.id,
+        actor: req.permission.type,
+        actorOrgId: req.permission.orgId,
+        actorAuthMethod: req.permission.authMethod
+      });
+
```
sheensantoscapadngan, line 809: Have we considered just merging these for simplicity?
Image

couldn't we assume that if the account itself is sensitive enough that connecting to it requires external approval, then we would need the same guardrails for viewing the credentials?

I think we made the same assumption here regarding MFA

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -745,6 +751,71 @@ export const registerPamAccountRouter = async (server: FastifyZodProvider) => {
...
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        orgId: req.permission.orgId,
+        projectId: req.internalPamProjectId,
```
sheensantoscapadngan, line 813: IGNORE PLACEMENT

should we be more specific here with regards to the difference between request pending and credential request pending? 
Image

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -745,6 +751,71 @@ export const registerPamAccountRouter = async (server: FastifyZodProvider) => {
...
+        event: {
+          type: EventType.PAM_ACCOUNT_CREDENTIALS_VIEW,
```
sheensantoscapadngan, line 815: IGNORE PLACEMENT

I think it would make sense for us to also show what kind of request it is from the table view. 
Image

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -1330,3 +1330,24 @@ export const isCredentialConfigured = (
     return typeof value === "string" && value.trim().length > 0;
   });
 };
+
+export const noRevealableCredentialMessage = (accountName: string) =>
+  `Account '${accountName}' has no stored credential to reveal. Its authentication method brokers access per session instead.`;
+
```
sheensantoscapadngan, line 1347: error with SSH and cert auth

Image

Image

Image

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
sheensantoscapadngan: same remark for this view here
Image

docs/documentation/platform/pam/concepts/access-control.mdx
```diff
@@ -48,9 +48,14 @@ Once someone is a product member, they still need to be granted access to specif
 | Role | What they can do |
 |------|------------------|
 | **Admin** | Full control — accounts, folders, sessions, memberships |
+| **Operator** | Launch sessions and [view account credentials](/documentation/platform/pam/accounts/viewing-credentials) |
 | **Connector** | Launch sessions and connect to accounts |
 | **Auditor** | View audit logs and session recordings |
 
+<Note>
+  **Operator** is Connector plus the right to read an account's stored password or key. Grant it only where someone genuinely needs the credential in hand, since a revealed credential leaves no session recording. An Operator cannot approve requests or change configuration, so an account gated with **Require Approval for Credentials** still needs someone else to sign off.
```
sheensantoscapadngan: same remark here regarding "approve request" framing

docs/docs.json
```diff
@@ -746,6 +746,7 @@
                     "group": "Accounts",
                     "pages": [
                       "documentation/platform/pam/accounts/overview",
+                      "documentation/platform/pam/accounts/viewing-credentials",
                       "documentation/platform/pam/accounts/postgresql",
```
sheensantoscapadngan, line 697: Does it make sense to have a dedicated page for viewing account credentials? We don't have one for accessing sessions, and so it might be that we can just embed this directly into the individual pages instead.

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
lb-vn: Not necessarily, we just dump everything from credentials into that modal. It could be useful as information though, if you can something like SSH private key as the auth method.

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
lb-vn: It's a POST to not break single-use MFA tokens. Also the "reason" argument can go into a body instead of a query parameter where it can be logged.

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
lb-vn: Yeah I guess we can merge it for now, good idea

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
lb-vn: Renamed first one to "Access Request Pending"

backend/src/ee/services/pam-account/pam-account-schemas.ts
lb-vn: Intentional, there's nothing to show

sheensantoscapadngan: Perhaps we don't show the view credentials then?

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -83,6 +87,23 @@ const SanitizedAccountDetailSchema = z.discriminatedUnion(
...
+);
+
+const PamAccountCredentialsSchema = z.discriminatedUnion(
+  "accountType",
+  accountCredentialVariants as [
+    (typeof accountCredentialVariants)[number],
+    ...(typeof accountCredentialVariants)[number][]
+  ]
+);
+
```
sheensantoscapadngan, line 106: IGNORE PLACEMENT

shouldn't we then mention "viewing credentials" here in the description as well now that we've consolidated it? 
Image

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -745,6 +766,71 @@ export const registerPamAccountRouter = async (server: FastifyZodProvider) => {
...
+            reason: result.reason,
+            grantExpiresAt: result.grantExpiresAt?.toISOString() ?? null
+          }
+        }
+      });
+
+      return { accountType: result.accountType, credentials: result.credentials } as unknown as z.infer<
+        typeof PamAccountCredentialsSchema
+      >;
+    }
```
sheensantoscapadngan, line 831: IGNORE PLACEMENT

There isn't really too much distinction between the view for requesting the credentials and requesting for session access. Should we make this clearer? 

Image
Image

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -1330,3 +1330,35 @@ export const isCredentialConfigured = (
     return typeof value === "string" && value.trim().length > 0;
   });
 };
+
```
sheensantoscapadngan, line 1333: iGNORE PLACEMENT

I just noticed now that triggering view credentials with MFA will result in email text that mentions sign in...

Image
Image

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -81,6 +88,11 @@ type TPamAccountServiceFactoryDep = {
   pamSessionDAL: Pick<TPamSessionDALFactory, "find" | "update">;
   pamDiscoverySourceDAL: Pick<TPamDiscoverySourceDALFactory, "find">;
   userDAL: Pick<TUserDALFactory, "findById">;
+  orgDAL: Pick<TOrgDALFactory, "findOrgById">;
```
sheensantoscapadngan, line 91: IGNORE PLACEMENT

What do you mean here by "outside a recorded session"? The reveal credential operation will result in an audit log record, though, no? Could we somehow adjust the description here to be clearer?

Image

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -81,6 +88,11 @@ type TPamAccountServiceFactoryDep = {
...
+  mfaSessionService: Pick<
+    TMfaSessionServiceFactory,
+    "createMfaSession" | "getMfaSession" | "deleteMfaSession" | "sendMfaCode"
+  >;
```
sheensantoscapadngan, line 95: IGNORE PLACEMENT

When I tried revoking an approved credential access, it tells me that any session using it will be terminated immediately, which is only really valid for session access...
Image

backend/src/ee/services/pam-account/pam-account-service.ts
lb-vn: I think we should leave that. This literally implies the session won't be recorded. An audit log does not tell you anything about the session. We can't even tell when the session even starts, since the user can use the credential at any time

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
lb-vn: Check slack DMs

frontend/src/pages/pam/components/RequestAccessSheet.tsx
```diff
@@ -275,6 +304,11 @@ export const RequestAccessSheet = ({
                           ))}
                         </SelectContent>
                       </Select>
+                      <FieldDescription>
```
sheensantoscapadngan, line 307: hmmm do we have to show "Rotating the account is what actually cuts off a copy" to the person asking for credentials?

sheensantoscapadngan (approved):

PR #7880, opened 2026-08-29
fix: cert dashboard and signer approval policy improvements

carlosmonastyrski:
## Context

Fixes a set of small UI issues across cert manager:

- Approval Requests tabs now show a pending count each, so the sidebar badge points at the right tab.
- The approval policy sheet says "Add approval policy" when there is no policy yet, instead of "Edit".
- Signing History hides its table headers when there is no activity.
- When a signer has requests but no policy, the Add Policy button goes full width instead of sitting squished in an empty state.
- Dashboard distribution charts move to a 2x2 grid, cap at 5 categories plus "Other", and no longer truncate labels.
- Slices under one percent read `<1%` instead of `0%`.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [x] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/frontend/src/pages/cert-manager/DashboardPage/components/DistributionCharts.tsx
+++ b/frontend/src/pages/cert-manager/DashboardPage/components/DistributionCharts.tsx
@@ -1,4 +1,5 @@
 import { Cell, Pie, PieChart, ResponsiveContainer, Tooltip as RechartsTooltip } from "recharts";
+import { twMerge } from "tailwind-merge";
 
 import {
   Card,
@@ -16,7 +17,12 @@ import {
 import type { TDashboardDistribution, TDashboardStats } from "@app/hooks/api/certificates";
 import { certKeyAlgorithmToNameMap } from "@app/hooks/api/certificates/constants";
 
-import { CHART_COLORS, CHART_COLORS_HEX } from "./chart-theme";
+import {
+  CHART_COLORS_HEX,
+  CHART_OTHER_COLOR_HEX,
+  formatShare,
+  MAX_DONUT_SEGMENTS
+} from "./chart-theme";
 
 type Props = {
   stats: TDashboardStats;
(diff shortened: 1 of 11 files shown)


frontend/src/pages/cert-manager/DashboardPage/components/DistributionCharts.tsx
scott-ray-wilson, line 1: nit but I feel like we may want to bump out the breakpoint one more width as these feel pretty narrow

Image

frontend/src/pages/cert-manager/DashboardPage/components/ExpirationTimeline.tsx
scott-ray-wilson, line 85: should this chart also use the `colorAt` fix? seems mismatched 

Image

frontend/src/pages/cert-manager/DashboardPage/components/chart-theme.tsx
```diff
@@ -25,6 +25,17 @@ export const CHART_COLORS_HEX = [
   "#b8a060" // olive gold
 ];
 
+export const MAX_DONUT_SEGMENTS = 6;
+
+export const CHART_OTHER_COLOR = "var(--color-muted)";
+
+export const formatShare = (count: number, total: number) => {
```
scott-ray-wilson, line 32: should we also use this for:
-  `PqcReadinessChart.tsx:114` + `:129`
-  `ValidityReadinessSection.tsx:295` + `:313`
-   `KpiCards.tsx:63`

also should we fix rounding on the top as well? 499/500 rounds to 100% so should we say >99%

scott-ray-wilson (changes requested):

scott-ray-wilson (approved):
LGTM

PR #7887, opened 2026-08-30
feat: set project retention limits

varonix0:
## Context

Added support for configuring project retention limits in the regular project update endpoint.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/server/routes/v1/project-router.ts
+++ b/backend/src/server/routes/v1/project-router.ts
@@ -516,7 +516,8 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
           .array(z.string())
           .optional()
           .describe(PROJECTS.UPDATE.secretDetectionIgnoreValues),
-        pitVersionLimit: z.number().min(1).max(100).optional()
+        pitVersionLimit: z.number().min(1).max(100).optional(),
+        auditLogsRetentionDays: z.number().int().min(0).optional().describe(PROJECTS.UPDATE.auditLogsRetentionDays)
       }),
       response: {
         200: z.object({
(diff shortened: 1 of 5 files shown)


backend/src/services/project/project-service.ts
```diff
@@ -943,6 +944,35 @@ export const projectServiceFactory = ({
...
+      if (!hasRole(ProjectMembershipRole.Admin)) {
+        throw new ForbiddenRequestError({
+          message: "Only project admins can update the audit logs retention period"
+        });
+      }
+
+      if (appCfg.isCloud) {
+        throw new BadRequestError({
+          message: "The audit logs retention period can not be updated on Infisical Cloud instances."
+        });
```
mathnogueira, line 957: Is there a reason we don't allow this in Cloud?

backend/src/server/routes/v1/project-router.ts
```diff
@@ -516,7 +516,8 @@ export const registerProjectRouter = async (server: FastifyZodProvider) => {
           .array(z.string())
           .optional()
           .describe(PROJECTS.UPDATE.secretDetectionIgnoreValues),
-        pitVersionLimit: z.number().min(1).max(100).optional()
+        pitVersionLimit: z.number().min(1).max(100).optional(),
+        auditLogsRetentionDays: z.number().min(0).max(365).optional().describe(PROJECTS.UPDATE.auditLogsRetentionDays)
```
mathnogueira: Image

In UI we don't allow retention to be `0`, we should have the same here as it can disable audit logs.

```suggestion
        auditLogsRetentionDays: z.number().min(1).max(365).optional().describe(PROJECTS.UPDATE.auditLogsRetentionDays)
```

mathnogueira (changes requested):

mathnogueira (approved):

PR #7893, opened 2026-08-31
fix(webhook): render the test event for the webhook's type

### LINKED ISSUE #7881: Webhook Test button always sends General payload; Slack-type webhooks fail with 400 no_text
### Summary

The **Test** button on a project webhook of type **Slack** (Project → Settings → Webhooks) always fails with HTTP 400 from Slack, even though the same webhook delivers real secret-change events correctly. The test payload is built without regard to the webhook type, so a Slack incoming-webhook URL receives a body with no `text` field and rejects it with `no_text`.

### Steps to reproduce

1. Create a Slack incoming webhook (`https://hooks.slack.com/services/...`).
2. Infisical Cloud → project → Settings → Webhooks → Add webhook, **type: Slack**, environment `prod`, secret path `/`, paste the URL.
3. Click **Test**.

### Expected

Test succeeds and a test message appears in the Slack channel.

### Actual

Test fails with status code 400. Slack's response body is `no_text`.

Modifying a secret under the configured path **does** post to Slack correctly ("A secret value has been added or modified." with the attachment fields), so only the test path is broken.

### Cause

In `backend/src/services/webhook/webhook-fns.ts`, `getWebhookPayload` switches on `type` for real events (e.g. the `WebhookType.SLACK` branch returns `{ text, attachments }`):

https://github.com/Infisical/infisical/blob/21a8841267b41f4089aee0f488f2ee36e8aed49c/backend/src/services/webhook/webhook-fns.ts#L73-L119

but the `TestEvent` branch has no type switch and always returns the General-shaped body `{ event, project: { workspaceId, projectId, projectName, environment, environmentName, secretPath } }`:

https://github.com/Infisical/infisical/blob/21a8841267b41f4089aee0f488f2ee36e8aed49c/backend/src/services/webhook/webhook-fns.ts#L580-L593

Slack incoming webhooks require `text` (or `blocks`/`attachments`), so that body is rejected. The same issue presumably affects `WebhookType.MICROSOFT_TEAMS` (Teams expects an Adaptive Card envelope).

Verified by replaying both bodies with curl against the same Slack URL: the Test-event body → `400 no_text`; the `SecretModified` Slack-type body → `200 ok`.

### Suggested fix

Give the `TestEvent` branch the same `switch (type)` as the other events, returning a minimal Slack payload (e.g. `{ text: "Infisical webhook test", attachments: [{ fields: [Project, Environment, Secret Path] }] }`) and a Teams Adaptive Card for `MICROSOFT_TEAMS`, falling through to the current General body otherwise.

### Environment

- Infisical Cloud (US), Pro plan, tested 2026-08-29
- Webhook type: Slack; Slack incoming webhook URL

lobocv:
## Context

The **Test** button on a project webhook always sent the General JSON body, whatever the webhook's type. Slack rejects that body with `400 no_text` because an incoming webhook needs `text` (or `blocks`/`attachments`), so the test failed on Slack webhooks that delivered real secret-change events correctly. Microsoft Teams has the same gap: it expects an adaptive card envelope.

The cause was in `getWebhookPayload` (`backend/src/services/webhook/webhook-fns.ts`). Every other event switches on the webhook type; only the `TestEvent` branch ignored it and returned the General shape. The service already passed `type: webhook.type` into the payload, so nothing else had to change.

Test bodies are now built per webhook type: a Slack body with `text` and one attachment, a Teams adaptive card with the same facts, and the unchanged General body. The builders live in a `Record<WebhookType, ...>` rather than a `switch`, so adding a webhook type fails the type check until it renders a body its provider accepts, and a unit test asserts the same at runtime. An unrecognised or null `type` still falls back to the General body, since `webhooks.type` is a nullable column.

The General body is unchanged, so nothing changes for existing General webhooks.

Closes ENG-5630. Fixes #7881.

## Steps to verify the change

1. Create a Slack [incoming webhook](https://api.slack.com/messaging/webhooks).
2. In a project, go to **Settings > Webhooks > Add webhook**, pick type **Slack**, set an environment and secret path, and paste the URL.
3. Click **Test**. The test succeeds and the message "This is a test message from Infisical." appears in the Slack channel with Project, Environment, Environment Name, and Secret Path fields.
4. Repeat with a **Microsoft Teams** webhook. The test posts an adaptive card carrying the same facts.
5. Repeat with a **General** webhook pointed at any HTTPS endpoint. The received body is unchanged: `{ "event": "test", "project": { ... }, "timestamp": ... }`.
6. `cd backend && npm run test:unit -- webhook-fns`

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/services/webhook/webhook-fns.test.ts
+++ b/backend/src/services/webhook/webhook-fns.test.ts
@@ -1,6 +1,6 @@
 import { ActorType } from "@app/services/auth/auth-type";
 
-import { getWebhookPayload, isWebhookPathSubscribed } from "./webhook-fns";
+import { getWebhookPayload, isWebhookPathSubscribed, TEST_EVENT_PAYLOAD_BUILDERS } from "./webhook-fns";
 import { AccessRequestWebhookAction, ChangeRequestWebhookAction, WebhookEvents, WebhookType } from "./webhook-types";
 
 const projectFields = {
@@ -402,6 +402,18 @@ describe("getWebhookPayload: secrets.access-request.modified", () => {
     }) as { attachments: { content: { body: { type: string }[] } }[] };
 
     expect(result.attachments[0].content.body.map((b) => b.type)).toEqual(["TextBlock", "FactSet"]);
+  });
+});
+
+describe("getWebhookPayload: test", () => {
+  test("every webhook type renders a test event body", () => {
+    expect(Object.keys(TEST_EVENT_PAYLOAD_BUILDERS).sort()).toEqual([...Object.values(WebhookType)].sort());
+  });
+
+  test("slack body carries the text field the incoming webhook API requires", () => {
+    expect(
+      getWebhookPayload({ type: WebhookEvents.TestEvent, payload: { ...projectFields, type: WebhookType.SLACK } })
+    ).toHaveProperty("text");
   });
 });
 
--- a/backend/src/services/webhook/webhook-fns.ts
+++ b/backend/src/services/webhook/webhook-fns.ts
@@ -78,7 +78,7 @@ type TTestEventFields = {
   secretPath?: string;
 };
 
-const TEST_EVENT_MESSAGE = "This is a test message from Infisical.";
+const TEST_EVENT_MESSAGE = "This is a test message from Infisical webhooks.";
 
 // Keyed by webhook type rather than switched on, so a new WebhookType fails the type check here
 // until it has a body its provider accepts. Slack rejects a body without `text` with 400 no_text.
--- a/backend/src/services/webhook/webhook-fns.ts
+++ b/backend/src/services/webhook/webhook-fns.ts
@@ -70,6 +70,91 @@ export const triggerWebhookRequest = async (
   return req;
 };
 
+type TTestEventFields = {
+  projectId: string;
+  projectName?: string;
+  environment: string;
+  environmentName: string;
+  secretPath?: string;
+};
+
+const TEST_EVENT_MESSAGE = "This is a test message from Infisical.";
+
+// Keyed by webhook type rather than switched on, so a new WebhookType fails the type check here
+// until it has a body its provider accepts. Slack rejects a body without `text` with 400 no_text.
+export const TEST_EVENT_PAYLOAD_BUILDERS: Record<WebhookType, (fields: TTestEventFields) => Record<string, unknown>> = {
+  [WebhookType.SLACK]: ({ projectName, environment, environmentName, secretPath }) => ({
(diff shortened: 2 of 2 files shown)


backend/src/services/webhook/webhook-fns.ts
```diff
@@ -70,6 +70,91 @@ export const triggerWebhookRequest = async (
...
 
+type TTestEventFields = {
+  projectId: string;
+  projectName?: string;
+  environment: string;
+  environmentName: string;
+  secretPath?: string;
+};
+
+const TEST_EVENT_MESSAGE = "This is a test message from Infisical.";
```
adilsitos: nit: Maybe we could say that this is coming from the webhook? "This is a test message from Infisical webhook"

backend/src/services/webhook/webhook-fns.test.ts
```diff
@@ -405,6 +405,18 @@ describe("getWebhookPayload: secrets.access-request.modified", () => {
...
+  test("every webhook type renders a test event body", () => {
+    expect(Object.keys(TEST_EVENT_PAYLOAD_BUILDERS).sort()).toEqual([...Object.values(WebhookType)].sort());
+  });
+
+  test("slack body carries the text field the incoming webhook API requires", () => {
+    expect(
+      getWebhookPayload({ type: WebhookEvents.TestEvent, payload: { ...projectFields, type: WebhookType.SLACK } })
+    ).toHaveProperty("text");
+  });
+});
```
adilsitos, line 418: doesn't it make sense to have a test that check that in case we don't sne da type (slack or team) the general one is triggered?

adilsitos (commented):
Just two nits.

backend/src/services/webhook/webhook-fns.ts
lobocv: Sure I can make that change!

backend/src/services/webhook/webhook-fns.test.ts
lobocv: That's a good idea. I'd prefer if we make the test fail if there is an unhandled event because that means we didn't code it right. Every webhook event should have handling code.

lobocv: Okay it turns out we can't make a good quality test that checks against this without hard-coding logic about every event type. At that point the test becomes worthless. This is further evidence how the code is poorly structured. 
We're going to move this all to shared alerting module soon so let's defer that

adilsitos (approved):

PR #7902, opened 2026-08-31
improvement(secret-syncs): modernize forms and harden destination selection

andrewhuhh:
## Context

Related ticket: PLATFOR-750

Modernizes Secret Sync setup and edit flows around the v3 component system and fixes destination-selection edge cases exposed by the migration.

Previously, destination forms relied on legacy React Select controls, discard behavior varied by exit path, and invalid fields often displayed an error without visually marking the control.

### Forms and shared components

- Migrate Secret Sync destination selectors to the v3 Combobox, with grouped options, server-side search, search keywords, rich option rendering, and saved-selection handling. Keep Vercel custom preview branches on CreatableSelect because Combobox does not support inline creation.
- Extend the shared Combobox to support these flows, including independently configurable missing-selection inclusion and select-all behavior for visible results.
- Standardize dirty-form tracking and discard confirmation across create, edit, sheet-close, and provider-reset paths. Revalidate create-form fields as they change.
- Add invalid styling to Secret Sync, App Connection, and related provider fields; add isError and aria-invalid support to SecretInput. Associate migrated controls with their visible labels, descriptions, and validation errors, and keep multi-select accessible names independent of search hints.
- Normalize optional destination values before submission and duplicate checks, and clear dependent selections when parent resources change.
- Use semantic v3 tokens for Secret Sync headers, status badges, details, and explanatory content. Add the shared inline Code component and adjust short Select menu sizing.

### Provider errors and duplicate detection

- Propagate GitLab rate limits for group/project lookups so the UI can distinguish them from empty results, including exhausted Gitbeaker retries.
- Throw explicit rate-limit errors when Checkly and Netlify retry budgets are exhausted.
- Guard Vercel retry handling against errors without an HTTP response.
- Treat empty and omitted optional scopes equivalently in Bitbucket, TeamCity, and Qovery duplicate-destination comparisons.
- Add focused tests for selected-item merging, accessible-name selection, destination schemas, duplicate comparisons, and provider error handling. Run frontend regression tests with `npm test` in CI; use leaf imports so schema tests do not load UI/CSS dependencies. Add a Storybook interaction check for multi-select label focus and naming.

## Screenshots

Secret Sync setup flow

## Steps to verify the change

1. Open a Secret Management project and navigate to **Integrations > Secret Syncs**. Exercise create and edit flows with single-select, multi-select, grouped, and server-searched destination fields.
2. Verify Vercel project grouping and search, saved selections outside the current result page, and custom preview branch entry. Verify AWS region search accepts both region names and slugs.
3. Select dependent resources, move Next and Back, and change the parent connection/account/repository. Verify obsolete child selections are cleared and the displayed values match the submitted configuration.
4. Select and clear optional Checkly, Netlify, and Travis CI fields. Verify valid submission and equivalent duplicate warnings for cleared versus omitted optional scopes.
5. Submit representative Secret Sync, App Connection, and Dynamic Secret forms with missing required fields. Verify invalid styling and that corrected fields clear their errors.
6. Verify untouched create/edit sheets close without prompting, while closing, canceling, or returning to provider selection after editing asks for discard confirmation. Verify Keep Editing preserves the form.
7. Verify keyboard selection, search, clear, and Escape behavior in comboboxes inside sheets, including grouped and multi-select menus.
8. Verify provider rate limits surface as errors rather than empty resource lists where changed, and Vercel errors without HTTP responses preserve the original failure path.
9. From `frontend`, run `npm test`, `npm run type:check`, and `npm run lint`. Run the focused backend regression tests added by this PR.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format: type(scope): short description.
- [x] Tested locally


frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AwsParameterStoreSyncFields.tsx
scott-ray-wilson, line 1: not sure of specific to this individual usage or combobox generic but the border doesn't display error states like the input does 

Image

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AwsParameterStoreSyncFields.tsx
```diff
@@ -68,15 +68,15 @@ export const AwsParameterStoreSyncFields = () => {
...
                     /demo/path/
                   </code>{" "}
                   and a key schema of{" "}
-                  <code className="rounded-sm bg-mineshaft-600 px-0.5 py-px text-sm text-mineshaft-300">
+                  <code className="rounded-sm bg-ring px-0.5 py-px text-sm text-label">
                     INFISICAL_{"{{secretKey}}"}
                   </code>
                   , then the result will be{" "}
-                  <code className="rounded-sm bg-mineshaft-600 px-0.5 py-px text-sm text-mineshaft-300">
+                  <code className="rounded-sm bg-ring px-0.5 py-px text-sm text-label">
```
scott-ray-wilson: I feel like this is a little hard to read/not enough contrast 

Image

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AzureEntraIdScimSyncFields.tsx
```diff
@@ -74,27 +74,31 @@ export const AzureEntraIdScimSyncFields = () => {
...
-                onInputChange={(newValue) => setSearchInput(newValue)}
-                filterOption={null}
+                onClear={() => {
+                  setSelectedSp(null);
+                  onChange("");
+                }}
+                onInputValueChange={(newValue) => setSearchInput(newValue)}
+                shouldFilter={false}
                 isLoading={isLoadingServicePrincipals}
                 options={options}
```
scott-ray-wilson: ah yeah doesn't look like we are passing error state to any of these comboboxes

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AzureEntraIdScimSyncFields.tsx
```diff
@@ -74,27 +74,31 @@ export const AzureEntraIdScimSyncFields = () => {
...
-              <FilterableSelect
+              <Combobox
                 value={options.find((sp) => sp.id === value) ?? null}
-                onChange={(option) => {
+                onValueChange={(option) => {
                   const selected = option as TAzureScimServicePrincipal | null;
                   setSelectedSp(selected);
                   onChange(selected?.id ?? "");
                 }}
-                onInputChange={(newValue) => setSearchInput(newValue)}
```
scott-ray-wilson, line 84: don't we need to preserve this?  `onInputValueChange={(newValue) => setSearchInput(newValue)}`

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitLabSyncFields.tsx
scott-ray-wilson, line 1: feels kinda janky transition wise

https://github.com/user-attachments/assets/db135ae7-3658-4610-9028-239d7b965484

scott-ray-wilson, line 1: also here, why does it show twice? 

https://github.com/user-attachments/assets/2887ba89-424b-4279-aac0-31d372158bbb

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/NetlifySyncFields.tsx
```diff
@@ -91,12 +89,12 @@ export const NetlifySyncFields = () => {
           <Field>
             <FieldLabel>Site (Optional)</FieldLabel>
             <FieldContent>
-              <FilterableSelect
+              <Combobox
                 isLoading={isSitesLoading && Boolean(accountId)}
                 isDisabled={!accountId}
```
scott-ray-wilson, line 121: given this is optional I think we should make this one clearable

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/NetlifySyncFields.tsx
```diff
@@ -125,17 +124,18 @@ export const NetlifySyncFields = () => {
           <Field>
             <FieldLabel>Context (Optional)</FieldLabel>
             <FieldContent>
-              <FilterableSelect
+              <Combobox
                 isDisabled={!accountId}
                 value={contexts.find((p) => p.value === value) ?? undefined}
-                onChange={(option) => {
-                  const v = option as SingleValue<{ label: string; value: NetlifySyncContext }>;
+                onValueChange={(option) => {
```
scott-ray-wilson: same here, context is optional so should have clear

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/TravisCISyncFields.tsx
scott-ray-wilson, line 1: I just noticed this here but might apply to others, if I select an option from a clearable combobox, navigate forward, then back, I can't clear it - we should see where this has implications else where

Image

frontend/src/components/secret-syncs/forms/SecretSyncOptionsFields/AwsParameterStoreSyncOptionsFields.tsx
```diff
@@ -176,33 +172,31 @@ export const AwsParameterStoreSyncOptionsFields = () => {
...
                     value={kmsKeys.find((org) => org.alias === value) ?? null}
-                    onChange={(option) =>
-                      onChange((option as SingleValue<TAwsConnectionKmsKey>)?.alias ?? null)
-                    }
+                    onValueChange={(option) => onChange(option.alias ?? null)}
                     isError={Boolean(error)}
                     // eslint-disable-next-line react/no-unstable-nested-components
-                    noOptionsMessage={({ inputValue }) =>
+                    emptyMessage={(inputValue) =>
                       inputValue ? undefined : (
```
scott-ray-wilson: Image looks like the empty state when theres an input value is just empty not the default

frontend/src/components/secret-syncs/forms/SecretSyncOptionsFields/AwsSecretsManagerSyncOptionsFields.tsx
```diff
@@ -181,33 +177,31 @@ export const AwsSecretsManagerSyncOptionsFields = () => {
...
                     value={kmsKeys.find((org) => org.alias === value) ?? null}
-                    onChange={(option) =>
-                      onChange((option as SingleValue<TAwsConnectionKmsKey>)?.alias ?? null)
-                    }
+                    onValueChange={(option) => onChange(option.alias ?? null)}
                     isError={Boolean(error)}
                     // eslint-disable-next-line react/no-unstable-nested-components
-                    noOptionsMessage={({ inputValue }) =>
+                    emptyMessage={(inputValue) =>
                       inputValue ? undefined : (
```
scott-ray-wilson: same here 

Image

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/1PasswordSyncFields.tsx
scott-ray-wilson, line 1: 1password is is breaking if you select a connection 

https://github.com/user-attachments/assets/04b1060b-e2aa-41e8-9696-b001795c797d

scott-ray-wilson (dismissed):

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AwsParameterStoreSyncFields.tsx
andrewhuhh: CleanShot 2026-09-02 at 3 08 23 PM@2x
fixd

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AwsParameterStoreSyncFields.tsx
andrewhuhh: fixed with a new defined and styled <Code> component

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AzureEntraIdScimSyncFields.tsx
andrewhuhh: so this was a gap in the existing code, just ported over to new shape

andrewhuhh: fixing!!

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AzureEntraIdScimSyncFields.tsx
andrewhuhh: it's preserved below as onInputValueChange={(newValue) => setSearchInput(newValue)} which is the new Combobox equivalent of onInputChange

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/shared/AwsRegionSelect.tsx
scott-ray-wilson, line 1: oh I can't search by region slug anymore

Image

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/NetlifySyncFields.tsx
scott-ray-wilson, line 1: this isn't specific to netlify, just using it as an example but there's an issue with clearing vs unset in duplication checking in the UI 

https://github.com/user-attachments/assets/3688fe7c-cbda-410c-bfe1-20e061646efa

backend/src/services/app-connection/gitlab/gitlab-connection-fns.ts
```diff
@@ -314,6 +314,28 @@ export const getGitLabConnectionClient = async (
...
+    }
+
+    throw new BadRequestError({
+      message: `Failed to fetch GitLab ${resource}: ${error.message ?? "Unknown error"}${error.cause?.description && error.message !== "Unauthorized" ? `. Cause: ${error.cause.description}` : ""}`
+    });
+  }
+
+  if (error instanceof InternalServerError) {
+    throw error;
+  }
```
scott-ray-wilson, line 337: claude is saying this is dead code:

**The GitLab 429 handling is dead code** (`gitlab-connection-fns.ts:317`). `throwGitLabListError` maps `GitbeakerRequestError` with a 429 to `RateLimitError`, but `@gitbeaker/rest` lists 429 in `retryCodes = [429, 502]` and only calls `throwFailedRequestError` (the sole producer of `GitbeakerRequestError`) for codes *not* in that list. After 10 retries it throws `GitbeakerRetryError`, which matches neither branch, so it falls to `InternalServerError` → the service's `catch` returns `[]` → HTTP 200. The user still sees "No groups found." on a rate limit, and the frontend's "GitLab is rate limiting groups" string is unreachable. Catching `GitbeakerRetryError` (its message carries `last status code: 429`) would make the path live.

frontend/src/components/v3/generic/Combobox/Combobox.tsx
```diff
@@ -284,39 +371,90 @@ const SingleCombobox = <TOption,>({
...
   const inputRef = React.useRef<HTMLInputElement | null>(null);
+  const highlightedOptionValueRef = React.useRef<string | null>(null);
   const [open, setOpen] = React.useState(false);
+  const selectedLabel = value == null ? "" : getOptionLabel(value);
+  const [search, setSearch] = React.useState("");
   const selectedOptions = React.useMemo(() => (value == null ? [] : [value]), [value]);
-  const items = useComboboxItems(options, selectedOptions, getOptionValue);
+  // Locally-filtered lists retain missing selections so the primitive can filter
+  // them normally. Externally-filtered lists must reflect only provider results.
+  const items = useComboboxItems(options, selectedOptions, getOptionValue, shouldFilter);
```
scott-ray-wilson: Coupling `includeMissingSelectedOptions` to `shouldFilter` ties together two things consumers need independently: "the server does my filtering" and "keep my saved selection visible." Every `shouldFilter={false}` caller loses the second one as a side effect of asking for the first.

Four of them noticed and rebuilt it by hand, each differently: `GitLabSyncFields` (`groupOptions`/`projectOptions`), `VercelSyncFields` (`teamOptions`/`projectOptions`), and `AzureEntraIdScimSyncFields` (component state, so it resets on remount). That is seven memos across three files, each fabricating fields the API owns, which will drift.

`BitbucketSyncFields` did not, and that one loses data. `bitbucket-connection-fns.ts` never paginates workspaces, so any saved workspace outside the first page is absent from `options`. On the Edit form the required Workspace field then renders empty while `destinationConfig.workspaceSlug` is still set, and saving from that state silently retargets the sync.

Repro: create a Bitbucket sync, point `workspaceSlug` at a workspace that is not in the first page of results, reopen Edit, and look at the Destination step.

Suggest making this its own prop (or keeping it unconditional) rather than deriving it from `shouldFilter`. That fixes Bitbucket and lets all seven hand-rolled memos be deleted.

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/TravisCISyncFields.tsx
scott-ray-wilson, line 1: the `onChange(null)` fix works, but I don't think this got swept the way I was hoping. `setValue(..., undefined)` has the identical problem and is still here on line 66 (and line 47).
>
> mechanism, for the record: RHF's `useController` computes `defaultValue: get(control._formValues, name, ...)` and hands it to `useWatch`, which stashes it in a ref on first render and returns that snapshot whenever the live value is `undefined`. `null` never hits the fallback, which is exactly why your fix works. so once a step remounts with a value already set, that field can no longer be reset to `undefined`.
>
> repro: pick repo A + branch `main`, Next, Back, then switch to repo B. `_formValues.branch` is correctly `undefined`, but the Branch combobox still renders `main`, because `branches.find(b => b.name === value)` resolves against the stale name. pick two repos that share a branch name or you'll just see it go blank, which looks correct. so the payload is right and the UI is lying, showing a branch from the other repo that we aren't going to save.
>
> same pattern in two more spots, though latent rather than visible, since there the stale value is an id that won't resolve against the new list:
> - `ChecklySyncFields.tsx:51` (connection change resets `groupId`)
> - `GitHubSyncFields.tsx:76` (connection change resets `selectedRepositoryIds`, only reachable when visibility is already `Selected`)
>
> those two just render blank today so there's no user-visible break, but they're wrong the same way. Vercel `targetProjects` and Spacelift `fileMountFormat` are genuinely fine, their Controllers unmount before the reset lands.
>
> all of this is pre-existing on main, so happy for the Checkly/GitHub ones to be a follow-up. would take the two TravisCI lines here though since they're actually wrong on screen.

scott-ray-wilson (dismissed):

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/TravisCISyncFields.tsx
```diff
@@ -84,23 +84,25 @@ export const TravisCISyncFields = () => {
...
                 }
                 isDisabled={!connectionId || !currentRepositoryId}
                 value={branches.find((branch) => branch.name === value) ?? null}
-                onChange={(option) => {
-                  const branch = option as SingleValue<TTravisCIBranch>;
-                  onChange(branch?.name ?? undefined);
+                onValueChange={(option) => {
+                  const branch = option;
+                  onChange(branch.name);
                 }}
```
andrewhuhh, line 113: Repo change correctly resets branch to `""`, but select still writes `undefined` and clear still writes `null`.

That is the same RHF remount trap: `useController` falls back to the first-render snapshot whenever the live value is `undefined`, so after Next/Back the Branch combobox can show a stale name from the previous repo even though `_formValues.branch` was cleared.

Please standardize the optional branch to `""` in all three paths (`onValueChange`, `onClear`, and the repo reset).

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/BitbucketSyncFields.tsx
```diff
@@ -107,26 +106,31 @@ export const BitbucketSyncFields = () => {
...
+                isError={Boolean(error)}
                 isLoading={isRepositoriesLoading && Boolean(workspace)}
                 isDisabled={!workspace}
-                value={repositories.find((r) => r.slug === value) ?? null}
-                onChange={(option) => {
-                  const v = option as SingleValue<TBitbucketRepo>;
-                  onChange(v?.slug ?? "");
+                value={value || null}
+                onValueChange={(option) => {
+                  if (option === value) return;
```
andrewhuhh, line 131: Workspace already uses a fallback so a saved slug outside the current search page still renders:

```ts
workspaces.find((w) => w.slug === value) ?? (value ? { slug: value } : null)
```

Repository still uses `?? null`. On Edit, a saved repo that is not in the first page / current query will look empty even though `destinationConfig.repositorySlug` is set. Please give this the same slug/`full_name` fallback.

frontend/src/components/v3/generic/Combobox/Combobox.tsx
```diff
@@ -284,39 +360,94 @@ const SingleCombobox = <TOption,>({
...
+          // typing itself opens the popup, preserve the query emitted just before
+          // this event instead of replacing the user's first keystroke.
+          if (eventDetails.reason !== "input-change") setSearch(selectedLabel);
+        } else {
+          highlightedOptionValueRef.current = null;
+          setSearch("");
+          onInputValueChange?.("");
+        }
+      }}
+      onItemHighlighted={(item) => {
```
andrewhuhh, line 439: Click-open seeds the local input with `selectedLabel` but never calls `onInputValueChange`. Close does reset the external query to `""`.

For `shouldFilter={false}` consumers (GitLab, Bitbucket, Vercel, Azure SCIM) that means the input looks like a search for the selected name while the API still has the previous / empty query, so the menu shows first-page results under a query that looks filtered.

Please also emit `onInputValueChange(selectedLabel)` when seeding from the selected label (and keep the `input-change` exception so the first keystroke is not overwritten).

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/NetlifySyncFields.tsx
```diff
@@ -65,19 +62,22 @@ export const NetlifySyncFields = () => {
...
                 isLoading={isAccountsLoading && Boolean(connectionId)}
                 isDisabled={!connectionId}
                 value={accounts.find((p) => p.id === value) ?? null}
-                onChange={(option) => {
-                  const v = option as SingleValue<TNetlifyAccount>;
+                onValueChange={(option) => {
+                  const v = option;
                   onChange(v?.id ?? null);
                   setValue("destinationConfig.accountName", v?.name ?? "");
                 }}
```
andrewhuhh, line 86: Changing account only updates `accountId` / `accountName`. `siteId`, `siteName`, and `context` from the previous account stay in form state and can be submitted against the new account.

This is pre-existing, but it is more visible now that site/context are explicitly clearable. Please also clear those fields here (same pattern as `SecretSyncConnectionField`).

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/ChecklySyncFields.tsx
```diff
@@ -61,19 +59,22 @@ export const ChecklySyncFields = () => {
...
                 isLoading={isAccountsLoading && Boolean(connectionId)}
                 isDisabled={!connectionId}
                 value={accounts.find((p) => p.id === value) ?? null}
-                onChange={(option) => {
-                  const v = option as SingleValue<TChecklyAccount>;
+                onValueChange={(option) => {
+                  const v = option;
                   onChange(v?.id ?? null);
                   setValue("destinationConfig.accountName", v?.name ?? "");
                 }}
```
andrewhuhh, line 81: Same class as Netlify: changing account does not reset `groupId` / `groupName`. Connection change does, account change does not.

Also, the group combobox is only disabled on `!connectionId` even though the groups query requires `accountId`. Please clear the group on account change and disable the control until an account is selected.

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/AzureEntraIdScimSyncFields.tsx
```diff
@@ -74,27 +61,32 @@ export const AzureEntraIdScimSyncFields = () => {
...
+                shouldFilter={false}
+                includeMissingSelectedOptions={!searchInput}
                 isLoading={isLoadingServicePrincipals}
-                options={options}
+                options={(servicePrincipals ?? []).map((sp) => sp.id)}
                 placeholder="Search for a SCIM service principal..."
-                getOptionLabel={(option) => option.displayName}
-                getOptionValue={(option) => option.id}
-                isClearable
+                getOptionLabel={(option) =>
```
andrewhuhh, line 96: `selectedSp` starts `null` and the combobox value is `options.find(...) ?? null`. On Edit, `servicePrincipalDisplayName` is already on the sync config, but this field never hydrates from it. If the saved principal is not in the first search page, the required control renders empty while `destinationConfig.servicePrincipalId` is still set.

This is pre-existing, but Combobox's missing-selected merge cannot help when `value` is `null`. Please initialize from the persisted id + display name (Bitbucket-style value fallback) instead of component state.

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitHubSyncFields.tsx
```diff
@@ -200,12 +197,13 @@ export const GitHubSyncFields = () => {
             <Field>
               <FieldLabel>Repository</FieldLabel>
               <FieldContent>
-                <FilterableSelect
+                <Combobox
+                  isError={Boolean(error)}
                   isLoading={isRepositoriesPending && Boolean(connectionId)}
                   isDisabled={!connectionId}
```
andrewhuhh, line 226: Value is still matched with `repo.name === value` even though the option label is `owner.login/name` and `getOptionValue` is the repo id.

On Edit, two repos with the same name under different owners can show the wrong selection. Pre-existing, but worth matching on `owner` + `name` (or the persisted owner + name pair) while this file is open.

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitLabSyncFields.tsx
```diff
@@ -269,30 +298,44 @@ export const GitLabSyncFields = () => {
                 </Tooltip>
               </FieldLabel>
               <FieldContent>
-                <FilterableSelect
-                  isLoading={isProjectsLoading && Boolean(connectionId)}
+                <Combobox
+                  isError={Boolean(error)}
+                  isLoading={isProjectSearchPending}
```
andrewhuhh, line 315: `useComboboxItems` now always appends the selected value when it is missing from `options`. That fixes Bitbucket-style edit hydration, but it undoes the `!search` injection guard in `groupOptions` / `projectOptions` above.

During an active server search, the saved group/project will reappear in the open list even when it does not match the query, because `value={selectedGroup}` is enough for Combobox to re-inject it.

Please either make missing-selected injection opt-in / search-aware on Combobox, or drop the `!search` injection blocks and keep display via these value fallbacks only.

andrewhuhh (commented):
Most of the earlier review items look fixed: invalid states, AWS region slug search, optional clear, GitLab 429, discard confirmation, 1Password, empty states, and always-visible missing selections.

I would not merge yet. The Travis `undefined`/`null`/`""` remount trap and the Bitbucket repo edit fallback are still user-visible. The Combobox click-open path also never syncs `onInputValueChange`, so server-searched fields show first-page results under a query that looks filtered.

The other comments are pre-existing leftovers that are worth fixing while these files are open (Netlify/Checkly cascade resets, Azure SCIM edit hydration, GitHub repo matching by name).

PR body still mentions Checkly/Netlify/Travis fixture credentials; those were added and then removed.

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/GitHubSyncFields.tsx
scott-ray-wilson, line 1: swapping orgs doesn't clear the selection of previous repos from the other org

https://github.com/user-attachments/assets/7d2c2d65-5d6f-4e1c-b347-a1daea0f8425

frontend/src/components/secret-syncs/forms/SecretSyncDestinationFields/VercelSyncFields.tsx
scott-ray-wilson, line 1: https://github.com/user-attachments/assets/96439290-bd30-403c-a5fd-77c4b87c2076

scott-ray-wilson, line 1: ignore file placement:

connection shows error when I click create connection 

https://github.com/user-attachments/assets/cfdd474d-6ef5-4877-bd56-6ae251aec487

scott-ray-wilson, line 1: oh this one should be clearable (probably same for AWS SM and AWS parameter store)

Image

scott-ray-wilson (dismissed):

scott-ray-wilson (approved):
lgtm

PR #7913, opened 2026-09-01
chore(backend): dependency updates

Vligai:
## Context

Five findings from a customer container scan, all still open on main.

| CVE | Package | Before | After |
|---|---|---|---|
| CVE-2025-55182, CVE-2025-55184 | react, react-dom | 19.1.0 | 19.1.4 |
| CVE-2026-12143 | form-data | 4.0.5 | 4.0.6 |
| CVE-2026-13149 | brace-expansion | 2.0.3 | 2.1.4 |
| CVE-2026-42508 and eight more | golang.org/x/crypto | v0.51.0 | v0.52.0 |

**brace-expansion is scoped, not global.** The tree carries four copies. Only one is production, reached through `glob` to `minimatch@9.0.9` on `^2.0.2`. A top-level override would have forced `minimatch@3` off its `^1.1.7` range and downgraded nodemon and react-email from 5.x to 2.x, so the override sits under `glob` instead. Production lands on 2.1.4 and the dev copies keep their own majors.

**x/crypto is hygiene rather than a fix.** Every affected path is in `x/crypto/ssh`, and `backend-go` imports `golang.org/x/crypto/bcrypt` and nothing else, so none of the nine advisories is reachable. v0.52.0 is the advisory floor. v0.54.0 would match the CLI but pulls six further `golang.org/x` modules along with it, so the smaller diff won.

## Steps to verify the change

```bash
cd backend-go && go build ./... && go test ./internal/... -count=1

cd backend && npm ci
npm run type:check
npm run test:unit
```

Confirm the brace-expansion override did not disturb the dev copies:

```bash
cd backend && node -e '
const l=require("./package-lock.json").packages;
for (const [k,v] of Object.entries(l)) if (k.endsWith("node_modules/brace-expansion"))
  console.log(v.version.padEnd(8), "dev="+!!v.dev, k);'
```

Expect 2.1.4 non-dev under `glob`, and 1.1.13, 5.0.5, 5.0.6 all still dev under their own parents.

### Verified locally

- `go build ./...` clean, 17 of 17 Go unit packages pass
- `backend-go` still imports `golang.org/x/crypto/bcrypt` only
- Lockfile audited for added, removed and changed entries. No entries removed

Standalone image built with these changes and smoke tested against a throwaway database:

- All four bumps present in the image: react 19.1.4, react-dom 19.1.4, form-data 4.0.6, brace-expansion 2.1.4 under `glob`
- Bootstrap, project create, secret write and secret read back returned the identical plaintext
- Signup verification email rendered and delivered, 10,575 bytes of HTML. That exercises the react bump directly, since the email templates are React server-rendered
- Zero error-level log lines across the run

Couldn't test `npm run type:check` locally due to Node version mismatch, but they should be covered by CI tests.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [x] Chore

## Checklist

- [x] Tested locally


akhilmhdh (approved):

akhilmhdh (changes requested):
Pause, Testing a preview

akhilmhdh (approved):

PR #7915, opened 2026-09-01
feat(auth): add signup flow experiment

andrewhuhh:
## Context

The current signup flow always presents the authentication form on the standard auth background, which makes it difficult to measure whether showing users a preview of the product improves conversion.

This change:

- initializes PostHog lazily through the shared telemetry client and keeps initialization failure-safe
- adds the cloud-only, inactive `signup-flow` multivariate feature flag contract with `control` and `test` variants
- centralizes feature-flag loading, timeout, error fallback, and unsubscribe behavior in a helper shared with the active secrets-activation experiment
- keeps the current signup flow as the control
- adds a test signup view with the existing auth options layered over a blurred, non-interactive dashboard preview
- keeps experiment assignment sticky within the tab and across US/EU region switches, then records assigned conversions with the dedicated `Signup Flow Completed` event
- falls back to control when PostHog is disabled, unavailable, or slow
- allows the `signupFlow` query override only in local development and `pr-*.preview.infisical.com` review builds; production cloud assignment remains exclusively PostHog-driven
- preserves the current-main onboarding transitions and focused layout for post-auth setup steps

PostHog experiment and flag configuration remain inactive/draft.

Linear: [ENG-5635](https://linear.app/infisical/issue/ENG-5635/run-a-test-based-on-the-blurry-background-ab-test)

### Test — organization overview preview

Signup experiment test variant

### Control — current main

Signup experiment control variant

## Steps to verify the change

1. Start the local development stack with `docker compose -f docker-compose.dev.yml up -d --build`.
2. Visit `http://localhost:8080/signup?signupFlow=control` and confirm the current signup screen renders unchanged.
3. Visit `http://localhost:8080/signup?signupFlow=test` and confirm the same auth options render over a blurred, non-interactive dashboard preview.
4. Navigate from the test signup screen to `/login`, then use browser Back and confirm the test variant returns.
5. Confirm invite-based signup continues to use the control layout.
6. Run `make reviewable-ui` from the repository root.
7. Run `npm exec -- tsx --test src/components/analytics/experiments/signupFlow/signupExperimentConfig.test.ts src/pages/secret-manager/OverviewPage/components/InviteMembersNudge/resolveActivationVariant.test.ts` and `npm run build` from `frontend/`.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


andrewhuhh:
@greptile review

andrewhuhh:
@greptile review

frontend/src/components/analytics/experiments/signupFlow/signupExperiment.ts
victorvhs017, line 1: We could use an `experiments` folder and one folder per experiment here if this is something we'll be creating more in the future.

victorvhs017, line 1: In general, I feel like there are a lot of things and patterns here that could be generic and reused in other experiments in the future.

For example, it looks like we already have feature flags being used in the app here: frontend/src/pages/secret-manager/OverviewPage/components/InviteMembersNudge/resolveActivationVariant.ts

The file above also has some things that we are missing in this signup implementation, like unsubscribes when it settles.

Do we have any documentation on how we want to approach A/B testing, with some guidelines? I think this would be a good design doc (if it doesn't already exist)

frontend/src/components/analytics/signupExperimentConfig.ts
```diff
@@ -0,0 +1,16 @@
+export const SIGNUP_FLOW_FEATURE_FLAG = "signup-flow";
+export const SIGNUP_COMPLETED_EVENT = "User Signed Up";
```
victorvhs017: This is the same event name that the backend already sends ("User Signed Up" from signup-router.ts), so self-serve signups will fire twice on the same distinctId. Funnels and signup counts will double, and the properties won't match (signupMethod/signupSource on the backend vs signup_method/signup_flow_variant here).

Either use a different name (e.g., "Signup Flow Completed") or send the variant to the backend and add it to the existing event.

frontend/src/pages/auth/SignUpPage/SignUpPage.tsx
```diff
@@ -44,6 +51,7 @@ export interface SignUpPageProps {
 
 export const SignUpPage = ({ invite }: SignUpPageProps) => {
   const isInvite = Boolean(invite);
+  const signupFlowVariant = useSignupFlowVariant(!isInvite);
```
victorvhs017: `useSignupFlowVariant` calls `getPostHog()`, which inits PostHog on `/signup` whenever `ENV === "production"` and telemetry is on.That includes self-hosted prod builds with default `TELEMETRY_ENABLED=true` and the default PostHog key, so unauthenticated signup pages start calling `app.posthog.com` and setting cookies. If PostHog is unreachable (common in air-gapped setups), the hook sits on null until the 2.5s timeout, so users see "Preparing sign up" on every visit.

This experiment should be cloud-only. Gate with `useSignupFlowVariant(!isInvite && isInfisicalCloud())`. The comment at `SelectOrgPage/route.tsx:62` about PostHog never initializing on signup is stale after this.

frontend/src/components/analytics/signupExperiment.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    let isSettled = false;
+    const settle = (nextVariant: SignupFlowVariant, shouldPersist = false) => {
+      if (isSettled) return;
+      isSettled = true;
+      if (shouldPersist) persistSignupFlowVariant(nextVariant);
+      setVariant(nextVariant);
+    };
+
+    const unsubscribe = client.onFeatureFlags((_flags, _variants, context) => {
```
victorvhs017: If flags arrive after the 2.5s timeout, this callback still calls `getFeatureFlag()`, which records a `$feature_flag_called `exposure in `PostHog. settle()` then no-ops because `isSettled` is already true, so the user stays on control but gets counted in the test arm. Add `if (isSettled) return` before the `getFeatureFlag `call. Slow-network users will otherwise be misattributed and the experiment results will be off.

frontend/src/components/analytics/signupExperiment.ts
```diff
@@ -0,0 +1,107 @@
...
+  }, [enabled, variant]);
+
+  return variant;
+};
+
+export const captureSignupCompleted = (signupMethod: "email" | "sso") => {
+  const telemetry = new Telemetry().getInstance();
+  telemetry.capture(SIGNUP_COMPLETED_EVENT, {
+    signup_method: signupMethod,
+    signup_flow_variant: getPersistedSignupFlowVariant() ?? SignupFlowVariant.Control
```
victorvhs017: Defaulting to `"control"` when sessionStorage is empty lumps in users who were never in the experiment: flag timeouts, PostHog off, self-hosted, SSO paths that skip `SignUpPage`, etc. That inflates the control arm in any analysis on `signup_flow_variant`. `resolveSignupFlowVariant` already separates enrolled vs not via `shouldPersist`; if nothing was persisted, emit `undefined` or `"unassigned"` instead of `"control"`.

victorvhs017 (changes requested):

frontend/src/components/analytics/experiments/signupFlow/signupExperiment.ts
andrewhuhh: this is a good point; i have a separate PR that tries to be more diligent about this!!

andrewhuhh: i think maidul tagged u in that one

andrewhuhh: yes this could be a good idea

victorvhs017: Any developer should be able to add AB testing; it would be good to have the guidelines and all helper functions ready to be used!

victorvhs017 (changes requested):
Another thing I realized, if I switch to EU will I lose the variant? will it have to do the "Preparing sign up" again?

Image

frontend/src/components/analytics/experiments/signupFlow/signupExperiment.ts
victorvhs017, line 1: Let's add something regarding PostHog experiments, feature flags, and A/B testing in the `frontend/CLAUDE.md`

frontend/src/pages/auth/SignUpSsoPage/SignUpSsoPage.tsx
```diff
@@ -74,6 +75,7 @@ export const SignupSsoPage = () => {
       window.dataLayer = window.dataLayer || [];
       window.dataLayer.push({ event: "signup_completed" });
     }
+    captureSignupCompleted("sso");
```
victorvhs017: `captureSignupCompleted("sso")` fires before the existing-org check, so invited users get counted as conversions with `signupFlowVariant: "unassigned"`. They never hit `SignUpPage`, so they were never in the experiment.

The email path already guards this with `if (!isInvite)`. Same idea here: move this (and the GTM `dataLayer` push above it) below the `userOrgs.length > 0` check, and only fire when the user is going to onboarding.

victorvhs017 (changes requested):

frontend/src/pages/auth/SignUpSsoPage/SignUpSsoPage.tsx
andrewhuhh: oo good catch

andrewhuhh: actual ball knowledge wtf

victorvhs017 (approved):

PR #7916, opened 2026-09-01
fix: bound membership removal and role changes by the target's privileges for legacy system

Thiago-AS:
## Context

You could remove people who outranked you. If your role let you delete members,
identities, or groups, nothing checked whether the person you were removing had more
privileges than you did. So a custom role scoped to "manage members" was enough to
remove an org admin.

Role changes had a related gap. Downgrading someone to no-access was skipped by the
boundary check even though it strips their access just as completely as removing them.

There was also a bug in how the boundary was measured: when a user, identity, or group
held several roles, only the first one was checked. Holding [member, some-admin-ish-role]
got you measured as if you were just a member.

## Steps to verify the change

Removal and downgrade checks only kick in for orgs on the legacy privilege system, so
flip your test org over first:

```update organizations set "shouldUseNewPrivilegeSystem" = false where id = '<org-id>'```

Then set up a weak actor: make a custom org role that can read and delete members and
nothing else, and assign it to a second user. Log in as them.

Should now get rejected (all of these used to go through):

1. Remove an org admin, both from the members list and from a bulk selection.
2. Do the same in a project with a custom project role that can only remove members.
3. Set a group that holds the Admin role down to no-access.
4. Give a test user two roles, member plus a custom role with broad permissions, then try
   to remove them. Used to pass because only the first role was looked at.

Should still work:

5. An org or project admin doing every one of the above.
6. Removing or downgrading someone who does not outrank you.

Works on either privilege system, so no flag flip needed:

7. As a non-admin who can invite members, try inviting someone as admin, and try promoting
   an existing member to admin. Both should be rejected.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [x] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/e2e-test/routes/v1/membership-downgrade-boundary.spec.ts
+++ b/backend/e2e-test/routes/v1/membership-downgrade-boundary.spec.ts
@@ -298,17 +298,18 @@ describe("Privilege boundary on org membership downgrade", () => {
         asIdentity(actor.token)
       );
       expect(res.statusCode).toBe(403);
-      expect(res.json().message).toContain("more privileged org member");
+      expect(res.json().message).toContain("Failed to change the roles or attributes of this org member");
     });
 
-    test("deactivating a more privileged member follows the removal boundary", async () => {
-      // Deactivation keys on member:delete, so it lands where a removal lands rather than where a role
-      // change does. Own target, because it succeeds on one of the two runs.
+    test("deactivating a more privileged member is bounded on the legacy system only", async () => {
+      // Deactivation keys on member:edit, which the route's own permission check already demanded, so
+      // the new system has nothing further to ask. Own target, because it succeeds on one of the two runs.
       const target = await createAdminTarget();
 
       const res = await patchMembership(target.membershipId, { isActive: false }, asIdentity(actor.token));
       expect(res.statusCode).toBe(newSystem ? 200 : 403);
-      if (!newSystem) expect(res.json().message).toContain("more privileged org member");
+      if (!newSystem)
+        expect(res.json().message).toContain("Failed to change the activation status of this org member");
     });
   });
 
--- a/backend/e2e-test/routes/v1/membership-downgrade-boundary.spec.ts
+++ b/backend/e2e-test/routes/v1/membership-downgrade-boundary.spec.ts
@@ -21,7 +21,7 @@ import crypto from "node:crypto";
 
 import { packRules } from "@casl/ability/extra";
 
-import { AccessScope, ProjectMembershipRole, TableName } from "@app/db/schemas";
+import { AccessScope, OrgMembershipRole, ProjectMembershipRole, TableName } from "@app/db/schemas";
 import { seedData1 } from "@app/db/seed-data";
 
 const adminHeaders = () => ({ authorization: `Bearer ${jwtAuthToken}` });
@@ -208,6 +208,109 @@ describe("Privilege boundary on project membership downgrade", () => {
       headers: adminHeaders(),
       body: { roles: [{ role: ProjectMembershipRole.NoAccess }] }
     });
+    expect(res.statusCode).toBe(200);
+  });
+});
+
+describe("Privilege boundary on org membership downgrade", () => {
+  // PATCH /api/v2/organizations/:orgId/memberships/:membershipId updates a member through
+  // org-service rather than through the scoped membership factory, and it only bounded the role
+  // being ASSIGNED. Downgrading an Admin to no-access, or deactivating them, assigns nothing the
+  // boundary could reject, so it was a way around both the update guard and the removal boundary:
+  // downgrade first, then remove the now-weaker member.
+  let actor: { identityId: string; token: string };
+  let adminTarget: { userId: string; membershipId: string };
+
+  const roleSlug = `e2e-org-downgrader-${crypto.randomUUID()}`;
+
+  const patchMembership = (membershipId: string, body: Record<string, unknown>, headers: Record<string, string>) =>
+    testServer.inject({
+      method: "PATCH",
+      url: `/api/v2/organizations/${seedData1.organization.id}/memberships/${membershipId}`,
+      headers,
+      body
+    });
+
+  const giveOrgAdminMembership = async (userId: string) => {
+    const [membership] = await testDb(TableName.Membership)
+      .insert({ scope: AccessScope.Organization, scopeOrgId: seedData1.organization.id, actorUserId: userId })
+      .returning("id");
+    const membershipId = (membership as { id: string }).id;
+    await testDb(TableName.MembershipRole).insert({ membershipId, role: OrgMembershipRole.Admin });
+    return membershipId;
+  };
+
+  beforeAll(async () => {
+    // member edit+delete and nothing else: the actor can legitimately reach the route, yet holds
+    // strictly less than the org Admin it is pointed at.
+    const [role] = await testDb(TableName.Role)
+      .insert({
+        name: roleSlug,
+        slug: roleSlug,
+        orgId: seedData1.organization.id,
+        permissions: JSON.stringify(packRules([{ subject: "member", action: ["read", "edit", "delete"] }] as never))
+      })
+      .returning("id");
+
+    actor = await createActorIdentity(`e2e-org-downgrade-actor-${crypto.randomUUID()}`);
+    const actorMembership = await testDb(TableName.Membership)
+      .where({
+        actorIdentityId: actor.identityId,
+        scope: AccessScope.Organization,
+        scopeOrgId: seedData1.organization.id
+      })
+      .first();
+    const actorMembershipId = (actorMembership as { id: string }).id;
+    await testDb(TableName.MembershipRole).where({ membershipId: actorMembershipId }).delete();
+    await testDb(TableName.MembershipRole).insert({
+      membershipId: actorMembershipId,
+      role: OrgMembershipRole.Custom,
+      customRoleId: (role as { id: string }).id
+    });
+
+    const target = await createTargetUser("org-downgrade");
+    adminTarget = { userId: target.userId, membershipId: await giveOrgAdminMembership(target.userId) };
+  });
+
+  afterAll(async () => {
+    await setNewPrivilegeSystem(true);
+    await testServer.inject({
+      method: "DELETE",
+      url: `/api/v1/identities/${actor.identityId}`,
+      headers: adminHeaders()
+    });
+    await testDb(TableName.Membership).where({ actorUserId: adminTarget.userId }).delete();
+    await testDb(TableName.Users).where({ id: adminTarget.userId }).delete();
+    await testDb(TableName.Role).where({ slug: roleSlug }).delete();
+  });
+
+  describe.each([
+    { label: "legacy", newSystem: false },
+    { label: "new", newSystem: true }
+  ])("$label privilege system", ({ newSystem }) => {
+    beforeAll(async () => {
+      await setNewPrivilegeSystem(newSystem);
+    });
+
+    test("downgrading a more privileged member to no-access is bounded", async () => {
+      const res = await patchMembership(
+        adminTarget.membershipId,
+        { role: OrgMembershipRole.NoAccess },
+        asIdentity(actor.token)
+      );
+      expect(res.statusCode).toBe(403);
+      expect(res.json().message).toContain("more privileged org member");
+    });
+
+    test("deactivating a more privileged member is bounded", async () => {
+      const res = await patchMembership(adminTarget.membershipId, { isActive: false }, asIdentity(actor.token));
+      expect(res.statusCode).toBe(403);
+      expect(res.json().message).toContain("more privileged org member");
+    });
+  });
+
+  test("regression: an admin can still downgrade an org Admin to no-access", async () => {
+    const res = await patchMembership(adminTarget.membershipId, { role: OrgMembershipRole.NoAccess }, adminHeaders());
     expect(res.statusCode).toBe(200);
   });
 });
--- a/backend/e2e-test/routes/v1/membership-downgrade-boundary.spec.ts
+++ b/backend/e2e-test/routes/v1/membership-downgrade-boundary.spec.ts
@@ -1,0 +1,213 @@
+/**
+ * The membership *update* guard has to bound the target's current roles, not just the roles being
+ * assigned.
+ *
+ * The update guards resolve `dto.data.roles` -- the roles being ASSIGNED -- and filter NoAccess out
+ * before resolution. A request of exactly [no-access] therefore resolved to an empty list, the
+ * boundary loop never ran, and nothing was checked on either privilege system. Since setting a
+ * member to no-access revokes their access just as a delete does, that was a way around the
+ * removal boundary rather than a separate concern:
+ *
+ *   DELETE an Admin target        -> 403 (bounded)
+ *   PATCH  that Admin -> viewer   -> 403 (bounded, actor cannot assign viewer)
+ *   PATCH  that Admin -> no-access-> 200 (unbounded)  <-- the gap
+ *
+ * Removing the filter alone does not fix the legacy system: no-access grants nothing, so
+ * dominating it passes trivially. The boundary has to run against the target's CURRENT roles,
+ * which is what the delete guard already does.
+ */
+
+import crypto from "node:crypto";
+
+import { packRules } from "@casl/ability/extra";
+
+import { AccessScope, ProjectMembershipRole, TableName } from "@app/db/schemas";
+import { seedData1 } from "@app/db/seed-data";
+
+const adminHeaders = () => ({ authorization: `Bearer ${jwtAuthToken}` });
+const asIdentity = (token: string) => ({ authorization: `Bearer ${token}` });
+
+const setNewPrivilegeSystem = async (enabled: boolean) => {
+  await testDb(TableName.Organization)
+    .where({ id: seedData1.organization.id })
+    .update({ shouldUseNewPrivilegeSystem: enabled });
+};
+
+const createActorIdentity = async (name: string) => {
+  const createRes = await testServer.inject({
+    method: "POST",
+    url: "/api/v1/identities",
+    headers: adminHeaders(),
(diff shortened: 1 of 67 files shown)

### CODE REVIEW

File: backend/src/services/membership-identity/org/org-membership-identity-factory.ts
Comment (scott-ray-wilson) [inline backend/src/services/membership-identity/org/org-membership-identity-factory.ts:1]: we have org routes that don't go through this so I can downgrade an admin as a custom role and then remove them 

 `PATCH /api/v2/organizations/:orgId/memberships/:membershipId` → `org.updateOrgMembership` (`org-service.ts:836`)

https://github.com/user-attachments/assets/58e85e86-ffc7-407e-a2c4-8a311e1544bd

File: backend/src/ee/services/group/group-service.ts [hunk 1]
```diff
@@ -1019,31 +1024,23 @@ export const groupServiceFactory = ({
...
+      ignoreUnresolvedRoles: true
+    });
     const { shouldUseNewPrivilegeSystem } = await requestMemoize(requestMemoKeys.orgFindById(actorOrgId), () =>
       orgDAL.findById(actorOrgId)
     );
 
-    // check if user has broader or equal to privileges than group
-    const permissionBoundary = validatePrivilegeChangeOperation(
+    // check if user has broader or equal to privileges than every role the group holds
+    assertRoleSetBoundary({
```
scott-ray-wilson, line 1066: this broke the new system: previously if a group had no access org role, and I as the actor had only read and edit groups, it would fail because I don't have the specific add members permission; now this incorrectly allows me to add to the group (repeated at 1134, 1229 and 1306)

main:

Image

branch:

Image

backend/src/services/membership-identity/project/project-membership-identity-factory.ts
```diff
@@ -178,6 +181,27 @@ export const newProjectMembershipIdentityFactory = ({
...
+      });
+      assertRoleSetBoundary({
+        shouldUseNewPrivilegeSystem,
+        opActions: [ProjectPermissionIdentityActions.AssignRole, ProjectPermissionIdentityActions.GrantPrivileges],
+        opSubject: ProjectPermissionSub.Identity,
+        actorPermission: permission,
+        targetPermissions,
+        baseMessage: "Failed to change the roles of a more privileged identity",
+        subjectFields: { identityId: dto.selector.identityId }
+      });
```
scott-ray-wilson: this broke the new system: previously if I as the actor had `identity:assign-role` scoped with an `assignableRole` condition (say `$in [viewer, member]`), I could move an identity from member to viewer, because the per-role check below passes `assignableRole` and my rule matched; now the new target check at runs first with only `{ identityId }`, so my `assignableRole` condition is evaluated against undefined and fails, and I get "Permission denied: assign-role on identity" for every identity that holds a real role (with a `$glob` condition it 500s instead, via `lib/casl/index.ts:19`). fix is to pass `assignableRole` per target role (the by-roles result has `role.slug` for built-ins too), and the user/group update guards want the same treatment since they pass no fields at all

backend/src/services/membership-user/project/project-membership-user-factory.ts
```diff
@@ -232,6 +234,26 @@ export const newProjectMembershipUserFactory = ({
...
+        ignoreUnresolvedRoles: true
+      });
+      assertRoleSetBoundary({
+        shouldUseNewPrivilegeSystem,
+        opActions: [ProjectPermissionMemberActions.AssignRole, ProjectPermissionMemberActions.GrantPrivileges],
+        opSubject: ProjectPermissionSub.Member,
+        actorPermission: permission,
+        targetPermissions,
+        baseMessage: "Failed to change the roles of a more privileged member"
+      });
```
scott-ray-wilson: this doesn't fully close the no-access gap on the new system: the new target check here passes no `subjectFields`, so CASL does a subject-type check and ignores rule conditions. if I as the actor have `member:assign-role` scoped with `userEmail $glob *@contractors.example`, PATCH alice@corp.example to viewer is correctly rejected by the per-role loop at below (it passes `userEmail`), but PATCH her to no-access still goes through, because that loop is empty for no-access and this check passes as soon as I hold any assign-role rule at all. so a scoped actor can still strip an out-of-scope Admin, which is the exact downgrade this PR is closing. `targetUser.email` is already loaded here, so pass `{ userEmail: targetUser.email || undefined }`; same thing in the group guard at `project-membership-group-factory.ts

backend/src/services/membership-group/project/project-membership-group-factory.ts
```diff
@@ -165,6 +167,27 @@ export const newProjectMembershipGroupFactory = ({
...
+      });
+
+      assertRoleSetBoundary({
+        shouldUseNewPrivilegeSystem,
+        opActions: [ProjectPermissionGroupActions.AssignRole, ProjectPermissionGroupActions.GrantPrivileges],
+        opSubject: ProjectPermissionSub.Groups,
+        actorPermission: permission,
+        targetPermissions,
+        baseMessage: "Failed to change the roles of a more privileged group"
+      });
```
scott-ray-wilson: same concern as user here

backend/src/ee/services/group/group-service.ts
scott-ray-wilson, line 314: this one still bypasses on the legacy version

backend/src/services/membership-identity/membership-identity-service.test.ts
scott-ray-wilson, line 1: not sure if out of scope but it looks like the `backend/src/services/identity/identity-service.ts` legacy mode still bypass

scott-ray-wilson (changes requested):

backend/e2e-test/routes/v1/membership-downgrade-boundary.spec.ts
```diff
@@ -0,0 +1,316 @@
...
+import { AccessScope, OrgMembershipRole, ProjectMembershipRole, TableName } from "@app/db/schemas";
+import { seedData1 } from "@app/db/seed-data";
+
+const adminHeaders = () => ({ authorization: `Bearer ${jwtAuthToken}` });
+const asIdentity = (token: string) => ({ authorization: `Bearer ${token}` });
+
+const setNewPrivilegeSystem = async (enabled: boolean) => {
+  await testDb(TableName.Organization)
+    .where({ id: seedData1.organization.id })
+    .update({ shouldUseNewPrivilegeSystem: enabled });
```
scott-ray-wilson, line 32: nit: I think we should add setting this back to false in `reset-shared-org-flags.ts` as a fallback/safety reset

backend/src/services/membership/membership-role-dal.ts
```diff
@@ -1,10 +1,44 @@
...
-  return orm;
+
+  // Role rows for a set of memberships, carrying the custom-role slug rather than its id so the
+  // result feeds resolveMembershipRoleSlugs directly. Batched so a bulk privilege-boundary check
+  // costs one query no matter how many memberships the caller is removing.
+  const findRolesByMembershipIds = async (membershipIds: string[], tx?: Knex): Promise<TMembershipRoleSlug[]> => {
+    if (!membershipIds.length) return [];
+
+    try {
+      return await (tx || db.replicaNode())(TableName.MembershipRole)
```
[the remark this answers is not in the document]
Thiago-AS: This was already like that before, not changing that in this scope.

backend/src/ee/services/permission/org-permission.ts
```diff
@@ -101,6 +101,14 @@ export enum OrgPermissionKmipServerActions {
   RevokeKmipServerAccess = "revoke-kmip-server-access"
 }
 
+export enum OrgPermissionMemberActions {
+  Read = "read",
+  Create = "create",
+  Edit = "edit",
+  Delete = "delete",
+  GrantPrivileges = "grant-privileges"
```
scott-ray-wilson, line 109: isn't this going to break any existing custom roles that previously could edit members or invite team members? invite is basically admin only now because no previous custom roles will have grant privs unless we back fill?

frontend/src/pages/project/AccessControlPage/components/MembersTab/components/AddMemberModal.tsx
```diff
@@ -109,7 +109,7 @@ export const AddMemberModal = ({ popUp, handlePopUpToggle }: Props) => {
     ProjectPermissionSub.Member
   );
   const canInviteNewMembers = orgPermission.can(
-    OrgPermissionActions.Create,
+    OrgPermissionMemberActions.Create,
```
scott-ray-wilson, line 112: this would no longer just be create right? because a role is always passed (or defaults to member) so all the invites would need the grant privileges check as well? same with editing role, no longer just edit but add. priv. so we'd need to audit all the usages and adjust them:

 The role selector at `OrgMembersTable.tsx:546`, the Edit User item at `OrgMembersTable.tsx:617`, the Edit User item at `UserDetailsByIDPage.tsx:147`, and the pencil at `UserDetailsSection.tsx:111` all gate on `edit`, but their save paths now need `grant-privileges`. Also, invite buttons at `Navbar.tsx:610` and `OrgMembersSection.tsx:175` gate on `create`, which is no longer sufficient.

frontend/src/pages/organization/RoleByIDPage/components/OrgRoleModifySection.utils.ts
```diff
@@ -327,24 +340,29 @@ export const ORG_PERMISSION_OBJECT: Record<string, TOrgPermissionConfig> = {
...
-        value: OrgPermissionActions.Create,
+        value: OrgPermissionMemberActions.Create,
         label: "Invite members",
         description: "Invite new users to join the organization"
       },
       {
-        value: OrgPermissionActions.Edit,
+        value: OrgPermissionMemberActions.Edit,
         label: "Edit members",
         description: "Modify member roles and access settings"
```
scott-ray-wilson: this still mentions modifying roles 

Image

backend/src/ee/services/permission/permission-fns.ts
```diff
@@ -634,6 +678,7 @@ export const interpolatePermissionRules = <T>(rules: T[], identityContext: Recor
 
 export {
   assertPermissionBoundary,
```
scott-ray-wilson: seems like we can remove `assertPermissionBoundary` as it's no longer used?

backend/src/services/org/org-service.ts
```diff
@@ -865,6 +867,42 @@ export const orgServiceFactory = ({
...
+        opAction: OrgPermissionMemberActions.GrantPrivileges,
+        baseMessage: "Failed to change the roles or attributes of a more privileged org member"
+      });
+    if (isActive !== undefined)
+      targetOps.push({
+        opAction: OrgPermissionMemberActions.Delete,
+        baseMessage: "Failed to change the activation status of a more privileged org member"
+      });
+
+    if (targetOps.length) {
```
scott-ray-wilson, line 886: doesn't this negate the empty set handling you added to `assertRoleSetBoundary` so the check always runs?

backend/src/services/membership-user/org/org-membership-user-factory.ts
```diff
@@ -321,20 +328,64 @@ export const newOrgMembershipUserFactory = ({
...
+      });
+
+    if (targetOps.length) {
+      const targetMembership = await membershipUserDAL.getUserById({
+        scopeData: dto.scopeData,
+        userId: dto.selector.userId
+      });
+      const targetRoles = targetMembership ? resolveMembershipRoleSlugs(targetMembership.roles) : [];
+
+      if (targetRoles.length) {
```
scott-ray-wilson: same concern here, shouldn't we allow the empty set still run the check?

backend/src/services/identity/identity-service.ts
```diff
@@ -243,6 +245,24 @@ export const identityServiceFactory = ({
...
+        });
+
+        assertRoleSetBoundary({
+          shouldUseNewPrivilegeSystem,
+          opActions: OrgPermissionIdentityActions.GrantPrivileges,
+          opSubject: OrgPermissionSubjects.Identity,
+          actorPermission: permission,
+          targetPermissions,
+          baseMessage: "Failed to change the roles of a more privileged identity"
+        });
```
scott-ray-wilson: there's two other identity delete paths that don't have this addition that I think should (respective of their scope) - ie the delete guards?

`backend/src/services/identity-v2/org/org-identity-factory.ts:55`

and

 `backend/src/services/identity-v2/project/project-identity-factory.ts:64`.

backend/src/ee/services/access-approval-request/access-approval-request-fns.ts
```diff
@@ -39,11 +43,21 @@ type TUnpackedAccessApprovalRequestRule = {
...
 // .strict() at both levels so no other operator ($in/$eq/...) or key
 // (secretName, secretTags, connectionId, metadata, ...) can be smuggled in.
-// The $glob value reuses the same validator as every other permission surface.
+// The $glob value reuses the same validator as every other permission surface, bounded here rather
+// than on the shared schema so the tightening cannot affect the other permission surfaces.
 const AccessApprovalRequestConditionsSchema = z
   .object({
-    environment: z.string().min(1),
-    secretPath: z.object({ $glob: PermissionConditionSchema[PermissionConditionOperators.$GLOB] }).strict()
+    environment: slugSchema({ max: 64, field: "Environment slug" }),
```
scott-ray-wilson: from claude locally: 
`access-approval-request-fns.ts:50` uses `slugSchema`, which trims. Validation and the environment lookup see `dev`, but the service persists the raw body at `access-approval-request-service.ts:357` and copies it verbatim into the granted privilege at lines 952 and 967. A request with `" dev "` is created, approved, and grants nothing, silently. Before the PR it failed with a 404. Either validate without trim or persist the parsed value.

scott-ray-wilson (changes requested):

backend/src/ee/services/permission/org-permission.ts
Thiago-AS: For new-privilege-system orgs the answer is yes. Any custom role with member:create/member:edit but no grant-privileges stops being able to invite with a role or change roles. We have little to no use of custom roles, we could do a migration to add these on the DB for those cases, but I do prefer to explicitly break so they have to go and manually add to the custom role if they want (more explicit, less boundary issues). WDYT?

frontend/src/pages/project/AccessControlPage/components/MembersTab/components/AddMemberModal.tsx
Thiago-AS: So my reasoning here is, we gate on the FE based on the permission we have for the route, the boundary itself we let fail so the user is aware of the change.

Thiago-AS: image

backend/src/ee/services/permission/org-permission.ts
scott-ray-wilson: I'm fine with it but we should consult the team 🙏

backend/src/ee/services/access-approval-request/access-approval-request-fns.ts
```diff
@@ -134,6 +148,17 @@ const parseAccessApprovalRequestPermissions = (permissions: TUnpackedAccessAppro
   });
 
 export const verifyRequestedPermissions = ({ permissions }: TVerifyPermission) => {
+  try {
+    validateHandlebarTemplate("Access Request Permissions", JSON.stringify(permissions ?? []), {
```
scott-ray-wilson, line 152: from claude after I shared the linear ticket:

the template half of the approval-grant gap is closed here, but the privilege-boundary half is still open in the service.

`reviewAccessApprovalRequest` calls `verifyRequestedPermissions` at `access-approval-request-service.ts:748` and rethrows on approve, so the check added here does now run on the approval path. But the two grants that follow, `additionalPrivilegeDAL.create` at `access-approval-request-service.ts:947` and `:962`, run with no privilege comparison at all: the file has zero references to `assertRoleSetBoundary`, `assertPermissionBoundary` or `validatePrivilegeChangeOperation`.

The direct surface writing to the same table does bound it, at `project-additional-privilege-factory.ts:238`, `:266` and `:342`. So approving a request is a way to create an additional privilege that the equivalent direct call would reject.

The tightened schema in this file limits the blast radius a lot (six subjects, fixed actions, only `environment` / `secretPath` conditions), so this is narrower than it was. What remains is that an approver can confer access on the requester that the approver does not hold.

backend/src/ee/services/access-approval-request/access-approval-request-fns.ts
scott-ray-wilson, line 1: ignore file placement:

from claude after I shared the linear ticket:

Two legs from the security ticket are not covered by this PR. Flagging so the ticket is not closed on the strength of this merge.

**Role writes have no privilege boundary.** `role-service.ts` `createRole` (86) and `updateRole` (123) apply `validateHandlebarTemplate` at 106 and 153 and nothing else. The scope factories only do a plain permission check: `org-role-factory.ts:45` and `:57`, `project-role-factory.ts:53` and `:66`. Nobody is compared against the permissions being written, so a role manager can author a role granting more than they hold and then assign it.

**The approval-time additional-privilege grant is unbounded.** Detail in the inline comment on `access-approval-request-fns.ts`.

Both are pre-existing rather than regressions from this PR, so they are reasonable follow-ups. Two notes for whoever picks them up:

- The reference implementation named in the ticket has moved. The bounded role write is now `assertRoleSetBoundary` with `OrgPermissionMemberActions.GrantPrivileges` at `org-service.ts:921`, not the old `assertPermissionBoundary` snippet.
- The ticket's "do not do" guidance is stale. It said to leave the eight `role?.name !== "no-access"` guards alone as dead code. This PR removed all eight and moved every site to pre-resolution filtering via `filterRolesNeedingPrivilegeBoundary`, which is the shape the ticket flagged as the real exemption. That is fine here, because no-access confers nothing and the target-side boundary now runs unconditionally, but the ticket text should be updated.

scott-ray-wilson (dismissed):

scott-ray-wilson (approved):
lgtm!

PR #7918, opened 2026-09-01
feat: adds Ed25519, Ed25519ph algorithms for sign/verify operation

Shaik-Sirajuddin:
Adds Ed25519, Ed25519ph algorithms for sign/verify operation

# Approach

OpenSSL lacks Ed25519ph support via CLI, available via C interface.
Created a minimal C executable to expose Ed25519ph sign/verify methods.
Patched Docker containers to bundle the binary.

# Followups required
FIPS Docker images use < 3.5 OpenSSL image variants. Ed25519ph variant is FIPS enabled starting in OpenSSL 3.5+.
Bumping it is needed to enable support for FIPS image. API detects and sends `unsupported` response in this case.
Ed25519ph itself is FIPS approved.

## Verification

2 algorithm compliance strategies are used.

Cross interoperability of same key with AWS, sign/verify algorithms:

* Sign at Infisical, verify at AWS work for Ed25519, Ed25519ph.
* AWS does dual hash Ed25519ph, needed a dual hash at client side before verification.
  This is not mandated by standard, Infisical doesn't do dual hash due to same reason.
* Also validated scenarios of Infisical verify, AWS sign, etc.

Cross interoperability with noble/curves Ed25519ph methods:

* Sign, verify with same keys at Infisical API, noble/curves cross-use pass.

Added tests to cover new algorithms, excluding one-time test cases.

Algorithm selection

Signing with Ed25519 SHA-512

Successful Ed25519 SHA-512 signature

Signing with Ed25519ph and prehashed input

Verifying an Ed25519ph signature

# Note for maintainers

PR has been handwritten, including all of backend. Use of AI was restricted to parts of verifications, some of frontend changes.

### DIFF
--- a/.github/workflows/publish-fips-toolchain.yml
+++ b/.github/workflows/publish-fips-toolchain.yml
@@ -3,7 +3,7 @@ name: "Publish FIPS test toolchain image"
 # Builds backend/Dockerfile.fips-toolchain and publishes it to GHCR so that
 # run-backend-tests.yml does not have to compile two OpenSSL trees on every PR.
 #
-# Tagged with the content hash of the toolchain Dockerfile: run-backend-tests.yml
+# Tagged with the content hash of the toolchain Dockerfile and OpenSSL helper sources: run-backend-tests.yml
 # computes the same hash and pulls that exact tag, falling back to building from
 # source when a PR changes the toolchain and the new tag isn't published yet.
 #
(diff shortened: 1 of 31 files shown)


Shaik-Sirajuddin:
Hi @sheensantoscapadngan 

Could you have a look , when you get a chance

PR #7928, opened 2026-09-01
feat(vault-migration): add support for folder creation

adilsitos:
## Context
This adds a new functionality for how secrets can be migrated from vault into infisical. This also fixes two previous bugs that were found: 
- Vault migration was not allowed in case the mount (i.e. secret engine) had more than one segment
- Add a maximum amount of paths that can be processed by the migration, so we don't overload vault instances and neither the creation of new folders and secrets with the new feature. 

Architecture of the folder creation: 
- A transaction is opened
- first we check the permissions of the user performing the migration. If folders need to be created, the folder creation permission is checked. If only secrets are required to be created, only the permission to create secrets is validated. Both of them can happen on just one call to migration
-  all the folders are created, then we create all the secrets
- any error will rollback the transaction, so we don't have mixed results 
- If the secret/path creation fall into a path that has a policy, the secret request will be created. 
- all the triggers that happen on the create secret are triggered after everything is committed

## Screenshots
With preserve structure 
image

Without preserve structure 

image

## Steps to verify the change

Create a setup with Vault (feel free to loop me in) and create several secrets with different paths 

### Scenario 1: Create policies in some folders 
- Try to import into folders that have policies and folders without policies. 
- Expected: the folders without policies receive the secrets. The folders with policies receive an approval request

### Scenario 2: Secret engine with more than one path (preserve folder) -- Previous bug
- Create a new secret engine on vault, with more than one segment (something like `kv/engine`)
- Try to import
- expected: everything is imported 

### Scenario 3: Secret engine with more than one path (without preserve folder) -- Previous bug
- Create a new secret engine on vault, with more than one segment (something like `kv/engine`)
- Try to import
- expected: everything is imported 

### Scenario 4: Folders already exist
- Try to import folders that already exist 
- expected: everything is imported, no new folder is created

### Scenario 5: Some folders exist, some does not
- Try to import folders that already exist 
- expected: everything is imported, folders that don't exist are created. The folders that already exist continue there

### Scenario 6: Permission (secrets)
- Try to import with a user that does not have permission to create secrets on a nested folder. Tip: use folder rbac for easily removing the permisison on a specific folder
- expected: Fail to import. Nothing is imported 

### Scenario 7: Permission (folders)
- Try to import paths that don't exist inside Infisical
- User cannot have the permission to create folders 
- expected: Fail to import. Nothing is imported
## Type

- [ ] Fix
- [x] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -2135,45 +2168,59 @@ export const secretV2BridgeServiceFactory = ({
...
-      secretPath,
-      deduplicatedSecrets,
-      project.secretDetectionIgnoreValues || []
-    );
+    // requestMemoize can hand back a value read outside a caller-supplied transaction, so it is bypassed here
+    const project = providedTx
+      ? await projectDAL.findById(projectId, providedTx)
+      : await requestMemoize(requestMemoKeys.projectFindById(projectId), () => projectDAL.findById(projectId));
+
+    if (!providedTx) {
```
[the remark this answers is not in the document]
adilsitos: this is fine. Secret scan is currently being evaluated in another place, so later we can return this.

adilsitos:
@greptile review it again (check the comments I added on the description for more context). Update the summary

mathnogueira:
@adilsitos can you resolve the conflict? Will review this Monday

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
```diff
@@ -1973,6 +1974,114 @@ export const secretApprovalRequestServiceFactory = ({
...
+    const cfg = getConfig();
+    const approvalUrl = `${cfg.SITE_URL}${approvalPath}?requestId=${secretApprovalRequest.id}`;
+
+    await triggerWorkflowIntegrationNotification({
+      input: {
+        projectId,
+        notification: {
+          type: TriggerFeature.SECRET_APPROVAL,
+          payload: {
+            userEmail: user?.email ?? "machine-identity",
```
mathnogueira: This type also has an `machineIdentityId`. You can use it instead of this in case of machine identities

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
```diff
@@ -1973,6 +1974,114 @@ export const secretApprovalRequestServiceFactory = ({
     return secretApprovalRequest;
   };
 
+  const dispatchSecretApprovalRequestCreateSideEffects = async ({
```
mathnogueira: I don't fully understand this name. Can't it be simply `createSecretApprovalRequest`?

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
mathnogueira: e.g.

```ts
userEmail: user?.email ?? undefined,
machineIdentityId: actor === ActorType.IDENTITY ? actorId : undefined,
```

mathnogueira (dismissed):
Check my previous comments. Rest is good

mathnogueira (approved):

PR #7936, opened 2026-09-01
feat: sync member email and name from the identity provider when the IdP is authoritative

Thiago-AS:
## Context

Identity providers identify a person using a stable identifier, so a person can be renamed at the IdP due to a legal name change, domain migration, or mailbox move without ceasing to be the same person. Infisical identifies a member by their email address, so we previously had no way to accommodate this.

As a result, our copy could become stale. Notifications continued to be sent to a mailbox that no longer existed, and audit entries written afterwards continued to record the outdated address. With SCIM, the behavior was more severe: we rejected the update entirely, leaving the provisioning job in a permanent error state. Entra can quarantine a connector after repeated failures, which can also prevent deprovisioning from functioning correctly.

This change makes the IdP authoritative for email and display name in organizations that **enforce SSO**. This is consistent with the existing behavior of SSO enforcement, which blocks password-based signup for verified domains and skips email verification during SSO sign-in.

## What changes for enforced organizations

* **SSO profile sync:** SAML, OIDC, and LDAP logins sync the asserted email and name to the account. Memberships and project access are unchanged.
* **SCIM updates:** `PATCH` and `PUT` now accept email changes, allowing IdP renames to propagate without requiring a login.
* **Self-service changes:** Members whose email is controlled by an enforced SSO organization can no longer change it manually.

Organizations that do not enforce SSO behave as before.

## Safeguards

Because an account can belong to multiple organizations:

* Sync requires SSO enforcement and a verified SSO alias.
* Both the current and new email addresses must use domains verified by the organization.
* An email cannot be changed to one already associated with another account. SCIM returns `409 Conflict`; SSO login records the conflict, preserves the email, and still syncs the name.
* Profile sync never blocks login. Failures leave the account unchanged.
* The previous email remains as an alias so in-flight logins can still resolve.

Two audit events are added: **profile synced** and **sync skipped due to conflict**.

## Local SSO test harness

The local SSO harness is documented in [`[sso.md](https://chatgpt.com/c/sso.md)`](sso.md). Unlock the EE feature gates first.

### Login-driven sync

Using OIDC with Keycloak:

1. Run `make up-dev-oidc` and `make seed-dev-oidc`, then open `http://localhost:8080`.
2. Sign in as `john@oidc.com` through SSO to create the account and alias.
3. Enable **Organization Settings → SSO → Enforce OIDC SSO** for the `oidc` organization. The `oidc.com` domain is already verified by the seed.
4. In Keycloak, change John's email to another `@oidc.com` address and sign in through SSO again. The email updates, while memberships and project access remain unchanged. The audit log records *Synced user profile from identity provider*.
5. Disable enforcement, change the email again, and sign in. The email remains unchanged.
6. Change the email to an unverified domain. The existing verified-domain check rejects the login.
7. Create another account using the target email, then rename John to that address in Keycloak. The login succeeds, John's email remains unchanged, and the audit log records *Skipped user email sync ... due to conflict*.

Repeat the login-driven tests with LDAP using `make up-dev-ldap` and `make seed-dev-ldap`.

### SCIM provisioning

Using SAML and Authentik:

1. Run `make up-dev-saml` and `make seed-dev-saml`. This uses real outbound SCIM.
2. In Authentik, change `john@saml.com`'s email and run the SCIM provider sync.
3. With SSO enforcement disabled, the sync is rejected by the immutable-email restriction.
4. Enable enforcement for the `saml` organization and sync again. John's email updates and memberships remain intact.
5. Change John's email to Alice's address and sync. Authentik receives `409 Conflict`.
6. Deactivate John in Authentik and sync. Deprovisioning succeeds, covering the case that was previously blocked.

### Self-service email changes

For an enforced organization, a member whose email uses a verified organization domain cannot change it from Personal Settings. The error explains that the identity provider controls the address.

A member whose email is outside the organization's verified domains can still change it.

## Type

- [x] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/scim/scim-service.ts
+++ b/backend/src/ee/services/scim/scim-service.ts
@@ -804,7 +804,7 @@ export const scimServiceFactory = ({
           {
             firstName: scimUser.name.givenName,
             lastName: scimUser.name.familyName,
-            ...(newEmail ? { email: newEmail, username: newEmail } : {})
+            ...(newEmail ? { email: newEmail, username: newEmail, isEmailVerified: false } : {})
           },
           tx
         );
@@ -921,7 +921,7 @@ export const scimServiceFactory = ({
           {
             firstName,
             lastName,
-            ...(newEmail ? { email: newEmail, username: newEmail } : {})
+            ...(newEmail ? { email: newEmail, username: newEmail, isEmailVerified: false } : {})
           },
           tx
         );
--- a/backend/src/ee/services/scim/scim-service.ts
+++ b/backend/src/ee/services/scim/scim-service.ts
@@ -645,7 +645,7 @@ export const scimServiceFactory = ({
     if (!assertedEmail) return null;
 
     const newEmail = sanitizeEmail(assertedEmail);
-    if (newEmail === user.email && newEmail === user.username) return null;
+    if (newEmail === user.email || newEmail === user.username) return null;
 
     if (!org.authEnforced) {
       throw new ScimRequestError({
@@ -667,7 +667,16 @@ export const scimServiceFactory = ({
       });
     }
 
-    await verifyEmailDomainOwnership({ email: newEmail, orgId: org.id, emailDomainDAL });
+    try {
+      await verifyEmailDomainOwnership({ email: newEmail, orgId: org.id, emailDomainDAL });
+    } catch (err) {
+      throw new ScimRequestError({
+        detail: `'${newEmail}' is not on a verified email domain for this organization. Add and verify the domain in Infisical, then retry.`,
+        status: 400,
+        scimType: "invalidValue",
+        error: err
+      });
+    }
 
     const conflictingUser = await userDAL.findOne({ username: newEmail });
     if (conflictingUser && conflictingUser.id !== user.id) {
@@ -684,9 +693,15 @@ export const scimServiceFactory = ({
   /**
    * The conflict check above is a read, not a lock, so a concurrent login or invite can take the
    * address first. Report that as the conflict it is instead of a 500 carrying a constraint name.
+   * Only where a rename was actually attempted: the transaction writes other rows too, and calling
+   * an unrelated unique violation an email conflict would send the IdP chasing the wrong address.
    */
-  const $toScimEmailConflictError = (err: unknown, newEmail: string) => {
-    if (err instanceof DatabaseError && (err.error as { code?: string })?.code === DatabaseErrorCode.UniqueViolation) {
+  const $toScimEmailConflictError = (err: unknown, newEmail: string | null) => {
+    if (
+      newEmail &&
+      err instanceof DatabaseError &&
+      (err.error as { code?: string })?.code === DatabaseErrorCode.UniqueViolation
+    ) {
       return new ScimRequestError({
         detail: `An Infisical account already exists for '${newEmail}'. Remove or merge that account before changing this user's email.`,
         status: 409,
@@ -820,7 +835,7 @@ export const scimServiceFactory = ({
         );
       });
     } catch (err) {
-      throw $toScimEmailConflictError(err, newEmail ?? "");
+      throw $toScimEmailConflictError(err, newEmail);
     }
 
     return scimUser;
@@ -944,7 +959,7 @@ export const scimServiceFactory = ({
         );
       });
     } catch (err) {
-      throw $toScimEmailConflictError(err, newEmail ?? "");
+      throw $toScimEmailConflictError(err, newEmail);
     }
 
     return buildScimUser({
--- a/backend/src/ee/services/scim/scim-service.ts
+++ b/backend/src/ee/services/scim/scim-service.ts
@@ -781,7 +781,7 @@ export const scimServiceFactory = ({
     const newEmail = await $resolveScimEmailChange({
       org,
       user,
-      assertedEmail: scimUser.emails?.[0]?.value
+      assertedEmail: (scimUser.emails?.find((email) => email.primary) ?? scimUser.emails?.[0])?.value
     });
 
     await verifyEmailDomainOwnership({
--- a/backend/src/ee/services/scim/scim-service.ts
+++ b/backend/src/ee/services/scim/scim-service.ts
@@ -11,6 +11,7 @@ import {
   TableName,
   TGroups,
   TMemberships,
+  TOrganizations,
   TUsers
 } from "@app/db/schemas";
 import { TGroupDALFactory } from "@app/ee/services/group/group-dal";
@@ -22,7 +23,9 @@ import { TScimDALFactory } from "@app/ee/services/scim/scim-dal";
 import { PgSqlLock } from "@app/keystore/keystore";
 import { getConfig } from "@app/lib/config/env";
 import { crypto } from "@app/lib/crypto";
-import { BadRequestError, NotFoundError, ScimRequestError, UnauthorizedError } from "@app/lib/errors";
+import { DatabaseErrorCode } from "@app/lib/error-codes";
+import { BadRequestError, DatabaseError, NotFoundError, ScimRequestError, UnauthorizedError } from "@app/lib/errors";
+import { unique } from "@app/lib/fn";
 import { logger } from "@app/lib/logger";
 import { ms } from "@app/lib/ms";
 import { alphaNumericNanoId } from "@app/lib/nanoid";
@@ -88,7 +91,7 @@ type TScimServiceFactoryDep = {
     TUserDALFactory,
     "find" | "findOne" | "create" | "transaction" | "findUserEncKeyByUserIdsBatch" | "findById" | "updateById"
   >;
-  userAliasDAL: Pick<TUserAliasDALFactory, "findOne" | "create" | "delete" | "update" | "find">;
+  userAliasDAL: Pick<TUserAliasDALFactory, "findOne" | "create" | "delete" | "update" | "updateById" | "find">;
   orgDAL: Pick<
     TOrgDALFactory,
     | "createMembership"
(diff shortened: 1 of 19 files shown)


backend/src/services/user-alias/user-alias-fns.ts
```diff
@@ -93,6 +98,194 @@ export const ensureSsoAccountVerified = async ({
...
+        event: {
+          type: EventType.SSO_USER_PROFILE_SYNC_CONFLICT,
+          metadata: {
+            userId: user.id,
+            aliasType: userAlias.aliasType,
+            externalId: userAlias.externalId,
+            currentEmail: user.username,
+            assertedEmail: assertedUsername,
+            conflictingUserId
+          }
```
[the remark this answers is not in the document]
Thiago-AS: We also have the emails there.

backend/src/services/user-alias/user-alias-fns.ts
```diff
@@ -93,6 +98,194 @@ export const ensureSsoAccountVerified = async ({
...
+      orgId,
+      event: {
+        type: EventType.SSO_USER_PROFILE_SYNCED,
+        metadata: {
+          userId: user.id,
+          aliasType: userAlias.aliasType,
+          externalId: userAlias.externalId,
+          ...(isEmailApplied ? { previousEmail: user.username, newEmail: assertedUsername } : {}),
+          ...(nameChanges.firstName ? { previousFirstName: user.firstName, newFirstName: nameChanges.firstName } : {}),
+          ...(nameChanges.lastName ? { previousLastName: user.lastName, newLastName: nameChanges.lastName } : {})
```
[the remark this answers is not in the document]
Thiago-AS: That intended.

docs/documentation/platform/sso/overview.mdx
```diff
@@ -77,6 +78,13 @@ This portal is accessible at `/login/admin` (e.g., https://app.infisical.com/log
...
+        With SSO enforcement on, the next login updates that member's Infisical email address and name to match the assertion. The account, its memberships, and its access are unchanged, and audit log entries written from then on carry the new address. Entries written earlier keep the address that was current when they were recorded.
+
+        If your organization provisions through [SCIM](/documentation/platform/scim/overview), the change arrives on the provisioning job instead, without waiting for the member to log in.
+
+        When another Infisical account already uses the new address, Infisical leaves the email address as it is and records the conflict in the audit log. Remove or merge the other account, and the change applies on the next login.
+    </Accordion>
     <Accordion title="Why do I get redirected to SSO when trying to use the Admin Login Portal?">
         You are likely being redirected because you do not have email authentication mode enabled, or you're not an **Organization Admin**. This portal requires **Organization Admin** status and direct credential login (email and password). **Server Admin** status alone is insufficient.
```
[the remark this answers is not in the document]
Thiago-AS: 1. It's not a line changed. Line 89 is a context line in the diff, not an addition. It only shows up because your new accordion was inserted directly above it. The bot is reviewing pre-existing prose that shifted position.

2. Its premise is factually wrong. I checked the whole docs tree:
- "email auth" appears nowhere in docs/. Its only match is inside "email authentication" on the very line being flagged. So the claim that it's "the authentication method documented elsewhere" is false, there is no "elsewhere."
- "direct credential login" likewise appears only on that line, so it isn't inconsistent with anything either.
- The actual house term across the docs is "email and password", auth-methods/email-password.mdx, cli/commands/login.mdx:44, getting-started/cli.mdx:95, documentation/platform/email-domain.mdx:177.

backend/src/services/user-alias/user-alias-fns.ts
```diff
@@ -93,6 +98,194 @@ export const ensureSsoAccountVerified = async ({
...
+  isAuthEnforced,
+  userDAL,
+  userAliasDAL,
+  emailDomainDAL,
+  auditLogService
+}: TSyncSsoUserProfileDTO): Promise<TUsers> => {
+  if (!isAuthEnforced || !userAlias.isEmailVerified) return user;
+
+  const assertedUsername = sanitizeEmail(assertedEmail);
+  const isEmailChanged = Boolean(assertedUsername) && assertedUsername !== user.username;
```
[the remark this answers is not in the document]
Thiago-AS: This is pre existing on main, won't change that for now to avoid blast radius on the PR.

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -621,6 +624,94 @@ export const scimServiceFactory = ({
...
+    assertedEmail
+  }: {
+    org: Pick<TOrganizations, "id" | "authEnforced">;
+    user: Pick<TUsers, "id" | "email" | "username">;
+    assertedEmail?: string | null;
+  }): Promise<string | null> => {
+    if (!assertedEmail) return null;
+
+    const newEmail = sanitizeEmail(assertedEmail);
+    if (newEmail === user.email || newEmail === user.username) return null;
```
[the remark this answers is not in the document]
Thiago-AS: Same as above.

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -687,51 +778,65 @@ export const scimServiceFactory = ({
...
             firstName: scimUser.name.givenName,
-            email: scimUser.userName,
             lastName: scimUser.name.familyName,
-            active: scimUser.active
-          }
-        },
-        tx
-      );
-    });
+            ...(newEmail ? { email: newEmail, username: newEmail } : {})
```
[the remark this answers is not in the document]
Thiago-AS: Fixed in 32e1536c75. both SCIM rename paths now clear isEmailVerified, which
gates resetPasswordV2

backend/src/services/user-alias/user-alias-fns.ts
```diff
@@ -93,6 +98,194 @@ export const ensureSsoAccountVerified = async ({
...
+
+    if (!isOrgOwnedRename) {
+      isEmailApplied = false;
+      logger.warn(
+        { userId: user.id, orgId, externalId: userAlias.externalId },
+        `Skipped SSO email sync, the organization does not own both addresses [userId=${user.id}] [orgId=${orgId}]`
+      );
+      if (!isNameChanged) return user;
+    }
+  }
```
victorvhs017, line 216: Even though this is theoretically impossible now, it could be possible in the future. We should audit log this skipped sync too, as we do in the other paths.

backend/src/services/user-alias/user-alias-fns.ts
victorvhs017, line 1: If we update the email, pending email invites would use the old one; we should update the emails in the invites to the new email.

The invites in this case could be from other orgs, as the flow requires a user to have a verified email.

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -621,6 +624,94 @@ export const scimServiceFactory = ({
...
+    const conflictingUser = await userDAL.findOne({ username: newEmail });
+    if (conflictingUser && conflictingUser.id !== user.id) {
+      throw new ScimRequestError({
+        detail: `An Infisical account already exists for '${newEmail}'. Remove or merge that account before changing this user's email.`,
+        status: 409,
+        scimType: "uniqueness"
+      });
+    }
+
+    return newEmail;
```
victorvhs017, line 690: I think we should also audit log this one, as we do for SSO

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -621,6 +624,94 @@ export const scimServiceFactory = ({
...
+      (err.error as { code?: string })?.code === DatabaseErrorCode.UniqueViolation
+    ) {
+      return new ScimRequestError({
+        detail: `An Infisical account already exists for '${newEmail}'. Remove or merge that account before changing this user's email.`,
+        status: 409,
+        scimType: "uniqueness",
+        error: err
+      });
+    }
+    return err;
```
victorvhs017, line 712: Same here

backend/src/ee/services/scim/scim-service.ts
```diff
@@ -794,61 +899,73 @@ export const scimServiceFactory = ({
...
             firstName,
-            email: email?.toLowerCase(),
             lastName,
-            active
-          }
-        },
-        tx
-      );
-    });
+            ...(newEmail ? { email: newEmail, username: newEmail, isEmailVerified: false } : {})
```
victorvhs017, line 924: Setting `isEmailVerified: false` on SCIM rename only blocks the last step when the user selects "Change your password" in account recovery (`resetPasswordV2`). It does not block requesting a recovery code or completing email verification.

For an SSO-only account renamed to an address the actor controls: they can complete recovery, use Domain or SSO change to enable email sign-in, then use Change your password, which still fails until the account is marked verified again (happens on the next SSO login, not in the recovery UI).

victorvhs017 (changes requested):

backend/src/services/user-alias/user-alias-fns.ts
Thiago-AS: Accepting the invite is unaffected. The user is looked up by their current email, and the membership is tied to the user ID, so `inviteEmail` is not involved.

Manual resends are also unaffected and already use the user’s current email address.

The issue is with reminder emails. After an email change, reminders are sent to the old address, and the invite link no longer works because it references the outdated email. This results in misdelivered reminders and a broken invite for anyone who follows the link.

The fix is to use the user’s current email address for reminders as well. This also keeps both invite flows consistent and avoids maintaining a duplicate email address.

This exists in the current change email flow as well, so I'll leave this fix to another PR.

backend/src/ee/services/scim/scim-service.ts
Thiago-AS: The important boundary here is domain ownership. SCIM email changes already require SSO enforcement and DNS-verified domains for both the old and new addresses, along with the existing account-conflict and MFA checks where applicable. `isEmailVerified` is not part of that security boundary.

Its purpose is to indicate whether control of `users.email` has been proven. For a SCIM rename, it hasn’t. A SCIM rename is a provisioning event, not an authentication event, so clearing the flag is a correctness property, not an anti-takeover control.

I’m deliberately not applying the same behavior to SSO profile sync. That path runs during an authenticated SSO login and uses an address asserted by the IdP, which is the same evidence used to restore the verified state. Clearing the flag there would be inconsistent and immediately undone on the next login.

The separate issue worth addressing is permanently enabling email authentication using only a recovery token, without considering whether the account is IdP-managed. Any fix needs to preserve the existing recovery path for cases where SSO is unavailable, so a blanket restriction would not work.

I’d leave the recovery flow itself unchanged. Gating recovery on `isEmailVerified` could lock out legitimate accounts, including SCIM-provisioned users, and recovery is intentionally the escape hatch for locked-out accounts.

backend/src/ee/services/scim/scim-service.ts
Thiago-AS: I added this only for SSO because there was a silent error. This one is a throwable, actionable error that is surfaced to the providers, so I don’t think we need to audit log it here.

backend/src/ee/services/scim/scim-service.ts
Thiago-AS: I added this only for SSO because there was a silent error. This one is a throwable, actionable error that is surfaced to the providers, so I don’t think we need to audit log it here.

backend/src/services/user-alias/user-alias-fns.ts
victorvhs017: The old invite won't work because the old email is baked into the invite link. But you're right that resend works fine! So it shouldn't be a big blocker, as you have the inviter's email in the email template, and you can just ask for a new one because your email changed.

backend/src/ee/services/scim/scim-service.ts
victorvhs017: I agree that the SCIM provider will have their log, but the sync is an action on our side. Also, to keep consistency between both places, since the sync is basically the same.

backend/src/ee/services/scim/scim-service.ts
victorvhs017: > The separate issue worth addressing is permanently enabling email authentication using only a recovery token, without considering whether the account is IdP-managed. Any fix needs to preserve the existing recovery path for cases where SSO is unavailable, so a blanket restriction would not work.

This is the part that could become another task.

I did the test again, and the attack vector here is really small; basically, it lets an IdP admin have access to any other org you have access to through that email. Which doesn't make sense, in a way, because they can just reset your password whenever they want and access it.

I don't see any other way that someone with access to the email to get the recovery code and the information that the IdP user was updated, could not already have another level of privileged access in the company.

victorvhs017 (approved):

PR #7939, opened 2026-09-02
feat: add pki license gating

sheensantoscapadngan:
## Context
This PR adds PKI license enforcement.

image

image

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/db/migrations/20260903051600_add-certificate-quota-key-index.ts
```diff
@@ -0,0 +1,56 @@
...
+          `SELECT 1 FROM pg_class c JOIN pg_index i ON i.indexrelid = c.oid WHERE c.relname = '${QUOTA_KEY_INDEX}' AND NOT i.indisvalid`
+        );
+        if (invalid.rows.length > 0) {
+          await raw(`DROP INDEX CONCURRENTLY IF EXISTS "${QUOTA_KEY_INDEX}"`);
+        }
+
+        await raw(`
+          CREATE INDEX CONCURRENTLY IF NOT EXISTS "${QUOTA_KEY_INDEX}"
+          ON "${TableName.Certificate}" ("projectId", "quotaKey")
+          INCLUDE ("notAfter", "status", "extendedKeyUsages")
```
sheensantoscapadngan: why do we include extendedKeyUsages though?

backend/src/keystore/keystore.ts
```diff
@@ -186,6 +186,11 @@ export const KeyStorePrefixes = {
   LicenseUsageReconcileMarker: (orgId: string) => `license-usage-reconcile-${orgId}` as const,
   LicenseUsageLastReported: (orgId: string, featureKey: string) =>
     `license-usage-last-reported-${orgId}-${featureKey}` as const,
+  // Cached distinct active certificate quota keys per org tree. Not LicenseUsageLastReported, which
+  // counts rows and is only written on a successful License Server report.
```
sheensantoscapadngan: why do you even mention LicenseUsageLastReported?

backend/src/services/certificate-common/certificate-quota-key.ts
```diff
@@ -0,0 +1,20 @@
+import { createHash } from "node:crypto";
+
+// Identity of "one logical certificate" for quotas. orderId already collapses renewal chains; this
+// exists for re-enrollment, where ACME mints a new order for the same names every cycle.
```
sheensantoscapadngan: I don't think it's even worth mentioning the orderID here in the comment...

backend/src/services/license-client/usage/usage-counter-dal.ts
```diff
@@ -50,6 +53,78 @@ export const usageCounterDALFactory = (db: TDbClient) => {
...
+      const row = (await db
+        .replicaNode()(TableName.Organization)
+        .where("id", orgId)
+        .select(db.raw(`COALESCE("rootOrgId", "id") as "rootOrgId"`))
+        .first()) as { rootOrgId?: string } | undefined;
+      return row?.rootOrgId ?? orgId;
+    } catch (error) {
+      throw new DatabaseError({ error, name: "Resolve root org id" });
+    }
+  };
```
sheensantoscapadngan, line 86: Hold up, don't we already implement this somewhere? I think it's in the org tree file? or is that different thing altogether?

backend/src/services/certificate-authority/certificate-authority-service.ts
```diff
@@ -326,15 +328,32 @@ export const certificateAuthorityServiceFactory = ({
...
+    // Two independent caps, whichever binds first: maxCas covers every type (the free tier's single
+    // "internal or ACME" allowance), maxInternalCas covers INTERNAL only and is the per-contract
+    // enterprise lever. In practice a tier sets one and leaves the other null.
+    if (typeof plan.maxCas === "number") {
+      const currentCaCount = await certificateAuthorityDAL.countCasByOrgId(actor.orgId);
+      if (currentCaCount >= plan.maxCas) {
+        throw new BadRequestError({
+          message: `Failed to create certificate authority due to plan limit reached (${currentCaCount} of ${plan.maxCas} certificate authorities). Upgrade plan to add more certificate authorities.`
+        });
+      }
```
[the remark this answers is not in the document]
sheensantoscapadngan: this doesn't have to be real-time. we'll be able to track for billing at the end of the day

backend/src/services/certificate-common/certificate-quota-fns.ts
```diff
@@ -0,0 +1,156 @@
...
+  if (isAlreadyActive) return { quotaKey, isNewQuotaKey: false, quotaOrgId, isWildcard };
+
+  // 0 is how a plan says it has no wildcard support at all.
+  if (isWildcard && maxWildcardCertificates === 0) {
+    throw new BadRequestError({
+      message: `Failed to issue certificate with a wildcard name due to plan restriction. Upgrade plan to issue wildcard certificates.`
+    });
+  }
+
+  const counts = await $getQuotaCounts(quotaOrgId, deps);
```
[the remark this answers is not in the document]
sheensantoscapadngan: thank you

backend/src/db/schemas/certificates.ts
```diff
@@ -54,7 +54,9 @@ export const CertificatesSchema = z.object({
   hsmKeyLabel: z.string().nullable().optional(),
   hsmPublicKeySpki: zodBuffer.nullable().optional(),
   subjectDomainComponents: z.string().nullable().optional(),
-  orderId: z.string().uuid().default("00000000-0000-0000-0000-000000000000")
+  orderId: z.string().uuid().default("00000000-0000-0000-0000-000000000000"),
+  quotaKey: z.string().nullable().optional(),
```
carlosmonastyrski, line 58: We should skip these in the public API

backend/src/db/migrations/20260903051500_add-certificate-quota-key.ts
```diff
@@ -0,0 +1,66 @@
+import { Knex } from "knex";
```
carlosmonastyrski, line 1: Ignore comment placement

The error is not intuitive for the user when hitting the max CA amount (generic 400)

Image

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -567,6 +566,14 @@ export const certificateV3ServiceFactory = ({
 
     const certificateRequestWithDefaults = applyProfileDefaults(certificateRequest, profile.defaults);
```
carlosmonastyrski, line 567: When applyProfileDefaults is called with CSR, the profile defaults for SAN is ignored but the resulting key has them

backend/src/db/migrations/20260903051500_add-certificate-quota-key.ts
sheensantoscapadngan: isn't there another error message that shows up after the error 400? this might be how we're doing it throughout the platform...

backend/src/services/certificate-authority/internal/internal-certificate-authority-service.ts
```diff
@@ -172,6 +175,15 @@ export const internalCertificateAuthorityServiceFactory = ({
   const $validatePqcLicense = (keyAlgorithm: string, projectId: string) =>
     validatePqcLicense({ keyAlgorithm, projectId, projectDAL, licenseService });
 
+  // Gates only Infisical's own distribution point. Custom URLs are gated where the CA is configured.
+  const $isManagedCrlDistributionAllowed = async (projectId: string) => {
```
carlosmonastyrski, line 179: we should receive a `tx` here so when this is called inside a transaction we don't open a second pool slot

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -1086,6 +1096,14 @@ export const pkiAcmeServiceFactory = ({
       throw new AcmeBadCSRError({ message: `Invalid CSR: ${validationResult.errors.join(", ")}` });
     }
 
+    await validateCertificateRequestLicense({
```
carlosmonastyrski: Should we parse this error to AcmeError? It won't be readable by certbot or any other ACME tool otherwise

backend/src/db/migrations/20260903051500_add-certificate-quota-key.ts
```diff
@@ -0,0 +1,66 @@
...
+
```
carlosmonastyrski, line 2: Ignore comment placement
Image
We should disable this option if it's not included in the plan instead of waiting to fail on save

carlosmonastyrski: Image
Same in the other places, maybe the upgrade to enterprise modal, something that tells you clear and simple why is it blocked earlier

backend/src/services/certificate-authority/internal/internal-certificate-authority-service.ts
sheensantoscapadngan: what do you mean? this reads from redis...

also, it's not ideal to have a long-running transaction since that will exhaust the client pool. if there's a cache miss it will try and hit the license server... a slow response there would be catastrophic

backend/src/db/migrations/20260903051500_add-certificate-quota-key.ts
sheensantoscapadngan: Added it for those that we can easily check in the frontend. Thus certificate quote is explicitly left out

sheensantoscapadngan: image

backend/src/db/migrations/20260903051500_add-certificate-quota-key.ts
```diff
@@ -0,0 +1,66 @@
...
+import {
```
carlosmonastyrski, line 3: Let's invalidate `caKeys.getCaQuota()` in the create and delete mutations for both internal and external CAs, otherwise it's only after refresh you get the upgrade to enterprise message

Image

backend/src/db/migrations/20260903051500_add-certificate-quota-key.ts
```diff
@@ -0,0 +1,66 @@
...
+  buildCertificateQuotaKey,
```
carlosmonastyrski, line 4: small nit, should be 1 certificate authority here

Image

backend/src/services/certificate-authority/internal/internal-certificate-authority-service.ts
carlosmonastyrski: Pointing at the `const project = await projectDAL.findById(projectId);` query, it's already being called from other functions where a tx is already open
Worst thing here would be to open 2 connections at the same time, true that the slow response can be worse, we can leave it like that 👍

backend/src/services/license-client/usage/usage-event-queue.ts
```diff
@@ -64,18 +57,21 @@ export const usageEventQueueFactory = ({
...
+    const reportOrgId = isCloud ? ((await metered.resolveReportOrgId?.(orgId)) ?? orgId) : SELF_HOSTED_LICENSE_ORG_ID;
 
     try {
-      const metered = featureByKey.get(dimensionKey);
-      if (!metered) {
-        logger.warn(`usage-metering: unknown metered feature, dropping event [dimensionKey=${dimensionKey}]`);
-        return;
-      }
-
       const value = await metered.count(reportOrgId);
```
carlosmonastyrski, line 75: We should add a constrain in this count as right now it can be SELF_HOSTED_LICENSE_ORG_ID and if the query goes into the DB it'll error

Image

carlosmonastyrski: Blocked by this on the self hosted testing, will finish the review after the fix 👍

carlosmonastyrski (dismissed):

carlosmonastyrski (approved):

PR #7943, opened 2026-09-02
feat: support custom X.509 extensions on certificate policies and profiles

carlosmonastyrski:
## Context

Adds support for arbitrary X.509 extensions on issued certificates, so a certificate can carry an OID beyond the explicitly mapped set (key usages, extended key usages, basic constraints, SANs).

A certificate policy declares one rule per object identifier with allow, require or deny, a value pattern and an optional criticality pin, following the same deny-by-default posture as SANs: unconfigured permits any OID, an empty configuration permits none, and a populated list permits only what it names.

A certificate profile supplies default values, and a request can carry its own, either through the API or in the extensions of a supplied CSR, which is what lets an MDM set a per-device value over SCEP.

Also adds the Smart Card Logon extended key usage, which AD requires for PKINIT, and fixes UPN subject alternative names being emitted as email addresses by the internal CA because the requested SAN type was discarded before reaching it.

## Type

- [x] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


carlosmonastyrski:
@greptile re-review and update the original Greptile Summary

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+const TABLES_WITH_CUSTOM_EXTENSIONS = [
```
bernie-g, line 5: ignore comment placement

Image
Can we add a tooltip explaining why this is greyed out for presets

bernie-g: Also would it make sense to make the Critical/Not Critical a checkbox instead of a dropdown in the profile

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
```
bernie-g, line 3: ignore comment placement

Image

Can we rename this button to "Add extension rule"

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
...
+  TableName.PkiCertificatePolicy,
+  TableName.CertificateRequests,
+  TableName.Certificate
+] as const;
+
+export async function up(knex: Knex): Promise<void> {
+  for await (const tableName of TABLES_WITH_CUSTOM_EXTENSIONS) {
+    if (await knex.schema.hasTable(tableName)) {
+      if (!(await knex.schema.hasColumn(tableName, "customExtensions"))) {
+        await knex.schema.alterTable(tableName, (t) => {
```
bernie-g, line 15: ignore comment placement


I think "Only allow the custom extension OIDs you configure here" is a bit misleading since you can deny OIDs there too. Can we change it to something like "Certificates may carry only the OIDs listed here, each subject to its rule."

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
+  for await (const tableName of TABLES_WITH_CUSTOM_EXTENSIONS) {
+    if (await knex.schema.hasTable(tableName)) {
+      if (await knex.schema.hasColumn(tableName, "customExtensions")) {
+        await knex.schema.alterTable(tableName, (t) => {
+          t.dropColumn("customExtensions");
```
bernie-g, line 28: ignore comment placement

I noticed that in the policy there's no way to restrict if the OID's value has to be set in the profile or can be submitted in the CSR. Do you think this is necessary?

bernie-g, line 28: Since you can't create multiple custom extension rules for the same OID you can't have multiple value regexes, so someone who for example has multiple domains in their AD wouldn't be able to allow multiple different SIDs as the value. However you can do this just fine for SANs

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+const TABLES_WITH_CUSTOM_EXTENSIONS = [
+  TableName.PkiCertificatePolicy,
+  TableName.CertificateRequests,
+  TableName.Certificate
```
bernie-g, line 8: Image
"Declare extension" seems a bit odd to me for the button name, can we just do "Add extension" to keep consistency with the rest of the UI?

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+const TABLES_WITH_CUSTOM_EXTENSIONS = [
+  TableName.PkiCertificatePolicy,
```
bernie-g, line 6: nit: Can't upload a picture because this plane wifi is terrible but in the cert profile creation could we validate the OID value right there in the textbox with an error instead of having it only show on submit? We do this in a lot of other places and it's better UX for the user imo

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
...
+  TableName.CertificateRequests,
+  TableName.Certificate
+] as const;
+
+export async function up(knex: Knex): Promise<void> {
+  for await (const tableName of TABLES_WITH_CUSTOM_EXTENSIONS) {
```
bernie-g, line 12: Ignore comment placement

In the cert profile creation dialog, custom extensions is in between subject attributes and SAN which seems kind of random. I feel like this is a pretty uncommon setting so maybe we should put it at the bottom?

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (!(await knex.schema.hasColumn(tableName, "customExtensions"))) {
+        await knex.schema.alterTable(tableName, (t) => {
+          t.jsonb("customExtensions");
+        });
+      }
+    }
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
```
bernie-g, line 23: Ignore comment placement

Image

I feel like it makes more sense for the value textbox to be wider than the name one

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
...
+  for await (const tableName of TABLES_WITH_CUSTOM_EXTENSIONS) {
+    if (await knex.schema.hasTable(tableName)) {
+      if (await knex.schema.hasColumn(tableName, "customExtensions")) {
+        await knex.schema.alterTable(tableName, (t) => {
+          t.dropColumn("customExtensions");
+        });
```
bernie-g, line 29: For the custom extension rules in the policy editor, do you think deny is needed here? I'm a bit confused about the custom extension rules since it seems that we're operating on both an allowlist and a denylist. If you have "Restrict custom extensions" set to on then add a DENY rule for an OID that OID is never allowed to be added UNLESS it matches a specific value. Seems very unintuitive to me. I think we should reconsider if we need deny at all. Also possibly change the rules to be more like the SANs rules

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
```diff
@@ -0,0 +1,33 @@
...
+      }
+    }
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
+  for await (const tableName of TABLES_WITH_CUSTOM_EXTENSIONS) {
+    if (await knex.schema.hasTable(tableName)) {
+      if (await knex.schema.hasColumn(tableName, "customExtensions")) {
+        await knex.schema.alterTable(tableName, (t) => {
```
bernie-g, line 27: Ignore comment placement

Should we enable "Restrict custom extensions" by default? Seems that it might be pretty lax to have it off by default

backend/src/db/schemas/certificate-requests.ts
```diff
@@ -42,7 +42,8 @@ export const CertificateRequestsSchema = z.object({
   encryptedPrivateKey: zodBuffer.nullable().optional(),
   pendingMessage: z.string().nullable().optional(),
   applicationId: z.string().uuid().nullable().optional(),
-  domainComponents: z.string().nullable().optional()
+  domainComponents: z.string().nullable().optional(),
+  customExtensions: z.unknown().nullable().optional()
```
bernie-g, line 46: Claude:
In CertificateRenewalModal.tsx, CSR mode returns ["setup", "csr", "extensions"], and the CSR step's help text was updated to say "custom extensions on the next step." But certificateRenewalUtils.ts:157 returns { ttl } and exits before customExtensions is ever attached. The user fills in the step, hits renew, and gets a cert without the values, no error.

The backend deliberately went the other way: customExtensions is marked csrEditable: true in RENEWAL_ATTRIBUTES so it can be set alongside a CSR. So this is just the frontend not following through, and the two sides now disagree about the rule.

bernie-g (changes requested):

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -3574,6 +3575,7 @@ interface UpdateCertificatePolicy {
     | {
         certificatePolicyId: string;
         name: string;
+        customExtensionOids?: string[];
```
sheensantoscapadngan, line 3593: IGNORE PLACEMENT
having it as one value per row looks a bit odd. Would it make sense for us to make it so that it's one extension per row instead?

Image
Image

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -32,8 +32,12 @@ import { ActorType } from "@app/services/auth/auth-type";
 import { TCertificateBodyDALFactory } from "@app/services/certificate/certificate-body-dal";
 import { CertSubjectAlternativeNameType } from "@app/services/certificate/certificate-types";
 import { TCertificateAuthorityDALFactory } from "@app/services/certificate-authority/certificate-authority-dal";
-import { CaType } from "@app/services/certificate-authority/certificate-authority-enums";
+import { CaCapability, CaType } from "@app/services/certificate-authority/certificate-authority-enums";
 import { assertCaInProfileProject } from "@app/services/certificate-authority/certificate-authority-fns";
+import {
+  caSupportsCapability,
+  CERTIFICATE_AUTHORITIES_TYPE_MAP
```
sheensantoscapadngan, line 39: IGNORE PLACEMENT

perhaps we disable adding multiple extensions of the same OID. Make it so that you can't declare a new one anymore instead of erroring on submit?

Image

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -124,6 +128,14 @@ import {
  */
 const STALE_PROCESSING_ORDER_TIMEOUT_MS = 60 * 60 * 1000;
 
+const assertAcmeCaSupportsCustomExtensions = (caType: CaType, count: number): void => {
```
sheensantoscapadngan, line 134: IGNORE PLACEMENT

Do we anticipate people using these custom extensions more commonly than subject alternative names? it seems like this is something that we can have at the bottom-most section after extended key usages...

Image

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -124,6 +128,14 @@ import {
...
+  if (!count || caSupportsCapability(caType, CaCapability.CUSTOM_EXTENSIONS)) return;
+
+  throw new AcmeBadCSRError({
+    message: `Invalid CSR: ${CERTIFICATE_AUTHORITIES_TYPE_MAP[caType] ?? caType} certificate authorities cannot carry custom extensions, but this request resolved ${count} of them.`
+  });
+};
+
```
sheensantoscapadngan, line 141: Why does this have its own validation, and yet we still show the value pattern field?

Image
Image

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -1106,6 +1120,7 @@ export const pkiAcmeServiceFactory = ({
       status: CertificateRequestStatus.PENDING,
       acmeOrderId: orderId,
       csr,
```
sheensantoscapadngan, line 1141: IGNORE PLACEMENT

should we be able to edit the custom extension too? right now i have to delete and recreate with new values
Image

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -1264,15 +1282,21 @@ export const pkiAcmeServiceFactory = ({
             // key actually is, which would fail a policy the CSR itself satisfies.
             const validationResult = await certificatePolicyService.validateCertificateRequest(
               policy.id,
-              applyProfileDefaults({ ...certificateRequest, ...extractAlgorithmsFromCSR(csr) }, profile.defaults)
+              applyProfileDefaults({ ...certificateRequest, ...extractAlgorithmsFromCSR(csr) }, profile.defaults),
```
sheensantoscapadngan, line 1304: IGNORE PLACEMENT

Image

backend/src/ee/services/pki-acme/pki-acme-service.ts
```diff
@@ -1264,15 +1282,21 @@ export const pkiAcmeServiceFactory = ({
...
             }
+            assertAcmeCaSupportsCustomExtensions(
+              finalizeCaType,
+              validationResult.resolvedCustomExtensions?.length ?? 0
+            );
 
             return {
               approvalPolicy: matchedApprovalPolicy,
               policy,
+              resolvedCustomExtensions: validationResult.resolvedCustomExtensions,
```
sheensantoscapadngan, line 1318: IGNOE PLACEMENT

Image

sheensantoscapadngan: That should be reflected in the approval request as well, right? and in the display?

backend/src/db/schemas/certificates.ts
```diff
@@ -54,7 +54,8 @@ export const CertificatesSchema = z.object({
   hsmKeyLabel: z.string().nullable().optional(),
   hsmPublicKeySpki: zodBuffer.nullable().optional(),
   subjectDomainComponents: z.string().nullable().optional(),
-  orderId: z.string().uuid().default("00000000-0000-0000-0000-000000000000")
+  orderId: z.string().uuid().default("00000000-0000-0000-0000-000000000000"),
+  customExtensions: z.unknown().nullable().optional()
```
sheensantoscapadngan: IGNORE PLACEMENT

It seems like importing a certificate with a custom extension does not parse the custom extension values correctly...

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
carlosmonastyrski: Profile defaults work as a "Use this value if the request does not provide one", so for example in ACME you will be now be able to fill fields that otherwise it's impossible to set with tools like certbot.
The policy restricts what the final value is, not from where it's coming from

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
carlosmonastyrski: All policy fields are off by default. Turning it on by default would also make custom extensions the only policy dimension that defaults to restricted.

backend/src/ee/services/pki-acme/pki-acme-service.ts
carlosmonastyrski: the preset validates a concrete value, the policy field is a *pattern* and takes wildcards like `S-1-5-21-1004336348-*`. The misleading part was the error message demanding a static SID and never mentioning wildcards, fixed, along with the criticality control that said "Critical and non-critical allowed" while claiming it was spec-fixed.

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
carlosmonastyrski: Rules now use the same `{ allowed, required, denied }` shape as SAN rules, so several patterns per OID work and "allowed X except Y" is a single rule.

I kept deny: a deny-only rule still means "permitted except these values", it's now identical to how SAN rules already behave rather than a one-off

backend/src/services/certificate-common/certificate-extension-fns.ts
```diff
@@ -0,0 +1,524 @@
...
+
+  for (const oid of resolvableOids) {
+    if (isReservedExtensionOid(oid)) {
+      errors.push(describeReservedExtensionOid(oid));
+      // eslint-disable-next-line no-continue
+      continue;
+    }
+
+    const declaration = declarationByOid.get(oid) ?? { oid };
+    const rule = rulesByOid === null ? ALLOW_ANY_RULE : rulesByOid.get(oid);
```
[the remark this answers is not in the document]
carlosmonastyrski: It can be restricted as any other field by setting the restriction manually

backend/src/db/migrations/20260831164512_add-custom-extensions-to-pki-policies-and-certificates.ts
bernie-g: Yes I know, I'm asking if we should have a setting in the policy that disables CSRs from filling in the values. I think the ticket says we need this 
CleanShot 2026-09-11 at 5 08 16 PM@2x

docs/documentation/platform/pki/settings/policies.mdx
```diff
@@ -192,6 +193,25 @@ Control whether certificates can act as CAs:
...
+|------|---------|
+| Allowed | The value may match this pattern |
+| Required | The extension must be present, and its value must match one of the required patterns |
+| Denied | The value must not match this pattern, even when an allowed pattern also matches it |
+
+Add several rules for the same OID to accept more than one value. An Active Directory estate with two domains, for example, takes two Allowed rules on the security identifier extension, one wildcard per domain.
+
+Patterns are matched against the value you enter rather than its encoded form, and matching is case sensitive. Use `*` for any value, or a wildcard such as `S-1-5-21-1004336348-1177238915-682003330-*` to accept only identifiers from one domain. A rule can also pin criticality, except where an extension's specification already fixes it.
+
+If you configure Custom Extensions, only the OIDs you list are permitted. Any other OID in a request is rejected, including one carried in a signing request, so the patterns are what limit what an enrolling client can assert. An empty configuration forbids all custom extensions, and leaving the section unconfigured allows any OID with any value.
```
bernie-g, line 210: Should we mention what CAs support custom extensions?

docs/documentation/platform/pki/settings/profiles.mdx
```diff
@@ -62,6 +62,23 @@ Defaults can be configured for:
   If the certificate policy is updated after the profile is saved, existing defaults may no longer comply. Certificate requests will fail at validation until you update the profile defaults to match the new policy.
 </Warning>
 
+### Custom extensions
+
+Set default values for custom X.509 extensions. The [certificate policy](/documentation/platform/pki/settings/policies) decides which object identifiers (OIDs) may be emitted, so a request can carry any extension the policy permits, whether or not the profile declares it. A request that supplies a signing request can carry them in the CSR, which is how an MDM enrolling over SCEP sets a per-device value.
+
+Set a value and Infisical uses it whenever the request doesn't supply one. Leave the value empty and the request supplies it, which suits a per-principal value such as an Active Directory security identifier. Either way, Infisical checks the value against the policy's rule for that OID.
```
bernie-g, line 69: Should we mention criticality here as well

bernie-g (approved):

PR #7946, opened 2026-09-02
docs: refresh secret versioning and point-in-time recovery screenshots

jakehulberg:
## Context

Every screenshot on **Secret versioning** and **Point-in-time recovery** predated the v3 UI, so all seven were stale. This recaptures them against the current interface.

The red box and arrow annotations from the originals are reproduced in the same places, in the same colour (`#E5484A`) and weight:

| Screenshot | Annotation |
| --- | --- |
| `secret-versioning-overview.png` | Red box + left-pointing arrow on a secret row's actions, matching the old box-and-arrow on the row's **More** button |
| `secret-versioning.png` | Red box around the version entries, matching the old box around the drawer's version history section |
| `pit-commits-button.png` | Red box on the **Commits** button |
| The other four | None in the originals, none added |

Two surfaces moved in the v3 UI, which is what made a few sentences stale as well:

- **Version history** was a section inside the secret drawer. It is now its own side panel, opened from a secret's three dot menu. The per-environment secrets dashboard the old drawer lived on now redirects to the project overview.
- **Commits** were three separate pages (list, detail, restore). They are now three views inside one **Commit History** side panel.

### Prose changes

Only sentences the new screenshots directly contradict were touched, so this stays a screenshot refresh. Each one, in case you'd rather drop it:

`secret-versioning.mdx`
- The link to the secret sidebar for reaching versions, which no longer exists, now points at the three dot menu and **Version History**.
- The `<Note>` said to copy and paste a version's value to roll back, and that automatic rollback was coming soon. It shipped: each earlier version has a **Restore Version** action, so the note now describes that.

`pit-recovery.mdx`
- "4 Commits" → "6 Commits" to match the new screenshot, and the button's location is now described as the secrets table header.
- "commits page" → "commit history panel".
- "Relative timestamp" → "Timestamp"; the list now shows absolute times. Added the resource counts each row carries.
- Search and sort bullets now name what's actually there: search across messages, keys, and folders, an **Author** filter, and an oldest/newest sort.
- Folder and secret changes no longer render as JSON, so "in JSON format" is gone and `skipMultilineEncoding` is now **Multi-line encoding**, its label in the UI.
- The `+`/`-` gutter indicators are now **Added**/**Updated**/**Deleted** badges with red previous-state and green new-state panels, plus **Expand all to full state**.
- **Restore All Child Folders** → **Restore all child folders**, matching the checkbox.
- One "Clicking on any commit" → "Selecting any commit", which clears the `Infisical.UIActions` warning on that line.

## Screenshots

The seven replaced files are the diff. All were captured at 1496x841 with a 2x device pixel ratio, so they land at 2992x1682 and are a consistent set.

Before and after for `secret-versioning-overview.png`, the one carrying both a box and an arrow, so the annotation call is easy to check. The old box wrapped the **More** tooltip and the two icons under it; the new one wraps the **Secret Actions** tooltip and the three icons under it, with the arrow entering from the same side:

| Before | After |
| --- | --- |
| Old secret versioning overview screenshot | New secret versioning overview screenshot |

## Steps to verify the change

1. `cd docs && mintlify dev`, then open **Secret versioning** and **Point-in-time recovery**.
2. Check each screenshot against the current UI: version history opens from a secret's three dot menu, and commit history opens from the **Commits** button in the secrets table header.
3. `make lint-docs-branch` — 0 errors, 0 warnings. The one remaining suggestion (`it is` → `it's` at `pit-recovery.mdx:8`) is pre-existing prose in the licensing callout and untouched here.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


adkah (approved):
LGTM

PR #7951, opened 2026-09-02
feat(secrets): implement copy secrets workflow

andrewhuhh:
## Context

Secret copying was split across separate duplicate-secret and replicate-folder experiences, with different entry points and behavior depending on whether a user started from a row, the selection toolbar, or a destination folder.

This change introduces one Copy Secrets sheet for those workflows. It:

- supports copying a single secret, a multi-secret selection, or secrets chosen from a source tree;
- lets users choose source and destination environments and paths while preventing invalid same-location copies;
- previews destination changes and distinguishes new secrets from conflicts before submission;
- optionally includes secret values and lets users overwrite conflicts;
- groups folder and secret operations by destination path;
- replaces the legacy duplicate-secret modal and replicate-folder flow with the shared sheet;
- clears bulk selection after a successful copy while preserving row and toolbar behavior.

Related ticket: [ENG-5631](https://linear.app/infisical/issue/ENG-5631/implement-copy-secrets-workflow)

## Screenshots

Screenshots are outstanding because browser-based UI verification was not run for this checkpoint.

Suggested capture setup:

1. Open a Secrets Management project overview at `/organizations/{orgId}/projects/secret-management/{projectId}/overview`. Use a project with at least two environments, nested folders, several secrets, and permission to read and create secrets.
2. From a secret row's action menu, choose **Copy Secret** and capture the sheet with the source secret preselected.
3. Select multiple secrets in the overview, choose **Copy Secrets** from the selection toolbar, and capture the destination and conflict-preview state.
4. Open a destination folder, choose the toolbar's copy action, select a source environment/path and nested folder in the source tree, then capture the preview showing both new and conflicting secrets.
5. Capture the completed state after copying, confirming the destination path and the cleared bulk selection.

## Steps to verify the change

1. Run the serialized frontend validation gate:
   `/Users/andrewhuang/Documents/Agent\ Workspace/skills/validate-infisical-frontend/scripts/frontend-validation-control gate --worktree "$PWD"`
2. Run the focused utility suite:
   `cd frontend && TSX_TSCONFIG_PATH=tsconfig.app.json ./node_modules/.bin/tsx --test src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/copySecrets.utils.test.ts`
3. Open a project with two environments and confirm **Copy Secret** from a row opens the sheet with only that secret selected.
4. Select multiple secrets, open **Copy Secrets**, complete a copy, and confirm the selection clears after success.
5. Open the toolbar copy flow from a destination folder. Choose a source environment, path, and a mix of direct and nested secrets; confirm destination previews preserve relative folder structure.
6. Confirm copying to the same environment, path, and key is prevented, hidden values cannot be included, and unreadable or invalid source paths show the appropriate empty/error state.
7. Create a destination conflict. Confirm the preview distinguishes create and overwrite operations, the overwrite option controls conflicting secrets, and the success notification appears only after completion.
8. Verify sheet dismissal and keyboard focus return from the row, bulk-selection, and toolbar entry points.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


andrewhuhh:
@greptile review

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/CopySecretsSheet.tsx
scott-ray-wilson, line 1: Image

1. nit but I think we should use label instead of field label here for the switch label
2. do you think not copying secret keys when the user doesn't have permission is the right call? I feel like it might be nice to just copy the secret value if present, otherwise just the key might be nice? otherwise I'd have to do two copies potentially but curious on your thoughts
3. I feel like we shouldn't display this unless there is a conflict and maybe not there? either in a tooltip on a conflict or bottom bar? the asymmetry throws me off a little
4. I think this is disabled because I did copy secret but is there a reason to not allow free edit from it? also this kind of illuminates the difference in disabled state styling between the select and input - just pointing out

frontend/src/pages/secret-manager/OverviewPage/components/SecretTableRow/DuplicateSecretModal.tsx
scott-ray-wilson, line 1: so I don't think we can remove this one unless we add feature parity to the new copy sheet because these were specific customer requests to be able to decide/configure what properties were duplicated or copied - so we should either role back duplication or extend this but that could probably be a separate PR and we cna just scope this to the duplicate board and bulk copy?

Image

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/CopySecretsSheet.tsx
scott-ray-wilson: ah yeah for 2 I think we should do that because sometimes you don't have permission to a subset and this basically breaks copying value for them since one inability to read forces all to not read

scott-ray-wilson, line 1: first time bulk selection copy doesn't select the item 

https://github.com/user-attachments/assets/c085a3e5-18d8-4e57-93d3-62a6ad79ba68

scott-ray-wilson, line 1: honey tokens shown as non-honey token secrets in copy sheet 

https://github.com/user-attachments/assets/505e1077-2677-4030-bef4-91420d6b9380

scott-ray-wilson, line 1: 1. confusion around folder behavior when copying and visualization of it
2. would expect whenever i open copy secrets it would open the source to my current env/folder
3. copy button disabled when I expect to be able to

https://github.com/user-attachments/assets/eadfc1aa-57df-4dc8-8f45-6be2fdcfa34f

scott-ray-wilson: think we should definitely add to tooltip probably (in addition to somewhere on the sheet)

scott-ray-wilson, line 1: hitting enter in the filters submits the form to copy 

https://github.com/user-attachments/assets/5e313ca5-3176-4bc0-864c-86a083b05e3a

scott-ray-wilson, line 1: copy secrets beaks when user has describe only in one env and read value in another 

https://github.com/user-attachments/assets/41268bca-e69b-40ad-911c-b15f6ea4bec3

scott-ray-wilson (dismissed):

andrewhuhh:
needs local test to addres honey tokens issue

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/CopySecretsSheet.tsx
```diff
@@ -0,0 +1,800 @@
...
+      // Mutation failures are surfaced by the global React Query error handler.
+    } finally {
+      setIsSubmitting(false);
+    }
+  };
+
+  const handleSubmit = async () => {
+    if (disabledReason || isSubmitting) return;
+
+    if (conflictingSecrets.length > 0) {
```
scott-ray-wilson: **R1 (medium): no way to overwrite when the destination preview comes back empty.**

This is the only path to `shouldOverwrite: true`, and it is gated on the preview having found conflicts. But the new metadata endpoint returns `{ secrets: [], nextCursor: null }` rather than an error when the caller cannot describe destination secrets (`secret-metadata-service.ts:60`, plus the per-secret filter at `:116`).

So for a user with create/edit but not describe at the destination: the preview is empty, `conflictingSecrets` is empty, this branch never fires, and the copy goes out with `shouldOverwrite: false`. `secret-service.ts:3522` then hard 400s on the first existing key with *"Set shouldOverwrite to true to replace it"*, and there is no control in the sheet that can do that. The modal this replaces had an unconditional `shouldOverwrite` checkbox.

Note the `destinationQuery.isError` state at `:516` ("Destination contents unavailable") does not cover this, because the permission case is a 200 with an empty list, not an error.

This is the same permission-asymmetry shape as my earlier comment about describe-only in one env and read-value in another. Two ways out: surface overwrite as a standing option (like the old modal) whenever the destination preview returns nothing, or have the endpoint distinguish "nothing here" from "you cannot see what is here" so the sheet can warn that the preview is incomplete.

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/copySecrets.utils.ts
```diff
@@ -0,0 +1,275 @@
...
+    invocation.sourceEnvironmentSlug ?? (availableSlugs.length === 1 ? availableSlugs[0] : "");
+  const sourceFolderName =
+    invocation.origin === "toolbar" ? getCopyPathName(invocation.sourcePath) : undefined;
+  return {
+    sourceEnvironmentSlug,
+    sourcePath: invocation.sourcePath,
+    destinationEnvironmentSlug: sourceEnvironmentSlug
+      ? getOtherCopyEnvironmentSlug(environments, sourceEnvironmentSlug) || sourceEnvironmentSlug
+      : "",
+    destinationPath: sourceFolderName ? joinCopyPath("/", sourceFolderName) : "/"
```
scott-ray-wilson: **R2 (medium): the toolbar destination default flattens nested source paths.**

`joinCopyPath("/", sourceFolderName)` only round-trips for a single-segment path. Running `getInitialCopyState` for the toolbar origin:

| source | destination |
|---|---|
| `/` | `/` |
| `/app` | `/app` |
| `/foo/bar` | `/bar` |
| `/a/b/c` | `/c` |

So from `/services/api` the sheet defaults to `prod:/api` and will create `/api` at the destination root rather than mirroring the path.

The test covering this is named "opens a folder copy with that folder as the exact destination" (`copySecrets.utils.test.ts:216`) but only asserts `/app`, so it passes without exercising the nested case.

Related: the `row` and `bulk` origins default to `/` regardless of source depth, so all three entry points disagree and none mirror the source. The modal this replaces duplicated to the same `secretPath`. Worth settling on one rule and pinning it with a nested-path test.

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
```diff
@@ -2861,11 +2645,20 @@ const OverviewPageContent = () => {
...
                     }}
-                    onReplicateSecrets={() => handlePopUpOpen("replicateFolder")}
+                    onCopySecrets={() =>
+                      handleOpenCopySecrets({
+                        origin: "toolbar",
+                        sourcePath: secretPath,
+                        sourceEnvironmentSlug: singleVisibleEnv?.slug ?? ""
+                      })
+                    }
+                    isCopySecretsDisabled={hasPendingBatchChanges}
```
scott-ray-wilson, line 2671: **R3 (medium): the `canCreateFolders` gate was dropped from this entry point.**

The old "Replicate Secrets" item was `isDisabled={!isReplicateSecretsAvailable || !canCreateFolders}`. "Copy Secrets" is now gated only on `hasPendingBatchChanges`.

The sheet auto-creates every missing destination folder segment through `getOrCreateFolder` (`CopySecretsSheet.tsx:329`), so a user without folder-create permission can open the sheet, pick a nested destination, and fail partway through the copy. The catch at `CopySecretsSheet.tsx:400` is deliberately silent, so all they get is the global error toast, potentially after some folders have already been created.

Dropping the single-environment gate is clearly intentional and reads fine. The folder-create one looks like collateral.

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/CopySecretsSheet.tsx
```diff
@@ -0,0 +1,800 @@
...
+                    </h3>
+                    <div className="grid grid-cols-[minmax(8rem,0.8fr)_minmax(0,1.2fr)] gap-2">
+                      <Field>
+                        <FieldLabel
+                          htmlFor="copy-secrets-destination-environment"
+                          className="sr-only"
+                        >
+                          Destination environment
+                        </FieldLabel>
+                        <Combobox
```
scott-ray-wilson, line 657: **R4 (low/medium): parity gaps with `DuplicateSecretModal` beyond the properties list.**

`CopySecretsProperties` covers the five include options exactly, which was my main ask on that thread, thanks. Two other behaviors did not come across:

1. The modal took `environmentIds: z.array(z.string()).min(1)`, so one submit fanned out to N destination environments. This is a single-destination select.
2. The modal duplicated to the same `secretPath`; the row flow now defaults the destination to `/` (see R2).

If either was part of the customer request behind that modal, they are still open. Fine by me as a follow-up rather than a blocker, but worth confirming explicitly rather than losing silently.

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/useCopySecretsQuery.ts
```diff
@@ -0,0 +1,43 @@
...
+  enabled
+}: {
+  projectId: string;
+  environment: string;
+  secretPath: string;
+  enabled: boolean;
+}) =>
+  useQuery({
+    queryKey: [
+      ...dashboardKeys.getDashboardSecrets({ projectId, secretPath }),
```
scott-ray-wilson: **R5 (low, efficiency): the copy loop invalidates its own destination query.**

`useDuplicateSecret.onSuccess` invalidates `dashboardKeys.getDashboardSecrets({ projectId, secretPath: destinationSecretPath })` (`hooks/api/secrets/mutations.tsx:457`), which is a *prefix* of this key. Secrets go out in chunks of 50 via `chunkCopySecretIds`, so a 500-secret copy fires roughly 10 invalidations, each cancelling and restarting the full recursive destination scan while the copy is still running.

That scan is itself multi-request (paged at 500) and sits behind `secretsLimit`, and `getCopySecretsRetryDelay` waits 60s on a 429. Cheapest fix is probably to key the copy queries outside the dashboard prefix, or to disable both queries while `isSubmitting`.

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/CopySecretsSecretTree.tsx
```diff
@@ -0,0 +1,553 @@
...
+];
+
+const getAllIds = (node: FolderNode): string[] => [
+  ...node.secrets.map(({ id }) => id),
+  ...node.folders.flatMap(getAllIds)
+];
+
+const getPreviewLabel = (status: NonNullable<CopySecretsSource["previewStatus"]>) => {
+  if (status === "new") return "New";
+  if (status === "overwrite") return "Overwrite";
```
scott-ray-wilson, line 146: **R6 (low): dead branches.**

`previewStatus: "overwrite"` is never assigned in app code, only in a fixture at `copySecrets.utils.test.ts:46`, so this branch is unreachable. `destinationContents` only ever sets `"conflict"` and `"new"`.

Same story for `getRestrictionLabel` at `:60-65`: the `isRotated` and `isHoneyToken` branches cannot fire, because `sourceSecrets` pre-filters both through `isCopySecretSelectable` (`CopySecretsSheet.tsx:154`), and on the destination side the badge renders under `restrictionLabel && !isReadOnly`. The `"This managed secret cannot be copied"` disabled state at `:252` is unreachable for the same reason.

Either delete them, or drop the pre-filter and render those secrets disabled with the badge, which is arguably the friendlier answer to my earlier honey-token comment than hiding them outright.

frontend/src/pages/secret-manager/OverviewPage/components/CopySecretsSheet/CopySecretsSheet.tsx
```diff
@@ -0,0 +1,800 @@
...
+        isValueHidden: secret.isValueHidden,
+        path,
+        previewStatus: "new" as const
+      };
+      contents.push(incoming);
+      byLocation.set(key, incoming);
+    });
+
+    return contents;
+  }, [destinationQuery.data, requestGroups, selectedSecrets]);
```
scott-ray-wilson: **R7 (nit): this `useMemo` never memoizes.** `requestGroups` and `selectedSecrets` are both rebuilt unconditionally on every render, so the dependency array always holds fresh identities and the body re-runs regardless.

scott-ray-wilson (dismissed):
Second pass over `68ac05c7..e237ca59`. All eight of my earlier threads are addressed, verified against the code, so I am leaving those resolved and scoping this review to what is new.

Heads up that the surface grew a lot since my last pass: a whole new backend endpoint (`GET /api/v1/dashboard/accessible-secrets/metadata`, a cursor-paginated recursive metadata scan), plus `copySecrets.data.ts`, `useCopySecretsQuery.ts` and `CopySecretsProperties.tsx`. None of that existed at `68ac05c7`.

That backend addition reads well against `CODE_QUALITY.md`: full zod validation on the querystring, opaque keystore-backed cursors scoped by an actor hash with a TTL, per-secret CASL subject checks (no inference from the root), replica reads, no encrypted columns touched, no transaction held across the scan, and readable errors on expired cursors and unsupported project versions.

Seven findings inline. **R1** and **R2** are the two I would want fixed before merge; the rest are smaller.

### Verification run

All at `e237ca59` in a clean worktree:

- `tsx --test copySecrets.utils.test.ts copySecrets.data.test.ts`: 30/30 pass
- `vitest run -c vitest.unit.config.mts secret-metadata-service.test.ts dashboard-secret-metadata-schemas.test.ts`: 37/37 pass
- `tsc --noEmit -p tsconfig.app.json` (frontend): clean
- `tsc --noEmit` (backend): clean
- `eslint --max-warnings=0` on every changed frontend and backend file: clean
- No dangling references to `DuplicateSecretModal`, `ReplicateFolderFromBoard`, `SecretTreeView` or `replicateSecrets`

### Still outstanding

The PR body says browser verification was not run and screenshots are outstanding. Several of my original comments were visual judgments, folder behavior and its visualization in particular, which I cannot close from code alone; R2 suggests that flow still needs a real look.

scott-ray-wilson (approved):
LGTM

andrewhuhh:
lets goo

PR #7957, opened 2026-09-03
chore(deps): base image and dependency updates from a customer contai…

Vligai:
## Context

**1. Base image packages.** Thirteen packages already installed in the runtime image are pinned forward to their `trixie-security` versions: the util-linux family (nine, one source package so apt moves them together), `libsystemd0`, `libudev1`, `liblzma5` and `sed`. They join the existing `--only-upgrade` block, so nothing new is installed. `apt-get -s upgrade` reports `16 upgraded, 0 newly installed, 0 to remove`. 

The pins go stale the same way `libssl3t64` did, which is the failure mode the earlier runtime-image PR wrote up. Dropping the `=version` suffixes so apt takes the candidate would avoid it, as `libcap2` in the same list already does. Out of scope here, but the list is now long enough that it is worth deciding.

**2. Bundled CLI.** The image pinned `infisical=0.43.90`, set on 9 June. Moved to `0.43.129`. Secret scanning is the only runtime consumer (`secret-scanning-v2-fns.ts` and the v1 equivalent, four call sites, all `infisical scan`). Flag and exit-code parity was checked against the shipped 0.43.90 with identical inputs: both accept `--exit-code=77`, `-r`, `-c`, `--source`, `--no-git`, both exit 77 on a planted secret and 0 on a clean tree.

**3. Application dependencies.** Five overrides, all for packages that ship in the image. 

`protobufjs` is the exception and is deliberate. `@opentelemetry/otlp-transformer` pins it to exactly `8.0.1`, which carries eleven advisories; the override forces `7.6.6`, which carries none. Removing the override lets npm honour the pin and install both copies, which is worse. The same reasoning applies to the pre-existing `postcss` override, which is why the dev-only overrides were dropped rather than kept: they matched zero rows in the report and are absent from the shipped image.

`axios` uses a tilde rather than a caret, the only one in the file. Ten advisories affect the previous `1.16.1` and all ten are fixed in `1.18.0`, but `^1.18.1` resolves to `1.20.0` and pulls in three further releases of behaviour changes that are not needed to clear them.

## Steps to verify the change

```bash
# apt pins resolve and nothing new is installed
docker run --rm --platform linux/amd64 --user root --entrypoint sh \
  infisical/infisical:v0.165.1 -c 'apt-get update -qq && apt-get -s upgrade | grep "^[0-9]* upgraded"'

# backend compiles against the new dependency set
docker build --platform linux/arm64 --target backend-build \
  -t infisical-backend-build:test -f Dockerfile.standalone-infisical .

cd backend-go && go build ./... && go test ./internal/... -count=1
```

Confirm the overrides actually resolved, since an override the lockfile ignores is a silent no-op:

```bash
cd backend && node -e '
const l=require("./package-lock.json").packages;
for (const n of ["dompurify","protobufjs","js-yaml","@grpc/grpc-js","fast-xml-builder","axios"])
  for (const [k,v] of Object.entries(l))
    if (k.endsWith("node_modules/"+n)) console.log(v.version.padEnd(9), k);'
```

Expect `dompurify 3.4.14`, `protobufjs 7.6.6`, `js-yaml 4.3.2`, `@grpc/grpc-js 1.14.4`, `fast-xml-builder 1.3.1`, `axios 1.18.1`, one copy each.

### Verified locally

- Thirteen apt pins all land at their pinned versions, CLI reports `0.43.128`
- grype on the thirteen target packages drops from 105 findings to 32
- `backend-build` stage compiles, `dist/main.mjs` present, all six target versions correct
- `backend-go` builds, 17 of 17 unit packages pass
- Published-image diff against v0.162.14: 283 packages to 180, 103 removed, none added

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [x] Chore

## Checklist

- [x] Tested locally


Dockerfile.standalone-infisical
victorvhs017, line 1: We should update Dockerfile.fips.standalone-infisical, too.

victorvhs017: On this note, the snyk check is only happening in the standalone image and never checks the FIPS one, should we do both?

Image

Dockerfile.standalone-infisical
```diff
@@ -213,7 +213,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libaio1t64 \
     smbclient \
     && curl -1sLf 'https://artifacts-cli.infisical.com/setup.deb.sh' | bash \
-    && apt-get update && apt-get install -y infisical=0.43.90 \
+    && apt-get update && apt-get install -y infisical=0.43.128 \
```
victorvhs017: backend/Dockerfile, backend/Dockerfile.dev, backend/Dockerfile.fips-toolchain, all of them have the CLI pinned at `infisical=0.43.79`, which would drift from this pin.

We can update them manually here, but is there a way to keep them always in sync?

victorvhs017 (changes requested):
Something is up with the integration tests, too. We should check that.

docker-compose.bdd.yml
```diff
@@ -41,6 +41,8 @@ services:
     build:
       context: ./backend
       dockerfile: Dockerfile.dev
+      args:
+        INFISICAL_CLI_VERSION: ${INFISICAL_CLI_VERSION}
```
victorvhs017: Compose passes an explicit empty build arg when INFISICAL_CLI_VERSION is unset (${VAR} → ""), which overrides the Dockerfile ARG default and still hits the test -n guard, but minutes into the build.

We can fail faster and hint the user on how to properly call these with:

`INFISICAL_CLI_VERSION: "${INFISICAL_CLI_VERSION:?not set - source build-versions.env before running compose}"`

docker-compose.dev-read-replica.yml
```diff
@@ -84,6 +84,8 @@ services:
     build:
       context: ./backend
       dockerfile: Dockerfile.dev
+      args:
+        INFISICAL_CLI_VERSION: ${INFISICAL_CLI_VERSION}
```
victorvhs017: Same here

docker-compose.dev-read-replica.yml
```diff
@@ -96,6 +98,8 @@ services:
...
```
victorvhs017: same here

Makefile
```diff
@@ -1,3 +1,9 @@
+# Versions pinned across multiple Dockerfiles live in one file and are exported
+# so compose can pass them as build args. Dockerfiles declare the ARG without a
+# default, so a build that misses this fails rather than installing an unpinned version.
```
victorvhs017: This comment is stale, the Dockerfiles do have a default now.

Makefile
```diff
@@ -1,3 +1,9 @@
...
+include build-versions.env
+export
```
victorvhs017: In GNU Make, export with no arguments exports every Make variable to the environment of every recipe, not just INFISICAL_CLI_VERSION.

That includes COMPOSE_PROJECT_NAME on line 27.

This means that env would leak, and even though it's not something that could cause issues, other envs could be added in the future.

Another issue is that breaks make up-dev in checkout dirs with uppercase characters: Compose normalizes the project name for volume/network prefixes, but Make passes the raw value.

We can replace it with `export INFISICAL_CLI_VERSION` only.

victorvhs017 (changes requested):

victorvhs017 (approved):

PR #7961, opened 2026-09-03
feat(pki): manage imported certificates by attaching a certificate profile

carlosmonastyrski:
## Context

Imported certificates had no certificate authority attached, so renewal, reissue and revocation all failed on them. You can now pick a certificate profile while importing into an Application.

The profile determines which authority signs the replacement when the certificate is renewed, and that is all Infisical needs to run the rest of the lifecycle.

Internal and self-signed profiles need nothing else. A DigiCert profile also asks for the CertCentral order ID, which you pick from a live list of that CA's orders instead of typing a number, and Infisical confirms the order really belongs to the certificate before linking it.

Revoking is refused when the profile's authority did not issue the certificate, because revocation only takes effect through the issuing authority.

## Type

- [ ] Fix
- [x] Feature
- [x] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/certificate/certificate-service.ts
```diff
@@ -474,6 +503,32 @@ export const certificateServiceFactory = ({
    * Note: Revoking a certificate adds it to the certificate revocation list (CRL)
    * of its issuing CA
    */
+  const $assertInternalCaIssuedCertificate = async (caId: string, certificatePem: string) => {
+    const caCerts = await getCaCertChains({
```
sheensantoscapadngan: Looks like the comment above this does not match?

backend/src/services/certificate/certificate-service.ts
```diff
@@ -905,6 +1125,14 @@ export const certificateServiceFactory = ({
 
     const { serialNumber, notBefore, notAfter } = leafCert;
 
+    if (linkage?.caType === CaType.DIGICERT && linkage.externalMetadata?.type === CaType.DIGICERT) {
+      await digicertFns.assertOrderMatchesCertificate({
+        caId: linkage.caId,
+        orderId: linkage.externalMetadata.orderId,
+        serialNumber
+      });
+    }
```
sheensantoscapadngan: Couldn't we do this inside the resolve import linkage call? Make it so that this entire certificate service method is as generic as possible.

sheensantoscapadngan: It might be cleaner to do this inside `CertificateImportLinkageMap`, where you can return some sort of callback function, perhaps?

Image

docs/documentation/platform/pki/applications/certificates.mdx
```diff
@@ -56,10 +56,42 @@ For automated issuance, configure an [enrollment method](/documentation/platform
...
-Click **Import** on the Certificates tab to bring in a certificate you already hold, either as PEM files or as a `.p12` / `.pfx` keystore with its password.
+Select **Import** on the Certificates tab to bring in a certificate you already hold, either as PEM files or as a `.p12` / `.pfx` keystore with its password.
 
 A keystore can hold several certificates. Infisical lists what is inside so you can pick which ones to import.
 
+### Managing an imported certificate
+
+An imported certificate is tracked for visibility and expiry alerts. Renewal, reissue and revocation stay wherever it was issued.
+
+To hand the rest of its lifecycle to Infisical, pick a **Certificate Profile** during import. The profile determines which authority signs the replacement when the certificate is renewed. Nothing is reissued at import time and your key pair is untouched.
```
sheensantoscapadngan: I think we can just remove the last sentence here since it's not really implied anyway?

docs/documentation/platform/pki/applications/certificates.mdx
```diff
@@ -56,10 +56,42 @@ For automated issuance, configure an [enrollment method](/documentation/platform
...
+  You can pick a profile when you import PEM files into an Application. Importing a keystore, or importing from the organization-wide Inventory, tracks the certificate without one.
+</Note>
+
+### Which profile to attach
+
+Attaching a profile backed by the authority that issued the certificate keeps the whole lifecycle in Infisical.
+
+You can also attach one backed by a different authority, including a self-signed profile. The certificate you already hold stays in service, and its replacement is issued by the authority you're moving to. Renewal works in both cases, because a renewal is a new certificate signed by the profile's authority.
+
+Revocation behaves differently. It only takes effect through the issuing authority, because that authority publishes the revocation list clients check. Infisical therefore refuses to revoke an imported certificate whose profile is backed by a different authority, rather than record a revocation no client would see. Revoke those certificates with their original issuer instead.
```
sheensantoscapadngan: We can also remove the "Rather than record a revocation, no client could see" part

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateImportProfileFields.tsx
```diff
@@ -0,0 +1,191 @@
+import { useEffect } from "react";
+import { Control, Controller, UseFormSetValue, useWatch } from "react-hook-form";
+import { OptionProps } from "react-select";
+
```
sheensantoscapadngan: IGNORE PLACEMENT

When selecting a certificate profile to import a certificate with, it looks like we don't validate the values against the actual policy...
Image

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateImportProfileFields.tsx
```diff
@@ -0,0 +1,191 @@
...
+  TooltipTrigger
+} from "@app/components/v3";
+import { useDigiCertConnectionListOrders } from "@app/hooks/api/appConnections/digicert";
+import { useListCasByProjectId } from "@app/hooks/api/ca";
+import { CaStatus, CaType } from "@app/hooks/api/ca/enums";
+import { useListCertificateProfiles } from "@app/hooks/api/certificateProfiles/queries";
+
+import { getCertificateImportReference, isCaTypeLinkable } from "./certificate-import-linkage";
+import { CertificateImportFormData } from "./types";
+
```
sheensantoscapadngan: IGNORE PLACEMENT

Looks like we don't handle certificate profile selection for PKCS12 import
Image

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateImportProfileFields.tsx
```diff
@@ -0,0 +1,191 @@
...
+  id: string;
+  slug: string;
+  caId: string | null;
+  caType: CaType | null;
+  isLinkable: boolean;
+};
+
+type ReferenceOption = { value: string; label: string };
+
+type Props = {
```
sheensantoscapadngan: IGNORE PLACEMENT

Are we intentionally not handling certificate profile assignment at the inventory level? I saw that in our backend implementation, this is something that's technically possible. 
Image

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateImportProfileFields.tsx
carlosmonastyrski: It's there, that flow expects you to first read the keystore first, and then before confirming select the profile
Before:
CleanShot 2026-09-04 at 21 13 58@2x
After
CleanShot 2026-09-04 at 21 14 01@2x

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateImportProfileFields.tsx
carlosmonastyrski: Issuance and renewal flows are not allowed at the inventory level, attaching it to a profile there would be a bad idea, if you want to keep the cert lifecycle and renew/revoke or any other operation, then that's the application responsability

frontend/src/pages/cert-manager/CertificatesPage/components/CertificateImportProfileFields.tsx
carlosmonastyrski: Added the restriction, the initial thought there is that it could be treated as allow imports but then only issuance/renewals are blocked (the real calls to the CA and by consequence the policy is enforced), something similar to issue something with an allow-all policy and then restrict it before renewal, but it can check it during import as well 👍

backend/src/services/certificate/certificate-service.ts
```diff
@@ -1125,12 +1146,59 @@ export const certificateServiceFactory = ({
...
+      const { keyAlgorithm, signatureAlgorithm } = certificateAlgorithms;
+
+      const validation = await certificatePolicyService.validateCertificateRequest(linkage.policyId, {
+        commonName: commonName || undefined,
+        organization: certificateFields.subjectOrganization ?? undefined,
+        organizationalUnit: certificateFields.subjectOrganizationalUnit ?? undefined,
+        country: certificateFields.subjectCountry ?? undefined,
+        state: certificateFields.subjectState ?? undefined,
+        locality: certificateFields.subjectLocality ?? undefined,
+        domainComponents: certificateFields.subjectDomainComponents?.split(",") ?? undefined,
```
sheensantoscapadngan: Wouldn't it be cleaner for us to do this inside the verify certificate function above? Basically make it so that anything that's linkage related is all in one place... that way if anything else imports a certificate in the future, we don't have to think about adding this validate check there...

backend/src/services/certificate/certificate-service.ts
```diff
@@ -1125,12 +1146,59 @@ export const certificateServiceFactory = ({
...
+          ? signatureAlgorithm
+          : undefined,
+        ...(certificateFields.isCA && {
+          basicConstraints: { isCA: true, pathLength: certificateFields.pathLength ?? undefined }
+        })
       });
+
+      if (!validation.isValid) {
+        throw new BadRequestError({
+          message: `This certificate does not satisfy the policy of certificate profile '${linkage.profileName}', so that profile could not renew it: ${validation.errors.join(", ")}. Attach a profile whose policy allows this certificate, or import it without one to track it for visibility and expiry alerts only.`
```
sheensantoscapadngan: I think we can just omit the part about the profile not being able to renew it policy...

backend/src/lib/api-docs/constants.ts
```diff
@@ -2232,6 +2232,8 @@ export const CERTIFICATES = {
     id: "The ID of the certificate to get.",
     serialNumber: "The serial number of the certificate to get.",
     hasPrivateKey: "Whether Infisical holds the private key for this certificate.",
+    externalMetadata:
+      "Identifies this certificate at the provider that issued it, for certificates issued by or linked to an external certificate authority. Null for everything else.",
     latestRenewalCertificateId:
```
sheensantoscapadngan, line 2241: IGNORE PLACEMENT

can't read the entire text...
Image

backend/src/lib/api-docs/constants.ts
```diff
@@ -2271,6 +2273,10 @@ export const CERTIFICATES = {
     chainPem: "Optional PEM-encoded chain of intermediate certificates.",
     friendlyName: "A friendly name for the certificate.",
     pkiCollectionId: "The ID of the PKI collection to add the certificate to.",
+    profileId:
+      "The certificate profile that will manage this certificate's lifecycle. The certificate must satisfy the profile's certificate policy, or the request is rejected with a 400 naming the attributes that failed. Omit to track the certificate without renewal, reissue or revocation.",
+    externalMetadata:
```
sheensantoscapadngan, line 2282: IGNORE PLACEMENT

maybe we can do one violation per line instead? seems like a hard read right now 
Image

sheensantoscapadngan: same with this one 
Image

carlosmonastyrski: CleanShot 2026-09-08 at 23 35 22@2x
CleanShot 2026-09-08 at 23 34 54@2x

backend/src/lib/api-docs/constants.ts
carlosmonastyrski: CleanShot 2026-09-08 at 23 36 35@2x

backend/src/services/certificate-common/external-metadata-schemas.ts
```diff
@@ -34,3 +34,44 @@ export const ExternalMetadataSchema = z.discriminatedUnion("type", [
 ]);
 
 export type TExternalMetadata = z.infer<typeof ExternalMetadataSchema>;
+
+export const ImportExternalMetadataSchema = z.discriminatedUnion("type", [DigiCertExternalMetadataSchema]);
+
```
sheensantoscapadngan, line 39: IGNORE PLACEMENT

User should be able to go back and select a different profile, right? Right now, I would have to go through the entire process of dropping the file again and reading the keystore...
Image

backend/src/services/certificate-common/external-metadata-schemas.ts
```diff
@@ -34,3 +34,44 @@ export const ExternalMetadataSchema = z.discriminatedUnion("type", [
...
+type TCertificateImportLinkage = {
+  externalMetadataSchema: z.ZodTypeAny | null;
+  referenceLabel?: string;
+  verifyCertificate?: (
+    args: { caId: string; externalMetadata: TImportExternalMetadata; serialNumber: string },
+    deps: TCertificateImportVerifierDeps
+  ) => Promise<void>;
+};
+
```
sheensantoscapadngan, line 56: IGNORE PLACEMENT

If a user is just trying to import one certificate, I wonder if we can just redirect directly to the details page, as opposed to showing this modal over here. It feels redundant because, before importing, I already saw these details anyway, since I copy-pasted the chain, key, etc. 

Image

sheensantoscapadngan (approved):

PR #7966, opened 2026-09-03
fix: allow read operations in cross project secret sharing

mathnogueira:
## Context

If our license server is down, this could lead to downtime to our users that are using cross project secret sharing. Now we only check the license when creating resources, but reading and resolving them don't require the plan.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/project-folder-grant/project-folder-grant-fns.ts
```diff
@@ -1,14 +0,0 @@
-import { TFeatureSet } from "@app/ee/services/license/license-types";
-import { requestMemoKeys } from "@app/lib/request-context/memo-keys";
-import { requestMemoize } from "@app/lib/request-context/request-memoizer";
-
-import { TOrgDALFactory } from "../org/org-dal";
-
-export const isCrossProjectEnabled = async (
```
mathnogueira, line 7: Now that we are not verifying the plan on all paths, I don't think this function makes too much sense. Better to inline those calls. One less indirection to deal with in the code.

varonix0 (approved):

PR #7979, opened 2026-09-04
fix(identity-ua): count client secret uses atomically

Vligai:
## Context

`clientSecretNumUsesLimit` caps how many times a Universal Auth client secret can be exchanged for an access token. The count and the cap were settled in two separate steps: the login handler compared a `clientSecretNumUses` value read off a row fetched earlier, and only later asked the database to increment it. Logins that overlap in that window all read the same pre-increment count, all conclude they are under the limit, and all mint a token, so a secret with `numUsesLimit: 1` can serve several logins and the stored count can end up above the limit it was given.

Both now happen in one statement:

```sql
update "identity_ua_client_secret"
set "clientSecretLastUsedAt" = $1, "clientSecretNumUses" = "clientSecretNumUses" + 1
where "id" = $2 and "clientSecretNumUsesLimit" > 0
  and "clientSecretNumUses" < "identity_ua_client_secret"."clientSecretNumUsesLimit"
returning "id"
```

### Compatibility

No migration, no schema change, no change to the login endpoint's request or response, no change to permissions. The rejection reuses the existing message and reason code.

One behavior change for the release notes: overlapping logins that previously slipped past the cap are now refused. A workload sharing one finite-use secret across replicas with the limit set below its replica count was getting away with it, and will now see the excess replicas fail to start. The fix is to raise the limit to the replica count, or use an unlimited secret with a short token TTL.

## Steps to verify the change

Create a single-use client secret and try to redeem it several times at once.

In the UI, under Access Control > Identities, create a machine identity, add Universal Auth, then create a client secret with **Max Number of Uses** set to `1`. Note the Client ID and the secret.

Redeem it twenty times in parallel:

```bash
CLIENT_ID=<client id>
CLIENT_SECRET=<client secret>

seq 20 | xargs -P 20 -I{} curl -s -o /dev/null -w '%{http_code}\n' \
  -X POST http://localhost:8080/api/v1/auth/universal-auth/login \
  -H 'Content-Type: application/json' \
  -d "{\"clientId\":\"$CLIENT_ID\",\"clientSecret\":\"$CLIENT_SECRET\"}" \
  | sort | uniq -c
```

Expected on this branch:

```
   1 200
  19 401
```

The e2e spec covers the same three cases if you would rather run it than click through: `e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts`

### Verified locally

`identity-ua-client-secret-usage-limit.spec.ts`, 4 passed:

- 20 concurrent logins yield exactly 1 token at `numUsesLimit: 1` and exactly 5 at `numUsesLimit: 5`, with distinct tokens, every rejection the usage-limit 401, and the stored count on the limit rather than above it
- An unlimited secret still serves all 20 and stays unrevoked
- A sequential second login on a single-use secret is still refused

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
+++ b/backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
@@ -141,6 +141,9 @@ describe("Universal Auth client secret usage limit", async () => {
           body: { clientId, clientSecret }
         });
         expect(afterBurst.statusCode).toBe(401);
+        // and it says why. Revoking mid-burst would make this attempt miss the secret entirely and
+        // answer "Invalid credentials", which is both untrue and a failed-credential strike
+        expect(afterBurst.json().message).toBe(USAGE_LIMIT_REACHED_MESSAGE);
         expect((await readClientSecretRow(clientSecretId))?.isClientSecretRevoked).toBe(true);
       } finally {
         await deleteIdentity(identityId);
--- a/backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
+++ b/backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
@@ -44,7 +44,9 @@ const createUaIdentity = async (name: string, numUsesLimit: number) => {
     body: {
       accessTokenTTL: 2592000,
       accessTokenMaxTTL: 2592000,
-      accessTokenNumUsesLimit: 0
+      accessTokenNumUsesLimit: 0,
+      // a burst of rejections must not be able to trip the lockout and change what these assert
+      lockoutEnabled: false
     }
   });
   expect(attachRes.statusCode).toBe(200);
@@ -127,7 +129,19 @@ describe("Universal Auth client secret usage limit", async () => {
         // exactly the limit: no increments lost by the winners, none burned by the losers
         const row = await readClientSecretRow(clientSecretId);
         expect(row?.clientSecretNumUses).toBe(numUsesLimit);
-        expect(row?.isClientSecretRevoked).toBe(true);
+
+        // a spent secret is revoked by the next attempt rather than mid-burst. Revoking during the
+        // burst would hide it from the secret lookup and send the rest of the burst down the
+        // invalid-credential path, which feeds the lockout counter for the whole clientId
+        expect(row?.isClientSecretRevoked).toBe(false);
+
+        const afterBurst = await testServer.inject({
+          method: "POST",
+          url: "/api/v1/auth/universal-auth/login",
+          body: { clientId, clientSecret }
+        });
+        expect(afterBurst.statusCode).toBe(401);
+        expect((await readClientSecretRow(clientSecretId))?.isClientSecretRevoked).toBe(true);
       } finally {
         await deleteIdentity(identityId);
       }
--- a/backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
+++ b/backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
@@ -1,0 +1,178 @@
+/**
+ * E2E tests for the Universal Auth client-secret usage limit under concurrency.
+ *
+ * Prerequisites (handled by vitest-environment-knex.ts):
+ *   - testServer: running Fastify instance
+ *   - jwtAuthToken: pre-authenticated admin JWT (user session)
+ *   - testDb: knex instance
+ *
+ * numUsesLimit has to hold when logins arrive together, not just in sequence, so
+ * these cover the concurrent case a sequential test cannot reach.
+ *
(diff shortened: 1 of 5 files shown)


backend/src/lib/telemetry/telemetry-attributes.ts
```diff
@@ -9,6 +9,9 @@ export const INFISICAL_CORE_METER_ATTRIBUTES = [
   // Bounded enums
   "infisical.auth.method",
   "infisical.auth.result",
+  // why an attempt failed, so a policy rejection is distinguishable from a bad credential.
+  // Bounded by AUTH_ATTEMPT_REASONS in metrics.ts, which folds anything unlisted into "other"
+  "infisical.auth.reason",
```
lobocv: should it be auth.failure_reason? Right now it implies it's the reason for authenticating

backend/src/services/identity-ua/identity-ua-service.ts
```diff
@@ -318,33 +318,63 @@ export const identityUaServiceFactory = ({
...
+            : // unlimited secret: numUses is informational, so collapse a login storm to one row write per window
+              keyStore
+                .setItemWithExpiryNX(
+                  KeyStorePrefixes.IdentityUaClientSecretUsageDebounce(clientSecretId),
+                  UA_CLIENT_SECRET_USAGE_DEBOUNCE_SECONDS,
+                  "1"
+                )
+                .then(Boolean),
+          shouldRecordIdentityLastLogin(keyStore, identity.id)
+        ]);
```
lobocv, line 360: nit: Maybe i'm not used to typescript yet but I had to stop and really think what this was doing. I think it could be simplified by not trying to force everything into a single func call. My suggestion to improve readability would be to do another IF statement on the hasFiniteUsageLimit and then append a list of promises, then call Promise.all() on that.

backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+  const identity = createRes.json().identity as { id: string };
+
+  const attachRes = await testServer.inject({
+    method: "POST",
+    url: `/api/v1/auth/universal-auth/identities/${identity.id}`,
+    headers: { authorization: `Bearer ${jwtAuthToken}` },
+    body: {
+      accessTokenTTL: 2592000,
+      accessTokenMaxTTL: 2592000,
+      accessTokenNumUsesLimit: 0,
```
lobocv, line 48: why is the limit 0 here? or is this a different usage limit? seems like we set the usage limit in a separate API call below?

backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+const deleteIdentity = async (identityId: string) => {
+  await testServer.inject({
+    method: "DELETE",
+    url: `/api/v1/identities/${identityId}`,
+    headers: { authorization: `Bearer ${jwtAuthToken}` }
+  });
+};
+
+const loginConcurrently = (clientId: string, clientSecret: string) =>
```
lobocv, line 79: nit: might be more readable to pass in the concurrency value as an arg here and define the CONCURRENCY locally within the test. It prevents less global state

backend/e2e-test/routes/v1/identity-ua-client-secret-usage-limit.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+
+const readClientSecretRow = async (clientSecretId: string) => {
+  const row = await testDb(TableName.IdentityUaClientSecret).where({ id: clientSecretId }).first<{
+    clientSecretNumUses: string | number;
+    isClientSecretRevoked: boolean;
+  }>("clientSecretNumUses", "isClientSecretRevoked");
+
+  // pg hands back int8 as a string, which is why the row schema coerces it
+  return row && { ...row, clientSecretNumUses: Number(row.clientSecretNumUses) };
+};
```
lobocv, line 98: Yeah having to look into the DB from an e2e test makes me uncomfortable. I think this test would better be served as an integration test but it's okay to leave it for now. I don't think we have good integration test examples. IMO E2E should test only observable behaviours, and looking into the columns to determine state violates that.
This is something I'd like to define more clearly in the testing strategy work

lobocv (dismissed):
Logic looks good. I left some minor comments on readability but they're not a blocker. Feel free to close them if you disagree! Interesting bug!

backend/src/lib/telemetry/telemetry-attributes.ts
Vligai: Good catch, renaming to infisical.auth.failure_reason.

PR #7992, opened 2026-09-04
feat(gateways): make relay optional

bernie-g:
## Context

Makes the relay optional for Gateways: instead of the Gateway dialing out to a relay, it can accept inbound mTLS on a registered address that Infisical dials, and health is now tracked per transport so a Gateway with both stays up while either path works. Direct listen is self-hosted only, since Infisical Cloud has no route into a customer network.

## Type

- [x] Fix
- [x] Feature
- [ ] Improvement
- [x] Breaking
- [ ] Docs
- [ ] Chore

Breaking applies to the Gateway Helm chart only (bumped to 1.5.0, see its CHANGELOG): the Service is now created only in direct listen mode, and `service.port` defaults to the port from `gateway.listenAddress` rather than `80`. `gateway.listenAddress` also needs an `image.tag` whose CLI supports `--listen-address`, so **the chart's tag has to be bumped once the CLI release ships**.

## Checklist

- [x] Tested locally

Paired CLI change: Infisical/cli#385


backend/src/lib/gateway-v2/gateway-v2.ts
```diff
@@ -236,30 +254,75 @@ export const setupRelayServer = async ({
...
+
+    const dialDirect = async () => {
+      const parsed = new URL(`tcp://${directAddress}`);
+      const serverName = parsed.hostname.startsWith("[") ? parsed.hostname.slice(1, -1) : parsed.hostname;
+      const [targetHost] = await verifyHostInputValidity({
+        host: serverName,
+        isGateway: true,
+        isDynamicSecret: false
+      });
+      return { conn: net.connect({ host: targetHost, port: Number(parsed.port) }), serverName, direct: true };
```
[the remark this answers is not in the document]
bernie-g: Fixed

helm-charts/infisical-gateway/templates/_helpers.tpl
```diff
@@ -60,3 +60,16 @@ Create the name of the service account to use
...
+Port the gateway listens on for direct connections, taken from gateway.listenAddress so the
+container port and the Service can never disagree with what the gateway actually binds.
+*/}}
+{{- define "infisical-gateway.listenPort" -}}
+{{- $addr := .Values.gateway.listenAddress -}}
+{{- $port := splitList ":" $addr | last -}}
+{{- if or (eq $port $addr) (not (regexMatch "^[0-9]+$" $port)) -}}
+{{- fail (printf "gateway.listenAddress must be host:port, got %q" $addr) -}}
+{{- end -}}
+{{- $port -}}
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/routes/v3/gateway-router.ts
```diff
@@ -717,33 +720,60 @@ export const registerGatewayV3Router = async (server: FastifyZodProvider) => {
...
+      if (connected.directAddress) transports.push("direct");
+      if (connected.relayHost) transports.push("relay");
+
+      await server.services.auditLog.createAuditLog({
+        orgId: req.permission.orgId,
+        actor: { type: ActorType.GATEWAY, metadata: { gatewayId: connected.gatewayId } },
+        event: {
+          type: EventType.GATEWAY_ENROLL,
+          metadata: {
+            gatewayId: connected.gatewayId,
```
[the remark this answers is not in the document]
bernie-g: Fixed

frontend/src/pages/organization/NetworkingPage/GatewayDetailsByIDPage/components/GatewayAuthMethod/EnrollmentTokenDialog.tsx
```diff
@@ -62,25 +66,31 @@ export const EnrollmentTokenContent = ({
...
   const expiryLabel = formatTimeRemaining(expiresAt, now);
   const isExpired = expiryLabel === "Expired";
 
   const cliCommand = useMemo(() => {
-    const relayPart = resolvedRelayName ? ` --target-relay-name=${resolvedRelayName}` : "";
-    return `infisical gateway start ${gatewayName} --enroll-method=token --token=${enrollmentToken}${relayPart} --domain=${siteURL}`;
-  }, [gatewayName, enrollmentToken, resolvedRelayName, siteURL]);
+    const relayPart = resolvedRelayName ? ` --relay=${resolvedRelayName}` : "";
+    const directPart = isDirect
+      ? ` --listen-address=${listenAddress.trim() || "<gateway-address>:8443"}`
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/routes/v3/gateway-router.ts
```diff
@@ -717,33 +720,60 @@ export const registerGatewayV3Router = async (server: FastifyZodProvider) => {
     schema: {
       operationId: "connectGateway",
       body: z.object({
-        relayName: z.string().trim().min(1).max(32).optional()
+        relayName: z.string().trim().min(1).max(32).optional(),
+        directAddress: z.string().trim().min(3).max(255).optional()
```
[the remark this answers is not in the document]
bernie-g: Fixed

helm-charts/infisical-gateway/templates/deployment.yaml
```diff
@@ -122,12 +122,21 @@ spec:
             {{- end }}
             {{- end }}
             {{- if .Values.gateway.relayName }}
-            - --target-relay-name={{ .Values.gateway.relayName }}
+            - --relay={{ .Values.gateway.relayName }}
+            {{- end }}
+            {{- if .Values.gateway.listenAddress }}
+            - --listen-address={{ .Values.gateway.listenAddress }}
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -54,6 +56,35 @@ import { TOrgGatewayConfigV2DALFactory } from "./org-gateway-config-v2-dal";
...
+  if (
+    !parsed.hostname ||
+    !parsed.port ||
+    parsed.username ||
+    parsed.password ||
+    parsed.pathname ||
+    parsed.search ||
+    parsed.hash
+  ) {
+    throw new BadRequestError({ message: `Gateway direct address "${address}" must use the host:port format` });
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/lib/gateway-v2/gateway-v2.ts
```diff
@@ -236,30 +248,72 @@ export const setupRelayServer = async ({
...
   const openUpstream = async (): Promise<TUpstream> => {
     const tunnelId = crypto.randomBytes(4).toString("hex");
+    const hasRelayFallback = Boolean(directAddress && relayHost && relay);
+
+    // Validated at registration; re-checking here would resolve DNS on every dial.
+    const dialDirect = () => {
+      const parsed = new URL(`tcp://${directAddress}`);
+      const serverName = parsed.hostname.startsWith("[") ? parsed.hostname.slice(1, -1) : parsed.hostname;
+      return Promise.resolve({
+        conn: net.connect({ host: serverName, port: Number(parsed.port) }),
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/services/dynamic-secret/dynamic-secret-fns.ts
```diff
@@ -6,31 +6,19 @@ import { BadRequestError } from "@app/lib/errors";
...
-  const appCfg = getConfig();
-
-  if (appCfg.isDevelopmentMode || appCfg.isTestMode) return [host];
-
-  if (isGateway) return [host];
+let reservedIpsCache: { ips: string[]; expiresAt: number } | null = null;
+const RESERVED_IPS_TTL_MS = 5 * 60 * 1000;
 
+const getReservedIps = async () => {
+  if (reservedIpsCache && reservedIpsCache.expiresAt > Date.now()) return reservedIpsCache.ips;
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
```diff
@@ -323,14 +323,22 @@ export const registerPamWebAccessRouter = async (server: FastifyZodProvider) =>
...
-          .describe("Target host to connect to, for accounts that allow multiple hosts")
+          .describe("Target host to connect to, for accounts that allow multiple hosts"),
+        supportedTransports: z
+          .array(z.enum(["relay", "direct"]))
+          .max(2)
+          .optional()
+          .describe(
+            "Transports the client can dial. Omit to let the platform choose. A CLI that predates direct connections omits it, so the platform treats an omitted list as relay-only for CLI access."
+          )
       }),
```
sheensantoscapadngan: Why again do we have to pass this explicitly? Couldn't we just send a null value for direct address if the gateway does not support direct connect?

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -51,14 +64,32 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
...
         actorAuthMethod: req.permission.authMethod,
+        directAddress: req.body.directAddress,
         name: req.body.name
       });
+
+      await server.services.auditLog.createAuditLog({
+        ...req.auditLogInfo,
+        orgId: req.permission.orgId,
+        event: {
+          type: EventType.GATEWAY_ENROLL,
```
sheensantoscapadngan: Enroll somehow implies that we are connecting for the first time. Would "Connect" work here instead?

backend/src/ee/services/gateway-v2/gateway-v2-constants.ts
```diff
@@ -1,6 +1,169 @@
...
+    directAddress?: string | null;
+    relayId?: string | null;
+    directHeartbeat?: Date | null;
+    heartbeatTTL?: number | null;
+  };
+  transport?: TGatewayTransport;
+}) => {
+  const directUsable =
+    Boolean(gateway.directAddress) &&
+    (transport === "direct" ||
```
sheensantoscapadngan: let's use an enum please...

backend/src/ee/services/gateway-v2/gateway-v2-constants.ts
```diff
@@ -1,6 +1,169 @@
...
+  directAddress,
+  relayHost
+}: {
+  directAddress?: string;
+  relayHost?: string;
+}): TGatewayTransport[] => {
+  const transports: TGatewayTransport[] = [];
+  if (directAddress) transports.push("direct");
+  if (relayHost) transports.push("relay");
+  return transports;
```
sheensantoscapadngan: same here

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
bernie-g: Technically the array isn't needed but we need to know if the CLI is on an old version that doesn't support direct listen mode. The absence of this field indicates that the CLI is too old so we can send back a good error message like "This gateway only supports direct connections. Upgrade the Infisical CLI to connect to it.". Otherwise there's no way to know

backend/src/ee/routes/v2/gateway-router.ts
bernie-g: This is a pre-existing event type but yes I can rename it

bernie-g: Done

backend/src/ee/services/gateway-v2/gateway-v2-constants.ts
bernie-g: Done

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
sheensantoscapadngan: And why do we have to know that the CLI is an old version? Isn't the CLI the one that chooses whether to use the direct address or the relay host anyway and if it was the old version, it wouldn't parse the direct address field?

backend/src/ee/services/gateway-v2/gateway-v2-constants.ts
```diff
@@ -1,6 +1,176 @@
...
+    !parsed.port ||
+    parsed.username ||
+    parsed.password ||
+    parsed.pathname ||
+    parsed.search ||
+    parsed.hash
+  ) {
+    throw invalidFormat();
+  }
+
```
sheensantoscapadngan: Just noticed that this is a constant file. Is this intentional, or should we put some of these in an fns file instead?

bernie-g: Good catch, I'll fix

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -680,6 +714,16 @@ export const gatewayV2ServiceFactory = ({
...
 
+    const subjectAlternativeNames: x509.JsonGeneralName[] = [
+      { type: "dns", value: "localhost" },
+      { type: "ip", value: "127.0.0.1" },
+      { type: "ip", value: "::1" }
+    ];
+    if (gateway.directAddress) {
+      const { host } = parseDirectAddress(gateway.directAddress);
+      subjectAlternativeNames.push(net.isIP(host) ? { type: "ip", value: host } : { type: "dns", value: host });
+    }
```
sheensantoscapadngan, line 724: Q: do we already handle IPV6?

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
```diff
@@ -926,10 +987,45 @@ export const gatewayV2ServiceFactory = ({
       }
     }
 
-    await gatewayV2DAL.updateById(gatewayId, {
-      heartbeat: new Date(),
-      heartbeatTTL
+    await gatewayV2DAL.updateById(
```
sheensantoscapadngan, line 989: IGNORE PLACEMENT OF COMMENT

Does it still make sense for us to refer to this as the setupRelayServer method, even when we're using direct address, there is no relay server in the first place?

Image

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
```diff
@@ -404,30 +400,25 @@ export const pamWebAccessServiceFactory = ({
...
+                    cancelSocket.end();
+                  });
+                  cancelSocket.setTimeout(PAM_CANCELLATION_FLUSH_TIMEOUT_MS, () => {
+                    cancelSocket.destroy();
+                    resolve();
+                  });
+                  cancelSocket.on("close", () => resolve());
+                  cancelSocket.on("error", reject);
+                }),
+              { ...certs, protocol: GatewayProxyProtocol.PamSessionCancellation }
```
sheensantoscapadngan: "certs" here isn't actually just certificates right?  conditional is a bit odd now that relays are optional 

Image

backend/src/lib/gateway-v2/gateway-v2.ts
```diff
@@ -236,30 +251,71 @@ export const setupRelayServer = async ({
...
+
+    const dialDirect = async () => {
+      const parsed = new URL(`tcp://${directAddress}`);
+      const serverName = parsed.hostname.startsWith("[") ? parsed.hostname.slice(1, -1) : parsed.hostname;
+      // Rechecked immediately before connecting, not just at registration: a name that passed then
+      // can be rebound since. Connecting by name rather than by the resolved address keeps Node's
+      // own multi-address fallback, which a dual-stack gateway relies on.
+      await assertHostNotInfisicalInfrastructure({ host: serverName });
+      return { conn: net.connect({ host: serverName, port: Number(parsed.port) }), serverName, direct: true };
+    };
```
sheensantoscapadngan, line 268: Same remark here. We might have to rename the methods because this being inside a setup relay server is contradictory to it being a direct dial

backend/src/lib/gateway-v2/gateway-v2.ts
```diff
@@ -236,30 +251,71 @@ export const setupRelayServer = async ({
...
+      const { conn, serverName, direct } = await dial();
+      try {
+        const gwConn = await createGatewayConnection(
+          conn,
+          gateway,
+          protocol,
+          tunnelId,
+          serverName,
+          direct ? DIRECT_HANDSHAKE_TIMEOUT_MS : undefined
+        );
```
sheensantoscapadngan, line 295: claude remark:
Image

backend/src/services/telemetry/telemetry-dal.ts
```diff
@@ -169,13 +169,7 @@ export const telemetryDALFactory = (db: TDbClient) => {
...
 
       const v2ActiveResult = (
-        await db(TableName.GatewayV2)
-          .whereNotNull("heartbeat")
-          .whereRaw(
-            `COALESCE("heartbeatTTL", 0) > 0 AND "heartbeat" + make_interval(secs => COALESCE("heartbeatTTL", 0) + ${HEARTBEAT_BUFFER_SECONDS}) > NOW()`
-          )
-          .count()
-          .first()
+        await db(TableName.GatewayV2).whereRaw(buildGatewayReachableSql(TableName.GatewayV2)).count().first()
```
sheensantoscapadngan: Q: why do we pass the table name as a parameter here when it will always be `gateway_v2`?

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
```diff
@@ -323,14 +324,22 @@ export const registerPamWebAccessRouter = async (server: FastifyZodProvider) =>
...
       response: {
         200: z.object({
           sessionId: z.string().describe("The ID of the created session"),
           accountType: z.nativeEnum(PamAccountType).describe("The account type"),
           metadata: z.record(z.string()).optional().describe("Account-type-specific metadata (e.g., username)"),
           relayHost: z.string().optional().describe("The relay host to connect to"),
+          directAddress: z.string().optional().describe("The gateway address for a direct connection"),
```
sheensantoscapadngan, line 334: IGNORE PLACEMENT OF COMMENT

should we also have bind field here? 
Image

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -1,12 +1,24 @@
...
+The Gateway is deployed inside your private network, alongside the resources Infisical needs to reach. Infisical never touches those resources directly. It sends every request through the Gateway, which forwards it to the database or internal API on the other side.
 
-## Connection model
+There are two ways the Gateway and Infisical reach each other, and they differ in one thing: which side opens the connection.
+
+- **Relay mode.** The Gateway dials out to a [relay server](/documentation/platform/gateways/relay-deployment/overview) and keeps that tunnel open. Infisical connects to the relay, and its requests travel back down the tunnel the Gateway already established. The Gateway accepts no inbound connections at all.
+- **Direct listen mode.** The Gateway listens on an address you register, and Infisical connects to it. There's no relay to deploy, but the address has to be reachable from wherever Infisical runs.
+
+Both modes carry the same mutually authenticated TLS between Infisical and the Gateway, with certificates issued by your organization's own certificate authorities. In relay mode the relay routes bytes without holding the keys to read them.
```
sheensantoscapadngan: Perhaps we can just omit this because this is the network architecture page anyway?

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -1,12 +1,24 @@
...
+
+<Info>
+  Direct listen is available on self-hosted Infisical. Cloud runs outside your network and has no
+  route into it, leaving no address it could dial, so Cloud gateways use relay mode instead.
+</Info>
```
sheensantoscapadngan: Maybe we can just briefly say that DirectListen is only available on self-hosted instances and end it at that?

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -42,33 +50,94 @@ On the Gateway host, allow **outbound** access only:
...
 - **Outbound TCP 443** to the Infisical instance host (relay API + certificates)
 
 ## Frequently asked questions
 
 <AccordionGroup>
   <Accordion title="Does the Gateway require any inbound ports?">
-    No. The Gateway makes only outbound connections (TCP 2222 to the relay and TCP 443 to the Infisical API). The platform reaches the Gateway back over its own outbound SSH tunnel via the relay, so no inbound firewall rules are needed.
+    In relay mode, no. The Gateway makes only outbound connections (TCP 2222 to the relay and TCP 443 to the Infisical API), and the platform reaches it back over that same tunnel.
+
+    In direct listen mode, yes. The Gateway accepts inbound connections on the port you set with `--listen-address`, and that port needs to be open to the Infisical instance host. Scope the rule to that host rather than opening the port broadly.
```
sheensantoscapadngan, line 92: IGNORE PLACEMENT

Should we still have this network segmentation?
Image

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -42,33 +50,94 @@ On the Gateway host, allow **outbound** access only:
...
+| Infisical API | Gateway | Infisical instance host | 443 | TCP (HTTPS) | API communication and certificate requests. **Egress** |
+| Resource access | Gateway | Private resource | the resource's own port (e.g. 5432) | TCP or HTTP | Gateway forwards platform requests to the internal service. **Local** |
+
+On the Gateway host, allow:
+
+- **Inbound** access on the port you set with `--listen-address`, from the Infisical instance host only
+- **Outbound TCP 443** to the Infisical instance host
+
+The address you register is the one Infisical dials, and it's written into the Gateway's server certificate, so it has to be stable. An internal DNS name or a load balancer in front of the Gateway both work. Changing it means re-running the deploy command so the Gateway is issued a certificate for the new address.
+
```
sheensantoscapadngan, line 75: I think it will make it easier to parse if you consistently refer to this as a listen address

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -42,33 +50,94 @@ On the Gateway host, allow **outbound** access only:
...
+| Gateway listener | Infisical Platform | Gateway's registered address | the port you set with `--listen-address` | TCP + mTLS | Platform connects to the gateway to serve every operation. **Ingress** on the Gateway host |
+| Infisical API | Gateway | Infisical instance host | 443 | TCP (HTTPS) | API communication and certificate requests. **Egress** |
+| Resource access | Gateway | Private resource | the resource's own port (e.g. 5432) | TCP or HTTP | Gateway forwards platform requests to the internal service. **Local** |
+
+On the Gateway host, allow:
+
+- **Inbound** access on the port you set with `--listen-address`, from the Infisical instance host only
+- **Outbound TCP 443** to the Infisical instance host
+
+The address you register is the one Infisical dials, and it's written into the Gateway's server certificate, so it has to be stable. An internal DNS name or a load balancer in front of the Gateway both work. Changing it means re-running the deploy command so the Gateway is issued a certificate for the new address.
```
sheensantoscapadngan: maybe we can just say "restarting the gateway"

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -42,33 +50,94 @@ On the Gateway host, allow **outbound** access only:
...
+
+| Flags | Mode |
+|---|---|
+| `--relay=my-relay` | Relay only |
+| `--relay=my-relay --listen-address=gateway.internal:8443` | Both |
+| `--listen-address=gateway.internal:8443` | Direct only |
+
+Stop at the middle step and check the Gateway's health. A wrong address costs you nothing there: operations keep completing over the relay, and the Gateway page reports the direct path as unreachable. Drop `--relay` once direct is healthy, or drop `--listen-address` to go back.
+
+Running both permanently only helps clients that can reach the direct address. PAM CLI sessions dial the Gateway from the user's own machine. A user outside the network waits for that attempt to time out before the session falls back to the relay. If most of your CLI users are outside the network, relay mode alone serves them better.
```
sheensantoscapadngan: hmm not sure if this is the right place to talk about how to run both modes at once. we would want to focus on the networking aspect here

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -42,33 +50,94 @@ On the Gateway host, allow **outbound** access only:
...
 <AccordionGroup>
   <Accordion title="Does the Gateway require any inbound ports?">
-    No. The Gateway makes only outbound connections (TCP 2222 to the relay and TCP 443 to the Infisical API). The platform reaches the Gateway back over its own outbound SSH tunnel via the relay, so no inbound firewall rules are needed.
+    In relay mode, no. The Gateway makes only outbound connections (TCP 2222 to the relay and TCP 443 to the Infisical API), and the platform reaches it back over that same tunnel.
+
+    In direct listen mode, yes. The Gateway accepts inbound connections on the port you set with `--listen-address`, and that port needs to be open to the Infisical instance host. Scope the rule to that host rather than opening the port broadly.
+  </Accordion>
+
+  <Accordion title="Which mode should I use?">
+    Use direct listen when Infisical can already reach the Gateway host at a stable address, which is the common case when you self-host Infisical in the same network as your resources. It needs no relay.
```
sheensantoscapadngan: nit - no need to mention "It needs no relay" since it would be self explanatory

docs/documentation/platform/gateways/network-architecture.mdx
```diff
@@ -42,33 +50,94 @@ On the Gateway host, allow **outbound** access only:
...
+
+    Use relay mode when the Gateway sits behind NAT or a firewall you can't open, when you're on Infisical Cloud, or when PAM CLI users connect from outside the Gateway's network. See [Gateway deployment](/documentation/platform/gateways/gateway-deployment) for the full comparison.
+  </Accordion>
+
+  <Accordion title="Do PAM sessions work in direct listen mode?">
+    Browser-based access is unaffected, because the platform brokers that connection. `infisical pam access` opens its own connection to the Gateway, so it works only from a machine that can reach the registered address.
+
+    See [Gateway deployment](/documentation/platform/gateways/gateway-deployment#frequently-asked-questions) for what to give users outside the network.
   </Accordion>
```
sheensantoscapadngan: I think we can remove this section too since it's PAM specific?

docs/documentation/platform/gateways/relay-deployment/overview.mdx
```diff
@@ -7,11 +7,23 @@ Infisical Relay is a secure routing layer that allows Infisical to connect to yo
...
 With this architecture, you can achieve secure, firewall-friendly access across network boundaries, making it possible for Infisical to interact with resources even in highly restricted environments.
 
-Before diving in, it's important to determine whether you actually need to deploy your own relay server or if you can use Infisical's managed infrastructure.
+Before diving in, work out whether you need a relay at all, and if you do, whether you can use Infisical's managed infrastructure.
 
-## Do you need to deploy a Relay?
+## Do you need a relay at all?
 
-Not all users need to deploy their own relay servers. Infisical provides managed relay infrastructure in US/EU regions for Infisical Cloud users, which requires no setup or maintenance. You only need to deploy a relay if you:
+If you self-host Infisical and it can reach your gateway host at a stable address, you can skip relays entirely. Deploy the gateway in [direct listen mode](/documentation/platform/gateways/gateway-deployment) instead, and Infisical connects to it.
```
sheensantoscapadngan: Can we point them to a specific section of the gateway deployment page referring to the direct listen mode?

docs/documentation/platform/gateways/relay-deployment/overview.mdx
```diff
@@ -7,11 +7,23 @@ Infisical Relay is a secure routing layer that allows Infisical to connect to yo
...
+
+A relay is what you need when:
+
+- The gateway sits behind NAT or a firewall you can't open for inbound connections
+- You're on Infisical Cloud, which runs outside your network and has no route into it, so a gateway there can't be dialed and has to open the connection itself
```
sheensantoscapadngan: I think the "has to open the connection itself" can be omitted?

docs/contributing/platform/pam-development.mdx
```diff
@@ -158,7 +158,7 @@ docker compose -f docker-compose.dev.yml logs -f backend
 
 <Accordion title="Gateway cannot connect to relay">
 - Verify relay is running and registered in the UI
-- Check `--target-relay-name` matches relay's `--name`
+- Check `--relay` matches relay's `--name`
 - Ensure port 2222 isn't blocked
```
sheensantoscapadngan: Perhaps for PAM local development, we can just advocate for the use of direct listen mode?

docs/cli/commands/gateway.mdx
```diff
@@ -279,9 +362,17 @@ The gateway name is provided as a positional argument.
...
+  </Accordion>
+  <Accordion title="Which mode do I get if I pass no connection flags?">
+      Relay mode, with a relay chosen automatically. Direct listen is opt-in: the gateway only listens for connections when you pass `--listen-address`.
+  </Accordion>
+  <Accordion title="Do PAM sessions work in direct listen mode?">
+      Browser-based sessions work in either mode, because the platform brokers the connection. `infisical pam access` connects from your own machine instead, so it needs to reach the gateway's registered address.
 
-      If the selected relay becomes unreachable while the gateway is running, the gateway will automatically switch to a different healthy one. Gateways started with an explicit `--target-relay-name` do not switch and will keep retrying the specified relay.
+      See [Gateway deployment](/documentation/platform/gateways/gateway-deployment#frequently-asked-questions) for what to give users outside the network.
```
sheensantoscapadngan: Same feedback here with regards to placement. Maybe we can explicitly mention this in one of the PAM-specific pages instead.

docs/cli/commands/gateway.mdx
```diff
@@ -208,17 +265,43 @@ sudo infisical gateway systemd install my-gateway --enroll-method=token --token=
...
+    ```bash
+    # Example
+    sudo infisical gateway systemd install my-gateway --relay=my-relay --enroll-method=token --token=<token> --domain=<domain>
+    ```
+
+    For Infisical Cloud users on instance relays, the relay infrastructure is already running and managed by Infisical. If using organization relays or self-hosted instance relays, start a relay server first. See the [Relay Deployment Guide](/documentation/platform/gateways/relay-deployment).
+
+    <Note>
+      This flag was previously named `--target-relay-name`, which still works but is deprecated.
+    </Note>
```
sheensantoscapadngan: We can just remove this

docs/cli/commands/gateway.mdx
```diff
@@ -208,17 +265,43 @@ sudo infisical gateway systemd install my-gateway --enroll-method=token --token=
...
+
+  </Accordion>
+
+  <Accordion title="--listen-address">
+    The `host:port` address that Infisical dials to reach this gateway, which runs it in direct listen mode. The systemd unit captures the flag, so every restart advertises the same address.
```
sheensantoscapadngan: The detail about system decapturing the same flag might be unnecessary here, since we pretty much do that for all the other flags too.

docs/cli/commands/gateway.mdx
```diff
@@ -120,17 +133,61 @@ Once started, the gateway component will:
...
     ```
 
-    **Note:** For Infisical Cloud users using instance relays, the relay infrastructure is already running and managed by Infisical. If using organization relays or self-hosted instance relays, you must first start a relay server. For more information on deploying relays, refer to the [Relay Deployment Guide](/documentation/platform/gateways/relay-deployment).
+    Can also be set with the `INFISICAL_RELAY_NAME` environment variable.
+
+    For Infisical Cloud users on instance relays, the relay infrastructure is already running and managed by Infisical. If using organization relays or self-hosted instance relays, start a relay server first. See the [Relay Deployment Guide](/documentation/platform/gateways/relay-deployment).
+
+    <Note>
+      This flag was previously named `--target-relay-name`, which still works but is deprecated.
+    </Note>
```
sheensantoscapadngan: Same remark here. We can just remove for simplicity.

docs/cli/commands/gateway.mdx
```diff
@@ -120,17 +133,61 @@ Once started, the gateway component will:
...
+
+    ```bash
+    infisical gateway start my-gateway --listen-address=gateway.internal:8443 --relay=my-relay --enroll-method=token --token=<token> --domain=<domain>
+    ```
+
+    Can also be set with the `INFISICAL_GATEWAY_LISTEN_ADDRESS` environment variable.
+
+    <Note>
+      Direct listen is available on self-hosted Infisical. Infisical Cloud runs outside your network and has no route into it, leaving no address it could dial. On Cloud, use `--relay` or omit both flags.
+    </Note>
```
sheensantoscapadngan, line 172: Same remark here about just cutting it off at the first sentence.

helm-charts/infisical-gateway/values.yaml
```diff
@@ -54,6 +54,11 @@ gateway:
   # Specific relay to connect to. Leave empty for automatic selection.
   relayName: ""
 
+  # Stable host:port advertised by a gateway accepting direct connections.
+  # Requires an image.tag whose CLI supports --listen-address; older images reject the flag and the
+  # pod crash-loops on startup.
+  listenAddress: ""
```
sheensantoscapadngan, line 60: should we care about bind here too?

helm-charts/infisical-gateway/templates/service.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.gateway.listenAddress }}
```
sheensantoscapadngan, line 1: Oh, interesting. Why did we have this before in the first place? Was this not used at all?

docs/self-hosting/helm-charts/gateway.mdx
```diff
@@ -122,6 +123,31 @@ helm uninstall infisical-gateway --namespace infisical-gateway
...
+  --set gateway.domain=<your-infisical-domain>
+```
+
+The chart takes the container port from `gateway.listenAddress`, points the Service at it, and exposes that same port, which keeps all three in step. A `host:port` value is required, and the chart fails the render if the port is missing. Set `service.port` only if you want the Service to answer on a different port than the gateway binds, in which case the address you register has to name the Service's port.
+
+The Service is created only in direct listen mode. A relay-mode gateway accepts no connections and needs none.
+
+The address has to be one Infisical can resolve and reach. An in-cluster Service DNS name works when you self-host Infisical in the same cluster. From outside the cluster, expose the gateway through an Ingress or a `LoadBalancer` Service and register that address instead.
+
+Direct listen is available on self-hosted Infisical. Infisical Cloud runs outside your cluster and has no route into it, leaving no address it could dial. On Cloud, leave `gateway.listenAddress` empty and use a relay.
```
sheensantoscapadngan: same here

docs/documentation/platform/gateways/overview.mdx
sheensantoscapadngan, line 51: Shouldn't we update the overview page of the Gateway accordingly too? Right now, I don't see it in the docs though which is odd

Image

docs/documentation/platform/gateways/gateway-deployment.mdx
```diff
@@ -3,23 +3,48 @@ title: "Gateway deployment"
...
+| Who connects | Infisical connects to the Gateway | The Gateway connects out to a relay |
+| Firewall | Inbound port open to the Infisical host | Outbound only |
+| Relay required | No | Yes |
+| Needs a stable Gateway address | Yes | No |
+| PAM CLI from outside the network | No | Yes |
+| Infisical Cloud | Not available | Yes |
+
+**Use direct listen** when Infisical can already reach the Gateway host at a stable address. This is the common case when you self-host Infisical in the same network as your resources, and it means there's no relay to deploy or operate.
+
+**Use relay mode** when the Gateway sits behind NAT or a firewall you can't open, when you're on Infisical Cloud, or when PAM CLI users connect from outside the Gateway's network.
```
sheensantoscapadngan: claude feedback:
```

  This opens from a mechanism and asks the reader to reason forward to the consequences.
  At this point they don't yet know that "which side opens the connection" is what
  determines their firewall work, so it can't help them decide anything until after
  they've read the table.

  Leading with cost instead would land better. The reader already knows whether they're
  on Cloud, whether the host has a stable address, and whether they can get a firewall
  rule opened. That last one is usually a ticket to the network team, which is the real
  cost of this decision. And for Cloud users this isn't a decision at all, it's a lookup,
  so that belongs on line one rather than in row four.

  Suggested rewrite:

  ## Choose a connection mode

  There are two ways to connect a Gateway, and your network usually decides which one
  you can use:

  - **On Infisical Cloud?** Use relay mode. Cloud runs outside your network and has no
    route into it.
  - **Self-hosted, and you can open an inbound port on the Gateway host?** Use direct
    listen. It's the simpler setup, with no relay in the path.
  - **Self-hosted, but the Gateway sits behind NAT or a firewall you can't change?**
    Use relay mode.

  |  | Direct listen | Relay |
  |---|---|---|
  | Firewall change | Inbound port open to the Infisical host | None, outbound only |
  | Relay to run | None | Managed on Cloud (US/EU), otherwise your own |
  | Stable Gateway address | Required | Not needed |
  | Infisical Cloud | Not available | Yes |
  | PAM CLI from outside the network | No | Yes |

  The difference underneath is which side opens the connection, and that's what drives
  every row above. See [Network architecture](/documentation/platform/gateways/network-architecture)
  for the ports and traffic flow.

```

docs/documentation/platform/gateways/gateway-deployment.mdx
```diff
@@ -45,13 +76,57 @@ This guide covers everything you need to deploy and configure Infisical Gateways
 
     If you're in a corporate environment with strict egress filtering, ensure outbound TCP 2222 to relay servers and outbound HTTPS 443 to Infisical API endpoints are allowed.
 
+    **Direct listen mode.** The Gateway accepts inbound connections. One port has to be open to Infisical.
+
+    | Protocol | Direction | Source or destination | Port | Purpose |
```
sheensantoscapadngan: Wait, why is this source or destination? Seems a bit confusing.

docs/documentation/platform/gateways/gateway-deployment.mdx
```diff
@@ -257,24 +332,49 @@ This guide covers everything you need to deploy and configure Infisical Gateways
...
+Direct listen is also unavailable on Infisical Cloud, which runs outside your network and has no route into it. On Cloud, use relay mode.
+
+</Accordion>
+
+<Accordion title="Do PAM sessions work in direct listen mode?">
+Only from a machine that can reach the gateway's registered address. The CLI connects to the gateway from the user's own machine rather than through Infisical, so a laptop outside the network can't reach a gateway listening on an internal address. Give those users a VPN into the network, or run a relay on the gateway alongside direct listen.
+
+Browser-based PAM access works in either mode, because the platform brokers that connection.
+
+Using direct listen also requires a current Infisical CLI. An older CLI can't connect to a gateway that only listens directly, and reports that the gateway needs a newer CLI.
```
sheensantoscapadngan: same here

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
```diff
@@ -323,14 +324,22 @@ export const registerPamWebAccessRouter = async (server: FastifyZodProvider) =>
           .max(255)
           .regex(hostPattern, "Must be a valid hostname or IP address")
           .optional()
-          .describe("Target host to connect to, for accounts that allow multiple hosts")
+          .describe("Target host to connect to, for accounts that allow multiple hosts"),
+        supportedTransports: z
+          .array(z.nativeEnum(GatewayTransport))
```
sheensantoscapadngan: IGNORE PLACEMENT

it might be better to show this as two different rows instead of saying "Direct + relay" up top...
Image

and would it make sense to only show the heartbeat details on hover of the badge?

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -1,6 +1,8 @@
 import z from "zod";
 
 import { GatewaysV2Schema } from "@app/db/schemas";
+import { EventType } from "@app/ee/services/audit-log/audit-log-types";
+import { gatewayTransports } from "@app/ee/services/gateway-v2/gateway-v2-transport-fns";
```
sheensantoscapadngan, line 5: IGNORE PLACEMENT

does it necessarily have to be in the exact same private network or is reachability the overarching requirement? 
Image

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
bernie-g: Switched to adding the version to the user agent and removed this field.

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
bernie-g: Yes its supported if you use brackets []

backend/src/ee/services/gateway-v2/gateway-v2-service.ts
bernie-g: Renamed to setupGatewayProxy

backend/src/ee/services/pam-web-access/pam-web-access-service.ts
bernie-g: Fixed

backend/src/lib/gateway-v2/gateway-v2.ts
bernie-g: Done

backend/src/lib/gateway-v2/gateway-v2.ts
bernie-g: Already fixed but yes

backend/src/services/telemetry/telemetry-dal.ts
bernie-g: Removed

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
bernie-g: I don't think so. This feels like internal state for the gateway only and the platform doesn't need to care about this.

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: done

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: done

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: Oops good catch

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: Simplified

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: done

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: done

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: done

docs/documentation/platform/gateways/network-architecture.mdx
bernie-g: done

docs/documentation/platform/gateways/relay-deployment/overview.mdx
bernie-g: done

docs/documentation/platform/gateways/relay-deployment/overview.mdx
bernie-g: done

docs/cli/commands/gateway.mdx
bernie-g: done

docs/cli/commands/gateway.mdx
bernie-g: done

docs/cli/commands/gateway.mdx
bernie-g: done

docs/cli/commands/gateway.mdx
bernie-g: done

docs/self-hosting/helm-charts/gateway.mdx
bernie-g: done

docs/cli/commands/gateway.mdx
bernie-g: done

docs/documentation/platform/gateways/gateway-deployment.mdx
bernie-g: done

docs/documentation/platform/gateways/gateway-deployment.mdx
bernie-g: done

docs/documentation/platform/gateways/gateway-deployment.mdx
bernie-g: done

docs/documentation/platform/gateways/overview.mdx
bernie-g: done

docs/contributing/platform/pam-development.mdx
bernie-g: done

helm-charts/infisical-gateway/values.yaml
bernie-g: I don't think so since a pod can only have one network interface

helm-charts/infisical-gateway/templates/service.yaml
bernie-g: Yea must not have been used before, so not sure

backend/src/ee/routes/v1/pam-routers/pam-session-router.ts
bernie-g: Done

backend/src/ee/routes/v2/gateway-router.ts
bernie-g: Updated it to be about reachability

backend/src/ee/services/dynamic-secret/dynamic-secret-fns.ts
```diff
@@ -42,6 +26,53 @@ export const verifyHostInputValidity = async ({
...
+    try {
+      hostIps = (await dns.lookup(host, { all: true })).map(({ address }) => address);
+    } catch {
+      // A gateway is often registered before its DNS record exists.
+      return;
+    }
+  }
+
+  const exclusiveIps = await getReservedIps();
+  if (hostIps.some((el) => exclusiveIps.includes(el))) {
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/routes/v2/gateway-router.ts
```diff
@@ -25,24 +30,32 @@ export const registerGatewayV2Router = async (server: FastifyZodProvider) => {
...
+        .refine((body) => Boolean(body.relayName || body.directAddress), {
+          message: "Either relayName or directAddress is required"
+        }),
       response: {
         200: z.object({
           gatewayId: z.string(),
-          relayHost: z.string(),
+          directAddress: z.string().optional(),
+          relayHost: z.string().optional(),
           pki: z.object({
```
sheensantoscapadngan, line 47: IGNORE PLACEMENT

migration is not idempotent
Image

sheensantoscapadngan:
@greptile-apps  review this again

docs/documentation/platform/pam/architecture.mdx
```diff
@@ -34,16 +34,20 @@ The actor never sees the credentials. The database/server never sees the actor d
...
 ## Network security
 
-The Gateway uses outbound connections only:
+In relay mode, the Gateway uses outbound connections only:
 
 - **No inbound ports** — your firewall doesn't need to allow any inbound traffic
 - **Outbound to Infisical** — for credential retrieval and session uploads
 - **Outbound to your infrastructure** — to connect to databases and servers
 
+In [direct listen mode](/documentation/platform/gateways/gateway-deployment#choose-a-connection-mode), Infisical opens the connection instead, so the Gateway accepts inbound traffic on its listen address from the Infisical host.
```
sheensantoscapadngan, line 43: Can we change the framing of this section here so that it's more about running both at once rather than putting them through the scenario of switching modes? Also, for the table, it might be better to do Relay only, direct only, and then both, so that there's a clear order. 

Image

docs/documentation/platform/pam/architecture.mdx
```diff
@@ -34,16 +34,20 @@ The actor never sees the credentials. The database/server never sees the actor d
...
+
 <Tip>
-Your databases and servers stay completely isolated from the internet. The Gateway sits inside your network and reaches out; nothing reaches in.
+Your databases and servers stay completely isolated from the internet either way. Only the Gateway is reachable, and only from Infisical.
 </Tip>
 
+Direct listen changes where a CLI session connects from. Browser-based sessions work in either mode, because the platform brokers the connection. `infisical pam access` connects from the user's own machine instead, so it needs to reach the Gateway's listen address. Users outside that network need a VPN into it, or a relay running on the Gateway alongside direct listen.
+
 ## Encryption
```
sheensantoscapadngan, line 52: Noticed that this page here is lost. Do you think you can include it as part of the PR? It should be quick
Image

docs/self-hosting/helm-charts/gateway.mdx
```diff
@@ -103,6 +103,7 @@ helm uninstall infisical-gateway --namespace infisical-gateway
 | `gateway.enrollment.kubernetes.externalTokenReview` | `false` | Set true when TokenReview permission comes from outside this chart, either a token reviewer JWT configured in Infisical or a binding you manage. Required to install with `serviceAccount.createAsAuthDelegator` disabled. |
 | `gateway.enrollment.credentialsSecretRef` | `""` | Kubernetes secret with credentials injected as env vars. |
 | `gateway.relayName` | `""` | Specific relay to connect to. Empty for automatic selection. |
+| `gateway.listenAddress` | `""` | `host:port` that Infisical dials to reach the gateway, which runs it in direct listen mode. Empty for relay mode. |
 | `persistence.enabled` | `true` | Enable persistent storage. Required for token auth, optional for AWS and Kubernetes auth. |
 | `persistence.storageClass` | `""` | Storage class for the PVC. |
 | `persistence.size` | `5Gi` | PVC size. |
```
sheensantoscapadngan, line 109: Could we instead make it so that the orientation here is horizontal? it looks a bit odd

Image

docs/self-hosting/helm-charts/gateway.mdx
```diff
@@ -122,6 +123,31 @@ helm uninstall infisical-gateway --namespace infisical-gateway
...
+  --create-namespace \
+  --set gateway.name=<gateway-name> \
+  --set gateway.listenAddress=infisical-gateway.infisical-gateway.svc.cluster.local:8443 \
+  --set gateway.enrollment.method=token \
+  --set gateway.enrollment.token.value=<enrollment-token> \
+  --set gateway.domain=<your-infisical-domain>
+```
+
+The chart takes the container port from `gateway.listenAddress`, points the Service at it, and exposes that same port, which keeps all three in step. A `host:port` value is required, and the chart fails the render if the port is missing. Set `service.port` only if you want the Service to answer on a different port than the gateway binds, in which case the listen address has to name the Service's port.
+
```
sheensantoscapadngan, line 145: Maybe this section over here should come under the relay mode section instead so it flows more naturally? 

Image

docs/self-hosting/helm-charts/gateway.mdx
```diff
@@ -181,6 +207,11 @@ gateway:
   # Specific relay to connect to. Leave empty for automatic selection.
   relayName: ""
 
+  # Stable host:port advertised by a gateway accepting direct connections.
+  # Requires an image.tag whose CLI supports --listen-address; older images reject the flag and the
```
sheensantoscapadngan, line 211: Could it make sense to use an accordion here so that it doesn't seem so overwhelming as a big block of text? 

Image

docs/documentation/platform/pam/architecture.mdx
sheensantoscapadngan: Or maybe it can be one of the FAQs? You did already mention above this that you can run both at the exact same time.

docs/documentation/platform/pam/architecture.mdx
```diff
@@ -34,16 +34,20 @@ The actor never sees the credentials. The database/server never sees the actor d
...
 - **No inbound ports** — your firewall doesn't need to allow any inbound traffic
 - **Outbound to Infisical** — for credential retrieval and session uploads
 - **Outbound to your infrastructure** — to connect to databases and servers
 
+In [direct listen mode](/documentation/platform/gateways/gateway-deployment#choose-a-connection-mode), Infisical opens the connection instead, so the Gateway accepts inbound traffic on its listen address from the Infisical host.
+
 <Tip>
-Your databases and servers stay completely isolated from the internet. The Gateway sits inside your network and reaches out; nothing reaches in.
+Your databases and servers stay completely isolated from the internet either way. Only the Gateway is reachable, and only from Infisical.
 </Tip>
```
sheensantoscapadngan, line 48: I still see an FAQ for PAM on the Gateway page. Is this absolutely necessary to have here? 
Image

docs/self-hosting/helm-charts/gateway.mdx
```diff
@@ -122,6 +123,31 @@ helm uninstall infisical-gateway --namespace infisical-gateway
...
+  --set gateway.domain=<your-infisical-domain>
+```
+
+The chart takes the container port from `gateway.listenAddress`, points the Service at it, and exposes that same port, which keeps all three in step. A `host:port` value is required, and the chart fails the render if the port is missing. Set `service.port` only if you want the Service to answer on a different port than the gateway binds, in which case the listen address has to name the Service's port.
+
+The Service is created only in direct listen mode. A relay-mode gateway accepts no connections and needs none.
+
+The address has to be one Infisical can resolve and reach. An in-cluster Service DNS name works when you self-host Infisical in the same cluster. From outside the cluster, expose the gateway through an Ingress or a `LoadBalancer` Service and register that address instead.
+
+Direct listen is only available on self-hosted Infisical.
```
sheensantoscapadngan, line 151: Shouldn't we also update this page accordingly? 

Image

backend/src/ee/services/dynamic-secret/dynamic-secret-fns.ts
```diff
@@ -42,6 +26,73 @@ export const verifyHostInputValidity = async ({
...
+    hostIps = [host];
+  } else {
+    try {
+      hostIps = (await dns.lookup(host, { all: true })).map(({ address }) => address);
+    } catch {
+      // A gateway is often registered before its DNS record exists.
+      return;
+    }
+  }
+
```
sheensantoscapadngan, line 61: Let's cross-reference with the gateway docs?

Image

backend/src/ee/routes/v2/gateway-router.ts
bernie-g: done

docs/documentation/platform/pam/architecture.mdx
bernie-g: done

docs/documentation/platform/pam/architecture.mdx
bernie-g: done

docs/self-hosting/helm-charts/gateway.mdx
bernie-g: done

docs/self-hosting/helm-charts/gateway.mdx
bernie-g: done

docs/self-hosting/helm-charts/gateway.mdx
bernie-g: done

docs/self-hosting/helm-charts/gateway.mdx
bernie-g: done

docs/documentation/platform/pam/architecture.mdx
bernie-g: I think it's pretty important to have since it's a major downside of not using a relay and we want customers to know ahead of time so they don't have to redo their setup and add a relay later when they realize PAM CLI doesn't work

backend/src/ee/services/dynamic-secret/dynamic-secret-fns.ts
bernie-g: done

backend/src/db/migrations/20260904120000_add-direct-gateway-transport.ts
```diff
@@ -0,0 +1,31 @@
...
+  await knex.schema.alterTable(TableName.GatewayV2, (table) => {
+    if (!hasDirectAddress) {
+      table.text("directAddress").nullable();
+    }
+    if (!hasDirectHeartbeat) {
+      table.timestamp("directHeartbeat", { useTz: true }).nullable();
+    }
+  });
+}
+
```
sheensantoscapadngan, line 18: Can we make it so that switching connection mode regenerates the enrollment token too? Encountered this during my testing

Image

sheensantoscapadngan (dismissed):

sheensantoscapadngan (approved):

PR #7995, opened 2026-09-04
fix(telemetry): bound HTTP metric cardinality and add core request counter

victorvhs017:
## Context

OpenTelemetry’s HTTP instrumentation was emitting unbounded metric series in production. The legacy HTTP semantic conventions label `http.server.duration` with `net.host.name` (the raw `Host` header), so scanners and misconfigured clients create a new time series per value. In multi-tenant/cloud deployments that set `OTEL_DROP_HIGH_CARDINALITY_METERS=true`, per-actor API counters are suppressed entirely, which also removes any low-cardinality request volume needed to compute error rates against `infisical.core.http.error.count`.

This change opts into stable HTTP semantic conventions (`OTEL_SEMCONV_STABILITY_OPT_IN=http`) before `HttpInstrumentation` loads, adds an allowlist View on the `@opentelemetry/instrumentation-http` meter to keep `http.server.request.duration` bounded, and introduces `infisical.core.http.request.count` on the InfisicalCore meter with labels limited to `http.request.method`, `http.route`, and `http.response.status_code`. The new counter is recorded on every response when telemetry is enabled, before the high-cardinality gate, so it survives `OTEL_DROP_HIGH_CARDINALITY_METERS=true`. HTTP methods are normalized to the OTel fixed set (`_OTHER` for unrecognized values) in both the new counter and existing error/rate-limit paths.

## Steps to verify the change

1. Start the backend with `OTEL_TELEMETRY_COLLECTION_ENABLED=true` and an OTLP metrics endpoint configured.
2. Send a few API requests (including one with a bogus `Host` header and one with a non-standard method like `FOOBAR`).
3. Confirm `infisical.core.http.request.count` increments with bounded labels (`http.request.method`, `http.route`, `http.response.status_code`) on every request.
4. Set `OTEL_DROP_HIGH_CARDINALITY_METERS=true`, restart, and repeat — the core request counter should still emit; `infisical.http.server.request.count` should not.
5. Inspect exported `http.server.request.duration` — labels should be `http.request.method`, `http.response.status_code`, and optionally `error.type`; `net.host.name` should be absent.
6. Trigger a 4xx/5xx and confirm `infisical.core.http.error.count` labels use normalized methods; compute error rate per route as `core.http.error.count / core.http.request.count`.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


victorvhs017:
@greptileai re-review

victorvhs017:
@greptileai re-review

backend/src/server/plugins/api-metrics.ts
```diff
@@ -24,14 +30,23 @@ const requestDurationHistogram = infisicalMeter.createHistogram("infisical.http.
...
+    const method = normalizeHttpMethod(request.method);
     const route = request.routeOptions.url;
     const { statusCode } = reply;
 
+    // Recorded ahead of the high-cardinality gate: these three labels are bounded (the route is Fastify's
+    // template, never the raw path), and they are the denominator for coreHttpErrorCounter, so a deployment
+    // that drops the per-actor meters must not lose them too.
+    coreHttpRequestCounter.add(1, {
+      "http.request.method": method,
+      "http.route": route ?? "unknown",
```
scott-ray-wilson, line 50: `STANDALONE_MODE` registers `@fastify/static` with `wildcard: false` (`serve-ui.ts:127`), which creates one Fastify route per file, so `request.routeOptions.url` here is a content-hashed asset path. Since `apiMetrics` is registered on the root instance at `app.ts:141` (before `registerServeUI` at `app.ts:184`), every asset request lands on this counter with a unique `http.route`, and the allowlist can't drop it because `http.route` is allowlisted. That's roughly 700 files x 2 methods x 200/304 on the one meter the docs tell large deployments to scrape exclusively. Could we skip recording when the route came from the static plugin, or fold asset routes to a single `/assets/*` label?

backend/src/lib/telemetry/telemetry-attributes.ts
```diff
@@ -54,3 +54,24 @@ export const INFISICAL_CORE_METER_ATTRIBUTES = [
 // labels (user.email, client.address, syncId, ...) flow through unchanged unless dropped wholesale via
 // OTEL_DROP_HIGH_CARDINALITY_METERS. Kept on by default for self-hosted; dropped in multi-tenant/cloud.
 export const HIGH_CARDINALITY_METER_NAMES = ["Infisical", "API", "SecretSyncs", "PkiSyncs", "Integrations"];
+
+// Meter used by @opentelemetry/instrumentation-http for http.server.duration / http.client.duration.
+export const HTTP_INSTRUMENTATION_METER_NAME = "@opentelemetry/instrumentation-http";
+
+// HttpInstrumentation reads OTEL_SEMCONV_STABILITY_OPT_IN in its constructor (no config hook), so instrumentation.ts
+// forces the "http" token via resolveHttpSemconvOptIn — not "http/dup", which doubles cardinality; other namespaces are preserved.
+export const resolveHttpSemconvOptIn = (current?: string): string => {
```
scott-ray-wilson, line 66: not sure if others consume this so maybe not problematic:

This renames `http.server.duration` (ms) to `http.server.request.duration` (s) for every existing deployment, and `resolveHttpSemconvOptIn` filters out `http/dup`, which is the OTel-sanctioned way to run both name sets during a dashboard migration. There's no log line when we override an operator's explicit value, `grep -rn OTEL_SEMCONV docs/` returns nothing, and `envars.mdx` doesn't list the variable as reserved. Two asks: honor an explicitly set `http/dup`, and add an upgrade note naming the old and new metric names plus the unit change.

backend/src/lib/telemetry/instrumentation.ts
```diff
@@ -80,6 +86,10 @@ const initTelemetryInstrumentation = ({
     {
       meterName: "InfisicalCore",
```
scott-ray-wilson, line 88: The InfisicalCore View sets `meterName` and `attributesProcessors` but no `aggregationCardinalityLimit`, so `MeterSharedState._registerMetricStorage` passes `undefined` and `DeltaMetricProcessor` falls back to `(undefined ?? 2000) - 1`. Reproduced with this exact View config: 2500 distinct `http.route` values export as 1999 points plus a single `otel.metric.overflow` bucket, injected after the attributes processor so the allowlist doesn't even hide it. Under `OTEL_EXPORT_TYPE=prometheus` (cumulative) that's permanent. Please set an explicit limit, or drop `http.response.status_code` to a status class to cut the product.

backend/src/server/plugins/api-metrics.ts
```diff
@@ -24,14 +30,23 @@ const requestDurationHistogram = infisicalMeter.createHistogram("infisical.http.
 
 export const apiMetrics = fp(async (fastify) => {
   fastify.addHook("onResponse", async (request, reply) => {
-    // Checked before assembling the attribute object below: it is built per response and every label on
-    // it is per-actor, so it is the largest source of the cardinality these meters would retain.
-    if (!shouldRecordHighCardinalityMetrics()) return;
+    if (!isTelemetryEnabled()) return;
 
-    const { method } = request;
+    const method = normalizeHttpMethod(request.method);
```
scott-ray-wilson: `normalizeHttpMethod` is assigned to the shared `method` local, so it also rewrites the label on `API_latency`, `infisical.http.server.request.count`, `infisical.http.server.request.duration`, and (via `error-handler.ts:112`) `infisical.http.server.error.count` and `infisical.rate_limit.exceeded.count`. Node accepts 35 methods against semconv's 10, so scanner traffic that used to report as PROPFIND/MKCOL/PURGE now folds to `_OTHER`. The joinability rationale only applies to `infisical.core.http.request.count`; the per-actor meters gain nothing and pay a label break that `monitoring-telemetry.mdx:405`/`:437` still contradicts. Suggest `const coreMethod = normalizeHttpMethod(request.method)` and leaving `method` raw.

backend/src/lib/telemetry/http-metrics.test.ts
```diff
@@ -0,0 +1,191 @@
...
+    const app = await buildApp();
+    await app.inject({ method: "GET", url: "/api/v1/workspace/abc-123" });
+
+    // The denominator has to survive the kill switch; that is the whole point of the split.
+    const core = await collect(CORE_REQUEST_COUNT);
+    expect(core).toHaveLength(1);
+    expect(core[0].value).toBe(1);
+
+    const legacy = await collect(LEGACY_REQUEST_COUNT);
+    expect(legacy).toHaveLength(0);
```
scott-ray-wilson, line 220: `new InMemoryMetricExporter(0)` at `:60` is DELTA, and `collect(CORE_REQUEST_COUNT)` at `:174` performs a full collection cycle that swaps `_activeCollectionStorage` for a fresh map, so the second `reader.collect()` at `:177` returns nothing for any instrument. I verified by mutation: setting `OTEL_DROP_HIGH_CARDINALITY_METERS = false` at `:167` leaves all 7 tests passing. This is the only assertion guarding the kill switch, so a regression that makes it a noop ships green. Suggest collecting once into a snapshot and filtering both metric names out of it, the way `http-instrumentation.test.ts` does.

scott-ray-wilson (changes requested):

backend/src/lib/telemetry/instrumentation.ts
victorvhs017: We made aggregationCardinalityLimit: 2000 explicit on both Views and will not raise it on the Prometheus path. We will not fold status codes. Overflow isn't what took ADOT down. We'll keep an eye out for any issues, but it's safe for now.

backend/src/server/plugins/error-handler.ts
scott-ray-wilson, line 301: just want to double check this one, otherwise looks good:

The follow-up restores raw methods for `API_latency` and the legacy request/error metrics, but [`infisical.rate_limit.exceeded.count` still calls `normalizeHttpMethod`](</Users/scott/Documents/projects/infisical/backend/src/server/plugins/error-handler.ts:298>). Before this PR it used `req.method`, so existing `PROPFIND`/`MKCOL` filters still break. This should use the raw method or be documented as a label migration.

scott-ray-wilson (dismissed):

backend/src/server/plugins/error-handler.ts
victorvhs017: that's expected! added a small note on the docs regarding the OTEL https conventions we are assuming here.

scott-ray-wilson (approved):
lgtm

PR #7998, opened 2026-09-04
fix(github-org-sync): safely sync teams for large orgs

scott-ray-wilson:
## Context

Supersedes #7989, whose two commits are included here unchanged. A user with ~500 GitHub teams and ~2k members can't run **Sync Now**: GitHub 502s on the nested `teams(first: 100) { members(first: 100) }` query.

Five things were broken. #7989 fixed the first two:

- **502 on large orgs.** The sync now paginates at `teams(first: 20)` with a fresh abort signal and retry per request. The old code shared one 30s `AbortSignal` across every page, so a large org couldn't finish even against a healthy GitHub.
- **Silent data loss.** `members(first: 100)` had no `pageInfo`, and the truncated list drives removals, so every "successful" sync deleted members 101+ from the group.
- **`groups.role` doesn't exist.** Migration `20260107083948_remove-old-memberships` dropped the column in January, but both sync paths still included it in group inserts. Creating any team group therefore failed before GitHub was reached.
- **The match loop and transaction consumed the request budget.** The old code made ~16M username-to-email comparisons on the event loop and held one transaction open across every team. The matcher is now a constant-time ID lookup, and each team reconciles in a short transaction whose reads use the transaction connection.
- **Bulk sync could grant access to the wrong person.** GitHub's team API identifies members by GitHub account, while the old code guessed which Infisical member owned each username from fragments of their email. The sync now requests GitHub's numeric account ID and joins it to the verified GitHub alias created during GitHub login. It never matches on username or public profile email. Unlinked GitHub accounts receive no new access. Before any writes, the sync verifies that every current member of an existing GitHub-named group has a verified GitHub link. If any do not, it stops and lists the affected members. Once identities are linked, exact reconciliation removes stale access.

Also, the **Sync Now** 403 demanded `admin:org` when `read:org` is all the feature needs, since it only reads from GitHub. The response now reports the actual number of removed memberships as well.

## Screenshots

N/A

## Steps to verify the change

- [x] `cd backend && npm run test:unit -- src/ee/services/github-org-sync` passes 11 tests
- [x] `cd backend && npx eslint --ext .ts src/ee/services/github-org-sync/ src/server/routes/index.ts` is clean
- [x] `cd backend && npm run type:check` reports 0 errors
- [x] `make lint-docs-branch` reports 0 errors, warnings, or suggestions
- [x] A live GraphQL query against the Infisical GitHub organization returns `databaseId` for a team member
- [x] Unit coverage verifies that an existing unlinked member blocks reconciliation before changes
- [ ] On an org with >100 teams and at least one team over 100 members, **Sync Now** completes and the large team's group keeps every linked member
- [ ] Logging in via GitHub OAuth to an org with a GitHub team that has no Infisical group yet succeeds
- [ ] A fine-grained token with only Organization > Members: read drives the whole sync
- [ ] A linked member is matched by GitHub account ID after changing their GitHub username
- [ ] An unlinked GitHub user whose username resembles an Infisical member's email receives no group access
- [ ] An existing group member without a verified GitHub link stops the sync before changes; after GitHub login, the next sync reconciles that access

Measured at the reported scale (640 teams, 2.3k members, 7k memberships) against a mock GitHub before the exact-identity follow-up:

| | before | paginated fetch |
| --- | --- | --- |
| Sync Now | 400 after 4x 502, page 1 never fetched | 32/32 team pages + 2 member pages, all 200 |
| 250-member team | HTTP 200, 150 memberships deleted | all 250 kept |

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/github-org-sync/github-org-sync-service.ts
+++ b/backend/src/ee/services/github-org-sync/github-org-sync-service.ts
@@ -37,6 +37,11 @@ import {
 
 const OctokitWithPlugin = Octokit.plugin(paginateGraphql);
 
+// The sync only ever reads from GitHub, so read:org is the whole requirement. Both 403 handlers
+// share this so they cannot drift apart again.
+const githubTeamAccessDeniedMessage = (githubOrgName: string) =>
+  `GitHub denied access to the teams in organization '${githubOrgName}'. A classic access token needs the 'read:org' scope. A fine-grained token needs Organization permissions > Members set to read, with the organization as its resource owner. GitHub also hides secret teams from anyone who is not an organization owner or a member of the team.`;
+
 // GitHub's GraphQL API times out (502) when one page resolves too many nodes; nesting members
 // under teams multiplies the per-page node count, so the team page is kept small.
 const GITHUB_TEAMS_PAGE_SIZE = 20;
@@ -672,8 +677,7 @@ export const githubOrgSyncServiceFactory = ({
         }
         if (statusCode === 403) {
           throw new BadRequestError({
-            message:
-              "GitHub access token lacks required permissions. Required: 1) 'read:org' scope for organization teams, 2) Token owner must be an organization member with team visibility access, 3) Organization settings must allow team visibility. Check GitHub token scopes and organization member permissions."
+            message: githubTeamAccessDeniedMessage(config.githubOrgName)
           });
         }
         if (statusCode === 404) {
@@ -773,8 +777,7 @@ export const githubOrgSyncServiceFactory = ({
         }
         if (statusCode === 403) {
           throw new BadRequestError({
-            message:
-              "GitHub access token lacks required permissions for organization team sync. Required: 1) 'admin:org' scope, 2) Token owner must be organization owner or have team read permissions, 3) Organization settings must allow team visibility. Check token scopes and user role."
+            message: githubTeamAccessDeniedMessage(config.githubOrgName)
           });
         }
         if (statusCode === 404) {
--- a/backend/src/ee/services/github-org-sync/github-org-sync-service.ts
+++ b/backend/src/ee/services/github-org-sync/github-org-sync-service.ts
@@ -206,6 +206,44 @@ export const fetchGithubOrgTeams = async (octokit: Pick<Octokit, "graphql">, org
   return teams.map(({ slug, name, description, members }) => ({ slug, name, description, members }));
 };
 
+type TMatchableMember<T> = { user?: { email?: string | null } | null; inviteEmail?: string | null } & T;
+
+// Matching runs once per (GitHub login, org member) pair, so for the reported org it is ~640 teams
+// x 7k memberships x 2.3k members. Deriving each member's comparison fields once per sync instead
+// of once per pair keeps that off the request's critical path; candidate order is preserved, so the
+// same member is still chosen. A member with no usable email could never match, so it is dropped
+// from the candidate list rather than re-tested and rejected on every comparison.
+export const buildGithubMemberMatcher = <T>(members: TMatchableMember<T>[]) => {
+  const emailSplitRegex = new RE2(/[._-]/);
+  const candidates = members.flatMap((member) => {
+    const email = member.user?.email || member.inviteEmail;
+    if (!email) return [];
+    const [rawPrefix, rawDomain] = email.split("@");
+    if (!rawPrefix || !rawDomain) return [];
+    const emailPrefix = rawPrefix.toLowerCase();
+    return [
+      {
+        member,
+        emailPrefix,
+        domainName: rawDomain.toLowerCase().split(".")[0],
+        longestEmailPart: emailPrefix.split(emailSplitRegex).reduce((a, b) => (a.length > b.length ? a : b), "")
+      }
+    ];
+  });
+
+  return (githubLogin: string) => {
+    const githubUsername = githubLogin.toLowerCase();
+    return candidates.find(
+      ({ emailPrefix, domainName, longestEmailPart }) =>
+        emailPrefix === githubUsername ||
+        (githubUsername.endsWith(domainName) &&
+          githubUsername.length > domainName.length &&
+          emailPrefix === githubUsername.slice(0, -domainName.length)) ||
+        (longestEmailPart.length >= 4 && githubUsername.includes(longestEmailPart))
+    )?.member;
+  };
+};
+
 // Type definitions for GitHub API errors
 interface GitHubApiError extends Error {
   status?: number;
@@ -758,6 +796,9 @@ export const githubOrgSyncServiceFactory = ({
       (member) => member.status === "accepted" && member.isActive
     ) as OrgMembershipWithUser[];
 
+    const matchGithubLogin = buildGithubMemberMatcher(activeMembers);
+    const activeMembersById = new Map(activeMembers.map((member) => [member.id, member]));
+
     const startTime = Date.now();
     const syncErrors: string[] = [];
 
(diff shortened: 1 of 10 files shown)


scott-ray-wilson:
@greptile review

scott-ray-wilson:
@greptile review

Thiago-AS (dismissed):

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
```diff
@@ -591,142 +811,101 @@ export const githubOrgSyncServiceFactory = ({
...
       orgId: orgPermission.orgId,
       $in: { name: allGithubTeamNames }
     });
     const existingTeamsMap = groupBy(existingTeamsOnInfisical, (i) => i.name);
 
+    if (existingTeamsOnInfisical.length) {
+      const existingGroupMemberships = await userGroupMembershipDAL.find({
+        $in: { groupId: existingTeamsOnInfisical.map((team) => team.id) }
+      });
+      assertGithubGroupMembersLinked({
```
Thiago-AS, line 1012: Basically, for this to work, users need to have logged in using GitHub SSO. Is that the new intended behavior? It seems like a pretty significant change that could potentially drive customers away from the feature.

Also, we reject social login when SSO is enforced for an organization. If they enforce SSO, users won’t be able to link their GitHub account to Infisical.

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
```diff
@@ -757,73 +936,54 @@ export const githubOrgSyncServiceFactory = ({
...
+          logger.info(
+            { githubLogin: githubMember.login, githubUserId: githubMember.databaseId, userId },
+            "Matched GitHub team member through a verified GitHub login"
+          );
+        } else {
+          unmatchedGithubUsers.add(githubMember.login);
+        }
+      });
+
+      const removedMembershipsForTeam = await groupDAL.transaction(async (tx) => {
```
Thiago-AS: Is there a way we can skip opening this transaction if we have nothing to write? 640 teams means 640 BEGIN/COMMIT pairs plus 640 pool acquisitions, even if the team has nothing to write.

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
```diff
@@ -848,21 +1008,31 @@ export const githubOrgSyncServiceFactory = ({
...
+    }
 
     if (createdTeams.size || updatedTeams.size) {
       // Team membership changes cascade into the group-expanded project identity meters.
       usageMeteringService.emit(orgPermission.orgId, SecretIdentities.key);
       usageMeteringService.emit(orgPermission.orgId, PamIdentities.key);
     }
 
+    if (unmatchedGithubUsers.size) {
+      logger.info(
```
Thiago-AS, line 1215: I feel this should be on syncErrors as well.

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
```diff
@@ -591,142 +811,101 @@ export const githubOrgSyncServiceFactory = ({
...
+          isEmailVerified: true,
+          $in: { userId: [...activeUserIds] }
+        })
+      : [];
+    const matchGithubMember = buildGithubMemberMatcher(githubAliases, activeUserIds);
+    const linkedActiveUserIds = new Set(githubAliases.map((alias) => alias.userId));
+    const activeUsers = activeMembers.flatMap((member) => (member.user ? [member.user] : []));
+
     const startTime = Date.now();
     const syncErrors: string[] = [];
```
Thiago-AS, line 954: This is returned but it's not consumed on the front end. Since it returns as 200 with this field, the user never sees if there were any sync errors along the way.

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
```diff
@@ -37,6 +38,199 @@ import {
...
+        });
+      }
+      team.members.push(...connection.edges.map((edge) => edge.node));
+      team.membersCursor = nextCursor(connection.pageInfo);
+      membersPage += 1;
+    }
+
+    if (team.membersCursor) {
+      throw new BadRequestError({
+        message: `GitHub team '${team.slug}' has more members than can be synced (${GITHUB_MAX_PAGES * GITHUB_TEAM_MEMBERS_PAGE_SIZE} member limit).`
```
Thiago-AS, line 210: GITHUB_MAX_PAGES * GITHUB_TEAM_MEMBERS_PAGE_SIZE, but the first page came from the teams query, so the ceiling is one page short.

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
```diff
@@ -591,142 +811,101 @@ export const githubOrgSyncServiceFactory = ({
...
-
-        if (!userTeamMap.has(username)) {
-          userTeamMap.set(username, []);
-        }
-        userTeamMap.get(username)!.push(teamName);
-      });
+    const githubTeamMembersByName = new Map<string, TGithubTeamMember[]>();
+    githubTeams.forEach((team) => {
+      const teamName = team.name.toLowerCase();
+      githubTeamMembersByName.set(teamName, [...(githubTeamMembersByName.get(teamName) ?? []), ...team.members]);
```
Thiago-AS: Two distinct GitHub teams whose names lowercase to the same string are unioned into one Infisical group. Should this be case sensitive?

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
```diff
@@ -757,73 +936,54 @@ export const githubOrgSyncServiceFactory = ({
...
 
         const membershipsToRemove = currentMemberships.filter((membership) => {
-          const activeMember = activeMembers.find((am) => am.id === membership.orgMembershipId);
-          return activeMember?.user?.id && !expectedUserIds.has(activeMember.user.id);
+          const userId = activeMembersById.get(membership.orgMembershipId)?.user?.id;
+          if (!userId) return false;
+          return !expectedUserIds.has(userId);
         });
 
         if (usersToAdd.length > 0) {
```
Thiago-AS: Just flagging here. No unique constraint on (groupId, userId) for some reason, so:

Each computes a delta against a prior read and inserts the difference. That is correct in isolation and racy in aggregate.

SyncUserGroups runs on every GitHub login and inserts memberships for the teams that login resolved. syncAllTeams reads current memberships and inserts what is missing. A member signing in with GitHub while an admin has Sync Now running is enough: both read a state without the membership, both insert, and the row lands twice. Any per-row listing will show the member twice.

Not blocking, rare case, just be aware to see if it can have a bad side effect other than data quality.

Thiago-AS (changes requested):
I think we should add Audit logs here. GitHub org sync is the outlier. It removes users from groups, group membership cascades into project and secret access, and there is no record of who was removed, from what, or why.

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
Thiago-AS: Already talked through!

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
scott-ray-wilson: this would break existing groups synced so opting to error on multiple matches

backend/src/ee/services/github-org-sync/github-org-sync-service.ts
scott-ray-wilson: will surface externally 👍

Thiago-AS (approved):

PR #8001, opened 2026-09-04
improvement(audit-log): tag stale-claim drops with stream provider

victorvhs017:
## Context

Audit log stream events that exhaust retries are counted on `auditLogStreamDeliveryExhaustedCounter`. Delivery-path drops already carry `audit_log_stream.provider`; stale-claim drops from the sweeper did not, so the same loss showed up under a different series and could not be broken down by provider.

`recoverStaleClaims` now looks up the stream provider before deleting exhausted rows and returns it with each drop. The sweeper emits the same counter attributes as the delivery path (`audit_log_stream.id` plus `audit_log_stream.provider` when the stream still exists). Provider is omitted only when the stream row is already gone, one reason a claim goes stale.

## Steps to verify the change

1. Configure an audit log stream and force a worker crash while rows are `processing` with `attempts` already at `MAX_ATTEMPTS - 1` (or leave them locked past `STALE_CLAIM_THRESHOLD_MS` = 10 minutes).
2. Run the stale-claim sweeper (`sweepStaleClaims`). Confirm the exhausted outbox rows are deleted.
3. Check `auditLogStreamDeliveryExhaustedCounter` for an increment labeled `audit_log_stream.id=<streamId>` and `audit_log_stream.provider=<provider>`.
4. Repeat after deleting the stream first. Confirm the counter still increments by stream id and does not include a provider attribute.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/ee/services/audit-log-stream-outbox/audit-log-stream-outbox-dal.ts
+++ b/backend/src/ee/services/audit-log-stream-outbox/audit-log-stream-outbox-dal.ts
@@ -167,7 +167,7 @@ export const auditLogStreamOutboxDALFactory = (db: TDbClient) => {
   const recoverStaleClaims = async (
     thresholdMs: number,
     maxAttempts: number
-  ): Promise<{ retried: number; dropped: { streamId: string; orgId: string }[] }> => {
+  ): Promise<{ retried: number; dropped: { streamId: string; orgId: string; provider: string | null }[] }> => {
     try {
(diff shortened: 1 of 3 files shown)


victorvhs017:
@greptileai re-review

Thiago-AS (approved):

PR #4429, opened 2025-08-28
Add Github Bulk Team Sync

carlosmonastyrski:
# Description 📣

Add new option for Github Team Sync feature to trigger a bulk sync manually from the Organization Settings Page, so now users can trigger this update without being forced to authenticate again using Github every time.

## Type ✨

- [ ] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Breaking change
- [ ] Documentation

# Tests 🛠️

```sh
# Here's some code block to paste some code snippets
```

---

- [ ] I have read the [contributing guide](https://infisical.com/docs/contributing/getting-started/overview), agreed and acknowledged the [code of conduct](https://infisical.com/docs/contributing/getting-started/code-of-conduct). 📝

PR #8002, opened 2026-09-04
fix(github-org-sync): match GitHub logins by rule strength, and skip ambiguous ones

scott-ray-wilson:
## Context

Stacked on #7998 (base that first). A customer report drove #7998; driving it end to end surfaced this separately.

Bulk sync has to guess which Infisical member each GitHub login belongs to, because unlike the login path it has no OAuth token per user. The heuristic from #4429 had two problems, and both hand one member another member's group membership, and so their project access.

- **No rule precedence.** Rules were tested per candidate, so array order picked the winner. A weak name-fragment hit on an early member beat an exact email match on a later one. Rules are now tried strongest first across all members.
- **The weakest rule matched a fragment anywhere in the login.** A 4+ character email part just had to appear somewhere, so unrelated people inherited teams. It now has to be a whole component of the login.
- **Ambiguity was resolved silently.** A login two members match equally well is now skipped and reported, not guessed.

Measured against a real 2,304-member org:

| GitHub login | Before | After |
| --- | --- | --- |
| `scott-ray-wilson` | `scott@infisical.com` | `scott-ray-wilson@acme-corp.test` (exact email) |
| `0xArshdeep` | `deep@acme-corp.test` | no match |
| `carlosmonastyrski` | `nast@acme-corp.test` | no match |
| `scott-ray-wilson`, with a colliding alias added | `scott@infisical.com` | reported ambiguous, skipped |

`deep` and `nast` are unrelated people with no GitHub presence who were being granted someone else's team.

**This narrows matching**, so a member matched only by an accidental infix loses that group on the next sync. That is the point of the change, and it errs toward removing access rather than granting it. The strongest rule also now ignores separator differences, so `jane.doe@acme.com` matches `janedoe` by email instead of falling through to the fragment rule.

Reported ambiguities populate the `errors` field on the sync response, which was previously declared, returned and never written to.

## Screenshots

N/A

## Steps to verify the change

- [ ] `cd backend && npm run test:unit -- src/ee/services/github-org-sync` passes 18 tests (needs Node 22 for `re2`), covering each rule, precedence, ambiguity, and the infix cases above
- [ ] `cd backend && npx eslint --ext .ts src/ee/services/github-org-sync/ && npx prettier --check src/ee/services/github-org-sync/` is clean
- [ ] `cd backend && NODE_OPTIONS=--max-old-space-size=12288 npm run type:check` reports 0 errors
- [ ] `make lint-docs-branch` is clean
- [ ] With two org members whose emails both reduce to one GitHub username, **Sync Now** skips that user and names both members in the result rather than adding one of them
- [ ] With a member whose email prefix is a substring of an unrelated GitHub username, **Sync Now** no longer adds them to that user's teams

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/github-org-sync/github-org-sync-service.test.ts
+++ b/backend/src/ee/services/github-org-sync/github-org-sync-service.test.ts
@@ -96,42 +96,98 @@ describe("buildGithubMemberMatcher", () => {
     user: email === null ? null : { email },
     inviteEmail
   });
+  const matched = (result: ReturnType<ReturnType<typeof buildGithubMemberMatcher>>) =>
+    result.status === "matched" ? { id: (result.member as { id: string }).id, rule: result.rule } : result;
 
-  test("matches a login equal to the email prefix", () => {
-    const match = buildGithubMemberMatcher([member("a", "jane.doe@acme.com")]);
-    expect(match("Jane.Doe")?.id).toBe("a");
-    expect(match("someone-else")).toBeUndefined();
+  describe("email rule", () => {
+    test("matches a login equal to the email prefix", () => {
+      const match = buildGithubMemberMatcher([member("a", "jane.doe@acme.com")]);
+      expect(matched(match("Jane.Doe"))).toEqual({ id: "a", rule: "email" });
+    });
+
+    test("matches across separator differences between email and login", () => {
+      const match = buildGithubMemberMatcher([member("a", "jane.doe@acme.com")]);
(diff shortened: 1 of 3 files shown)


scott-ray-wilson:
Folded into #7998 as `6afc4fca0e` instead of keeping it stacked. Closing this.

PR #8012, opened 2026-09-05
fix: split query into 3 ones to reduce duplicate alloc by joins

mathnogueira:
## Context

findByFolderId in `secret-v2-bridge-dal.ts` uses a single query with two LEFT JOINs (tags via junction table + metadata). This creates a Cartesian product: a folder with 500 secrets, 3 tags each, and 2 metadata each produces 500 × 3 × 2 = 3,000 rows instead of 500.

The fix is to split that single query into 3 different queries and merging them together using NodeJS.

Benchmark results (500 secrets, 3 tags, 2 metadata each, 20 iterations):

| | Time (avg) | Heap (avg) | Heap (max) |
|---|---|---|---|
| OLD (single JOIN) | 60.8ms | 13.94 MB | 26.53 MB |
| NEW (three queries) | 21.7ms | 4.43 MB | 10.45 MB |
| **Delta** | **64% faster** | **68% less** | **61% less** |

Benchmark results (3000 secrets, 3 tags, 2 metadata each, 20 iterations):

| | Time (avg) | Heap (avg) | Heap (max) |
|---|---|---|---|
| OLD (single JOIN) | 392.8ms | 6.61 MB | 44.61 MB |
| NEW (three queries) | 92.6ms | 3.72 MB | 30.03 MB |
| **Delta** | **76% faster** | **44% less** | **33% less** |

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
+++ b/backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
@@ -414,89 +414,111 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
         userId = undefined;
       }
 
-      const conn = tx || db.replicaNode();
+      const execQueries = async (conn: Knex) => {
+        const secrets = await conn(TableName.SecretV2)
+          .where({ folderId })
+          .where((bd) => {
+            void bd
+              .whereNull(`${TableName.SecretV2}.userId`)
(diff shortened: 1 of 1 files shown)


mathnogueira:
@greptile review again

akhilmhdh (approved):

PR #7748, opened 2026-08-21
improvement(ui): modernize secrets overview and project list

andrewhuhh:
## Context

Expands [SECRETS-542](https://linear.app/infisical/issue/SECRETS-542/migrate-project-overview-table-to-v3) into a broader modernization of Secret Management and organization project lists.

### What changed

- Rebuilt the Secret Management Overview with v3 components, responsive layouts, canonical resource rows, improved empty/loading states, shift-click selection, and cleaner folder/environment actions.
- Added alphabetical secret sorting controls in both directions. Secret Last edited/Created sorting, environment-scoped timestamp selection, and the associated API/pagination changes are isolated in stacked PR #8022.
- Standardized row actions across secrets, folders, imports, rotations, dynamic secrets, honey tokens, and proxied services. Actions remain available on touch devices and keyboard focus, with reduced-motion-safe transitions.
- Expanded secret workflows: multi-secret creation, inline quick-add, multi-entry `.env` paste, a unified import sheet, and deduplicated copy/download export with value-free preflight.
- Modernized secret editing, overrides, rename/delete confirmations, pending-change reconciliation, bulk actions, and reference/dependency views. Save and undo now support Cmd/Ctrl+Enter and Cmd/Ctrl+U.
- Refreshed organization project lists with responsive grid/list layouts and sortable Name, Description, and Created columns. Organization-wide sorting is backed by the API/DAL.
- Extended shared v3 primitives used by these screens, including PageHeader back links, plain SecretInput, link buttons, scrollable AlertDialogs, embedded SelectedActionBars, table overflow handling, and accessible validation/input states. Removed superseded v2 and duplicate overview components.

## Screenshots

CleanShot 2026-08-21 at 15 12 32@2x
CleanShot 2026-08-21 at 15 12 44@2x
CleanShot 2026-08-21 at 19 43 38@2x
CleanShot 2026-08-21 at 19 46 18@2x
CleanShot 2026-08-21 at 19 46 45@2x
CleanShot 2026-08-23 at 12 01 12@2x

## Steps to verify the change

1. Go to **Organization → Projects → any project type**. Verify Mine/All, search, pagination, responsive grid/list layouts, and Name/Description/Created sorting in both directions.
2. Go to **Secret Management → a project → Overview** with multiple environments and resource types. Verify navigation, search/filters, environment selection, sticky columns, expanded rows, shift-click selection, loading states, and empty states.
[...]

PR #8022, opened 2026-09-07
improvement(secrets): add environment-scoped recency sorting

andrewhuhh:
## Context

Implements [SECRETS-500 — Sort secrets by last edited and created](https://linear.app/infisical/issue/SECRETS-500/sort-secrets-by-last-edited-and-created), stacked on #7748 (SECRETS-542).

The base PR keeps the filter UI, alphabetical secret sorting, and project-list Name/Description/Created sorting. This PR adds Last edited and Created ordering for secrets, in either direction. Single-environment views use that environment's timestamps; multi-environment views choose recency ordering from an environment column's menu.

The dashboard API validates the sort environment, ranks only authorized secrets before pagination, and hydrates the selected page. Missing timestamps sort last and equal timestamps use secret names as a deterministic tie-breaker. Other resource types retain alphabetical ordering during secret recency sorting. Removing the active sort environment in a multi-environment view resets sorting to Name ascending.

This is a selective extraction across 12 files. The combined stack restores the exact source tree from before the split. No dependencies, environment variables, Docker configuration, or database migrations change.

## Steps to verify the change

1. In a single-environment Overview, select Last edited and Created in both directions. Check ordering across page boundaries.
2. Select multiple environments and choose recency sorting from one environment column. Verify that environment determines ordering and keys missing there appear last.
3. Remove the active sort environment while multiple environments remain. Verify the table resets to Name ascending.
4. Verify restricted secrets cannot affect page order or pagination. Confirm the dashboard rejects a missing sort environment for multi-environment recency sorting and rejects a sort environment outside the requested set.
5. Confirm alphabetical sorting, filters, environment actions, and project-list sorting still work.
6. Run the focused backend sort tests, `make reviewable-api`, and the serialized frontend validation gate.

Validation: all four focused recency unit tests pass. Both the extracted base and stacked head pass full backend and frontend lint/type checks. The stacked head has the same Git tree as pre-split commit `ae3a37a05b`. Browser verification and screenshots remain pending.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


andrewhuhh:
@greptile review

backend/src/services/secret/secret-types.ts
```diff
@@ -193,6 +193,14 @@ export enum SecretsOrderBy {
   Name = "name" // "key" for secrets but using name for use across resources
 }
 
+export enum DashboardSecretsOrderBy {
+  Name = "name",
+  CreatedAt = "createdAt",
+  UpdatedAt = "updatedAt"
+}
+
```
lobocv: "Dashboard" is a UI concept and i'm not sure it belongs at this level. This is the service that serves the backend API (which the UI relies on) but there are other consumers as well. Is there a place we can put this that is closer to the front end, or if this is the default for the API, name it `defaultSecretsOrderBy`

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
```diff
@@ -308,7 +344,7 @@ const SECRET_NAME_SORT_OPTIONS = [
   }
 ] as const;
 
-const SECRET_SORT_OPTIONS = SECRET_NAME_SORT_OPTIONS;
+const SECRET_SORT_OPTIONS = [...SECRET_NAME_SORT_OPTIONS, ...SECRET_RECENCY_SORT_OPTIONS] as const;
```
lobocv: nit: Why are we defining two separate sorting options lists just to join them together? Why not just have a list of sorting options?

frontend/src/pages/secret-manager/OverviewPage/OverviewPage.tsx
```diff
@@ -571,8 +608,33 @@ const OverviewPageContent = () => {
     [userAvailableEnvs]
   );
 
+  const shouldClearSortEnvironment = Boolean(
+    sortEnvironment &&
+      (visibleEnvs.length === 1 ||
+        !visibleEnvs.some((environment) => environment.slug === sortEnvironment))
+  );
+  const wouldAggregateRecency =
```
lobocv: nit: the naming of this function is confusing. Not sure what it's supposed to do. the "would" part is what's confusing me.

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1181,34 +1185,95 @@ export const secretV2BridgeServiceFactory = ({
...
-    const scannedSecrets = await secretDAL.findByFolderIds({
-      folderIds: folderMappings.map((folderMapping) => folderMapping.folderId),
-      userId,
-      tx: undefined,
-      filters: limit ? { ...filters, limit: limit + 1 } : filters
-    });
-
-    const { items: secrets, isLimitReached } = takeDistinctKeyScanWindow(scannedSecrets, limit, (secret) => secret.key);
+    const timestampOrderBy =
+      filters.orderBy === DashboardSecretsOrderBy.CreatedAt || filters.orderBy === DashboardSecretsOrderBy.UpdatedAt
```
lobocv: related to the other comment, we shouldn't be mentioning "dashboard" at this layer

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1181,34 +1185,95 @@ export const secretV2BridgeServiceFactory = ({
...
+            filters: {
+              ...filters,
+              keys: page.orderedKeys,
+              limit: undefined,
+              offset: undefined,
+              orderBy: DashboardSecretsOrderBy.Name,
+              orderDirection: OrderByDirection.ASC
+            }
+          })
+        : [];
```
lobocv: Something seems wrong here, we're making 3 separate database queries and essentially doing an in memory join. Given that this is a hot-path, I wouldn't recommend this approach. I'm not very familiar with the DB schemas here but is it possible to do this at the database layer?

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1181,34 +1185,95 @@ export const secretV2BridgeServiceFactory = ({
...
+      secrets.sort((left, right) => (keyOrder.get(left.key) ?? 0) - (keyOrder.get(right.key) ?? 0));
+    } else {
+      // findByFolderIds windows on distinct keys, so scan one key past the limit to tell a full window from a truncated one
+      const scannedSecrets = await secretDAL.findByFolderIds({
```
lobocv: I would have expected that the logic to add sorting would have gone inside this DAL query, not a separate branch path. The branching logic in this func is a code smell.

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1263,10 +1328,26 @@ export const secretV2BridgeServiceFactory = ({
...
+      | "actorAuthMethod"
+      | "search"
+      | "tagSlugs"
+      | "orderBy"
+      | "orderDirection"
+      | "limit"
+      | "offset"
+    >,
+    "orderBy"
   > & {
```
lobocv: nit: I wonder what your opinions are on this pattern, and I know you didn't add this. 

I think the overuse of Pick has destroyed the readability in the codebase. This seems crazy that we're taking 12 fields from the TGetSecretsDTO. IMO it would be clearer if we just used the type itself or define a new named type. But these anonymous DTOs are brutal to read

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -1299,11 +1380,24 @@ export const secretV2BridgeServiceFactory = ({
       environment: folder.environment.slug
     }));
 
+    const isRecencySort = params.orderBy && params.orderBy !== DashboardSecretsOrderBy.Name;
```
lobocv: this is brittle. All orderBy that is NOT name automatically is assumed to be a recencySort?
I'm also not a fan of the term "recencySort". It encodes too much. It's not clear to me what that even means. It seems like it's a combo of multiple individual sorting options. At the API level, we should be granular so that we can combine different sorting fields and orders, not pre-define ones, otherwise we're assuming too much about how a customer wants to access data.

Now at the UI level you can do that, because you own the experience of the customer. But at the API level we need to be more flexible

backend/src/services/secret-v2-bridge/secret-v2-bridge-types.ts
```diff
@@ -501,6 +503,10 @@ export type TFindSecretsByFolderIdsFilter = {
   excludeRotatedSecrets?: boolean;
 };
 
+export type TSecretSortCandidate = Pick<TSecretsV2, "id" | "key" | "folderId" | "createdAt" | "updatedAt"> & {
+  tags: { id: string; slug: string }[];
+};
```
lobocv: Do we really need a new type here? I think this is conflating two separate things. You're joining the secret with tags, but then i see in another part of the code you use Omit to strip the tags out.. this is very confusing and makes me think that we're not modelling the code right

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -720,6 +728,131 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
...
+    let { userId } = dto;
+
+    try {
+      if (userId && !uuidValidate(userId)) {
+        userId = undefined;
+      }
+
+      const readDb = tx || db.replicaNode();
+      const matchingSecretsQuery = readDb(TableName.SecretV2)
+        .distinct(`${TableName.SecretV2}.id`)
```
lobocv: Why do we need distinct on ID? Isn't ID unique? Could this be achieved with a join?

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -720,6 +728,131 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
...
+        .whereIn(`${TableName.SecretV2}.folderId`, folderIds)
+        .where((bd) => {
+          if (filters?.search) {
+            const searchPattern = `%${sanitizeSqlLikeString(filters.search)}%`;
+            void bd.whereILike(`${TableName.SecretV2}.key`, searchPattern);
+            if (filters.includeTagsInSearch) {
+              void bd.orWhereILike(`${TableName.SecretTag}.slug`, searchPattern);
```
lobocv: you say include tags but we're searching on the slug? is that the same thing? why not name it includeSlugInSearch?

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -720,6 +728,131 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
...
+            }
+            if (filters.includeMetadataInSearch) {
+              void bd
+                .orWhereILike(`${TableName.ResourceMetadata}.key`, searchPattern)
+                .orWhereILike(`${TableName.ResourceMetadata}.value`, searchPattern);
+            }
```
lobocv: is this a requirement of the PR? I thought we are trying to sort by last edited/created?

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -720,6 +728,131 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
...
+        .leftJoin(
+          TableName.SecretV2JnTag,
+          `${TableName.SecretV2}.id`,
+          `${TableName.SecretV2JnTag}.${TableName.SecretV2}Id`
+        )
+        .leftJoin(
+          TableName.SecretTag,
+          `${TableName.SecretV2JnTag}.${TableName.SecretTag}Id`,
+          `${TableName.SecretTag}.id`
+        )
```
lobocv: Not sure if you know the answer but why are there 2 secret tag tables being joined here? Do we need both?

backend/src/services/secret-v2-bridge/secret-v2-bridge-dal.ts
```diff
@@ -720,6 +728,131 @@ export const secretV2BridgeDALFactory = ({ db, keyStore }: TSecretV2DalArg) => {
     }
   };
 
+  const findSortCandidatesByFolderIds = async (dto: {
```
lobocv: This is a monster query and it makes me very uncomfortable. My gut is telling me that this is not going to be very performant. TBH i don't think we should be doing such complex searches in postgres, or atleast not this way. At the very least we need some integration tests to make sure this query actually works. Its way too large for me to understand

lobocv (dismissed):
Hey @andrewhuhh 
I haven't had the time to go through this all but I read enough to have the opinion that we should rethink this approach. 
The design seems wrong to me and the database query scares me. This is a hot path so I would not take this lightly.

I'm not familiar enough with the schema to provide better direction, maybe @akhilmhdh can help

PR #8025, opened 2026-09-07
feat(app-connections): add EasyDNS as an ACME DNS-01 provider

syangsao:
# Add EasyDNS as an ACME DNS-01 provider

## Summary

Adds [EasyDNS](https://www.easydns.net) as a supported DNS provider for ACME certificate issuance, so self-hosted Infisical instances can issue Let's Encrypt (and other ACME) certificates natively for domains hosted on EasyDNS — no external `acme.sh` sidecar needed.

The implementation follows the existing provider pattern; **DNS Made Easy** is the closest analogue (two-credential Basic-auth style).

## Changes

| Area | File(s) |
|------|---------|
| ACME provider enum | `acme-certificate-authority-enums.ts` — add `EasyDNS = "easydns"` |
| App connection registration | `app-connection-enums.ts`, `app-connection-maps.ts`, `app-connection-fns.ts`, `app-connection-service.ts`, router `index.ts` |
| New app-connection module | `services/app-connection/easydns/` (enum, schema, types, fns, service) |
| New ACME DNS provider | `acme/dns-providers/easydns.ts` — insert/delete TXT records |
| ACME wiring | `acme-certificate-authority-fns.ts` — insert/delete switch cases + connection validation (create & update paths) |
| URLs / env | `integration-list.ts`, `env.ts` — EasyDNS API base URLs + sandbox flag |

## How it works

- **Credentials:** an EasyDNS *App Connection* storing `apiKey` (API token) + `secretKey` (API key), authenticated via HTTP Basic (`token:key`).
- **Add TXT:** `PUT /zones/records/add/{zone}/TXT` with `{ host, rdata }`.
- **Find record:** `GET /zones/records/all/{zone}/search/{host}` — the API returns a substring match, so results are filtered client-side by exact `host` + `rdata`.
- **Delete TXT:** `DELETE /zones/records/{zone}/{id}`.
- **Record naming:** EasyDNS uses full record names (like Cloudflare), so it stays in the default branch of `getAcmeChallengeRecord` and needs no CNAME-resolution skip.

## API notes (verified against the live API)

- The EasyDNS REST API returns **HTML unless** the request sends `Accept: application/json` — the provider always sets that header.
- Duplicate records are reported via a non-2xx response containing `"Record already exists"` rather than a dedicated status code; the insert path treats this as success (idempotent), matching the DNS Made Easy behavior.
- There is no zone-listing endpoint, so credential validation probes a search against the apex host instead.

## Testing

Verified end-to-end against the live EasyDNS API on a real zone:

```
PASS  PUT add TXT — http=201, status=201, id=...
PASS  GET search TXT — http=200, count=1, exact-match=true
PASS  dig propagation — TXT visible via Cloudflare DNS
PASS  DELETE record — http=200, status=200
PASS  record deleted — gone
```

TypeScript typecheck (`tsc --noEmit`) reports no errors in any changed or new file.

## Usage (after merge)

1. Create an **App Connection** of type *EasyDNS* with your API token + key.
2. In the ACME certificate authority configuration, set `dnsProviderConfig.provider = "easydns"` and `hostedZoneId` to the zone name (e.g. `example.com`).


syangsao:
Hi — thanks for reviewing. A few notes that may help:

**Live-tested against the real EasyDNS API.** The provider was verified end-to-end on a production zone (add → search → public-DNS propagation via `dig` → delete), not just unit-level:

```
PASS  PUT add TXT      — http=201, status=201, record id returned
PASS  GET search TXT   — http=200, exact host+rdata match found
PASS  dig propagation  — TXT visible via public DNS in ~9s
PASS  DELETE record    — http=200, status=200
PASS  record deleted   — gone on re-query
```

**API quirks handled** (EasyDNS is unusual among the existing providers):
- Returns **HTML unless** `Accept: application/json` is sent — the provider always sets it.
- The search endpoint (`/zones/records/all/{zone}/search/{host}`) is a *substring* match, so results are filtered client-side by exact `host` + `rdata` before deletion.
- Duplicate records come back as a non-2xx containing `"Record already exists"` (no dedicated status code); insert treats this as idempotent success, matching DNS Made Easy.
- No zone-listing endpoint exists, so credential validation probes an apex search instead.

**Typecheck:** full backend `tsc --noEmit` reports no errors in any added/changed file. Happy to address anything the gated backend test suite or review surfaces.

PR #8029, opened 2026-09-07
improvement(security): derive cookie signing key from KMS root key

victorvhs017:
## Context

Self-hosted installs that never set `COOKIE_SECRET_SIGN_KEY` all used the same hardcoded default from `env.ts`. That meant session cookies for SAML, OIDC, and LDAP flows could be signed with a predictable, shared secret across deployments.

This change removes that default and derives a per-instance cookie signing key from the KMS root key (HKDF, context `infisical-cookie-signing-key-v1`) after `kmsService.startService()` runs. The key is stable across restarts, replicas, and KEK rotations, and does not require a second secret at rest. Operators can still set `COOKIE_SECRET_SIGN_KEY` explicitly (`openssl rand -base64 32`); when set, it takes precedence over the derived key.

`@fastify/cookie` registration moves from `app.ts` to `routes/index.ts` so signing happens only after the KMS root key is loaded. All SSO session routers (SAML, OIDC, LDAP, SSO, identity LDAP) read the key via `getCookieSigningKey()` instead of config directly.

**Note:** Upgrades on instances that relied on the old default will get a new signing key and invalidate in-flight SSO/login sessions. Instances that already set `COOKIE_SECRET_SIGN_KEY` are unaffected.

## Steps to verify the change

1. Start a fresh self-hosted instance **without** `COOKIE_SECRET_SIGN_KEY` and confirm the backend boots and `@fastify/cookie` registers after KMS init.
2. Complete an SSO flow (SAML, OIDC, or LDAP) and confirm session cookies are set and subsequent authenticated requests succeed.
3. Restart the backend and repeat the SSO flow; confirm existing sessions remain valid (derived key is stable).
4. Set `COOKIE_SECRET_SIGN_KEY` to an explicit 32+ byte value, restart, and confirm SSO still works with the env-provided key.
5. Confirm `docs/self-hosting/configuration/envars.mdx` documents the optional env var and per-instance default behavior.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/lib/crypto/cookie-signing-key.ts
+++ b/backend/src/lib/crypto/cookie-signing-key.ts
@@ -1,10 +1,6 @@
 import crypto from "node:crypto";
 
-import { InternalServerError } from "@app/lib/errors";
-
 const COOKIE_SIGNING_KEY_CONTEXT = "infisical-cookie-signing-key-v1";
-
-let cookieSigningKey: string | undefined;
 
 /**
  * Derives the cookie signing key from the KMS root key, which is already generated and persisted
@@ -13,17 +9,3 @@ let cookieSigningKey: string | undefined;
  */
 export const deriveCookieSigningKey = (rootKey: Buffer) =>
   Buffer.from(crypto.hkdfSync("sha256", rootKey, Buffer.alloc(0), COOKIE_SIGNING_KEY_CONTEXT, 32)).toString("base64");
-
-export const setCookieSigningKey = (key: string) => {
-  cookieSigningKey = key;
-};
-
-export const getCookieSigningKey = () => {
-  if (!cookieSigningKey) {
-    throw new InternalServerError({
-      message: "Cookie signing key was requested before the KMS root key was loaded"
-    });
-  }
-
-  return cookieSigningKey;
-};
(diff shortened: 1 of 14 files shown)


backend/src/lib/crypto/cookie-signing-key.ts
```diff
@@ -0,0 +1,29 @@
+import crypto from "node:crypto";
+
+import { InternalServerError } from "@app/lib/errors";
+
+const COOKIE_SIGNING_KEY_CONTEXT = "infisical-cookie-signing-key-v1";
+
+let cookieSigningKey: string | undefined;
```
akhilmhdh: I think we should avoid this pattern. I was just doing a security boundary check, and this could accidentally leak in a heap error. Not critical but feels as we introduce more of this internal state pattern it becomes harder to maintain.

I think we should move this inside the KMS service and use the KMS service as required for cookie secret.

Let me know what you think about that

victorvhs017: I’m good with dropping the module singleton from a code organization standpoint, but I don't see how this pattern increases heap leaks. Do you mean by someone accidentally logging it?

Updating the code!

akhilmhdh (approved):

PR #8035, opened 2026-09-08
feat(pam): restore Oracle Database account support

carlosmonastyrski:
## Context

Oracle Database is available in Infisical PAM again. You can add an Oracle account, launch a session to it, see at a glance whether its stored password still works, and rotate that password on a schedule, either the account changing its own password, or a separate admin account doing it on its behalf.

Because Oracle logs in differently from the other databases, rotation and health checks are handled by the Gateway rather than by Infisical directly, which also means Oracle databases requiring an encrypted connection are supported.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
+++ b/backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
@@ -17,6 +17,8 @@ import { checkAccountAccess, TActorContext } from "../pam/pam-permission";
 import {
   buildGatewayConnectionTest,
   CLOUD_CONNECTION_VALIDATORS,
+  exceedsOraclePasswordLimit,
+  ORACLE_MAX_PASSWORD_LENGTH,
   TestConnectionMode
 } from "../pam-account/pam-account-connection-test";
 import { TPamAccountDALFactory, TPamAccountDetail } from "../pam-account/pam-account-dal";
@@ -215,8 +217,15 @@ export const pamAccountHeartbeatServiceFactory = ({
     const probeCredentials =
       accountType === PamAccountType.SSH ? await mintEphemeralSshCertificate(account, credentials) : credentials;
 
+    if (exceedsOraclePasswordLimit(accountType, credentials)) {
+      return {
+        status: PamHeartbeatStatus.CannotCheck,
+        message: `This account's password is longer than ${ORACLE_MAX_PASSWORD_LENGTH} characters, so this credential was not checked`
+      };
+    }
+
     const test = await buildGatewayConnectionTest(accountType, connectionDetails, probeCredentials, orgId, {
-      allowWindowsAuthSql: true
+      allowNewerGatewayTests: true
     });
     if (!test) {
       return { status: PamHeartbeatStatus.Unknown, message: "This account type cannot be checked yet" };
--- a/backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
+++ b/backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
@@ -35,6 +35,8 @@ import {
   classifyCloudProbeError,
   computeNextHeartbeatAt,
   describeFailure,
+  GATEWAY_MISSING_CHECK_NOTE,
+  gatewayIsMissingCheckSupport,
   HEARTBEAT_SSH_CERT_TTL_SECONDS,
   HEARTBEAT_TIMEOUT_MS,
   isHeartbeatScheduled,
@@ -263,6 +265,9 @@ export const pamAccountHeartbeatServiceFactory = ({
       return { status: PamHeartbeatStatus.CannotCheck, message: "The gateway could not be reached" };
     }
     if (!result.ok) {
+      if (gatewayIsMissingCheckSupport(result.errorMessage)) {
+        return { status: PamHeartbeatStatus.CannotCheck, message: GATEWAY_MISSING_CHECK_NOTE };
+      }
       return {
         status: statusForFailureKind(result.kind),
         message: describeFailure(result.kind, result.errorMessage)
--- a/backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
+++ b/backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
@@ -163,17 +163,15 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
   ): TGeneratedPasswordRequirements => {
     if (accountType !== PamAccountType.OracleDB) return requirements;
     const effective = requirements ?? DEFAULT_PASSWORD_REQUIREMENTS;
+    if (effective.length <= ORACLE_MAX_PASSWORD_LENGTH) return requirements;
+
     const requiredTotal = Object.values(effective.required).reduce((sum, count) => sum + count, 0);
-    if (effective.length <= ORACLE_MAX_PASSWORD_LENGTH && requiredTotal <= ORACLE_MAX_PASSWORD_LENGTH) {
-      return requirements;
+    if (requiredTotal > ORACLE_MAX_PASSWORD_LENGTH) {
+      throw new BadRequestError({
+        message: `This account's template requires at least ${requiredTotal} characters, which is more than the ${ORACLE_MAX_PASSWORD_LENGTH} an Oracle password allows. Lower the character requirements on the template, then rotate again`
+      });
     }
-    return {
-      ...effective,
-      length: Math.min(effective.length, ORACLE_MAX_PASSWORD_LENGTH),
-      required: Object.fromEntries(
-        Object.entries(effective.required).map(([type, count]) => [type, count > 0 ? 1 : 0])
-      ) as typeof effective.required
-    };
+    return { ...effective, length: ORACLE_MAX_PASSWORD_LENGTH };
   };
 
   const getPasswordRequirements = (templateSettings: unknown) =>
--- a/backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
+++ b/backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
@@ -329,16 +329,15 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
     a: Record<string, unknown>,
     b: Record<string, unknown>
   ): boolean => {
-    if (accountType === PamAccountType.WindowsAd) return a.dcAddress === b.dcAddress && a.domain === b.domain;
-    if (isWindowsRotatableType(accountType)) return a.host === b.host;
+    const sameField = (field: string) =>
+      (a[field] as string | undefined)?.toLowerCase() === (b[field] as string | undefined)?.toLowerCase();
+
+    if (accountType === PamAccountType.WindowsAd) return sameField("dcAddress") && sameField("domain");
+    if (isWindowsRotatableType(accountType)) return sameField("host");
     if (accountType === PamAccountType.OracleDB) {
-      return (
-        a.host === b.host &&
-        a.port === b.port &&
-        (a.database as string | undefined)?.toLowerCase() === (b.database as string | undefined)?.toLowerCase()
-      );
+      return sameField("host") && a.port === b.port && sameField("database");
     }
-    return a.host === b.host && a.port === b.port;
+    return sameField("host") && a.port === b.port;
   };
 
   const assertRotatorSameResource = (
@@ -574,12 +573,12 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
     const targetUsername = targetCredentials.username;
     const handler = rotationHandlers[accountType];
 
+    handler.validateTarget({ accountType, authMethod: targetCredentials.authMethod });
+
     const connectionDetails = resolveConnectionDetails(
       accountType,
       await decrypt(projectId, account.encryptedConnectionDetails)
     );
-
-    handler.validateTarget({ accountType, authMethod: targetCredentials.authMethod });
     const gatewayId = account.gatewayId ?? account.templateGatewayId;
     const gatewayPoolId = account.gatewayPoolId ?? account.templateGatewayPoolId;
     const now = new Date();
--- a/backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
+++ b/backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
@@ -5,7 +5,7 @@ import { TGatewayPoolServiceFactory } from "@app/ee/services/gateway-pool/gatewa
 import { TGatewayV2ServiceFactory } from "@app/ee/services/gateway-v2/gateway-v2-service";
 import { TPermissionServiceFactory } from "@app/ee/services/permission/permission-service-types";
 import { ResourcePermissionPamResourceActions } from "@app/ee/services/permission/resource-permission";
-import { generatePassword } from "@app/ee/services/secret-rotation-v2/shared/utils";
+import { DEFAULT_PASSWORD_REQUIREMENTS, generatePassword } from "@app/ee/services/secret-rotation-v2/shared/utils";
 import { KeyStorePrefixes, TKeyStoreFactory } from "@app/keystore/keystore";
 import { BadRequestError, NotFoundError } from "@app/lib/errors";
 import { WinRmRpcEndpoint } from "@app/lib/gateway-v2/winrm-rpc";
@@ -23,6 +23,7 @@ import {
   TActorContext,
   verifyProductMembership
 } from "../pam/pam-permission";
+import { ORACLE_MAX_PASSWORD_LENGTH } from "../pam-account/pam-account-connection-test";
 import { TPamAccountDALFactory, TPamAccountDetail } from "../pam-account/pam-account-dal";
 import { validateConnectionDetails, validateCredentials } from "../pam-account/pam-account-schemas";
 import { computeNextHeartbeatAt, isHeartbeatScheduled } from "../pam-account-heartbeat/pam-heartbeat-fns";
@@ -56,6 +57,8 @@ import {
   winrmPortFromConn,
   winrmTransportFromConn
 } from "./pam-rotation-handlers";
+
+type TGeneratedPasswordRequirements = Parameters<typeof generatePassword>[0];
 
 export const ROTATION_STATUS = { Success: "success", Failed: "failed" } as const;
 
@@ -153,6 +156,25 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
 
   const getRotationConfig = (templateSettings: unknown) =>
     PamTemplateSettingsSchema.safeParse(templateSettings).data?.rotation;
+
+  const capRequirementsForAccountType = (
+    accountType: PamAccountType,
+    requirements: TGeneratedPasswordRequirements
+  ): TGeneratedPasswordRequirements => {
+    if (accountType !== PamAccountType.OracleDB) return requirements;
+    const effective = requirements ?? DEFAULT_PASSWORD_REQUIREMENTS;
+    const requiredTotal = Object.values(effective.required).reduce((sum, count) => sum + count, 0);
+    if (effective.length <= ORACLE_MAX_PASSWORD_LENGTH && requiredTotal <= ORACLE_MAX_PASSWORD_LENGTH) {
+      return requirements;
+    }
+    return {
+      ...effective,
+      length: Math.min(effective.length, ORACLE_MAX_PASSWORD_LENGTH),
+      required: Object.fromEntries(
+        Object.entries(effective.required).map(([type, count]) => [type, count > 0 ? 1 : 0])
+      ) as typeof effective.required
+    };
+  };
 
   const getPasswordRequirements = (templateSettings: unknown) =>
     PamTemplateSettingsSchema.safeParse(templateSettings).data?.passwordRequirements;
@@ -302,7 +324,8 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
   };
 
   // A delegated rotator must sit on the same resource as its target so its credential can reach it: same
-  // host for SQL / local Windows, same domain controller for Windows AD.
+  // host for SQL / local Windows, same domain controller for Windows AD. Oracle users live in a PDB rather
+  // than the instance, so a rotator reaching a different service name cannot see the target user.
   const isSameResource = (
     accountType: PamAccountType,
     a: Record<string, unknown>,
@@ -310,6 +333,13 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
   ): boolean => {
     if (accountType === PamAccountType.WindowsAd) return a.dcAddress === b.dcAddress && a.domain === b.domain;
     if (isWindowsRotatableType(accountType)) return a.host === b.host;
+    if (accountType === PamAccountType.OracleDB) {
+      return (
+        a.host === b.host &&
+        a.port === b.port &&
+        (a.database as string | undefined)?.toLowerCase() === (b.database as string | undefined)?.toLowerCase()
+      );
+    }
     return a.host === b.host && a.port === b.port;
   };
 
@@ -322,11 +352,13 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
     let detail = "on the same resource (host and port) as";
     if (accountType === PamAccountType.WindowsAd) detail = "in the same domain as";
     else if (isWindowsRotatableType(accountType)) detail = "on the same host as";
+    else if (accountType === PamAccountType.OracleDB) detail = "on the same resource (host, port and service name) as";
     throw new BadRequestError({ message: `Rotation account is no longer ${detail} this account` });
   };
 
   // Stable key for the identity an account authenticates as, to detect two PAM objects on the same credential:
-  // domain accounts on domain, local Windows on host, SQL on host+port, all on the bare name. Null if not rotatable.
+  // domain accounts on domain, local Windows on host, SQL on host+port (plus service name for Oracle), all on
+  // the bare name. Null if not rotatable.
   const identityKey = (accountType: PamAccountType, conn: Record<string, unknown>, username: string): string | null => {
     const user = toBareAccountName(username).toLowerCase();
     if (accountType === PamAccountType.WindowsAd) {
@@ -339,7 +371,12 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
     }
     if (isSqlRotatableType(accountType)) {
       const host = (conn.host as string | undefined)?.toLowerCase();
-      return host ? `sql|${host}|${String(conn.port)}|${user}` : null;
+      if (!host) return null;
+      if (accountType === PamAccountType.OracleDB) {
+        const service = (conn.database as string | undefined)?.toLowerCase();
+        return service ? `sql|${host}|${String(conn.port)}|${service}|${user}` : null;
+      }
+      return `sql|${host}|${String(conn.port)}|${user}`;
     }
     return null;
   };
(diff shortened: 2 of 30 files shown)


carlosmonastyrski:
@greptile re-review this PR and update the top Greptile Summary

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
```diff
@@ -49,13 +51,125 @@ type TTestCredentialInput = {
...
+
+const assertOracleCredentialIsUsable = (password: string) => {
+  if (password.length > ORACLE_MAX_PASSWORD_LENGTH) {
+    throw new BadRequestError({
+      message: `Oracle credentials longer than ${ORACLE_MAX_PASSWORD_LENGTH} characters cannot be used to rotate. Shorten the rotation account's password and try again.`
+    });
+  }
+};
+
+const ORACLE_GATEWAY_TOO_OLD =
```
lb-vn: v0.43.130 is already tagged on the cli repo (2026-09-08) and it does not contain sql_handler.go, since Infisical/cli#391 is still open. so anyone sitting on 0.43.130 gets told they are on a good version, upgrades to nothing, and hits this again. we should point at whatever version actually ships #391, or drop the number and say update the gateway to the latest release. same string is copied into two docs pages.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: "Oracle credential rotation"
+sidebarTitle: "Oracle"
+description: "Rotate Oracle account passwords with Infisical PAM."
+---
+
+This page covers what is specific to rotating an [Oracle account](/documentation/platform/pam/accounts/oracledb). For how rotation is configured and scheduled, see the [overview](/documentation/platform/pam/product-reference/credential-rotation/overview).
+
+<Note>
+  Rotating an Oracle account needs a [Gateway](/documentation/platform/gateways/overview) running v0.43.130 or later. Rotation fails on an older Gateway.
```
lb-vn: same as the backend constant: v0.43.130 is already released and does not have the sql rotation handler in it, that lands in whatever version #391 ships in.

docs/documentation/platform/pam/product-reference/credential-health/overview.mdx
```diff
@@ -49,6 +50,10 @@ Checks are spread across the interval rather than fired together, so a template
   Microsoft SQL Server accounts using Windows authentication need a gateway running a version that supports the Windows authentication handshake. An older gateway reports these accounts as **Out of Sync** even when the credential is valid. Update the gateway to get an accurate result.
 </Note>
 
+<Note>
+  Checking an Oracle credential needs a gateway running v0.43.130 or later, and a password of 30 characters or fewer. An Oracle account that doesn't meet both shows as **Unreachable**.
```
lb-vn: same version issue here, v0.43.130 is already out and predates the oracle dialect.

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -215,8 +217,15 @@ export const pamAccountHeartbeatServiceFactory = ({
     const probeCredentials =
       accountType === PamAccountType.SSH ? await mintEphemeralSshCertificate(account, credentials) : credentials;
 
+    if (exceedsOraclePasswordLimit(accountType, credentials)) {
```
lb-vn: connected to the version note: on a gateway that predates the oracle dialect, openSQLTestDB falls into its default branch and returns `unsupported SQL dialect: "oracle"`, doSQLConnectionTest wraps that in connectFailure so kind is transport, and describeFailure passes the raw string through untouched. so the account shows Unreachable with an internal driver message and no hint that the fix is a gateway upgrade. that is the default state for every existing gateway on day one, not an edge case. we should catch it and reuse the same copy ORACLE_GATEWAY_TOO_OLD uses.

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
```diff
@@ -49,13 +51,125 @@ type TTestCredentialInput = {
...
+  dialect: "oracle",
+  username: auth.username,
+  password: auth.password,
+  database: connectionDetails.database,
+  sslEnabled: connectionDetails.sslEnabled,
+  sslRejectUnauthorized: connectionDetails.sslRejectUnauthorized,
+  sslCertificate: connectionDetails.sslCertificate
+});
+
+const verifyOracleViaGateway = async (
```
lb-vn: the windows testCredential wraps every probe in withGatewayRetry precisely so a transient tunnel blip is not read as a verdict, but this one throws on the first transport failure. after a successful apply that lands in the changed-but-could-not-verify branch, which does self-heal on the next run via the pending credential, but it still fails the rotation and fires an alert off one blip. we should wrap it in withGatewayRetry(..., "verify") for parity.

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -313,6 +315,55 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      sslEnabled: z.boolean(),
+      sslRejectUnauthorized: z.boolean(),
+      sslCertificate: optionalTrimmedString
+    }),
+    credentials: z.object({
+      username: z.string().trim().min(1).max(128),
+      password: z
+        .string()
+        .trim()
+        .max(256)
```
lb-vn: we take up to 256 here, but the docs say max 30, the health check refuses to probe anything over 30, and rotation refuses to use a rotator over 30. so an account can be saved in a state where it can never be checked and never rotated, and the only signal is the heartbeat message after the fact. if 256 is deliberate so people can onboard an existing long password, fine, but then we should say so in the account form rather than silently degrading. otherwise cap it at ORACLE_MAX_PASSWORD_LENGTH.

frontend/src/pages/pam/PamTemplatesPage/components/TemplateDetailSheet.tsx
```diff
@@ -582,6 +586,7 @@ const SettingsTab = ({
                       <FieldLabel>{label}</FieldLabel>
                       <Input
                         type="number"
+                        max={name === "settings.pwLength" ? maxPwLength : undefined}
```
lb-vn: `max` on a controlled number input is only a spinner hint. the Controller has no rules and handleSubmit does not run native constraint validation, so typing 50 on an oracle template still submits and comes back as a server error. we should add rules={{ max: { value: maxPwLength, message: ... } }}, and probably a FieldDescription so the 30 cap is visible before you hit save rather than after.

frontend/src/hooks/api/pam/enums.ts
```diff
@@ -30,13 +30,19 @@ export const ROTATABLE_PAM_ACCOUNT_TYPES = [
...
   PamAccountType.MsSQL,
+  PamAccountType.OracleDB,
   PamAccountType.Windows,
   PamAccountType.WindowsAd
 ];
 
 export const isRotatablePamAccountType = (type: PamAccountType | string) =>
   (ROTATABLE_PAM_ACCOUNT_TYPES as string[]).includes(type);
 
+export const ORACLE_MAX_PASSWORD_LENGTH = 30;
```
lb-vn: this 30 is a copy of ORACLE_MAX_PASSWORD_LENGTH in pam-account-schemas.ts, and the 250 below is a copy of PasswordRequirementsSchema's max, with nothing tying either to its source. if one moves these drift silently. minor, but a comment pointing at the backend constants would help.

backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
```diff
@@ -539,12 +574,12 @@ export const pamAccountRotationServiceFactory = (deps: TPamAccountRotationServic
...
     const handler = rotationHandlers[accountType];
 
-    handler.validateTarget({ accountType, authMethod: targetCredentials.authMethod });
-
     const connectionDetails = resolveConnectionDetails(
       accountType,
       await decrypt(projectId, account.encryptedConnectionDetails)
     );
+
+    handler.validateTarget({ accountType, authMethod: targetCredentials.authMethod });
```
lb-vn: why did validateTarget move below resolveConnectionDetails? nothing in it reads connectionDetails, and now we decrypt the blob before finding out the account is not rotatable at all. if that was incidental we should put it back above.

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
lb-vn: v0.43.130 is already tagged on the cli repo (2026-09-08) and it does not contain sql_handler.go, since Infisical/cli#391 is still open. so anyone sitting on 0.43.130 gets told they are on a good version, upgrades to nothing, and hits this again. we should point at whatever version actually ships #391, or drop the number and say update the gateway to the latest release. same string is copied into two docs pages.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
lb-vn: same as the backend constant: v0.43.130 is already released and does not have the sql rotation handler in it, that lands in whatever version #391 ships in.

docs/documentation/platform/pam/product-reference/credential-health/overview.mdx
lb-vn: same version issue here, v0.43.130 is already out and predates the oracle dialect.

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
lb-vn: connected to the version note: on a gateway that predates the oracle dialect, openSQLTestDB falls into its default branch and returns `unsupported SQL dialect: "oracle"`, doSQLConnectionTest wraps that in connectFailure so kind is transport, and describeFailure passes the raw string through untouched. so the account shows Unreachable with an internal driver message and no hint that the fix is a gateway upgrade. that is the default state for every existing gateway on day one, not an edge case. we should catch it and reuse the same copy ORACLE_GATEWAY_TOO_OLD uses.

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
lb-vn: the windows testCredential wraps every probe in withGatewayRetry precisely so a transient tunnel blip is not read as a verdict, but this one throws on the first transport failure. after a successful apply that lands in the changed-but-could-not-verify branch, which does self-heal on the next run via the pending credential, but it still fails the rotation and fires an alert off one blip. we should wrap it in withGatewayRetry(..., "verify") for parity.

backend/src/ee/services/pam-account/pam-account-schemas.ts
lb-vn: we take up to 256 here, but the docs say max 30, the health check refuses to probe anything over 30, and rotation refuses to use a rotator over 30. so an account can be saved in a state where it can never be checked and never rotated, and the only signal is the heartbeat message after the fact. if 256 is deliberate so people can onboard an existing long password, fine, but then we should say so in the account form rather than silently degrading. otherwise cap it at ORACLE_MAX_PASSWORD_LENGTH.

frontend/src/pages/pam/PamTemplatesPage/components/TemplateDetailSheet.tsx
lb-vn: `max` on a controlled number input is only a spinner hint. the Controller has no rules and handleSubmit does not run native constraint validation, so typing 50 on an oracle template still submits and comes back as a server error. we should add rules={{ max: { value: maxPwLength, message: ... } }}, and probably a FieldDescription so the 30 cap is visible before you hit save rather than after.

frontend/src/hooks/api/pam/enums.ts
lb-vn: this 30 is a copy of ORACLE_MAX_PASSWORD_LENGTH in pam-account-schemas.ts, and the 250 below is a copy of PasswordRequirementsSchema's max, with nothing tying either to its source. if one moves these drift silently. minor, but a comment pointing at the backend constants would help.

backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
lb-vn: why did validateTarget move below resolveConnectionDetails? nothing in it reads connectionDetails, and now we decrypt the blob before finding out the account is not rotatable at all. if that was incidental we should put it back above.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+An account that rotates its own password restates the old one, which Oracle requires when a password policy is in force:
+
+```sql
+ALTER USER "<target_username>" IDENTIFIED BY "<generated_password>" REPLACE "<current_password>";
+```
+
+The connection uses the account's existing connection details, including the SSL settings.
+
+The statement matches the stored username exactly, so an account whose username doesn't match the case Oracle stored fails to rotate even though its sessions launch fine. Oracle reports this as a privilege error, not a missing user.
```
lb-vn: tested this against oracle 23.8 free and the last sentence is only half right, which matters because it is the sentence someone will grep for when they are debugging. delegated rotation with a properly privileged rotator gives `ORA-01918: user 'app_user' does not exist`, not a privilege error. it is only self-rotation that reports `ORA-01031: insufficient privileges` (oracle checks the privilege before it tells you the quoted name is missing). we should split it: self-rotation reports a privilege error, delegated reports a missing user.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
```diff
@@ -0,0 +1,57 @@
...
+GRANT CREATE SESSION TO app_user;
+```
+
+<Note>
+  A delegated rotation account has to use the same service name as its target, not just the same host and port. Oracle scopes a user to a single database, so a rotation account pointed at a different service can't see the target user.
+</Note>
+
+## Password length
+
+An Oracle template won't accept a password length above 30 characters, and rotation never generates a longer one.
```
lb-vn: worth recording why this cap exists, because the failure mode without it is nasty and silent. on 23.8 free, `ALTER USER ... IDENTIFIED BY "<31+ chars>"` returns success, but you then cannot authenticate with that password at all. i tried 31, 48, 128 and 129 chars: every one was accepted by the ALTER and every one gave ORA-01017 on the way back in. so with no cap a rotation would apply cleanly, fail verification, and leave the account with neither the old nor the new password working, which is the one state the pending-credential recovery cannot dig out of. the 30 is correct and capRequirementsForAccountType is load-bearing, not belt-and-braces, so it would be good to say so here and in a comment on the constant.

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
lb-vn: v0.43.130 is already tagged on the cli repo (2026-09-08) and it does not contain sql_handler.go, since Infisical/cli#391 is still open. so anyone sitting on 0.43.130 gets told they are on a good version, upgrades to nothing, and hits this again. we should point at whatever version actually ships #391, or drop the number and say update the gateway to the latest release. same string is copied into two docs pages.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
lb-vn: same as the backend constant: v0.43.130 is already released and does not have the sql rotation handler in it, that lands in whatever version #391 ships in.

docs/documentation/platform/pam/product-reference/credential-health/overview.mdx
lb-vn: same version issue here, v0.43.130 is already out and predates the oracle dialect.

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
lb-vn: connected to the version note: on a gateway that predates the oracle dialect, openSQLTestDB falls into its default branch and returns `unsupported SQL dialect: "oracle"`, doSQLConnectionTest wraps that in connectFailure so kind is transport, and describeFailure passes the raw string through untouched. so the account shows Unreachable with an internal driver message and no hint that the fix is a gateway upgrade. that is the default state for every existing gateway on day one, not an edge case. we should catch it and reuse the same copy ORACLE_GATEWAY_TOO_OLD uses.

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
lb-vn: the windows testCredential wraps every probe in withGatewayRetry precisely so a transient tunnel blip is not read as a verdict, but this one throws on the first transport failure. after a successful apply that lands in the changed-but-could-not-verify branch, which does self-heal on the next run via the pending credential, but it still fails the rotation and fires an alert off one blip. we should wrap it in withGatewayRetry(..., "verify") for parity.

backend/src/ee/services/pam-account/pam-account-schemas.ts
lb-vn: we take up to 256 here, but the docs say max 30, the health check refuses to probe anything over 30, and rotation refuses to use a rotator over 30. so an account can be saved in a state where it can never be checked and never rotated, and the only signal is the heartbeat message after the fact. if 256 is deliberate so people can onboard an existing long password, fine, but then we should say so in the account form rather than silently degrading. otherwise cap it at ORACLE_MAX_PASSWORD_LENGTH.

frontend/src/pages/pam/PamTemplatesPage/components/TemplateDetailSheet.tsx
lb-vn: `max` on a controlled number input is only a spinner hint. the Controller has no rules and handleSubmit does not run native constraint validation, so typing 50 on an oracle template still submits and comes back as a server error. we should add rules={{ max: { value: maxPwLength, message: ... } }}, and probably a FieldDescription so the 30 cap is visible before you hit save rather than after.

frontend/src/hooks/api/pam/enums.ts
lb-vn: this 30 is a copy of ORACLE_MAX_PASSWORD_LENGTH in pam-account-schemas.ts, and the 250 below is a copy of PasswordRequirementsSchema's max, with nothing tying either to its source. if one moves these drift silently. minor, but a comment pointing at the backend constants would help.

backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
lb-vn: why did validateTarget move below resolveConnectionDetails? nothing in it reads connectionDetails, and now we decrypt the blob before finding out the account is not rotatable at all. if that was incidental we should put it back above.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
lb-vn: tested this against oracle 23.8 free and the last sentence is only half right, which matters because it is the sentence someone will grep for when they are debugging. delegated rotation with a properly privileged rotator gives `ORA-01918: user 'app_user' does not exist`, not a privilege error. it is only self-rotation that reports `ORA-01031: insufficient privileges` (oracle checks the privilege before it tells you the quoted name is missing). we should split it: self-rotation reports a privilege error, delegated reports a missing user.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
lb-vn: worth recording why this cap exists, because the failure mode without it is nasty and silent. on 23.8 free, `ALTER USER ... IDENTIFIED BY "<31+ chars>"` returns success, but you then cannot authenticate with that password at all. i tried 31, 48, 128 and 129 chars: every one was accepted by the ALTER and every one gave ORA-01017 on the way back in. so with no cap a rotation would apply cleanly, fail verification, and leave the account with neither the old nor the new password working, which is the one state the pending-credential recovery cannot dig out of. the 30 is correct and capRequirementsForAccountType is load-bearing, not belt-and-braces, so it would be good to say so here and in a comment on the constant.

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -215,8 +217,15 @@ export const pamAccountHeartbeatServiceFactory = ({
...
+      return {
+        status: PamHeartbeatStatus.CannotCheck,
+        message: `This account's password is longer than ${ORACLE_MAX_PASSWORD_LENGTH} characters, so this credential was not checked`
+      };
+    }
+
     const test = await buildGatewayConnectionTest(accountType, connectionDetails, probeCredentials, orgId, {
-      allowWindowsAuthSql: true
+      allowNewerGatewayTests: true
```
lb-vn: verified end to end against a real oracle through a real gateway, and the consequence of the classification bug i left on Infisical/cli#391 lands here. i created an oracle account with a wrong service name; the health check returned `invalid-credentials` with `ORA-12514: TNS:listener does not currently know of service requested in connect descriptor`.

because stopsSchedule halts on InvalidCredentials, a closed PDB or a service-name typo permanently stops the heartbeat and reports the credential as wrong. the gateway side is the right place to fix it, but worth knowing the blast radius is here.

for the record, everything else on this path checked out against oracle 23.8: valid credential to healthy, wrong password to invalid-credentials quoting ORA-01017, and a 40-char password to cannot-check with the new over-30 message.

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
lb-vn: v0.43.130 is already released and doesn't have this feature in it. we should point at whatever version the cli PR actually ships in right before we merge and deploy this

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
lb-vn: same here, v0.43.130 is already out and doesn't have oracle rotation.

docs/documentation/platform/pam/product-reference/credential-health/overview.mdx
lb-vn: same here, v0.43.130 is already out and doesn't have oracle support.

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
lb-vn, line 222: IGNORE PLACEMENT

on a gateway that's too old the health check shows `unsupported SQL dialect: "oracle"` with no hint to upgrade. every existing gateway does this until it's updated. we should show the same "update the gateway" message rotation uses.

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
lb-vn, line 101: the windows verify path retries with withGatewayRetry, this one doesn't. Should we swap it over?

backend/src/ee/services/pam-account/pam-account-schemas.ts
lb-vn, line 344: we accept 256 here but rotation and health checks both refuse anything over 30. is 256 deliberate so people can onboard an existing password? if not we should cap it at 30.

frontend/src/pages/pam/PamTemplatesPage/components/TemplateDetailSheet.tsx
lb-vn, line 591: max on a number input doesn't block submitting, so you still get a server error. we should add a validation rule and show the limit in the field.

frontend/src/hooks/api/pam/enums.ts
lb-vn, line 43: this duplicates ORACLE_MAX_PASSWORD_LENGTH from the backend. worth a comment pointing at it so they don't drift.

backend/src/ee/services/pam-account-rotation/pam-account-rotation-service.ts
lb-vn: why did this move below resolveConnectionDetails?

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
lb-vn: Claude:

tested this and it's the other way round for delegated rotation, you get ORA-01918 user does not exist. the privilege error only happens on self-rotation. we should mention both.

docs/documentation/platform/pam/product-reference/credential-rotation/oracledb.mdx
lb-vn, line 60: Claude:

tested this and oracle accepts a longer password but then won't let you log in with it. so the cap is preventing a dead account, not just a rejected statement. worth saying that here so nobody relaxes it later.

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
lb-vn, line 230: IGNORE PLACEMENT

Claude found this during tests: an oracle account with a wrong service name comes back as Out of Sync quoting ORA-12514. that stops the heartbeat schedule, so a closed PDB or a typo permanently stops checks and blames the password. the fix belongs in the gateway (commented on the cli PR) but the damage shows up here.

lb-vn (changes requested):

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
carlosmonastyrski: Changed it, but will update it before releasing to the next available tag 👍

backend/src/ee/services/pam-account/pam-account-schemas.ts
carlosmonastyrski: Yeah, it's deliberate so people can onboard an existing password

backend/src/ee/services/pam-account-rotation/pam-rotation-handlers.ts
```diff
@@ -49,13 +51,124 @@ type TTestCredentialInput = {
...
+const oracleGatewayRequest = (
+  connectionDetails: TPamSqlConnectionDetails,
+  auth: { username: string; password: string }
+) => ({
+  dialect: "oracle",
+  username: auth.username,
+  password: auth.password,
+  database: connectionDetails.database,
+  sslEnabled: connectionDetails.sslEnabled,
+  sslRejectUnauthorized: connectionDetails.sslRejectUnauthorized,
```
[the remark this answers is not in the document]
carlosmonastyrski: Pre-existing, will be addressed in a separate PR (fixing the other types as well)

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -35,6 +35,8 @@ import {
   classifyCloudProbeError,
   computeNextHeartbeatAt,
```
lb-vn, line 36: IGNORE PALCEMENT

I couldn't get this to work with **sqlplus 23.3**, could you test this version yourself? It's the latest homebrew one

lb-vn (changes requested):

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -743,7 +743,8 @@ export const pamSessionServiceFactory = ({
         (account.accountType === PamAccountType.Postgres ||
```
lb-vn, line 743: IGNORE PLACEMENT

session logs mark a failed oracle statement as OK. i ran `SELECT version()` through a session, oracle rejected it with ORA-00904, and the log entry says "SELECT version() OK".

extractResponseOutcome in the cli's query_logger.go returns "OK" whenever it can't parse the error block, so anything it doesn't understand reads as a success in the audit trail. we should default to unknown instead of OK.

lb-vn: image

lb-vn: also maybe look into making these have a similar format to stuff like postgres, but this is more optional

lb-vn (changes requested):

lb-vn (approved):

PR #8043, opened 2026-09-08
fix(pam): atomic recording secret mint

bernie-g:
## Context

Recording secrets were minted with an unguarded read-decide-write on `encryptedSessionKey`, so two concurrent credential fetches for one session both minted and the later write won, leaving the earlier caller with a session key and upload token the row no longer matched. That gateway's chunk uploads then 400ed forever, and anything it had already uploaded became undecryptable at playback.

Fixes [PAM-463](https://linear.app/infisical/issue/PAM-463/investigate-session-recording-endpoint-error-spike), which accounted for 99.75% of Infisical Cloud US platform errors. The gateway-side half is in Infisical/cli#392.

## Steps to verify the change

Race concurrent credential fetches at one live, never-minted session and check that every response carries the same session key, exactly one carries an upload token, and the stored hash is `sha256` of that token.

Six concurrent fetches through the real stack, before and after:

| | Before | After |
|---|---|---|
| Runs | 6 | 8 |
| Session keys diverged | 5 of 6 | 0 |
| Distinct keys issued for one session | up to 6 | always 1 |
| Verdict | 5 fail | 8/8 pass |

Also verified the guarded `UPDATE` against Postgres directly: eight concurrent claims on one row yield exactly one winner over five runs, where the unguarded version yields eight winners.

The second commit only renames error variants, so the four `verifyGatewayUploadToken` branches are distinguishable in logs. The client-facing message is deliberately unchanged, which keeps the response shape and the router's `PAM_SESSION_UPLOAD_TOKEN_INVALID` audit trigger working. After deploy, group chunk-route errors by `errorName like /PamUploadToken/`: all `PamUploadTokenHashMismatch` confirms the race.

## Type

- [x] Fix

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
+++ b/backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
@@ -73,9 +73,8 @@ export const decryptSessionKey = async ({
   return decrypted.subarray(32, 32 + SESSION_KEY_LENGTH);
 };
 
-// Every rejection keeps the same client-facing message; only the error name differs, so the
-// branch is recoverable from logs without widening what the caller learns. A malformed token and
-// a token that simply does not match point at completely different causes.
+// The message must stay identical across these: the chunk router substring-matches it to write
+// the upload-token audit event.
 export const PamUploadTokenRejection = {
   Missing: "PamUploadTokenMissing",
   MalformedLength: "PamUploadTokenMalformedLength",
@@ -96,8 +95,7 @@ export const verifyGatewayUploadToken = (
     });
   }
 
-  // Buffer.from drops invalid base64 characters rather than throwing, so a bad encoding arrives
-  // here as a short decode, never as an exception.
+  // Buffer.from drops invalid base64 rather than throwing, so a bad encoding lands here short.
   const presentedBuf = Buffer.from(presentedTokenBase64, "base64");
   if (presentedBuf.length !== UPLOAD_TOKEN_LENGTH) {
     throw new BadRequestError({
--- a/backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
+++ b/backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
@@ -73,15 +73,15 @@ export const decryptSessionKey = async ({
   return decrypted.subarray(32, 32 + SESSION_KEY_LENGTH);
 };
 
-// The message must stay identical across these: the chunk router substring-matches it to write
-// the upload-token audit event.
-export const PamUploadTokenRejection = {
+const PamUploadTokenRejection = {
   Missing: "PamUploadTokenMissing",
   MalformedLength: "PamUploadTokenMalformedLength",
   StoredHashMalformed: "PamUploadTokenStoredHashMalformed",
   HashMismatch: "PamUploadTokenHashMismatch"
 } as const;
 
+// The chunk router substring-matches this to write the upload-token audit event, and the gateway
+// treats HashMismatch as the one failure worth giving up on, so both are a contract.
 const UPLOAD_TOKEN_INVALID_MESSAGE = "Invalid upload token";
 
 export const verifyGatewayUploadToken = (
--- a/backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
+++ b/backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
@@ -73,30 +73,51 @@ export const decryptSessionKey = async ({
   return decrypted.subarray(32, 32 + SESSION_KEY_LENGTH);
 };
 
+// Every rejection keeps the same client-facing message; only the error name differs, so the
+// branch is recoverable from logs without widening what the caller learns. A malformed token and
+// a token that simply does not match point at completely different causes (PAM-463).
+export const PamUploadTokenRejection = {
+  Missing: "PamUploadTokenMissing",
+  MalformedLength: "PamUploadTokenMalformedLength",
+  StoredHashMalformed: "PamUploadTokenStoredHashMalformed",
+  HashMismatch: "PamUploadTokenHashMismatch"
+} as const;
+
+const UPLOAD_TOKEN_INVALID_MESSAGE = "Invalid upload token";
+
 export const verifyGatewayUploadToken = (
   presentedTokenBase64: string | undefined | null,
   storedTokenHash: Buffer | null | undefined
 ) => {
   if (!presentedTokenBase64 || !storedTokenHash) {
     throw new BadRequestError({
+      name: PamUploadTokenRejection.Missing,
       message: "Gateway upload token missing or session not configured for chunked uploads"
     });
   }
-  let presentedBuf: Buffer;
-  try {
-    presentedBuf = Buffer.from(presentedTokenBase64, "base64");
-  } catch {
-    throw new BadRequestError({ message: "Invalid upload token" });
-  }
+
+  // Buffer.from drops invalid base64 characters rather than throwing, so a bad encoding arrives
+  // here as a short decode, never as an exception.
+  const presentedBuf = Buffer.from(presentedTokenBase64, "base64");
   if (presentedBuf.length !== UPLOAD_TOKEN_LENGTH) {
-    throw new BadRequestError({ message: "Invalid upload token" });
+    throw new BadRequestError({
+      name: PamUploadTokenRejection.MalformedLength,
+      message: UPLOAD_TOKEN_INVALID_MESSAGE
+    });
   }
+
   const presentedHash = crypto.nativeCrypto.createHash("sha256").update(presentedBuf).digest();
   if (presentedHash.length !== storedTokenHash.length) {
-    throw new BadRequestError({ message: "Invalid upload token" });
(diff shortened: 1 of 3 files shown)


backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -312,32 +313,56 @@ export const pamSessionServiceFactory = ({
...
+        recording = {
+          sessionKey: secrets.sessionKey.toString("base64"),
+          uploadToken: secrets.uploadToken.toString("base64"),
+          storageBackend: resolvedBackend,
+          projectId: session.projectId,
+          sessionId
+        };
+      } else {
+        // Another fetch claimed it first; adopt its secrets rather than our own discarded ones.
+        const claimedSession = await pamSessionDAL.findById(sessionId);
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -312,32 +313,56 @@ export const pamSessionServiceFactory = ({
...
+        storedSessionKey = claimedSession?.encryptedSessionKey ?? null;
+        if (!storedSessionKey) {
+          throw new InternalServerError({
+            message: `Recording secrets for session ${sessionId} are missing after a lost mint`
+          });
+        }
+      }
```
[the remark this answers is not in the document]
bernie-g: Skipping tests here to keep this PR small.

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -312,32 +313,56 @@ export const pamSessionServiceFactory = ({
...
+          sessionKey: secrets.sessionKey.toString("base64"),
+          uploadToken: secrets.uploadToken.toString("base64"),
+          storageBackend: resolvedBackend,
+          projectId: session.projectId,
+          sessionId
+        };
+      } else {
+        // Another fetch claimed it first; adopt its secrets rather than our own discarded ones.
+        const claimedSession = await pamSessionDAL.findById(sessionId);
+        storedSessionKey = claimedSession?.encryptedSessionKey ?? null;
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -312,32 +313,46 @@ export const pamSessionServiceFactory = ({
...
+      if (claimed.encryptedSessionKey.equals(secrets.encryptedSessionKey)) {
+        recording = {
+          sessionKey: secrets.sessionKey.toString("base64"),
+          uploadToken: secrets.uploadToken.toString("base64"),
+          storageBackend: resolvedBackend,
+          projectId: session.projectId,
+          sessionId
+        };
+      } else {
+        storedSessionKey = claimed.encryptedSessionKey;
```
lb-vn: the loser lands here with the right key but no upload token, and the comment just below explains that away with "the gateway restores its own from disk". that's true for the restart case the branch was originally written for, but it's false here - a loser in a first-mint race has nothing on disk yet. we only keep the hash so we genuinely can't reissue the winner's token.

what actually saves us today is the gateway coalescing: credentials.go keeps an already-cached token when the response's token is empty. but that only works if the winner's response landed first, and it's the singleflight change that's missing from Infisical/cli#392. so the correctness of this path is resting on gateway-side behaviour that isn't merged. at minimum the comment should say so instead of implying the empty token is always recoverable.

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -312,32 +313,46 @@ export const pamSessionServiceFactory = ({
...
 
-      recording = {
-        sessionKey: secrets.sessionKey.toString("base64"),
-        uploadToken: secrets.uploadToken.toString("base64"),
-        storageBackend: resolvedBackend,
-        projectId: session.projectId,
-        sessionId
-      };
-    } else {
+      if (claimed.encryptedSessionKey.equals(secrets.encryptedSessionKey)) {
```
lb-vn: deciding the winner by comparing kms ciphertexts only works because the cipher is randomized. that's true today, but it's an invisible dependency - a deterministic cipher would make both callers believe they won and hand out two tokens for one stored hash. returning `gatewayUploadTokenHash` and comparing it to `secrets.uploadTokenHash` is the same cost and doesn't rely on it.

backend/src/ee/services/pam-session/pam-session-service.ts
```diff
@@ -312,32 +313,46 @@ export const pamSessionServiceFactory = ({
...
         kmsService
       });
 
-      await pamSessionDAL.updateById(sessionId, {
-        encryptedSessionKey: secrets.encryptedSessionKey,
-        gatewayUploadTokenHash: secrets.uploadTokenHash
-      });
+      // A gateway fetches credentials per connection, so two can mint at once; without a single
+      // claim the loser keeps a key the row no longer holds and its uploads fail forever.
+      const claimed = await pamSessionDAL.claimRecordingSecrets(
```
lb-vn: we should add a test here. `backend/src/ee/services/pam-*` is full of `*.test.ts` and this is branchy now - winner, loser, missing row - which a stubbed `claimRecordingSecrets` covers in a few lines. the six-concurrent-fetch verification in the description is the convincing part of this pr and none of it is committed, so the next person to touch the dal method has nothing catching them.

backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
```diff
@@ -73,30 +73,49 @@ export const decryptSessionKey = async ({
   return decrypted.subarray(32, 32 + SESSION_KEY_LENGTH);
 };
 
+// The message must stay identical across these: the chunk router substring-matches it to write
```
lb-vn: this isn't quite right as written - `Missing` deliberately keeps a different message, so it doesn't contain "Invalid upload token" and pam-session-chunk-router.ts:81 never writes the `PAM_SESSION_UPLOAD_TOKEN_INVALID` audit event for it. that's the branch a gateway holding an empty token hits, which is exactly the failure mode the other half of this fix can produce. we should scope the comment to the three that share `UPLOAD_TOKEN_INVALID_MESSAGE`, and separately decide whether `Missing` ought to be audited too.

backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
```diff
@@ -73,30 +73,49 @@ export const decryptSessionKey = async ({
...
+// the upload-token audit event.
+export const PamUploadTokenRejection = {
```
lb-vn: three small ones. `PamUploadTokenRejection` is exported but only used in this file. the response's `error` field changes from `BadRequest` to `PamUploadToken*` for these branches - gateway-only route so almost certainly fine, just flagging it's wire-visible. and on the "all HashMismatch confirms the race" plan in the description: `Buffer.from` silently drops invalid base64, so a malformed token that still decodes to 32 bytes lands in `HashMismatch` too. probably a small share, but the signal isn't clean.

backend/src/ee/services/pam-session/pam-session-service.ts
lb-vn: the loser gets no upload token, and the comment below says the gateway restores its own from disk - but a loser on first mint has nothing on disk yet. it only works because the gateway keeps its cached token when the response is empty, and that's the singleflight change missing from the cli pr.

backend/src/ee/services/pam-session/pam-session-service.ts
lb-vn: comparing kms ciphertexts to pick the winner only works because the cipher is randomized. comparing the token hash instead does the same thing without depending on that.

backend/src/ee/services/pam-session/pam-session-service.ts
lb-vn: can we get a test on this? winner, loser and missing row are easy to cover with a stubbed dal, and the concurrency verification in the description isn't committed anywhere.

backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
lb-vn: This comment does not seem fully accurate because the Missing error on line 94 seems to use a different message. Is this intentional?

backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
lb-vn: nit: PamUploadTokenRejection is exported but only used here

lb-vn (changes requested):

backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
bernie-g: Fixed

backend/src/ee/services/pam-session-recording/pam-recording-secrets.ts
bernie-g: Fixed

lb-vn (approved):

PR #8044, opened 2026-09-08
feat(pam): snowflake account

lb-vn:
## Context

New Snowflake account type for PAM with 3 auth methods. Supports web access and CLI access (proxy).

## Steps to verify the change

Read the added docs for testing different auth types. I recommend testing with snowsql (cli client) and DBeaver (ui client). I tested with both personally.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
+++ b/backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
@@ -21,6 +21,7 @@ import {
 } from "../pam-account/pam-account-connection-test";
 import { TPamAccountDALFactory, TPamAccountDetail } from "../pam-account/pam-account-dal";
 import {
+  collectCredentialSecrets,
   parseInternalMetadata,
   validateConnectionDetails,
   validateCredentials
@@ -145,10 +146,7 @@ export const pamAccountHeartbeatServiceFactory = ({
       accountType,
       await decrypt(projectId, account.encryptedCredentials)
     ) as Record<string, unknown>;
-    for (const field of ["password", "privateKey", "serviceAccountKeyJson", "clientSecret", "serviceAccountToken"]) {
-      const value = credentials[field];
-      if (typeof value === "string" && value) usedSecrets.push(value);
-    }
+    usedSecrets.push(...collectCredentialSecrets(accountType, credentials));
 
     const validateCloud = CLOUD_CONNECTION_VALIDATORS[accountType];
     if (validateCloud) {
--- a/backend/src/ee/services/pam-account/pam-account-connection-test.ts
+++ b/backend/src/ee/services/pam-account/pam-account-connection-test.ts
@@ -17,6 +17,7 @@ export enum TestConnectionMode {
   LDAP = "ldap",
   Kubernetes = "kubernetes",
   SSH = "ssh",
+  Snowflake = "snowflake",
   Tcp = "tcp"
 }
 
@@ -79,6 +80,20 @@ export type TestConnectionRequest =
       password?: string;
       privateKey?: string;
       certificate?: string;
+    }
+  | {
+      mode: TestConnectionMode.Snowflake;
+      account: string;
+      authMethod: string;
+      username: string;
+      password?: string;
+      token?: string;
+      privateKey?: string;
+      privateKeyPassphrase?: string;
+      warehouse?: string;
+      database?: string;
+      schema?: string;
+      role?: string;
     }
   | { mode: TestConnectionMode.Tcp };
 
@@ -297,6 +312,42 @@ export const buildGatewayConnectionTest = async (
           password: c.password,
           privateKey: c.privateKey,
           certificate: c.certificate
+        }
+      };
+    }
+    case PamAccountType.Snowflake: {
+      const cd = connectionDetails as {
+        account: string;
+        warehouse?: string;
+        database?: string;
+        schema?: string;
+        role?: string;
+      };
+      const c = creds as {
+        authMethod: string;
+        username: string;
+        password?: string;
+        token?: string;
+        privateKey?: string;
+        privateKeyPassphrase?: string;
+      } | null;
+      if (!c) return tcp(host, port);
+      return {
+        host,
+        port,
+        request: {
+          mode: TestConnectionMode.Snowflake,
+          account: cd.account,
+          authMethod: c.authMethod,
+          username: c.username,
+          password: c.password,
+          token: c.token,
+          privateKey: c.privateKey,
+          privateKeyPassphrase: c.privateKeyPassphrase,
+          warehouse: cd.warehouse,
+          database: cd.database,
+          schema: cd.schema,
+          role: cd.role
         }
       };
     }
--- a/backend/src/ee/services/pam-web-access/snowflake/pam-snowflake-connection-controller.ts
+++ b/backend/src/ee/services/pam-web-access/snowflake/pam-snowflake-connection-controller.ts
@@ -1,0 +1,245 @@
+import snowflake from "snowflake-sdk";
+
+import { logger } from "@app/lib/logger";
+import { executeSnowflakeSql } from "@app/services/app-connection/snowflake";
+
+import { type ControllerParams } from "../pam-data-explorer-session-handler";
+import {
+  DataExplorerClientMessageType,
+  DataExplorerServerMessageType,
+  type TConnectionController,
+  type TTabScopedMessage
+} from "../pam-data-explorer-ws-types";
+import { extractCommand, nextTransactionState, splitSnowflakeStatements } from "./pam-snowflake-data-explorer-fns";
+import {
+  COLUMNS_QUERY,
+  showImportedKeysQuery,
+  showPrimaryKeysQuery,
+  TColumnRow,
+  toColumns,
+  toForeignKeys,
+  toPrimaryKeys
+} from "./pam-snowflake-data-explorer-metadata";
+import { snowflakeAccountOf } from "./pam-snowflake-metadata";
+import { connectThroughRelay } from "./pam-snowflake-relay-client";
+
+const MAX_ROWS = 1000;
+
+const readRows = (
+  statement: snowflake.RowStatement
+): Promise<{ rows: Record<string, unknown>[]; truncated: boolean }> =>
+  new Promise((resolve, reject) => {
+    const rows: Record<string, unknown>[] = [];
+    let truncated = false;
+    const stream = statement.streamRows();
+
+    stream.on("data", (row: Record<string, unknown>) => {
+      if (rows.length < MAX_ROWS) {
+        rows.push(row);
+        return;
+      }
+      truncated = true;
+      stream.destroy();
+      resolve({ rows, truncated });
+    });
+    stream.on("error", (err) => {
+      if (truncated) resolve({ rows, truncated });
+      else reject(err);
+    });
+    stream.on("end", () => resolve({ rows, truncated }));
+  });
+
+export const createSnowflakeConnectionController = async (params: ControllerParams): Promise<TConnectionController> => {
+  const { sessionId, connectionId, sendResponse, onUnexpectedTermination } = params;
+
+  const client = await connectThroughRelay(params.relayPort, snowflakeAccountOf(params));
+
+  let isInTransaction = false;
+  let disposing = false;
+  let runningStatement: snowflake.RowStatement | null = null;
+
+  const execute = (sqlText: string): Promise<snowflake.RowStatement> =>
+    new Promise((resolve, reject) => {
+      runningStatement = client.execute({
+        sqlText,
+        streamResult: true,
+        complete: (err, statement) => {
+          if (err) reject(err);
+          else resolve(statement as snowflake.RowStatement);
+        }
+      }) as snowflake.RowStatement;
+    });
+
+  const cancelRunningStatement = () => {
+    const statement = runningStatement;
+    if (!statement) return;
+    statement.cancel((err) => {
+      if (err) {
+        logger.debug(err, `Failed to cancel Snowflake query [sessionId=${sessionId}] [connectionId=${connectionId}]`);
+      }
+    });
+  };
+
+  const sendQueryError = async (id: string, err: unknown) => {
+    // The tunnel is gone once the session is terminated, so the tab is closed rather than left waiting
+    if (!disposing && !client.isUp()) {
+      disposing = true;
+      onUnexpectedTermination("Session ended");
+      return;
+    }
+
+    // Snowflake leaves a failed statement's transaction open, holding its locks until the session ends
+    if (isInTransaction) {
+      try {
+        await executeSnowflakeSql(client, "ROLLBACK");
+      } catch {
+        // nothing to roll back
+      }
+      isInTransaction = false;
+    }
+
+    sendResponse({
+      type: DataExplorerServerMessageType.Error,
+      id,
+      connectionId,
+      transactionOpen: isInTransaction,
+      error: (err as Error)?.message ?? "Query execution failed",
+      detail: (err as { sqlState?: string })?.sqlState
+    });
+  };
+
+  const runTableDetail = async (schema: string, table: string) => {
+    const columns = await executeSnowflakeSql<TColumnRow>(client, COLUMNS_QUERY, [schema, table]);
+    if (columns.length === 0) return null;
+
+    const primaryKeyRows = await executeSnowflakeSql(client, showPrimaryKeysQuery(schema, table));
+    const importedKeyRows = await executeSnowflakeSql(client, showImportedKeysQuery(schema, table));
+
+    return {
+      columns: toColumns(columns),
+      primaryKeys: toPrimaryKeys(primaryKeyRows),
+      foreignKeys: toForeignKeys(importedKeyRows)
+    };
+  };
+
+  const runQuery = async (sql: string) => {
+    const startTime = performance.now();
+
+    let lastRows: Record<string, unknown>[] = [];
+    let lastFields: { name: string }[] = [];
+    let lastRowCount: number | null = null;
+    let lastCommand = "";
+    let lastIsTruncated = false;
+
+    try {
+      for (const statementSql of splitSnowflakeStatements(sql)) {
+        // eslint-disable-next-line no-await-in-loop
+        const statement = await execute(statementSql);
+        // eslint-disable-next-line no-await-in-loop
+        const { rows, truncated } = await readRows(statement);
+
+        lastCommand = extractCommand(statementSql);
+        isInTransaction = nextTransactionState(lastCommand, isInTransaction);
+        lastRows = rows;
+        lastFields = (statement.getColumns() ?? []).map((column) => ({ name: column.getName() }));
+        // The driver reports -1, not undefined, when nothing was updated, so a SELECT falls through
+        const updatedRows = statement.getNumUpdatedRows();
+        lastRowCount =
+          typeof updatedRows === "number" && updatedRows >= 0 ? updatedRows : (statement.getNumRows() ?? null);
+        lastIsTruncated = truncated;
+      }
+    } finally {
+      runningStatement = null;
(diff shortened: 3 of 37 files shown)


backend/src/ee/services/pam-web-access/snowflake/pam-snowflake-relay-client.ts
```diff
@@ -0,0 +1,23 @@
...
+    accessUrl: `http://127.0.0.1:${relayPort}`,
+    application: "Infisical"
+  });
+
+  try {
+    return await client.connectAsync();
+  } catch (err) {
+    client.destroy(() => {});
+    throw new BadRequestError({
+      message: `Unable to reach Snowflake through the gateway: ${(err as Error)?.message ?? "connection failed"}`
```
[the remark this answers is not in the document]
lb-vn: not a real issue

backend/src/ee/services/pam-web-access/snowflake/pam-snowflake-connection-controller.ts
```diff
@@ -0,0 +1,245 @@
...
+  statement: snowflake.RowStatement
+): Promise<{ rows: Record<string, unknown>[]; truncated: boolean }> =>
+  new Promise((resolve, reject) => {
+    const rows: Record<string, unknown>[] = [];
+    let truncated = false;
+    const stream = statement.streamRows();
+
+    stream.on("data", (row: Record<string, unknown>) => {
+      if (rows.length < MAX_ROWS) {
+        rows.push(row);
```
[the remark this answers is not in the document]
lb-vn: same as postgres, ignoring for now

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -21,6 +21,7 @@ import {
 } from "../pam-account/pam-account-connection-test";
```
carlosmonastyrski, line 21: Ignore comment placement
Table names issue on web view
Image

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -21,6 +21,7 @@ import {
...
 import { TPamAccountDALFactory, TPamAccountDetail } from "../pam-account/pam-account-dal";
```
carlosmonastyrski, line 22: Ignore comment placement
Pagination does not work, page size does, but it's not shown correctly and can not navigate to the next page

Image

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -21,6 +21,7 @@ import {
...
 import {
```
carlosmonastyrski, line 23: Ignore comment placement
Two issues on "Check now", I can edit it, but if I click that it's taking the original credentials when it checks (tried with an invalid cred and worked). And it clears the form losing the chages

https://github.com/user-attachments/assets/0c2bc3b0-fc4b-484d-a4b5-595fcca20617

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -21,6 +21,7 @@ import {
...
+  collectCredentialSecrets,
```
carlosmonastyrski, line 24: booleans come back blank in the grid and sometimes long numbers shows the wrong number (9007199254740993 renders as …992, dates show as full timestamps)

Image

Image

backend/src/ee/services/pam-web-access/snowflake/pam-snowflake-relay-client.ts
```diff
@@ -0,0 +1,23 @@
+import snowflake from "snowflake-sdk";
+
+import { BadRequestError } from "@app/lib/errors";
+
+// The gateway answers Snowflake's REST API on the relay port and authenticates for the session
+export const connectThroughRelay = async (relayPort: number, account: string): Promise<snowflake.Connection> => {
+  const client = snowflake.createConnection({
```
carlosmonastyrski, line 10: no timeout set here so the SDK default of 90s applies while the gateway allows 30 min. a long query gets re-sent to snowflake after the 90s teardown, so one click runs the statement twice, and only one of the two is in the session recording.

docs/documentation/platform/pam/accounts/snowflake.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+Snowflake accounts let you manage access to a Snowflake database. Users connect through Web Access or the CLI, and every statement they run is recorded in the [session](/documentation/platform/pam/sessions/overview).
+
+Snowflake has no database wire protocol, so the [gateway](/documentation/platform/gateways/overview) answers the REST API its drivers speak and runs each statement itself. The credential never reaches the client, and because the gateway sits in your own network, an account behind AWS PrivateLink or a network policy is reachable like any other target.
+
+## Prerequisites
+
+- A Snowflake user for Infisical to connect as.
+- A warehouse the user's role can use. Snowflake bills for the compute a session's queries consume.
+- A [gateway](/documentation/platform/gateways/overview) that can reach your Snowflake account. Snowflake returns large results from cloud object storage rather than from the account host, so a gateway behind an egress allowlist needs that storage reachable too.
```
carlosmonastyrski, line 15: should we name the minimum cli version here?

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
```diff
@@ -21,6 +21,7 @@ import {
...
   parseInternalMetadata,
```
carlosmonastyrski, line 25: Ignore comment placement
In pam-project-bootstrap, should we add a snowflake default template there? so a fresh org does not have to create one by hand before it can add an account?

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -80,6 +81,20 @@ export type TestConnectionRequest =
       privateKey?: string;
       certificate?: string;
     }
+  | {
```
sheensantoscapadngan, line 84: What are your thoughts on automatically creating the snowflake account templates for existing organizations? Seems a bit odd for somebody to have to manually create a new one from scratch for every single new account type that we add 

Image

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -80,6 +81,20 @@ export type TestConnectionRequest =
...
+      account: string;
+      authMethod: string;
+      username: string;
+      password?: string;
+      token?: string;
+      privateKey?: string;
+      privateKeyPassphrase?: string;
+      warehouse?: string;
+      database?: string;
+      schema?: string;
```
sheensantoscapadngan, line 95: It would be best for the command blocking placeholders to be specific to Snowflake
Image

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -300,6 +315,42 @@ export const buildGatewayConnectionTest = async (
...
+      const cd = connectionDetails as {
+        account: string;
+        warehouse?: string;
+        database?: string;
+        schema?: string;
+        role?: string;
+      };
+      const c = creds as {
+        authMethod: string;
+        username: string;
```
sheensantoscapadngan, line 328: What do you mean by "Snowflake charges for its compute" here?

Image

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -300,6 +315,42 @@ export const buildGatewayConnectionTest = async (
...
+        password?: string;
+        token?: string;
+        privateKey?: string;
+        privateKeyPassphrase?: string;
+      } | null;
+      if (!c) return tcp(host, port);
+      return {
```
sheensantoscapadngan, line 335: Could this instead be a list of allowed roles that you can use with the account, and when the user does try to access, they can pick one?

Image

Same question for the following fields. I know we spoke about how we could potentially model Postgres and all the other databases like this too, given that when you say PAM account, you're referring to the actual service user under the hood (not necessarily a selection of database schema and role). 
Image

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -300,6 +315,42 @@ export const buildGatewayConnectionTest = async (
...
+        request: {
+          mode: TestConnectionMode.Snowflake,
+          account: cd.account,
+          authMethod: c.authMethod,
+          username: c.username,
+          password: c.password,
+          token: c.token,
+          privateKey: c.privateKey,
+          privateKeyPassphrase: c.privateKeyPassphrase,
+          warehouse: cd.warehouse,
```
sheensantoscapadngan, line 347: What do you mean by default here? Can somebody open a connection to a different schema altogether? 

Image

docs/documentation/platform/pam/accounts/snowflake.mdx
```diff
@@ -0,0 +1,149 @@
+---
+title: "Snowflake accounts"
+sidebarTitle: "Snowflake"
+description: "Add and connect to Snowflake accounts."
+---
+
+Snowflake accounts let you manage access to a Snowflake database. Users connect through Web Access or the CLI, and every statement they run is recorded in the [session](/documentation/platform/pam/sessions/overview).
+
+Snowflake has no database wire protocol, so the [gateway](/documentation/platform/gateways/overview) answers the REST API its drivers speak and runs each statement itself. The credential never reaches the client, and because the gateway sits in your own network, an account behind AWS PrivateLink or a network policy is reachable like any other target.
```
sheensantoscapadngan: I don't think we have to explicitly mention the database wire protocol aspect. Maybe we can just say that things are being executed within the gateway, and if you're using private link or something, make sure that the gateway is able to reach it?

docs/documentation/platform/pam/accounts/snowflake.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+## Prerequisites
+
+- A Snowflake user for Infisical to connect as.
+- A warehouse the user's role can use. Snowflake bills for the compute a session's queries consume.
```
sheensantoscapadngan: same Q here re: last sentence

docs/documentation/platform/pam/accounts/snowflake.mdx
```diff
@@ -0,0 +1,149 @@
...
+    The command prints the port and a ready-made connection string for each client.
+
+    **Flags:**
+    - `--port <port>` — use a specific local port (otherwise one is assigned automatically)
+    - `--reason <reason>` — provide an access reason (if required by template)
+
+  </Tab>
+</Tabs>
+
+Results are capped at 10,000 rows per statement, and the Data Explorer shows the first 1,000 rows of a result and tells you when there are more.
```
sheensantoscapadngan: Is this a technical limit from Snowflake, or are we enforcing this cap ourselves? Why is it capped?

lb-vn: We add the cap to not exhaust memory. I updated this line to be more specific

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: Made the description more clear. It just means where we default the connection to, and yeah they can

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: removed cause its unnecessary knowledge, but it literally meant that the warehouse would get charged for running queries.

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: Honestly this has been the pattern for every new account type. Ever since the initial seed through the migration we haven't auto added any account types when implementing them.

Perhaps we need to redesign this system so that all the new account types get seeded automatically without need to run any migraiton.

lb-vn: I did add it to project bootstrap though, so new orgs get it

docs/documentation/platform/pam/accounts/snowflake.mdx
lb-vn: I honestly think we need some robust system for feature management rather than guessing the next CLI version 🤣 Like you did this for oracle and then it was already innacurate after a release.

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
lb-vn: This is intentional, that's just the database name + table name. It's the same for postgres:

image

backend/src/ee/services/pam-account-heartbeat/pam-account-heartbeat-service.ts
lb-vn: fixed

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: You cant really force an account to only use a list of roles, cause they can always swap after connecting. I honestly think this is ok to leave as is, cause the default role thing makes the most sense to me.

lb-vn: For other fields as well

backend/src/ee/services/pam-account/pam-account-connection-test.ts
sheensantoscapadngan: can we not just run a migration now?

docs/documentation/platform/pam/accounts/snowflake.mdx
sheensantoscapadngan: which memory? Infisical's memory or the gateway? do we have the same cap elsewhere?

docs/documentation/platform/pam/accounts/snowflake.mdx
sheensantoscapadngan: what do you mean inaccurate? it's the minimum version which will have the snowflake changes... 

we could even apply the gateway capabilities check here right?

backend/src/ee/services/pam-account/pam-account-connection-test.ts
sheensantoscapadngan: should we instead call the field "Default Role" then?

docs/documentation/platform/pam/accounts/snowflake.mdx
lb-vn: I'm saying we'd need to make this change right before merging the PR for it to be accurate or else the CLI could be released before we merge

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: Would we just add snowflake or other account types as well?

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: Done

docs/documentation/platform/pam/accounts/snowflake.mdx
lb-vn: The gateways memory

backend/src/ee/services/pam-account/pam-account-connection-test.ts
sheensantoscapadngan: i think we can just do it for snowflake

docs/documentation/platform/pam/accounts/snowflake.mdx
sheensantoscapadngan: yeah that should still be worth doing i think. better than having nothing at this point

sheensantoscapadngan: Also, I think we should add that gateway capabilities check, which will ensure that when people are using the latest version of the platform, they are prompted to upgrade their gateways accordingly if they're trying to use Snowflake.

lb-vn: we already do that check actually

backend/src/ee/services/pam-web-access/snowflake/pam-snowflake-connection-controller.ts
```diff
@@ -0,0 +1,261 @@
...
+        const columns = statement.getColumns() ?? [];
+        lastRows = asDateOnly(
+          rows,
+          columns.filter((column) => column.getType()?.toLowerCase() === "date").map((column) => column.getName())
+        );
+        lastFields = columns.map((column) => ({ name: column.getName() }));
+        // The driver reports -1, not undefined, when nothing was updated, so a SELECT falls through
+        const updatedRows = statement.getNumUpdatedRows();
+        lastRowCount =
+          typeof updatedRows === "number" && updatedRows >= 0 ? updatedRows : (statement.getNumRows() ?? null);
```
carlosmonastyrski, line 164: getNumRows() returns data.returned (capped), not data.total, so the SQL editor still shows “Showing 1,000 of 10,000” for a 12,000-row table. The grid is fine because it runs its own COUNT(*).

Image

frontend/src/pages/pam/PamDataExplorerPage/PamDataExplorerPage.tsx
```diff
@@ -40,11 +45,18 @@ export const PamDataExplorerPage = ({ reason, mfaSessionId }: Props = {}) => {
...
-      : "public";
+  const dialect: SqlDialect =
+    (account?.accountType && DIALECT_BY_ACCOUNT_TYPE[account.accountType]) ?? "postgres";
+  const connectionDetails = account?.connectionDetails as { database?: string; schema?: string };
+
+  let defaultSchema: string;
+  if (dialect === "mysql") {
+    defaultSchema = connectionDetails?.database ?? "";
+  } else if (dialect === "snowflake") {
+    defaultSchema = connectionDetails?.schema ?? "PUBLIC";
```
carlosmonastyrski, line 57: found by Claude
Snowflake upper-cases unquoted identifiers and loadSchemas matches exactly, so a configured schema that doesn’t match is silently replaced by result[0]. Verified live: this account is configured with Default Schema PUBLIC and the explorer browses TPCDS_SF100TCL instead, with no warning. Probe() uses EqualFold, so the connection test won’t catch it.
Configured PUBLIC:
Image
Browsing TPCDS_SF100TCL:
Image

backend/src/ee/services/pam-web-access/snowflake/pam-snowflake-connection-controller.ts
lb-vn: The thing is we hard-cap at 10,000 either way. So its not possible to reach rows 10,001 - 12,000. This is a hard limitation we have for any account type in order to not exhaust gateway memory.

backend/src/ee/services/pam/pam-policies.ts
```diff
@@ -105,8 +105,13 @@ export const PAM_POLICY_DEFINITIONS: Record<PamPolicyType, TPamPolicyDefinition>
   [PamPolicyType.CommandBlocking]: {
     label: "Command Blocking",
     description: "Matching commands will be rejected (one regex per line).",
-    appliesTo: [PamAccountType.SSH],
-    schema: patternsStringSchema()
+    appliesTo: [PamAccountType.SSH, PamAccountType.Snowflake],
```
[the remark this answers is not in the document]
lb-vn: documented issue, ignoring

carlosmonastyrski (approved):

PR #8072, opened 2026-09-10
feat: add transactional event outbox and event-triggered alerts

Thiago-AS:
## Context

The alerting module learned about event-triggered alerts. It was built around a daily cron that scans for things that are due. Now each event a provider declares says how it fires: scheduled events keep the cron scan and per-target dedup, event-triggered ones skip both (the target is already known, and an event that happened won't be rediscovered tomorrow). Each event also owns its own condition shape, so an expiry threshold and a change notification don't have to share one. The frontend form follows along and only shows the threshold and daily reminder fields for the expiry event.

A generic event outbox. An auth method change is a notification you can't afford to drop, so the event is written as a row inside the same transaction as the business write instead of being pushed to Redis after commit. A relay picks rows up every 10s, retry state lives on the row with exponential backoff, and a claim is a lease so a dead pod hands its work back. It's not alert-specific: alerting is registered as one consumer, and anything else that needs at-least-once delivery of a domain event can register next to it.

## Steps to verify the change

1. Create a machine identity in the org and give it Universal Auth.
2. Open the identity, select Alert in the top right, and create an alert with event Auth Method Change. Notice the condition and daily reminder fields disappear once you pick it. Add an Email channel with yourself as recipient.
3. Trigger a change on that identity: add another auth method, edit the existing one, or create a new UA client secret.
4. Within ~10s a notification lands in Mailhog naming the auth method, the change, who made it, and when. No cron run and no waiting for the next day.
5. Repeat with a few more changes (revoke the client secret, delete an auth method) and confirm each one sends its own notification rather than being deduped.
6. Test with a machine identity that doesn't have an alert set and check that is no-op, the event is consumed but nothing is triggered.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/alert/alert-engine.ts
```diff
@@ -218,13 +187,137 @@ export const alertEngineFactory = ({
...
     }
 
-    await alertHistoryDAL.createWithTargets(alert.id, { status }, deliveries);
+    // Deliberately not allowed to fail the run: the event path may retry, and a throw after channels
+    // have already sent would re-notify.
+    try {
+      await alertHistoryDAL.createWithTargets(alert.id, { status }, deliveries);
+    } catch (err) {
+      logger.error(err, `Failed to record alert history after delivery [alertId=${alert.id}]`);
+    }
```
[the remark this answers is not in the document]
Thiago-AS: This isn't a regression from this change. On main, createWithTargets was awaited with no try/catch, so it threw out of runAlert into the AlertDispatch handler, which logs and rethrows. That job runs with attempts: 1 and removeOnFail: true, so BullMQ dropped it. Nothing retried the history write and the notification had already been sent, so the next daily scan re-alerted the same target either way. Throwing after the send never protected dedup; it just failed the job.

backend/src/services/event-outbox/event-outbox-dal.ts
```diff
@@ -0,0 +1,282 @@
...
+    maxAttempts: number
+  ): Promise<{ retried: number; failed: number }> => {
+    try {
+      return await db.transaction(async (tx) => {
+        const staleRows = await tx(TableName.EventOutbox)
+          .where("status", EventOutboxStatus.Processing)
+          .andWhereRaw(`"lockedAt" < NOW() - (? || ' milliseconds')::INTERVAL`, [thresholdMs])
+          .forUpdate()
+          .skipLocked()
+          .select<{ id: string; attempts: number }[]>("id", "attempts");
```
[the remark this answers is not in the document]
Thiago-AS: The transaction is bounded, but implicitly. Only processing rows qualify, and a row is only processing between claimBatch and commitResults; drain handles one batch per key at a time, so the ceiling is pods × FLUSH_WORKER_CONCURRENCY × OUTBOX_CLAIM_BATCH_SIZE, not the outage backlog. The pending backlog never enters this query, and a Redis outage can't grow it since nothing claims.

Thiago-AS:
@greptile review

Thiago-AS:
@greptile review

backend/src/services/alert/alert-event-consumer.ts
```diff
@@ -0,0 +1,155 @@
+import { z } from "zod";
+
+import { TEventOutbox } from "@app/db/schemas";
+import { AlertDispatchOutcome, recordAlertDispatchOutcomeMetric } from "@app/lib/telemetry/metrics";
+import {
+  EventOutboxStatus,
+  IEventOutboxConsumer,
+  TOutboxRowResult
```
lobocv: suggestion: I think the fact that this is an outbox is not relevant to consumers. From their perspective it is an event that we subscribe to. It would help readability to abstract this away. Only the actually outbox code itself needs to know about the pattern, consumers deal with a simpler event interface.

I find that the codebase is hard to follow because we don't separate these concerns and let them leak through to consumers

backend/src/services/alert/alert-event-consumer.ts
```diff
@@ -0,0 +1,155 @@
...
+  // subscribesTo only sees the event type, so a valid event key on the wrong resourceType gets past it.
+  // Fail here, terminally and naming both, instead of marking the row delivered as "no matching alert".
+  const $isDeclaredEvent = (row: TEventOutbox): boolean =>
+    Boolean(
+      alertProviderRegistry
+        .get(row.resourceType)
+        ?.events.some((event) => event.key === row.eventType && event.triggerType === AlertTriggerType.Event)
+    );
+
+  const $handleRow = async (row: TEventOutbox, findAlerts: TAlertLookup): Promise<TOutboxRowResult> => {
```
lobocv: Similar comment, i dont think we should be talking about "handling rows". We're handling events here

backend/src/services/event-outbox/event-outbox-types.ts
```diff
@@ -0,0 +1,93 @@
...
+  progress?: Record<string, unknown> | null;
+};
+
+export type TOutboxFlushKey = {
+  consumer: string;
+  resourceType: string;
+  resourceId: string;
+};
+
+export interface IEventOutboxConsumer<TPayload = unknown> {
```
lobocv: I'd suggest renaming this to Subscriber or EventConsumer and leaving out the outbox language.

backend/src/services/event-outbox/event-outbox-types.ts
```diff
@@ -0,0 +1,93 @@
...
+  // Checked at emit. Don't make it `.strict()`: one event can feed several consumers, so each schema
+  // has to tolerate fields it doesn't care about.
+  payloadSchema: z.ZodType<TPayload>;
+
+  // Runs on the request path for every emit, so keep it pure and in-memory. Whether anyone actually
+  // wants the event is decided later, in handle().
+  subscribesTo(eventType: string): boolean;
+
+  // Rows for a single (resourceType, resourceId), in id order. Must return one result per row.
+  handle(rows: TEventOutbox[]): Promise<TOutboxRowResult[]>;
```
lobocv: I'd also suggest having the public types be named "Event" without outbox

backend/src/db/migrations/20260908000000_event-outbox.ts
```diff
@@ -0,0 +1,70 @@
...
+import { createOnUpdateTrigger, dropOnUpdateTrigger } from "../utils";
+
+export async function up(knex: Knex): Promise<void> {
+  if (!(await knex.schema.hasTable(TableName.EventOutbox))) {
+    await knex.schema.createTable(TableName.EventOutbox, (t) => {
+      t.bigIncrements("id").primary();
+      t.string("consumer").notNullable();
+      t.string("eventType").notNullable();
+      t.string("resourceType").notNullable();
+      t.string("resourceId").notNullable();
```
lobocv: Why is resourceType and Id needed at the outbox table? It doesn't seem right to be a top level column. I couldn't see where it was actually used in the queries. Maybe it should be inside the payload and let each event decide. 

I'm not even sure orgId and projectId need to be here either. Are they queried directly? They should be in the payload as well IMO. The event payload should contain the tennant info. I dont see why the outbox processor should need to query on those fields? Am i missing something?

backend/src/db/migrations/20260908000000_event-outbox.ts
```diff
@@ -0,0 +1,70 @@
...
+      t.jsonb("payload").notNullable();
+      t.string("idempotencyKey");
+      t.timestamp("occurredAt", { useTz: true }).notNullable().defaultTo(knex.fn.now());
+      t.string("status").notNullable().defaultTo("pending");
+      t.check(`"status" IN ('pending', 'processing', 'retry', 'delivered', 'failed')`);
+      t.integer("attempts").notNullable().defaultTo(0);
+      t.timestamp("nextRetryAt", { useTz: true }).notNullable().defaultTo(knex.fn.now());
+      t.timestamp("lockedAt", { useTz: true });
+      t.uuid("lockToken");
+      t.jsonb("progress");
```
lobocv: This doesn't seem to be used (the alert module sets it to null on makeEvent). Do we really need it for v1 of this outbox? I think we should add it later if we have a real need for it. I'm not even sure what it's doing since each row is for it's own event and consumer. and there is already a status column

backend/src/db/migrations/20260908000000_event-outbox.ts
```diff
@@ -0,0 +1,70 @@
...
+      CREATE INDEX IF NOT EXISTS "${TableName.EventOutbox}_terminal_idx"
+      ON "${TableName.EventOutbox}" ("updatedAt")
+      WHERE status IN ('delivered', 'failed')
+    `);
+
+    await knex.schema.raw(`
+      CREATE UNIQUE INDEX IF NOT EXISTS "${TableName.EventOutbox}_idempotency_idx"
+      ON "${TableName.EventOutbox}" ("consumer", "idempotencyKey")
+      WHERE "idempotencyKey" IS NOT NULL
+    `);
```
lobocv, line 56: i dont think this is necessary. The generation of the idempotency key should be a single place, internal to the outbox package. It uses a UUID plus, we prune old events, so there is zero chance of collision.
Having this unique index is overkill and consumes postgres resources and slows down insertions.

If you want to keep it, i'd suggest removing the consumer part of the compound index, it doesn't add anything (again the idempotencyKey is UUID)

lobocv (commented):
Not done review yet but getting distracted by a meeting and i want to give you enough time to read the comments rather than bomabard you all at once!

backend/src/db/migrations/20260908000000_event-outbox.ts
Thiago-AS: ResourceType and resourceId were the partition key for delivery. findDueFlushKeys groups pending rows by (consumer, resourceType, resourceId) to find due partitions, claimBatch filters on the same key, the drain index leads with them, and the flush job id is derived from them so only one worker drains a resource at a time. 

We can drain per consumer. The cost is a throughput ceiling of one batch in flight per consumer. Ok for now, so will change.

backend/src/db/migrations/20260908000000_event-outbox.ts
Thiago-AS: It is used, it tracks partial success on consumer delivery. But let me abstract that for now.

backend/src/db/migrations/20260908000000_event-outbox.ts
Thiago-AS: The outbox doesn't generate this. It's an optional key the emitter supplies so a re run of the same business event collapses into one row, and the unique index is what makes the ON CONFLICT DO NOTHING work. Consumer has to be in it because one emit becomes one row per consumer with the same key, so a key-only index would drop every consumer after the first

docs/documentation/platform/identities/machine-identities.mdx
```diff
@@ -100,21 +109,30 @@ Alerts currently cover **Universal Auth client secrets that have a TTL**. A clie
...
+
+It also fires when a credential inside a method changes: a Universal Auth client secret is created or revoked, or a Token Auth token is created, renamed, or revoked. Access tokens minted by logging in with a credential aren't credentials themselves and don't fire it.
+
+Each notification names the auth method, the kind of change, the credential where one is involved, who made the change (a user's email or a machine identity's name), and when. It takes no condition and has no daily run: the notification is sent as soon as the change is saved.
+
 ### Create an alert
 
-Open the machine identity you want to monitor and select **Alert** in the top-right of the page to create one.
+Open the machine identity you want to monitor and select **Alert** in the top-right of the page to create one. An identity holds one alert, so pick the event you care about most.
```
victorvhs017, line 122: This really doesn't make sense from a UI/UX point of view. What's blocking us from allowing at least one alert per event type? This is already possible through the API.

Image

backend/src/lib/cron/cron-job.ts
```diff
@@ -47,6 +47,8 @@ export const CronJobName = {
   PamHeartbeatQueueChecks: "pam-heartbeat-queue-checks",
   MonthlyNativeIntegrationDeprecationNotice: "monthly-native-integration-deprecation-notice",
   DailyAlertProcessing: "daily-alert-processing",
+  EventOutboxStaleClaimSweeper: "event-outbox-stale-claim-sweeper",
+  EventOutboxCleanup: "event-outbox-cleanup",
```
victorvhs017, line 51: Reminder to get the infra PR up, adding these cron jobs and with alerts in the new metrics, now that OTEL is up

backend/src/lib/telemetry/metrics.ts
victorvhs017, line 1: We need to update the metrics docs to add these new ones.

backend/src/services/event-outbox/event-outbox-service.ts
```diff
@@ -0,0 +1,316 @@
...
+    try {
+      return await consumer.handle(claimed);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : "Unknown error";
+      logger.error(error, `event-outbox: consumer '${key.consumer}' threw while handling a batch`);
+      return claimed.map((row) => ({ id: String(row.id), status: EventResultStatus.Retry, error: message }));
+    } finally {
+      clearInterval(heartbeat);
+    }
+  };
```
victorvhs017, line 268: The heartbeat can extend the claim indefinitely. We could put a timeout on the consumer.handle to handle this edge case

frontend/src/hooks/api/alerts/types.ts
```diff
@@ -28,7 +29,15 @@ export const ALERT_RESOURCE_TYPE_LABELS: Record<AlertResourceType, string> = {
...
-  [AlertEventType.IdentityAuthenticationExpiry]: "Expiration"
+  [AlertEventType.IdentityAuthenticationExpiry]: "Client Secret Expiration",
+  [AlertEventType.IdentityAuthMethodChanged]: "Auth Method Change"
+};
+
+export const ALERT_EVENT_TYPE_DESCRIPTIONS: Record<AlertEventType, string> = {
+  [AlertEventType.IdentityAuthenticationExpiry]:
+    "Notify a set number of days before a Universal Auth client secret expires.",
+  [AlertEventType.IdentityAuthMethodChanged]:
+    "Notify whenever an auth method is added, updated, or removed."
```
victorvhs017: We need to mention that this alerts on credential changes too.

victorvhs017 (changes requested):

docs/documentation/platform/identities/machine-identities.mdx
Thiago-AS: Agree! but it was wired that way in the UI initially. We can definitely change it, but I wouldn’t do that in this PR since it’s outside the scope and requires an UI/UX exploration.

backend/src/services/event-outbox/event-outbox-service.test.ts
victorvhs017, line 1: ignore placement:

I get the credential expiry notice test when I set the "auth method change":

Image

Image

victorvhs017, line 1: ignore placement:

You are able to create an alert for a machine identity passing the suborg id and that alert would never run because it would mismatch the identity org Id with the alert suborg Id

victorvhs017 (changes requested):

backend/src/services/event-outbox/event-outbox-service.test.ts
Thiago-AS: Made it more abstract!

Thiago-AS: Fixed, you should not be able to do that. UI already blocks, added a BE block.

victorvhs017 (approved):

PR #8087, opened 2026-09-11
feat(secret-sync): recursive secret syncs, stage one

lobocv:
## Context

Stage one of recursive secret syncs ([SECRETS-525](https://linear.app/infisical/issue/SECRETS-525/recursive-secret-syncs)). Adds a `recursive` sync option: when it is on, a sync pushes every secret in its source folder and in all folders beneath it.

Infisical does not decide where those secrets land at the destination. Each provider does. Today every provider lands them in one flat list. Providers whose destination has its own folder structure, such as AWS Parameter Store, can later keep the paths. That is separate work.

**Stacked on #<e2e-harness-pr> (`calvin/secret-sync-e2e-tests`), which must merge first.**

### The contract

Providers receive a payload object with two views:

```ts
all()       // every secret, each with its name, folder path, id and value
flatten()   // the same secrets as a single flat list, keyed by name
```

`flatten()` is where duplicate names are caught. Two secrets in different folders can share a name and a flat destination has room for only one, so calling `flatten()` reports every clash, naming both folders, and the sync fails. A provider that keeps the folder structure calls `all()` instead and never meets the problem, because it does not have one.

Obtaining the flat view is what runs the check, so there is no separate call to forget. The payload is an object rather than a record for the same reason: every provider previously destructured a `Record` with `Object.entries` and reconciled deletions with `key in secretMap`, and both of those compile against an array or a path-keyed record, so an unconverted provider would have typechecked and silently either named secrets `0`, `1`, `2` or deleted the whole destination.

### Who does what

| Part | Responsibility |
| --- | --- |
| Infisical | Collect every secret below the source folder and hand it over. Knows folders, permissions and imports. Knows nothing about destinations. |
| The payload | Offer two ways to read the same secrets. Decides nothing. |
| Provider | Choose one, and place the secrets at the destination. Knows what its own destination supports. |

### Permissions

The sync worker performs no authorization of its own, so create and update time is the only gate. Folder grants let someone read `/backend` while being denied `/backend/api`, so without a check a recursive sync would have read the denied folder on their behalf and delivered it to a destination they control. Every folder a sync will read is now authorized, on create and on every update that leaves the sync recursive.

### Secret imports

Inside one folder a secret that lives there still beats one pulled in, which is existing behaviour and unchanged. Across folders, two folders offering the same name is a clash and the sync fails.

## Screenshots

Not added yet. The `Include subfolders` toggle sits on the source step, disabled with a stated reason when an import initial-sync behavior is selected.

## Steps to verify the change

```
make test-api-unit SPEC=secret-sync     # 37 tests
make test-api-e2e  SPEC=secret-sync     # 29 tests, 4 files
cd backend && npm run type:check        # zero errors
```

Note: the end-to-end harness does not reset state between runs, so a docker stack left up across many runs accumulates enough data to start timing out. If you see intermittent "Timed out waiting for a sync run to reach the destination", tear it down first:

```
docker compose -f docker-compose.test.yml --profile runner down -v
```

Five consecutive runs pass on a freshly rebuilt stack, with no degradation across them.

## Other things worth a look

- **A pre-existing hole, not introduced here.** An actor with `SecretSyncs.Edit` on a **non-recursive** sync's source path, but no read permission on those secrets, can still repoint that sync at a destination they control. The update gate only fires when the source changes. This is on `main` today. Fixing it would re-check permissions on every edit of every existing sync and could refuse edits that work now, so it is left for a separate decision.
- **A bug this work uncovered and fixed.** `UpdateAwsSecretsManagerSyncSchema` dereferenced `sync.syncOptions.syncSecretMetadataAsTags` without a guard while `syncOptions` is optional on the update schema, so a `PATCH` omitting it threw. Fixed in `f16d30e65a`.
- Authorization happens at create and update time while the folder set resolves at run time, so a subfolder created under the source after authorization is read with no further check.
- **Two unrelated fixes rode along, both found while debugging the suite.** Queue job failures were logged nowhere, only counted, so an uncaught handler exception was invisible (`cee9aca2cc`). And `PINO_LOG_LEVEL=debug` was a no-op product-wide, because the pino transport target was pinned to `info` while the logger read the variable (`f123037698`). Both are small and separable if you would rather they went in on their own.
- Import-from-destination is rejected while `recursive` is on, because there is no single folder to import into.

## Type

- [x] Feature

## Checklist

- [x] Tested locally (37 unit, 29 e2e, type check clean)

### DIFF
--- a/backend/src/lib/api-docs/constants.ts
+++ b/backend/src/lib/api-docs/constants.ts
@@ -3101,7 +3101,8 @@ export const SecretSyncs = {
     return {
       initialSyncBehavior: `Specify how Infisical should resolve the initial sync to the ${destinationName} destination.`,
       keySchema: `Specify the format to use for structuring secret keys in the ${destinationName} destination.`,
-      disableSecretDeletion: `Enable this flag to prevent removal of secrets from the ${destinationName} destination when syncing.`
+      disableSecretDeletion: `Enable this flag to prevent removal of secrets from the ${destinationName} destination when syncing.`,
+      recursive: `Whether to sync secrets from folders beneath the source path as well.`
     };
   },
   ADDITIONAL_SYNC_OPTIONS: {
--- a/backend/src/services/secret-sync/external-infisical/external-infisical-sync-fns.ts
+++ b/backend/src/services/secret-sync/external-infisical/external-infisical-sync-fns.ts
@@ -6,6 +6,7 @@ import { blockLocalAndPrivateIpAddresses } from "@app/lib/validator";
 import { getExternalInfisicalAccessToken } from "@app/services/app-connection/external-infisical";
 import { SecretSyncError } from "@app/services/secret-sync/secret-sync-errors";
 import { matchesSchema } from "@app/services/secret-sync/secret-sync-fns";
+import { TSecretSyncPayload } from "@app/services/secret-sync/secret-sync-payload";
 import { TSecretMap } from "@app/services/secret-sync/secret-sync-types";
 
 import { TExternalInfisicalSyncWithCredentials } from "./external-infisical-sync-types";
@@ -143,8 +144,11 @@ const batchDeleteSecrets = async (
 };
 
 export const ExternalInfisicalSyncFns = {
-  syncSecrets: async (secretSync: TExternalInfisicalSyncWithCredentials, secretMap: TSecretMap) =>
+  syncSecrets: async (secretSync: TExternalInfisicalSyncWithCredentials, payload: TSecretSyncPayload) =>
     withExternalInfisicalErrorHandling(async () => {
+      // Key schema must never be applied for Infisical-to-Infisical syncs, or the prefixed key
+      // triggers another sync cycle on the remote project.
+      const secretMap = payload.flatten({ applySchema: false });
       const ctx = await getRemoteContext(secretSync);
       const remoteSecrets = await fetchRemoteSecrets(secretSync, ctx);
       const environmentSlug = secretSync.environment?.slug || "";
@@ -189,8 +193,9 @@ export const ExternalInfisicalSyncFns = {
       return Object.fromEntries(remoteSecrets.map((s) => [s.secretKey, { value: s.secretValue ?? "" }]));
     }),
 
-  removeSecrets: async (secretSync: TExternalInfisicalSyncWithCredentials, secretMap: TSecretMap) =>
+  removeSecrets: async (secretSync: TExternalInfisicalSyncWithCredentials, payload: TSecretSyncPayload) =>
     withExternalInfisicalErrorHandling(async () => {
+      const secretMap = payload.flatten({ applySchema: false });
       const ctx = await getRemoteContext(secretSync);
       const secretsToDelete = Object.keys(secretMap);
       await batchDeleteSecrets(secretSync, secretsToDelete, ctx);
--- a/backend/src/services/secret-sync/secret-sync-payload.test.ts
+++ b/backend/src/services/secret-sync/secret-sync-payload.test.ts
@@ -1,4 +1,9 @@
-import { createSecretSyncPayload, dedupeEntriesByDestinationKey, TSecretPayload } from "./secret-sync-payload";
+import {
+  createSecretSyncPayload,
+  dedupeEntriesByDestinationKey,
+  findFlattenConflicts,
+  TSecretPayload
+} from "./secret-sync-payload";
 
 const secret = (key: string, path: string, value = "v"): TSecretPayload => ({ key, path, value });
 
@@ -81,6 +86,43 @@ describe("createSecretSyncPayload", () => {
     });
 
     expect(() => payload.flatten()).not.toThrow();
+  });
+});
+
+describe("findFlattenConflicts", () => {
+  test("returns every conflict, uncapped, unlike flatten()'s thrown message", () => {
+    const secrets = Array.from({ length: 7 }, (_, i) => [secret(`K${i}`, "/"), secret(`K${i}`, "/api")]).flat();
+
+    const conflicts = findFlattenConflicts({ secrets, environment: "dev" });
+
+    expect(conflicts).toHaveLength(7);
+    expect(conflicts.map((c) => c.key).sort()).toEqual(["K0", "K1", "K2", "K3", "K4", "K5", "K6"]);
+  });
+
+  test("reports the destination key and every path it appears in", () => {
+    const conflicts = findFlattenConflicts({
+      secrets: [secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")],
+      environment: "dev"
+    });
+
+    expect(conflicts).toEqual([{ key: "DB_URL", paths: ["/backend", "/backend/api"] }]);
+  });
+
+  test("returns an empty array when there are no conflicts", () => {
+    expect(findFlattenConflicts({ secrets: [secret("DB_URL", "/")], environment: "dev" })).toEqual([]);
+  });
+
+  test("applySchema: false reports raw-key collisions instead of schema-applied ones", () => {
+    const conflicts = findFlattenConflicts(
+      {
+        secrets: [secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")],
+        environment: "dev",
+        keySchema: "{{environment}}_{{secretKey}}"
+      },
+      { applySchema: false }
+    );
+
+    expect(conflicts).toEqual([{ key: "DB_URL", paths: ["/backend", "/backend/api"] }]);
   });
 });
 
--- a/backend/src/services/secret-sync/secret-sync-payload.test.ts
+++ b/backend/src/services/secret-sync/secret-sync-payload.test.ts
@@ -1,4 +1,4 @@
-import { createSecretSyncPayload, TSecretPayload } from "./secret-sync-payload";
+import { createSecretSyncPayload, dedupeEntriesByDestinationKey, TSecretPayload } from "./secret-sync-payload";
 
 const secret = (key: string, path: string, value = "v"): TSecretPayload => ({ key, path, value });
 
@@ -61,6 +61,41 @@ describe("createSecretSyncPayload", () => {
       environment: "dev",
       keySchema: "{{secretKey}}"
     });
+
+    expect(() => payload.flatten()).not.toThrow();
+  });
+});
+
+describe("dedupeEntriesByDestinationKey", () => {
+  test("keeps one entry per destination key when two folders share a name", () => {
+    const deduped = dedupeEntriesByDestinationKey([secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")], {
+      environment: "dev"
+    });
+
+    expect(deduped).toHaveLength(1);
+  });
+
+  test("leaves distinct names untouched", () => {
+    const deduped = dedupeEntriesByDestinationKey([secret("DB_URL", "/"), secret("API_KEY", "/api")], {
+      environment: "dev"
+    });
+
+    expect(deduped.map((entry) => entry.key).sort()).toEqual(["API_KEY", "DB_URL"]);
+  });
+
+  test("dedupes by the schema-applied destination key, not the raw name", () => {
+    const deduped = dedupeEntriesByDestinationKey([secret("DB_URL", "/"), secret("dev_DB_URL", "/api")], {
+      environment: "dev",
+      keySchema: "{{secretKey}}"
+    });
+
+    expect(deduped).toHaveLength(2);
+  });
+
+  test("the deduped result never conflicts when flattened", () => {
+    const entries = [secret("DB_URL", "/backend", "one"), secret("DB_URL", "/backend/api", "two")];
+    const deduped = dedupeEntriesByDestinationKey(entries, { environment: "dev" });
+    const payload = createSecretSyncPayload(deduped, { environment: "dev" });
 
     expect(() => payload.flatten()).not.toThrow();
   });
--- a/backend/src/services/secret-sync/secret-sync-payload.test.ts
+++ b/backend/src/services/secret-sync/secret-sync-payload.test.ts
@@ -1,0 +1,67 @@
+import { createSecretSyncPayload, TSecretPayload } from "./secret-sync-payload";
+
+const secret = (key: string, path: string, value = "v"): TSecretPayload => ({ key, path, value });
+
+describe("createSecretSyncPayload", () => {
+  test("all() returns every secret, including ones that share a name", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/"), secret("DB_URL", "/api")], {
+      environment: "dev"
+    });
+
+    expect(payload.all()).toHaveLength(2);
+  });
+
+  test("flatten() returns one entry per secret when names are unique", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/", "one"), secret("API_KEY", "/api", "two")], {
+      environment: "dev"
+    });
+
+    expect(payload.flatten()).toEqual({
+      DB_URL: { value: "one", id: undefined, comment: undefined, secretMetadata: undefined },
+      API_KEY: { value: "two", id: undefined, comment: undefined, secretMetadata: undefined }
+    });
+  });
+
+  test("flatten() throws naming both folders when a name appears twice", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")], {
+      environment: "dev"
+    });
+
+    expect(() => payload.flatten()).toThrow(/"DB_URL" in \/backend and \/backend\/api/);
+  });
+
+  test("flatten() reports every conflict in one throw, capped at five with a remainder count", () => {
+    const secrets = Array.from({ length: 7 }, (_, i) => [secret(`K${i}`, "/"), secret(`K${i}`, "/api")]).flat();
+    const payload = createSecretSyncPayload(secrets, { environment: "dev" });
+
+    let message = "";
+    try {
+      payload.flatten();
+    } catch (err) {
+      message = (err as Error).message;
+    }
+
+    expect(message).toContain("K0");
+    expect(message).toContain("K4");
+    expect(message).not.toContain("K5");
+    expect(message).toContain("and 2 more");
+  });
+
+  test("flatten() applies the key schema to destination keys", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/")], {
+      environment: "dev",
+      keySchema: "{{environment}}_{{secretKey}}"
+    });
+
+    expect(Object.keys(payload.flatten())).toEqual(["dev_DB_URL"]);
+  });
+
+  test("flatten() does not treat a name that only resembles a schema-prefixed one as a conflict", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/"), secret("dev_DB_URL", "/api")], {
+      environment: "dev",
+      keySchema: "{{secretKey}}"
+    });
+
+    expect(() => payload.flatten()).not.toThrow();
+  });
+});
--- a/backend/src/services/secret-sync/secret-sync-payload.test.ts
+++ b/backend/src/services/secret-sync/secret-sync-payload.test.ts
@@ -3,12 +3,12 @@ import { createSecretSyncPayload, dedupeEntriesByDestinationKey, TSecretPayload 
 const secret = (key: string, path: string, value = "v"): TSecretPayload => ({ key, path, value });
 
 describe("createSecretSyncPayload", () => {
-  test("all() returns every secret, including ones that share a name", () => {
+  test("secrets includes every secret, including ones that share a name", () => {
     const payload = createSecretSyncPayload([secret("DB_URL", "/"), secret("DB_URL", "/api")], {
       environment: "dev"
     });
 
-    expect(payload.all()).toHaveLength(2);
+    expect(payload.secrets).toHaveLength(2);
   });
 
   test("flatten() returns one entry per secret when names are unique", () => {
@@ -54,6 +54,24 @@ describe("createSecretSyncPayload", () => {
     });
 
     expect(Object.keys(payload.flatten())).toEqual(["dev_DB_URL"]);
+  });
+
+  test("flatten({ applySchema: false }) ignores the key schema", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/")], {
+      environment: "dev",
+      keySchema: "{{environment}}_{{secretKey}}"
+    });
+
+    expect(Object.keys(payload.flatten({ applySchema: false }))).toEqual(["DB_URL"]);
+  });
+
+  test("flatten({ applySchema: false }) still rejects a raw-key collision", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")], {
+      environment: "dev",
+      keySchema: "{{environment}}_{{secretKey}}"
+    });
+
+    expect(() => payload.flatten({ applySchema: false })).toThrow(/"DB_URL" in \/backend and \/backend\/api/);
   });
 
   test("flatten() does not treat a name that only resembles a schema-prefixed one as a conflict", () => {
--- a/backend/src/services/secret-sync/secret-sync-payload.test.ts
+++ b/backend/src/services/secret-sync/secret-sync-payload.test.ts
@@ -87,6 +87,31 @@ describe("createSecretSyncPayload", () => {
 
     expect(() => payload.flatten()).not.toThrow();
   });
+
+  test("findConflicts() reports the same collisions as flatten() throws, without throwing", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")], {
+      environment: "dev"
+    });
+
+    expect(payload.findConflicts()).toEqual([{ key: "DB_URL", paths: ["/backend", "/backend/api"] }]);
+  });
+
+  test("findConflicts() returns an empty array when there are no conflicts", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/")], { environment: "dev" });
+
+    expect(payload.findConflicts()).toEqual([]);
+  });
+
+  test("findConflicts({ applySchema: false }) reports raw-key collisions instead of schema-applied ones", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")], {
+      environment: "dev",
+      keySchema: "{{environment}}_{{secretKey}}"
+    });
+
+    expect(payload.findConflicts({ applySchema: false })).toEqual([
+      { key: "DB_URL", paths: ["/backend", "/backend/api"] }
+    ]);
+  });
 });
 
 describe("findFlattenConflicts", () => {
--- a/backend/src/services/secret-sync/secret-sync-payload.ts
+++ b/backend/src/services/secret-sync/secret-sync-payload.ts
@@ -15,8 +15,9 @@ export type TSecretPayload = {
 };
 
 export type TSecretSyncPayload = {
-  all: () => TSecretPayload[];
-  flatten: () => TSecretMap;
+  secrets: TSecretPayload[];
+  environment: string;
+  flatten: (opts?: { applySchema?: boolean }) => TSecretMap;
 };
 
 export const getKeyWithSchema = ({
@@ -87,41 +88,42 @@ export const dedupeEntriesByDestinationKey = (
 export const createSecretSyncPayload = (
   secrets: TSecretPayload[],
   { environment, keySchema }: { environment: string; keySchema?: string }
-): TSecretSyncPayload => {
-  let flattened: TSecretMap | undefined;
+): TSecretSyncPayload => ({
+  secrets,
+  environment,
+  // A destination whose sync targets can't carry the configured key schema (eg a many-to-one
+  // JSON body whose fields are app-facing variable names, or an Infisical-to-Infisical sync,
+  // where a schema-renamed key would look like a new secret and retrigger a sync cycle) calls
+  // flatten({ applySchema: false }) to get the raw-key view instead. Either way flatten() still
+  // groups by the resulting destination key and rejects collisions, so duplicate-name detection
+  // never depends on whether the schema was applied.
+  flatten: ({ applySchema = true }: { applySchema?: boolean } = {}) => {
+    const schema = applySchema ? keySchema : undefined;
+    const grouped = new Map<string, TSecretPayload[]>();
 
-  return {
-    all: () => secrets,
-    flatten: () => {
-      if (flattened) return flattened;
+    for (const entry of secrets) {
+      const destinationKey = getKeyWithSchema({ key: entry.key, environment, schema });
+      const group = grouped.get(destinationKey);
 
-      const grouped = new Map<string, TSecretPayload[]>();
-
-      for (const entry of secrets) {
-        const destinationKey = getKeyWithSchema({ key: entry.key, environment, schema: keySchema });
-        const group = grouped.get(destinationKey);
-
-        if (group) group.push(entry);
-        else grouped.set(destinationKey, [entry]);
-      }
-
-      const conflicts = [...grouped.entries()].filter(([, group]) => group.length > 1);
-
-      if (conflicts.length) throw buildConflictError(conflicts);
-
-      const map: TSecretMap = {};
-
-      for (const [destinationKey, [entry]] of grouped) {
-        map[destinationKey] = {
-          value: entry.value,
-          id: entry.id,
-          comment: entry.comment,
-          secretMetadata: entry.secretMetadata
-        };
-      }
-
-      flattened = map;
-      return map;
+      if (group) group.push(entry);
+      else grouped.set(destinationKey, [entry]);
     }
-  };
-};
+
+    const conflicts = [...grouped.entries()].filter(([, group]) => group.length > 1);
+
+    if (conflicts.length) throw buildConflictError(conflicts);
+
+    const map: TSecretMap = {};
+
+    for (const [destinationKey, [entry]] of grouped) {
+      map[destinationKey] = {
+        value: entry.value,
+        id: entry.id,
+        comment: entry.comment,
+        secretMetadata: entry.secretMetadata
+      };
+    }
+
+    return map;
+  }
+});
--- a/backend/src/services/secret-sync/secret-sync-payload.ts
+++ b/backend/src/services/secret-sync/secret-sync-payload.ts
@@ -58,6 +58,32 @@ const buildConflictError = (conflicts: [string, TSecretPayload[]][]) => {
   });
 };
 
+// The remove path deletes by destination key and does not care which duplicate wins, so it
+// dedupes before the payload is built. That keeps flatten() itself unconditional: a sync that
+// has drifted into a duplicate-name state must still be removable, since deleteSyncOnComplete
+// only drops the sync row after a successful remove.
+export const dedupeEntriesByDestinationKey = (
+  entries: TSecretPayload[],
+  { environment, keySchema }: { environment: string; keySchema?: string }
+): TSecretPayload[] => {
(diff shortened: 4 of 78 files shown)


backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -121,6 +137,140 @@ export const secretSyncServiceFactory = ({
 
   const preSaveTransformDeps = { secretV2BridgeDAL, appConnectionDAL, kmsService };
 
+  // The sync worker reads its folders with every access check disabled, so this is the only place
+  // the actor's read access to them is established. A recursive sync reads the whole subtree, and
+  // folder grants let an actor hold read on a parent while being denied on a child.
```
lobocv: the last sentence about folder grants is a little confusing. Maybe we should say why we can't rely on folder grants alone because they dont allow access to child folders? Clear this up but don't make the comment too long, 1-3 sentences

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -121,6 +137,140 @@ export const secretSyncServiceFactory = ({
...
+        });
+      }
+    }
+  };
+
+  // Flattening a subtree onto a destination that holds one flat list can produce two secrets with
+  // the same destination key. flatten() is what detects that, and the job runs it on every sync, so
+  // calling it here means a user reads the same sentence at save time and when the sync later drifts
+  // into the same state.
+  const $assertSyncableAtDestination = async ({
```
lobocv: this name is describing functionality that is required by the consumer elsewhere. Rename it to describe what it is doing, which is checking if the secret payload can flatten. maybe call it `assertSyncedSecretsAreFlattenable`?

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -121,6 +137,140 @@ export const secretSyncServiceFactory = ({
...
+  }: {
+    projectId: string;
+    actorOrgId: string;
+    environment: string;
+    sourcePath: string;
+    sourceFolderId: string;
+    recursive: boolean;
+    keySchema?: string;
+  }) => {
+    if (!recursive) return;
```
lobocv: remove this recursive argument and only make the caller call this when recursive == true

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -414,22 +564,36 @@ export const secretSyncServiceFactory = ({
...
+    await $assertCanReadSyncedFolders(projectPermission, {
+      projectId,
+      environment,
+      sourcePath: secretPath,
+      sourceFolderId: folder.id,
+      recursive: isRecursive
+    });
+
+    // Runs after the permission check above: the conflict it raises names every folder beneath the
+    // source, including ones this actor is denied on.
```
lobocv: low value comment, remove it.

lobocv: replace it with a comment saying "some destinations do not support heirarchical secret organization, so they must flatten the secrets to a single folder. Check for this early so that the sync doesn't get created and fail immediately on the first run"

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -612,29 +776,59 @@ export const secretSyncServiceFactory = ({
...
-    if (
-      (secretPath && secretPath !== secretSync.folder?.path) ||
-      (environment && environment !== secretSync.environment?.slug)
-    ) {
+    const isSourceChanged =
+      (Boolean(secretPath) && secretPath !== secretSync.folder?.path) ||
+      (Boolean(environment) && environment !== secretSync.environment?.slug);
+
+    const wasRecursive = Boolean((secretSync.syncOptions as { recursive?: boolean } | undefined)?.recursive);
+    const isRecursive = Boolean((params.syncOptions as { recursive?: boolean } | undefined)?.recursive ?? wasRecursive);
```
lobocv: why is this written this way? is there a way to simplify? why do we need to cast with `as`?

backend/src/services/secret-sync/secret-sync-recursive-fns.test.ts
```diff
@@ -0,0 +1,322 @@
...
+      [
+        { path: "/backend", secrets: [imported("DB_URL", "one")] },
+        { path: "/backend/api", secrets: [imported("DB_URL", "two")] }
+      ]
+    );
+
+    expect(merged.map((entry) => entry.path).sort()).toEqual(["/backend", "/backend/api"]);
+  });
+
+  test("a later import wins over an earlier one in the same folder, as today", () => {
```
lobocv: remove "as today" and explain (in comment) that this is released functionality that we need to pin so we dont make breaking changes. Mention ideally it would have errored on conflict

lobocv: also what does "later" mean? is the order deterministic?

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
```diff
@@ -0,0 +1,251 @@
...
+    projectId: string;
+    environment: string;
+    sourcePath: string;
+    sourceFolderId: string;
+    recursive: boolean;
+    keySchema?: string;
+    includeImports: boolean;
+    // The remove path passes true so a duplicate name across folders cannot leave the sync
+    // undeletable; every other caller must keep the duplicate-name check intact.
+    dedupeForRemoval?: boolean;
```
lobocv: what is this comment even saying? I don't understand it. It seems like this function has too much information on downstream callers. Whats the problem that this flag is attempting to solve? Is there a cleaner way to solve this?

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
```diff
@@ -0,0 +1,251 @@
...
+export const assertWithinSecretLimit = (count: number) => {
+  if (count <= SECRET_SYNC_MAX_SECRETS) return;
+
+  throw new SecretSyncError({
+    message: `This sync covers ${count} secrets, which is above the limit of ${SECRET_SYNC_MAX_SECRETS} for a single sync. Point the sync at a narrower secret path, or split it into several syncs.`,
+    shouldRetry: false
+  });
+};
+
+export const resolveSyncFolders = async ({
```
lobocv: the name of this func is not clear, maybe `getSyncedFolders`? Is seems like this would produce duplicate folders? Or does recursivelyGetSecretPaths return one path per folder? I guess i'm getting confused because of using both secret path and folder. Are they the same?

backend/src/services/secret-sync/secret-sync-schemas.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  test("accepts recursive with the overwrite behavior", () => {
+    const result = schema.safeParse({
+      initialSyncBehavior: SecretSyncInitialSyncBehavior.OverwriteDestination,
+      recursive: true
+    });
+
+    expect(result.success).toBe(true);
+  });
+
+  test("rejects recursive combined with importing from the destination", () => {
```
lobocv, line 24: Can we leave a comment explaining why this is rejected?

backend/src/services/secret-sync/secret-sync-schemas.ts
```diff
@@ -9,6 +9,15 @@ import { SecretSync, SecretSyncInitialSyncBehavior } from "@app/services/secret-
 import { SECRET_SYNC_CONNECTION_MAP, SECRET_SYNC_NAME_MAP } from "@app/services/secret-sync/secret-sync-maps";
 import { TSyncOptionsConfig } from "@app/services/secret-sync/secret-sync-types";
 
+const RECURSIVE_SYNC_REFINEMENT = {
+  path: ["recursive"],
+  message:
+    "A sync that includes subfolders cannot also import existing secrets from the destination, because there is no single folder to import them into. Turn off subfolders, or set the first sync to overwrite the destination."
```
lobocv, line 17: could we not just create the folders in infisical when we import them from a destination? We can handle this in the next PR when we add support for saving secrets into heirarchical paths in the destination (for destinations that support it). For those that do not, we don't want to import when recursive it set true because it would squash everything to one directory (ie the secret sync root) and mess up the user's folder structure of secrets.

backend/src/services/secret-sync/secret-sync-schemas.ts
```diff
@@ -63,16 +72,28 @@ const BaseSyncOptionsSchema = <T extends AnyZodObject | undefined = undefined>({
...
+  >;
+
+  // `merge` is supplied exactly when `T` is a ZodObject, which is the condition this type branches
+  // on, and the compiler cannot correlate a runtime ternary with a type parameter. Passing an
+  // explicit type argument that disagrees with the `merge` argument would make the assertion false.
+  const refinedSchema = (
+    merge
+      ? baseSchema.merge(merge).refine(isRecursiveCombinationAllowed, RECURSIVE_SYNC_REFINEMENT)
+      : baseSchema.refine(isRecursiveCombinationAllowed, RECURSIVE_SYNC_REFINEMENT)
+  ) as TRefinedSchema;
```
lobocv: What is this comment saying? is there a way we can make this clearer? what does refinedSchema represent and why do we need it?

backend/src/services/secret-sync/secret-sync-schemas.ts
```diff
@@ -149,5 +170,5 @@ export const GenericUpdateSecretSyncFieldsSchema = <T extends AnyZodObject | und
       .optional()
       .describe(SecretSyncs.UPDATE(destination).secretPath),
     isAutoSyncEnabled: z.boolean().optional().describe(SecretSyncs.UPDATE(destination).isAutoSyncEnabled),
-    syncOptions: BaseSyncOptionsSchema({ destination, syncOptionsConfig, merge, isUpdateSchema: true })
+    syncOptions: BaseSyncOptionsSchema({ destination, syncOptionsConfig, merge, isUpdateSchema: true }).optional()
```
lobocv, line 184: why did we make this optional?

backend/src/services/secret-sync/secret-sync-payload.test.ts
```diff
@@ -112,6 +112,25 @@ describe("createSecretSyncPayload", () => {
       { key: "DB_URL", paths: ["/backend", "/backend/api"] }
     ]);
   });
+
+  test("dedupeConflicts() returns a payload whose flatten() no longer throws", () => {
+    const payload = createSecretSyncPayload([secret("DB_URL", "/backend"), secret("DB_URL", "/backend/api")], {
+      environment: "dev"
+    });
+
```
lobocv, line 120: test that this does throw before we check it doesn't throw

backend/src/services/secret-sync/secret-sync-queue.ts
```diff
@@ -843,7 +839,9 @@ export const secretSyncQueueFactory = ({
             credentials
           }
         } as TSecretSyncWithCredentials,
-        payload,
+        // A sync that has drifted into a duplicate-name state must still be removable: dedupe
+        // before any provider's own flatten() call gets a chance to throw on it.
```
lobocv: you can just say that when we remove secrets. we dont really care about duplicates when removing secrets because duplicates would still map to the secret we are removing.

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
```diff
@@ -48,7 +47,11 @@ export const assertWithinSecretLimit = (count: number) => {
   });
 };
 
-export const resolveSyncFolders = async ({
+// A secret path is just a folder's location written as a filesystem-style string, so "folder"
+// and "path" name the same node here; each folder has exactly one path and each path names
+// exactly one folder, so this never returns the same folder twice. Non-recursive: just the
+// source folder. Recursive: the source folder plus every folder beneath it.
```
lobocv: remove this comment

backend/src/services/secret-sync/secret-sync-schemas.ts
```diff
@@ -9,6 +9,10 @@ import { SecretSync, SecretSyncInitialSyncBehavior } from "@app/services/secret-
 import { SECRET_SYNC_CONNECTION_MAP, SECRET_SYNC_NAME_MAP } from "@app/services/secret-sync/secret-sync-maps";
 import { TSyncOptionsConfig } from "@app/services/secret-sync/secret-sync-types";
 
+// Importing would need to write each secret back into the subfolder it came from, which we don't
+// support yet (hierarchical writes to destinations that can represent folders are planned for a
+// later PR). Until then, allowing this combination would silently squash every subfolder's
+// secrets into the sync's root folder, restructuring the user's existing secrets tree.
```
lobocv: remove comment on "later PR".  Just say we don't allow initial sync import with recursive because it can change the shape of how secrets are organized on infisical, which will then influence the future syncs.

backend/src/services/secret-sync/secret-sync-schemas.ts
```diff
@@ -170,5 +177,10 @@ export const GenericUpdateSecretSyncFieldsSchema = <T extends AnyZodObject | und
       .optional()
       .describe(SecretSyncs.UPDATE(destination).secretPath),
     isAutoSyncEnabled: z.boolean().optional().describe(SecretSyncs.UPDATE(destination).isAutoSyncEnabled),
+    // Optional like every other field here, so an update can touch just the name or description
+    // without resending sync options. This can't be used to sneak past
+    // isRecursiveCombinationAllowed: initialSyncBehavior inside BaseSyncOptionsSchema stays
+    // required, so a request that includes syncOptions at all must still submit a complete,
+    // valid one.
```
lobocv: remove this comment

backend/src/lib/api-docs/constants.ts
```diff
@@ -3144,7 +3144,8 @@ export const SecretSyncs = {
     return {
       initialSyncBehavior: `Specify how Infisical should resolve the initial sync to the ${destinationName} destination.`,
       keySchema: `Specify the format to use for structuring secret keys in the ${destinationName} destination.`,
-      disableSecretDeletion: `Enable this flag to prevent removal of secrets from the ${destinationName} destination when syncing.`
+      disableSecretDeletion: `Enable this flag to prevent removal of secrets from the ${destinationName} destination when syncing.`,
+      recursive: `Whether to sync secrets from folders beneath the source path as well.`
```
varonix0: I think we can rename this to `includeSubFolders` to make it more user-friendly on the api side

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -121,6 +142,220 @@ export const secretSyncServiceFactory = ({
...
+      actorId: actor.id,
+      actorAuthMethod: actor.authMethod,
+      actorOrgId: actor.orgId,
+      actionProjectType: ActionProjectType.SecretManager,
+      projectId
+    });
+
+    ForbiddenError.from(projectPermission).throwUnlessCan(
+      ProjectPermissionSecretSyncActions.Read,
+      ProjectPermissionSub.SecretSyncs
```
varonix0: q: looks like we're asserting on the secret syncs read, but not the creation of secret syncs. should we also be checking for if the user is able to create secret syncs here?

also fyi this will only work if the user has unconditional permission on secret syncs. if its scoped to a path this won't work

backend/src/services/secret-sync/external-infisical/external-infisical-sync-fns.ts
```diff
@@ -143,8 +144,11 @@ const batchDeleteSecrets = async (
 };
 
 export const ExternalInfisicalSyncFns = {
-  syncSecrets: async (secretSync: TExternalInfisicalSyncWithCredentials, secretMap: TSecretMap) =>
+  syncSecrets: async (secretSync: TExternalInfisicalSyncWithCredentials, payload: TSecretSyncPayload) =>
     withExternalInfisicalErrorHandling(async () => {
+      // Key schema must never be applied for Infisical-to-Infisical syncs, or the prefixed key
+      // triggers another sync cycle on the remote project.
```
varonix0: why is that? I don't think we disallowed the use of key schemas on the infiscial to infsical sync historically

backend/src/services/secret-sync/secret-sync-payload.ts
```diff
@@ -0,0 +1,165 @@
...
+  return grouped;
+};
+
+// Returns every destination-key collision, unordered and untruncated: the caller decides how much
+// of this to show. findConflicts() below is the public way to reach this on a built payload; the
+// recursive-conflicts preview endpoint calls that method, not this function directly, so
+// keySchema never has to be a field the payload object exposes to its own callers.
+export const findFlattenConflicts = (
+  payload: { secrets: TSecretPayload[]; environment: string; keySchema?: string },
+  { applySchema = true }: { applySchema?: boolean } = {}
```
varonix0: I think `applySchema` defaulting to true meaning it gets applied if its configured, but false meaning it gets forcefully not-set even if its defined, can be confusing for future readers.

I think it might be better to call this `forceDisableKeySchema` and have it default to false. behavior is the same but easier to reason about when reading

backend/src/services/secret-sync/secret-sync-payload.ts
```diff
@@ -0,0 +1,165 @@
...
+): TSecretSyncPayload => ({
+  secrets,
+  environment,
+  // A destination whose sync targets can't carry the configured key schema (eg a many-to-one
+  // JSON body whose fields are app-facing variable names, or an Infisical-to-Infisical sync,
+  // where a schema-renamed key would look like a new secret and retrigger a sync cycle) calls
+  // flatten({ applySchema: false }) to get the raw-key view instead. Either way flatten() still
+  // groups by the resulting destination key and rejects collisions, so duplicate-name detection
+  // never depends on whether the schema was applied.
+  flatten: ({ applySchema = true }: { applySchema?: boolean } = {}) => {
```
varonix0: same thing applies everywhere we call it `applySchema` like above

backend/src/services/secret-sync/secret-sync-queue.ts
```diff
@@ -275,13 +278,12 @@ export const secretSyncQueueFactory = ({
...
       });
 
-    const secretMap: TSecretMap = {};
-
     const { decryptor: secretManagerDecryptor } = await kmsService.createCipherPairWithDataKey({
       type: KmsDataKey.SecretManager,
       projectId
     });
     const actorOrgId = secretSync.connection.orgId;
+    const syncOptions = secretSync.syncOptions as { recursive?: boolean; keySchema?: string } | undefined;
```
varonix0: not a huge fan of casting in this case. can't we parse it from the base secret sync options schema for added type safety? (would also need to remember to update this if we rename it to `includeSubFolders` like mentioned in the first comment. if we parsed from the base schema we would see type errors if we forgot to update it)

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
```diff
@@ -0,0 +1,225 @@
...
+  deps: {
+    folderDAL: Pick<TSecretFolderDALFactory, "find" | "findByManySecretPath">;
+    projectEnvDAL: Pick<TProjectEnvDALFactory, "findOne">;
+    secretV2BridgeDAL: Pick<TSecretV2BridgeDALFactory, "findByFolderId" | "findByFolderIds" | "find">;
+    secretImportDAL: Pick<TSecretImportDALFactory, "findByFolderIds" | "findByIds">;
+    expandSecretReferences: TExpandSecretReferences;
+    decryptSecretValue: (value?: Buffer | null) => string;
+    fnSecretsV2FromImportsDeps: TFnSecretsV2FromImportsDeps;
+  },
+  args: {
```
varonix0: nits:

1. I think all the args should be passed in before the dependencies
2. instead of passing options like `keySchema` and `recursive`, can't we just pass the actual secret sync and do any computation inside this function? makes the call site cleaner. the args would just be the secret sync itself in that case

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
```diff
@@ -0,0 +1,225 @@
...
+    projectId: string;
+    environment: string;
+    sourcePath: string;
+    sourceFolderId: string;
+    recursive: boolean;
+    keySchema?: string;
+    includeImports: boolean;
+  }
+): Promise<TSecretSyncPayload> => {
```
varonix0, line 125: We don't need to do `: Promise<TSecretSyncPayload>`, because its implicit since what the function returns in the end will always have that shape and typescript will automatically infer it

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
```diff
@@ -0,0 +1,225 @@
...
+    projectId,
+    environment,
+    sourcePath,
+    sourceFolderId,
+    recursive
+  });
+
+  const pathByFolderId = new Map(folders.map(({ folderId, path }) => [folderId, path]));
+
+  // findByFolderIds joins tags, metadata, rotation, honey token, reminder, recipients and users
```
varonix0: I would opt for readability over optimization here; the overhead we get from calling `findByFolderIds` with a single input folder is very tiny, and we could get rid of 3 lines of comments and a ternary operator if we just called it directly

backend/src/services/secret-sync/secret-sync-schemas.ts
```diff
@@ -9,6 +9,50 @@ import { SecretSync, SecretSyncInitialSyncBehavior } from "@app/services/secret-
 import { SECRET_SYNC_CONNECTION_MAP, SECRET_SYNC_NAME_MAP } from "@app/services/secret-sync/secret-sync-maps";
 import { TSyncOptionsConfig } from "@app/services/secret-sync/secret-sync-types";
 
+// We don't allow initial sync import combined with recursive, because it can change the shape of
+// how secrets are organized in Infisical, which would then influence future syncs.
+const RECURSIVE_SYNC_REFINEMENT = {
```
varonix0: general note; I think this is fine for now but I can envision that people will want import support, but we can tackle that in a follow up down the road

backend/src/services/secret-sync/secret-sync-schemas.ts
```diff
@@ -149,5 +181,5 @@ export const GenericUpdateSecretSyncFieldsSchema = <T extends AnyZodObject | und
       .optional()
       .describe(SecretSyncs.UPDATE(destination).secretPath),
     isAutoSyncEnabled: z.boolean().optional().describe(SecretSyncs.UPDATE(destination).isAutoSyncEnabled),
-    syncOptions: BaseSyncOptionsSchema({ destination, syncOptionsConfig, merge, isUpdateSchema: true })
+    syncOptions: BaseSyncOptionsSchema({ destination, syncOptionsConfig, merge, isUpdateSchema: true }).optional()
```
varonix0, line 184: why is the sync options now optional?

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -59,17 +67,27 @@ import {
 } from "./secret-sync-maps";
 import { TSecretSyncQueueFactory } from "./secret-sync-queue";
 
+// secretSyncDAL rows carry syncOptions as an untyped jsonb column, not the destination-specific
+// shape TSecretSync narrows it to elsewhere, so reading a field back out needs a cast.
+type TRawSyncOptions = { recursive?: boolean; keySchema?: string };
```
varonix0: I think we should try to parse from the base secret sync schema instead of overriding the type system

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -121,6 +142,220 @@ export const secretSyncServiceFactory = ({
...
+              ? `You do not have permission to read secrets at path "${path}" in environment "${environment}".`
+              : `You do not have permission to read secrets at path "${path}" in environment "${environment}". This sync includes subfolders, so it reads every folder beneath "${sourcePath}".`
+        });
+      }
+    }
+  };
+
+  // A SecretSyncError is written for the end user already (see secret-sync-payload.ts); every
+  // caller that can surface one directly just needs it translated to the house error envelope.
+  const $rethrowSecretSyncErrorsAsBadRequest = async <T>(fn: () => T | Promise<T>): Promise<T> => {
```
varonix0: i'd get rid of this, super hard to read. better to just wrap the actual call sites in a try/catch

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -121,6 +142,220 @@ export const secretSyncServiceFactory = ({
...
+    keySchema
+  }: {
+    projectId: string;
+    actorOrgId: string;
+    environment: string;
+    sourcePath: string;
+    sourceFolderId: string;
+    keySchema?: string;
+  }) =>
+    $rethrowSecretSyncErrorsAsBadRequest(async () => {
```
varonix0: very hard to read, also nit I would avoid anonymous arrow functions for anything that is not a one-liner. better to use block returns

backend/src/services/secret-sync/secret-sync-service.ts
varonix0: this check should happen later i think

1. it checks for the `read` permission without conditions, so if a user has secret sync permission only scoped to certain paths in a project, this will never work
2. I think we are missing the check to see if the user is allowed to create or update secret syncs? if they're accessing this endpoint, it would be because they're trying to update or create a secret sync

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -612,29 +863,61 @@ export const secretSyncServiceFactory = ({
...
-      folderId = newFolder.id;
+      await $assertCanReadSyncedFolders(permission, {
+        projectId: secretSync.projectId,
+        environment: updatedEnvironment,
+        sourcePath: updatedSecretPath,
+        sourceFolderId: folderId,
+        recursive: isRecursive
+      });
+
+      if (isRecursive) {
```
varonix0: `const isRecursive = Boolean(requestedSyncOptions?.recursive ?? wasRecursive);`

Wouldn't this also run even if the user is trying to update their secret sync to **disable** recursive mode? I don't think we want to assert that the secrets are flattenable if they're trying to _disable_ recursive mode

frontend/src/components/secret-syncs/forms/SecretSyncSourceFields.tsx
```diff
@@ -97,6 +133,76 @@ const DefaultSecretSyncSourceFields = () => {
...
         )}
       />
+      <Controller
+        control={control}
+        name="syncOptions.recursive"
+        render={({ field: { value, onChange }, fieldState: { error } }) => (
+          <Field>
+            <Field orientation="horizontal">
+              <FieldContent>
+                <Label htmlFor="recursive">Recursively sync secrets</Label>
```
varonix0: I think we should stray away from using technical terms like "recursively" I think "Include secrets from subfolders" reads more naturally

frontend/src/components/secret-syncs/forms/SecretSyncSourceFields.tsx
```diff
@@ -97,6 +133,76 @@ const DefaultSecretSyncSourceFields = () => {
           </Field>
         )}
       />
+      <Controller
```
varonix0, line 129: Disregard comment placement: the DX here is not great. i wasn't able to actually enable the option until I went into the initial sync step, set it to overwrite destination secrets, and then went back to source to enable it.

I think instead what we should do is allow the user to toggle it on, and when they toggle it on, display a small banner telling the user that when syncing sub-folders, its not possible to import into infisical from the destination, and the only available initial sync option is overwriting the destination

backend/src/services/secret-sync/secret-sync-queue.ts
```diff
@@ -1084,7 +1055,23 @@ export const secretSyncQueueFactory = ({
...
+    const candidateSyncs = await secretSyncDAL.find({
+      $in: { folderId: [...ancestorFolders.map((entry) => entry.id), folder.id] },
+      isAutoSyncEnabled: true
+    });
+
+    // A sync on the changed folder itself always matches, recursive or not. A sync on an
+    // ancestor folder only matches when it is recursive, so a non-recursive sync rooted above
+    // this path is never triggered by a write it was never configured to cover.
+    const secretSyncs = candidateSyncs.filter(
+      (sync) => sync.folderId === folder.id || Boolean((sync.syncOptions as { recursive?: boolean } | null)?.recursive)
```
varonix0: same here, avoid type casting

varonix0 (changes requested):
couple comments; ill need to go deeper on the application testing side of things

backend/src/services/secret-sync/external-infisical/external-infisical-sync-fns.ts
lobocv: I'm guessing it's because inf to inf is a replication, so you'd want the secrets to be exactly the same, but I guess the user's could decide that by not setting the keySchema if they wanted it to be identical. I'll change that

backend/src/services/secret-sync/secret-sync-payload.ts
lobocv: i'm not sure I agree. It reads pretty straight forward to me. "applySchema == true" means apply the keySchema it if it exists, otherwise don't. If there is no keySchema, there's nothing to apply obviously. I find the wording "force" is not the right word to use here and is even more confusing to me. 

If you still disagree with that I can change it, I don't feel that strongly about it

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
lobocv: Good call, agreed we should be passing in the secret sync object in instead of it's separate fields.

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
lobocv: Even if it's not required, I think it greatly helps readability. Now I can look at this function and know exactly what it returns. Otherwise, i need to look through it's body to find the return statement, and in this case, it returns the results of `createSecretSyncPayload`, which I then need to track down, just to find the return type.

backend/src/services/secret-sync/secret-sync-recursive-fns.ts
lobocv: Yes, good catch. Agreed. This probably happened because these methods return different shapes of the Folder they are looking up (which isn't great design).

backend/src/services/secret-sync/secret-sync-schemas.ts
lobocv: Yes the import stuff isn't trivial, it should be a follow up ticket

frontend/src/components/secret-syncs/forms/SecretSyncSourceFields.tsx
lobocv: Sure, that's a good idea

backend/src/lib/api-docs/constants.ts
lobocv: I actually disagree here. `recursive` already matches what we have in the infisical CLI and other places and consistency is important in API design.

It  makes more sense to me than "includeSubFolders". `includeSubFolders` doesn't make it clear that this happens recursively and not just for one level of child folders. When I see `includeSubFolders` I can easily think that it just contains the synced folder's children and not their children.

It's a developer tool so the word `recursive` is pretty well understood. 

CleanShot 2026-09-16 at 8 52 40 AM@2x

backend/src/services/secret-sync/secret-sync-payload.ts
varonix0: I think for the most part it's better to have optional default inputs, default to the unset variant of the variable (an undefined boolean would resolve as false as an example). Having optional parameters that carry the opposite default adds more to think about for the next person to work on this

backend/src/services/secret-sync/secret-sync-queue.ts
```diff
@@ -302,6 +302,14 @@ export const secretSyncQueueFactory = ({
     });
 
     return buildSyncPayload(
+      {
+        projectId,
```
varonix0, line 306: Why aren't we doing something like this instead:


```typescript
return buildSyncPayload(secretSync, { // pass the entire secret sync
   folderDAL,
   projectEnvDAL,
   secretV2BridgeDAL,
   secretImportDAL,
   expandSecretReferences,
   decryptSecretValue,
   fnSecretsV2FromImportsDeps: {
      projectFolderGrantDAL,
      actorOrgId,
      orgDAL,
      kmsService
   }
});
```

backend/src/services/secret-sync/secret-sync-service.ts
```diff
@@ -121,6 +138,221 @@ export const secretSyncServiceFactory = ({
...
+      throw error;
+    }
+  };
+
+  // Lets the frontend check for duplicate-name conflicts while the user is still on the Source
+  // step, before a provider or destination is even chosen: the check is destination-agnostic
+  // (flatten() never uses the destination or the source folder path to build a key), so nothing
+  // downstream can change the answer. Returns every conflict, untruncated, unlike the
+  // BadRequestError $assertSyncedSecretsAreFlattenable throws at create/update time.
+  const findRecursiveConflicts = async (
```
varonix0, line 308: bug: if a user that only has edit access on secret syncs tries to edit a sync, this endpoint will throw a 403 error because we are only checking if the actor has create access (it should be either update or create)

varonix0 (changes requested):
went through application testing and a couple notes:

1. Deleting a sub-folder doesn't trigger a secret sync. You can delete a sub-folder that's covered by a sync and nothing happens. This is partially a problem today in production but less of a problem because on a single folder (folder gets deleted = sync has no base target anymore), but for recursive sub-folders it feels bugged. let me know what you think!

2. sync options are now optional (already mentioned earlier) but didn't see a response to that comment



other than that this is pretty great! a few comments remain open / unanswered, and I pushed some smaller changes for zod input validation / UX

backend/src/services/secret-sync/secret-sync-queue.ts
lobocv: This is called before a secret sync is created in two areas: 

1. When turning on the recursive flag in the UI in order to show the conflict error early
2. Before creating the secret in createSecretSync so that we don't create a sync job that will immediately error out

I'm not sure i'd like a solution that fabricates a fake secret sync just to pass in these args as a package. I think a better solution would be to group common fields within the secret sync into their own objects. For example, there could be a `source` object inside secret sync that contains the env, project, source path, folder id. 

Passing in the full secret sync IMO would be wrong because this function doesn't care about a lot of  it's fields such as app connection, destination, destinationConf. And it also makes us require that full object just to call this function.

I do see the issue and I think i'll refactor the secret sync object to be less flat. That shoudl help aleviate what I think the underlying concern you have is: readability and preventing mistakes of passing in parameters individually

backend/src/services/secret-sync/secret-sync-schemas.ts
lobocv: It's optional because the recursive parameter is in the syncOptions and we need a way to update it in the PATCH endpoint. I'm not sure it was a good decision to make the PATCH endpoint accept syncOptions as a single parameter. It would have been simpler to make each option inside it's own field in the PATCH but that's a beaking change. 
Now we have to deal with the complexity of checking which fields inside the syncOptions have changed when it is passed in.

I'm open to better ideas if you have any

backend/src/services/secret-sync/secret-sync-queue.ts
lobocv: Sigh.. it's not a simple refactor because we are inferring the API shape from the domain objects using zod `infer`.. Making the change is possible but it's risky and shouldn't be added to this PR. I'm going to make notes on this because this is a bad coupling between the REST API and the domain layer...

backend/src/services/secret-sync/secret-sync-service.ts
lobocv: been racking my brain on this one... i think github is doing something strange here because i can't see that code anymore. Right now it should only run the flat assertion if the recursion is on, which is what we want because we want to catch any issues with the job early.

backend/src/services/secret-sync/secret-sync-payload.ts
lobocv: I'll make the parameter name so that the default is false "disableKeySchema"

backend/src/lib/api-docs/constants.ts
varonix0: tldr; we are doing includeSubFolders

varonix0:
@northflank rebuild the preview please?

varonix0 (approved):

PR #8089, opened 2026-09-11
fix(queue): log job failures and make PINO_LOG_LEVEL work

lobocv:
## Context

Two logging gaps found while debugging a flaky test suite. Neither is specific to that suite, and both hide failures in production.

### Queue job failures were counted, never logged

`worker.on("failed")` recorded OpenTelemetry counters and nothing else. A counter carries no message, no stack and no job identity, so an uncaught exception in a queue handler left no usable trace. Several handlers await work with no `try`/`catch` (eg the `SecretSync` handler in `secret-queue.ts`), so this is reachable rather than theoretical.

It was hiding real volume. A single run of the secret sync end-to-end suite logs around 150 job failures that nobody could see, mostly background jobs firing against projects a finished test had already deleted. One of them is a genuine null dereference in `secret-integration-pull` (`Cannot read properties of undefined (reading 'orgId')`) that is worth chasing separately.

### PINO_LOG_LEVEL below info did nothing, anywhere

`initLogger` built its transport target with a hardcoded `level: "info"` while the logger instance read `PINO_LOG_LEVEL`. A pino transport filters independently of the logger, so debug records were discarded at the transport no matter what the variable said. Setting `PINO_LOG_LEVEL=debug` has been a no-op product-wide.

## What changed

- Job failures log at `error`, with queue, job name, job id, attempt and whether attempts are exhausted.
- Enqueued, picked up and completed log at `debug`, so a run being debugged can tell "never queued" from "queued but never ran" from "ran and hung".
- Stalling logs at `warn`. A worker that stopped heartbeating is rare and always worth seeing, so it does not belong behind debug.
- The transport target now follows `PINO_LOG_LEVEL`.

Two details worth a reviewer's eye:

- All four calls pass structured fields with a static message rather than an interpolated string. Pino swaps a disabled level for a noop, but JavaScript still evaluates the arguments at the call site, so building a string per job completion would cost work in production for output nobody reads.
- The enqueue line is guarded on the queue existing. `q` is undefined only where `QUEUE_WORKERS_ENABLED` is false, and a pod in that mode neither produces nor consumes, so a no-op there is expected rather than a dropped job worth reporting.

## Screenshots

Not applicable.

## Steps to verify the change

```
cd backend && npm run type:check
```

Then, on a branch that has the end-to-end harness, `PINO_LOG_LEVEL=debug make test-api-e2e SPEC=secret-sync` shows the lifecycle lines and the previously invisible failures.

## Type

- [x] Fix
- [x] Improvement

## Checklist

- [x] Tested locally

### DIFF
--- a/backend/src/lib/logger/logger.ts
+++ b/backend/src/lib/logger/logger.ts
@@ -106,13 +106,15 @@ const extractOrgId = () => {
   }
 };
 
+// A transport target filters independently of the logger instance, so both have to be set from
+// the same place. Pinning the target to info while the instance read the variable is what made
+// PINO_LOG_LEVEL=debug do nothing.
+const getLogLevel = () => process.env.PINO_LOG_LEVEL || "info";
+
 export const initLogger = () => {
   const targets: pino.TransportMultiOptions["targets"][number][] = [
     {
-      // Follows the same env var as the logger instance below. A transport target filters
-      // independently of the logger, so pinning this to info silently discarded every debug
-      // record no matter what PINO_LOG_LEVEL said.
-      level: process.env.PINO_LOG_LEVEL || "info",
+      level: getLogLevel(),
       target: "pino/file",
       options: {
         destination: 1,
@@ -154,7 +156,7 @@ export const initLogger = () => {
       mixin(_context, level) {
         return { severity: logLevelToSeverityLookup[level] || logLevelToSeverityLookup["30"] };
       },
-      level: process.env.PINO_LOG_LEVEL || "info",
+      level: getLogLevel(),
       formatters: {
         bindings: (bindings) => ({
           pid: bindings.pid,
--- a/backend/src/lib/logger/logger.ts
+++ b/backend/src/lib/logger/logger.ts
@@ -106,9 +106,6 @@ const extractOrgId = () => {
   }
 };
 
-// A transport target filters independently of the logger instance, so both have to be set from
-// the same place. Pinning the target to info while the instance read the variable is what made
-// PINO_LOG_LEVEL=debug do nothing.
 const getLogLevel = () => process.env.PINO_LOG_LEVEL || "info";
 
 export const initLogger = () => {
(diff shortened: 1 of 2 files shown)


backend/src/lib/logger/logger.ts
```diff
@@ -109,7 +109,10 @@ const extractOrgId = () => {
 export const initLogger = () => {
   const targets: pino.TransportMultiOptions["targets"][number][] = [
     {
-      level: "info",
+      // Follows the same env var as the logger instance below. A transport target filters
+      // independently of the logger, so pinning this to info silently discarded every debug
+      // record no matter what PINO_LOG_LEVEL said.
+      level: process.env.PINO_LOG_LEVEL || "info",
```
lobocv: factor this out into a `getLogLevel` function and replace it in both the transport and instance. This will keep them in sync

backend/src/queue/queue-service.ts
```diff
@@ -882,17 +877,34 @@ export const queueServiceFactory = (redisCfg: TRedisConfigKeys): TQueueServiceFa
         "error.type": errorType,
         "attempts.exhausted": attemptsExhausted ? "true" : "false"
       });
+
+      // A handler that throws without catching leaves no other trace: the metrics above carry no
+      // message and no stack, so without this the only evidence is a counter moving.
```
lobocv: remove useless comment

backend/src/lib/logger/logger.ts
lobocv: Done in b428a5e6c5. Extracted `getLogLevel` and both the transport target and the logger instance now read through it, so they cannot drift apart again. Moved the explanation onto the function, since the reason they have to stay in sync is the bug itself.

backend/src/queue/queue-service.ts
lobocv: Removed in b428a5e6c5.

backend/src/lib/logger/logger.ts
```diff
@@ -106,10 +106,15 @@ const extractOrgId = () => {
   }
 };
 
+// A transport target filters independently of the logger instance, so both have to be set from
+// the same place. Pinning the target to info while the instance read the variable is what made
+// PINO_LOG_LEVEL=debug do nothing.
```
lobocv: remove overfit comment

akhilmhdh (dismissed):

akhilmhdh (approved):

PR #8094, opened 2026-09-11
feat: add validation rule unique values across the whole project

mathnogueira:
## Context

Add a new option to prevent reusing values that are already present in the project.

## Screenshots

CleanShot 2026-09-14 at 17 29 40

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -413,14 +421,6 @@ export const secretV2BridgeServiceFactory = ({
       project.secretDetectionIgnoreValues || []
     );
 
-    await secretValidationRuleService.validateSecrets({
```
mathnogueira, line 426: Moved those calls to be inside the transaction so we can use the advisory lock to make sure those checks are atomic. Otherwise, 2 concurrent request calls can write 2 secrets with the same value.

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -745,6 +747,20 @@ export const secretV2BridgeServiceFactory = ({
...
+            projectId,
+            environment,
+            envId: folder.envId,
+            secretPath,
+            secrets: [{ key: finalKey, value: secretValue, secretId }],
+            canAccessLocation: $canDescribeSecretAt(permission)
+          },
+          tx
+        );
+      }
```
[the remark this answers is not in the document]
mathnogueira: Not an issue.

backend/src/services/secret-validation-rule/secret-validation-rule-service.ts
```diff
@@ -372,8 +396,108 @@ export const secretValidationRuleServiceFactory = ({
...
+      generateSecretBlindIndex
+    }: Pick<TValidateSecretsDTO, "projectId" | "environment" | "secretPath" | "secrets" | "canAccessLocation"> & {
+      scope: { envId?: string | null; secretPath: string };
+      generateSecretBlindIndex: (value: Buffer) => Promise<string>;
+    },
+    tx?: Knex
+  ): Promise<Record<string, TDuplicateSecret>> => {
+    const candidates = secrets.filter(
+      (secret): secret is typeof secret & { value: string } =>
+        secret.value !== undefined && !containsSecretReference(secret.value)
```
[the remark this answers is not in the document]
mathnogueira: That's expected, we expect that secret can be referenced by multiple secrets, even if using a template such as: `static-${dev.SECRET}`

mathnogueira:
@greptile review again

backend/src/services/secret-validation-rule/secret-validation-rule-service.ts
```diff
@@ -372,8 +396,109 @@ export const secretValidationRuleServiceFactory = ({
...
+    );
```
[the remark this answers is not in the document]
mathnogueira: This is expected, if we include secret references in the duplicate detection, it does not make any sense to support secret references as you cannot use them to prevent duplications.

mathnogueira:
@greptile review again

mathnogueira:
@varonix0 didn't touch the UI of the other rules in this PR as it's kind of large already.

backend/src/ee/services/secret-approval-request/secret-approval-request-service.ts
```diff
@@ -206,6 +207,14 @@ export const secretApprovalRequestServiceFactory = ({
   queueService,
   secretValidationRuleService
 }: TSecretApprovalRequestServiceFactoryDep) => {
+  // A duplicate-value violation may only name where the value is already used when the writer can read there.
+  const $canDescribeSecretAt =
+    (permission: MongoAbility<ProjectPermissionSet>) => (environment: string, secretPath: string) =>
+      permission.can(
+        ProjectPermissionSecretActions.DescribeSecret,
+        subject(ProjectPermissionSub.Secrets, { environment, secretPath })
+      );
```
lobocv: Not related to this PR, but we should think about abstracting away the implementation details of permissioning. Right now this exposes CASL and Mongo to the consumer when all the consumer should care about is "can this actor read this secret". The interface should be some sort of function like "CanDescribeSecret(...)"

lobocv: Oh I see that this is essentially defining a function like that... its just that I think that should be part of the permissioning service / library, not define here

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -2270,6 +2275,18 @@ export const secretV2BridgeServiceFactory = ({
...
 
     const executeBulkInsert = async (tx: Knex) => {
+      await secretValidationRuleService.validateSecrets(
+        {
+          projectId,
+          environment,
+          envId: folder.envId,
+          secretPath,
+          secrets: deduplicatedSecrets.map((s) => ({ key: s.secretKey, value: s.secretValue })),
+          canAccessLocation: $canDescribeSecretAt(permission)
```
lobocv: seems strange to be passing in the permission check here... why is that handled internally by the secretValidationRuleService? Or should it even be? Seems like validation and access are separate things. Maybe we should be checking the access as a separate step from validation

backend/src/services/secret-validation-rule/static-secrets/static-secrets-rule-fns.ts
```diff
@@ -2,10 +2,20 @@ import { CONSTRAINT_LABELS, evaluateConstraints, TConstraintViolation } from "..
 import { ConstraintKind, ConstraintTarget } from "../secret-validation-rule-enums";
 import { TStaticSecretsRuleConfig } from "./static-secrets-rule-types";
 
+// Where a value that another secret already holds was found, once the service has looked it up.
+export type TDuplicateSecret = {
+  key: string;
+  environment: string;
+  secretPath: string;
+  // Set when the writer cannot read the location, so the message has to stay vague.
+  hidden?: boolean;
```
lobocv: same thing.. seems like a code smell to me. In my mind, the permission check should happen separately before we check validation. From the perspective of the validation service, only a caller able to see the secret would call it, or (better yet) the validation service should have it's own set of permission requirements (ability to read the secret), rather than being dependency injected in.

backend/src/services/secret-validation-rule/secret-validation-rule-service.ts
```diff
@@ -372,8 +396,109 @@ export const secretValidationRuleServiceFactory = ({
...
+    if (!candidates.length) return {};
+
+    // Two writes of the same value would each find no duplicate and both land. The lock makes this
+    // check and the write that follows it atomic, so the second one sees the first.
+    if (tx) await tx.raw("SELECT pg_advisory_xact_lock(?)", [PgSqlLock.SecretValueReuseCheck(projectId)]);
```
lobocv: I'm curious, what other solutions have you considered? The advisory lock works but it is a pretty critical design decision of this PR. What were the alternatives?
Adding an advisory lock on a process as important as secret update is a bit scary. Are you sure this is a hard requirement? It is essentially making updates to all secrets within a project be atomic for an edge case.

lobocv (commented):
This is a tricky one, but i have some concerns on the advisory lock. It seems to be the only solution if we want to have the contraint on the path. If we want to drop that, then it's possible to do this with a unique partial index in postgres.

I'm worried about the lock slowing down all writes. This will be even more of a problem for a customer with a lot of secrets. I think the advisory lock is on the whole project, but perhaps we can scope it down to just the exact policy (project, environment, path) so that we're not locking writes to every single secret?

backend/src/services/secret-validation-rule/secret-validation-rule-service.ts
mathnogueira: Well, this lock only affects projects that have this validation rule in place, this lock will not prevent other orgs from updating secrets. 

Also, it's only acquired in the validation path of the unique value across the whole project, so it will not be used if the project doesn't have the validation rule.

> I'm curious, what other solutions have you considered?

1. A redis lock: might be an option, we use it for creating folders
2. An advisory lock with the blind index as part of the lock name: problematic, we can exhaust the number of locks per transaction (64 by default)
3. Not handle this concurrency issue: very likely we won't see this issue in the wild


I see your point and I think it might make sense not to handle this, especially if it can slowdown the rest of the updates of a project.

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
mathnogueira: > why is that handled internally by the secretValidationRuleService? 

The main reason we check the permission is not to block the write or anything, that is done before this validation. We only check it here so during the validation we decide which message to show:

* if has permission to describe secrets from scope (path/env) A: detailed message displaying the secret that has the same value in scope A
* if user doesn't have permission to describe secrets in scope A, we should not show the location of conflicting secrets in that scope, as it would allow someone to guess secrets

backend/src/services/secret-validation-rule/static-secrets/static-secrets-rule-fns.ts
mathnogueira: See the explanation above

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
lobocv: That seems like the consumers' concern, not the validators. I'd suggest a cleaner separation of concerns. The validator should return all the conflicting secrets and the consumer should decide how to render it in an error message.

backend/src/services/secret-validation-rule/secret-validation-rule-service.ts
lobocv: I think always locking at the project level may be overkill. Doesn't the validation rule have a more granular scope (ie env or folder)? If we only want unique values in a specific (project, environment, folder) then why lock the entire project?  

Perhaps we should add this as a separate follow up so we have some more time to think about it? Is it really a problem if a concurrent write uses the same value? Whats the likelyhood of that even happening? 

I'm worried about this causing issues when there are lots of secret syncs happening or other platform related updates to secrets. It's in the hot-path

mathnogueira: > Perhaps we should add this as a separate follow up so we have some more time to think about it? Is it really a problem if a concurrent write uses the same value? Whats the likelyhood of that even happening?

Yes, this only happens if 2 concurrent write requests happen to write the same conflicting value. TBH, pretty low chance, but not impossible.

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -423,13 +451,16 @@ export const secretV2BridgeServiceFactory = ({
...
     );
 
-    await secretValidationRuleService.validateSecrets({
-      projectId,
-      environment,
-      envId: folder.envId,
-      secretPath,
-      secrets: [{ key: inputSecret.secretName, value: inputSecret.secretValue }]
-    });
+    await $validateSecrets(
```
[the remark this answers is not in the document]
mathnogueira: Expected for now.

backend/e2e-test/routes/v1/secret-validation-rules.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    beforeAll(async () => {
+      ({ projectId, authToken, cleanup } = await createIsolatedOrgAndProject("secret-validation-rules-e2e"));
+
+      await createStaticSecretsValidationRule({
+        projectId,
+        name: "no-duplicate-values-in-prod",
+        environment: PROD_ENV,
+        secretPath: "/",
+        valueConstraints: { reusePrevention: { otherSecretsInScope: true } },
```
lobocv: This API reads confusing to me. It does make sense but it made me think hard. 
Just my opinion so take it with a grain of salt:

- reusePrevention --> "reuse" and "otherSecretsInScope" --> "uniqueWithinScope"

then the API reads more clearly IMO:

`valueConstraints: { reuse: { uniqueWithinScope: true } },`

reads to me as

> "the value constraint on resuse is that secrets must be unique within scope"

backend/e2e-test/routes/v1/secret-validation-rules.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+  // validation rule: only writes inside the rule's environment and path are held against it.
+  describe("Preventing reuse of a value another secret holds", () => {
+    let projectId: string;
+    let authToken: string;
+    let cleanup: () => Promise<void>;
+
+    beforeAll(async () => {
+      ({ projectId, authToken, cleanup } = await createIsolatedOrgAndProject("secret-validation-rules-e2e"));
+
+      await createStaticSecretsValidationRule({
```
lobocv, line 20: just my preference, but given that the secretValidationRule is central to this test, I think each test case should define it so that each test is self contained and readable. Right now as I look through the tests, I need to go back up here to see how the validation rule was configured. 

This also forces the tests to use a single validation rule across all tests, which over time as we test edge cases, can accumulate test-specific behaviour and cause a coupling that makes tests break, flakey or hard to refactor. 

I would suggest calling this inside each test so that each so that each test has full control of how it creates the validation rules.

For non-central entities such as orgs and projects, I think it's fine to have a common setup shared among tests, but we should still be careful if those shared entities need specific tuning between tests

backend/e2e-test/routes/v1/secret-validation-rules.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      await createSecretV2({
+        workspaceId: projectId,
+        environmentSlug: PROD_ENV,
+        secretPath: "/",
+        key: "ORIGINAL",
+        value: "duplicated-value",
+        authToken
+      });
+
+      const res = await testServer.inject({
```
lobocv: what if we add an option function argument to createSecretV2 where we define the expectation. If not provided, it can do the current expectation (200, no error). This way we can re-use the createSecretV2 and essentially pass in a closure for the `expect` calls you have below. 
I think it would improve readability. Checking that a secret cannot be created will be a common testing pattern I'd assume. It would be nice to streamline it.

The func arg should accept the `res` and the caller should fill in any expectations

backend/e2e-test/routes/v1/secret-validation-rules.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      await createSecretV2({
+        workspaceId: projectId,
+        environmentSlug: PROD_ENV,
+        secretPath: "/",
+        key: "UPDATE_HOLDER",
+        value: "value-held-by-another-secret",
+        authToken
+      });
+
+      const res = await testServer.inject({
```
lobocv: same for this, do we have have a testUtil helper for updating secrets? We should probably add one as it'll be very common practice in tests and improve readability

backend/src/services/secret-v2-bridge/secret-v2-bridge-service.ts
```diff
@@ -423,13 +451,16 @@ export const secretV2BridgeServiceFactory = ({
...
+      {
+        projectId,
+        environment,
+        envId: folder.envId,
+        secretPath,
+        secrets: [{ key: inputSecret.secretName, value: inputSecret.secretValue }]
+      },
+      permission
+    );
```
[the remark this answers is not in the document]
mathnogueira: expected

backend/src/services/secret-validation-rule/secret-validation-rule-service.ts
```diff
@@ -372,6 +395,87 @@ export const secretValidationRuleServiceFactory = ({
...
+    // Two secrets in one request can collide with each other before either exists in the database.
+    const seenInBatch = new Map<string, TDuplicateSecret>();
+    candidates.forEach((secret, idx) => {
+      const seen = seenInBatch.get(blindIndexes[idx]);
+      if (seen) duplicates[secret.key] = seen;
+      else seenInBatch.set(blindIndexes[idx], { key: secret.key, environment, secretPath });
+    });
+
+    const excludedSecretIds = candidates.map((secret) => secret.secretId).filter(Boolean) as string[];
+    const existing = await secretDAL.findExistingSecretsByBlindIndexes(
```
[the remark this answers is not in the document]
mathnogueira: expected

lobocv (approved):
API looks great and I love the improvements to the test utils. Great work @mathnogueira

PR #8102, opened 2026-09-12
chore(auth): revert grouped input autofill styling and signup form restructure

scott-ray-wilson:
## Context

Reverts Infisical/infisical#8084.

That PR made two changes:

- `frontend/src/index.css` gained an autofill override for grouped inputs (`input[data-slot="input-group-control"]:-webkit-autofill`) that clipped the native autofill background to the text and dropped the inset shadow, so the opaque rectangle no longer covered the input group's border.
- `UserInfoStep` wrapped the signup credential fields in a native `<form>` with `onSubmit`, and folded the hidden username input and the invite email display into a single `autocomplete="username"` field.

This reverts both. Grouped inputs fall back to the global autofill rule again, and the signup step submits through the button's `onClick={handleSubmit(onSubmit)}` rather than a native form, so Enter-to-submit from the credential fields goes away with it. The hidden username input and the disabled invite email field are separate elements again.

`frontend/src/components/auth/UserInfoStep.tsx` and `frontend/src/index.css` return to their state on main before #8084. No other files change.

## Steps to verify the change

1. Sign up with a new email and complete verification to reach the password step.
2. Enter a name, organization name, and password, then submit. Confirm validation runs and the account is created once.
3. Repeat through an invite link. Confirm the email appears in a disabled field and the organization field is absent.
4. Autofill a saved password in the field. Confirm the styling matches the pre-#8084 behavior, with the global inset-shadow autofill background back.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [x] Chore

## Checklist

- [ ] Tested locally
- [x] Updated docs (if needed): not needed, the reverted change had no docs.
- [x] Updated CLAUDE.md files (if needed): not needed.

### DIFF
--- a/frontend/src/components/auth/UserInfoStep.tsx
+++ b/frontend/src/components/auth/UserInfoStep.tsx
@@ -162,11 +162,7 @@ export default function UserInfoStep({
   };
 
   return (
-    <form
-      className="mx-auto flex w-full flex-col items-center justify-center"
-      onSubmit={handleSubmit(onSubmit)}
-      noValidate
-    >
+    <div className="mx-auto flex w-full flex-col items-center justify-center">
       <AuthPagePanel>
         <CardHeader className={isInvite ? "mb-6 gap-2" : "mb-4 gap-2"}>
           <CardTitle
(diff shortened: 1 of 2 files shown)


andrewhuhh (approved):
ltgm

andrewhuhh (approved):
lgtm

PR #8105, opened 2026-09-12
fix: avoid lockout when removing a lost factor and honor org enforcement from any org

Thiago-AS:
## Context

Removing an authenticator app or passkey required proving the factor being removed, which locked out anyone who lost the device. The step-up also fired for users with MFA disabled and no enforcing org, where it protects nothing, and pending org invitations counted as memberships when deciding whether MFA can be disabled.

## Validation

Setup: one user, two orgs. Org A does not enforce MFA. Org B enforces MFA with passkey as the method.

1. Lost factor fallback. In Org A, with no enforcing org yet, enable MFA with an authenticator app and add one passkey. Remove the authenticator app. Expect a passkey challenge, not an authenticator code. Re-add the authenticator, then remove the passkey. Expect an authenticator challenge.
2. Non-last passkey. Add two passkeys. Remove one. Expect the normal challenge for your current method. Remove the second. Expect the fallback to another factor.
3. Org enforcement wins. Join Org B with passkey enforcement. From Org A, try to remove your last passkey. Expect a passkey challenge even though Org A does not enforce anything. Repeat from Org B. Same result.
4. Recovery path. Log out, log in with a recovery code, then remove the passkey right away. Expect no challenge.
5. No pointless challenge. With a user in no enforcing org and MFA disabled, add and remove a passkey or authenticator. Expect no MFA challenge.
6. Pending invite. Invite a user to an enforcing org but do not accept. In their existing org, disable MFA. Expect it to succeed. Accept the invite, then try again. Expect it to be blocked.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally

### DIFF
--- a/backend/src/server/routes/mfa-step-up-fns.ts
+++ b/backend/src/server/routes/mfa-step-up-fns.ts
@@ -36,6 +36,8 @@ export const getStepUpSessionId = (req: FastifyRequest): string => {
  * enforced method, or the user's preference otherwise). Pass `mfaMethod` to override
  * this when the action dictates the method independently of org enforcement — e.g.
  * enabling MFA challenges the factor being enabled, not a stronger org-enforced one.
+ * Pass `excludeMfaMethod` when the action removes a factor, so the challenge never
+ * demands the factor being removed (which is usually the one the user has lost).
  */
 export const ensureStepUpMfa = async (
   server: FastifyZodProvider,
@@ -46,7 +48,8 @@ export const ensureStepUpMfa = async (
     resourceId,
     mfaSessionId,
     message,
-    mfaMethod: mfaMethodOverride
+    mfaMethod: mfaMethodOverride,
+    excludeMfaMethod
   }: {
     userId: string;
     orgId: string;
@@ -55,6 +58,7 @@ export const ensureStepUpMfa = async (
     mfaSessionId?: string;
     message: string;
     mfaMethod?: MfaMethod;
+    excludeMfaMethod?: MfaMethod;
   }
 ) => {
   if (
@@ -69,18 +73,17 @@ export const ensureStepUpMfa = async (
     return;
   }
 
-  if (
-    resourceId === MfaStepUpResource.MfaManagement &&
-    (await server.services.mfaSession.hasRecentMfaAuth(userId, tokenVersionId))
-  ) {
-    return;
+  if (resourceId === MfaStepUpResource.MfaManagement) {
+    if (await server.services.mfaSession.hasRecentMfaAuth(userId, tokenVersionId)) return;
+    if (!(await server.services.user.isStepUpMfaRequired(userId))) return;
   }
 
   await server.services.mfaSession.enforceStepUpMfaLockout(userId);
 
   const user = await server.services.user.getMe(userId);
 
-  const mfaMethod = mfaMethodOverride ?? (await server.services.user.getStepUpMfaMethod(userId, orgId));
+  const mfaMethod =
+    mfaMethodOverride ?? (await server.services.user.getStepUpMfaMethod(userId, orgId, excludeMfaMethod));
 
   const newMfaSessionId = await server.services.mfaSession.createMfaSession(
     userId,
--- a/backend/src/server/routes/v1/user-router.ts
+++ b/backend/src/server/routes/v1/user-router.ts
@@ -6,7 +6,7 @@ import { getConfig } from "@app/lib/config/env";
 import { logger } from "@app/lib/logger";
 import { authRateLimit, readLimit, writeLimit } from "@app/server/config/rateLimiter";
 import { verifyAuth } from "@app/server/plugins/auth/verify-auth";
-import { AuthMode } from "@app/services/auth/auth-type";
+import { AuthMode, MfaMethod } from "@app/services/auth/auth-type";
 
 import { ensureStepUpMfa, getStepUpSessionId, MfaStepUpResource } from "../mfa-step-up-fns";
 
@@ -240,7 +240,8 @@ export const registerUserRouter = async (server: FastifyZodProvider) => {
         tokenVersionId: getStepUpSessionId(req),
         resourceId: MfaStepUpResource.MfaManagement,
         mfaSessionId: req.query.mfaSessionId,
-        message: "MFA verification is required to remove your authenticator app"
+        message: "MFA verification is required to remove your authenticator app",
+        excludeMfaMethod: MfaMethod.TOTP
       });
       return server.services.totp.deleteUserTotpConfig({
         userId: req.permission.id
@@ -591,7 +592,8 @@ export const registerUserRouter = async (server: FastifyZodProvider) => {
         tokenVersionId: getStepUpSessionId(req),
         resourceId: MfaStepUpResource.MfaManagement,
         mfaSessionId: req.query.mfaSessionId,
-        message: "MFA verification is required to remove a passkey"
+        message: "MFA verification is required to remove a passkey",
+        excludeMfaMethod: MfaMethod.WEBAUTHN
       });
       await server.services.webAuthn.deleteWebAuthnCredential({
         userId: req.permission.id,
--- a/backend/src/services/user/user-service.ts
+++ b/backend/src/services/user/user-service.ts
@@ -1,7 +1,7 @@
 import { ForbiddenError } from "@casl/ability";
 import { Knex } from "knex";
 
-import { AccessScope, OrganizationActionScope, TUsers } from "@app/db/schemas";
+import { AccessScope, OrganizationActionScope, OrgMembershipStatus, TUsers } from "@app/db/schemas";
 import { TEmailDomainDALFactory } from "@app/ee/services/email-domain/email-domain-dal";
 import { EmailDomainStatus } from "@app/ee/services/email-domain/email-domain-types";
 import { OrgPermissionMemberActions, OrgPermissionSubjects } from "@app/ee/services/permission/org-permission";
@@ -215,7 +215,8 @@ export const userServiceFactory = ({
   const hasMfaEnforcingOrg = async (userId: string) => {
     const userOrgMemberships = await membershipUserDAL.find({
       actorUserId: userId,
-      scope: AccessScope.Organization
+      scope: AccessScope.Organization,
+      status: OrgMembershipStatus.Accepted
     });
     if (!userOrgMemberships.length) return false;
 
--- a/backend/src/services/user/user-service.ts
+++ b/backend/src/services/user/user-service.ts
@@ -141,6 +141,17 @@ export const userServiceFactory = ({
     });
   };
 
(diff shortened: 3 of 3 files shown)


Thiago-AS:
@greptile review

backend/src/server/routes/v1/user-router.ts
scott-ray-wilson: ignore placement:

seeing passkey confirmation when trying to remove passkey, not sure if because it's my default but seems counter to what the PR is trying to solve?

https://github.com/user-attachments/assets/71138433-ec93-4e66-8d3c-ae086df4e4fb

scott-ray-wilson, line 1: question: if an org I'm in enforces webpass, should we block removing the last passkey and notify them an org requires it? or not a concern cause they just get asked to add it back next login?

scott-ray-wilson (changes requested):
validated and confirmed PR description flows as working, one comment from me and posting local agent findings separately

backend/src/server/routes/mfa-step-up-fns.ts
```diff
@@ -69,18 +73,17 @@ export const ensureStepUpMfa = async (
...
+    if (!(await server.services.user.isStepUpMfaRequired(userId))) return;
   }
 
   await server.services.mfaSession.enforceStepUpMfaLockout(userId);
 
   const user = await server.services.user.getMe(userId);
 
-  const mfaMethod = mfaMethodOverride ?? (await server.services.user.getStepUpMfaMethod(userId, orgId));
+  const mfaMethod =
+    mfaMethodOverride ?? (await server.services.user.getStepUpMfaMethod(userId, orgId, excludeMfaMethod));
```
scott-ray-wilson: **Blocker:** this makes the challenged method depend on the route, but the accepted session is still scoped only to `MfaManagement`. The early return above checks `userId`, `resourceId`, and `tokenVersionId`; all management routes pass the same resource, and it never checks which method or operation the session authorized. A session minted here with EMAIL to remove TOTP can therefore be replayed on recovery-code read/rotation, method changes, MFA disable, or passkey deletion, where the handler returns before recalculating the stronger required method. Please bind this proof to the exact operation and target, or store and compare an assurance requirement before accepting it.

backend/src/server/routes/mfa-step-up-fns.ts
```diff
@@ -69,18 +73,17 @@ export const ensureStepUpMfa = async (
     return;
   }
 
-  if (
-    resourceId === MfaStepUpResource.MfaManagement &&
-    (await server.services.mfaSession.hasRecentMfaAuth(userId, tokenVersionId))
-  ) {
-    return;
+  if (resourceId === MfaStepUpResource.MfaManagement) {
+    if (await server.services.mfaSession.hasRecentMfaAuth(userId, tokenVersionId)) return;
```
scott-ray-wilson: **Blocker:** fixing active-session replay alone would still leave this generic recent-auth path. `verifyMfaSession` calls `recordRecentMfaAuth` for every `MfaManagement` challenge without retaining the method or operation, and this branch accepts that marker for every management route. Completing an EMAIL fallback issued for TOTP removal therefore grants the same grace to read or rotate recovery codes and perform other stronger actions. Please avoid recording global management grace for a factor-removal recovery challenge, or scope and check the grace by action and required assurance.

backend/src/services/user/user-service.ts
```diff
@@ -201,28 +212,45 @@ export const userServiceFactory = ({
     return { user: updatedUser, recoveryCodes };
   };
 
+  const findMfaEnforcingOrgs = async (userId: string) => {
+    const userOrgMemberships = await membershipUserDAL.find({
+      actorUserId: userId,
+      scope: AccessScope.Organization,
+      status: OrgMembershipStatus.Accepted
```
scott-ray-wilson: **Blocker:** this raw query does not implement the `Every org is checked` invariant below. `actorUserId` only finds direct membership rows; group-derived org membership is stored under `actorGroupId` and is included by `orgDAL.findEffectiveOrgMemberships`, so an enforcing linked org reached through a group is absent here. It also counts direct memberships whose `isActive` is false. Because this helper decides both whether step-up can be skipped and whether the factor being removed is enforced anywhere, the first gap can allow that factor to be removed and the second can require an unavailable factor for an inactive membership. Please resolve accepted, active effective memberships across all orgs. The existing effective-membership helper is scoped to one org, so this needs either a global equivalent or org enumeration followed by per-org resolution.

backend/src/server/routes/v1/user-router.ts
```diff
@@ -240,7 +240,8 @@ export const registerUserRouter = async (server: FastifyZodProvider) => {
         tokenVersionId: getStepUpSessionId(req),
         resourceId: MfaStepUpResource.MfaManagement,
         mfaSessionId: req.query.mfaSessionId,
-        message: "MFA verification is required to remove your authenticator app"
+        message: "MFA verification is required to remove your authenticator app",
+        excludeMfaMethod: MfaMethod.TOTP
```
scott-ray-wilson, line 246: **High:** this alternate-factor path exposes the unconditional EMAIL fallback in `deleteUserTotpConfig`, as well as the analogous last-passkey branch. On an instance without SMTP, a user with selected TOTP and a working passkey can prove the passkey and delete TOTP, after which the service sets `selectedMfaMethod` to EMAIL. They are then locked out even though a usable passkey remains. The inverse occurs when selected WebAuthn is deleted while TOTP remains. Please select a configured remaining factor, ideally the one just proven, or reject removal if none is usable instead of always setting EMAIL.

backend/src/server/routes/v1/user-router.ts
```diff
@@ -585,13 +586,21 @@ export const registerUserRouter = async (server: FastifyZodProvider) => {
     },
     onRequest: verifyAuth([AuthMode.JWT]),
     handler: async (req) => {
+      // Other passkeys can still answer the challenge, so only removing the last one
+      // needs to fall back to a different factor.
+      const { credentials } = await server.services.webAuthn.getUserWebAuthnCredentials({
+        userId: req.permission.id
+      });
+      const isRemovingLastPasskey = credentials.every((credential) => credential.id === req.params.id);
```
scott-ray-wilson: Please determine the target before asking for step-up. `every()` is true for an empty credential list, and a single matching UUID with different casing compares false even though the database UUID lookup accepts it. Missing or non-UUID IDs are also rejected only by the deletion query after the user completes MFA; a non-UUID currently becomes a database 500. Validate `id` as a UUID, find and authorize the credential before the challenge, return 404 when it is absent, and then use `credentials.length === 1` to decide whether it is the last passkey.

scott-ray-wilson (changes requested):
I found five additional cases that should be resolved before merge. The first three affect authorization or organization-policy enforcement, the fourth can preserve the lockout this PR is intended to fix, and the fifth is request validation. Please add regression coverage for the session/grace scoping and alternate-factor removal cases.

backend/src/server/routes/v1/user-router.ts
Thiago-AS: Not a concern. If the org enforces webauthn and they have no passkeys, at next login they will be prompt to add it again. Blocking would hurt the main reason people delete their last passkey, a lost device, and there's no escape hatch from settings.

backend/src/services/user/user-service.ts
```diff
@@ -580,13 +643,42 @@ export const userServiceFactory = ({
...
+  ): Promise<{ challenge: MfaMethod; accepted: MfaMethod[] }> => {
+    const [user, sessionOrg] = await Promise.all([userDAL.findById(userId), orgDAL.findById(orgId)]);
+    const org =
+      sessionOrg?.rootOrgId && sessionOrg.rootOrgId !== sessionOrg.id
+        ? await orgDAL.findById(sessionOrg.rootOrgId)
+        : sessionOrg;
     const { requiredMfaMethod } = getRequiredMfaMethod(org ?? {}, user ?? {});
-    return requiredMfaMethod;
+    const asRequired = { challenge: requiredMfaMethod, accepted: [requiredMfaMethod] };
+    if (!user || !excludeMethod || requiredMfaMethod !== excludeMethod) return asRequired;
```
scott-ray-wilson: **Blocker:** this early return reintroduces the org-switch bypass. If the current org requires EMAIL while another active effective org enforces TOTP, deleting TOTP reaches this line with `requiredMfaMethod = EMAIL` and `excludeMethod = TOTP`, then returns before `findMfaEnforcingOrgs` runs. The action consequently accepts an EMAIL proof and can remove the other org's enforced TOTP. In `3c03565`, the all-org check ran before the `requiredMfaMethod !== excludeMethod` return; this follow-up reversed that order. Please check enforcing orgs whenever an exclusion is present and return `{ challenge: excludeMethod, accepted: [excludeMethod] }` when any org enforces it. A regression test needs a non-enforcing current org plus a different enforcing org; the current test uses the enforcing org as the current context and misses this case.

backend/src/services/totp/totp-service.ts
```diff
@@ -178,29 +186,20 @@ export const totpServiceFactory = ({ totpConfigDAL, kmsService, userDAL, keyStor
...
-        });
-      }
+    if (!totpConfig) {
+      throw new NotFoundError({
+        message: "TOTP configuration not found"
+      });
+    }
 
-      await totpConfigDAL.deleteById(totpConfig.id, tx);
+    const replacementMfaMethod = await userService.resolveMfaMethodAfterRemoval(userId, MfaMethod.TOTP);
```
scott-ray-wilson, line 197: **High:** replacement selection now happens outside the deletion transaction, and these ORM reads use `replicaNode()` when no `tx` is supplied. This creates a check/use race across both removal services. With selected TOTP, one passkey, and no SMTP, concurrent TOTP and passkey deletions can each observe the other factor, both proceed, and commit with both factors gone and `selectedMfaMethod` pointing to the deleted passkey; I reproduced that interleaving with a synchronized unit test. Replica lag can produce the same stale replacement decision without concurrency. Please keep the target/count/user/replacement reads, deletion, and preference update in one primary transaction and serialize factor removals for the same user (for example with a user-row lock and a consistent lock order, without nesting transactions). Add a concurrent-removal regression test.

scott-ray-wilson (changes requested):
The follow-up addresses the original proof replay/grace, effective-membership, EMAIL replacement, and target-validation paths in their normal flows. Two policy/lockout regressions remain, and I reproduced both with focused tests; details are inline.

backend/src/services/user/user-service.ts
Thiago-AS: Not really a blocker, the user is prompted to add the MFA method again on login, will fix.

scott-ray-wilson (dismissed):
lgtm

scott-ray-wilson (approved):
lgtm

PR #8107, opened 2026-09-12
feat(pki-syncs): choose sync certificates by filter

carlosmonastyrski:
## Context

Certificate syncs inside an Application now pick their certificates by filter instead of one at a time. You describe what the sync should hold, by profile, by certificate order, or by metadata: new certificates are picked up as they are issued, renewals replace their predecessor, and anything that stops matching, expires, or is revoked is dropped and removed from the destination when removal is enabled.

Syncs that can only hold a limited number of certificates, such as Nutanix Prism Central, take a certificate order filter naming at most that many orders and nothing else, so a sync can never be asked to hold more than it can.

Syncs created before Applications existed keep syncing what they already hold and are read-only, moving them means creating the sync inside an Application.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
```
bernie-g, line 5: ignore comment placement

Image
This UI looks weird to me. I feel like there's gotta be a better way to display this. At the very least I think the +3 button should have a hover

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  await knex.raw(`
+    UPDATE "${TableName.PkiSync}" AS s
+    SET "filters" = jsonb_build_object(
+      'certificateOrderIds',
+      COALESCE(
+        (
+          SELECT jsonb_agg(DISTINCT c."orderId")
+          FROM "${TableName.CertificateSync}" cs
+          JOIN "${TableName.Certificate}" c ON c.id = cs."certificateId"
```
bernie-g, line 22: ignore comment placement

Image

Why did we rename this to Certificate Orders? Do we think people will understand what that means?

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasFilters = await knex.schema.hasColumn(TableName.PkiSync, "filters");
+
+  if (!hasFilters) {
+    await knex.schema.alterTable(TableName.PkiSync, (t) => {
+      t.jsonb("filters").nullable();
+    });
```
bernie-g, line 11: ignore comment placement

Image
Can we also replace this X with the trash icon? In general lets make this look more like the other UIs to maintain consistency. Here's an example from proxied services

Image

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  await knex.raw(`
+    UPDATE "${TableName.PkiSync}" AS s
+    SET "filters" = jsonb_build_object(
+      'certificateOrderIds',
+      COALESCE(
```
bernie-g, line 18: Image
I think it's not super obvious that these filters are AND'd together. Can we add a big AND inbetween each one to make it more clear

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
bernie-g, line 22: ignore comment placement

Should we have a way to just sync all certs in the application?

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+          WHERE cs."pkiSyncId" = s.id
+        ),
+        '[]'::jsonb
+      )
+    )
+    WHERE s."filters" IS NULL AND s."applicationId" IS NOT NULL
+  `);
+}
+
+export async function down(knex: Knex): Promise<void> {
```
bernie-g, line 32: Image

This sounds AI

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+  const hasFilters = await knex.schema.hasColumn(TableName.PkiSync, "filters");
+  if (hasFilters) {
+    await knex.schema.alterTable(TableName.PkiSync, (t) => {
+      t.dropColumn("filters");
+    });
```
bernie-g, line 37: Image
nit: can we lowercase "application" so it's consistent with everywhere else

backend/src/db/schemas/pki-syncs.ts
```diff
@@ -39,7 +39,8 @@ export const PkiSyncsSchema = z.object({
   encryptedCredentials: zodBuffer.nullable().optional(),
   lastHealthCheckRanAt: z.date().nullable().optional(),
   lastHealthCheckStatus: z.string().nullable().optional(),
-  lastHealthCheckMessage: z.string().nullable().optional()
```
bernie-g, line 42: ignore comment placement

Image
This UI looks very weird, there's no indication that hovering over these will do anything.

bernie-g: I think these should be badges too instead of just numbers

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
```
bernie-g, line 3: Image
I think there should be pagination here, it just renders an infinite list right now

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+  const hasFilters = await knex.schema.hasColumn(TableName.PkiSync, "filters");
+
+  if (!hasFilters) {
+    await knex.schema.alterTable(TableName.PkiSync, (t) => {
+      t.jsonb("filters").nullable();
+    });
+  }
```
bernie-g, line 12: Image
The UI for selecting a certificate for syncs that only let you choose one is very strange, I feel we shouldn't be using a checkbox selector here

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
+import { Knex } from "knex";
+
+import { TableName } from "../schemas";
+
+export async function up(knex: Knex): Promise<void> {
+  const hasFilters = await knex.schema.hasColumn(TableName.PkiSync, "filters");
```
bernie-g, line 6: Image
This sounds AI, can we rephrase?

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+
```
bernie-g, line 7: Image
I think for cert sync integrations that only allow one cert we don't need this whole UI. We can just have a cert picker like before, right?

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (!hasFilters) {
+    await knex.schema.alterTable(TableName.PkiSync, (t) => {
+      t.jsonb("filters").nullable();
```
bernie-g, line 10: Image
Could we say something like "These certificates will no longer be synced" instead? I think that's more clear than detached

backend/src/server/routes/v1/pki-sync-routers/pki-sync-router.ts
```diff
@@ -279,6 +292,47 @@ export const registerPkiSyncRouter = async (server: FastifyZodProvider, enableOp
     }
   });
 
+  server.route({
+    method: "POST",
+    url: "/certificates/preview",
```
bernie-g: /preview reads a bit weird to me - maybe /resolve or /search would be better?

backend/src/services/certificate/certificate-dal.ts
```diff
@@ -499,48 +517,87 @@ export const certificateDALFactory = (db: TDbClient) => {
...
+        .orderBy([
+          { column: `${TableName.Certificate}.orderId` },
+          { column: `${TableName.Certificate}.notBefore`, order: "desc" }
+        ])
+        .select(
+          `${TableName.Certificate}.id`,
+          `${TableName.Certificate}.commonName`,
+          `${TableName.Certificate}.altNames`,
+          `${TableName.Certificate}.serialNumber`,
+          `${TableName.Certificate}.notAfter`,
```
bernie-g, line 660: Ignore comment placement

From Claude:
1. The preview ignores certificate-level permissions; the picker doesn't

findCertificatesMatchingSyncFilters takes no permission filters at all. Its siblings findActiveCertificatesForSync and countActiveCertificatesForSync both accept permissionFilters and apply applyProcessedPermissionRulesToQuery, and the order picker goes through that path with ProjectPermissionCertificateActions.Read conditions.

So in the same feature, on adjacent screens: the picker hides certificates a user isn't allowed to read, and the filter preview returns them — common name, SANs, serial, expiry, profile. More importantly the filter then syncs them to an external destination. The only gate is Edit on the sync.

frontend/src/components/pki-syncs/forms/schemas/base-pki-sync-schema.ts
```diff
@@ -101,7 +101,21 @@ export const BasePkiSyncSchema = <T extends AnyZodObject | undefined = undefined
     description: z.string().optional(),
     isAutoSyncEnabled: z.boolean().default(true),
     subscriberId: z.string().nullable().optional(),
-    certificateIds: z.array(z.string()).optional(),
+    filters: z
+      .object({
+        profileIds: z.array(z.string().uuid()).max(20).optional(),
+        certificateOrderIds: z.string().uuid().array().max(500).optional(),
```
bernie-g: Doesn't this mean that existing customers who have more than 500 certs won't be able to edit their certs at all

backend/src/services/pki-sync/pki-sync-service.ts
```diff
@@ -1180,17 +1328,153 @@ export const pkiSyncServiceFactory = ({
...
+      await auditLogService.createAuditLog({
+        ...(auditLogInfo ?? { actor: { type: ActorType.PLATFORM, metadata: {} } }),
+        projectId: current.projectId,
+        event: {
+          type: EventType.UPDATE_PKI_SYNC,
+          metadata: {
+            pkiSyncId: current.id,
+            name: current.name,
+            destination: current.destination,
+            hasFilters: true,
```
bernie-g: Why is this hardcoded to true?

docs/documentation/platform/pki/applications/certificate-syncs/overview.mdx
```diff
@@ -115,19 +115,103 @@ flowchart LR
...
+
+Enter a key on its own to match any certificate carrying that key, whatever its value. Enter a key and a value to match that exact pair. A certificate has to carry every pair listed, so each pair narrows the filter further.
+
+### Attaching and detaching one certificate
+
+`POST` and `DELETE` on `/api/v1/cert-manager/syncs/:pkiSyncId/certificates` take certificate ids, resolve each to the order it belongs to, and add or remove that order in the **Certificate Order** filter. `infisical_certificate_sync_certificate` uses this in Terraform, so a client that names a certificate still follows its renewals.
+
+The same resolution applies to the `certificateIds` field on sync creation.
+
+Both work only on a sync whose **Certificate Order** filter is its only filter. On a sync that also filters by profile or metadata, or on one with no filters, they're rejected, because rewriting the order list would narrow the sync and detach everything else. Edit the filters instead.
```
bernie-g: Can we run /[humanize](https://github.com/blader/humanizer) on all this new doc? It sounds very AI generated imo

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
carlosmonastyrski: It's a new concept we introduced a few weeks ago. A 'Certificate Order' doesn't refer to a single certificate instance, but rather the entire lineage, from the initial certificate through all subsequent renewals. By referencing the order instead of a specific certificate, infrastructure-as-code flows like Terraform remain clean and avoid drift during renewals. (I'll add a new concept in the docs and make it more explicit there 👍 )

carlosmonastyrski: Not for now

backend/src/db/migrations/20260910091447_pki-sync-filters.ts
carlosmonastyrski: Add-All checkbox added here
CleanShot 2026-09-15 at 21 33 16@2x

frontend/src/components/pki-syncs/forms/schemas/base-pki-sync-schema.ts
carlosmonastyrski: Increased to 10000 (should not handle more than that for now)

bernie-g:
CleanShot 2026-09-16 at 1 41 40 PM@2x

bernie-g:
CleanShot 2026-09-16 at 1 42 00 PM@2x

bernie-g (approved):

PR #8108, opened 2026-09-13
fix(lib): handle multiple consecutive trailing slashes in removeTrailingSlash

Tyagiquamar:
Fixes removeTrailingSlash helper to remove all consecutive trailing slashes while preserving single root slash.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #8117, opened 2026-09-14
feat(pki): certificate audit log details

bernie-g:
## Context

[PKI-424](https://linear.app/infisical/issue/PKI-424/improve-details-in-certificate-audit-log-events). Certificate audit log events were missing details customers need. Adds fields to existing events; no new event types.

- Approval events: `requestId` renamed to `approvalRequestId`, plus the requester, the certificate request it gates, and the request's real policy type (it previously reported the route's, showing `cert-code-signing` for TLS certs)
- Issuance and renewal: `status`, `certificateRequestId`, `serialNumber`
- Signer events: `requestId` renamed to `approvalRequestId`, plus the signer name
- `revoke-cert`: `revocationReason`
- ACME: `download-acme-certificate` gains the certificate id, common name and serial; `finalize-acme-order` gains a parsed common name
- `get`/`cancel-certificate-request`, `automated-renew-certificate`: identifying fields
- Every event carrying an `applicationId` now carries the application name too, across enrollment, syncs, alerts, inventory views, profile attach/detach, SCEP challenges, issuance, and the certificate lifecycle (revoke, delete, import, renew, key/bundle/PKCS#12 export, metadata and renewal config)

The name is resolved inside the operation that emits the event rather than by a separate lookup in the router, so an audit label can never fail the request. Where a DAL already joined the application, or the service already loaded it, this costs no extra query.

Also adds an audit log section to `backend/CODE_QUALITY.md`, drawn from what this PR got wrong the first time.

## Steps to verify the change

1. Issue a cert with no approval policy: `issue-certificate-from-profile` has `status: issued` and a `serialNumber`.
2. Issue one with an approval policy: same event now reads `status: pending-approval` instead of looking identical to a success.
3. Approve it: `approval-request-approve` has `policyType: cert-request` and a `certificateRequestId` matching the Certificate Requests tab.
4. Revoke a certificate issued against an application: `revoke-cert` carries the revocation reason plus the application id and name.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/services/certificate-v3/certificate-v3-queue.ts
```diff
@@ -91,10 +91,12 @@ export const certificateV3QueueServiceFactory = ({
                     type: EventType.AUTOMATED_RENEW_CERTIFICATE,
                     metadata: {
                       certificateId: certificate.id,
+                      newCertificateId: renewed.certificateId,
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/services/certificate-v3/certificate-v3-queue.ts
```diff
@@ -91,10 +91,12 @@ export const certificateV3QueueServiceFactory = ({
...
                       commonName: certificate.commonName || "",
                       profileId: certificate.profileId!,
-                      renewBeforeDays: certificate.renewBeforeDays?.toString() || "",
-                      profileName: certificate.profileName || ""
+                      renewBeforeDays: certificate.renewBeforeDays ?? undefined,
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/services/certificate-v3/certificate-v3-queue.ts
```diff
@@ -91,10 +91,12 @@ export const certificateV3QueueServiceFactory = ({
...
+                      profileName: certificate.profileName || "",
+                      serialNumber: renewed.serialNumber
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/server/routes/v1/approval-policy-routers/approval-policy-endpoints.ts
```diff
@@ -419,9 +420,11 @@ export const registerApprovalPolicyEndpoints = ({
...
             metadata: {
-              policyType,
+              policyType: request.type,
+              approvalRequestId: request.id,
               justification: req.body.justification || undefined,
-              requestDuration: req.body.requestDuration || "infinite"
+              requestDuration: req.body.requestDuration || "infinite",
+              ...getApprovalRequestSubjectMetadata(request)
             }
           }
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -808,9 +821,9 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
           type: EventType.CANCEL_CERTIFICATE_REQUEST,
           metadata: {
             certificateRequestId: req.params.requestId,
+            commonName: certificateRequest?.commonName ?? undefined,
             cancelled,
-            previousStatus,
-            previousPendingMessage
+            previousStatus
           }
```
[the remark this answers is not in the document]
bernie-g: Fixed

backend/src/server/routes/v1/approval-policy-routers/approval-policy-endpoints.ts
```diff
@@ -539,9 +547,12 @@ export const registerApprovalPolicyEndpoints = ({
           event: {
             type: EventType.APPROVAL_REQUEST_APPROVE,
             metadata: {
-              policyType,
-              requestId: req.params.requestId,
-              comment: req.body.comment
+              policyType: request.type,
+              approvalRequestId: request.id,
```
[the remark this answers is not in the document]
bernie-g: Known.

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4028,8 +4038,11 @@ interface IssueCertificateFromProfile {
   metadata: {
     certificateProfileId: string;
     certificateId: string;
+    certificateRequestId: string;
     commonName: string;
     profileName: string;
+    status: CertificateRequestStatus;
+    serialNumber?: string;
```
sheensantoscapadngan, line 4087: application name too?

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4039,8 +4052,11 @@ interface SignCertificateFromProfile {
   metadata: {
     certificateProfileId: string;
     certificateId: string;
+    certificateRequestId: string;
     profileName: string;
     commonName: string;
+    status: CertificateRequestStatus;
```
sheensantoscapadngan, line 4101: same here, application name?

backend/src/ee/services/audit-log/audit-log-types.ts
```diff
@@ -4049,7 +4065,11 @@ interface OrderCertificateFromProfile {
   type: EventType.ORDER_CERTIFICATE_FROM_PROFILE;
   metadata: {
     certificateProfileId: string;
+    certificateRequestId: string;
+    commonName: string;
     profileName: string;
+    status: CertificateRequestStatus;
+    serialNumber?: string;
     applicationId?: string;
```
sheensantoscapadngan, line 4117: same here re: application name

backend/src/services/certificate-authority/digicert/digicert-revocation-sync-queue.ts
```diff
@@ -71,7 +71,8 @@ export const digicertRevocationSyncQueueFactory = ({
         metadata: {
           certId: cert.id,
           cn: cert.commonName,
-          serialNumber: cert.serialNumber
+          serialNumber: cert.serialNumber,
+          revocationReason: CrlReason.UNSPECIFIED
```
sheensantoscapadngan, line 75: Do you know if this being hard-coded as unspecified is fine? Wondering if this was a hack that we did before to get it out ASAP. Not really relevant to the PR, but if it is indeed a hack, we can create a ticket

bernie-g: Claude says its not a hack and that it's because this queue reconciles revocations that happened somewhere in Digicert so we have no idea why they were revoked since it wasn't in infisical

backend/src/services/pki-application/pki-application-service.ts
```diff
@@ -217,6 +217,11 @@ export const pkiApplicationServiceFactory = ({
     return application;
   };
 
+  const getApplicationNameById = async (applicationId: string) => {
+    const application = await pkiApplicationDAL.findById(applicationId);
+    return application?.name;
+  };
```
sheensantoscapadngan: shouldn't we be throwing an error if application does not exist for the passed in ID?

backend/src/server/routes/v1/pki-application-profile-router.ts
```diff
@@ -128,13 +131,16 @@ export const registerPkiApplicationProfileRoutes = async (server: FastifyZodProv
         profileId: req.params.profileId
       });
 
+      const applicationName = await server.services.pkiApplication.getApplicationNameById(result.applicationId);
+
```
sheensantoscapadngan: Could we return application name from `detachProfile` the same way that we return application ID there? That way, we don't have to do a different fetch altogether just to satisfy the audit logs... This makes it so that the operation being performed is self-sufficient and no other process can cause the non-publishing of audit logs

sheensantoscapadngan: same comment applies all throughout the diff

backend/src/ee/routes/scep/pki-scep-router.ts
```diff
@@ -185,6 +185,8 @@ export const registerPkiScepRouter = async (server: FastifyZodProvider) => {
         actorOrgId: req.permission.orgId
       });
 
+      const applicationName = await server.services.pkiApplication.getApplicationNameById(applicationId);
+
```
sheensantoscapadngan: IGNORE PLACEMENT

events like renew certificate and revoke certificate should also have application ID and name if relevant right? can we add it there too otherwise we'd only have these identifiers for certificate issuance but not for the other lifecycle operations

Image

backend/src/services/pki-application/pki-application-service.ts
bernie-g: This would always run after the actual operation has completed and just when we are writing the audit log so it wouldn't make sense to throw here and skip logging. Also not sure how we would ever get in a situation where there's no application for an ID

backend/src/server/routes/v1/pki-application-profile-router.ts
bernie-g: Done

backend/src/ee/routes/scep/pki-scep-router.ts
bernie-g: Added it to every PKI event

backend/src/services/pki-application/pki-application-service.ts
sheensantoscapadngan: it's currently written as a very generic method inside the application service so i don't think it should matter what the actual intended use is. afaik, we follow this pattern validation consistently throughout the platform so perhaps we rename it?

backend/src/services/certificate-v3/certificate-v3-service.ts
```diff
@@ -2171,6 +2183,21 @@ export const certificateV3ServiceFactory = ({
...
+
+  const issueCertificateFromProfile = async (dto: Parameters<typeof $issueCertificateFromProfile>[0]) =>
+    $withApplicationName(dto.applicationId, await $issueCertificateFromProfile(dto));
+
+  const signCertificateFromProfile = async (dto: Parameters<typeof $signCertificateFromProfile>[0]) =>
+    $withApplicationName(dto.applicationId, await $signCertificateFromProfile(dto));
+
+  const orderCertificate = async (dto: Parameters<typeof $orderCertificate>[0]) =>
+    $withApplicationName(dto.applicationId, await $orderCertificate(dto));
+
```
sheensantoscapadngan: this pattern feels odd. would it be bad to simply return the application name from within the respective methods directly? 

also, if we want to add more metadata to the audit log, is the idea here that we'd have to extend this?

backend/src/ee/services/pki-scep/pki-scep-service.ts
```diff
@@ -105,6 +106,7 @@ type TPkiScepServiceFactoryDep = {
   auditLogService: Pick<TAuditLogServiceFactory, "createAuditLog">;
   permissionService: Pick<TPermissionServiceFactory, "getProjectPermission" | "getResourcePermission">;
   pkiApplicationProfileDAL?: Pick<TPkiApplicationProfileDALFactory, "findOneByApplicationAndProfile">;
+  pkiApplicationDAL?: Pick<TPkiApplicationDALFactory, "findById">;
```
sheensantoscapadngan: why is this optional?

backend/src/services/certificate-v3/certificate-v3-service.ts
bernie-g: Removed

backend/src/ee/services/pki-scep/pki-scep-service.ts
bernie-g: Because the one above was optional too, but there's no reason for that either so I made them both required.

sheensantoscapadngan (approved):

PR #8130, opened 2026-09-15
feat(agent-vault): method and path rules, custom headers, and substitutions

saifsmailbox98:
## Context

A service could only say which hosts it covers. It can now also say which HTTP methods and path prefixes it allows, carry extra headers alongside its credential, and swap placeholders an agent already sends for the real value.

Methods and paths are filters on a service that already matched, not part of the match key, so the host conflict rule is unchanged. A request outside them gets a 403 from the proxy rather than a silently withheld credential.

**Data model:** `allowedMethods` and `allowedPathPrefixes` are nullable `text[]`. `NULL` is the only "unrestricted", since `[]` is a 422, so there is one representation of "all". Arrays rather than the comma-separated string `hostPattern` uses, because a comma is legal in a path. Custom headers and substitutions are two child tables, `agent_vault_service_custom_headers` and `agent_vault_service_substitutions`, sealed with the project cipher and never returned.

**Naming:** the concept is `customHeaders` in every layer, on the management API, on the resolve payload, in the table and in the UI. Unqualified "header" in this product means the credential's own header, and the two names sitting one word apart is what made a custom header silently overwriting the credential easy to miss. The migration is renamed in place rather than given a follow-up, since it has not shipped; anyone who already applied it locally needs `ALTER TABLE agent_vault_service_headers RENAME TO agent_vault_service_custom_headers` plus the matching trigger rename.

**Header collisions:** a custom header may not reuse the credential's effective header name, checked on every write carrying a credential or custom headers, against `mergeCredential`'s output rather than the request body, because a credential-only PATCH can move the bearer onto a name a stored header holds and a custom-headers-only PATCH never reaches `mergeCredential`. That check reads the stored credential inside the transaction, after the bundle lock: `ormify` serves reads from a replica, and a headers-only PATCH judged on a stale copy would admit exactly the collision the check exists to stop. The proxy also writes the credential last now, so a pairing that reached a service some other way costs the custom header rather than the token.

**Path prefixes:** the grammar is an allowlist, letters, digits and `- . _ ~ $ & + , : = @`. A prefix is compared against the request's escaped path, so one carrying anything Go's path encoder rewrites could never match: `/café` would be judged against `/caf%C3%A9`, and `/repos/{owner}` against `/repos/%7Bowner%7D`. `[`, `]`, `!`, `'`, `(`, `)` and `*` are worse still, matching or not depending on what the rest of the request path happens to contain, so all of them are refused on write rather than saved and silently matching nothing.

**Frontend:** hosts and path prefixes are now the same chip field, a new v3 `TagsInput` built on the Base UI combobox parts `Combobox` already uses, not `CreatableSelect`. The credential step lays its fields out on one row, header name, prefix and token, with the widths taken from the custom header row on the transformations step so the two line up.

CLI PR: https://github.com/Infisical/cli/pull/397. The `customHeaders` field rename spans both repos, so they land together. An older proxy forwards without enforcing any of this.

### Members, and the lists behind them

The same branch reworks the member surface, because both land in the same preview window and the docs reference publishes them together.

**Endpoints:** `/agent-vault/memberships` was GET, POST, PATCH and DELETE repeated once per actor kind, twelve routes answering with six response shapes. It is now `/agent-vault/members` with `GET /`, `POST /`, `PATCH /{actorType}/{actorId}` and `POST /revoke`, all speaking the actor union the access bundle endpoints already use. The prefix matches the docs nav, which has said Members since the reference was reorganised, and reads properly against its sibling: `/members` is who is in Agent Vault, `/access-bundles/{id}/members` is who is on one bundle. `PATCH` stays single-actor because role varies per person where add and revoke are the same operation for everyone named.

**Naming:** `AgentVaultMemberType.Identity` becomes `MachineIdentity`, so `actor.type`, the audit discriminator and the PostHog `memberType` move together rather than needing a translation layer. Agent Vault used "identity" for a machine identity on a surface that sits it next to users and groups; every other product spells it out, and `group-router` already publishes `/machine-identities`. Nothing persists the enum, since a member's type is derived from whichever actor column is non-null under `only_one_actor_type`, so there is no migration.

**Revoke:** `POST /members/revoke` rather than a DELETE per actor kind, and the same on bundles, where three DELETEs become `POST /access-bundles/{id}/members/revoke` taking the same three id arrays as the grant it undoes. That breaks REST, which `CODE_QUALITY.md` sanctions for a bulk operation that cannot be expressed per resource and asks to have pointed out rather than shipped quietly: DELETE carrying a body is mangled by enough proxies and clients to be unusable, so the alternative was three routes that cannot batch at all. One consequence worth knowing: an actor holding no membership or grant now comes back in `skipped` with a 200 where the DELETE returned 404, so revoke is idempotent. Only the bundle keeps its 404, because that one is the path resource.

**Batch semantics:** the two add paths checked organization membership differently, direct rows only on the bulk user add and inherited-through-a-group on the single group and identity adds, so `org-dal` gains a batched `findActiveEffectiveOrgMemberActorIds` and the more permissive check wins, one query per actor kind rather than a hundred lookups against a pool of ten. Anyone who already has access is skipped rather than failing the call, and anything that is a caller mistake still fails whole and names every offender, so a batch of a hundred never has to be bisected. The last-admin guard runs once for the whole batch with every membership id in `excludeMembershipIds`, because checking per actor passes for each of two admins individually and removes both.

**Pagination:** every Agent Vault list loaded its whole table and filtered in the browser. Members, bundle members, access bundles and proxies now take `search`, `limit`, `offset` and, where the table offers it, `orderBy` and `orderDirection`, with the count and the page sharing one filter closure so the pager describes the filtered set. Two details are deliberate: the role comes from a subselect, because `membership_roles` allows several rows per membership and a join would split a member across a page; and every order carries the membership id as a tiebreak, because `createdAt` defaults to the transaction's start time, so a bulk add writes up to a hundred rows sharing one timestamp and offset paging would show some twice and never show others. The bundle query keeps the session list's two-hop shape, since it joins services and a LIMIT over that fan-out would cut a bundle's services instead of the bundle list. With the bundle member list paginated, the `members` array embedded in `GET /access-bundles/{id}` is dropped rather than left as a second source that disagrees past one page.

**Frontend:** the three Access Control tabs, the bundles and proxies lists and the manage access sheet all follow the sessions page now, with a debounced search term, page reset on filter change and per-page in table preferences. The three tabs share one per-page key, since they are three tabs of one page. The client-side exclusion lists in the add dialogs are gone: they only worked while the client held every member, and they report the server's `skipped` instead, which also fixes the silent no-op when the picker went stale between render and submit. `AddMemberDialog` is renamed `GrantAccessDialog`, matching its title, its button and the docs.

## Steps to verify the change

Anything on the wire needs the CLI branch. In the cli repo: `gh pr checkout 397 && go build -o infisical .`
A proxy built from `main` forwards without enforcing any of this, so a stale binary looks like nothing works.

### The happy path, all four at once

Point a service at `httpbin.org`, which echoes the method, path, headers and body back. Configure all four on it:

- **Methods:** GET, POST
- **Path prefix:** `/anything`
- **Custom header:** `X-Org-Id` with value `acme`
- **Substitution:** `__TOKEN__` -> any value, on the header surface
- **Credential:** bearer on `Authorization`

1. Through the proxy, run `curl https://httpbin.org/anything -H "X-Api-Key: __TOKEN__"`.

   One response shows all four working: the method and path were allowed, `X-Org-Id: acme` is attached, `X-Api-Key` carries the real value, and `Authorization` carries the credential. The agent held none of them.
2. `POST /anything/nested` also succeeds. POST is allowed and the prefix matches on segments.
3. Substitute into the path: add `__PROJECT__` -> `group/project` on the path surface, then call `/anything/__PROJECT__`. It arrives as `/anything/group%2Fproject`, one segment, and the prefix still accepts it.

### Now each rule blocking

4. `DELETE /anything` -> `403 ... does not allow DELETE: blocked by service policy`, and it never leaves the network.
5. `GET /status/200` -> same 403, outside the prefix. So does `/anything/../status/200`.
6. Add a second custom header named the same as the credential's header. The credential wins, since it's written last.

### Without a proxy (API and UI only)

7. Reopen the service. Methods and path prefixes come back as chips, and header and substitution values come back masked, never in the response.
8. Add the path prefix `/repos/{owner}`. Refused. Same for `/café`. Neither could ever match an escaped path, so they're refused on write instead of saved and silently matching nothing.
9. On a bearer service using `Authorization`, add a custom header named `authorization`. Refused, case-insensitively. Try `Host` too.
10. The one worth your time: PATCH **only the credential**, moving the bearer onto a header name a stored custom header already holds. Refused. Then PATCH **only custom headers**, adding one that collides with the stored credential. Also refused. Two separate paths, and neither sees the other's value in the request body.

### Members, pagination and the rename

11. `GET /api/v1/agent-vault/members` answers for all three kinds in one response. Pass `?actorType=machineIdentity` and the discriminator reads `machineIdentity`, not `identity`. `?search=` matches a full name server-side, and `totalCount` follows the search rather than the table.
12. Walk a page: add enough members in one call that they share a `createdAt`, then read them with `limit=1` and increasing `offset`. Every row appears exactly once, which is the tiebreak doing its job.
13. `POST /members/revoke` naming an actor who holds no membership returns 200 with them in `skipped`. Naming every admin at once is refused and removes nobody. Naming yourself anywhere in a batch refuses the whole call and leaves the others alone.
14. One add naming all three kinds dedupes a repeat and skips whoever already had access, and the rest of the batch still lands.
15. In the UI, set per-page to fifty on the Users tab and switch to Machine Identities and Groups. It stays fifty. The bundle member sheet keeps its own ten.
16. Open the manage access sheet on one bundle, search, close it, then open a different bundle. The search is cleared, rather than the new bundle's members looking as though they are missing.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/agent-vault-access-bundle/agent-vault-access-bundle-service.ts
```diff
@@ -607,28 +777,111 @@ export const agentVaultAccessBundleServiceFactory = (deps: TAgentVaultAccessBund
...
+        // Under the lock and on the primary. A replica read here could be stale enough to admit exactly
+        // the collision this check exists to stop, and it is the only thing between a custom header and a
+        // silently overwritten credential. Runs whether the credential, the headers, or both arrived.
+        if (credential || headers) {
+          const effectiveHeaders =
+            headers ??
+            (await agentVaultServiceHeaderDAL.findByServiceIds([service.id], tx)).map((row) => ({
+              name: row.name
+            }));
+          assertHeadersDoNotShadowCredential(effectiveCredentialConfig, effectiveHeaders);
```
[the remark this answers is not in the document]
saifsmailbox98: Valid, already handled. The credential is re-read through `tx` after the bundle lock when the PATCH sends no credential. One correction: custom headers are no longer injected last, the credential is, so a collision now costs the custom header rather than the token.

frontend/src/components/agent-vault/service-sheet/ServiceSheet.tsx
```diff
@@ -185,22 +221,47 @@ export const ServiceSheet = ({ isOpen, onOpenChange, accessBundleId, service }:
...
         await updateService.mutateAsync({
           accessBundleId,
           serviceId: service.id,
           name: data.name,
-          hostPattern: data.hostPattern,
-          credential: buildCredentialPatch(data)
+          hostPattern: data.hosts.join(","),
+          ...buildPolicy(data),
+          credential: buildCredentialPatch(data),
+          ...buildTransformations(data)
```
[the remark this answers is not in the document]
saifsmailbox98: Real, but it is not specific to this sheet. Every form in Infisical sends the whole form on save, and every API here replaces child lists wholesale. Proxied services does the same and deletes all its credential rows on update without even keeping row ids or stored values, this one at least diffs and keeps both. Not making one sheet behave differently from the rest, if we want partial payloads it should be a product wide change.

backend/src/ee/services/agent-vault-proxy/agent-vault-proxy-service.ts
```diff
@@ -388,12 +403,29 @@ export const agentVaultProxyServiceFactory = ({
...
+
     const services: TResolvedService[] = rows.map((row) => ({
       id: row.id,
       name: row.name,
       accessBundleName: row.accessBundleName,
       hostPattern: row.hostPattern,
-      credential: $decryptCredential(row, decryptor)
+      allowedMethods: row.allowedMethods,
+      allowedPathPrefixes: row.allowedPathPrefixes,
+      credential: $decryptCredential(row, decryptor),
```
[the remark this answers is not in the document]
saifsmailbox98: Known and deliberate, not an oversight. There is no version negotiation with the proxy for now, it is written down in the Agent Vault CLAUDE.md and the docs carry a warning to upgrade proxies before relying on method and path rules. Accepted while the product is in preview.

frontend/src/components/agent-vault/service-sheet/ServiceSheet.tsx
```diff
@@ -289,7 +382,9 @@ export const ServiceSheet = ({ isOpen, onOpenChange, accessBundleId, service }:
         requestDiscard();
       }}
     >
-      <SheetContent className="sm:max-w-6xl">
+      {/* Three columns live in here: the step rail, the form and the explainer. w-3/4 (the Sheet default)
+          left the middle one about 445px, which crammed the Transformations rows. */}
+      <SheetContent className="w-[90vw] sm:max-w-[1500px]">
```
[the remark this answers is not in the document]
saifsmailbox98: Width is now the same string as the proxied service sheet, which this whole layout already mirrors. On the rule itself, `SheetContent` has no size variant and its default caps at `sm:max-w-md`, about 445px, so every wide sheet in the app sets the width through className. There are 78 of them.

backend/src/ee/routes/v1/agent-vault-routers/agent-vault-access-bundle-router.ts
```diff
@@ -282,7 +288,11 @@ export const registerAgentVaultAccessBundleRouter = async (server: FastifyZodPro
       body: z.object({
         name: AgentVaultNameSchema.describe(AGENT_VAULT.SERVICE.name),
         hostPattern: AgentVaultHostPatternSchema,
-        credential: AgentVaultCredentialInputSchema
+        allowedMethods: AgentVaultAllowedMethodsSchema.optional(),
```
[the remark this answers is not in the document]
saifsmailbox98: Same as the other thread on this PR. Known and deliberate, there is no version negotiation with the proxy for now, it is in the Agent Vault CLAUDE.md and the docs warn to upgrade proxies before relying on the rules. Accepted while the product is in preview.

saifsmailbox98:
@greptile please re-review.

Since the last pass:
- The credential is re-read inside the transaction after the bundle lock, so a headers-only PATCH cannot be judged against a stale replica copy.
- Path prefixes are now an allowlist, so one carrying anything the path encoder rewrites is refused on write instead of being saved and silently matching nothing.
- The sheet width is the same string the proxied service sheet uses, which this layout already mirrors.
- `headers` is `customHeaders` in every layer now, including the table and the resolve payload.

Two things were left as they are, both answered inline: the sheet sending untouched lists, which is how every form in Infisical behaves, and the lack of proxy version negotiation, which is a documented preview tradeoff.

Companion CLI PR: https://github.com/Infisical/cli/pull/397

saifsmailbox98:
@greptile review

backend/src/ee/services/agent-vault/agent-vault-transformation-schemas.ts
```diff
@@ -0,0 +1,129 @@
...
+export const AgentVaultCustomHeaderUpdateSchema = z.object({
+  id: z.string().uuid().optional().describe(AGENT_VAULT.SERVICE.customHeaderId),
+  name: agentVaultHeaderNameSchema.describe(AGENT_VAULT.SERVICE.customHeaderName),
+  prefix: headerPrefixSchema.optional().describe(AGENT_VAULT.SERVICE.updateCustomHeaderPrefix),
+  value: secretValueSchema.optional().describe(AGENT_VAULT.SERVICE.updateCustomHeaderValue)
+});
+
+export const AgentVaultSubstitutionInputSchema = z.object({
+  placeholder: placeholderSchema.describe(AGENT_VAULT.SERVICE.placeholder),
+  surfaces: surfacesSchema.describe(AGENT_VAULT.SERVICE.surfaces),
```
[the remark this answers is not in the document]
saifsmailbox98: We only substitute one way, placeholder to real value on the request. Responses pass through untouched, so yes this is possible. Containment is the service's method and path rules, not binding each substitution to a fixed location.

docs/docs.json
scott-ray-wilson, line 1: do we maybe want to put these in nested accordions like the other products?

Image

vs

Image

docs/api-reference/endpoints/agent-vault-proxies/generate-enrollment-token.mdx
```diff
@@ -0,0 +1,4 @@
+---
+title: "Generate enrollment token"
+openapi: "POST /api/v1/agent-vault/proxies/{proxyId}/token-auth/generate-enrollment-token"
```
scott-ray-wilson: not blocking but after seeing this url I wonder if we can make it more succinct?

I assume we need the path as we may provide alternative auth methods but would generate ever apply to anything other than enrollment token for token auth?

docs/api-reference/endpoints/agent-vault-memberships/update-user-role.mdx
```diff
@@ -0,0 +1,4 @@
+---
+title: "Update user role"
+openapi: "PATCH /api/v1/agent-vault/memberships/users/{userId}"
```
scott-ray-wilson: more just question thinking out loud - will these endpoints ever do anything beyond roles where we may want to include role in the path? or is this extendable as you see for future entity mutations?

docs/api-reference/endpoints/agent-vault-access-bundles/list.mdx
scott-ray-wilson, line 1: more just a general schema comment not specific to this endpoint but I believe we have a way to label these options, rather than the generic default labels

Image

vs

Image

docs/api-reference/endpoints/agent-vault-access-bundles/get.mdx
scott-ray-wilson, line 1: just noticed but why don't we have an updated at field for access bundles? 

Image

docs/api-reference/endpoints/agent-vault-services/update.mdx
```diff
@@ -0,0 +1,4 @@
+---
+title: "Update"
+openapi: "PATCH /api/v1/agent-vault/access-bundles/{accessBundleId}/services/{serviceId}"
```
scott-ray-wilson, line 3: also not saying we should but just want to exhaust any thoughts before we make this public; is there any value in having services be a subpath? just requires an additional ID when the service ID alone could narrow the scope - not like services are shared across bundles where the relation plays a factor in scoping?

docs/api-reference/endpoints/agent-vault-access-bundles/get.mdx
scott-ray-wilson, line 1: not specific to this url but wondering if we should use back ticks for these so they get formatted? not sure

Image

docs/api-reference/endpoints/agent-vault-access-bundle-members/list.mdx
scott-ray-wilson, line 1: yeah this one in particular I really feel like we should discriminate on, this response looks so confusing and should probably be differentiated by an enum?

Image

docs/api-reference/endpoints/agent-vault-access-bundle-members/remove-user.mdx
scott-ray-wilson, line 1: also wonder if we should do like we did for agent vault members here because we don't have any easy way to get member ID via the api other than from listing - might be more intuitive just to use the direct resource ID with resource scoped paths

backend/src/lib/api-docs/constants.ts
```diff
@@ -4245,7 +4244,29 @@ export const AGENT_VAULT = {
...
     value: "The secret. Never returned once saved.",
     password:
-      "The password half of the basic credential. May be empty if a username is set, for APIs that carry the whole key in the username. Never returned once saved."
+      "The password half of the basic credential. May be empty if a username is set, for APIs that carry the whole key in the username. Never returned once saved.",
+    allowedMethods:
+      "The HTTP methods this service allows. Null allows every method. Anything else is refused by the proxy with a 403.",
+    allowedPathPrefixes:
+      "The path prefixes this service allows, matched on whole segments, so '/repos' covers '/repos/octo' but not '/repositories'. Null allows every path. A path-restricted service also refuses any request whose path would have to be normalised to judge.",
+    customHeaders:
+      "Extra headers the proxy attaches to every request to this service, on top of the credential. Send the whole list: a row is matched to a stored one by 'id' when you send one and otherwise by its name, a name matching nothing is created, and a stored custom header you leave out is deleted.",
```
scott-ray-wilson: this one is a little confusing to read; The first sentence makes sense but the "Send the whole list..." beyond - not sure what it's trying to communicate? from the UI it just seems like they're always sent but here seems like there's some resolution? and what is this ID? I don't see in API

Image

scott-ray-wilson: oh I see, in update... yeah we need to communicate this more clearly or I'm wondering if we should omit sub/custom header ID entirely - given how you edit it I'd almost think it just adds complexity

backend/src/ee/services/agent-vault/agent-vault-transformation-schemas.ts
```diff
@@ -0,0 +1,129 @@
+import { z } from "zod";
+
+import { AGENT_VAULT } from "@app/lib/api-docs";
+
+import {
+  AGENT_VAULT_HEADER_NAME_MESSAGE,
+  AGENT_VAULT_HEADER_NAME_RE,
+  AGENT_VAULT_NO_CONTROL_CHARS_MESSAGE,
+  AGENT_VAULT_NO_CONTROL_CHARS_RE
+} from "./agent-vault-credential-schemas";
```
scott-ray-wilson, line 10: (oh I think we should be using RE2 for all of these (technically not ReDos vuln right now but probably best for conformity and safety net for future copies/additions)

docs/documentation/platform/agent-vault/services.mdx
```diff
@@ -106,9 +106,33 @@ Wildcards match exactly one label. The pattern below matches `api.atlassian.net`
...
-### Paths/routes
+## Methods and paths
 
-You can't configure a specific path/route on a host. Patterns like `api.github.com/repos` aren't supported.
+By default, a service allows every HTTP method and every path on its hosts. When configuring a service, you have the option to narrow it to only the methods and routes your agent needs.
+
+<Frame>
+  <img src="/images/platform/agent-vault/service-details.png" alt="The Details step of a GitHub service, allowing GET and POST on the path prefixes /repos and /search" />
+</Frame>
+
```
scott-ray-wilson, line 128: I don't see this image? 🤔 

Image

docs/documentation/platform/agent-vault/services.mdx
```diff
@@ -123,4 +147,43 @@ In the above example, requests to `api.github.com` would get the credential from
...
+</Note>
+
+
+## Transformations
+
+A service can add two things to a request that the agent never holds: custom headers and placeholder substitutions. Both are optional, and both work with any credential type, including pass-through.
+
+<Frame>
+  <img src="/images/platform/agent-vault/service-transformations.png" alt="The Transformations step with an API version custom header and a substitution that swaps the agent's placeholder for the real token, matched in the header" />
+</Frame>
```
scott-ray-wilson, line 173: same, don't see this

docs/documentation/platform/agent-vault/services.mdx
```diff
@@ -123,4 +147,43 @@ In the above example, requests to `api.github.com` would get the credential from
...
+
+The swap is a plain find-and-replace, so pick a placeholder distinctive enough that it can't appear anywhere else in the request.
+
+A value containing a `/` stays inside one path segment rather than adding another, so a GitLab project reaches the API as `group%2Fproject` and a path prefix still accepts it.
+
+The proxy doesn't rewrite a body over 10 MB, or one sent compressed with `Content-Encoding`. The placeholder goes out unchanged, and the proxy logs a warning naming the service.
+
+<Warning>
+Method rules, path prefixes, custom headers, and substitutions all need a recent [proxy](/documentation/platform/agent-vault/proxies). An older proxy forwards the request without them: the credential still goes out, the restrictions aren't enforced, and the upstream rejects a request that needed a custom header. Upgrade your proxies before you rely on any of these.
+</Warning>
```
scott-ray-wilson: should we say at least the minimum cli version that works with this?

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
```diff
@@ -34,21 +38,143 @@ export const DetailsFields = () => {
...
+          />
+
+          <Controller
+            control={control}
+            name="methods"
+            render={({ field, fieldState }) => (
+              <FieldContent>
+                {/* Indented under the checkbox label so the seven read as what "All Methods" governs.
+                    Fixed columns line the boxes up; w-fit keeps them hugging the labels rather than
+                    stretching across the width of the sheet. */}
```
scott-ray-wilson: probs don't need

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
```diff
@@ -34,21 +38,143 @@ export const DetailsFields = () => {
...
+                render={({ field: draft }) => (
+                  <TagsInput
+                    value={field.value}
+                    onValueChange={field.onChange}
+                    inputValue={draft.value}
+                    onInputValueChange={(next) => {
+                      draft.onChange(next);
+                      if (fieldState.error) clearErrors("pathPrefixes");
+                    }}
+                    // A comma is legal inside a path, so only Enter, Tab and blur commit here.
```
scott-ray-wilson: would space be valid here for entry as well? I like this component but it doesn't feel super intuitive to know I need to hit enter or tab (maybe an info tooltip would also help?)

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
```diff
@@ -34,21 +38,143 @@ export const DetailsFields = () => {
...
+                    separators={[]}
+                    validateTag={pathPrefixError}
+                    onValidationError={(reason) =>
+                      reason ? trigger("pathPrefixes") : clearErrors("pathPrefixes")
+                    }
+                    isError={Boolean(fieldState.error)}
+                    aria-label="Path prefixes"
+                    placeholder="/api/v1"
```
scott-ray-wilson, line 169: either in the placeholder or somewhere else we should probably communicate leave empty to allow all

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
scott-ray-wilson: also tab actually still focuses the next element so we should prevent default / stop propagation if we want tab to submit

scott-ray-wilson: applies to hosts as well

scott-ray-wilson: applies to hosts as well

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
scott-ray-wilson, line 1: substitution transparency concern:

https://github.com/user-attachments/assets/16d025a5-c6ce-4586-9e78-db5b1a0cac0c

frontend/src/components/agent-vault/service-sheet/ServiceSheet.tsx
```diff
@@ -289,7 +375,9 @@ export const ServiceSheet = ({ isOpen, onOpenChange, accessBundleId, service }:
         requestDiscard();
       }}
     >
-      <SheetContent className="sm:max-w-6xl">
+      {/* Same width as the proxied service sheet, which this whole layout mirrors: three columns need more
+          than the Sheet default, which caps at sm:max-w-md and leaves the middle one about 445px. */}
```
scott-ray-wilson: probs don't need

frontend/src/components/agent-vault/service-sheet/TransformationsFields.tsx
scott-ray-wilson, line 1: when editing, clicking in a sub value field doesn't remove the placeholder 

Image

frontend/src/components/v3/generic/TagsInput/TagsInput.tsx
scott-ray-wilson, line 1: can you ask claude to create a story for this? no need to get picky with it just ask for stories and to follow combobox as ref

backend/src/ee/services/agent-vault/agent-vault-transformation-schemas.ts
```diff
@@ -0,0 +1,129 @@
...
+        code: z.ZodIssueCode.custom,
+        message: `The custom header "${header.name}" is listed twice.`,
+        path: [index, "name"]
+      });
+    }
+    seen.add(key);
+  });
+};
+
+export const addDuplicatePlaceholderIssues = (substitutions: TPlaceheld[], ctx: z.RefinementCtx) => {
```
scott-ray-wilson, line 117: You can save two placeholders where one is the beginning of the other, like __TOKEN__ and __TOKEN__V2. The proxy swaps them one at a time in list order with a plain find-and-replace, so the short one gets swapped inside the long one first. The header that should have received secret B receives secret A with a stray V2 glued on, and secret B never sends

backend/src/ee/services/agent-vault/agent-vault-path-prefix.ts
```diff
@@ -0,0 +1,62 @@
...
+// The grammar of record. packages/agentvault/policy.go does the matching at runtime and reimplements these
+// rules, so a change here needs the same change there.
+
+export const AGENT_VAULT_MAX_PATH_PREFIX_LENGTH = 512;
+export const AGENT_VAULT_MAX_PATH_PREFIXES = 20;
+
+// An allowlist, because a prefix is compared against the escaped path and only these survive that encoding:
+// `/café` would be judged against `/caf%C3%A9` and could never match. '%' is left out so the comparison stays
+// exact, ';' and '\' because servers disagree about them, '?' and '#' because both end the path.
+const PATH_PREFIX_RE = new RE2(/^\/[A-Za-z0-9\-._~$&+,/:=@]*$/);
```
scott-ray-wilson: two here from claude:

1. The allowed-character list is backwards at both edges. It accepts characters that normal HTTP clients escape before sending (@ + , : = & $), and the proxy compares raw bytes, so those prefixes match nothing and every request 403s. It refuses characters clients leave untouched (( ) [ ]), so OData-shaped APIs cannot be restricted at all. The comment explaining the rule is wrong in both directions.

2. packages/agentvault/rewrite.go:98 (CLI PR, verified by running applySubstitutions)
A placeholder in the query string is only swapped when it appears literally. Every standard client escapes { and } first, so ?key={{API_KEY}} goes upstream unsubstituted, the third party answers 401, and the log has no field explaining it. The path surface already has the escaped-form fallback; query needs the same three lines.

docs/documentation/platform/agent-vault/services.mdx
scott-ray-wilson: The docs teach putting a specific host and a wildcard host in one bundle as the way to override. With the new method rules that breaks: the proxy picks the most specific service and stops, so if that one is limited to GET, a POST is refused even though the unrestricted wildcard sibling would have served it. CLAUDE.md files this as a future multi-bundle concern, but it is reachable today inside one bundle.

backend/src/ee/routes/v1/agent-vault-routers/agent-vault-access-bundle-router.ts
```diff
@@ -388,6 +410,10 @@ export const registerAgentVaultAccessBundleRouter = async (server: FastifyZodPro
               req.body.credential?.type === AgentVaultCredentialType.Bearer
                 ? req.body.credential.headerPrefix
                 : undefined,
+            allowedMethods: req.body.allowedMethods,
+            allowedPathPrefixes: req.body.allowedPathPrefixes,
+            customHeaderNames: req.body.customHeaders?.map((header) => header.name),
+            substitutionPlaceholders: req.body.substitutions?.map((substitution) => substitution.placeholder),
             credentialReplaced: isStoredSecretReplaced(req.body.credential)
```
scott-ray-wilson, line 420: The audit row records which header names were in the request but not whether a header’s secret value changed. Rotating a stored header secret writes a byte-identical row to saving with no change, so nobody can answer “when was this rotated”. The line right below already does this correctly for the credential with credentialReplaced.

scott-ray-wilson (changes requested):

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
saifsmailbox98: Removed. The class names already say it.

frontend/src/components/agent-vault/service-sheet/ServiceSheet.tsx
saifsmailbox98: Removed. Same width string as the two proxied service sheets, so grep finds them.

docs/documentation/platform/agent-vault/services.mdx
saifsmailbox98: Both images were pushed about an hour before this, so the preview was likely stale. Reshot them since, the step layout changed.

docs/documentation/platform/agent-vault/services.mdx
saifsmailbox98: Same as above.

docs/documentation/platform/agent-vault/services.mdx
saifsmailbox98: Dropped that warning instead. The product is in preview and this ships with it, so warning about older proxies is noise.

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
saifsmailbox98: Good catch, though the Tab branch was doing less than it looked. onBlur already commits, so tabbing away was always creating the tag — that branch only trapped focus when the value was invalid. Removed it, so now Enter commits and leaving the field commits.

saifsmailbox98: Went the other way instead: a comma now creates a chip in Paths too, same as Hosts. Commas were legal in a path prefix, which is why Paths couldn't use the separator — made them illegal in the grammar, since they only turn up in the variable tail of a real path (Cloudinary's w_300,h_200), never in a prefix you'd configure. Worth flagging two things: POST/PATCH with a comma in allowedPathPrefixes now returns 422, and I dropped the separators prop from TagsInput since nothing varies it any more — comma and newline are fixed.

saifsmailbox98: Yes, same component, so hosts behaves the same.

saifsmailbox98: Yes, same component, so hosts behaves the same.

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
saifsmailbox98: Added it under the field. Also moved the segment rule into a tooltip on the label, since the description was carrying two different things.

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
saifsmailbox98: On space specifically — it is now, good call. Space can't appear in a hostname or a path prefix either, so it was free to take. Comma, space, or both work in either field, and pasting splits on all of them.

backend/src/ee/services/agent-vault/agent-vault-transformation-schemas.ts
saifsmailbox98: Done, both are RE2 now. You're right they weren't a risk — single character classes, no backtracking — but they were the only two in agent-vault still on the built-in engine, with agent-vault-host-pattern and agent-vault-path-prefix already using it. The frontend mirrors stay plain since re2 is native.

frontend/src/components/agent-vault/service-sheet/TransformationsFields.tsx
saifsmailbox98: Real bug, and worse than it looked. It was selecting the sentinel rather than clearing it, which a mouse click undoes — mousedown selects, then mouseup puts the caret where you clicked, so typing went into the middle of the sentinel. It only ever worked when you tabbed in.

Clears on focus now. Leaving the field untouched puts the sentinel back rather than sending an empty secret, and only where empty is refused, so a basic username or password can still be cleared deliberately.

backend/src/ee/services/agent-vault/agent-vault-transformation-schemas.ts
saifsmailbox98: Confirmed, and it depended on the order the server happened to return them in — __TOKEN__ first gave X-B = SECRET_AV2, __TOKEN__V2 first gave the right answer. The proxy now sorts longest-first when it builds a session's substitutions, so the longer placeholder is swapped before a shorter one can eat its prefix. Sorted at resolve time rather than per request, since that slice is shared by every request a session serves.

Left the schema alone: once sorted, that pair is a perfectly valid config, so rejecting it on write would forbid something that works. Fix is in the CLI PR.

backend/src/ee/services/agent-vault/agent-vault-path-prefix.ts
saifsmailbox98: Checked the first half and it doesn't hold — @ + : = & $ go through curl and Go untouched, so those prefixes match fine, no 403.

Second half is right that we refuse ( ) [ ] ! ' * and they do survive. But allowing them wouldn't get you OData: prefixes match whole segments, so /Products never covers /Products(1) whatever the grammar says — you'd be listing exact keys one at a time. Leaving the list alone and fixing the comment instead, which claimed the exclusions were about encoding.

Query one is real. It only bites a brace-style placeholder sent by a client that percent-encodes — __PAT__ style and raw-string clients were always fine — but the path surface already falls back to the escaped form and query should too, especially as the failure is silent. Fixed on the CLI PR.

docs/documentation/platform/agent-vault/services.mdx
saifsmailbox98: Reproduced — with api.github.com GET-only and *.github.com unrestricted, a POST is refused and the wildcard never gets a look.

Keeping the behaviour though. Falling through would mean adding a permissive wildcard silently loosens a restriction you set on the specific host, with nothing in the UI saying so. A restriction you can lose that way isn't one.

The docs were the actual problem — they framed it as which credential you get, when it's which service handles the request, rules included. Fixed that.

backend/src/ee/routes/v1/agent-vault-routers/agent-vault-access-bundle-router.ts
saifsmailbox98: Right, and the same gap was next door on substitutions, so both get it. customHeadersReplaced and substitutionsReplaced list which ones arrived carrying a value — omitting a value means keep the stored secret, so sending one is the rotation signal.

Added as new fields rather than reshaping customHeaderNames, since audit rows go out through log streams and changing an existing field's type would break whatever is parsing them. Still names and placeholders only, never a value. Create doesn't get them, same reason it doesn't get credentialReplaced: nothing is being replaced.

backend/src/lib/api-docs/constants.ts
saifsmailbox98: Rewrote all four, they were one long sentence carrying four rules. Now: send the full list, anything you leave out is deleted, send an id to change a row in place and keep its stored value, no id means matched by name.

Keeping the id though. It's what lets you rename a header without resending its value — without it a rename is a delete plus a create, so you'd have to paste the secret again into a field that never shows it back. It also lets you swap two names in one request, which name-matching can't express. It is in the API by the way, it comes back on every service response.

Also dropped a docs line that was explaining this to users. It's plumbing they see working in the UI anyway.

docs/api-reference/endpoints/agent-vault-access-bundles/get.mdx
saifsmailbox98: Yes, good catch. Backticked the literals across the agent-vault descriptions — host patterns, Authorization, the path prefix examples, null, the traffic policy values. It's what the rest of constants.ts already does, about 190 lines of it; we were the outlier using single quotes.

docs/api-reference/endpoints/agent-vault-access-bundles/list.mdx
saifsmailbox98: Yes — .describe(JSON.stringify({ title: "..." })). jsonDescription in fastify-zod.ts parses that and spreads it into the JSON schema, which is how app connections and secret syncs get their labels.

Added it to all nine branches across the three credential unions: response summary, create body, update body. Using Bearer / Basic Auth / Pass-through, the same words as the credential type dropdown in the UI.

saifsmailbox98: Went wider on this after a second look, since you said it wasn't specific to the endpoint. The proxies list was showing Option 1 / Option 2 for the admin and member views of the same proxy, and the resolve payload had the same three credential shapes unlabelled. Both named now, so that's every union in agent-vault.

docs/api-reference/endpoints/agent-vault-access-bundle-members/list.mdx
saifsmailbox98: done — a member is now `id`, `createdAt` and one `actor`:

```jsonc
{ "actor": { "type": "user", "id": "3c90...", "username": "jake", "email": "jake@acme.com" } }
{ "actor": { "type": "group", "id": "7d81...", "name": "Platform" } }
```

`skipped` is `{ type, id }` too. went with `actor` since `subject` already means the CASL resource.

frontend/src/components/v3/generic/TagsInput/TagsInput.tsx
saifsmailbox98: done, 7 stories following combobox.

kept the examples generic rather than hosts and paths — hosts is just in the validation one.

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
saifsmailbox98: this is on one step now — credential type and substitutions are together, so you see both at once.

docs/api-reference/endpoints/agent-vault-access-bundles/get.mdx
saifsmailbox98: the column and its trigger were always there, just never selected. added it on bundles and services.

left it off proxies — the heartbeat is a raw UPDATE on that row, so it was moving every poll interval rather than when settings changed. heartbeat already covers that.

docs/api-reference/endpoints/agent-vault-access-bundle-members/remove-user.mdx
saifsmailbox98: done — matches the memberships endpoint now:

```
DELETE /access-bundles/{id}/members/users/{userId}
DELETE /access-bundles/{id}/members/groups/{groupId}
DELETE /access-bundles/{id}/members/identities/{identityId}
```

no more listing first to find an id. the old one also handed you a memberships row id, and that table holds org and project memberships too, so it was one missing scope filter away from deleting the wrong row. that whole shape is gone now.

docs/api-reference/endpoints/agent-vault-services/update.mdx
saifsmailbox98: looked into it properly. the thing that usually decides this is whether the child can move parents — a nested url that can be reparented goes stale, so flat wins there. ours can't move, `accessBundleId` isn't in the update body.

our own codebase does both:

```
flat     PATCH  /api/v1/pam/accounts/{accountId}                    folderId goes in the create body
nested   PATCH  /api/v1/projects/{projectId}/roles/{roleId}
nested   PATCH  /api/v1/projects/{projectId}/trusted-ips/{trustedIpId}
nested   DELETE /api/v1/pam/folders/{folderId}/users/{userId}
```

so does the industry, and both camps are deliberate:

```
nested   PATCH  /v1/projects/{p}/locations/{l}/instances/{i}    google, AIP-122 — hierarchy is the identifier
nested   POST   /Accounts/{AccountSid}/Messages/{MessageSid}    twilio, even though message sids are global
flat     POST   /v1/subscription_items/{id}                     stripe — subscription goes in the create body
```

stripe is exactly what you're describing, and pam accounts already follow it here.

two small things lean nested for us: `POST` has to name the bundle anyway, so flattening only PATCH/DELETE splits the three verbs across two roots, and the members endpoints next door are nested.

honestly a coin flip. happy either way, your call.

docs/docs.json
saifsmailbox98: good catch, done. every other product wraps its groups in one — Secrets management, Certificate management, Infisical KMS — ours were sitting at the top level next to KMIP servers and Gateways.

```
before                              after
Agent Vault access bundles          Agent Vault > Access bundles
Agent Vault services                            > Services
Agent Vault access bundle members               > Access bundle members
Agent Vault members                             > Members
Agent Vault proxies                             > Proxies
Agent Vault sessions                            > Sessions
```

the prefix was doing the parent group's job, so it could go. also makes the two membership groups readable instead of two names that differ three words in.

docs/api-reference/endpoints/agent-vault-proxies/generate-enrollment-token.mdx
saifsmailbox98: your assumption's right, that's exactly why the segment is there. everything else about an auth method goes through a discriminated body on POST/PATCH so it needs no path of its own — minting an enrollment token is the one thing that can't, so it sits under `/token-auth/`.

the verbosity was deliberate. it used to be `/token-auth/configure`, and configure didn't say whether it set the method up once or handed you something back, so it got renamed to say what it does.

and on your literal question — you're right, under `token-auth` there's nothing to generate but an enrollment token, so `generate-` is redundant. `POST .../token-auth/enrollment-tokens` would be shorter and closer to how we do it elsewhere (`POST /auth/token-auth/identities/{id}/tokens`).

catch is it's in four products:

```
POST /api/v3/gateways/{gatewayId}/token-auth/generate-enrollment-token
POST /api/v2/relays/{relayId}/token-auth/generate-enrollment-token
POST /api/v1/kmip/servers/{kmipServerId}/token-auth/generate-enrollment-token
POST /api/v1/agent-vault/proxies/{proxyId}/token-auth/generate-enrollment-token
```

renaming means 4 routes, 4 frontend callers, 3 doc pages, a doc string with the url written into it, and the CLI's pam e2e helper — across two repos, and it's breaking on v3 gateways which is public.

leaving agent vault matching the other three for now. happy to open a separate PR if you think it's worth it before we go public.

saifsmailbox98: changed it — `POST .../token-auth/enrollment-token`. dropped `generate-`, since under token-auth there's nothing else it could be minting.

singular rather than `enrollment-tokens`: minting deletes the outstanding row and writes one, so a proxy has at most one ever. a plural would promise a list that can't exist. closest things anyone else ships are singular too — github's runner `registration-token`, k8s serviceaccount `token`.

**this now diverges from the other three** — gateways, relays and kmip keep `generate-enrollment-token`. deliberate: they're public on v3/v2/v1 and the cli's pam e2e calls the gateway one, whereas agent vault is still preview. theirs is also a first mint where ours only ever reissues, since the first token comes back from create. worth bringing them over whenever there's a reason to touch them.

docs/api-reference/endpoints/agent-vault-memberships/update-user-role.mdx
saifsmailbox98: extendable as-is — the resource being patched is the membership, and role is just its only mutable field right now. anything else we let people change goes in the same body.

`isActive` is the obvious next one: it's already on the membership and we return it, we just don't let you set it. if we add deactivate-without-removing it's a body field, not a new endpoint.

putting role in the path would mean a new endpoint per field, which is the thing PATCH exists to avoid. pam does the same shape.

only thing it costs is the page title reading "update user role" when the endpoint is really "update user membership" — worth renaming if it ever takes a second field.

docs/api-reference/endpoints/agent-vault-services/update.mdx
scott-ray-wilson: fine to leave as is

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
scott-ray-wilson: we may want to auto expand the accordion if the erroring field is inside it and also maybe scroll the section to it if it is off screen?

CleanShot 2026-09-17 at 13 45 53@2x

CleanShot 2026-09-17 at 13 46 08@2x

frontend/src/components/agent-vault/service-sheet/TransformationsFields.tsx
scott-ray-wilson: feels good for substitutions but the clearing off the username/password feels frustrating if I'm not intending too; no way to reset it unless you exit and restart
https://github.com/user-attachments/assets/3d2577f4-4954-474a-97ac-e9ea487f124f

frontend/src/components/agent-vault/service-sheet/DetailsFields.tsx
saifsmailbox98: done both. the expand was actually already there, it just only fired the first time — we gate steps with `trigger` rather than `handleSubmit`, so `submitCount` never moves and pressing Continue again after collapsing it by hand did nothing. the sheet counts the blocked attempts itself now.

it also scrolls to the first invalid field. that one waits a beat before scrolling because the accordion unmounts its content while it's closed, so there's nothing to scroll to until it's opened and taken its height.

frontend/src/components/agent-vault/service-sheet/TransformationsFields.tsx
saifsmailbox98: yeah, fair — it cleared on focus and never put it back. every secret field restores the dots now when you leave it empty, so clicking in and wandering off can't wipe anything.

username and password are the two that can legitimately be empty, so they get an ✕ to clear on purpose. it greys out once you've pressed it and puts the caret in the field, since you're usually clearing to retype. the review step already said "Cleared" for those so that half needed nothing.

one thing worth knowing: if a basic credential only has a password and you clear it, the server refuses, which is right — it needs one of the two. what's off is the display: the form shows dots for a username that isn't there, so the refusal reads as arbitrary. the response tells us the type but not which halves are set. that error lands under the field now rather than as a toast, and I can add `hasUsername`/`hasPassword` to the credential summary if you want the empty one to actually look empty.

backend/src/ee/services/agent-vault/agent-vault-transformation-schemas.ts
```diff
@@ -0,0 +1,129 @@
...
+  .string()
+  .min(1)
+  .max(8192)
+  .regex(AGENT_VAULT_NO_CONTROL_CHARS_RE, AGENT_VAULT_NO_CONTROL_CHARS_MESSAGE);
+
+// No minimum length: a short placeholder over-matches, but that is the author's own doing.
+const placeholderSchema = z
+  .string()
+  .trim()
+  .min(1)
```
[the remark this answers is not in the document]
saifsmailbox98: fixed in the CLI, where the substitution actually happens: https://github.com/Infisical/cli/pull/397 — the body rewrite divides now instead of multiplying, same as the helper next to it already did.

the cap here is fine as it stands. a 1 byte placeholder with an 8KB value is a legal service and should be, the arithmetic downstream just has to cope with it.

backend/src/ee/services/agent-vault-proxy/agent-vault-proxy-service.ts
```diff
@@ -388,12 +401,29 @@ export const agentVaultProxyServiceFactory = ({
...
+      return (JSON.parse(decryptor({ cipherTextBlob: encryptedValue }).toString("utf-8")) as { value: string }).value;
+    };
+
     const services: TResolvedService[] = rows.map((row) => ({
       id: row.id,
       name: row.name,
       accessBundleName: row.accessBundleName,
       hostPattern: row.hostPattern,
-      credential: $decryptCredential(row, decryptor)
+      allowedMethods: row.allowedMethods,
```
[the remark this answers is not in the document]
saifsmailbox98: not doing this one, same answer as on the CLI side: https://github.com/Infisical/cli/pull/397

`_method` is a Rails/Laravel/Symfony convention rather than anything in HTTP. we forward an honest POST and the upstream decides to read a body field and treat it as something else. chasing that means chasing `?action=delete` and json rpc envelopes next, and it can't be made complete anyway — GraphQL does every mutation over POST, so a GET/POST-only rule never stopped a delete there.

a method rule constrains the method on the wire. the override headers are stripped because review asked for them by name, not because that's where the principle lands.

frontend/src/components/agent-vault/service-sheet/TransformationsFields.tsx
```diff
@@ -0,0 +1,339 @@
...
+                      name={`substitutions.${index}.surfaces`}
+                      render={({ field, fieldState }) => (
+                        <Field>
+                          <FieldContent>
+                            <FieldLabel className="text-xs">
+                              Replace In
+                              <Tooltip>
+                                <TooltipTrigger asChild>
+                                  <InfoIcon />
+                                </TooltipTrigger>
```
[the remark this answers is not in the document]
saifsmailbox98: This matches how we do help tooltips everywhere else, bare icon in ~107 spots, including the Paths tooltip a few fields up in this same sheet. The a11y point is fair though. Making these focusable would be a change to the shared label/tooltip components so it lands everywhere at once, rather than a one-off here.

frontend/src/components/agent-vault/service-sheet/SubstitutionExample.tsx
```diff
@@ -0,0 +1,57 @@
...
+  title,
+  isResolved,
+  token
+}: {
+  title: string;
+  isResolved?: boolean;
+  token: string;
+}) => (
+  <div className="overflow-hidden rounded-md border border-border">
+    <div className="flex items-center gap-1.5 border-b border-border bg-container px-2.5 py-1.5 font-mono text-[10px] tracking-wider text-muted uppercase">
```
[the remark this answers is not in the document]
saifsmailbox98: done — dropped the uppercase, the captions read "Agent sends" / "Service receives" now. took the letter-spacing off with it since it was the paired small-caps treatment and looks odd on sentence case.

scott-ray-wilson (approved):
lgtm

PR #8132, opened 2026-09-15
feat: add time units to secret scanning env variables

mathnogueira:
## Context

Replace the secret scanning timeout variables to allow time units to be added, making it simpler to configure those values.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


backend/src/lib/config/env.ts
```diff
@@ -63,7 +63,75 @@ export const runModesSchema = zpStr(z.string().optional())
...
+  defaultValue: string;
+  legacyMsEnvVar: string;
+}) =>
+  zpStr(
+    z
+      .string()
+      .optional()
+      .describe(description)
+      .transform((val, ctx) => {
+        const legacyValue = process.env[legacyMsEnvVar]?.trim() || undefined;
```
varonix0, line 83: I think we should even do a log here saying "Warning: The environment variable `<LEGACY_ENV_VAR_NAME>` has been deprecated. Please use `<NEW_ENV_NAME>` instead."

varonix0 (changes requested):

varonix0 (approved):

PR #8139, opened 2026-09-15
feat: add PowerDNS as a DNS-01 provider for ACME certificate authorities

carlosmonastyrski:
## Context

Adds PowerDNS as a DNS-01 provider so an ACME certificate authority can solve challenges against a self-hosted PowerDNS server. The connection can route through a Gateway or connect directly, and accepts a CA certificate for servers behind a TLS reverse proxy.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/services/certificate-authority/certificate-authority-queue.ts
```diff
@@ -92,7 +98,9 @@ export const certificateAuthorityQueueFactory = ({
     pkiSubscriberDAL,
     projectDAL,
     pkiSyncDAL,
-    pkiSyncQueue
+    pkiSyncQueue,
+    gatewayV2Service,
+    gatewayPoolService
```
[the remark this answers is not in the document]
carlosmonastyrski: Subscribers are a legacy entity scheduled to be removed and no longer supported

backend/src/keystore/keystore.ts
```diff
@@ -63,6 +63,8 @@ export const KeyStorePrefixes = {
     `app-connection-concurrency-${connectionId}${targetHost ? `-${targetHost.toLowerCase()}` : ""}` as const,
   AppConnectionCommandLock: (connectionId: string, targetHost?: string) =>
     `app-connection-command-mutex-${connectionId}${targetHost ? `-${targetHost.toLowerCase()}` : ""}` as const,
+  AcmeDnsRecordLock: (connectionId: string, zoneId: string, recordName: string) =>
+    `acme-dns-record-mutex-${connectionId}-${zoneId.toLowerCase()}-${recordName.toLowerCase()}` as const,
```
bernie-g, line 66: ignore comment placement

Image
Can we make this automatically strip the /api/v1 part of the URL once you click out of the textbox

backend/src/server/routes/v1/app-connection-routers/powerdns-connection-router.ts
```diff
@@ -0,0 +1,51 @@
...
+    app: AppConnection.PowerDns,
+    server,
+    sanitizedResponseSchema: SanitizedPowerDnsConnectionSchema,
+    createSchema: CreatePowerDnsConnectionSchema,
+    updateSchema: UpdatePowerDnsConnectionSchema
+  });
+
+  server.route({
+    method: "GET",
+    url: `/:connectionId/zones`,
```
bernie-g, line 26: Image
Are we planning on supporting more auth methods here? If not we can remove this

bernie-g: According to claude the dns resolver goes through the public internet even if you have a gateway configured. Doesn't that kinda defeat the point of supporting gateways here?

backend/src/server/routes/index.ts
```diff
@@ -3443,7 +3443,9 @@ export const registerRoutes = async (
     pkiSyncDAL,
     pkiSyncQueue,
     internalCertificateAuthorityDAL,
-    hsmConnectorService
+    hsmConnectorService,
```
bernie-g, line 3429: Image
This picker lets me select app connections that aren't powerdns when I have PowerDNS as the dns provider selected

bernie-g (commented):
From Claude:
Automatic renewals run without the new distributed lock. certificate-authority-queue.ts:89 builds AcmeCertificateAuthorityFns without keyStore, though it has one in scope; certificate-issuance-queue.ts:220 passes it. So the subscriber renewal path falls back to the in-process lock only, which is exactly where a wildcard plus apex pair of challenges lands on a schedule across several API instances. Related: the lock-timeout message says "Retry the order", which nobody reads on an unattended renewal.

backend/src/server/routes/v1/app-connection-routers/powerdns-connection-router.ts
```diff
@@ -0,0 +1,51 @@
...
+            id: z.string().describe("The PowerDNS zone ID."),
+            name: z.string().describe("The name of the PowerDNS zone.")
+          })
+          .array()
+      }
+    },
+    onRequest: verifyAuth([AuthMode.JWT, AuthMode.OAUTH]),
+    handler: async (req) => {
+      const { connectionId } = req.params;
+      const zones = await server.services.appConnection.powerDns.listZones(connectionId, req.permission);
```
bernie-g, line 47: Image
Should we have hints/tooltips for these too

backend/src/keystore/keystore.ts
```diff
@@ -63,6 +63,8 @@ export const KeyStorePrefixes = {
     `app-connection-concurrency-${connectionId}${targetHost ? `-${targetHost.toLowerCase()}` : ""}` as const,
```
bernie-g, line 62: Image
Can we remove the trailing dots here? Looks a bit weird

bernie-g: Biggest nit of all time but shouldn't Zone have an asterisk next to it to indicate it's required 
CleanShot 2026-09-18 at 2 05 44 PM@2x

backend/src/server/routes/v1/app-connection-routers/app-connection-router.ts
```diff
@@ -436,6 +441,7 @@ const AppConnectionOptionsSchema = z.discriminatedUnion("app", [
   LiteLLMConnectionListItemSchema,
   FireworksConnectionListItemSchema,
   NutanixPrismCentralConnectionListItemSchema,
+  PowerDnsConnectionListItemSchema,
```
bernie-g, line 444: Image
For all the other dns providers we require https but here we allow http. Can we require https or at least warn if not?

backend/src/server/routes/v1/app-connection-routers/app-connection-router.ts
```diff
@@ -436,6 +441,7 @@ const AppConnectionOptionsSchema = z.discriminatedUnion("app", [
   LiteLLMConnectionListItemSchema,
```
bernie-g, line 441: Claude:
keyStore not passed to AcmeCertificateAuthorityFns in certificate-authority-queue.ts:89. Auto-renewals lose the cross-instance lock. The dep is optional in the type, so nothing catches it.

backend/src/server/routes/v1/app-connection-routers/app-connection-router.ts
carlosmonastyrski: PowerDNS does not natively support HTTPS, only if you configure a proxy

bernie-g (approved):

PR #8148, opened 2026-09-15
fix(identity-gcp-auth): enforce GCE project and zone restrictions

victorvhs017:
## Context

GCP GCE identity auth allowed `allowedProjects` and `allowedZones` to be configured but only enforced them when the ID token included `google.compute_engine`. That claim is present only with `format=full` metadata tokens; a `format=standard` token (or any service account minting an ID token via `generateIdToken`) could skip project/zone checks entirely while still passing signature, audience, and optional service-account allowlist validation.

This change fails closed: for `gce` identities with `allowedProjects` or `allowedZones` set, login is rejected when `computeEngineDetails` is missing (`reasonCode: compute_engine_details_missing`). Project and zone checks now always run when those fields are configured, since the missing-details gate guarantees the claim is present.

Validation error handling in `identity-gcp-auth-fns.ts` is also tightened: JWT verify/decode failures map to explicit 401 messages instead of bubbling library errors (which could include the full token in logs), and IAM auth validates `header.kid` and signing-key lookup before verification.

## Steps to verify the change

1. Attach GCP auth to an identity with type `gce`, set `allowedProjects` and/or `allowedZones`, and note the identity ID.
2. **Negative — missing GCE details:** POST `/api/v1/auth/gcp-auth/login` with a standard-format ID token (no `google.compute_engine`). Expect `401` with `reasonCode: compute_engine_details_missing`.
3. **Negative — wrong project/zone:** From a GCE VM outside the allowlist, fetch a metadata token with `format=full` and `audience=<identityId>`. Expect `401` with `project_not_allowed` or `zone_not_allowed`.
4. **Positive:** From an allowed VM, use `format=full` and the same audience. Expect `200` and an access token.
5. **Regression — no project/zone restrictions:** Identity with only `allowedServiceAccounts` (or empty project/zone fields) should still accept a valid token without `compute_engine` when the service account is allowed.
6. **IAM auth:** POST login with a valid IAM signed JWT; confirm malformed tokens, missing `kid`, and bad signatures return the new explicit 401 messages without leaking the JWT in server logs.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


victorvhs017:
@greptileai re-review

akhilmhdh (approved):

PR #8160, opened 2026-09-16
fix(ldap-auth): enforce bind password when changing LDAP URL

victorvhs017:
## Context
Updating a machine identity’s LDAP auth let you change the LDAP URL and omit `bindPass`. The stored bind password is encrypted and never returned, so the update kept the previous server’s password against the new URL.
Custom-scope updates now require `bindPass` whenever the URL changes. Template-backed configs are unchanged (credentials come from the template). The identity LDAP form blocks submit and tells the operator to re-enter the bind password when the URL field differs from the saved value. Leaving the URL alone still allows a blank password (keep current / rotate only).
## Steps to verify the change
1. Attach LDAP auth to a machine identity (custom scope, not a template). Note the current URL.
2. **UI:** Organization → Access Control → identity → LDAP auth. Change the URL, leave Bind Pass blank, submit. The form should fail with “Bind Pass is required when the LDAP URL changes” and the helper should read “The LDAP URL changed, so the bind password must be re-entered.” Restore the original URL and leave Bind Pass blank — save should succeed.
3. **API:** `PATCH /api/v1/auth/ldap-auth/identities/<identityId>` with a new `url` and no `bindPass`. Expect `400` with: `Changing the LDAP URL requires supplying bindPass, because the stored bind password cannot be read back. Send the bind password for the new server with this change.`
4. Repeat the PATCH with the new `url` **and** `bindPass`. Expect `200`.
5. PATCH a URL change while the identity uses an LDAP auth template (`templateId`). Expect `200` without sending `bindPass`.
## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


victorvhs017:
@greptileai re-review

backend/src/ee/services/identity-auth-template/identity-auth-template-schemas.ts
```diff
@@ -10,10 +10,10 @@ import { formatOidcAudiences } from "@app/services/identity-oidc-auth/identity-o
 import { TEMPLATE_VALIDATION_MESSAGES } from "./identity-auth-template-enums";
 
 export const ldapTemplateFieldsSchema = z.object({
-  url: z.string().min(1, TEMPLATE_VALIDATION_MESSAGES.LDAP.URL_REQUIRED),
-  bindDN: z.string().min(1, TEMPLATE_VALIDATION_MESSAGES.LDAP.BIND_DN_REQUIRED),
+  url: z.string().trim().min(1, TEMPLATE_VALIDATION_MESSAGES.LDAP.URL_REQUIRED),
+  bindDN: z.string().trim().min(1, TEMPLATE_VALIDATION_MESSAGES.LDAP.BIND_DN_REQUIRED),
   bindPass: z.string().min(1, TEMPLATE_VALIDATION_MESSAGES.LDAP.BIND_PASSWORD_REQUIRED),
-  searchBase: z.string().min(1, TEMPLATE_VALIDATION_MESSAGES.LDAP.SEARCH_BASE_REQUIRED),
+  searchBase: z.string().trim().min(1, TEMPLATE_VALIDATION_MESSAGES.LDAP.SEARCH_BASE_REQUIRED),
```
scott-ray-wilson: we use this for the response schema so technically the `trim` could break existing templates in the database as whitespace could have been stored - we encountered this with OIDC below so we only kept the trim on the request validation, not response 

```// OIDC templates hold no write-only credentials, so the response is the stored shape.
// The two normalized fields are re-declared as plain strings so no request-side transform
// or refinement runs on serialization and one stored row cannot fail a list response
export const oidcTemplateFieldsResponseSchema = oidcTemplateFieldsSchema.extend({
...```

backend/src/services/identity-ldap-auth/identity-ldap-auth-service.ts
scott-ray-wilson, line 1: question: if the legacy template is already attached to an existing identity prior to this patch, updating the template will succeed, even though an attach would fail - should we also include these guards on the template update?

scott-ray-wilson (changes requested):

backend/src/ee/services/identity-auth-template/identity-auth-template-schemas.ts
victorvhs017: Removed the trim for the ldapTemplateFieldsResponseSchema.

backend/src/services/identity-ldap-auth/identity-ldap-auth-service.ts
victorvhs017: you're right, they could also only update the URL in the patch and that would cause the same initial issue reported. Fixing that!

scott-ray-wilson (approved):
lgtm

PR #8162, opened 2026-09-16
chore(ci): scan released images with Trivy

Vligai:
## Context

This adds a Trivy scan of each published image. Fixable CRITICAL and HIGH findings go to this repo's code scanning, under Security -> Code scanning filtered by tool "Trivy". 

The scan cannot fail a release. It is `continue-on-error` with `exit-code: 0`, the upload is gated on the scan having succeeded, and if either step fails the job emits a warning annotation. The image is already pushed by the time these steps run, so a red job would flag the problem too late to prevent it.

Most of the choices below are hard to read off the diff:

- Scans by digest. The build steps now carry an `id`, and the scan targets `image@${{ steps.build.outputs.digest }}`. Scanning by tag leaves a window where the report could describe something else.
- Both the action and the scanner are pinned. The action is pinned by commit SHA like everything else in this file. The Trivy binary is pinned separately through `TRIVY_VERSION`, because the action's own default lags it (v0.70.0 at the pinned SHA).
- Results upload to `refs/heads/main`. This workflow runs only on tag pushes, and the code scanning alerts UI is scoped to branches, so results left on `refs/tags/*` land where nobody reads them and never reconcile against the next release. This assumes release tags are cut from main. A tag that is not gets rejected by the API, and the guard step reports it.
- The SARIF category carries the release channel. Code scanning reconciles per (ref, category), closing any alert the newest analysis omits. Nightlies share this trigger and this ref, so without the suffix tonight's nightly would close the alerts describing the stable image users are running.
- amd64 only. Both images publish amd64 and arm64, but the OS packages are pinned to identical versions for both arches (`Dockerfile.standalone-infisical:226-243`), and Debian advisories match on package version, so arm64's findings are amd64's. Scanning both mostly produces duplicate alerts to dismiss twice.
- `ignore-unfixed: true` keeps the list to what can be acted on, which is what the Debian pin refreshes in the Dockerfiles already do.
- The vulnerability DB comes from the ECR mirror. Trivy defaults to GHCR, which rate-limits anonymous pulls from the shared Actions egress IPs.

## Screenshots

Not applicable, CI only.

## Steps to verify the change

Static, runnable now:

1. `actionlint .github/workflows/release-standalone-docker-img-postgres-offical.yml` reports only the four issues that predate this change: two deprecated `::set-output` calls and two `steps.version.*` references to a step that does not exist.
2. The workflow parses and all five jobs survive, including `infisical-fips-standalone`, which `generate-upgrade-impact` depends on.
3. The pinned SHAs resolve to their claimed tags. `aquasecurity/trivy-action@ed142fd` is `v0.36.0` and `github/codeql-action@b96794f` is `v4.38.0`.
4. Every `with:` key exists in trivy-action's `action.yaml` at the pinned SHA. `limit-severities-for-sarif` matters most, since `severity` does not apply to SARIF output without it.

On the next tag, after merge:

5. Both build jobs show a scan step and an upload step, and the job is green whatever the scan found.
6. Security -> Code scanning lists analyses under `trivy-infisical-core` and `trivy-infisical-fips`, attributed to `main`.
7. A nightly tag lands under the `-nightly` categories and leaves the stable alerts open.
8. For the failure path, point `image-ref` at a nonexistent digest on a branch build. The scan should fail, the upload should be skipped, the job should stay green, and the run should carry a warning annotation.

## Type

- [ ] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [x] Chore

## Checklist

- [ ] Tested locally


Vligai:
@greptileai review

.github/workflows/release-standalone-docker-img-postgres-offical.yml
```diff
@@ -189,6 +264,72 @@ jobs:
...
+                    echo "| $total | $crit | $cves |"
+                  } >> "$GITHUB_STEP_SUMMARY"
+                  if [ "$total" -gt 0 ]; then
+                    echo "::warning::infisical/infisical-fips: $total fixable CRITICAL/HIGH findings across $cves CVEs. See the trivy-infisical-fips artifact."
+                  fi
+            # A scan that breaks is otherwise indistinguishable from a clean one, since
+            # both leave the job green with no report attached.
+            - name: Flag an incomplete Trivy scan of infisical/infisical-fips
+              if: steps.trivy_infisical_fips.outcome == 'failure' || steps.trivy_upload_infisical_fips.outcome == 'failure'
+              run: echo "::warning::Trivy did not report on infisical/infisical-fips for this release. No report does not mean no findings."
```
victorvhs017, line 232: This step is exactly the same as the one above, we just change from infisical/infisical to infisical/infiical-fips. Can we make something reusable?

.github/workflows/release-standalone-docker-img-postgres-offical.yml
```diff
@@ -120,11 +126,79 @@ jobs:
...
                 args: --file=Dockerfile.standalone-infisical --project-name="infisical-core-docker-image"
+            # Reports into the run's artifacts, never through the job status: the
+            # image is already published by the time this runs.
+            - name: 🔍 Scan infisical/infisical with Trivy
+              id: trivy_infisical_core
+              continue-on-error: true
+              uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0, Apr 22, 2026
+              env:
+                  # The GHCR default rate-limits anonymous pulls from Actions egress IPs.
+                  TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db:2
```
victorvhs017: `TRIVY_DB_REPOSITORY` accepts a comma-separated list in priority order. We can leave GHCR as a fallback:

`TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db:2,mirror.gcr.io/aquasec/trivy-db:2`

.github/workflows/release-standalone-docker-img-postgres-offical.yml
```diff
@@ -120,11 +126,79 @@ jobs:
...
+                    echo "| $total | $crit | $cves |"
+                  } >> "$GITHUB_STEP_SUMMARY"
+                  if [ "$total" -gt 0 ]; then
+                    echo "::warning::infisical/infisical: $total fixable CRITICAL/HIGH findings across $cves CVEs. See the trivy-infisical-core artifact."
+                  fi
+            # A scan that breaks is otherwise indistinguishable from a clean one, since
+            # both leave the job green with no report attached.
+            - name: Flag an incomplete Trivy scan of infisical/infisical
+              if: steps.trivy_infisical_core.outcome == 'failure' || steps.trivy_upload_infisical_core.outcome == 'failure'
+              run: echo "::warning::Trivy did not report on infisical/infisical for this release. No report does not mean no findings."
```
victorvhs017, line 144: The summarize step can also fail, we should handle it here, too

victorvhs017 (changes requested):

.github/workflows/release-standalone-docker-img-postgres-offical.yml
Vligai: Extracted .github/actions/trivy-image-scan/action.yml

.github/workflows/release-standalone-docker-img-postgres-offical.yml
Vligai: Thanks! Fixed

.github/workflows/release-standalone-docker-img-postgres-offical.yml
Vligai: Good catch, fixed: all three steps now run inside one composite action, so if the summary fails the whole step is marked failed and the warning annotation still fires.

victorvhs017 (approved):

.github/workflows/release-standalone-docker-img-postgres-offical.yml
```diff
@@ -120,11 +126,79 @@ jobs:
...
+            - name: 📤 Keep the Trivy report for infisical/infisical
+              id: trivy_upload_infisical_core
+              if: steps.trivy_infisical_core.outcome == 'success'
+              continue-on-error: true
+              uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+              with:
+                  name: trivy-infisical-core-${{ steps.extract_version.outputs.version }}
+                  path: trivy-infisical-core.json
+                  retention-days: 90
+                  if-no-files-found: error
```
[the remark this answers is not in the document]
Vligai: will be addressed in a follow up PR

PR #8166, opened 2026-09-16
fix(secrets): improve search all folders scoping and results

andrewhuhh:
## Context

Resolves [SECRETS-668](https://linear.app/infisical/issue/SECRETS-668/improve-secrets-management-project-searchability).

- Scope Search All Folders by environments and a lazy-loaded folder tree. Folder selection is available only for one environment, includes descendants, and resets when environments change.
- Keep filters in an always-open inset rail, widened from 288px to 376px, with a shared reset action.
- Show the four resource types as checked-by-default checkboxes; unchecking all types shows no results.
- Render the metadata `is` operator as a separate static box before the value input.
- Remove the filter dropdown focus-restoration path that reopened its tooltip, and dismiss nested comboboxes with Escape without closing search.
- Keep results independently scrollable and preserve readable name columns on narrow screens.
- Show every secret metadata pair in a permanent Metadata column, with an em dash for resources without metadata.
- Open an interactive hover card for each metadata pair and append its “Apply Filter” action to the existing metadata conditions without adding duplicates.
- Add the permission-aware copy action to metadata-search results.
- Clarify the fixed metadata-search limit without exposing information about inaccessible secrets.
- Hide stale results while scope changes are pending.
- Clarify that the inline empty state only covers the current folder, preserve the table header and its actions, and offer a Search All Folders action that carries the current query into the global search.

## Screenshots

Wider filter rail with resource checkboxes and a separate metadata operator

Environment-scoped folder tree

Metadata hover card with Apply Filter

Current-folder empty state with Search All Folders

<details><summary>Narrow-screen layout</summary>

Narrow-screen search

</details>

## Steps to verify the change

1. Open Search All Folders and search by name across multiple environments.
2. Select one environment, expand the folder tree, and choose a folder. Confirm its descendants are included while siblings are excluded. Add another environment or clear the environment selection: the folder resets to `/` and the picker becomes disabled.
3. Confirm every result table includes the Metadata column, every metadata pair is visible, and resources without metadata show an em dash.
4. Hover a metadata pair and select **Apply Filter**. Confirm the pair is appended to the metadata conditions and an existing identical condition is not duplicated.
5. Add multiple metadata conditions and switch between ALL and ANY. Confirm the result set follows the selected matching mode.
6. Hover a tag badge, then copy a secret value without leaving search.
7. Change the environment or folder scope and confirm old results are replaced by a loading state until the scoped response arrives.
8. Toggle the four resource checkboxes, including unchecking all four. Confirm excluded types disappear. Dismiss the environment or folder picker with Escape; search must remain open.
9. Enter a query with no match in the current folder. Confirm the table header and its actions remain visible and the empty state says so, then select **Search All Folders** and confirm the global search opens with that query already submitted.
10. Reset filters and confirm scopes and metadata conditions clear, all four resource types are checked, and the text query remains.
11. Check a narrow viewport: the filter rail stacks above the independently scrollable results.

Validation for the filter refinements: `make reviewable-ui`, 19 related frontend tests, and 15 browser scenarios against the disposable local app covering recursive folder scope, environment changes, checkbox inclusion/exclusion, metadata scope, keyboard selection, Escape handling, request-error recovery, root selection, reset behavior, and desktop/narrow-screen geometry.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Title follows the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) format.
- [x] Tested locally


andrewhuhh:
@greptile review

andrewhuhh:
@greptile review

PR #8185, opened 2026-09-17
feat(pam): clickhouse

lb-vn:
## Context

Adds clickhouse support to PAM. HTTP protocol only. Web access and CLI. Credential blocking included.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -26,6 +26,7 @@ export enum TestConnectionMode {
   Kubernetes = "kubernetes",
   SSH = "ssh",
   Snowflake = "snowflake",
+  ClickHouse = "clickhouse",
```
bernie-g, line 29: Running `SELECT map('a', 'b') AS m` crashes the PAM web ui

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -103,6 +104,15 @@ export type TestConnectionRequest =
       schema?: string;
       role?: string;
     }
```
bernie-g, line 106: ignore comment placement

Image
Should we have a default role field like snowflake?

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -26,6 +26,7 @@ export enum TestConnectionMode {
   Kubernetes = "kubernetes",
   SSH = "ssh",
   Snowflake = "snowflake",
+  ClickHouse = "clickhouse",
   Tcp = "tcp"
 }
```
bernie-g, line 31: ignore comment placement

Image
Seems that the primary key badge randomly appears on certain columns

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -26,6 +26,7 @@ export enum TestConnectionMode {
   Kubernetes = "kubernetes",
   SSH = "ssh",
   Snowflake = "snowflake",
+  ClickHouse = "clickhouse",
   Tcp = "tcp"
```
bernie-g, line 30: From Claude:

Command blocking and session logging can't see query-parameter substitutions

ClickHouse's HTTP interface fills {name:Type} placeholders from param_name query parameters, so the value never appears in the statement text. inspect() in packages/pam/handlers/clickhouse/proxy.go builds the inspected statement from the query parameter and the body only, which has two consequences.

A blocking policy can be sidestepped when the placeholder is an Identifier. With the policy set to (?i)\busers\b:

# refused, as expected
curl 'http://127.0.0.1:<port>/' --data-binary 'SELECT count() FROM users'
→ Code: 497. DB::Exception: This statement is blocked by the command blocking policy on this account. (ACCESS_DENIED)

# same read, allowed
curl 'http://127.0.0.1:<port>/?param_t=users' --data-binary 'SELECT count() FROM {t:Identifier}'
→ 250

The session log has the same blind spot: it records the placeholder rather than the value, so the recording does not show which object a statement touched. This already applies to the explorer's own metadata queries, which log WHERE database = {schema:String} AND table = {table:String} with no values attached.

Grants remain the real access boundary, so this is not a privilege escalation. It does defeat the stated purpose of the policy, and it weakens the audit trail for every parameterized statement.

Suggested fix: fold the param_* values into the text inspect() returns, so the policy and the recording both see them. Sorting the names and appending one -- parameters: n=3 t=users line keeps existing anchored patterns working and makes the log show what was read.

backend/src/ee/services/pam-web-access/clickhouse/pam-clickhouse-client.ts
```diff
@@ -0,0 +1,133 @@
...
+const CONNECT_TIMEOUT_MS = 30 * 1000;
+
+export const MAX_RESULT_BYTES = 16 * 1024 * 1024;
+
+export const createRelayClient = (opts: {
+  relayPort: number;
+  database?: string;
+  sessionId?: string;
+}): ClickHouseClient =>
+  createClient({
```
bernie-g, line 16: Image
I don't understand this error msg on a view. Can we just say this table is read only?

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -593,6 +593,50 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+      sslRejectUnauthorized: z.boolean(),
+      sslCertificate: boundedOptionalString(16384)
+    }),
+    credentials: z.object({
+      username: z.string().trim().min(1).max(255),
+      password: boundedOptionalString(256)
+    }),
+    sanitizedCredentials: z.object({ username: z.string() }),
+    ui: {
+      port: {
```
bernie-g, line 613: ignore comment placement

Can we mention in the comments that clickhouse doesn't support rotation yet

backend/src/ee/services/pam-account/pam-account-schemas.ts
```diff
@@ -593,6 +593,50 @@ export const ACCOUNT_TYPE_CONFIGS = {
...
+        defaultValue: 8123,
+        tooltip:
+          "The HTTP interface port: 8123 for plain HTTP, 8443 for HTTPS. Sessions never use the native protocol on 9000."
```
bernie-g, line 616: ignore comment placement

Can we add clickhouse to the capabilities array in the gateway and then show an error on account creation if the gateway is too old

backend/src/ee/services/pam-account/pam-account-schemas.ts
lb-vn: What comments? Other account types that don't rotate don't have any special comments as far as I'm aware

backend/src/ee/services/pam-account/pam-account-schemas.ts
lb-vn: So we do have supported account types / resources being reported but they're only checked on session create if I'm not mistaken.

I think we need to extend the gateway heartbeat functionality or do some shared logic that can be reused across many types rather than doing it in this PR. Especially since this is on a deadline

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: Intentional. Clickhouse reports sorting key columns as "Primary Key". This is also documented in the clickhouse docs we added for this PR

backend/src/ee/services/pam-account/pam-account-connection-test.ts
lb-vn: For clickhouse its common to pin roles to users so this is unnecessary. I say we skip it to keep scope low and add it if users ask for it

backend/src/ee/services/pam-web-access/clickhouse/pam-clickhouse-connection-controller.ts
```diff
@@ -0,0 +1,230 @@
...
+    const queryId = randomUUID();
+    abortController = new AbortController();
+    runningQueryId = queryId;
+
+    const { stream, summary } = await client.exec({
+      query: statementSql,
+      query_id: queryId,
+      abort_signal: abortController.signal,
+      clickhouse_settings: {
+        default_format: "JSONCompact",
```
bernie-g, line 108: Can we set max_execution_time: 30s here

bernie-g, line 108: Using FORMAT seems to break everything

Image

backend/src/ee/services/pam-account/pam-account-schemas.ts
bernie-g: I meant in the docs. Freudian slip

lb-vn: Resolved in DMs

backend/src/ee/services/pam-account/pam-account-connection-test.ts
```diff
@@ -371,6 +381,29 @@ export const buildGatewayConnectionTest = async (
...
     }
+    case PamAccountType.ClickHouse: {
+      const cd = connectionDetails as {
+        database: string;
+        sslEnabled?: boolean;
+        sslRejectUnauthorized?: boolean;
+        sslCertificate?: string;
+      };
+      const c = creds as { username: string; password?: string } | null;
+      if (!c) return tcp(host, port);
```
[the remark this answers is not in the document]
lb-vn: Intentional

backend/src/ee/services/pam-web-access/clickhouse/pam-clickhouse-client.ts
```diff
@@ -0,0 +1,133 @@
...
+    );
+    this.name = "ClickhouseResultTooLargeError";
+  }
+}
+
+export const clickhouseErrorFields = (err: unknown): { message?: string; detail?: string } => {
+  if (err instanceof ClickHouseError) {
+    return { message: err.message, detail: err.type ? `${err.type} (code ${err.code})` : `Code ${err.code}` };
+  }
+  return { message: (err as Error)?.message };
```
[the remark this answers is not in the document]
lb-vn: This is intentional and part of the prduct. we forward them to the web explorer as we do for postgres and other accounts

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -556,6 +557,17 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
       throw new BadRequestError({ message: "A gateway must be attached to this account." });
     }
 
+    const attachedGateway = await gatewayV2DAL.findOne({ id: gatewayId });
+    const capabilities = attachedGateway?.capabilities as { supported_account_types?: string[] } | null;
+    if (!gatewaySupportsAccountType(accountType, capabilities?.supported_account_types)) {
+      throw new BadRequestError({
```
bernie-g, line 633: This will make the user fill out the entire account form only to then figure out the gateway doesn't support the account type. I feel like we should gray out the account type or throw an error before filling out the whole form instead

lb-vn: image

done

bernie-g (approved):

PR #8186, opened 2026-09-17
feat: secret scanning batching

mathnogueira:
## Context

A historical scan can take hours to run and are uses lots of resources. This PR batches commits in a repository so we can have "checkpoints" to resume from in case of a failing worker. 

## Steps to verify the change

1. Use a large repo (> 10k commits): Redis is a good one
2. Notice that during the scan, you see logs such as:
```
Full Scan Batch Complete [scanId=e40b916c-2b72-4622-a3a2-da33e041f1c9] [resourceId=71bf0007-846f-44de-adf7-705efd3db1a3] [dataSourceId=e77acfb5-0399-4be7-8dc2-e5bfca2a47fd] [jobId=e40b916c-2b72-4622-a3a2-da33e041f1c9] retryCount=[1/3] batch=[1/3] findings=[2] durationMs=[39092]
```
3. If you restart your worker (API if you are running everything in the same container), the job is picked up from where it left and finishes

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


mathnogueira:
@greptile review again

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-fns.ts
```diff
@@ -155,11 +155,135 @@ export const cloneRepository = async ({ cloneUrl, repoPath }: TCloneRepository):
   });
 };
 
-export async function scanDirectory(inputPath: string, outputPath: string, configPath?: string): Promise<void> {
+/**
+ * Commits are enumerated oldest-first and scanned in slices of that order, this list
+ * of commits will be used to create the batches of commits for historical scans.
+ */
+const COMMIT_LIST_ARGS = ["rev-list", "--full-history", "HEAD"];
```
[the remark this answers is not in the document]
mathnogueira: Fine for now that historical scans only scan the default branch.

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -191,26 +241,55 @@ export const secretScanningV2QueueServiceFactory = ({
...
+          scanningStartedAt: new Date(),
+          progressUpdatedAt: new Date()
         }
       );
 
+      if (!startedScans.length) {
+        logger.warn(
+          `secretScanningV2Queue: Full Scan skipped, scan was already closed out ${logDetails} [scanType=${SecretScanningScanType.FullScan}]`
+        );
+        return;
```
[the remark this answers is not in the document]
mathnogueira: We might need the event table that Thiago presented in the last eng demo.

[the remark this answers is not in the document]
mathnogueira: Will update the schema for the migration, so we can add this in.

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-fns.ts
```diff
@@ -155,11 +155,135 @@ export const cloneRepository = async ({ cloneUrl, repoPath }: TCloneRepository):
...
+  repoPath: string;
+  batchSize: number;
+  resumeAfterCommit?: string | null;
+  resumeAfterCommitDigest?: string | null;
+}): Promise<{
+  totalCommits: number;
+  batches: TCommitBatch[];
+  resumed: boolean;
+  /** Set when the resume point is still present but the history before it is not the history that was scanned. */
+  prefixChanged: boolean;
```
lobocv: nit: Seems like an error mode. In Go we return errors explicitly but I think the pattern in Node is to throw and catch an exception? Is there a reason why we return a boolean here instead of that pattern? (just curious for my own learning, im totally fine with this approach too)

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -88,6 +94,36 @@ export const secretScanningV2QueueServiceFactory = ({
...
 }: TSecretRotationV2QueueServiceFactoryDep) => {
+  /**
+   * Full scans of one resource are serialized across pods by a lease in Redis holding the ID of the
+   * scan that owns it. A retry or a stalled-job recovery carries the same scan ID, so it reclaims
+   * its own lease and resumes where it left off; a second, distinct scan of the same resource finds
+   * the lease held and gives up rather than cloning the repository alongside the first. Renewing is
+   * the same operation as acquiring, and returns false once the lease has been lost.
+   */
+  const acquireFullScanLease = async (resourceId: string, scanId: string) => {
+    const key = KeyStorePrefixes.SecretScanningFullScanLease(resourceId);
```
lobocv, line 105: What is resource here? Is it an abstraction over the thing being scanned (eg repo)? Is it to prep for scanning other things that are not repos?

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -191,26 +241,55 @@ export const secretScanningV2QueueServiceFactory = ({
...
+        return;
       }
 
-      await secretScanningV2DAL.scans.update(
-        { id: scanId },
+      // Guarded so a retry cannot reopen a scan that already reached a terminal status. The work
+      // itself succeeded and only a side effect after it (a notification enqueue, an audit log)
+      // threw, so BullMQ hands the whole handler back; without this an already-Completed scan would
+      // be reset to `scanning` and, on the last attempt, closed out as Failed.
+      const startedScans = await secretScanningV2DAL.scans.update(
```
lobocv: nit: IMO this should be a method on the secret scanning DAL so that we're not leaking SQL logic into the application layer (service). It should be named in application-terms for example `setScansInProgress`. That would make this function read much clearer

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -191,26 +241,55 @@ export const secretScanningV2QueueServiceFactory = ({
...
 
     try {
-      try {
-        lock = await keyStore.acquireLock(
-          [KeyStorePrefixes.SecretScanningLock(dataSource.id, resource.externalId)],
-          getFullScanLockTtlMs()
+      holdsLease = await acquireFullScanLease(resourceId, scanId);
+
+      if (!holdsLease) {
+        await secretScanningV2DAL.scans.update(
```
lobocv: same comment here, this is infrastructure level concerns leaking into the app layer. It should be behind a DAL method that abstracts the SQL query

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -241,66 +320,155 @@ export const secretScanningV2QueueServiceFactory = ({
...
+       * Each batch is made durable on its own: its findings and the commit it reached are committed
+       * before the next `infisical scan` starts, so a worker killed mid-scan resumes from there
+       * rather than re-walking history it has already paid for. Returns whether this run still owns
+       * the scan and its lease.
+       */
+      const persistBatch = async (
+        batchFindings: TFindingsPayload,
+        resumePoint?: { lastScannedCommit: string; lastScannedCommitDigest: string }
+      ) => {
+        const owned = await secretScanningV2DAL.findings.transaction(async (tx) => {
```
lobocv, line 358: same here

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -241,66 +320,155 @@ export const secretScanningV2QueueServiceFactory = ({
...
+          );
+
+          return Boolean(progressed.length);
+        });
+
+        if (!owned) return false;
+
+        // Renewed on the same beat progress is recorded, so the lease outlives a scan of any length
+        // without a timer of its own. Losing it means another scan of this resource has taken over.
+        holdsLease = await acquireFullScanLease(resourceId, scanId);
```
lobocv, line 389: nit: One could argue that the act of acquiring the least could be inside the DAL and encapsulated in the methods that put the scan in progress / update progress.

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -241,66 +320,155 @@ export const secretScanningV2QueueServiceFactory = ({
...
+        }
+
+        return holdsLease;
+      };
+
+      let stillOwned = true;
+
       switch (resource.type) {
         case SecretScanningResource.Repository:
         case SecretScanningResource.Project: {
```
lobocv, line 402: nit: These large switch statements are really painful to read. It would be nice to move the concents of them into separate functions

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -241,66 +320,155 @@ export const secretScanningV2QueueServiceFactory = ({
...
+        if (!holdsLease) {
+          logger.warn(`secretScanningV2Queue: Full Scan lost its lease on the resource ${logDetails}`);
+        }
+
+        return holdsLease;
+      };
+
+      let stillOwned = true;
+
       switch (resource.type) {
```
lobocv, line 400: Not for this PR: these switch statements are code smells to me. They don't scale well. We should think more about what the interface is between the application and the thing we are scanning and define implementations for each. This makes the code more modular and testable. We should register the implementation on startup with dependency injection. 
However, this is not a pattern i see in the codebase so it would be a big change to introduce it now. This is how I will be pushing things to be when we move to Go

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
```diff
@@ -241,66 +320,155 @@ export const secretScanningV2QueueServiceFactory = ({
...
+            batchSize,
+            resumeAfterCommit: scan.lastScannedCommit,
+            resumeAfterCommitDigest: scan.lastScannedCommitDigest
+          });
+
+          // Two ways a resume point stops meaning anything: a rewritten history (force push, or a
+          // rebase landing between runs) takes the commit out of the repository entirely, or a newly
+          // reachable ref puts commits ahead of it that this scan has never looked at. Either way the
+          // repository is re-walked from the start rather than resumed past commits nobody scanned.
+          if (scan.lastScannedCommit && !plan.resumed) {
```
lobocv: If the logic for scanning a Project was encapsulated behind some sort of object, this sort of implementation leak wouldn't show up in the application layer. You would instead be asking that object `project.CanResume()` and wouldnt need to know the details of "lastScannedCommit".

Later on if that logic changes, we only change it on the project itself (which is really the thing that decides resumability)


I would highly recommend moving all this logic under a Project class and keeping the behaviours close to it rather than in the application layer

lobocv:
Some minor feedback:
1. Perhaps the page should poll for updates because I came back to it after a long time and it still showed "scanning" but after a refresh it said it was done
2. I think the sort on status is FE only because I marked two findings as "Resolved" but can no longer see them. When I hit sort by status it still only ever shows me Unresolved issues. I did see that the Filter button does work though. I think moving the sort into the backend is a potentially bigger task and we have likely have the same problem in other parts of the UI. It's okay to defer IMO given we have the filter button.
3. I'm not sure what the Search findings bar is searching on. Doesn't seem to pull up anything when I use it.  I would expect it to search on things like the file path, commit hash, author, description, commit message

Overall it was really great. Good work @mathnogueira !

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
mathnogueira: it's a `secret_scanning_resource` we will rename those in the revamp. And yes, we plan to scan more thing other than repos.

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
mathnogueira: DALs are only for DB, this is acquiring the lease from Redis.

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
mathnogueira: I agree, we will fix this in the revamp. Will not do it in this PR, taking notes

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
mathnogueira: same, will be done in a separated PR

backend/src/ee/services/secret-scanning-v2/secret-scanning-v2-queue.ts
mathnogueira: Now everything is inside `planCommitBatches` instead.

I like this approach of having domain objects, I'm considering moving the implementation to use more of it as it would simplify a few things.

lobocv (approved):
Looks great. Nice readability improvements!

PR #8194, opened 2026-09-17
feat(secret-rotation): add Stripe app connection and API key rotation

lobocv:
## Context

[ENG-5685](https://linear.app/infisical/issue/ENG-5685/add-stripe-api-key-secret-rotation-integration). This started as the Stripe app connection alone and now carries the rotation provider too. The connection sections below are unchanged; the rotation sections were added when that work landed on this branch.

Stripe does not let any API key create another API key. The only principal that can is an installed Stripe App holding the `api_key_write` permission, so Stripe API key rotation requires Infisical to be a Stripe app vendor. That is why this is more than a normal app connection: it ships the app manifest too.

### How auth works, and what is deliberately not stored

Infisical authenticates with **its own** Stripe secret key and names the customer's account with the `Stripe-Context` header. The tokens the OAuth exchange returns are therefore never used, and are not persisted. The exchange still runs, because it is the only proof that the installer controls the account they are claiming — a client-supplied account ID would be unverified.

The connection stores one thing: the customer's `acct_...`. That means there is no access token to expire, nothing to refresh, and no rolled-refresh-token hazard. Revocation happens when the customer uninstalls the app.

### Validation probes the real thing

`validateStripeConnectionCredentials` calls `GET /v2/iam/api_keys` with the same credentials rotation will use, so a connection that cannot manage keys fails at creation rather than silently at rotation time. This has been confirmed working against a gated sandbox.

### Three env vars, not two

Stripe embeds a per-app, per-mode channel link ID in the install URL path (`/oauth/v2/chnlink_.../authorize`). It cannot be derived from the client ID, so the authorize URL is configuration. A sandbox link paired with a live key will fail at the exchange, so the link and the key must be for the same mode.

## Steps to verify the change

Requires a Stripe sandbox that has been gated into the Managed API Keys private preview. See `stripe-app/README.md` and the [setup doc](https://app.notion.com/p/infisical/Stripe-secret-rotation-setup-3d8564692229802f9082f2aa848b5eea); ask Daniel for sandbox access rather than gating a new one.

1. Point `allowed_redirect_uris` in `stripe-app/stripe-app.json` at your HTTPS dev origin, bump `version`, and `stripe apps upload`. Set External test version to the new version.
2. Set `SITE_URL` to that same origin, plus `INF_APP_CONNECTION_STRIPE_OAUTH_CLIENT_ID`, `INF_APP_CONNECTION_STRIPE_OAUTH_AUTHORIZE_URL` (copied from the dashboard's External test tab) and `INF_APP_CONNECTION_STRIPE_SECRET_KEY`. Recreate the containers; `restart` does not re-read `env_file`.
3. Reach the UI through that origin, not `localhost`, or the `redirect_uri` will not match the manifest.
4. Organization → App Connections → Add Connection → Stripe → Connect. Approve on Stripe.
5. The connection is created and its `credentials.accountId` is the `acct_...` you installed into.

### Rotation, added on top of the connection

Rotation is create-then-retire. On every rotation we mint a new restricted key and expire the one from two rotations ago. We never call Stripe's `rotate` endpoint, because it kills the old token instantly and would leave running applications holding a dead credential. The framework's two-slot credential model supplies the handover window instead, so two keys are deliberately live at once.

Stripe encrypts a newly minted key's secret to a public key we supply, so each create generates a throwaway RSA-2048 keypair and decrypts the returned JWE in process. The private key is never stored and never reused.

Three things worth a reviewer's attention:

- Stripe's list endpoint returns every key's secret in full plaintext, for every key in the account. The new `/api-keys` endpoint strips it at three layers (inside the list function, again in the service mapping, and again in the route's allowlist response schema). The layering is deliberate, so no future caller can reach a raw object.
- The pagination loop follows a URL out of Stripe's response body while carrying our platform key, which is shared across every tenant. It now checks that URL points at the Stripe API keys endpoint before the credential goes out.
- Stripe has no wildcard permission and no endpoint that lists the permission names, so the 158 names are maintained by hand and ride on the rotation's list option for the form to read. Nothing in the build can validate that list.

### Steps to verify the rotation

Automated, all run locally: `make reviewable-api`, `make reviewable-ui` and `make lint-docs-branch` are clean, and `make test-api-unit SPEC=stripe` passes 36 tests across 4 files.

The full `make test-api-unit` run is red with 8 files and 9 tests failing (Ed25519 under FIPS, PQC key sizes, a timing-sensitive process group test, and three collection errors). Every one of those files has zero commits on this branch and the same set fails before and after the merge of main, so they are inherited rather than introduced here.

Against a gated sandbox, still outstanding:

1. Create one rotation and inspect the raw create response. Record whether `secret_key.token` is present, whether `secret_key.encrypted_secret.ciphertext` is present, and whether both are. `readStripeSecret` prefers `token`, so if Stripe returns both then the rotation succeeds down the plaintext path and never exercises the JWE code. A working rotation is not evidence the encrypted path works.
2. Confirm a key appears in the dashboard named `infisical-<secret name>-<timestamp>`.
3. Confirm the mapped secret holds an `rk_` value that authenticates. Do not assert on an `sk_` prefix.
4. Rotate twice and confirm the first key is expired while the second stays live.
5. Delete the rotation with revoke and confirm both keys expire.
6. Confirm the copy-from-an-existing-key control prefills the permission picker.
7. In devtools, confirm the `api-keys` response carries no `secret_key` field.
8. Expire a key, then GET it, and record the status. See the open questions below.
9. Record the exact `status` values Stripe returns in the list response.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally

## Before this leaves draft

- [ ] `stripe-app.json` still holds one developer's ngrok host. Needs a shared redirect URI story. This is now sharper than when it was written: the root `CLAUDE.md` and `stripe-app/README.md` both present that manifest as authoritative, and the new connection page tells cloud users both prerequisites are already in place. Either the production app lives elsewhere and those three claims are misleading, or this cannot work in production as committed.
- [x] Docs page (`docs/integrations/app-connections/stripe.mdx`) and `docs.json` entries. Also the rotation page, eight API reference pages, the rotations browser entry, and the three `INF_APP_CONNECTION_STRIPE_*` vars in `envars.mdx`.
- [x] Tests. 36 unit tests across the JWE decryption, the public client, the connection service and the rotation factory.
- [ ] The production app should be published from a **non-Connect** account. The dev sandbox is a Connect platform, and publishing a public app from one only works via an undocumented private-then-public workaround that Stripe has not confirmed is supported.
- [ ] App review, required for marketplace listing before Cloud.
- [ ] Sandbox verification of the rotation, items 1 to 9 above.

## Open questions rotation depends on

- [x] Which JWE `alg` Stripe uses when returning a new secret, and whether the FIPS image can decrypt it. Answered: the FIPS image decrypts both `RSA-OAEP` (SHA-1) and `RSA-OAEP-256`, verified under `--force-fips` with Node 22.23.2 and OpenSSL 3.5.7. The suite asserts `crypto.getFips()` is 1 so a toolchain bump cannot break this silently.
- [x] Whether `rotate` leaves any grace period. Moot: rotation never calls `rotate`, it creates then retires, so the handover window comes from our own two-slot model rather than from Stripe.
- [ ] Whether `expire` takes effect immediately or after a delay. Still unanswered. Nothing we ship or document claims a retired key is immediately dead.

Two further premises are unobserved and shipped code depends on both:

- [ ] `$tryRetireKey` treats a 404 on retrieve as proof a key is already gone, and that is the whole mechanism stopping a rotation wedging permanently after a commit failure. Nobody has confirmed an expired key actually 404s on retrieve. Retained 404-on-expire handling means this is never worse than before, so the honest claim today is "no worse", not "closed". Verification item 8.
- [ ] The copy-from-key control filters on `status === "active"`. That literal came from our own fixtures, not from a probe of Stripe's real status vocabulary. If it is wrong the control never renders. Verification item 9.


docs/superpowers/plans/2026-09-18-stripe-api-key-rotation.md
lobocv, line 1: Remove superpowers design specs and plans from the PR

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-jwe.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    const { publicKey, privateKey } = await generateStripeEncryptionKeyPair();
+
+    for (const alg of ["RSA-OAEP", "RSA-OAEP-256"]) {
+      const jwe = encryptJwe("rk_live_secret", publicKey, { alg });
+      expect(decryptStripeJwe(jwe, privateKey)).toBe("rk_live_secret");
+    }
+  });
+
+  it("rejects a JWE without exactly five segments", async () => {
```
lobocv, line 52: Are any of these tests low value or redundant given that generateStripeEncryptionKeyPair is just a light wrapper around the crypto library?

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.test.ts
```diff
@@ -0,0 +1,304 @@
+/* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-argument */
```
lobocv: Are these tests valuable given that they are mocking the stripe API? These tests are hard to read. Improve the readability and remove any low value tests

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.ts
```diff
@@ -0,0 +1,233 @@
...
+      }
+
+      return { retired: false, error: expireError };
+    }
+  };
+
+  /**
+   * Expire is the whole retirement. Stripe decides when it takes effect, so nothing here claims the
+   * key is dead. Narrowing the key's permissions first would stop it sooner, and was measured
+   * working, but is deliberately not in v1. See the design doc.
```
lobocv: remove comment about v1 and design doc. Say that expiry doesn't happen right away but usually within several minutes, which is good enough for us given that rotations happen on the time scale of weeks to months

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.ts
```diff
@@ -0,0 +1,233 @@
...
+          throw new BadRequestError({
+            message: `Stripe API key ${credentialsToRevoke.keyId} could not be retired (${getStripeErrorMessage(retireResult.error)}), and the newly created key ${newCredentials.keyId} could not be cleaned up either (${getStripeErrorMessage(cleanupResult.error)}). Both may need to be removed manually from the Stripe dashboard.`
+          });
+        }
+
+        throwStripeApiKeyManagementError(accountId, retireResult.error);
+      }
+    }
+
+    return $retireOnFailure(newCredentials.keyId, () => callback(newCredentials));
```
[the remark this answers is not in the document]
lobocv: Pushing back on this one. The scenario is a Postgres commit that durably succeeds but whose acknowledgement is lost, after which `$retireOnFailure` expires the key the committed row now points at. Real, but two things make it the wrong trade to fix:

At the call site you cannot tell a lost ack apart from an ordinary rollback, so the only way to avoid the rare bad case is to stop cleaning up at all. That leaks a live Stripe API key on every normal commit failure, which is the exact thing `$retireOnFailure` was added for.

Every other rotation factory already does the unguarded version. `awsIamUserSecretRotationFactory` is a bare `return callback(credentials)` with no cleanup on any failure path. So this is strictly safer than the house pattern, not less safe.

Leaving as is.

frontend/src/pages/organization/AppConnections/OauthCallbackPage/OauthCallbackPage.tsx
```diff
@@ -605,6 +607,56 @@ export const OAuthCallbackPage = () => {
...
+        });
+      }
+    } catch {
+      navigate({
+        to: returnUrl,
+        params: {
+          projectId
+        }
+      });
+      return null;
```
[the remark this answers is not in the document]
lobocv: Fair observation, but it is not specific to Stripe. `handleStripe` is a line-for-line copy of `handleHeroku`, silent catch included, and a cancelled authorization dead-ends the same way for Heroku, GitLab, GitHub and all three Azure connections on this page.

Fixing it properly means one change across every handler so the behaviour stays consistent, which belongs in its own PR rather than leaving Stripe as the only connection that explains itself. Not changing it here.

stripe-app/stripe-app.json
```diff
@@ -0,0 +1,20 @@
...
+  "extensions": null,
+  "permissions": [
+    {
+      "permission": "api_key_write",
+      "purpose": "Lets Infisical create, rotate and expire API keys in your account so your Stripe credentials can be rotated automatically."
+    }
+  ],
+  "allowed_redirect_uris": [
+    "https://provable-triceps-maximum.ngrok-free.dev/organization/app-connections/stripe/oauth/callback"
+  ],
```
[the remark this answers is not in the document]
lobocv: Known and documented, see the Redirect URIs section of `stripe-app/README.md`. Stripe requires HTTPS on every entry so `localhost` is not an option, the manifest is not editable in the dashboard, and a shared redirect URI every developer can use is still an open decision.

This is a deploy prerequisite rather than a code defect: the manifest is uploaded by hand with `stripe apps upload`, so the production URI goes in at that point along with a version bump. Not changing the checked-in manifest here.

backend/src/services/app-connection/stripe/stripe-connection-schemas.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const CreateStripeConnectionSchema = ValidateStripeConnectionCredentialsSchema.and(
+  GenericCreateAppConnectionFieldsSchema(AppConnection.Stripe)
+);
+
+export const UpdateStripeConnectionSchema = z
+  .object({
+    credentials: StripeConnectionOAuthCredentialsSchema.optional().describe(
+      AppConnections.UPDATE(AppConnection.Stripe).credentials
+    )
```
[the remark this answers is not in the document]
lobocv: Fixed in 73d9ca865d. The update path now rejects a reconnect whose exchanged account ID differs from the stored one, naming both accounts in the message. Pointing a connection at a different account means creating a new connection.

The failure chain checked out rather than being theoretical. `UpdateStripeConnectionSchema` accepts an optional OAuth code, so a PATCH with a fresh code is a supported path and not delete-and-create only, and `$tryRetireKey` treats a 404 from expire as already retired (`stripe-api-key-rotation-fns.ts:73`), which is exactly what a key ID minted in the old account returns in the new one.

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-jwe.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+  }
+
+  if (header.enc !== SUPPORTED_CONTENT_ALGORITHM) {
+    throw new BadRequestError({
+      message: `Stripe encrypted the secret with an unsupported content algorithm '${header.enc ?? "none"}'. Infisical supports ${SUPPORTED_CONTENT_ALGORITHM}.`
+    });
+  }
+
+  const decodedAuthTag = Buffer.from(authTag, "base64url");
```
varonix0: (whole file) this might be overkill. if this breaks its on Stripe's side. the encryption here is very well defined

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-constants.ts
```diff
@@ -0,0 +1,183 @@
...
+  "top_up_read",
+  "top_up_write",
+  "transaction_read",
+  "transfer_read",
+  "transfer_write",
+  "treasury_transaction_read",
+  "usage_record_read",
+  "usage_record_write",
+  "webhook_read",
+  "webhook_write"
```
varonix0: did we double check that these are indeed the available ones? my PoC might've missed some

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-constants.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export const STRIPE_API_KEY_ROTATION_LIST_OPTION: TSecretRotationV2ListItem = {
+  name: "Stripe API Key",
+  type: SecretRotation.StripeApiKey,
+  connection: AppConnection.Stripe,
+  template: {
+    secretsMapping: {
+      apiKey: "STRIPE_API_KEY"
+    },
+    // The frontend picker reads the permission names from here. It cannot import this file, since
```
varonix0: "It cannot import this file, since
    // the frontend is a separate package, and Stripe has no endpoint that lists them." 

i dont think we need to say this

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.test.ts
```diff
@@ -0,0 +1,297 @@
...
+  describe("checkActiveCredentials", () => {
+    it("treats 403 as a live key and 401 as a dead one", async () => {
+      getMock.mockRejectedValueOnce(httpError(403, "Insufficient permissions"));
+      await expect(makeFactory().checkActiveCredentials!(NEW_KEY)).resolves.toBeUndefined();
+
+      getMock.mockRejectedValueOnce(httpError(401, "Invalid API Key"));
+      await expect(makeFactory().checkActiveCredentials!(NEW_KEY)).rejects.toThrow("Invalid API Key");
+    });
+  });
+});
```
varonix0: entire file: seems overkill when we have the fakes e2e tests perhaps?

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-schemas.ts
```diff
@@ -0,0 +1,69 @@
...
+  .min(1)
+  .max(2);
+
+const StripeApiKeyRotationParametersSchema = z.object({
+  permissions: StripeApiKeyPermissionSchema.array()
+    .min(1, "At least one permission is required")
+    .max(STRIPE_API_KEY_PERMISSIONS.length)
+    .describe(SecretRotations.PARAMETERS.STRIPE_API_KEY.permissions),
+  connectPermissions: StripeApiKeyPermissionSchema.array()
+    .max(STRIPE_API_KEY_PERMISSIONS.length)
```
varonix0: should we have a zod refine to we can remove duplicate permissions from both permissions and connect permissions?

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-schemas.ts
```diff
@@ -0,0 +1,69 @@
...
+    keyId: z.string(),
+    apiKey: z.string()
+  })
+  .array()
+  .min(1)
+  .max(2);
+
+const StripeApiKeyRotationParametersSchema = z.object({
+  permissions: StripeApiKeyPermissionSchema.array()
+    .min(1, "At least one permission is required")
```
varonix0: connect permissions is optional but regular account permissions are required. its not possible to only pass connect permissions without normal permissions?

backend/src/lib/api-docs/constants.ts
```diff
@@ -3648,6 +3648,12 @@ export const SecretRotations = {
       username:
         "The Snowflake user whose RSA key pair will be rotated. If the user does not exist, it is created as a key-pair-only SERVICE user.",
       modulusLength: "The modulus length in bits of the generated RSA key pairs. Defaults to 2048."
+    },
+    STRIPE_API_KEY: {
+      permissions:
+        "The permissions granted to the generated Stripe API key. Stripe has no wildcard permission, so this is the full list of what the key may do.",
+      connectPermissions:
+        "The permissions the generated Stripe API key has over connected accounts. Only meaningful when the account is a Connect platform."
```
varonix0: nit: I would rephrase as:

"The permissions that the generated Stripe API key has over connected accounts. If your Stripe account is not a Connect platform account, you cannot configure Connect permissions for the key."

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.ts
```diff
@@ -0,0 +1,242 @@
...
+        throw new BadRequestError({
+          message: `${getStripeErrorMessage(actionError)} The newly created Stripe API key ${keyId} could not be expired and may need to be removed manually: ${getStripeErrorMessage(cleanupError)}`
+        });
+      }
+
+      throw actionError;
+    }
+  };
+
+  const $createApiKey = async () => {
```
varonix0: When we create new keys, I think we should check:

1. Has the user configured connectPermissions?
2. If yes, use the Stripe connection `validateStripeConnectionCredentials` to get the account type. If the account type is not a stripe connect (express, custom, or standard), if the user's account is not stripe connect, throw an error early so they won't get confused later

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.ts
```diff
@@ -0,0 +1,242 @@
...
+    // The published key is retired last. A failure part way through aborts the delete, so retiring
+    // it first would leave the rotation row and its mapped secret in place pointing at a key that
+    // is on its way out, which is an outage until the operator retries.
+    const { activeIndex } = secretRotation;
+    const retirementOrder = [
+      ...credentials.filter((_, index) => index !== activeIndex),
+      ...credentials.filter((_, index) => index === activeIndex)
+    ];
+
+    for (const { keyId } of retirementOrder) {
```
varonix0: nit: you can do `for await (const ...) {}` and get rid of the eslint disable

backend/src/server/routes/v1/app-connection-routers/stripe-connection-router.ts
```diff
@@ -0,0 +1,61 @@
...
+      rateLimit: readLimit
+    },
+    schema: {
+      operationId: "listStripeApiKeys",
+      params: z.object({
+        connectionId: z.string().uuid()
+      }),
+      response: {
+        200: z.object({
+          // An allowlist, deliberately. Stripe's own list response carries every key's plaintext
```
varonix0: no need to say this, this is understood broadly

backend/src/services/app-connection/stripe/stripe-connection-fns.ts
```diff
@@ -0,0 +1,96 @@
...
+  const { INF_APP_CONNECTION_STRIPE_OAUTH_CLIENT_ID, INF_APP_CONNECTION_STRIPE_OAUTH_AUTHORIZE_URL } = getConfig();
+
+  return {
+    name: "Stripe" as const,
+    app: AppConnection.Stripe as const,
+    methods: Object.values(StripeConnectionMethod) as [StripeConnectionMethod.OAuth],
+    oauthClientId: INF_APP_CONNECTION_STRIPE_OAUTH_CLIENT_ID,
+    // Stripe embeds a per-app, per-mode channel link ID in the install URL path, so it cannot be
+    // derived from the client ID and has to be copied out of the Stripe dashboard.
+    oauthAuthorizeUrl: INF_APP_CONNECTION_STRIPE_OAUTH_AUTHORIZE_URL
```
varonix0: This env var actually won't be that intuitive I think. As an example, A stripe app can have multiple redirect URL's (like we'll have 1 app that'll work for both EU and US cloud). Asking users to paste the literal authorize URL with the actual redirect URL baked in is not ideal. I think we need to derive the redirect_url parameter from the side URL instead of betting that the user will get it right

varonix0: Ideally they would be able to provide just an app ID and we can compute the rest of the URL. For live-mode stripe apps, the URL is predictable from the Stripe App's ID. It's only for sandboxes where we can't determine the oauth URL

backend/src/services/app-connection/stripe/stripe-connection-public-client.ts
```diff
@@ -0,0 +1,152 @@
...
+      message: "Stripe is not configured on this instance. Set the Stripe App Connection credentials to enable it."
+    });
+  }
+
+  return INF_APP_CONNECTION_STRIPE_SECRET_KEY;
+};
+
+/** Infisical acting on a customer's account: our own key, with Stripe-Context naming the account. */
+export const getStripePlatformRequestConfig = (accountId: string): AxiosRequestConfig => ({
+  auth: { username: getStripeSecretKey(), password: "" },
```
varonix0: huh, is this the documented way to go about it? why not just pass `Authorization: Bearer <secret-key>`?

backend/src/services/app-connection/stripe/stripe-connection-public-client.ts
```diff
@@ -0,0 +1,152 @@
...
+    throw new InternalServerError({
+      message: "Stripe is not configured on this instance. Set the Stripe App Connection credentials to enable it."
+    });
+  }
+
+  return INF_APP_CONNECTION_STRIPE_SECRET_KEY;
+};
+
+/** Infisical acting on a customer's account: our own key, with Stripe-Context naming the account. */
+export const getStripePlatformRequestConfig = (accountId: string): AxiosRequestConfig => ({
```
varonix0: also i dont think we should refer to ourselves as the "stripe platform", someone might read this in the future and take it out of context and read it like we are a Stripe Connect platform

backend/src/services/app-connection/stripe/stripe-connection-service.test.ts
```diff
@@ -0,0 +1,101 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
```
varonix0: same here as for earlier tests, you think we need these if we're doing fakes? at the e2e layer?

docs/api-reference/endpoints/secret-rotations/stripe-api-key/create.mdx
```diff
@@ -0,0 +1,77 @@
...
+openapi: "POST /api/v2/secret-rotations/stripe-api-key"
+---
+
+<Note>
+    Check out the configuration docs for [Stripe API Key Rotations](/documentation/platform/secret-rotation/stripe-api-key) to learn how to obtain the required parameters.
+</Note>
+
+Creates a Stripe API key secret rotation. Infisical will create and expire restricted API keys via your [Stripe connection](/integrations/app-connections/stripe) and write them to the specified secret path.
+
+## Request body
```
varonix0: we don't define these manually. when we set the `openapi` attribute, it will auto polulate from our API spec

docs/documentation/platform/secret-rotation/stripe-api-key.mdx
```diff
@@ -0,0 +1,121 @@
...
+
+<Note>
+  **Rotation Type: Dual-Phase**
+
+  This rotation maintains two active credential sets with overlapping validity, ensuring zero-downtime during rotation cycles.
+</Note>
+
+Stripe API key rotation mints a new [restricted key](https://docs.stripe.com/keys#limit-access-with-restricted-api-keys) on the interval you set, so your applications always have a working credential without a manual key swap. A few things about how it behaves are worth knowing before you set it up.
+
+- **The minted key's mode follows your Stripe account's own key.** If your Stripe account is in test mode, every key this rotation generates is a test-mode key. Your account decides the mode; the rotation doesn't, and there's no setting to override it.
```
varonix0: hmm do we need to state this? seems pretty clear that an api key created in test mode or a sandbox, the created api key will be tied to the that account

docs/self-hosting/configuration/envars.mdx
```diff
@@ -131,7 +131,7 @@ Example values:
   Defaults to `false` so an operator-configured forward proxy keeps working.
 
   <Note>
-    Enable this only if your instance does **not** rely on an outbound
```
varonix0: why this change? personally prefer 'does not'

docs/self-hosting/configuration/envars.mdx
```diff
@@ -283,7 +283,7 @@ and raising it is the right move.
 <Tip>
   Prefer scaling reads with `DB_READ_REPLICAS` over enlarging the primary pool. A connection pooler such as
   PgBouncer or RDS Proxy is only worth the extra hop and failure domain once instance count alone pushes you
-  near `max_connections`, and note that a pooler in transaction mode does **not** reclaim connections held
+  near `max_connections`, and note that a pooler in transaction mode **doesn't** reclaim connections held
```
varonix0: same

docs/snippets/RotationsBrowser.jsx
```diff
@@ -239,6 +239,15 @@ export const RotationsBrowser = () => {
       category: "AI & LLM",
       rotationType: "Dual-Phase",
     },
+    {
+      name: "Stripe API Key",
```
varonix0: we forgot api reference for stripe app connection, and for the app connection browser as we did here for rotations

stripe-app/.gitignore
```diff
@@ -0,0 +1,4 @@
+/node_modules
```
varonix0: lets not put this in the main repo I think. lets create a new github repo for this, the infisical/infisical one is already very large

CLAUDE.md
```diff
@@ -23,6 +23,7 @@ infisical/
 ├── frontend/              # React 18 SPA (see frontend/CLAUDE.md)
 ├── wasm/                  # Rust crates compiled to WASM for the frontend (see wasm/<crate>/CLAUDE.md)
 ├── e2e/                   # External Playwright suite — gates prod deploys against gamma (see e2e/CLAUDE.md)
+├── stripe-app/            # Stripe App manifest granting Infisical api_key_write (see stripe-app/README.md)
```
varonix0: remove here too

varonix0 (changes requested):

backend/src/lib/api-docs/constants.ts
varonix0: see comment further down for more details

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.ts
varonix0: update: we can update `validateStripeConnectionCredentials` to also get the account type when we get the account ID. a stripe account can never be changed back to a "normal" stripe account after it becomes a stripe connect platform, so we only need to save it once and we can assume that the account type will never change

varonix0: when we do this we can also check on UI side and disable the input field for connect permissions

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.ts
lobocv: I think we want to retire in a specific order here (inactive first). If we did that wouldn't it send both requests concurrently? It's a bit safer this way because if it fails while the published key is tired, we're in a broken state

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-schemas.ts
lobocv: Good catch, this should require at least one permission between `permissions` and `connectPermission`

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-schemas.ts
lobocv: We could add deduplication within each array but i dont know if it's really necessary. The UI at least won't be sending duplicates. If it's easy to do in zod then i can add it no problem. If it involves writing deduplication code in the API I wouldn't say it's worth doing.

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-constants.ts
lobocv: This was derived from the restricted key permission set but ill check again just to be sure

backend/src/ee/services/secret-rotation-v2/stripe-api-key/stripe-api-key-rotation-fns.test.ts
lobocv: I'm sure there is some overlap with those tests but i think there are a few things that the E2E tests don't cover as easily (retirement ordering which is critical). I'll remove the redundancies.

docs/api-reference/endpoints/secret-rotations/stripe-api-key/create.mdx
lobocv: Hmm i thought this would have been done properly by our docs skill... not sure why it missed that

docs/self-hosting/configuration/envars.mdx
lobocv: yes this shouldn't have been changed. I too prefer "does not". ill revert it

stripe-app/.gitignore
lobocv: yeah i was meaning to ask you about that. This didn't seem like the right place. ill make a new repo

PR #8196, opened 2026-09-17
feat: built-in PAM masking

sheensantoscapadngan:
## Context
This PR adds support for built-in PAM masking

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -238,6 +244,66 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
...
+    const gatewayIds = [
+      ...new Set([
+        ...candidates.filter((c) => c.enabled && c.gatewayId).map((c) => c.gatewayId!),
+        ...[...poolGatewayIds.values()].flat()
+      ])
+    ];
+    const gatewayIdSet = new Set(gatewayIds);
+    const orgGateways = gatewayIdSet.size ? await gatewayV2DAL.find({ orgId }) : [];
+    const gateways = orgGateways.filter((g) => gatewayIdSet.has(g.id));
+    const unsupported = new Set(
```
sheensantoscapadngan, line 294: we do not care about the legacy gateway at all

backend/src/ee/services/pam/pam-enums.ts
```diff
@@ -101,3 +101,14 @@ export enum PamNotificationEvent {
...
+// Informational conditions surfaced on an account. Unlike PamAccountAccessibilityIssue these gate
+// nothing: the account launches, records and rotates as normal.
+export enum PamAccountWarning {
+  SessionLogMaskingDegraded = "session-log-masking-degraded"
+}
+
+// These are brokered over RDP, whose recordings are base64 frames rather than terminal text, so
+// the gateway never masks them.
+export const accountTypeSupportsSessionLogMasking = (accountType: PamAccountType): boolean =>
+  accountType !== PamAccountType.Windows && accountType !== PamAccountType.WindowsAd;
```
sheensantoscapadngan: interesting. so we already have session log masking for everything else except these? how about AWS?

backend/src/ee/services/pam-account-template/pam-account-template-schemas.ts
```diff
@@ -36,14 +36,17 @@ export const PamTemplateSettingsSchema = z.object({
...
   heartbeat: PamHeartbeatConfigSchema.optional(),
-  sessionLogMaskingPatterns: z.string().optional()
+  sessionLogMaskingPatterns: z.string().optional(),
+  sessionLogMaskingBuiltInDetection: z.boolean().default(false)
 });
 
 export const PamTemplateSettingsInputSchema = PamTemplateSettingsSchema.extend({
-  sessionLogMaskingPatterns: patternsStringSchema().optional()
+  sessionLogMaskingPatterns: patternsStringSchema().optional(),
+  sessionLogMaskingBuiltInDetection: z.boolean().optional()
```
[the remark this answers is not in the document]
sheensantoscapadngan: This is an issue with our API contract because, even though it's a patch endpoint, we do a full replace of the object.  Will not resolve that contract issue here.

frontend/src/pages/pam/PamTemplatesPage/components/TemplateDetailSheet.tsx
```diff
@@ -947,31 +961,64 @@ const SettingsTab = ({
...
+                  name="settings.sessionLogMaskingBuiltInDetection"
+                  render={({ field }) => (
+                    <Field orientation="horizontal" className="items-center!">
+                      <FieldContent>
+                        <FieldTitle>
+                          Built-in Detection
+                          <Badge variant="success">Recommended</Badge>
+                        </FieldTitle>
+                        <FieldDescription>
+                          API keys, tokens, private keys, and high-entropy values are detected and
```
lb-vn, line 984: we should drop private keys from this copy, or fix them. the gateway masks one rendered terminal line at a time and the private-key rule needs the whole BEGIN/END block in one fragment, so cat id_rsa over ssh still lands in the recording unmasked.

backend/src/ee/services/pam-account/pam-account-service.ts
```diff
@@ -238,6 +244,68 @@ export const pamAccountServiceFactory = (deps: TPamAccountServiceFactoryDep) =>
...
+          poolId,
+          healthy.some((gateway) => !supportsMasking(gateway))
+        );
+      })
+    );
+
+    const directIds = new Set(candidates.filter((c) => c.enabled && c.gatewayId).map((c) => c.gatewayId!));
+    const unsupported = new Set(
+      directIds.size
+        ? (await gatewayV2DAL.find({ orgId }))
```
lb-vn, line 296: we should scope this with $in on the direct ids rather than pulling every gateway in the org on each account list

backend/src/ee/services/pam-account-template/pam-account-template-schemas.ts
```diff
@@ -36,14 +36,17 @@ export const PamTemplateSettingsSchema = z.object({
   passwordRequirements: PasswordRequirementsSchema.optional(),
   rotation: PamRotationConfigSchema.optional(),
   heartbeat: PamHeartbeatConfigSchema.optional(),
-  sessionLogMaskingPatterns: z.string().optional()
+  sessionLogMaskingPatterns: z.string().optional(),
+  sessionLogMaskingBuiltInDetection: z.boolean().default(false)
```
lb-vn: update replaces settings wholesale, so an api caller that patches settings without this key silently turns masking off. we should merge on update instead of letting an omitted key mean off.

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
```diff
@@ -2,7 +2,7 @@ import z from "zod";
 
 import { PamAccountsSchema } from "@app/db/schemas";
 import { EventType } from "@app/ee/services/audit-log/audit-log-types";
```
lb-vn, line 4: ignore placement

Shouldn't we enable this for all templates by default?

frontend/src/hooks/api/pam/enums.ts
```diff
@@ -39,6 +39,25 @@ export const ROTATABLE_PAM_ACCOUNT_TYPES = [
...
+}
+
+// Windows is brokered over RDP, whose recordings the gateway never masks; AWS IAM is gateway-less
+// and produces no session log. Mirrors accountTypeSupportsSessionLogMasking in
+// backend/src/ee/services/pam/pam-enums.ts.
+export const UNMASKABLE_PAM_ACCOUNT_TYPES = [
+  PamAccountType.Windows,
+  PamAccountType.WindowsAd,
+  PamAccountType.AwsIam
+];
```
lb-vn, line 56: I'm curious if keeping a negative-style list is the right call. Instead of opt-in. What if an account type is added but we forget to put it into this array, what will happen? Or would tests catch the issue

lb-vn (changes requested):

frontend/src/hooks/api/pam/enums.ts
sheensantoscapadngan: I think in this case it makes sense because session log masking is only really invalid for very few account types meaning for most additions, they shouldn't have to care about this at all

backend/src/ee/routes/v1/pam-routers/pam-account-router.ts
sheensantoscapadngan: wdym? all newly created templates will have this built-in masking enabled by default. for existing ones, we keep them as it is since it might break user expectations for the accounts they're already using. it's similar to the idea of not automatically pre-creating templates when we add new account types

lb-vn: Makes sense

frontend/src/hooks/api/pam/enums.ts
lb-vn: Sounds good, we can keep it

frontend/src/pages/pam/PamTemplatesPage/components/TemplateDetailSheet.tsx
sheensantoscapadngan: Dropped it from the list here since we're not able to mask arbitrary private keys, but if the auth method was public key for SSH, then we're able to handle it appropriately in the latest patch that I pushed out. 

image

lb-vn: Looks good to me!

lb-vn (approved):

PR #8199, opened 2026-09-17
fix(pki): carry non-text custom extensions through issuance and renewal

carlosmonastyrski:
## Context

Every publicly trusted certificate carries Certificate Transparency timestamps, and we were copying them into renewals, which made every imported cert impossible to renew.

We now leave those to the issuing CA. Custom extension values can also be sent as base64 DER, not just text, so must-staple and qcStatements survive a renewal instead of blocking it.

**How to test**

Import any cert from a public site into an internal-CA profile and renew it. It should issue, keeping the original's extensions minus the CT ones.

Then issue a cert with a custom extension whose value isn't text: OID `1.3.6.1.5.5.7.1.24`, format DER, value `MAMCAQU=`. `openssl x509 -text` should print `TLS Feature: status_request`, which only happens if those bytes landed untouched. If it had been treated as text you'd see the raw base64 string instead.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


carlosmonastyrski:
@greptile re-review and update your Confidence Score above

carlosmonastyrski:
@greptile re-review and update your Confidence Score above

PR #8202, opened 2026-09-17
feat(audit-log): normalize free text in audit events

Vligai:
## Context

Audit events carry user-supplied free text: display names, PAM access reasons, `User-Agent`, and the client IP. All of it is stored and forwarded verbatim. The values are legal JSON, so well-formed output says nothing about how they render in a terminal, in a CSV export, or in a SIEM that splits on newlines.

Two changes, one in the ingest path and one at the API boundary.

**Normalize on the way in.** `lib/validator/log-safe-text` strips C0/C1 controls, ANSI escape sequences, bidirectional overrides, the zero-width space and the BOM. It runs in `enrichEntry` (`audit-log-queue.ts`), the one value handed to both the batch insert and the stream outbox, so persistence and forwarding are covered in the same place.

**Bound at the API boundary.** `auditSafeText` rejects the same set on `POST` / `PATCH /v1/identities` (`name`) and on the three PAM `reason` fields. A NUL byte previously reached the storage layer and faulted with a 500; it now returns a 422 naming the field. Reasons are multi-line inputs, so they accept LF, CR and tab. Names do not.

Metadata keys are normalized alongside values. A collision is suffixed so one field cannot overwrite another, and nesting past the recursion cap becomes a marker, so truncation stays visible in the record.

### What this does not cover

The sink-side normalization is not exercised against a running server. `buildStreamEntry` drops every entry at push time when `plan.auditLogsRetentionDays` is falsy, the default without an audit-log entitlement, so no rows are written and `GET /v1/organization/audit-logs` returns `[]`. The unit tests cover the sanitizer directly. What stays unverified live is the wiring in `enrichEntry`, which needs an entitled org or the e2e suite's license fake.

`ipAddress` and `userAgent` are narrower in practice than they look. Node rejects NUL and ESC in header values outright, so only the multi-byte invisibles reach those two fields.

`name` on `/v1/identities` is an existing field gaining validation. Any caller currently sending control characters in it will start getting a 422.

## Steps to verify the change

```bash
cd backend
npm run type:check
npx vitest run --config vitest.unit.config.mts src/lib/validator/log-safe-text.test.ts
```

Against a local dev stack, confirm the boundary rejects what it should:

```bash
TOKEN=<admin jwt>; ORG=<org uuid>
for payload in 'probename' 'probe\nname' '[31mprobe' 'probe‮name' 'ad​min' 'probe] tail'; do
  printf '%-22s ' "$payload"
  curl -s -o /dev/null -w '%{http_code}\n' -X POST http://localhost:4000/api/v1/identities \
    -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
    --data-binary "{\"name\":\"$payload\",\"organizationId\":\"$ORG\"}"
done
```

Multi-line reasons reach the handler instead of failing validation. A 422 mentioning control characters means the rule fired; any other response means the field passed:

```bash
curl -s -X POST http://localhost:4000/api/v1/pam/access-requests \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  --data-binary '{"accountId":"<uuid>","reason":"line one\r\nline two","duration":"1h","accessType":"session"}'
```

### Verified locally

- Type check and lint clean; 27 tests in `log-safe-text.test.ts`, 3446 in the full unit suite, no failures
- Six malformed identity names return 422 with the field named, including NUL, which previously returned 500
- Emoji ZWJ sequences and Persian ZWNJ return 200, and the joiners are still present when the value is read back
- Zero-width space, RTL override and unterminated OSC still return 422
- LF, CRLF and tab in a PAM reason pass the character rule; NUL, ANSI and ZWSP in the same field do not

Not verified: the sink-side path, for the entitlement reason above, and ClickHouse insertion, which the local stack does not run.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


Vligai:
@greptileai review again

PR #8210, opened 2026-09-18
improvement(secret-sync): surface Include All Subfolders in the UI and docs

lobocv:
## Context

Follow up to #8087, which added recursive secret syncs. That PR shipped the behavior but left no way to tell a recursive sync apart from a normal one after it was created, and the option was undocumented.

The fix was to add a small badge beside the source folder on the individual secret sync in the UI. See screenshots

CleanShot 2026-09-18 at 11 03 42 AM@2x

There is also a new section in the API docs that explain the new recursive option
CleanShot 2026-09-18 at 11 05 07 AM@2x


andrewhuhh (approved):
lgtm

PR #8213, opened 2026-09-18
feat: block certificate deletion until expiry

carlosmonastyrski:
## Context

Deleting a certificate removes its history, its place in the inventory, and any expiry alerts attached to it, but it does not revoke anything: the certificate stays valid wherever it is deployed until it expires. That made deletion an easy way to lose visibility of a certificate that is still live.

Deletion is now available only once a certificate has expired, so revocation is the way to retire one early. Certificates that were discovered or imported can still be deleted at any time, since they are inventory records rather than certificates Infisical issued.

Revoked certificates also wait until expiry, because removing one early would drop it from the certificate revocation list and clients would start trusting it again.

The delete action stays visible in the UI and explains why it is unavailable, and the audit event now records which certificate was deleted and why it was eligible.

## Type

- [ ] Fix
- [x] Feature
- [ ] Improvement
- [ ] Breaking
- [x] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -1841,7 +1842,7 @@ export const registerCertificateRouter = async (server: FastifyZodProvider) => {
       hide: false,
       operationId: "deleteCertificate",
       tags: [ApiDocsTags.PkiCertificates],
-      description: "Delete certificate",
+      description: "Delete certificate. Only expired, discovered and imported certificates can be deleted.",
```
sheensantoscapadngan: there should be a comma between "discovered" and "and" here right?

backend/src/server/routes/v1/deprecated-certificate-router.ts
```diff
@@ -539,7 +540,7 @@ export const registerDeprecatedCertRouter = async (server: FastifyZodProvider) =
     schema: {
       hide: false,
       tags: [ApiDocsTags.PkiCertificates],
-      description: "Delete certificate",
+      description: "Delete certificate. Only expired, discovered and imported certificates can be deleted.",
```
sheensantoscapadngan: same remark here about comma

backend/src/server/routes/v1/certificate-router.ts
```diff
@@ -3,6 +3,7 @@ import RE2 from "re2";
 import { z } from "zod";
 
 import { EventType } from "@app/ee/services/audit-log/audit-log-types";
+import { CertificateSource } from "@app/ee/services/pki-discovery/pki-discovery-types";
```
sheensantoscapadngan: I don't think we have to explicitly instruct folks here to revoke it. I feel like we should just explain to them why the specific certificate can't be deleted...

Image

backend/src/server/routes/v1/certificate-router.ts
sheensantoscapadngan: also "or" might be more appropriate here?

sheensantoscapadngan (approved):

PR #8214, opened 2026-09-18
improvement(sso): stop duplicating surnames when OIDC omits given_name

victorvhs017:
## Context

When an IdP omitted `given_name`, OIDC login stored the composite `name` claim as `firstName` and still took `lastName` from `family_name`. The UI concatenates those fields, so users rendered as `Jane Doe Doe`. LDAP had the same pairing of `cn` and `sn`. Only some users were affected — the ones whose IdP sent a full name in `name` and no structured given name.

OIDC and LDAP now share `resolveAssertedProfileName`. Structured `given_name` / `givenName` wins. If that is missing, a trailing `family_name` / `sn` is stripped from `name` / `cn` instead of being appended. A family-name-first locale or a suffix after the surname is left whole rather than guessed at. Orgs that enforce SSO already re-assert the IdP profile on login, so those users correct on the next sign-in. Docs now state that `email` plus `given_name` or `name` is required, and that `family_name` is only needed when Infisical should store a separate last name.

## Steps to verify the change

1. In an org with OIDC (or LDAP) configured, log in as a user whose IdP sends `name=Jane Doe` and `family_name=Doe` with no `given_name`. The stored name should be `Jane` / `Doe`, not `Jane Doe` / `Doe`.
2. Repeat with `given_name` and `family_name` both present — mapping should stay `given_name` / `family_name` unchanged.
3. If the org enforces SSO, an already-duplicated profile should correct on the next login. If it does not enforce SSO, the stored name stays as-is.

## Type

- [ ] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


backend/src/ee/routes/v1/ldap-router.ts
```diff
@@ -82,12 +83,18 @@ export const registerLdapRouter = async (server: FastifyZodProvider) => {
               ? (fastifyReq.body as { callbackPort?: number })?.callbackPort
               : undefined;
 
+          const assertedName = resolveAssertedProfileName({
```
Thiago-AS, line 86: on the LDAP side, the name attributes can come back as arrays when a directory stores more than one value for them, and the new name resolver chokes on that and fails the login. Before this change those users could still sign in, just with a messy name, so it's a small regression. Intended?

victorvhs017: good catch, updated to handle arrays!

Thiago-AS (approved):

PR #8221, opened 2026-09-19
fix(build): align Node and npm installation toolchains

andrewhuhh:
## Context

Refs [ENG-5697](https://linear.app/infisical/issue/ENG-5697/standardize-nodenpm-and-dependency-installation-safeguards).

Standardize builds on Node 22.23.2 and npm 11.19.1. Previously, standalone/backend images installed npm 10.9.9 despite requiring npm >=11.10.0, CI left npm implicit, and several Docker install stages omitted the existing `.npmrc`.

- Add canonical Node/npm pins in `build-versions.env`, matching Docker defaults and `.nvmrc`, and a shared CI setup action. Reconcile all six package roots with `engines`, early `devEngines` enforcement, and `engine-strict`. Restore npm download caches only after installing the pinned npm.
- Add a dependency-free runtime/drift check and local-registry regression tests. Copy `.npmrc` before container installs, use locked `npm ci` in development/CI, and preserve backend lifecycle/native builds. Remove redundant unpinned global `tsconfig-paths` installs; it is already a backend dependency.
- Document contributor setup and the safeguard boundary: `min-release-age=7` filters dependency resolution, but `npm ci` trusts locked versions and does not recheck publication dates. Tests demonstrate both behaviors.

No dependency versions change: all six lockfile diffs contain only root engine metadata. React, TypeScript, Zod, and unrelated upgrades are excluded.

**Omnibus boundary:** inspected [Omnibus #6](https://github.com/Infisical/infisical-omnibus/pull/6), which is open and changes only Node/checksums. Omnibus must also install npm 11.19.1 before packaging application revisions from this PR. That repository is not configured for delivery here; this PR neither modifies its branch nor duplicates its separate full-package verification, and does not close ENG-5697's remaining Omnibus scope.

## Steps to verify the change

1. Run `nvm install && nvm use`, `npm install -g npm@11.19.1`, `node scripts/check-node-toolchain.mjs`, and `node --test scripts/node-toolchain.test.mjs`. All 22 policy checks pass, including early mismatch rejection, release-age resolution, locked replay, lifecycle behavior, and drift rejection. Bootstrap from the Node image's bundled npm 10.9.8 to the pinned npm also passes.
2. Run `npm ci` in the root, backend, frontend, e2e, upgrade-impact, and sink/oidc-server directories. All clean installs pass; frontend also passes with `--ignore-scripts`. Backend/frontend production builds, typechecks, and lint pass; all 285 frontend tests pass. Auxiliary tooling typechecks/tests and data validation pass (historical git-range validation is skipped where local release tags are absent).
3. Build all eight affected Dockerfiles on Linux amd64: both standalone images; backend production, dev, FIPS toolchain, and FIPS dev; frontend dev; docs. All full builds pass. For FIPS dev, pass the locally built toolchain as `TOOLCHAIN_IMAGE`. Images report Node 22.23.2, npm 11.19.1, and release age 7.
4. Start backend production/dev/FIPS-dev and standalone/FIPS images against isolated PostgreSQL/Redis. All five return HTTP 200 from `/api/status`. Required production native modules load; bcrypt/argon2/RE2 operations pass, and FIPS activation/SHA256 pass.
5. Run `make lint-docs-branch` with Vale 3.17.1: clean. Changed workflow YAML and composite Bash parse. Actionlint is clean except four unchanged release-workflow findings reproduced on main (`set-output` and nonexistent `steps.version` references).

Existing image caveats, confirmed independently of the npm change: FIPS-dev Oracle thick-client initialization lacks the amd64 `libaio.so.1` compatibility symlink; backend-only production Docker health references a missing `healthcheck.js`, despite successful direct HTTP startup. These unrelated issues are not changed here. CI results are tracked on this PR.

## Type

- [x] Fix
- [ ] Feature
- [x] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [x] Tested locally


package.json
```diff
@@ -27,5 +27,22 @@
...
+  },
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": "^22.23.2",
+      "onFail": "error"
+    },
+    "packageManager": {
+      "name": "npm",
+      "version": "11.19.1",
```
Vligai: Is this intended to be a strict match? Think it will break local if someone moves to latest 11.20.0. Could we use ^ here?

backend/package.json
```diff
@@ -318,5 +319,18 @@
...
+  },
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": "^22.23.2",
+      "onFail": "error"
+    },
+    "packageManager": {
+      "name": "npm",
+      "version": "11.19.1",
```
Vligai: Our deploy pipeline runs npm ci in backend/ on the GitHub runner before the Docker build, and that runner seems to have npm 10.9.8 based on the latest deploy log. This may break that pipeline, I think the infra repo needs a bump to 11.19.1 before this goes out.

frontend/package.json
```diff
@@ -190,5 +190,18 @@
...
+  },
+  "devEngines": {
+    "runtime": {
+      "name": "node",
+      "version": "^22.23.2",
+      "onFail": "error"
+    },
+    "packageManager": {
+      "name": "npm",
+      "version": "11.19.1",
```
Vligai: Ah, same thing with omnibus, it builds backend and frontend with 10.9.8 like the v0.165.13 release bork. So it may need a version bump similar to infra repo.

e2e/.npmrc
```diff
@@ -0,0 +1,2 @@
+min-release-age=7
```
Vligai: Claude.md in this directory has the following clause:
> Don't add a .npmrc here for the 7-day minimum-release-age rule — e2e/ is a test harness, not a runtime artifact. The supply-chain concern that motivates the rule in backend//frontend/ doesn't apply.

It's worth aligning the two.

scripts/node-toolchain.test.mjs
```diff
@@ -0,0 +1,390 @@
...
+  await new Promise((resolve, reject) =>
+    registry.close((error) => (error ? reject(error) : resolve())),
+  );
+});
+
+test("regressions run with the pinned npm version", async (t) => {
+  const cwd = await project(t);
+  const result = await npm(cwd, "--version");
+  succeeded(result);
+  assert.equal(result.stdout.trim(), "11.19.1");
```
Vligai: Could it read them from build-versions.env like the checker does, or could CLAUDE.md list this file with the others to update together?

backend/Dockerfile.dev.fips
```diff
@@ -1,17 +1,22 @@
 ARG TOOLCHAIN_IMAGE=ghcr.io/infisical/backend-fips-toolchain:main
 FROM ${TOOLCHAIN_IMAGE}
 
+ARG NPM_VERSION=11.19.1
+RUN npm install -g npm@${NPM_VERSION} && test "$(npm --version)" = "${NPM_VERSION}"
```
Vligai: nit: this one is different because it starts from the finished tool chain image where the npm_config_min_release_age=7 is already set so if the pin for the NPM versions moves to something that is less than 7 days old, you will get ETARGET error.

Vligai (changes requested):

PR #8224, opened 2026-09-19
fix(agent-vault): expose agent vault routes in the openapi spec

saifsmailbox98:
## Context

Every Agent Vault API reference page renders blank in prod, e.g. https://infisical.com/docs/api-reference/endpoints/agent-vault-sessions/list

`createJsonSchemaTransform` in `backend/src/server/plugins/fastify-zod.ts` defaults a route out of the spec:

```ts
if (typeof schema.hide === "undefined") {
  schema.hide = true;
}
```

So a route only reaches `/api/docs/json` if it explicitly sets `hide: false`. 693 routes across the backend do. None of the Agent Vault ones did, so `app.infisical.com/api/docs/json` serves 1495 paths and zero of them match `agent-vault`. The routes themselves are live in prod (`GET /api/v1/agent-vault/sessions` returns `401 Token missing`, a nonexistent path returns `404`) — they were just never in the spec.

`docs.json` points `openapi` at that live spec, so Mintlify looks up `GET /api/v1/agent-vault/sessions`, finds nothing, and renders an empty page. Same for all 24 Agent Vault pages, not just sessions.

Adds `hide: false` to the 24 routes that have a docs page, matching the placement convention in `relay-router.ts`:

- `agent-vault-access-bundle-router.ts` — 11 routes (access bundles, services, bundle members)
- `agent-vault-proxy-router.ts` — 6 routes
- `agent-vault-membership-router.ts` — 4 routes
- `agent-vault-session-router.ts` — 3 routes

Left alone: `agent-vault-project-router.ts` (already explicitly `hide: true`) and the three `agent-vault-proxy-agent-router.ts` routes (`/proxy/login`, `/proxy/heartbeat`, `/proxy/resolve`) — proxy-to-server endpoints with no docs page.

Note the spec is built once per process and cached for the pod's lifetime (`swagger.ts`, a `Map` that is never invalidated), so this only takes effect after the backend redeploys.

## Type

- [x] Fix
- [ ] Feature
- [ ] Improvement
- [ ] Breaking
- [ ] Docs
- [ ] Chore

## Checklist

- [ ] Tested locally


scott-ray-wilson (approved):
lgtm

